PR #13, opened 2023-09-13
[E2B-318] - Create env

jakubno:
- Handle creating env vars
- Changed the database to [sqlboiler](https://github.com/volatiletech/sqlboiler)

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -134,16 +134,33 @@ resource "google_storage_bucket" "e2b-envs-docker-context" {
   uniform_bucket_level_access = true
 }
 
+resource "google_service_account" "e2b-api-service-account" {
+  account_id   = "e2b-api-service-account"
+  display_name = "E2B API Service Account"
+}
+
+
+resource "google_project_iam_member" "e2b-api-service-account-storage-permissions" {
+  project = var.gcp_project_id
+  role    = "roles/storage.editor"
+  member  = "serviceAccount:${google_service_account.e2b-api-service-account.email}"
+}
+
+resource "google_service_account_key" "e2b-api-service-account-key" {
+  service_account_id = google_service_account.e2b-api-service-account.name
+}
+
 module "api" {
   source = "./packages/api"
 
   gcp_zone = var.gcp_zone
 
-  logs_proxy_address = "http://${module.cluster.logs_proxy_ip}"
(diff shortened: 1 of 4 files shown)


packages/api/internal/db/client.go
```diff
@@ -1,42 +1,36 @@
...
 var (
-	supabaseURL = os.Getenv("SUPABASE_URL")
-	supabaseKey = os.Getenv("SUPABASE_KEY")
+	databaseURL = os.Getenv("DATABASE_CONNECTION_STRING")
 )
 
 func NewClient() (*DB, error) {
-	// The /rest/v1/ at the end of url is required
-	parsedURL, err := url.Parse(supabaseURL)
+	fmt.Println("databaseURL", databaseURL)
```
jakubno:
```suggestion
```

packages/api/.env.template
```diff
@@ -0,0 +1,10 @@
+NOMAD_TOKEN=
+CONSUL_TOKEN=
+POSTHOG_API_KEY=
+DATABASE_CONNECTION_STRING=
+
+PSQL_PASS=
```
jakubno, line 8:
```suggestion
# DEV envs - generating db models
PSQL_PASS=
```

spec/openapi.yml
```diff
@@ -196,10 +180,25 @@ paths:
...
+                dockerfile:
+                  type: string
+                  description: Dockerfile content
+                envID:
+                  type: string
+                  description: ID of the environment, only present if the environment should be rebuilt
             encoding:
               buildContext:
-                contentType: application/gzip
+                contentType: application/octet-stream
```
jakubno, line 200: not sure about this yet, there's issue with openapi3filter middleware. If I send it through postman, it will automatically assign `application/gzip` encoding, but then it will fail in the middleware, because the middleware wants to parse it as a value

spec/openapi.yml
```diff
@@ -196,10 +180,25 @@ paths:
         content:
           multipart/form-data:
             schema:
-              $ref: "#/components/schemas/EnvironmentToBuild"
```
jakubno, line 199: openapi3filter middleware can work with multipart only if the schema type is `object` , it can't be ref

PR #24, opened 2023-10-23
Push docker image

jakubno:


packages/api/internal/nomad/env.go
```diff
@@ -109,7 +112,7 @@ func (n *NomadClient) BuildEnvJob(
 		return fmt.Errorf("error waiting for env '%s' build: %w", envID, finishErr.Err)
 	}
 
-	delErr := n.DeleteEnvBuild(*job.ID, false)
+	delErr := n.DeleteEnvBuild(*job.ID, true)
```
jakubno, line 115: Is this intended? @ValentaTomas

ValentaTomas: Yes, this should delete the Job from Nomad only after it *successfully* completes. If there was error the job will stay.

PR #40, opened 2023-11-27
Manage start command in envd

ValentaTomas:
Right now the start cmd is managed by systemd. This PR changed it so it is managed by envd which allows us better control over it and also allows integration with logging.

Tasks
- [ ] Handle/communicate that the start cmd will be started during build without internet now
- [x] Modify old start cmd log handlers in SDKs
- [x] Permissions bug - use root for the start command
- [x] Fix start cmd not starting jupyter server (waiting...)

### DIFF
--- a/packages/envd/internal/process/service.go
+++ b/packages/envd/internal/process/service.go
@@ -112,7 +112,7 @@ func (s *Service) scanRunCmdOut(pipe io.Reader, t output.OutType, process *Proce
 	}
 }
 
-func (s *Service) Start(id ID, cmd string, envVars *map[string]string, rootdir string) (ID, error) {
+func StartWithUser(s *Service, id ID, cmd string, envVars *map[string]string, rootdir string, user string) (ID, error) {
 	s.logger.Infow("Start process",
 		"processID", id,
 		"cmd", cmd,
@@ -133,7 +133,7 @@ func (s *Service) Start(id ID, cmd string, envVars *map[string]string, rootdir s
 			id = xid.New().String()
 		}
 
-		newProc, err := s.processes.Add(id, s.env.Shell, cmd, envVars, rootdir)
+		newProc, err := s.processes.Add(id, s.env.Shell, cmd, envVars, rootdir, user)
 		if err != nil {
 			s.logger.Errorw("Failed to create new process",
 				"processID", id,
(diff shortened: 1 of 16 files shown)


packages/env-build-task-driver/internal/env/rootfs.go
ValentaTomas, line 1: Not sure if this properly escapes the start cmd.

ValentaTomas:
We may want to rethink the proper way to handle start command — even when we move the management to envd there are questions about storing start command logs and how to handle output subscriptions/reconnecting.

packages/envd/internal/process/service.go
```diff
@@ -112,7 +112,7 @@ func (s *Service) scanRunCmdOut(pipe io.Reader, t output.OutType, process *Proce
 	}
 }
 
-func (s *Service) Start(id ID, cmd string, envVars *map[string]string, rootdir string) (ID, error) {
+func StartWithUser(s *Service, id ID, cmd string, envVars *map[string]string, rootdir string, user string) (ID, error) {
```
jakubno, line 115: Why not have it on `Service` and have it as an "independent" function?

jakubno:
Jupyter + info about starting the sandbox are exported
image

packages/envd/main.go
```diff
@@ -202,9 +188,23 @@ func main() {
 	// The /file route used for downloading and uploading files via SDK.
 	router.HandleFunc("/file", fileHandler)
 
+	// TODO: Run the start command as root
```
jakubno: we are doing it already

packages/envd/main.go
```diff
@@ -202,9 +188,23 @@ func main() {
...
+	// TODO: How to differentiate the start command from other running commands? Metadata? EnvVars?
```
jakubno, line 206: why do we need that?

packages/envd/internal/process/service.go
ValentaTomas: I think by default all receiver methods are exposed as JsonRpc services, so I didn't want to expose two services.

If we have the startWithUser starting with lowercase it might be ok, but I haven't tested it.

This will also be different in the new rpc.

packages/envd/main.go
ValentaTomas: Yeah, I think this is outdated.

packages/envd/main.go
ValentaTomas: I was thinking that sometimes you might want to interact with the start command from the SDK, but how can you tell that a process is started as a start command?
I have a way to do this now, don't worry.

ValentaTomas:
> Jupyter + info about starting the sandbox are exported image

What is the format of results/stdout/stderr from the jupyter in the logs?

PR #50, opened 2024-01-18
Improve docker security

jakubno:


packages/env-build-task-driver/internal/env/rootfs.go
```diff
@@ -82,7 +83,25 @@ func NewRootfs(ctx context.Context, tracer trace.Tracer, env *Env, docker *clien
 		env:          env,
 	}
 
-	err = rootfs.buildDockerImage(childCtx, tracer)
+	network, err := rootfs.client.NetworkCreate(childCtx, uuid.New().String(), types.NetworkCreate{})
```
ValentaTomas: Is using the build ID as the network name (or something derived from the build ID) possible here?

PR #60, opened 2024-01-25
Update kernel version

ValentaTomas:
- Create system for migrating from old kernel or allow using different versions of kernels.


packages/shared/migrations/20240202120312.sql
```diff
@@ -0,0 +1,2 @@
+-- Modify "envs" table
+ALTER TABLE "public"."envs" ADD COLUMN "kernel_version" character varying NOT NULL DEFAULT 'vmlinux-5.10.186';
```
jakubno, line 2: Change to only use `vmlinux-5.10.186` and use the new one as default

PR #61, opened 2024-01-25
Update FC

ValentaTomas:
Update FC and regenerate FC API.

- [ ] Check the performance


ValentaTomas:
We are postponing the update.

Issue #55, opened 2024-01-24
Allow network access in start command during custom sandbox building


mlejva:

ValentaTomas:
Duplicate of https://github.com/e2b-dev/E2B/issues/264

PR #64, opened 2024-02-01
Enable internet access during build

ValentaTomas:
During build the startup command does not have access to internet.

- https://github.com/e2b-dev/E2B/issues/264
- https://github.com/e2b-dev/infra/issues/55

What needs to be done:
- [ ] Use shared code for handling FC sockets inside build and instance driver
  - Check how the socket is created in both drivers and unify this
- [ ] Acquire ip slot inside build drive. The ID will be based on the acquired slot not buildID
- [ ] Ensure that the `/etc/hosts` is properly managed now that two drivers will be using it (Should we use dnsmasq, envoy proxy, etc., instead? Should the slot acquisition and local sandbox dns setup be handled by API instead?)
- [ ] Use the networking from instance driver moved to shared for FC networking in build driver - this will setup internet access


ValentaTomas:
We may be reworking how the start command works so changes to the networking may get outdated.

PR #67, opened 2024-02-07
Add endpoint for killing sandboxes

ValentaTomas:
- [ ] Decide if the naming and ergonomy is right


packages/api/internal/handlers/sandbox_kill.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+
+	deleteErr := a.nomad.DeleteInstance(sandboxID, true)
+	if deleteErr != nil {
+		errMsg := fmt.Errorf("error when killing sandbox: %s", deleteErr.ClientMsg)
+		a.sendAPIStoreError(c, deleteErr.Code, fmt.Sprintf("Error killing sandbox - sandbox '%s' was not found", sandboxID))
+		telemetry.ReportCriticalError(ctx, errMsg)
+
+		return
+	}
```
jakubno: It's enough to call `Delete` on `instanceCache`, this way the sandbox will stay in cache

jakubno (approved):

PR #81, opened 2024-02-19
Parametrize Firecracker version

ValentaTomas:
Parametrize Firecracker so each template can use a different Firecracker binary.

- [x] Add FCs versions downloading, building and uploading to bucket
- [x] Mount the bucket with FC versions
- [x] Copy the FCs versions to the specified path in root disk


packages/fc-versions/build.sh
```diff
@@ -0,0 +1,33 @@
...
+  # The format will be: latest_tag_latest_commit_hash — v1.7.0-dev_g8bb88311
+  version_name=$(git describe --tags --abbrev=0 $(git rev-parse HEAD))_$(git rev-parse --short HEAD)
+  echo "Version name: $version_name"
+
+  echo "Building Firecracker version: $version_name"
+  tools/devtool -y build --release
+
+  echo "Copying finished build to builds directory"
+  mkdir -p "../builds/${version_name}"
+  cp build/cargo_target/x86_64-unknown-linux-musl/release/firecracker "../builds/${version_name}/firecracker"
```
jakubno, line 21: not the nicest path, but it looks it should stay the same

packages/fc-versions/firecracker_versions.txt
```diff
@@ -0,0 +1,2 @@
+v1.5.0
+8bb883111ff608dc2ca1a08d5e4eb2ddbd5c1c8b
```
jakubno, line 2: it would be nice to have some comments to it, otherwise it's hard to read

jakubno (approved):
Lgtm

packages/fc-versions/firecracker_versions.txt
ValentaTomas: I added a readme where I explained the whole package together with the versioning system.

PR #84, opened 2024-02-20
Enable huge pages to improve sandboxes performance

ValentaTomas:
- [x] Enable huge pages on host
- [x] Reconfigure FC to use huge pages


ValentaTomas:
This PR is rebased against `parametrize-firecracker-version` because it is directly depending on the changes made there.

PR #86, opened 2024-02-27
Cleanup kernel building

ValentaTomas:
Cleanup kernel building so we can later update kernel easily.


jakubno (commented):
Code lgtm
we need to test it and let's go

ValentaTomas:
Made separate issue from updating kernel to 6.1. This PR just cleanup up kernel build so we can add it later.

PR #91, opened 2024-03-05
Add API endpoint for retrieving sandbox logs

ValentaTomas:
Remaining tasks:

- [x] Loki service config
  - [x] Bucket persistence config
- [x] Handle connection between services (logs collector to loki, loki to API)
- [x] Redirecting logs to the Loki instance
- [x] Query the Loki instance
  - [x] Pass the env vars and initialize client in API
  - [x] Sanitize user provided info
  - [x] Handle offset/range


ValentaTomas:
The problems with service interconnectivity (collector-loki-api) may be the `icc: false` setup that prevent container networking on the default network.

packages/nomad/otel-collector.hcl
```diff
@@ -171,7 +171,7 @@ exporters:
 service:
   telemetry:
     logs:
-      level: debug
+      level: warn
```
jakubno, line 174: What exactly are those?

packages/buckets/main.tf
```diff
@@ -1,3 +1,14 @@
+resource "google_storage_bucket" "loki_storage_bucket" {
+  name     = "${var.gcp_project_id}-loki-storage"
+  location = var.gcp_region
+
+  public_access_prevention    = "enforced"
+  storage_class               = "STANDARD"
```
jakubno, line 6: Would you add a lifecycle policy?

spec/openapi.yml
```diff
@@ -286,6 +310,39 @@ paths:
...
 
+  /sandboxes/{sandboxID}/logs:
+    get:
+      description: Get sandbox logs
+      tags: [sandboxes]
+      security:
+        - ApiKeyAuth: []
+      parameters:
+        - $ref: "#/components/parameters/sandboxID"
+        - in: query
```
jakubno, line 321: can we also add end? We may not use it right now, but it IMHO makes a lot of sense, especially if the sandbox ran for several hours, for sharing specific logs with colleagues etc.

packages/api/internal/handlers/sandbox_logs.go
```diff
@@ -0,0 +1,94 @@
...
+	end := time.Now()
+
+	if params.Start != nil {
+		start = time.UnixMilli(int64(*params.Start))
+	} else {
+		start = end.Add(-oldestLogsLimit)
+	}
+
+	fmt.Printf("Start: %v\n End: %v\n", start, end)
+
```
jakubno:
```suggestion
```

packages/api/internal/handlers/sandbox_logs.go
```diff
@@ -0,0 +1,94 @@
...
+	// Sanitize ID
+	// https://grafana.com/blog/2021/01/05/how-to-escape-special-characters-with-lokis-logql/
+	id := strings.ReplaceAll(sandboxID, "`", "")
```
jakubno, line 53: is there a possibility for some "loki injections"? We can make sure there are only letters and numbers and then we don't even need backticks

packages/api/internal/handlers/sandbox_logs.go
```diff
@@ -0,0 +1,94 @@
...
+	query := fmt.Sprintf("{source=\"logs-collector\", service=\"envd\", teamID=`%s`, sandboxID=`%s`}", teamID.String(), id)
+
+	// TODO: Check if the nanoseconds conversion is correct
```
jakubno: is it?

packages/api/internal/handlers/sandbox_logs.go
```diff
@@ -0,0 +1,94 @@
...
+	oldestLogsLimit = 168 * time.Hour // 7 days
+)
+
+func (a *APIStore) GetSandboxesSandboxIDLogs(
+	c *gin.Context,
+	sandboxID string,
+	params api.GetSandboxesSandboxIDLogsParams,
+) {
+	ctx := c.Request.Context()
+	sandboxID = utils.ShortID(sandboxID)
```
jakubno, line 32: ~shouldn't we check if it belongs to the team? But I wouldn't do it in this PR~

it's handled by team.id in query ✔️

packages/nomad/loki.hcl
```diff
@@ -0,0 +1,153 @@
...
+
+variable "loki_bucket_name" {
+  type = string
+}
+
+job "loki" {
+  datacenters = [var.gcp_zone]
+  type        = "service"
+
+  priority = 75
```
jakubno, line 21: Not entirely related, but based on what you choose the priority?

packages/api/internal/handlers/sandbox_logs.go
ValentaTomas: Backticks are used as string literals in LogQL to escape everything. I could have escaped the backtick to prevent injection but I just removed it here.

spec/openapi.yml
ValentaTomas: Right now we are not exposing the `start` in CLI too. I thought about adding this, but let's not invest more work until necessary — we can add the flags and `end` parameter whenever we want without breaking anything.

Let's see how useful this is first.

packages/api/internal/handlers/sandbox_logs.go
ValentaTomas: Yes

packages/nomad/loki.hcl
ValentaTomas: I tried to prioritize the services by:
1. Us knowing what is happening (otel-collector -> logs-collector)
2. Services necessary for sandboxes to be running, in the order in which they are interacted with (API -> proxies, docker-reverse-proxy?)
3. Services necessary for users to know what is happening (logs-collector -> loki)

ValentaTomas: I think there are some hidden deps (like logs-collector healthchecking loki), that are not accounted for right now.

packages/buckets/main.tf
ValentaTomas: Yes, we want to add this — probably deletion after bucket objects being 192h old.

jakubno (approved):

PR #100, opened 2024-03-13
Add temporary limit for concurrent new sandbox requests

ValentaTomas:
We are improving handling of creating sandboxes during spikes and we are adding a temporal limit to the maximum concurrent requests for new sandboxes.
This will result in smoother handling of traffic instead of blocking the whole system.

We will remove this limit after improving how we handle the spikes.

- [x] Test


packages/api/internal/handlers/sandbox_create.go
```diff
@@ -89,6 +95,18 @@ func (a *APIStore) PostSandboxesWithoutResponse(c *gin.Context, ctx context.Cont
 		attribute.String("env.firecracker.version", firecrackerVersion),
 	)
 
+	telemetry.ReportEvent(ctx, "waiting for parallel lock")
+	limitErr := postSandboxParallelLimit.Acquire(ctx, 1)
+	if limitErr != nil {
+		errMsg := fmt.Errorf("error when acquiring parallel lock: %w", limitErr)
+		telemetry.ReportCriticalError(ctx, errMsg)
+
+		a.sendAPIStoreError(c, http.StatusInternalServerError, "Too many requests")
```
jakubno: The error message isn't probably true, acquire returns only error when context is cancelled

jakubno (approved):
Let's test and merge

packages/api/internal/handlers/sandbox_create.go
ValentaTomas: 👀 Not sure if the error message is then overriden by the gin handled. Will check this.

jakubno (approved):

PR #104, opened 2024-03-15
Don't cleanup nomad logs on restart

ValentaTomas:
Nomad logs are cleaned when the instance is restarted which makes debugging unexpected restarts harder.

- [ ] Test


ValentaTomas:
We are handling this differently

PR #112, opened 2024-03-26
Cleanup Nomad job handling and drivers

ValentaTomas:
The sudden restarts we are sometimes experiencing may be connected to how we are using Nomad. This PR should improve that.

- [x] Check that incomplete cleanup and 502 errors is not caused by this change


jakubno (approved):

ValentaTomas:
I need to finish checking some possibel problems with deleting sandboxes here.

jakubno (approved):

Issue #123, opened 2024-04-05
Logs for some sandboxes are missing


ValentaTomas:
Logs for some sandboxes are missing both in Grafana and in our internal Loki when testing this on staging.
The problem should be either in our `logs-collector` (Vector) setup or in the `envd` in the logs pushing that involves MMDS data (we can test this inside sandboxes).

ValentaTomas:
Fixed by https://github.com/e2b-dev/infra/pull/124

PR #124, opened 2024-04-05
Fix missing MMDS in some sandboxes

ValentaTomas:
This moves MMDS setup from a goroutine — errors in MMDS setup will also cancel the sandbox spawn.

Fixes https://github.com/e2b-dev/infra/issues/123


jakubno (approved):

ValentaTomas:
The cause of the problem was using the childCtx that gets closed after the function end.
Getting the MMDS out of the goroutine have very little time cost and ensures it gets set.

PR #125, opened 2024-04-08
Use observable counter for measuring running sandboxes in monitoring

ValentaTomas:
This PR should improve the reliability of reporting of running sandboxes.

Tasks
- [ ] Test


ValentaTomas:
Observable up-down counter and gauge are both visible, but they are not correctly reporting closing sandboxes.
This could be caused by the cumulative setup, but I'm not sure.

ValentaTomas:
Observable counter does not work well with Grafana. Closing.

PR #126, opened 2024-04-08
Add separate orchestrator for sandboxes

ValentaTomas:
Right now we are using Nomad to schedule and place sandboxes. There are several performance problems with this.
This PR creates a separate service from the orchestrator.

Tasks:
- [x] Fix permission problem with starting permissions
- [x] Improve communication between services
- [x] Update Nomad config to enable deployment on all clients
- [x] Remove hardcoded params
- [x] Improve the deployment (the job is not redeployed when the file changes → hash?)


packages/nomad/orchestrator.hcl
```diff
@@ -0,0 +1,80 @@
...
+        timeout      = "5s"
+        grpc_use_tls = false
+        port         = var.port
+      }
+    }
+
+    task "start" {
+      driver = "raw_exec"
+
+      resources {
```
ValentaTomas, line 75: @jakubno Let's check if the raw exec resources limit the spawned FC resources (should not) and in case they do make it unlimited (https://developer.hashicorp.com/nomad/docs/drivers/raw_exec#resource-isolation).

packages/api/internal/orchestrator/create_instance.go
```diff
@@ -0,0 +1,81 @@
...
+		KernelVersion:      kernelVersion,
+		FirecrackerVersion: firecrackerVersion,
+		Metadata:           string(metadataSerialized),
+		MaxInstanceLength:  int32(maxInstanceLengthHours),
+		HugePages:          features.HasHugePages(),
+	})
+	if err != nil {
+		errMsg := fmt.Errorf("failed to create sandbox of environment '%s': %w", templateID, err)
+
+		return nil, fmt.Errorf("failed to create sandbox of environment '%s': %w", templateID, errMsg)
```
mlejva: I think this error should be wrapped just once

packages/api/internal/orchestrator/create_instance.go
```diff
@@ -0,0 +1,81 @@
...
+			attribute.String("env.id", templateID),
+		),
+	)
+	defer childSpan.End()
+
+	metadataSerialized, err := json.Marshal(metadata)
+	if err != nil {
+		errMsg := fmt.Errorf("failed to marshal metadata: %w", err)
+
+		return nil, fmt.Errorf("failed to marshal metadata: %w", errMsg)
```
mlejva: I think this error should be wrapped just once

packages/api/internal/orchestrator/create_instance.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	telemetry.ReportEvent(childCtx, "Marshalled metadata")
+
+	features, err := sandbox.NewVersionInfo(firecrackerVersion)
+	if err != nil {
+		errMsg := fmt.Errorf("failed to get features for firecracker version '%s': %w", firecrackerVersion, err)
+
+		return nil, fmt.Errorf("failed to get features for firecracker version '%s': %w", firecrackerVersion, errMsg)
```
mlejva: I think this error should be wrapped just once

packages/nomad/orchestrator.hcl
jakubno: There's no resource (cpu/ram) isolation for raw exec

Makefile
```diff
@@ -30,14 +30,15 @@ endif
...
+DESTROY_TARGETS := $(shell terraform state list | grep module | cut -d'.' -f1,2 | grep -v -e "fc_envs_disk" -e "buckets" | uniq | awk '{print "-target=" $$0 ""}' | xargs)
 
 # Login for Packer and Docker (uses gcloud user creds)
 # Login for Terraform (uses application default creds)
 .PHONY: login-gcloud
 login-gcloud:
 	gcloud auth login
 	gcloud config set project "$(GCP_PROJECT_ID)"
-	gcloud --quiet auth configure-docker us-central1-docker.pkg.dev
+	gcloud --quiet auth configure-docker "$(GCP_REFION)-docker.pkg.dev"
```
ValentaTomas: There is typo in "REFION" — is the GCP_REGION env var accessible?

jakubno: Thanks! yes, GCP_REGION is

PR #136, opened 2024-04-16
New specification for sandbox interaction and management

ValentaTomas:
**Draft** of the improved spec for `envd` and API.

### `envd`
- REST API server for file upload and download
- (probably) gRPC for interacting with the sandbox

### API
- Managing sandboxes and their lifecycle in a more ergonomic way

Tasks:
- [ ] Handle oom_score_adj for processes spawned by `envd`
- [ ] Handle start command by `envd`
- [ ] Should we support the client side and bidirectional streaming?
  - `WriteFileStream`
  - Stdin stream in process handling (`ConnectStream`, `StartStream`)
  - Having fewer streams seems like a more stateless option that we should think about.
  - Also streams may be harder for people to use.
- [ ] REST support via gRPC gateway
- [ ] Pass process exit error message
- [ ] Handle auth via gRPC auth
  - Should file upload/download via REST be also behind auth?
- [ ] Watch for new processes?
- [ ] Does `ReadFile` make sense to return a stream so we can chunk the response? Can this be done by returning only parts of the file instead?
- [ ] Check the HTTP/3 (gRPC)?
- [ ] Add "CreateParents" parameter when uploading a file via REST API


ValentaTomas:
Changes from this PR are included in the https://github.com/e2b-dev/infra/pull/161

PR #146, opened 2024-04-19
Set limits for orchestrator based on the machine

jakubno:
Set the limits based on the machine


main.tf
```diff
@@ -10,11 +10,11 @@ terraform {
     }
     google = {
       source  = "hashicorp/google"
-      version = "5.6.0"
+      version = "5.25.0"
```
jakubno, line 13: There wasn't the resource we needed here in the old version

PR #150, opened 2024-04-23
Consul DNS resolving

jakubno:
Allow to use consul for DNS, we need this for specifying a service on an exact node.


jakubno:
It's possible this won't be needed

ValentaTomas:
> Not needed

Why, what are we using now?

jakubno:
We use nomad client library to get all available nodes and it also gives us the IP address of the node directly, so there's no need for extra consul DNS resolving

jakubno:
Actually, we probably need this 😅  to reference services from all nodes, because they probably won't be accessible on localhost

PR #151, opened 2024-04-23
Add swap and adjust OOM handling in sandbox

ValentaTomas:
This PR configures small swaps space (limited by free disk) and OOM handling of envd and start command to prevent the most frequent OOM during sandbox usage. For proper OOM handling the zram/zswap and other OOM handling techniques (envd child process oom adjust) should be explored.

After we merge this we should rebuild the templates:
- [ ] base
- [ ] code-interpreter
- [ ] code-interpreter-stateful


mlejva:
A little bit OT but we should get rid of `code-interpreter-stateful` name and have just one `code-interpreter` template

jakubno (approved):

PR #154, opened 2024-04-29
Allow specifying sandbox timeout when creating the sandbox and modifying it later

ValentaTomas:
This PR adds a parameter during sandbox creation that allows specifying the sandbox timeout and a route that allows modifying the timeout after creation.

This is compatible with the way sandboxes are currently handled but allows us to change how the sandbox lifecycle is managed (without constant refreshes, setting a maximum timeout, or extending timeout) just by modifying SDKs' code.

Tasks
- [ ] Regenerate SDK API clients
- [ ] Update this code to handle saving/loading timeout from serialized sandbox data that can be used to recover
- [ ] Add method for getting the current timeout/sandbox close time?

### DIFF
--- a/spec/openapi.yml
+++ b/spec/openapi.yml
@@ -65,6 +65,20 @@ components:
             $ref: "#/components/schemas/Error"
 
   schemas:
+    CPUCount:
+      type: integer
+      format: int32
+      minimum: 1
+      maximum: 8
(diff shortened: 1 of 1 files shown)


ValentaTomas:
Merging with new envd/rpc branch.

PR #164, opened 2024-07-18
Migrate to AWS

ValentaTomas:
Migrate to using AWS for deploying our infra instead of GCP, mainly because of bare mental support.


ValentaTomas:
We will work on the migration after we tackle the current tasks (beta SDK, scaling, etc.).

PR #167, opened 2024-07-18
Add streaming sandbox snapshot files from storage

ValentaTomas:
To enable multiple clients we are adding Network Block Device that allows us to stream sandbox data (rootfs, memfile, snapfile) from single storage (GCP Cloud Storage for now).


ValentaTomas:
Closed because we continue the work here — https://github.com/e2b-dev/infra/pull/194.

PR #171, opened 2024-07-29
Move UFFD handling to orchestrator

ValentaTomas:
There is a problem with our current Rust UFFD handler. We will move the handler to orchestrator so we can properly develop and fix it.


ValentaTomas:
After investigating this might not be enough to fix all the Uffd problems. We are checking alternative solutions and then ideally getting rid of the Uffd.

jakubno (approved):
This fixes the uffd bug

PR #176, opened 2024-08-09
Add support for file backed loading of FC snapshots with huge pages

ValentaTomas:
Tasks:
- [ ] Return uffd handling for old templates
- [ ] Change firecracker naming, so it can distinguish between uffd and non uffd
- [ ] Copy linux from staging
- [ ] Restart client / copy firecracker and kernel to the host and chmod firecracker


ValentaTomas:
For now, we are using Uffd instead.

PR #184, opened 2024-09-17
Enforce max allowed sandbox time

jakubno:
# Description

This makes sure users can't set a higher duration for sandbox than the max sandbox duration length. This could be bypassed by setting higher timeout when creating the sandbox


packages/api/internal/cache/instance/monitoring.go
```diff
@@ -8,10 +8,15 @@ import (
 )
 
 func (c *InstanceCache) UpdateCounter(instance InstanceInfo, value int64) {
+	if c.counter == nil {
+		return
+	}
+
+	counter := *c.counter
```
ValentaTomas: Why did you add separate dereference here?

packages/api/internal/cache/instance/monitoring.go
```diff
@@ -8,10 +8,15 @@ import (
 )
 
 func (c *InstanceCache) UpdateCounter(instance InstanceInfo, value int64) {
+	if c.counter == nil {
```
ValentaTomas: If the InstanceCache was created via `NewInstanceCache` it should be always initialized. The same checks could then be for the `c.cache` everywhere.

packages/api/internal/cache/instance/instance.go
```diff
@@ -39,13 +39,13 @@ type InstanceCache struct {
 
 	logger *zap.SugaredLogger
 
-	counter   metric.Int64UpDownCounter
-	analytics analyticscollector.AnalyticsCollectorClient
+	counter   *metric.Int64UpDownCounter
```
ValentaTomas: I think I understand the dereferences now, but it gets more complicated. These fields are interfaces and interfaces in Go are pointers (to the underlying implementation, struct), so there is rarely a point of using pointer to an interface.
The default Go syntax of dereferencing pointers when accessing fields plays directly into this.

packages/api/internal/cache/instance/operations.go
```diff
@@ -67,12 +69,49 @@ func (c *InstanceCache) GetInstances(teamID *uuid.UUID) (instances []InstanceInf
...
+	if instance.StartTime.IsZero() || instance.EndTime.IsZero() || instance.StartTime.After(instance.EndTime) {
+		return fmt.Errorf("instance %s has invalid start(%s)/end(%s) times", instance.Instance.SandboxID, instance.StartTime, instance.EndTime)
+	}
+
+	if instance.EndTime.Sub(instance.StartTime) > instance.MaxInstanceLength {
+		instance.EndTime = instance.StartTime.Add(instance.MaxInstanceLength)
+	}
+
+	// Get actual TTL for the instance
+	var ttl time.Duration
```
ValentaTomas: Will the edited ttl be correctly reported to the billing?

packages/api/internal/cache/instance/instance.go
```diff
@@ -63,31 +63,36 @@ func NewCache(analytics analyticscollector.AnalyticsCollectorClient, logger *zap
...
-			InstanceId:    instanceInfo.Instance.SandboxID,
-			EnvironmentId: instanceInfo.Instance.TemplateID,
-			BuildId:       instanceInfo.BuildID.String(),
-			TeamId:        instanceInfo.TeamID.String(),
-			Timestamp:     timestamppb.Now(),
-		})
-		if err != nil {
-			errMsg := fmt.Errorf("error when sending analytics event: %w", err)
-			telemetry.ReportCriticalError(ctx, errMsg)
+		if analytics != nil {
```
ValentaTomas: If we really need to check these just check them during the initialization and fail the whole initialization.

packages/api/internal/cache/instance/operations.go
ValentaTomas: Could this calculation also be made outside of the cache so we don't have to concern ourselves with distinguishing new/old?

packages/api/internal/cache/instance/sync.go
```diff
@@ -10,17 +10,13 @@ import (
 	analyticscollector "github.com/e2b-dev/infra/packages/api/internal/analytics_collector"
 )
 
-func getMaxAllowedTTL(startTime time.Time, duration, maxInstanceLength time.Duration) time.Duration {
-	runningTime := time.Since(startTime)
-	timeLeft := maxInstanceLength - runningTime
-
+func getMaxAllowedTTL(now time.Time, startTime time.Time, duration, maxInstanceLength time.Duration) time.Duration {
+	timeLeft := maxInstanceLength - now.Sub(startTime)
```
ValentaTomas, line 14: Can we do without passing the now here?

ValentaTomas (commented):
Can we only deploy the minimal changes to fix the longer-than-maximal timeout?
As you mentioned, it can only happen for the initial sandbox creation timeout.

We can make the other improvements later, but some parts of the logic are becoming more obscure, and there are also some unnecessary changes, as I mentioned (https://github.com/e2b-dev/infra/pull/184#discussion_r1769652853).

PR #190, opened 2024-10-02
Update and Improve Documentation and Automation of Initializing Infra

Just-en-Time:
Improve documentation for infra repo setup


terraform.md
```diff
@@ -1,29 +1,83 @@
...
+8. Run `make build-and-upload-fc-components` Note: This needs to be done on a Linux machine due to case-sensitive requirements for the file system--you'll error out during the automated git section with a complaint about unsaved changes. Kernel and versions could alternatively be sourced elsewhere.
+9. At the time of this writing, several versions are required. The script may not fully create and upload these. As of 9/27/24, your Storage buckets should look like this:  
+```
+<prefix>-fc-env-pipeline/envd  
+                        /envd-v.0.0.1 (this is legacy)  
+                        /orchestrator  
+                        /template-manager
+
+<prefix>-fc-kernels/vmlinux-5.10.186/vmlinux.bin
+                   /vmlinux-6.1.99/vmlinux.bin
```
jakubno: I believe you don't this one (yet)
```suggestion
```

terraform.md
```diff
@@ -1,29 +1,83 @@
...
+
+<prefix>-fc-versions/v1.7.0-dev_8bb88311/firecracker
+                                        /uffd
+                    /v1.8.0-hugepages-state_4778a02/firecracker
+                    /v1.8.0-hugepages-state_9e0b47a/firecracker
+                    /v1.8.0-main_43ff620/firecracker
+                    /v1.9.0_fake-2476009/firecracker
+                                        /uffd
```
jakubno: I believe you don't need these 
```suggestion
```

packages/init/main.tf
```diff
@@ -1,3 +1,60 @@
+
+# Enable Secrets Manager API
+resource "google_project_service" "secrets_manager_api" {
+  #project = var.gcp_project_id
```
jakubno: Is there a reason why to keep the project here?

jakubno (commented):
Looks great, thanks for the improvements, especially enabling APIs was annoying, didn't realize you can use terraform for that as well

packages/init/main.tf
```diff
@@ -1,3 +1,66 @@
...
+
+  disable_on_destroy = false
+}
+
+resource "time_sleep" "secrets_api_wait_60_seconds" {
+  depends_on = [google_project_service.secrets_manager_api]
+
+  create_duration = "60s"
+}
+
```
jakubno: This is 80/20 improvement. Feel free to improve it by changing it to local exec, which would check if the api is already ready

terraform.md
```diff
@@ -1,29 +1,83 @@
...
+
+```
+10. Run `make apply-without-jobs`
+11. Secrets are created and stored in GCP Secrets Manager. Once created, that is the source of truth--you will need to update values there to make changes.   
+Some notes:  
+- You can optionally add Grafana and Posthog API keys for monitoring
```
jakubno, line 60: postgres_connection_string is still required

packages/init/main.tf
justine-time: No, good call--had that in there for testing. Removing it

packages/init/main.tf
justine-time: LGTM, I think it's useful to have and lean towards 80/20 as a rule. I reduced the time to 20s, When I ran through enabling the APIs they seemed to take about 10s at most, so we can split the difference.

terraform.md
justine-time: good call, added

jakubno (approved):

Makefile
```diff
@@ -15,7 +15,9 @@ tf_vars := TF_VAR_client_machine_type=$(CLIENT_MACHINE_TYPE) \
 	TF_VAR_prefix=$(PREFIX) \
 	TF_VAR_terraform_state_bucket=$(TERRAFORM_STATE_BUCKET) \
 	TF_VAR_otel_tracing_print=$(OTEL_TRACING_PRINT) \
-	TF_VAR_environment=$(TERRAFORM_ENVIRONMENT)
+	TF_VAR_environment=$(TERRAFORM_ENVIRONMENT) \
+	TF_VAR_gcp_project_id=$(GCP_PROJECT_ID) \
+	TF_VAR_cloudflare_api_token=$(CLOUDFLARE_API_TOKEN)
```
jakubno: gcp project is doubled, and cloudflare_api_token isn't used

jakubno (changes requested):

Makefile
```diff
@@ -15,7 +15,7 @@ tf_vars := TF_VAR_client_machine_type=$(CLIENT_MACHINE_TYPE) \
 	TF_VAR_prefix=$(PREFIX) \
 	TF_VAR_terraform_state_bucket=$(TERRAFORM_STATE_BUCKET) \
 	TF_VAR_otel_tracing_print=$(OTEL_TRACING_PRINT) \
-	TF_VAR_environment=$(TERRAFORM_ENVIRONMENT)
+	TF_VAR_environment=$(TERRAFORM_ENVIRONMENT) \
```
jakubno:
```suggestion
	TF_VAR_environment=$(TERRAFORM_ENVIRONMENT)
```

PR #191, opened 2024-10-02
Debug and Stabilize Logging and Loki

Just-en-Time:
Updates to increase default resource allocations for loki and logging agents, increase buffer and batch for logging. Should review before implementing in general prod, tested in smaller but high traffic prod environments

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,8 +24,8 @@ else
 endif
 
 WITHOUT_JOBS := $(shell echo $(ALL_MODULES) | tr ' ' '\n' | grep -v -e "nomad" | awk '{print "-target=module." $$0 ""}' | xargs)
-ALL_MODULES_ARGS := $(shell echo $(ALL_MODULES) | tr ' ' '\n' | awk '{print "-target=module." $$0 ""}' | xargs)
-DEBUG_ALL_MODULES_ARGS := -target=module.nomad.nomad_job.loki -target=module.nomad.nomad_job.logs-collector
+ALL_MODULES_ARGS := $(shell echo $(ALL_MODULES) | tr ' ' '\n' | awk '{print "-target=module." $$0 ""}' | xargs)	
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -24,7 +24,8 @@ else
 endif
 
 WITHOUT_JOBS := $(shell echo $(ALL_MODULES) | tr ' ' '\n' | grep -v -e "nomad" | awk '{print "-target=module." $$0 ""}' | xargs)
-ALL_MODULES_ARGS := $(shell echo $(ALL_MODULES) | tr ' ' '\n' | awk '{print "-target=module." $$0 ""}' | xargs)
+# TEMPORARILY DISABLED FOR DEBUG ALL_MODULES_ARGS := $(shell echo $(ALL_MODULES) | tr ' ' '\n' | awk '{print "-target=module." $$0 ""}' | xargs)
+ALL_MODULES_ARGS := -target=module.nomad.nomad_job.loki -target=module.nomad.nomad_job.logs-collector
```
ValentaTomas: Let's not forget to later remove this debugging edit done for the deploy convenience.

packages/nomad/logs-collector.hcl
```diff
@@ -136,21 +136,22 @@ teamID = "{{ teamID }}"
...
-auth.user = "${var.grafana_logs_username}"
-auth.password = "${var.grafana_api_key}"
-
-[sinks.grafana.labels]
-source = "logs-collector"
-service = "{{ service }}"
-teamID = "{{ teamID }}"
-envID = "{{ envID }}"
-sandboxID = "{{ sandboxID }}"
+# [sinks.grafana]
```
ValentaTomas: We might probably also cleanup the vars that were passed here from TF that might no longer be used before this goes out of draft.

Makefile
justine-time: I just added a `make debug-apply` make target to keep both and restore `make apply` to its original purpose

packages/nomad/logs-collector.hcl
justine-time: Good call, cleaned up

jakubno:
Important parts merged to `dev`

PR #192, opened 2024-10-02
Add sandbox SSH alias

jakubno:
To improve DX you can now run only 
```bash
sbx-ssh <sandboxID>
```
and you will be directly connected to the sandbox.

### DIFF
--- a/packages/cluster/scripts/start-client.sh
+++ b/packages/cluster/scripts/start-client.sh
@@ -153,3 +153,11 @@ echo $overcommitment_hugepages >/proc/sys/vm/nr_overcommit_hugepages
 # These variables are passed in via Terraform template interpolation
 /opt/consul/bin/run-consul.sh --client --consul-token "${CONSUL_TOKEN}" --cluster-tag-name "${CLUSTER_TAG_NAME}" --enable-gossip-encryption --gossip-encryption-key "${CONSUL_GOSSIP_ENCRYPTION_KEY}" &
 /opt/nomad/bin/run-nomad.sh --client --consul-token "${CONSUL_TOKEN}" &
+
+# Add alias for ssh-ing to sbx
+echo "_sbx_ssh() {
+  local address=$(dig @localhost $1. A +short 2>/dev/null)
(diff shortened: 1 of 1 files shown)


mlejva:
Is this a legit ssh now? I remember we had to fame it before

jakubno:
It's inside the host. You can connect to client and then to sandbox, it's DX for e2b devs.

ValentaTomas (approved):

0div:
Thx for this! 🙏 FYI i built the equivalent as util func in the orchestrator go code, using it for tests for now.

packages/cluster/scripts/start-client.sh
```diff
@@ -153,3 +153,11 @@ echo $overcommitment_hugepages >/proc/sys/vm/nr_overcommit_hugepages
 # These variables are passed in via Terraform template interpolation
 /opt/consul/bin/run-consul.sh --client --consul-token "${CONSUL_TOKEN}" --cluster-tag-name "${CLUSTER_TAG_NAME}" --enable-gossip-encryption --gossip-encryption-key "${CONSUL_GOSSIP_ENCRYPTION_KEY}" &
 /opt/nomad/bin/run-nomad.sh --client --consul-token "${CONSUL_TOKEN}" &
+
+# Add alias for ssh-ing to sbx
+echo '_sbx_ssh() {
+  local address=$(dig @localhost $1. A +short 2>/dev/null)
```
0div, line 159: on my host at least, localhost doesn't return the IP of a running instance but this does:
```
jonas@jtdevorch-client-fkl5:~/infra/packages/orchestrator$ dig @127.0.0.4 test-instance-1 A +short
192.168.2.197

jonas@jtdevorch-client-fkl5:~/infra/packages/orchestrator$ dig @localhost test-instance-1 A +short
<nada>
```

jakubno: looks like there must be something different, works for me on a clean deploy

0div: maybe worth noting: im looking for instances created via `make test`

PR #193, opened 2024-10-03
Add a way to upload several files to a sandbox at once

0div:
# Description
[E2B-954](https://linear.app/e2b/issue/E2B-954/add-a-way-to-upload-several-files-to-a-sandbox-at-once)

For now, simply reusing `PostFiles` logic.


ValentaTomas:
This can be closed, because of the existing multipart implementation, right?

PR #196, opened 2024-10-04
Add multi-file write support

0div:
# Description
Create separate `multipart.Part` method to get file name with full path with directories.

# Tests
See the tests in https://github.com/e2b-dev/E2B/pull/451


packages/envd/internal/api/upload.go
```diff
@@ -110,7 +110,7 @@ func processFile(r *http.Request, path string, part *multipart.Part, user *user.
 func resolvePath(part *multipart.Part, paths *UploadSuccess, u *user.User, params PostFilesParams) (string, error) {
 	var pathToResolve string
 
-	if params.Path != nil {
+	if params.Path != nil && *params.Path != "" {
```
ValentaTomas: I think we would ideally make it so that the SDK does not need this change.

0div: This is adapting to the SDK (the `js-sdk` at least), as in, in the SDK you can only send `string` not `string | undefined` , or else you get an API error

ValentaTomas: I would like to know if the API error is caused by the client generated from the OpenAPI spec or if this is because of the envd implementation.

0div: it comes from `handleEnvdApiError`

```
 ❯ Module.handleEnvdApiError src/envd/api.ts:20:14
     18|   switch (res.response.status) {
     19|     case 400:
     20|       return new InvalidArgumentError(message)
       |              ^
     21|     case 401:
     22|       return new AuthenticationError(message)
```

0div: it says it's optional from the OpenAPI spec: https://github.com/e2b-dev/infra/blob/main/packages/envd/internal/api/api.gen.go#L235

0div:
@ValentaTomas per discussion in https://github.com/e2b-dev/E2B/pull/451, I overloaded the `multipart.Part.FileName()` method so that it's possible to get the full path with directories. It's unsafe for regular web apps, but in our sandboxes it's ok, correct?  I adapted the `js-sdk` to include nested paths and it works.

ValentaTomas:
Yes, I think it should be ok for us, because we are giving users access to the whole sandbox so this is not security risk.

packages/envd/internal/utils/multipart.go
```diff
@@ -0,0 +1,41 @@
...
+
+// CustomPart is a wrapper around multipart.Part that overloads the FileName method
+type CustomPart struct {
+	*multipart.Part
+}
+
+// FileName returns the filename parameter of the Part's Content-Disposition header.
+// This method overrides the original FileName method to return the full filename
+// without using filepath.Base.
+func (p *CustomPart) FileName() (string, error) {
```
ValentaTomas: We probably don't have to overload the `FileName`, just create a new one method like `FilePath` to indicate we are not using the default expected behavior via the naming.

ValentaTomas: Also if you are uploading files with just the URL in browser will this be backwards compatible? I'm not sure if the paths that the browser adds to the files are just filenames by default of full paths and it relies on the server just getting the filenames.

0div: It seems like most browsers (besides some IE exceptions..) only add filename in the input element target for security reasons.

ValentaTomas: Ok, nice!

ValentaTomas:
Is it tested for the old file uploads via SDK when passing relative and absolute path and uploads via URL (with the path being part of the multipart)?
I just want to make sure it is compatible with the current usage.

0div:
> Is it tested for the old file uploads via SDK when passing relative and absolute path and uploads via URL (with the path being part of the multipart)? I just want to make sure it is compatible with the current usage.

New write tests have passed as follow: https://github.com/e2b-dev/E2B/pull/503/files#diff-7a70570a2ea87cb7d3e355b0aa6f011a4bec4d1733f40bf94b05e4024f232b01

ValentaTomas (approved):

jakubno (approved):

ValentaTomas:
If everything is ok here we can merge to `main`—from there we can then deploy to clusters.

PR #210, opened 2024-12-02
Add watch dir recursive option

dobrac:
Add recursive directory watch support to the envd. It uses not yet released recursive feature of [fsnotify](https://github.com/fsnotify/fsnotify) library. This library was forked to https://github.com/e2b-dev/fsnotify and is now included instead of the original one.

Updated API with the `recursive` parameter as:
```
message CreateWatcherRequest {
    string path = 1;
    bool recursive = 2;
}
```

It should be backwards compatible, as all fields in protobuf are optional. The default value is recursive=false.

**Important restriction**
This change limits the compatibility to Linux, Windows only.

<details open>
<summary>:sparkles: <i><h3>Description by Callstackai</h3></i></summary>

This PR adds recursive directory watch support to the envd, utilizing a forked version of the fsnotify library. It introduces a new `recursive` parameter in the WatchDirRequest and CreateWatcherRequest messages, ensuring backward compatibility.

<details>
<summary><b>Diagrams of code changes</b></summary>

```mermaid
sequenceDiagram
    participant Client
    participant FileSystem Service
    participant FSNotify Watcher

    Client->>FileSystem Service: CreateWatcher(path, recursive)
    FileSystem Service->>FSNotify Watcher: Create new watcher
    FSNotify Watcher-->>FileSystem Service: Return watcher instance
    
    alt recursive watching
        FileSystem Service->>FSNotify Watcher: Add path with "..." suffix
    else non-recursive watching
        FileSystem Service->>FSNotify Watcher: Add path without suffix
    end
    
    FileSystem Service-->>Client: Return watcher_id
    
    loop File System Events
        FSNotify Watcher->>FileSystem Service: Emit events (Create/Write/Remove/Rename/Chmod)
        FileSystem Service->>Client: Stream events
    end
    
    Client->>FileSystem Service: RemoveWatcher(watcher_id)
    FileSystem Service->>FSNotify Watcher: Close watcher
    FileSystem Service-->>Client: Confirm removal
```

</details>

<details>
<summary><b>Files Changed</b></summary>
<table>
<tr><th>File</th><th>Summary</th></tr>
<tr><td><a href=https://github.com/e2b-dev/infra/pull/210/files#diff-20cee3a725d1c5d2c595ff5372e6593d3bcaa7aebd8319866f65e90c6924f247>go.work.sum</a></td><td>Updated dependencies to include the new fsnotify library.</td></tr>
<tr><td><a href=https://github.com/e2b-dev/infra/pull/210/files#diff-6ff03137b9e02ed40dde5a0274654f4104cbb0b8d1a41c976192ce9b39d5bb1c>packages/envd/internal/services/filesystem/watch.go</a></td><td>Modified the WatchDir function to use the new recursive path utility.</td></tr>
<tr><td><a href=https://github.com/e2b-dev/infra/pull/210/files#diff-834039e24b75798bfaf83a9fc463fdb0f5d07445d59ea3d823957da0f8dc3c30>packages/envd/internal/services/filesystem/watch_sync.go</a></td><td>Updated CreateFileWatcher to accept a recursive parameter.</td></tr>
<tr><td><a href=https://github.com/e2b-dev/infra/pull/210/files#diff-36c5d1624fa1e707a62a13b6f50a21b6a8d301128ac194bf8b3fb0af7c8bea24>packages/envd/internal/services/spec/filesystem/filesystem.pb.go</a></td><td>Added recursive field to WatchDirRequest and CreateWatcherRequest messages.</td></tr>
<tr><td><a href=https://github.com/e2b-dev/infra/pull/210/files#diff-278166b3a6429b01212b538bbf754d5e50b14852903ab5cc017779512500625d>packages/envd/internal/utils/rfsnotify.go</a></td><td>Created a utility function to handle recursive path creation.</td></tr>
<tr><td><a href=https://github.com/e2b-dev/infra/pull/210/files#diff-e5cfdaf8d2af9c0bc1401f62098c2768acad24ba196a5f9b3a84ff3d0280fd47>packages/envd/main.go</a></td><td>Updated version number to reflect changes.</td></tr>
<tr><td><a href=https://github.com/e2b-dev/infra/pull/210/files#diff-04025526e58674c4387d89e7a2034999e18685eab4863244d4d8418cd59e8cfc>packages/envd/spec/filesystem/filesystem.proto</a></td><td>Updated proto definitions to include recursive parameter.</td></tr>

</table>
</details>

*This PR includes files in programming languages that we currently do not support. We have not reviewed files with the extensions `.sum`, `.go`, `.proto`. <a href=https://docs.callstack.ai/introduction>See list of supported languages.</a>*

</details>


packages/envd/spec/filesystem/filesystem.proto
```diff
@@ -91,6 +92,7 @@ message WatchDirResponse {
 
 message CreateWatcherRequest {
     string path = 1;
+    bool recursive = 2;
```
0div, line 95: same comment as for the SDK PR, do you think it would make sense to not add this option and always watch recursively, and also as to avoid changing the communication body too much.

packages/envd/internal/utils/rfsnotify.go
```diff
@@ -0,0 +1,12 @@
+package utils
+
+import "path/filepath"
+
+// FsnotifyPath creates an optionally recursive path for fsnotify/fsnotify internal implementation
+func FsnotifyPath(path string, recursive bool) string {
```
0div, line 6: ❤️ this, i was initially turned off by the fact that the original code hardcoded the branch to never allow to use the `"..."` path with a cryptic comment saying "allowed only for test". The tests you wrote seem to show that this works! 
This not necessarily a request for code change but what would you do to test/find proof that this doesn't listen to  special linux files (Block file, Character device file, Symbolic link file, Socket file, etc)? Also, in the same vein, what root folders would you exclude, if any, to avoid spamming?

0div (changes requested):
Same comment as for SDK PR:

From a product perspective, do you think it would be better not to make the recursive an option always and just always watch recursively ? Also taking into account that the instructions said:
> Make as few changes to the communication between envd and SDKs as possible

packages/envd/spec/filesystem/filesystem.proto
dobrac: Yes, I think so. But as per the instructions

> make it so that the updated version is backward compatible

changing the behavior might introduce a breaking change for current users. I was thinking, that this parameter could be eventually changed to true by default and then deprecated and removed.

But if you think that this kind of change is on par with other changes you do in between versions, then I’m fine with removing non-recursive watching.

packages/envd/internal/utils/rfsnotify.go
dobrac: This is a very good point.

**Special Linux files**
Basic files should be handled the same as without recursive watch. The implementation of recursion uses either isDir() or isDir := mask&unix.IN_ISDIR == unix.IN_ISDIR to watch nested directories (based on the source code). I've tested block files, device files (/dev), socket files (using python), etc. and I didn't find any issues (excessive updates or crashing). Symbolic links of folders are followed.

**Spam**
About spamming, maybe it would make sense to prevent users for listening on certain directories (/dev, /proc, /var, /sys). But, even without the recursive watch, they can watch these folders or too many files and break the containers. So it might not be necessarily an issue.

packages/envd/spec/filesystem/filesystem.proto
0div: I read that as backward compatible as far as the API goes, but you raise a good point about what our users expect in terms of behaviour. Do you think that given that these SDKs are versioned and that from this version on we explicitly state in the docs and comments that this is recursive could help?

packages/envd/internal/utils/rfsnotify.go
0div: Cool thanks for checking the behaviour for special files 👍 
For the spamming I was thinking of not allowing those folders to be watched at all, sometimes a user may want to watch root `/` folder and with this we could lower the default linux files being watched from ~4000 to ~800 last i checked. Not needed for this PR tho, just wanted to see your thought process, thanks.

dobrac: Ok 👍 I agree with the follow-up PR

packages/envd/spec/filesystem/filesystem.proto
dobrac: The SDKs are versioned, but without updating the API (.proto), this update would be server (infra/envd) only. We would effectively push it immediately everywhere. It sounds to me much more risky if anything goes wrong.

I can make the default in the SDK recursive=true and remove the option only on the SDK side — users won't be able to change it. This way, we won't influence old deployments, but we will make it dependent on the SDK version. (If I saw it correctly, you had few potentially breaking changes between minor SDK versions, so it might be ok.) Would you prefer this option?

ValentaTomas: Yeah, I prefer having the parameter added in the API, adding the recursive flag to SDK methods, and then when we do breaking SDK update, changing the default behavior to `recursive=True` while you can still specify it.

jakubno (approved):

0div (approved):

PR #224, opened 2024-12-19
Show 404 for not running/existent sandbox

dobrac:


ValentaTomas:
One consideration is whether we should return 404 specifically when the sandbox is not found.
The request is not to an API about a resource but to a "server" that is not accessible.

Also, because we might return 404 for some requests from the SDK (for example, if a file is not found in the sandbox), it would make it harder to distinguish between these errors.

jakubno:
Good point, but what are the options, 503 maybe? In the end, the text should explain it

ValentaTomas:
Maybe even 502 could make sense, but as you said, if we explain in the next what is happening, it should not be a problem.
I think thinking here in terms of what are the other errors we will want to return for OOM/port could make it clearer what should this one be.

PR #231, opened 2025-01-07
Show useful 502 message for not running/existent sandbox  

0div:
# Description

[re-using previous work from @dobrac in [#224](https://github.com/e2b-dev/infra/pull/224)]

To return a meaningful error during the DNS resolution in API from `client-proxy` we route the traffic to a "dummy" server that serves the exception if the sandbox cannot be found. 
Per feedback, also split the DNS package between api/ and orchestrator/ , with only the latter having the special "sandbox not found" routing.

# Test 
- [x] deploy 
- [x] [test](https://github.com/e2b-dev/E2B/pull/540) ✅ 

<details open>
<summary>:sparkles: <i><h3>Description by Callstackai</h3></i></summary>

This PR introduces a new DNS handling feature that provides a meaningful 502 error message when a sandbox is not running or does not exist. It also refactors the DNS package structure and updates the configuration for the client proxy.

<details>
<summary><b>Diagrams of code changes</b></summary>

```mermaid
sequenceDiagram
    participant Client
    participant DNS Server
    participant Records Map
    
    Note over DNS Server: Initialized with empty records map
    
    alt Add Sandbox Record
        Client->>DNS Server: Add(sandboxID, ip)
        DNS Server->>Records Map: Insert(hostname, ip)
    end

    alt DNS Request
        Client->>DNS Server: DNS Query
        DNS Server->>Records Map: Get(sandboxID)
        alt Record Found
            Records Map-->>DNS Server: Return IP
            DNS Server-->>Client: Response with Sandbox IP
        else Record Not Found
            Records Map-->>DNS Server: Not Found
            DNS Server-->>Client: Response with 127.0.0.1
        end
    end

    alt Remove Sandbox Record
        Client->>DNS Server: Remove(sandboxID, ip)
        DNS Server->>Records Map: Remove Entry
    end
```

</details>

<details>
<summary><b>Files Changed</b></summary>
<table>
<tr><th>File</th><th>Summary</th></tr>
<tr><td><a href=https://github.com/e2b-dev/infra/pull/231/files#diff-961fb1f2c35561d42542795a834f81078ed60fb9e7606a558c96fc8274da83ea>packages/api/internal/dns/server.go</a></td><td>Added a new DNS server implementation that handles sandbox requests and returns appropriate IP addresses or a default routing IP.</td></tr>
<tr><td><a href=https://github.com/e2b-dev/infra/pull/231/files#diff-b13187f1770e1ab7c360c4be9e23b3d45806471d6f22b6d3d935be2b06b4324a>packages/api/internal/orchestrator/orchestrator.go</a></td><td>Updated import path for the DNS package to reflect the new structure.</td></tr>
<tr><td><a href=https://github.com/e2b-dev/infra/pull/231/files#diff-6c27f7f9086581bf590ffa990fe7731852146f3f010f12a7c4cedaf747e0e85d>packages/nomad/proxies/client.conf</a></td><td>Modified the Nginx configuration to return a 404 error with a message for invalid sandbox URLs and added a mock server for handling non-existent sandboxes.</td></tr>
<tr><td><a href=https://github.com/e2b-dev/infra/pull/231/files#diff-99efebff49c5c6e67467074b4aaa0eac6b81392af893783fa64dbd2791c481df>packages/orchestrator/cmd/mock-sandbox/mock.go</a></td><td>Updated import path for the DNS package to reflect the new structure.</td></tr>
<tr><td><a href=https://github.com/e2b-dev/infra/pull/231/files#diff-409c739871a0085c2b9e6aa644708715345bfe4cde320617023aee382b819bee>packages/orchestrator/cmd/mock-snapshot/mock.go</a></td><td>Updated import path for the DNS package to reflect the new structure.</td></tr>
<tr><td><a href=https://github.com/e2b-dev/infra/pull/231/files#diff-98c0a06d62146b351f4bc3d1e0bbd0b0ba6d45736f1c90c0cb6cc01bb403506c>packages/orchestrator/internal/sandbox/sandbox.go</a></td><td>Updated import path for the DNS package to reflect the new structure.</td></tr>
<tr><td><a href=https://github.com/e2b-dev/infra/pull/231/files#diff-5b708ea611cbb317ed859293c637da1fb8bd3793d2aa8c16e97c0f8eed0ab6ee>packages/orchestrator/internal/server/main.go</a></td><td>Updated import path for the DNS package to reflect the new structure.</td></tr>

</table>
</details>

*This PR includes files in programming languages that we currently do not support. We have not reviewed files with the extensions `.conf`. <a href=https://docs.callstack.ai/introduction>See list of supported languages.</a>*

</details>


jakubno (changes requested):
the DNS is now shared for API and orchestrator. I think it will behave the same, because there's nothing running on 3003 port, but it would be better and cleaner to split it

packages/nomad/proxies/client.conf
```diff
@@ -85,6 +85,14 @@ server {
   }
 }
 
+# Mock for sandbox server when the sandbox is not running, 127.0.0.1 is returned by the DNS resolver
+server {
+  listen 3003;
+
+  default_type text/plain;
+  return 502 'Sandbox is not running or not found.';
```
mlejva: What's the difference between "not running" and "not found" state? We should be more explicit in the error message. It sounds like the same state.

mlejva (changes requested):

packages/nomad/proxies/client.conf
0div: Sometimes our users get this after starting a sandbox and expecting it to still run vs a user using a bad sandbox ID (a sandbox that was never created). In the first scenario, if they only get "not found" they might think there's a problem on our end?

mlejva: > Sometimes our users get this after starting a sandbox and expecting it to still run 

Are you talking about `Sandbox.connect()` here?

My main problem with this error message is that it kind of sounds like two same states. Do you have any ideas for better phrasing it? Or explicitly distinguishing between these two different errors?

0div: Examples i can think of are when they want to make a request to a sandbox (using `getHost` with a specific port) or when they have a sandbox instance in memory and try to call methods on it, expecting it to still exist. 

What do you think of `Sandbox does not exist` ?

mlejva: Yeah that's good.

mlejva: Yeah that's good. 

I was discussing this with Tomas - eventually we must think more about what error codes we're using. A lot of people see 502, that means server-side and they immediately think it's something wrong with us

0div:
@jakubno split the DNS package between api/ and orchestrator/ , with only the latter having the special "sandbox not found" routing

jakubno (approved):

PR #234, opened 2025-01-09
[Examples] Optimize snapshots storage

ValentaTomas:
Optimize snapshot storage in preparation to enable the automatic pausing of snapshots.

Subtasks:
- [x] _Handle empty block in orchestrator (was already implemented)_
- [ ] Skip empty blocks during snapshotting
  - [x] Memfile (untested)
  - [ ] _Rootfs (lower priority)_
- [x] Skip unchanged blocks during snapshotting
  - [x] Memfile (untested)
- [x] Start creating empty blocks during template building and upload diff with header instead of the full template
  - [x] Memfile (untested)
  - [x] Rootfs (untested)
- [x] Handle diffs instead of just full snapshots in orchestrator for normal sandbox start (right now we handle diffs only for resume)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,6 @@ ENV := $(shell cat .last_used_env || echo "not-set")
 -include .env.${ENV}
 
 OTEL_TRACING_PRINT ?= false
-IMAGE := e2b-orchestration/api
 EXCLUDE_GITHUB ?= 1
 
 tf_vars := TF_VAR_client_machine_type=$(CLIENT_MACHINE_TYPE) \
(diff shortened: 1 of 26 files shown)


ValentaTomas:
The remaining work to do here is:

- [ ] Test if everything works
- [ ] Profile the impact and optimize

ValentaTomas:
When checking pause on some bigger snapshots (around 3 GB total) the post-processing of memory took most of the time—when we work on the outlines optimizations we should take into account how they might slow down this further+optimize this speed also.

ValentaTomas:
This PR will probably be cherry picked for the actual changes after we add tests/benchmarks to the code that would be affected. Otherwise this could result in regression.

ValentaTomas:
We are keeping the branch around as a reference, but this PR won't be merged. We partially implemented some of the improvements already.

PR #237, opened 2025-01-10
api: polish api service shutdown and top-level context handling

tychoish:
This makes shutdown, in response to SIGTERM, cleaner and more correct: 
- avoided creating a new context in several key places, notably, the base context handed to requests.
- the context canceled when the signal is caught (SIGTERM). 
- the process will not exit until all request handlers have returned and any shutdown operations.

As an added bonus, I made some changes to error handling and logging which should remove the case of "process exited without logging anything useful". 

There are still some aspects which could require more attention in a follow up PR:
- I haven't looked at all the places where we either create a background context (should only really be done in tests (maybe) and in `main()`, but this requires more churn and should happen in a later PR. 
- I haven't audited all the places where we create go routines to make sure that we give them reasonable contexts and wait for them during shut down, if needed. 

The changes for shutdown handling are pretty generic (though not implemented here. If needed later, we can either  use stuff from one of my repos (if suitable) or factor the generic parts out to be used in other processes.

### DIFF
--- a/packages/api/main.go
+++ b/packages/api/main.go
@@ -1,11 +1,18 @@
 package main
 
 import (
+	"context"
+	"errors"
 	"flag"
 	"fmt"
 	"log"
+	"net"
 	"net/http"
 	"os"
+	"os/signal"
+	"sync"
+	"sync/atomic"
+	"syscall"
 	"time"
 
 	"github.com/getkin/kin-openapi/openapi3"
@@ -34,7 +41,7 @@ const (
 	defaultPort          = 80
 )
 
-func NewGinServer(apiStore *handlers.APIStore, swagger *openapi3.T, port int) *http.Server {
+func NewGinServer(ctx context.Context, apiStore *handlers.APIStore, swagger *openapi3.T, port int) *http.Server {
 	// Clear out the servers array in the swagger spec, that skips validating
 	// that server names match. We don't know how this thing will be run.
 	swagger.Servers = nil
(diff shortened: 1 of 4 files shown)


packages/api/main.go
```diff
@@ -114,44 +121,138 @@ func NewGinServer(apiStore *handlers.APIStore, swagger *openapi3.T, port int) *h
...
+		}
+	case <-ctx.Done():
+		log.Println("http service (%d) shutdown outside of signal", port)
 	}
+
+	// call cleanup explicitly because defers do not run on os.Exit
+	cleanup()
+
+	// Exit, with appropriate code.
+	os.Exit(int(exitCode.Load()))
```
jakubno, line 258: LGTM. I have a small question, what's the logic behind the exit code?

jakubno:
Hey @tychoish, I saw you started adding some changes in orchestrator and template-manager. Are those relevant to API part? Ideally we would have several smaller PRs rather than one PR.

tychoish:
@jakubno definitely can split it up: the change to the API wanted a changed to a shared package. There's no way to change the shared code _and_ the calling code (in the API and orchestrator packages,) in different commits, but I can pull the shared change from the larger API changes. Regardless the orchestrator change is pretty small.

jakubno:
No problem with that, just making sure it has limited scope

packages/api/main.go
tychoish: What do you mean "logic": the exit code is 0 for non-zero "normal" exits, and greater than 1 for others. This is the same as before (exits of > 0 for errors, as a result of calling `log.Fatalf` or `os.Exit(1)`. I'm incrementing this counter from different threads, so the exact number would indicate the number of errors.

The oddity, upon reflection, is that I increment it by 2 on line 230, that should probably just be 1.

packages/api/main.go
```diff
@@ -115,44 +122,138 @@ func NewGinServer(apiStore *handlers.APIStore, swagger *openapi3.T, port int) *h
...
+		// ended outside of shutdown, (we want everything else
+		// to cleanup, and not to get stuck waiting).
+		cancel()
+	}()
+
+	select {
+	case <-signalCtx.Done():
+		// shutdown blocks until all active http handlers have returned.
+		if err := s.Shutdown(ctx); err != nil {
+			exitCode.Add(1)
```
tychoish, line 247:
```suggestion
			exitCode.Add(0)
```

packages/api/main.go
tychoish: I've removed the increment by 2, so now we just increment by 1 for every error that we get in any of the background or shutdown threads.

All errors will be logged.

tychoish (commented):
this PR, is a superset of #248 (which can be merged at any time.) 

Getting #248 out of the way will make this easier to review, but it's testable in its current form, will set this PR to draft while I test, but there's no further work to do here.

packages/orchestrator/main.go
```diff
@@ -24,6 +24,7 @@ func main() {
 
 	if !env.IsLocal() {
 		shutdown := telemetry.InitOTLPExporter(ctx, server.ServiceName, "no")
+
 		defer shutdown(context.TODO())
```
tychoish:
```suggestion
		defer shutdown(context.TODO())
```

tychoish:
Status:
- With #248 merged, this PR is more minimally scoped, and ready to review/merge
- would like to do one final round of of validation/smoke test before merging

tychoish:
Update: I've thoroughly tested this and am confident that it's safe to merge and deploy.

packages/api/main.go
jakubno: I thought there's some logic based on `2` and this comment
```
	// Exit, with appropriate code
```

Something like error code 137 is indicates OOM

packages/api/main.go
```diff
@@ -116,47 +122,138 @@ func NewGinServer(apiStore *handlers.APIStore, swagger *openapi3.T, port int) *h
...
+			log.Printf("http service (%d) shutdown error: %v", port, err)
+		}
+	case <-ctx.Done():
+		log.Println("http service (%d) shutdown outside of signal", port)
 	}
+
+	// call cleanup explicitly because defers do not run on os.Exit
+	cleanup()
```
jakubno, line 255: TIL, that would get me one day :)

jakubno (approved):

packages/api/main.go
tychoish: let's not talk about how I learned this one 😅

packages/api/main.go
tychoish: I don't think we can do logic based on the specific return value, but we couldn't already... We could just exit zero or 1, but it seemed interesting to propagate something from the counter?

PR #247, opened 2025-01-21
Increase memory during build when memory for the template is small

jakubno:
# Description

Increase memory for the part of the build when exporting rootfs

Before this change:
```
error | error creating rootfs for env 'ffj7ru56ykc5yfcflniu' during build 'ee6830b4-dd2f-4e6c-aa66-b14f91d8ab32': error creating rootfs file: container exited with status 137:
```

After I was able to build the template and start it:

```sh
~ E2B_DOMAIN=krindapana.dev e2b sbx sp ffj7ru56ykc5yfcflniu
Terminal connecting to template ffj7ru56ykc5yfcflniu with sandbox ID ildfkac242oujw2ntorzm-447dc250
user@d40650b7c9d8:~$ free -h
               total        used        free      shared  buff/cache   available
Mem:           106Mi        59Mi       5.3Mi       936Ki        51Mi        47Mi
Swap:          127Mi          0B       127Mi
user@d40650b7c9d8:~$ 
```

### DIFF
--- a/packages/template-manager/internal/build/rootfs.go
+++ b/packages/template-manager/internal/build/rootfs.go
@@ -255,6 +255,7 @@ func (r *Rootfs) createRootfsFile(ctx context.Context, tracer trace.Tracer) erro
 	telemetry.ReportEvent(childCtx, "created network")
 
 	pidsLimit := int64(200)
+	memory := int64(math.Max(float64(r.env.MemoryMB)/2, 512)) << ToMBShift
 
 	cont, err := r.client.ContainerCreate(childCtx, &container.Config{
 		Image:        r.dockerTag(),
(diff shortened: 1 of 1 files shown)


ValentaTomas:
Are we changing the sandbox minimal ram because there is some problem when it is <256?
Also this would need to change it docs then.

@mlejva Was there some discussion about this?

ValentaTomas:
@jakubno Pinging here

ValentaTomas:
@jakubno @mlejva We should add the change in docs/landing page and merge this if the templates with 128 are not viable.

ValentaTomas:
I'm okay with changing this, especially if the 128 MB is not a big enough size to run anything.

ValentaTomas:
Let's reflect this in the docs+landing page, but I think we should add this, as the 128MB option is not properly usable now.

dobrac:
> Let's reflect this in the docs+landing page, but I think we should add this, as the 128MB option is not properly usable now.

Just a note, if we increase the minimum like this, it might be for some users a breaking change. The requests could start failing because it changes the API spec (if they were passing any value under 256MB).

jakubno:
I am not sure if it was the build or the start of the sandbox wasn't failing with 128MB, i will test so we can move forward with this

ValentaTomas:
We can also check if there are even any valid builds with 128 in DB. Let's move quickly on this. We can add the doc change right away too stop advertising this.

dobrac:
It looks like only one template with 128MB was built on prod. So I think it's safe to bump it to 256MB.

ValentaTomas:
@jakubno This modifies just the docker during build?

jakubno:
Yes, the OOM was during the "docker phase" when we export rootfs

PR #249, opened 2025-01-21
Add autoscaling (only up) for client

jakubno:
The scaling down is turned off for obvious reasons.


ValentaTomas:
I think if we parametrize the `max_replicas` we can merge this to `main`.

ValentaTomas (approved):

PR #250, opened 2025-01-22
Prevent closed block cache usage

dobrac:
Prevent the ability to call read/write methods on already closed block cache in orchestrator, as it might cause invalid reads/writes. This might be simulated by closing the build cache right after creation while fetchToCache is still running.

### DIFF
--- a/packages/orchestrator/internal/sandbox/block/cache.go
+++ b/packages/orchestrator/internal/sandbox/block/cache.go
@@ -137,10 +137,10 @@ func (m *Cache) Close() error {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
-	if m.isClosed() {
+	succ := m.closed.CompareAndSwap(false, true)
+	if !succ {
 		return NewErrCacheClosed(m.filePath)
 	}
-	m.closed.Store(true)
 
 	return errors.Join(
 		m.mmap.Unmap(),
--- a/packages/orchestrator/internal/sandbox/block/cache.go
+++ b/packages/orchestrator/internal/sandbox/block/cache.go
@@ -14,6 +14,8 @@ import (
 	"github.com/e2b-dev/infra/packages/shared/pkg/storage/header"
 )
 
+const errClosedTemplate = "block cache already closed for path %s"
+
 type Cache struct {
 	filePath  string
 	size      int64
@@ -22,6 +24,7 @@ type Cache struct {
 	mu        sync.RWMutex
 	dirty     sync.Map
 	dirtyFile bool
+	closed    chan struct{}
 }
 
 // When we are passing filePath that is a file that has content we want to server want to use dirtyFile = true.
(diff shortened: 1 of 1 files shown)


ValentaTomas (commented):
I would consider using [`atomic.Bool`](https://pkg.go.dev/sync/atomic#Bool) for the close state instead of the channel. The channel should work ok, but the bool should be more clear.

packages/orchestrator/internal/sandbox/block/cache.go
```diff
@@ -14,6 +14,8 @@ import (
 	"github.com/e2b-dev/infra/packages/shared/pkg/storage/header"
 )
 
+const errClosedTemplate = "block cache already closed for path %s"
```
ValentaTomas: If the error is used at multiple places I think we would ideally create a regular error instead of ad-hoc error with templating, so we can potentially check for this type of error later.

One solution that I think could work is having an error like this (because error is just an interface with `Error() string` method):

```go
type ErrCacheClosed struct {
    filePath string
}

func (e *ErrCacheClosed) Error() string {
    return fmt.Sprintf("block cache already closed for path %s", e.filePath)
}

func NewErrCacheClosed(filePath string) *ErrCacheClosed {
  return &ErrCacheClosed{
    filePath: filePath,
  }
}

// return nil, NewErrCacheClosed(filePath)

```

dobrac: Makes sense, I'll update it.

packages/orchestrator/internal/sandbox/block/cache.go
```diff
@@ -98,26 +126,43 @@ func (m *Cache) WriteAt(b []byte, off int64) (int, error) {
...
+	}
+
 	return m.WriteAtWithoutLock(b, off)
 }
 
 func (m *Cache) Close() error {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
+	if m.isClosed() {
```
ValentaTomas: Btw, there might be and improvement here with using https://pkg.go.dev/sync/atomic#Bool.CompareAndSwap if you want to add it.

dobrac: Updated, hopefully it's correct 🙏

ValentaTomas: I think it is, also using something similar here—https://github.com/e2b-dev/infra/blob/main/packages/orchestrator/internal/sandbox/block/overlay.go#L49

ValentaTomas:
I think this looks good. If the (manual) testing is ok I think we can merge.

dobrac:
> I think this looks good. If the (manual) testing is ok I think we can merge.

Manual testing is ok. Tested using low TTL value in build cache causing closing the file before it's fully downloaded.

ValentaTomas:
I'll just do some testing before merging, but otherwise this looks good.

ValentaTomas (approved):

PR #253, opened 2025-01-23
Automate build cache cleanup in orchestrator

dobrac:
### Description
Build cache cleanup automation is based on the percentage of free disk space. When the total disk usage reaches `75%`, the oldest item in the cache is evicted. This is checked every second.

File deletion (Close) is delayed by `60 seconds` to prevent race conditions with exposed slices, pending data fetches, or data uploads. If the data scheduled for eviction is requested during this time, the eviction is canceled.

### Tests
There are automated tests for basic usage. I've done also manual testing by running `make mock` as well as using SDK.


packages/orchestrator/internal/sandbox/block/cache.go
```diff
@@ -160,3 +161,32 @@ func (m *Cache) WriteAtWithoutLock(b []byte, off int64) (int, error) {
...
+func (m *Cache) DirtySize() (int64, error) {
+	var blocks int64
+	m.dirty.Range(func(key, value any) bool {
+		blocks += 1
+		return true
+	})
+
+	return blocks * m.blockSize, nil
+}
+
```
jakubno: this isn't used

packages/orchestrator/internal/sandbox/build/cache.go
```diff
@@ -4,58 +4,76 @@ import (
...
+		cachePath: cachePath,
+		cache:     cache,
+		ctx:       ctx,
+		pdSizes:   make(map[string]*deleteDiff),
+		pdDelay:   delay,
+	}
+
 	cache.OnEviction(func(ctx context.Context, reason ttlcache.EvictionReason, item *ttlcache.Item[string, Diff]) {
 		buildData := item.Value()
+		defer ds.resetDelete(item.Key())
```
jakubno, line 57: Could you add comment, that it will be deleted by closing calling `buildData.Close()`?

packages/orchestrator/internal/sandbox/build/cache.go
```diff
@@ -73,8 +91,153 @@ func (s *DiffStore) Get(buildId string, diffType DiffType, blockSize int64) (Dif
...
+
+	cancelCh := make(chan struct{})
+	s.pdSizes[key] = &deleteDiff{
+		size:   dSize,
+		cancel: cancelCh,
+	}
+
+	// Delay cache (file close/removal) deletion,
+	// this is to prevent race conditions with exposed slices,
+	// pending data fetching, or data upload
```
jakubno, line 226: If anything access to that data during that time, the delete is cancelled. Do I understand it correctly?

packages/orchestrator/internal/sandbox/build/cache.go
```diff
@@ -73,8 +91,153 @@ func (s *DiffStore) Get(buildId string, diffType DiffType, blockSize int64) (Dif
...
+	var pendingSize int64
+	for _, value := range s.pdSizes {
+		pendingSize += value.size
+	}
+	return pendingSize
+}
+
+// deleteOldestFromCache deletes the oldest item (smallest TTL) from the cache
+// ttlcache has items in order by TTL
+func (s *DiffStore) deleteOldestFromCache() (bool, error) {
```
jakubno, line 169: could you also add test for this? 🙏🏻

packages/orchestrator/internal/sandbox/build/cache.go
```diff
@@ -73,8 +91,153 @@ func (s *DiffStore) Get(buildId string, diffType DiffType, blockSize int64) (Dif
...
+func (s *DiffStore) Add(d Diff) {
+	s.resetDelete(d.CacheKey())
+	s.cache.Set(d.CacheKey(), d, ttlcache.DefaultTTL)
+}
+
+func (s *DiffStore) Has(d Diff) bool {
+	return s.cache.Has(d.CacheKey())
+}
+
+func (s *DiffStore) startDiskSpaceEviction(threshold float64) {
```
jakubno, line 108: this isn't cancelled on close

packages/orchestrator/internal/sandbox/block/cache.go
dobrac: it's not, but it took me quite some time to find it and implement it, can we keep it? 🥺 🙏

packages/orchestrator/internal/sandbox/build/cache.go
dobrac: yes, that should be correct

packages/orchestrator/internal/sandbox/build/cache.go
dobrac: it's canceled on context, but I'll add also on close

packages/orchestrator/internal/sandbox/build/cache.go
dobrac: it's an internal method which is transitively tested somewhere else, but sure

PR #254, opened 2025-01-23
Add condition to instant replace the orchestrator on staging

jakubno:
This should make development faster and easier

### DIFF
--- a/packages/nomad/main.tf
+++ b/packages/nomad/main.tf
@@ -186,26 +186,29 @@ data "google_compute_machine_types" "client" {
   filter = "name = \"${var.client_machine_type}\""
 }
 
-resource "nomad_job" "orchestrator" {
-  jobspec = file("${path.module}/orchestrator.hcl")
+data "external" "orchestrator_checksum" {
+  program = ["bash", "${path.module}/checksum.sh"]
 
-  hcl2 {
-    vars = {
-      gcp_zone     = var.gcp_zone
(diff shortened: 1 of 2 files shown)


packages/nomad/main.tf
```diff
@@ -186,28 +186,31 @@ data "google_compute_machine_types" "client" {
...
-  jobspec = file("${path.module}/orchestrator.hcl")
-
-  hcl2 {
-    vars = {
-      gcp_zone     = var.gcp_zone
-      port         = var.orchestrator_port
-      environment  = var.environment
-      consul_token = var.consul_acl_token_secret
-      cpu_mhz      = var.environment == "prod" ? floor(data.google_compute_machine_types.client.machine_types[0].guest_cpus * 1.5) * 1000 : 1000
-      memory_mb    = var.environment == "prod" ? floor(data.google_compute_machine_types.client.machine_types[0].memory_mb * 0.6 / 1024) * 1024 : 1024
```
jakubno, line 199: I removed these 2 because, they are not doing anything, just causing troubles, if you would like to change machine size and do something like opportunistic deploy, you can't

packages/nomad/orchestrator.hcl
jakubno, line 1: I replaced nomad template variables with terraform templating, it should solve some weird errors we are seeing (when the zone changes?) and also terraform allows to use `if` statements

ValentaTomas (approved):

PR #255, opened 2025-01-24
Fix mmap flush to the filesystem

dobrac:
Flushing mmap with prot "unix.PROT_EXEC" doesn't flush the bytes to the filesystem file. This PR is trying to address that.

Size of sparse file on disk can be verified using command `du fileName` in build cache. Manually tested by successful start of a new sandbox.

### DIFF
--- a/packages/orchestrator/internal/sandbox/block/cache.go
+++ b/packages/orchestrator/internal/sandbox/block/cache.go
@@ -39,7 +39,7 @@ func NewCache(size, blockSize int64, filePath string, dirtyFile bool) (*Cache, e
 		return nil, fmt.Errorf("error allocating file: %w", err)
 	}
 
-	mm, err := mmap.MapRegion(f, int(size), unix.PROT_READ|unix.PROT_WRITE, mmap.RDWR|mmap.EXEC, 0)
+	mm, err := mmap.MapRegion(f, int(size), unix.PROT_READ|unix.PROT_WRITE, 0, 0)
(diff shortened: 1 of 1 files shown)


packages/orchestrator/internal/sandbox/block/cache.go
```diff
@@ -39,7 +39,7 @@ func NewCache(size, blockSize int64, filePath string, dirtyFile bool) (*Cache, e
 		return nil, fmt.Errorf("error allocating file: %w", err)
 	}
 
-	mm, err := mmap.MapRegion(f, int(size), unix.PROT_READ|unix.PROT_WRITE|unix.PROT_EXEC, mmap.RDWR|mmap.EXEC, 0)
+	mm, err := mmap.MapRegion(f, int(size), unix.PROT_READ|unix.PROT_WRITE, mmap.RDWR|mmap.EXEC, 0)
```
ValentaTomas: It might be possible to remove `mmap.EXEC` too.

ValentaTomas: Per discussion on call, the third argument with `mmap.RDWR` might also have no effect here.

dobrac: Updated

ValentaTomas:
I think it is ready to merge, I'll just do some additional tests.

ValentaTomas (approved):

PR #257, opened 2025-01-28
Add autopause handling to the API

ValentaTomas:
This PR adds autopause to the API—when you create or resume sandbox with the `autoPause` flag (opt-in), instead of the sandbox being killed on timeout, it will be paused.


ValentaTomas:
All functionality should be in place, but this is not tested yet.

packages/api/internal/orchestrator/cache.go
```diff
@@ -117,8 +117,14 @@ func (o *Orchestrator) syncNode(ctx context.Context, node *Node, nodes []*node.N
 	node.SyncBuilds(builds)
 }
 
-func (o *Orchestrator) getDeleteInstanceFunction(ctx context.Context, posthogClient *analyticscollector.PosthogClient, logger *zap.SugaredLogger) func(info instance.InstanceInfo) error {
+func (o *Orchestrator) getDeleteInstanceFunction(
+	ctx context.Context,
```
ValentaTomas: I think the ctx duration here should be at least somehow limited, because otherwise request can get stuck infinitely. It can be significant time like 5 minutes, but them just hanging is a problem because they hog resources and we won't see errors from them.

dobrac: I don't see any awaiting in getDeleteInstanceFunction, where it could get stuck?

ValentaTomas: On the `err = o.PauseInstance(ctx, o.tracer, &info, *info.TeamID)` for example— we are using context in the body of this function, but the context we are passing here is the context of the whole API server, which effectively won't end.

ValentaTomas: In the normal request it is limited by the server request ctx timeout, but not for the eviction, etc I think.

dobrac: I see, so basically limit the communication between API and Orchestrator by some timeout, I think it makes sense

PR #245, opened 2025-01-17
API: Sandbox routing: Use Redis for client-proxy DNS instead of local map.

jaytaylor:
*API: Sandbox routing:*

Use Redis for client-proxy DNS instead of local map.

It also cleans up logging a bit to be more consistent and informative about activity.

<details open>
<summary>:sparkles: <i><h3>Description by Callstackai</h3></i></summary>

This PR refactors the DNS handling in the API to use Redis for client-proxy DNS instead of a local map. It also improves logging consistency and adds a Redis job specification.

<details>
<summary><b>Diagrams of code changes</b></summary>

```mermaid
sequenceDiagram
    participant Client
    participant DNS Server
    participant Redis
    participant Orchestrator
    participant Node

    Note over DNS Server: New DNS service with Redis integration

    alt Add DNS Entry
        Orchestrator->>DNS Server: Add(sandboxID, ip)
        DNS Server->>Redis: Set key-value with 24h expiration
    end

    alt DNS Query Flow
        Client->>DNS Server: Query(sandboxID)
        DNS Server->>Redis: Get(sandboxID)
        alt Found in Redis
            Redis-->>DNS Server: Return IP
        else Not Found in Redis
            DNS Server->>Orchestrator: FallbackResolver(sandboxID)
            Orchestrator-->>DNS Server: Return IP from Node
            DNS Server->>Redis: Cache IP (async)
        end
        DNS Server-->>Client: Return IP Address
    end

    alt Remove DNS Entry
        Orchestrator->>DNS Server: Remove(sandboxID)
        DNS Server->>Redis: Delete key
    end
```

</details>

<details>
<summary><b>Files Changed</b></summary>
<table>
<tr><th>File</th><th>Summary</th></tr>
<tr><td><a href=https://github.com/e2b-dev/infra/pull/245/files#diff-ad2c33b49112c808ea24699b01e9e020735e894c9438ace109477b3e6098cc43>.terraform.lock.hcl</a></td><td>Updated provider constraints and added new hashes.</td></tr>
<tr><td><a href=https://github.com/e2b-dev/infra/pull/245/files#diff-961fb1f2c35561d42542795a834f81078ed60fb9e7606a558c96fc8274da83ea>packages/api/internal/dns/server.go</a></td><td>Refactored DNS handling to use Redis for storage and improved logging.</td></tr>
[...]

PR #258, opened 2025-01-28
nomad: add configuration for deploying redis service

tychoish:
This pulls from #245 but moves the redis service off of the same machines as the API service and onto a separate machine that can be accessed from different services.

This is pretty primitive, but good to keep is separate from changes to the API server code.

Currently, there is only one API server so the previous approach is to colocate the redis server and the API server would work, but wouldn't as soon as we had more than one machine running the API service. I'm also not sure that if we did a rolling restart of the API service (with one instance,) if it would necessarily continue to run on the same host. Regardless someone with more broad context for nomad and our nomad deployment might be more well suited to ironing these details out.

Apart from this we are discussing if we want to run a redis service ourselves or use a vendor-hosted service. (Either GCP or Redis Labs)

### DIFF
--- a/packages/nomad/api.hcl
+++ b/packages/nomad/api.hcl
@@ -83,6 +83,12 @@ variable "otel_collector_grpc_endpoint" {
   default = ""
 }
 
+variable "redis_url" {
+  type = string
+  default = ""
+}
+
+
 job "api" {
   datacenters = [var.gcp_zone]
   node_pool = "api"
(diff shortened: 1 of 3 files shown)


packages/nomad/api.hcl
```diff
@@ -143,6 +143,16 @@ job "api" {
...
+
+      template {
+	env = true
+	destination = "local/env.txt"
+	data = <<EOF
+{{ range nomadService "redis" }}
+REDIS_URL=redis://{{ .Address }}:{{ . Port }}
+{{ end }}
+EOF
+      }
```
jakubno: you can use `redis.service.consul` to dynamically resolve the IP. I am not sure if redeploy of redis wouldn't cause redeploy of api in this case

jakubno (changes requested):

packages/nomad/api.hcl
tychoish: I think I fixed it in a way that's more idiomatic.

jakubno (approved):

PR #260, opened 2025-01-28
api(minor): clean up several logging related oddities

tychoish:
I stumbled across these logging issues as I was working, and thought it was easier to push a small patch than outline them in another way. 

The time format defined in the logging encoders was RFC3339 (or equivalent given that it was always UTC.) and there's already a library function for this operation.

I didn't change [this occurance](https://github.com/tychoish/infra/blob/logging-formatting-cleanup/packages/api/internal/handlers/sandbox.go#L105) which appears to mostly be for human consumption (and therefore the extra spaces are desireable. Having said that, this almost certianly should be coerced into UTC.)

### DIFF
--- a/packages/api/main.go
+++ b/packages/api/main.go
@@ -248,7 +248,7 @@ func main() {
 			log.Printf("http service (%d) shutdown error: %v", port, err)
 		}
 	case <-ctx.Done():
-		log.Println("http service (%d) shutdown outside of signal", port)
+		log.Printf("http service (%d) shutdown outside of signal", port)
 	}
 
 	// call cleanup explicitly because defers do not run on os.Exit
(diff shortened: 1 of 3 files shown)


packages/api/main.go
```diff
@@ -248,7 +248,7 @@ func main() {
 			log.Printf("http service (%d) shutdown error: %v", port, err)
 		}
 	case <-ctx.Done():
-		log.Println("http service (%d) shutdown outside of signal", port)
+		log.Printf("http service (%d) shutdown outside of signal", port)
```
jakubno:
```suggestion
		log.Printf("http service (%d) shutdown outside of signal\n", port)
```

jakubno (approved):

jakubno:
There's no standard for logging now, we are just finishing this internally.

PR #261, opened 2025-01-28
api: implement redis cache lookups for dns when configured

tychoish:
This is still a draft but there are a couple of questions in TODOs, and I wanted to just validate that I put the logic in a reasonable way before I cycle back to this later in the week.


packages/api/internal/dns/server.go
```diff
@@ -63,35 +81,127 @@ func (d *DNS) handleDNSRequest(w resolver.ResponseWriter, r *resolver.Msg) {
...
+				res, err := d.redis.Get(ctx, d.getCacheKey(sandboxID)).Result()
+				if err == nil {
+					// TODO: do we need to do
+					// anything with the error or
+					// distinguish between "can't
+					// find" and "server error"
+
+					ip = net.ParseIP(res)
+
+					// TODO: should we add this to
```
jakubno: Can you test how fast is the Redis request, if there's like 1000 keys? If we don't store in cache, it's probably easier to reason about, but is it worth the extra time? Every request to sandbox will do DNS request.

packages/api/internal/dns/server.go
```diff
@@ -2,99 +2,242 @@ package dns
...
-		return v == ip
-	})
+func (d *DNS) Remove(ctx context.Context, sandboxID, ip string) {
+	switch {
+	case d.remote != nil:
+		if err := d.remote.Delete(ctx, d.cacheKey(sandboxID)); err != nil {
+			d.logger.Debug("removing item from DNS cache", zap.Error(err), zap.String("sandbox", sandboxID))
+		}
+	case d.local != nil:
+		d.local.RemoveCb(d.cacheKey(sandboxID), func(k string, v string, ok bool) bool { return v == ip })
```
tychoish, line 75: this "check the value of the IP" before deleting things, is previous behavior. I'm not sure what it's trying to protect against, and if it's worth retaining. I think we probably should either unconditionally remove keys from the cache or add this additional check to the remote case.

packages/api/internal/dns/server.go
```diff
@@ -2,99 +2,242 @@ package dns
...
 }
 
-func (*DNS) hostname(sandboxID string) string {
-	return fmt.Sprintf("%s.", sandboxID)
+func (d *DNS) cacheKey(id string) string {
+	switch {
+	case d.remote != nil:
+		return fmt.Sprintf("%s%s", cachedDnsPrefix, id)
+	case d.local != nil:
+		return fmt.Sprintf("%s.", id)
```
tychoish: adding the terminal dot for the cache keys is a thing that happened previously, and I've retained that: the "hostname" value isn't used outside of the cache, so I'm not sure what benefit we get from this, but if it's cruft we should remove.

packages/api/internal/dns/server.go
```diff
@@ -2,99 +2,242 @@ package dns
...
+		}
+	}
+
+	addr := net.ParseIP(res)
+	if addr == nil {
+		if res != "" {
+			d.logger.Error("malformed address in cache", zap.Bool("local", d.local != nil), zap.String("addr", res))
+		}
+
+		addr = net.ParseIP(defaultRoutingIP)
```
tychoish, line 104: I put the "fall back to localhost" in the Get method from the cache, where it was previously in the handler.

packages/api/internal/orchestrator/orchestrator.go
```diff
@@ -32,23 +33,23 @@ func New(
...
-		logger.Info("Running locally, skipping starting DNS server")
+		fmt.Printf("Running locally, skipping starting DNS server\n")
 	} else {
 		go func() {
-			logger.Info("Starting DNS server")
+			fmt.Printf("Starting DNS server\n")
 
-			if err := dnsServer.Start(ctx, "127.0.0.4", 53); err != nil {
-				logger.Panic("Failed starting DNS server", zap.Error(err))
+			dnsErr := dnsServer.Start(ctx, "127.0.0.4", 53)
```
jakubno: The type doesn't match, there's no error

packages/api/internal/orchestrator/orchestrator.go
```diff
@@ -32,23 +33,23 @@ func New(
...
-	redisClient *redis.Client,
 ) (*Orchestrator, error) {
 	analyticsInstance, err := analyticscollector.NewAnalytics()
 	if err != nil {
-		logger.Error("Error initializing Analytics client", zap.Error(err))
+		logger.Errorf("Error initializing Analytics client\n: %v", err)
 	}
 
-	dnsServer := dns.New(redisClient)
+	dnsServer := dns.New()
```
jakubno: The type doesn't match, you should pass variables here

packages/api/internal/dns/server.go
jakubno: We introduce pause possibility, which can cause the same `sandboxID` is used multiple times and there were some issues with resolving the hostname. We though there's some race condition, but turned out it's nginx issue

packages/api/internal/dns/server.go
jakubno: I think it's based on `.` used in `dig`

packages/api/internal/dns/server.go
```diff
@@ -2,99 +2,242 @@ package dns
...
-func (*DNS) hostname(sandboxID string) string {
-	return fmt.Sprintf("%s.", sandboxID)
+func (d *DNS) cacheKey(id string) string {
+	switch {
+	case d.remote != nil:
+		return fmt.Sprintf("%s%s", cachedDnsPrefix, id)
+	case d.local != nil:
+		return fmt.Sprintf("%s.", id)
+	default:
+		return id
```
jakubno, line 123: Could you just add comment here 🙏🏻

packages/api/internal/dns/server.go
```diff
@@ -2,99 +2,242 @@ package dns
...
+				// the service to continue without any
+				// DNS service.
+				panic(errors.Join(errors.New("problem starting DNS service"), err, errOnStartup))
+			case "server already started":
+				// this only happens if you call start
+				// more than once, which shouldn't be
+				// possible.
+				errChan <- errors.Join(err, errOnStartup)
+			default:
+				// this should only happen if we
```
jakubno, line 198: I think this can happen when the port is already used

jakubno (changes requested):

packages/api/internal/dns/server.go
```diff
@@ -2,99 +2,242 @@ package dns
...
 
-func (d *DNS) Add(_ context.Context, sandboxID, ip string) {
-	d.records.Insert(d.hostname(sandboxID), ip)
+func (d *DNS) Add(ctx context.Context, sandboxID, ip string) {
+	switch {
+	case d.remote != nil:
+		d.remote.Set(&cache.Item{
+			Ctx:   ctx,
+			TTL:   redisTTL,
+			Key:   sandboxID,
```
jakubno:
```suggestion
			Key:   d.cacheKey(sandboxID),
```

packages/api/internal/dns/server.go
jakubno: Okay this should explain it
image

jakubno: Looking more into it, I would strip it in `handleDNSRequest` and then remove it everywhere, it's just confusing

packages/api/internal/dns/server.go
tychoish: do we need to preserve this behavior? If so I'd like to add a comment just so our future selves don't get confused by this again...

packages/api/internal/dns/server.go
tychoish: I think the errors you saw were because I wasn't putting the cacheKey on the on when inserting into the remote cache.

packages/api/internal/dns/server.go
jakubno: No, we don't want to preserve this behavior

tychoish: cool, it's been removed in my latest push.

packages/api/internal/dns/server.go
tychoish: I guess the question is "should we panic here"

if this is a "port in use error" then it's an initialization error, panicking makes sense. By not panicking we let a process run in a (potentially bad state.)

If there's a problem that happens later and the service aborts early then the process panics at an arbitrary point. This seems unlikely, but there's not a lot we can do except for... maybe a retry loop?

jakubno: I think we should panic, it will look healthy, but it won't be doing one of the most important functionalities

packages/api/internal/orchestrator/orchestrator.go
```diff
@@ -80,19 +77,27 @@ func New(
...
 
-func (o *Orchestrator) Close() error {
-	var err error
-	for _, node := range o.nodes.Items() {
-		closeErr := node.Client.Close()
-		if closeErr != nil {
-			err = errors.Join(err, closeErr)
+func (o *Orchestrator) Close(ctx context.Context) error {
+	nodes := o.nodes.Items()
+	errs := make([]error, 0, len(nodes)+2)
```
jakubno: do you think this is better? I would expect that in 99% it will be empty. Also if somebody adds some new close with potential error, I am little doubtful they will increase the number

jakubno (approved):

PR #264, opened 2025-01-29
Add paused sandboxes to the list of running sandboxes

mishushakov:
Includes some changes for: [E2B-1494: Update snapshot metadata and timestamp on subsequent pauses](https://linear.app/e2b/issue/E2B-1494/update-snapshot-metadata-and-timestamp-on-subsequent-pauses)


mishushakov:
found a bug, sorry, 1s

spec/openapi.yml
```diff
@@ -229,6 +229,7 @@ components:
         - cpuCount
         - memoryMB
         - endAt
+        - state
```
jakubno, line 262: It's little bit weird `RunningSandbox` can have state `paused`

packages/api/internal/handlers/sandbox_get.go
```diff
@@ -24,50 +24,60 @@ func (a *APIStore) GetSandboxesSandboxID(c *gin.Context, id string) {
 
 	sandboxId := strings.Split(id, "-")[0]
 
+	// Try to get the running instance first
 	info, err := a.orchestrator.GetInstance(ctx, sandboxId)
-	if err != nil {
-		c.JSON(http.StatusNotFound, fmt.Sprintf("instance \"%s\" doesn't exist or you don't have access to it", id))
+	if err == nil && *info.TeamID == team.ID {
```
jakubno: You should return if team ids don't match (you will, but there's unnecessary query and it's rather confusing

packages/api/internal/handlers/sandboxes_list.go
```diff
@@ -27,66 +27,18 @@ func (a *APIStore) GetSandboxes(c *gin.Context, params api.GetSandboxesParams) {
...
+		instanceSandboxIDs = append(instanceSandboxIDs, info.Instance.SandboxID)
+	}
 
-			if matchesAll {
-				instanceInfo[n] = instance
-				n++
-			}
-		}
+	// all snapshots where env team is same as team.ID and sandbox_id is not included in instanceInfo.SandboxID
+	snapshotEnvs, err := a.db.GetTeamSnapshots(ctx, team.ID)
```
jakubno: We will have to do some pagination in future, this can get pretty big

packages/api/internal/handlers/sandboxes_list.go
```diff
@@ -152,6 +105,118 @@ func (a *APIStore) GetSandboxes(c *gin.Context, params api.GetSandboxesParams) {
...
+				n++
+			}
+		}
+
+		// Trim slice
+		sandboxes = sandboxes[:n]
+	}
+
+	// filter sandboxes by state
+	if params.State != nil {
```
jakubno: I think you should handle it differently, check the state at start and do only the work that's needed, listing all snapshots if you don't need them can be very wasteful

jakubno (changes requested):

packages/api/internal/handlers/sandbox_get.go
```diff
@@ -24,50 +24,66 @@ func (a *APIStore) GetSandboxesSandboxID(c *gin.Context, id string) {
...
-		c.JSON(http.StatusNotFound, fmt.Sprintf("instance \"%s\" doesn't exist or you don't have access to it", id))
+	if err == nil {
+		// Check if instance belongs to the team
+		if *info.TeamID != team.ID {
+			c.JSON(http.StatusNotFound, fmt.Sprintf("instance \"%s\" doesn't exist or you don't have access to it", id))
+			return
+		}
+
+		// Instance exists and belongs to the team - return running sandbox info
+		build, err := a.db.Client.EnvBuild.Query().Where(envbuild.ID(*info.BuildID)).First(ctx)
```
jakubno: the check if build is not `nil` was there for reason you already deleted the template. It will result in `panic`

packages/api/internal/handlers/sandboxes_list.go
```diff
@@ -19,14 +19,105 @@ import (
...
+				}
+
+				instance := api.ListedSandbox{
+					ClientID:   info.Instance.ClientID,
+					TemplateID: info.Instance.TemplateID,
+					Alias:      info.Instance.Alias,
+					SandboxID:  info.Instance.SandboxID,
+					StartedAt:  info.StartTime,
+					CpuCount:   int32(buildsMap[*info.BuildID].Vcpu),
+					MemoryMB:   int32(buildsMap[*info.BuildID].RAMMB),
```
jakubno: you should check build is not `nil` if the user deleted template it can happen the the build won't be present

packages/api/internal/handlers/sandboxes_list.go
```diff
@@ -66,96 +157,59 @@ func (a *APIStore) GetSandboxes(c *gin.Context, params api.GetSandboxesParams) {
...
-		for _, instance := range instanceInfo {
+		for _, instance := range sandboxes {
 			if instance.Metadata == nil {
 				continue
 			}
 
 			matchesAll := true
 			for key, value := range filters {
-				if metadataValue, ok := instance.Metadata[key]; !ok || metadataValue != value {
+				if metadataValue, ok := (*instance.Metadata)[key]; !ok || metadataValue != value {
```
jakubno: what if `instance.Metadata` is nil

packages/api/internal/handlers/sandboxes_list.go
```diff
@@ -66,96 +157,59 @@ func (a *APIStore) GetSandboxes(c *gin.Context, params api.GetSandboxesParams) {
...
-			CpuCount:   int32(buildsMap[*info.BuildID].Vcpu),
-			MemoryMB:   int32(buildsMap[*info.BuildID].RAMMB),
-			EndAt:      info.EndTime,
-		}
-
-		if info.Metadata != nil {
-			meta := api.SandboxMetadata(info.Metadata)
-			instance.Metadata = &meta
+	// filter sandboxes by state
+	if params.State != nil {
```
jakubno: this isn't needed

packages/api/internal/handlers/sandboxes_list.go
```diff
@@ -66,96 +157,59 @@ func (a *APIStore) GetSandboxes(c *gin.Context, params api.GetSandboxesParams) {
...
 	// Sort sandboxes by start time descending
-	slices.SortFunc(sandboxes, func(a, b api.RunningSandbox) int {
+	slices.SortFunc(sandboxes, func(a, b api.ListedSandbox) int {
 		return a.StartedAt.Compare(b.StartedAt)
 	})
 
+	// Report analytics
+	a.posthog.IdentifyAnalyticsTeam(team.ID.String(), team.Name)
+	properties := a.posthog.GetPackageToPosthogProperties(&c.Request.Header)
+	a.posthog.CreateAnalyticsTeamEvent(team.ID.String(), "listed running instances", properties)
```
jakubno: the name of the event isn't true anymore

packages/shared/pkg/db/snapshot.go
```diff
@@ -139,13 +142,44 @@ func (db *DB) GetLastSnapshot(ctx context.Context, sandboxID string, teamID uuid
...
+		Client.
+		Env.
+		Query().
+		Where(
+			env.HasBuildsWith(envbuild.StatusEQ(envbuild.StatusSuccess)),
+			env.TeamID(teamID),
+		).
+		WithSnapshots().
+		WithBuilds(func(query *models.EnvBuildQuery) {
+			query.Where(envbuild.StatusEQ(envbuild.StatusSuccess)).Order(models.Desc(envbuild.FieldFinishedAt))
```
jakubno: wouldn't this work?
```suggestion
			query.Where(envbuild.StatusEQ(envbuild.StatusSuccess)).Order(models.Desc(envbuild.FieldFinishedAt)).Only()
```

packages/shared/pkg/db/snapshot.go
```diff
@@ -139,13 +142,44 @@ func (db *DB) GetLastSnapshot(ctx context.Context, sandboxID string, teamID uuid
...
+		}).
+		All(ctx)
+
+	notFound := models.IsNotFound(err)
+
 	if notFound {
```
jakubno: I think `All()` can't be notFound...

packages/shared/pkg/schema/snapshots.go
```diff
@@ -22,6 +22,7 @@ func (Snapshot) Fields() []ent.Field {
 			Annotations(
 				entsql.Default("CURRENT_TIMESTAMP"),
 			),
+		field.Time("sandbox_started_at"),
```
jakubno: This probably requires migration, also it is not backward compatible is it? Maybe we can fill in date from sandbox logs?

jakubno (changes requested):

packages/shared/pkg/schema/snapshots.go
mishushakov: I can make it nullable, but you can also get the sandbox_started_at date from the billing logs by id?

packages/shared/pkg/db/snapshot.go
mishushakov: For some reason I am seeing a difference:

with .Only()

```json
{
    "clientID": "00000000",
    "cpuCount": -1,
    "endAt": "2025-01-31T22:03:30.613128Z",
    "memoryMB": -1,
    "sandboxID": "imrfja7a6fja50jwrtklc",
    "startedAt": "2025-01-31T22:03:30.399452Z",
    "state": "paused",
    "templateID": "1gny15juzaejkjgw2t9s"
  }
```

without .Only()

```json
  {
    "clientID": "00000000",
    "cpuCount": 2,
    "endAt": "2025-01-31T22:03:30.613128Z",
    "memoryMB": 1024,
    "sandboxID": "imrfja7a6fja50jwrtklc",
    "startedAt": "2025-01-31T22:03:30.399452Z",
    "state": "paused",
    "templateID": "1gny15juzaejkjgw2t9s"
  }
```

(it cannot get the build)

packages/api/internal/handlers/sandboxes_list.go
mishushakov: But we're doing a check on line 161 already? Am I not understanding it correctly?

packages/api/internal/handlers/sandboxes_list.go
mishushakov: wdym? same as was before

https://github.com/e2b-dev/infra/blob/20aff1d6ba1ef1f627e4d71b79f1bc654390150d/packages/api/internal/handlers/sandboxes_list.go#L92-L94

packages/shared/pkg/db/snapshot.go
mishushakov: the build and env exist in the database - I checked, just something messed up with the query

mishushakov: more over, here's the output of

```go
log.Printf("snapshot %s, build count %d", snapshot.ID, len(snapshotBuilds))
```

in the for loop

with .Only()

```
2025/02/06 15:46:13 snapshot 65c5c66e-68c3-4a11-9dad-26ea66382cd5, build count 0
2025/02/06 15:46:13 snapshot bc3df4e0-bb77-4fbe-9c25-548c5c00320c, build count 0
2025/02/06 15:46:13 snapshot c74219c1-5ac8-42b8-8ab5-c995b1a46e98, build count 0
2025/02/06 15:46:13 snapshot ea48706d-fb72-4583-a996-a0cdac974d87, build count 0
2025/02/06 15:46:13 snapshot 23fb5c04-0ade-4302-b129-ef57b6d25cc7, build count 0
2025/02/06 15:46:13 snapshot 5617fc30-dabe-4f8a-801d-b85aae219b92, build count 0
2025/02/06 15:46:13 snapshot 299fc69f-2e3e-482c-a4fb-01f9becfb33a, build count 0
2025/02/06 15:46:13 snapshot 9939891d-d7d8-4f17-8004-2eb3e5d11427, build count 0
2025/02/06 15:46:13 snapshot 59894b9b-73a8-463d-a4fd-52433cf5e61f, build count 0
2025/02/06 15:46:13 snapshot 28777bf3-b675-4251-93a0-af9c0a5393b3, build count 0
2025/02/06 15:46:13 snapshot fde42022-b9ae-4a97-9fd3-a3cb60f42cb5, build count 0
2025/02/06 15:46:13 snapshot d2d55f44-c9b1-4809-98d4-9856a3c48c1d, build count 0
2025/02/06 15:46:13 snapshot b9ab25c9-dac8-4146-9ee2-962caca2b080, build count 0
2025/02/06 15:46:13 snapshot dd1b7a3b-1bc6-4c2b-9601-397c6d0e5e5c, build count 0
2025/02/06 15:46:13 snapshot c4595dea-9d7e-4d6a-9c86-5b45e24d62bb, build count 0
2025/02/06 15:46:13 snapshot 3eb51681-1729-4877-a7a2-b5f171a62cb7, build count 0
2025/02/06 15:46:13 snapshot 6d7b425f-b90f-4451-b290-45c9998a3400, build count 0
2025/02/06 15:46:13 snapshot a0e9815d-f457-4207-bf36-2a1714e5a424, build count 0
2025/02/06 15:46:13 snapshot 3ac2fd20-207e-46d1-9605-58650b735194, build count 0
2025/02/06 15:46:13 snapshot a2a628fe-1448-462f-a625-e2cf166a3e43, build count 0
2025/02/06 15:46:13 snapshot 54aa44f6-f1a1-4a8d-b74b-4d6ffaf4e7c4, build count 0
2025/02/06 15:46:13 snapshot a0e1dbae-38c6-49c9-9911-233dd386c374, build count 0
2025/02/06 15:46:13 snapshot e8296ee8-4645-441c-b6d6-b53e407fa563, build count 0
2025/02/06 15:46:13 snapshot c3e56ebb-c26d-4f34-a1fc-54db01d810a8, build count 0
2025/02/06 15:46:13 snapshot 751b50f7-97d8-45d5-8ab6-f3d35de84b76, build count 0
2025/02/06 15:46:13 snapshot 4a364365-9c7b-4236-bf99-2ad719b92201, build count 1
2025/02/06 15:46:13 snapshot b8817a89-460d-4664-a5a3-2ad2682d4700, build count 0
2025/02/06 15:46:13 snapshot 480f8a1e-28bb-4953-bed6-7c0b9d6344ad, build count 0
2025/02/06 15:46:13 snapshot 42f8cf57-7450-40da-b833-f07e00a33cc5, build count 0
2025/02/06 15:46:13 snapshot af6ee060-4580-4135-8a98-a160c8d951d2, build count 0
2025/02/06 15:46:13 snapshot 46c46612-144d-46fa-b015-64fcc467156b, build count 0
2025/02/06 15:46:13 snapshot 14d050a9-0494-4be8-9dbe-03b57dd242cd, build count 0
2025/02/06 15:46:13 snapshot 2c7ce9f8-fdf6-4adf-8836-d003fb830fe6, build count 0
2025/02/06 15:46:13 snapshot 311b2e2f-11a0-497e-810c-ff4dfe3f40d3, build count 0
2025/02/06 15:46:13 snapshot b9608ac7-617b-46e3-b551-9374d3283e0d, build count 1
```

without .Only()

```
2025/02/06 15:48:11 snapshot 65c5c66e-68c3-4a11-9dad-26ea66382cd5, build count 1
2025/02/06 15:48:11 snapshot bc3df4e0-bb77-4fbe-9c25-548c5c00320c, build count 1
2025/02/06 15:48:11 snapshot c74219c1-5ac8-42b8-8ab5-c995b1a46e98, build count 1
2025/02/06 15:48:11 snapshot ea48706d-fb72-4583-a996-a0cdac974d87, build count 1
2025/02/06 15:48:11 snapshot 23fb5c04-0ade-4302-b129-ef57b6d25cc7, build count 1
2025/02/06 15:48:11 snapshot 5617fc30-dabe-4f8a-801d-b85aae219b92, build count 1
2025/02/06 15:48:11 snapshot 299fc69f-2e3e-482c-a4fb-01f9becfb33a, build count 1
2025/02/06 15:48:11 snapshot 9939891d-d7d8-4f17-8004-2eb3e5d11427, build count 1
2025/02/06 15:48:11 snapshot 59894b9b-73a8-463d-a4fd-52433cf5e61f, build count 2
2025/02/06 15:48:11 snapshot 28777bf3-b675-4251-93a0-af9c0a5393b3, build count 1
2025/02/06 15:48:11 snapshot fde42022-b9ae-4a97-9fd3-a3cb60f42cb5, build count 1
2025/02/06 15:48:11 snapshot d2d55f44-c9b1-4809-98d4-9856a3c48c1d, build count 1
2025/02/06 15:48:11 snapshot b9ab25c9-dac8-4146-9ee2-962caca2b080, build count 1
2025/02/06 15:48:11 snapshot dd1b7a3b-1bc6-4c2b-9601-397c6d0e5e5c, build count 1
2025/02/06 15:48:11 snapshot c4595dea-9d7e-4d6a-9c86-5b45e24d62bb, build count 1
2025/02/06 15:48:11 snapshot 3eb51681-1729-4877-a7a2-b5f171a62cb7, build count 1
2025/02/06 15:48:11 snapshot 6d7b425f-b90f-4451-b290-45c9998a3400, build count 1
2025/02/06 15:48:11 snapshot a0e9815d-f457-4207-bf36-2a1714e5a424, build count 1
2025/02/06 15:48:11 snapshot 3ac2fd20-207e-46d1-9605-58650b735194, build count 1
2025/02/06 15:48:11 snapshot a2a628fe-1448-462f-a625-e2cf166a3e43, build count 1
2025/02/06 15:48:11 snapshot 54aa44f6-f1a1-4a8d-b74b-4d6ffaf4e7c4, build count 1
2025/02/06 15:48:11 snapshot a0e1dbae-38c6-49c9-9911-233dd386c374, build count 1
2025/02/06 15:48:11 snapshot e8296ee8-4645-441c-b6d6-b53e407fa563, build count 1
2025/02/06 15:48:11 snapshot c3e56ebb-c26d-4f34-a1fc-54db01d810a8, build count 1
2025/02/06 15:48:11 snapshot 751b50f7-97d8-45d5-8ab6-f3d35de84b76, build count 1
2025/02/06 15:48:11 snapshot 4a364365-9c7b-4236-bf99-2ad719b92201, build count 1
2025/02/06 15:48:11 snapshot b8817a89-460d-4664-a5a3-2ad2682d4700, build count 1
2025/02/06 15:48:11 snapshot 480f8a1e-28bb-4953-bed6-7c0b9d6344ad, build count 1
2025/02/06 15:48:11 snapshot 42f8cf57-7450-40da-b833-f07e00a33cc5, build count 1
2025/02/06 15:48:11 snapshot af6ee060-4580-4135-8a98-a160c8d951d2, build count 1
2025/02/06 15:48:11 snapshot 46c46612-144d-46fa-b015-64fcc467156b, build count 1
2025/02/06 15:48:11 snapshot 14d050a9-0494-4be8-9dbe-03b57dd242cd, build count 1
2025/02/06 15:48:11 snapshot 2c7ce9f8-fdf6-4adf-8836-d003fb830fe6, build count 1
2025/02/06 15:48:11 snapshot 311b2e2f-11a0-497e-810c-ff4dfe3f40d3, build count 1
2025/02/06 15:48:11 snapshot b9608ac7-617b-46e3-b551-9374d3283e0d, build count 1
```

I don't know why this happens

packages/shared/pkg/schema/snapshots.go
jakubno: The only problem is with paused sandboxes, it would require some non trivial matching

packages/api/internal/handlers/sandboxes_list.go
jakubno: sorry, i missed that

packages/api/internal/handlers/sandboxes_list.go
jakubno: `running` just NIT

packages/shared/pkg/schema/snapshots.go
mishushakov: I made it nillable for now at least

packages/api/internal/handlers/sandbox_get.go
```diff
@@ -24,26 +24,57 @@ func (a *APIStore) GetSandboxesSandboxID(c *gin.Context, id string) {
```
jakubno, line 24: Can you change the event above? 🙏🏻

packages/api/internal/handlers/sandbox_get.go
```diff
@@ -24,26 +24,57 @@ func (a *APIStore) GetSandboxesSandboxID(c *gin.Context, id string) {
...
-		c.JSON(http.StatusNotFound, fmt.Sprintf("instance \"%s\" doesn't exist or you don't have access to it", id))
-		return
-	}
-
-	if *info.TeamID != team.ID {
-		c.JSON(http.StatusNotFound, fmt.Sprintf("instance \"%s\" doesn't exist or you don't have access to it", id))
+	if err == nil {
+		// Check if instance belongs to the team
+		if *info.TeamID != team.ID {
+			c.JSON(http.StatusNotFound, fmt.Sprintf("instance \"%s\" doesn't exist or you don't have access to it", id))
```
jakubno: Let's use sandbox 🙏🏻 
```suggestion
			c.JSON(http.StatusNotFound, fmt.Sprintf("sandbox \"%s\" doesn't exist or you don't have access to it", id))
```

packages/api/internal/handlers/sandbox_get.go
```diff
@@ -24,26 +24,57 @@ func (a *APIStore) GetSandboxesSandboxID(c *gin.Context, id string) {
...
+			return
+		}
+
+		// Instance exists and belongs to the team - return running sandbox info
```
jakubno:
```suggestion
		// Sandbox exists and belongs to the team - return running sandbox info
```

packages/api/internal/handlers/sandbox_get.go
```diff
@@ -24,26 +24,57 @@ func (a *APIStore) GetSandboxesSandboxID(c *gin.Context, id string) {
...
+		build, err := a.db.Client.EnvBuild.Query().Where(envbuild.ID(*info.BuildID)).First(ctx)
+		if err != nil {
+			telemetry.ReportCriticalError(ctx, err)
```
jakubno: can we start logging on error? 🙏🏻

packages/api/internal/handlers/sandbox_get.go
```diff
@@ -24,26 +24,57 @@ func (a *APIStore) GetSandboxesSandboxID(c *gin.Context, id string) {
...
+		c.JSON(http.StatusOK, instance)
 		return
 	}
 
-	a.posthog.IdentifyAnalyticsTeam(team.ID.String(), team.Name)
-	properties := a.posthog.GetPackageToPosthogProperties(&c.Request.Header)
-	a.posthog.CreateAnalyticsTeamEvent(team.ID.String(), "get running instance", properties)
-
-	build, err := a.db.Client.EnvBuild.Query().Where(envbuild.ID(*info.BuildID)).First(ctx)
+	// If instance not found try to get the latest snapshot
```
jakubno:
```suggestion
	// If sandbox not found try to get the latest snapshot
```

packages/api/internal/handlers/sandbox_get.go
```diff
@@ -24,26 +24,57 @@ func (a *APIStore) GetSandboxesSandboxID(c *gin.Context, id string) {
...
+	snapshot, build, err := a.db.GetLastSnapshot(ctx, sandboxId, team.ID)
 	if err != nil {
-		telemetry.ReportCriticalError(ctx, err)
-		c.JSON(http.StatusInternalServerError, fmt.Sprintf("Error getting build for instance %s", id))
-
+		fmt.Println(err)
```
jakubno: use `a.logger`

packages/api/internal/handlers/sandbox_get.go
```diff
@@ -55,19 +86,20 @@ func (a *APIStore) GetSandboxesSandboxID(c *gin.Context, id string) {
 		cpuCount = int32(build.Vcpu)
 	}
 
-	instance := api.RunningSandbox{
-		ClientID:   info.Instance.ClientID,
-		TemplateID: info.Instance.TemplateID,
-		Alias:      info.Instance.Alias,
-		SandboxID:  info.Instance.SandboxID,
-		StartedAt:  info.StartTime,
+	// optional
```
jakubno: What's optional about this? Also the type for `snapshot.SandboxStartedAt` is incorrect, `ListedSandbox` expects `StartedAt` defined

packages/api/internal/handlers/sandbox_get.go
jakubno, line 1: This isn't used anywhere yet, right? Because it would be backward incompatible

packages/api/internal/handlers/sandboxes_list.go
```diff
@@ -19,14 +20,119 @@ import (
...
+			if info.TeamID != nil && *info.TeamID == team.ID && info.BuildID != nil {
+				buildIDs = append(buildIDs, *info.BuildID)
+			}
+		}
+
+		// Only fetch builds if we have running instances
+		if len(buildIDs) > 0 {
+			builds, err := a.db.Client.EnvBuild.Query().Where(envbuild.IDIn(buildIDs...)).All(ctx)
+			if err != nil {
+				telemetry.ReportCriticalError(ctx, err)
```
jakubno: can you log the error? 🙏🏻

packages/api/internal/handlers/sandboxes_list.go
```diff
@@ -19,14 +20,119 @@ import (
...
+				instance := api.ListedSandbox{
+					ClientID:   info.Instance.ClientID,
+					TemplateID: info.Instance.TemplateID,
+					Alias:      info.Instance.Alias,
+					SandboxID:  info.Instance.SandboxID,
+					StartedAt:  info.StartTime,
+					CpuCount:   cpuCount,
+					MemoryMB:   memoryMB,
+					EndAt:      info.EndTime,
+					State:      "running",
```
jakubno:
```suggestion
					State:      api.ListedSandboxStateRunning,
```

packages/api/internal/handlers/sandboxes_list.go
```diff
@@ -19,14 +20,119 @@ import (
...
+
+			instance := api.ListedSandbox{
+				ClientID:   "00000000",
+				TemplateID: e.ID,
+				SandboxID:  snapshot.SandboxID,
+				StartedAt:  snapshot.SandboxStartedAt,
+				CpuCount:   cpuCount,
+				MemoryMB:   memoryMB,
+				EndAt:      snapshot.CreatedAt,
+				State:      "paused",
```
jakubno:
```suggestion
				State:      api.ListedSandboxStatePaused,
```

packages/api/internal/handlers/sandboxes_list.go
```diff
@@ -19,14 +20,119 @@ import (
...
+			memoryMB := int32(-1)
+			cpuCount := int32(-1)
+
+			if len(snapshotBuilds) > 0 {
+				memoryMB = int32(snapshotBuilds[0].RAMMB)
+				cpuCount = int32(snapshotBuilds[0].Vcpu)
+			}
+
+			instance := api.ListedSandbox{
+				ClientID:   "00000000",
```
jakubno: Can you add comment with explanation? 🙏🏻

packages/api/internal/handlers/sandbox_get.go
```diff
@@ -55,19 +86,20 @@ func (a *APIStore) GetSandboxesSandboxID(c *gin.Context, id string) {
...
+	instance := api.ListedSandbox{
+		ClientID:   "00000000",
```
jakubno: Can you add comment with explanation? 🙏🏻

packages/api/internal/handlers/sandbox_get.go
```diff
@@ -55,19 +86,20 @@ func (a *APIStore) GetSandboxesSandboxID(c *gin.Context, id string) {
...
+		TemplateID: snapshot.EnvID,
+		SandboxID:  snapshot.SandboxID,
+		StartedAt:  snapshot.SandboxStartedAt,
 		CpuCount:   cpuCount,
 		MemoryMB:   memoryMB,
-		EndAt:      info.EndTime,
+		EndAt:      snapshot.CreatedAt,
+		State:      "paused",
```
jakubno:
```suggestion
		State:      api.ListedSandboxStatePaused,
```

packages/api/internal/handlers/sandbox_get.go
```diff
@@ -24,26 +24,57 @@ func (a *APIStore) GetSandboxesSandboxID(c *gin.Context, id string) {
...
+		instance := api.ListedSandbox{
+			ClientID:   info.Instance.ClientID,
+			TemplateID: info.Instance.TemplateID,
+			Alias:      info.Instance.Alias,
+			SandboxID:  info.Instance.SandboxID,
+			StartedAt:  info.StartTime,
+			CpuCount:   cpuCount,
+			MemoryMB:   memoryMB,
+			EndAt:      info.EndTime,
+			State:      "running",
```
jakubno:
```suggestion
			State:      api.ListedSandboxStateRunning,
```

packages/api/internal/handlers/sandboxes_list.go
```diff
@@ -19,14 +20,119 @@ import (
...
+
+		// Get build IDs for running instances
+		buildIDs := make([]uuid.UUID, 0)
+		for _, info := range instanceInfo {
+			if info.TeamID != nil && *info.TeamID == team.ID && info.BuildID != nil {
+				buildIDs = append(buildIDs, *info.BuildID)
+			}
+		}
+
+		// Only fetch builds if we have running instances
```
jakubno:
```suggestion
		// Only fetch builds if we have running sandboxes
```

packages/api/internal/handlers/sandboxes_list.go
```diff
@@ -19,14 +20,119 @@ import (
...
 	teamInfo := c.Value(auth.TeamContextKey).(authcache.AuthTeamInfo)
 	team := teamInfo.Team
 
 	telemetry.ReportEvent(ctx, "list running instances")
 
-	instanceInfo := a.orchestrator.GetSandboxes(ctx, &team.ID)
+	// Initialize empty slice for results
+	sandboxes := make([]api.ListedSandbox, 0)
+
+	// Only fetch running instances if we need them (state is nil or "running")
```
jakubno:
```suggestion
	// Only fetch running sandboxes if we need them (state is nil or "running")
```

packages/api/internal/handlers/sandboxes_list.go
```diff
@@ -19,14 +20,119 @@ import (
...
+				telemetry.ReportCriticalError(ctx, err)
+				return
+			}
+
+			buildsMap := make(map[uuid.UUID]*models.EnvBuild, len(builds))
+			for _, build := range builds {
+				buildsMap[build.ID] = build
+			}
+
+			// Add running instances to results
```
jakubno:
```suggestion
			// Add running sandboxes to results
```

packages/api/internal/handlers/sandboxes_list.go
```diff
@@ -19,14 +20,119 @@ import (
...
+					instance.Metadata = &meta
+				}
+
+				sandboxes = append(sandboxes, instance)
+			}
+		}
+	}
+
+	// Only fetch snapshots if we need them (state is nil or "paused")
+	if params.State == nil || *params.State == "paused" {
```
jakubno:
```suggestion
	if params.State == nil || *params.State == api.ListedSandboxStatePaused {
```

packages/api/internal/handlers/sandboxes_list.go
```diff
@@ -19,14 +20,119 @@ import (
...
 	team := teamInfo.Team
 
 	telemetry.ReportEvent(ctx, "list running instances")
 
-	instanceInfo := a.orchestrator.GetSandboxes(ctx, &team.ID)
+	// Initialize empty slice for results
+	sandboxes := make([]api.ListedSandbox, 0)
+
+	// Only fetch running instances if we need them (state is nil or "running")
+	if params.State == nil || *params.State == "running" {
```
jakubno:
```suggestion
	if params.State == nil || *params.State == api.ListedSandboxStateRunning {
```

packages/api/internal/handlers/sandboxes_list.go
```diff
@@ -66,96 +172,38 @@ func (a *APIStore) GetSandboxes(c *gin.Context, params api.GetSandboxesParams) {
 
 		// Filter instances to match all filters
```
jakubno:
```suggestion
		// Filter sandboxes to match all filters
```

packages/api/internal/handlers/sandboxes_list.go
```diff
@@ -66,96 +172,38 @@ func (a *APIStore) GetSandboxes(c *gin.Context, params api.GetSandboxesParams) {
...
 		n := 0
-		for _, instance := range instanceInfo {
+		for _, instance := range sandboxes {
```
jakubno:
```suggestion
		for _, sandbox := range sandboxes {
```

packages/api/internal/handlers/sandboxes_list.go
```diff
@@ -19,14 +20,119 @@ import (
...
+
+			memoryMB := int32(-1)
+			cpuCount := int32(-1)
+
+			if len(snapshotBuilds) > 0 {
+				memoryMB = int32(snapshotBuilds[0].RAMMB)
+				cpuCount = int32(snapshotBuilds[0].Vcpu)
+			}
+
+			instance := api.ListedSandbox{
```
jakubno:
```suggestion
			sandbox := api.ListedSandbox{
```

packages/api/internal/handlers/sandboxes_list.go
```diff
@@ -19,14 +20,119 @@ import (
...
+
+				memoryMB := int32(-1)
+				cpuCount := int32(-1)
+
+				if buildsMap[*info.BuildID] != nil {
+					memoryMB = int32(buildsMap[*info.BuildID].RAMMB)
+					cpuCount = int32(buildsMap[*info.BuildID].Vcpu)
+				}
+
+				instance := api.ListedSandbox{
```
jakubno:
```suggestion
				sandbox := api.ListedSandbox{
```

packages/api/internal/handlers/sandboxes_list.go
```diff
@@ -19,14 +20,119 @@ import (
...
 
 	telemetry.ReportEvent(ctx, "list running instances")
 
-	instanceInfo := a.orchestrator.GetSandboxes(ctx, &team.ID)
+	// Initialize empty slice for results
+	sandboxes := make([]api.ListedSandbox, 0)
+
+	// Only fetch running instances if we need them (state is nil or "running")
+	if params.State == nil || *params.State == "running" {
+		instanceInfo := a.orchestrator.GetSandboxes(ctx, &team.ID)
```
jakubno:
```suggestion
		sandboxInfo := a.orchestrator.GetSandboxes(ctx, &team.ID)
```

packages/api/internal/handlers/sandbox_get.go
jakubno: could log this?

packages/api/internal/handlers/sandboxes_list.go
```diff
@@ -19,14 +20,119 @@ import (
...
+				sandboxes = append(sandboxes, instance)
+			}
+		}
+	}
+
+	// Only fetch snapshots if we need them (state is nil or "paused")
+	if params.State == nil || *params.State == "paused" {
+		snapshotEnvs, err := a.db.GetTeamSnapshots(ctx, team.ID)
+		if err != nil {
+			telemetry.ReportCriticalError(ctx, err)
```
jakubno: could you log the error

jakubno (dismissed):
Looks good, there are some nitpicks and few stuff changed due to the change in database type

packages/api/internal/handlers/sandbox_get.go
mishushakov: I am not seeing the line you're talking about?

packages/api/internal/handlers/sandbox_get.go
mishushakov: What line are you mentioning? StartedAt is now nullable (but required) - it is used in the CLI and Dashboard

packages/api/internal/handlers/sandboxes_list.go
mishushakov: Sorry, am still not understanding what is required from me, can you suggest the change?

mishushakov:
I think we should make StartedAt non-nullable to avoid breaking existing types and instead set 1/1/1970 for unknown snapshot date?

mishushakov:
Here's the script for filling in sandbox_started_at from logs

```
UPDATE public.snapshots s
SET sandbox_started_at = bl.started_at
FROM billing.sandbox_logs bl
WHERE s.sandbox_id = bl.sandbox_id;
```

mishushakov:
Doing some refactoring after Tomas feedback-call

packages/api/internal/handlers/sandbox_get.go
```diff
@@ -24,26 +24,58 @@ func (a *APIStore) GetSandboxesSandboxID(c *gin.Context, id string) {
...
+	if err == nil {
+		// Check if sandbox belongs to the team
+		if *info.TeamID != team.ID {
+			zap.L().Error("sandbox %s doesn't exist or you don't have access to it", zap.String("sandbox_id", id))
+			c.JSON(http.StatusNotFound, fmt.Sprintf("sandbox \"%s\" doesn't exist or you don't have access to it", id))
+			return
+		}
+
+		// Sandbox exists and belongs to the team - return running sandbox info
+		_, build, err := a.templateCache.Get(ctx, info.Instance.TemplateID, team.ID, true)
```
ValentaTomas: Why do we need to fetch build here?

packages/api/internal/handlers/sandboxes_list.go
```diff
@@ -155,13 +303,47 @@ func (a *APIStore) GetSandboxes(c *gin.Context, params api.GetSandboxesParams) {
...
+	})
+	if err != nil {
+		zap.L().Error("Error fetching sandboxes", zap.Error(err))
+		a.sendAPIStoreError(c, http.StatusInternalServerError, fmt.Sprintf("Error returning sandboxes for team '%s': %s", team.ID, err))
+
+		return
+	}
+
+	// Filter sandboxes
+	sandboxes, err = filterSandboxes(sandboxes, SandboxesListFilter{
```
ValentaTomas: I think the filter could be used via the DB directly to match for attributes in the metadata JSON.

packages/api/internal/handlers/sandboxes_list.go
```diff
@@ -155,13 +303,47 @@ func (a *APIStore) GetSandboxes(c *gin.Context, params api.GetSandboxesParams) {
 	properties := a.posthog.GetPackageToPosthogProperties(&c.Request.Header)
 	a.posthog.CreateAnalyticsTeamEvent(team.ID.String(), "listed running instances", properties)
 
-	sandboxes, err := a.getSandboxes(ctx, team.ID, params.Query)
+	sandboxes, err := a.getSandboxes(ctx, team.ID, SandboxesListParams{
```
ValentaTomas: The return limit should be probably already enforced here (especially for the DB) so we don't return everything at once.

packages/shared/pkg/db/snapshot.go
```diff
@@ -200,3 +204,29 @@ func (db *DB) GetSnapshotBuilds(ctx context.Context, sandboxID string, teamID uu
...
+	[]*models.Env,
+	error,
+) {
+	e, err := db.
+		Client.
+		Env.
+		Query().
+		Where(
+			env.HasBuildsWith(envbuild.StatusEQ(envbuild.StatusSuccess)),
+			env.HasSnapshotsWith(snapshot.SandboxIDNotIn(excludeSandboxIDs...)),
```
ValentaTomas: This might be dangerous, because we don't know the upper cap of how many we will send. Might be better to fetch limit, then deduplicate and refetch until we get at least one valid or we reach the end.

packages/shared/pkg/db/snapshot.go
```diff
@@ -200,3 +204,29 @@ func (db *DB) GetSnapshotBuilds(ctx context.Context, sandboxID string, teamID uu
...
 }
+
+func (db *DB) GetTeamSnapshots(ctx context.Context, teamID uuid.UUID, excludeSandboxIDs []string) (
+	[]*models.Env,
+	error,
+) {
+	e, err := db.
+		Client.
+		Env.
+		Query().
```
ValentaTomas, line 220: I think the query could be on snapshots table (it contains sandbox ID we will be using plus the metadata to filter for).
After finding the corresponding entries we will add `env` plus the latest `env_build` that corresponds to this.

packages/api/internal/handlers/sandbox_get.go
mishushakov: originally it was required to fetch vCPU and RAM count, but I see that InstanceInfo struct contains that

packages/api/internal/handlers/sandbox_get.go
jakubno: I think I meant changing line 23 to 
```
	telemetry.ReportEvent(ctx, "get running sandbox")
```

packages/api/internal/handlers/sandboxes_list.go
```diff
@@ -15,131 +19,295 @@ import (
...
+	sandboxes := make([]api.ListedSandbox, 0)
+	snapshots, err := a.db.GetTeamSnapshotsWithCursor(ctx, teamID, runningSandboxesIDs, int(*limit), metadataFilter, cursorTime, cursorID)
+	if err != nil {
+		return nil, fmt.Errorf("error getting team snapshots: %s", err)
+	}
+
+	// Add snapshots to results
+	for _, snapshot := range snapshots {
+		env := snapshot.Edges.Env
+		if env == nil {
```
jakubno: can it be null?

packages/shared/pkg/db/snapshot.go
```diff
@@ -200,3 +205,73 @@ func (db *DB) GetSnapshotBuilds(ctx context.Context, sandboxID string, teamID uu
...
+					snapshot.SandboxStartedAtEQ(*cursorTime),
+					snapshot.SandboxIDGT(*cursorID),
+				),
+			),
+		)
+	}
+
+	// Apply metadata filtering
+	if metadataFilter != nil {
+		query = query.Where(snapshot.MetadataEQ(*metadataFilter))
```
ValentaTomas: Just to check here—does this correctly include rows if there are additional fields in their metadata or does the metadata needs to strictly equal the metadata we are filtering for?

packages/shared/pkg/db/snapshot.go
```diff
@@ -200,3 +205,73 @@ func (db *DB) GetSnapshotBuilds(ctx context.Context, sandboxID string, teamID uu
...
+	}
+
+	// Order by sandbox_started_at (descending), then by sandbox_id (ascending) for stability
+	query = query.Order(models.Desc(snapshot.FieldSandboxStartedAt), models.Asc(snapshot.FieldSandboxID))
```
ValentaTomas: I looked into this further and it seems that if you chain sorts they are not guaranteed to be stable (in the sorting term—keeping order), so I think we might just remove the startedAt sorting.

packages/shared/pkg/db/snapshot.go
```diff
@@ -200,3 +205,73 @@ func (db *DB) GetSnapshotBuilds(ctx context.Context, sandboxID string, teamID uu
...
+
+	// Apply limit + 1 to check if there are more results
+	query = query.Limit(limit + 1)
```
ValentaTomas, line 255: I think this works equeally well with just limit, because we never know where the "stop" will be so checking just one more further is kinda the same as just returning the limit and checking if we return at least something.

packages/shared/pkg/db/snapshot.go
```diff
@@ -200,3 +205,73 @@ func (db *DB) GetSnapshotBuilds(ctx context.Context, sandboxID string, teamID uu
...
+			snapshot.HasEnvWith(
+				env.And(
+					env.TeamID(teamID),
+					env.HasBuildsWith(envbuild.StatusEQ(envbuild.StatusSuccess)),
+				),
+			),
+		).
+		WithEnv(func(query *models.EnvQuery) {
+			query.WithBuilds(func(query *models.EnvBuildQuery) {
+				query.Order(models.Desc(envbuild.FieldFinishedAt))
```
ValentaTomas, line 231: Is it possible to only return the one latest build for the env here?

packages/shared/pkg/db/snapshot.go
```diff
@@ -200,3 +205,73 @@ func (db *DB) GetSnapshotBuilds(ctx context.Context, sandboxID string, teamID uu
...
+			})
+		})
+
+	// Apply cursor-based filtering if cursor is provided
+	if cursorTime != nil && cursorID != nil {
+		query = query.Where(
```
ValentaTomas: I'm thinking if the condition here should not be:

`sandbox_created_at < cursor_time && sandbox_id > cursor_id`.

If we are invoking this for the first time we would pass the cursor time as "now" and the sandbox id as arbitrary string that is always lexically after any sandbox ID (could be `0` I think).

ValentaTomas (commented):
I'm adding a partial review here—I'll be adding some few more comments.

packages/shared/pkg/db/snapshot.go
mishushakov: It will check if the row contains the given json, the naming is bit off

packages/api/internal/handlers/sandboxes_list.go
jakubno: just the name of the analytics event, it isn't only listing running sandboxes anymore

packages/shared/pkg/db/snapshot.go
mishushakov: so the diff will look like this in the end?

```diff
- query = query.Order(models.Desc(snapshot.FieldSandboxStartedAt), models.Asc(snapshot.FieldSandboxID))
+ query = query.Order(models.Asc(snapshot.FieldSandboxID))
```

packages/shared/pkg/db/snapshot.go
mishushakov: The reason this was implemented is so if the limit is + 1 you can trim it and return the cursor to the first item of the next page

packages/shared/pkg/db/snapshot.go
mishushakov: This was on purpose - for some reason, if I add Limit(n) or First(ctx) to WithBuilds, the snapshot list only returns one item. I had a similar issue with some other query like this somewhere else, maybe it's ent-go.

```go
query.WithBuilds(func(query *models.EnvBuildQuery) {
    query.Where(envbuild.StatusEQ(envbuild.StatusSuccess)).
      Order(models.Desc(envbuild.FieldFinishedAt)).
      Limit(1)
})
```

packages/shared/pkg/db/snapshot.go
mishushakov: I just had another look at this and think it makes sense, if page len <= limit then we can assume we're looking at the last page already. The only change needed here would be to fetch items before cursor - right now the cursor is pointing at the first item after limit, thus n + 1

Screenshot 2025-03-20 at 17 27 07

mishushakov: The problem with what you are suggesting is that if we remove +1, we cannot reliably tell if there are more items in the list left to paginate. For example if last page is same size as the limit you might think there is a next page, but maybe there is not. It would also complicate the logic on the client - it will have to rely on the response body (empty list) to tell if there are any more items left

ValentaTomas: I see what you mean—what I had in mind is that you will loop in the API *until* you either get one item or you go over all of the items, so in body you either return some item or you found nothing so this will be the last request.

ValentaTomas: I think the +1 in the end should work ok, we can continue with this.

packages/shared/pkg/db/snapshot.go
ValentaTomas: As mentioned, we should keep the two sorting, just use the createdAt instead.

packages/api/internal/handlers/sandboxes_list.go
```diff
@@ -15,132 +19,306 @@ import (
...
+	// Get all sandbox instances
+	runningSandboxes := a.orchestrator.GetSandboxes(ctx, &teamID)
+
+	// Running Sandbox IDs
+	runningSandboxesIDs := make([]string, 0)
+	for _, info := range runningSandboxes {
+		runningSandboxesIDs = append(runningSandboxesIDs, utils.ShortID(info.Instance.SandboxID))
+	}
+
+	if params.State == nil || (slices.Contains(*params.State, api.Running) && slices.Contains(*params.State, api.Paused)) {
```
ValentaTomas, line 214: Wasn't the default behavior supposed to be still return just running when no state is passed to keep backward compatibility as we do now also in the SDK?

ValentaTomas (commented):
Other than the last comment I left I think we should be good to merge.

packages/api/internal/handlers/sandboxes_list.go
mishushakov: I will check with @mlejva

mishushakov: He said breaking change would be fine

mishushakov:
Here's the migration script for prod

```
BEGIN;

ALTER TABLE "public"."snapshots"
  ADD COLUMN IF NOT EXISTS "sandbox_started_at" timestamp with time zone NOT NULL 
  DEFAULT TIMESTAMP WITH TIME ZONE '1970-01-01 00:00:00 UTC';

-- Update records with actual data
UPDATE public.snapshots s
SET sandbox_started_at = latest_starts.started_at
FROM (
  SELECT sandbox_id, MAX(started_at) as started_at
  FROM billing.sandbox_logs
  GROUP BY sandbox_id
) latest_starts
WHERE s.sandbox_id = latest_starts.sandbox_id;

-- Remove the default constraint after populating data
ALTER TABLE "public"."snapshots"
ALTER COLUMN "sandbox_started_at" DROP DEFAULT;

COMMIT;
```

ValentaTomas (approved):

ValentaTomas:
@mishushakov Resolve the comments and you can merge.

ValentaTomas:
@jakubno Can you handle applying of the migration script so this can be deployed?

dobrac:
> @jakubno Can you handle applying of the migration script so this can be deployed?

The migration script should be part of migrations, otherwise any autodeploy will not work for us, and for people selfhosting the infra.

PR #265, opened 2025-01-29
Fix forward_ports service and disable envd-v0.0.1 service

0div:
Use `iptables` to forward all `localhost` <> `fc-ip` ports instead of [legacy port forwarding in envd-v0.0.1](https://github.com/e2b-dev/infra/blob/v0.0.75/packages/envd/internal/port/forward.go#L45)

### DIFF
--- a/packages/template-manager/internal/build/provision.sh
+++ b/packages/template-manager/internal/build/provision.sh
@@ -134,7 +134,7 @@ echo "nameserver 8.8.8.8" >/etc/resolv.conf
 
 # Start systemd services
 systemctl enable envd
-systemctl enable envd-v0.0.1
+# systemctl enable envd-v0.0.1
 systemctl enable chrony 2>&1
 
 cat <<EOF >/etc/systemd/system/forward_ports.service
@@ -146,12 +146,12 @@ Type=simple
 Restart=no
 User=root
 Group=root
-ExecStart=/bin/bash -l -c "(echo 1 | tee /proc/sys/net/ipv4/ip_forward) && iptables-legacy -t nat -A POSTROUTING -s 127.0.0.1 -j SNAT --to-source {{ .FcAddress }} && iptables-legacy -t nat -A PREROUTING -d {{ .FcAddress }} -j DNAT --to-destination 127.0.0.1"
+ExecStart=/bin/bash -l -c "(echo 1 | tee /proc/sys/net/ipv4/ip_forward) && (echo 1 | tee /proc/sys/net/ipv4/conf/eth0/route_localnet) && iptables-legacy -t nat -A POSTROUTING -s 127.0.0.1 -j SNAT --to-source {{ .FcAddress }} && iptables-legacy -t nat -A PREROUTING -d {{ .FcAddress }} -j DNAT --to-destination 127.0.0.1"
 
 [Install]
 WantedBy=multi-user.target
 EOF
 
-# systemctl enable forward_ports
+systemctl enable forward_ports
 
 echo "Finished provisioning script"


ValentaTomas:
One consideration here—one of the other reasons why we had the old envd there was because of the compatibility with the old SDK for now.

Does this change works when the old env is still present and tries to also forward the ports?

0div:
> Does this change works when the old env is still present and tries to also forward the ports?

It works with both running yeah, but :sad face: i was hoping we could stop using old envd for good

mlejva:
@0div can you please replace the error message with following code?

https://gist.github.com/mlejva/06d01a605eae36c9c330a8a6a5908afb

Just replace 
```
<img src="https://hebbkx1anhila5yf.public.blob.vercel-storage.com/Symbol%20Gradient-Kr5pnWlK3ZhzBcRGf6Am4cNbJvY1Ge.svg" alt="Logo" class="logo">
```
with the actual SVG code please

The result should be this:

Screenshot 2025-02-04 at 8 46 18 PM

mlejva:
Actually use [this one](https://gist.github.com/mlejva/d295c850c9a61535562b09599330f72d) please. I updated it to support dark theme as well.


Screenshot 2025-02-04 at 8 53 36 PM

0div:
> Actually use [this one](https://gist.github.com/mlejva/d295c850c9a61535562b09599330f72d) please. I updated it to support dark theme as well.
> 
> Screenshot 2025-02-04 at 8 53 36 PM

I tired inline but it's too big, I'm gonna have to load it as a file in that case, gonna require some changes in nginx config and how we provision the nomad job.

mlejva:
> > Actually use [this one](https://gist.github.com/mlejva/d295c850c9a61535562b09599330f72d) please. I updated it to support dark theme as well.
> > Screenshot 2025-02-04 at 8 53 36 PM
> 
> I tried inline but it's too big, I'm gonna have to load it as a file in that case, gonna require some changes in nginx config and how we provision the nomad job.

Just to be sure we try all the options - have you tried minifying the code?

0div:
> > > Actually use [this one](https://gist.github.com/mlejva/d295c850c9a61535562b09599330f72d) please. I updated it to support dark theme as well.
> > > Screenshot 2025-02-04 at 8 53 36 PM
> > 
> > 
> > I tried inline but it's too big, I'm gonna have to load it as a file in that case, gonna require some changes in nginx config and how we provision the nomad job.
> 
> Just to be sure we try all the options - have you tried minifying the code?

Yep, doesn't seem to work, it can't find the closing quote even though i made sure there were no other quotes.

mlejva:
@0div if this is content length problem, I think this could be fixed by changing the default value the `client_max_body_size` property in the nginx config file https://nginx.org/en/docs/http/ngx_http_core_module.html#client_max_body_size

jakubno:
Can nginx do something like template file? To have the base template as html as then just replace variables?

0div:
> Can nginx do something like template file? To have the base template as html as then just replace variables?

@jakubno I didn't find any obvious out-of-the-box way of doing it purely via nginx.

jakubno (changes requested):
I think we still can't stop supporting old envd. I would separate this PR in 2. 
For turning on port forwarding we would rebuild old envd without this functionality. I can help with that if needed.

0div:
> I think we still can't stop supporting old envd. I would separate this PR in 2. For turning on port forwarding we would rebuild old envd without this functionality. I can help with that if needed.

@jakubno forgot to uncomment it, ready for re-review

jakubno (approved):
I removed port forwarding, because it requires additional work, doesn't make much sense why there 2 should be together

PR #266, opened 2025-01-29
Use dynamic service discovery for client dns routing

jakubno:
# Description

Use consul service discovery for sandbox DNS, this will allow us to swap API servers without problems with resolving DNS in client proxy

### DIFF
--- a/packages/api/internal/orchestrator/orchestrator.go
+++ b/packages/api/internal/orchestrator/orchestrator.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 
+	"github.com/e2b-dev/infra/packages/shared/pkg/utils"
 	"github.com/go-redis/redis/v8"
 	nomadapi "github.com/hashicorp/nomad/api"
 	"go.opentelemetry.io/otel/trace"
@@ -40,6 +41,7 @@ func New(
 	}
 
 	dnsServer := dns.New(redisClient)
+	port := utils.RequiredEnv("DNS_PORT", "Local DNS server resolving IPs for sandboxes")
 
 	if env.IsLocal() {
 		logger.Info("Running locally, skipping starting DNS server")
@@ -47,7 +49,7 @@ func New(
 		go func() {
 			logger.Info("Starting DNS server")
 
-			if err := dnsServer.Start(ctx, "127.0.0.4", 53); err != nil {
+			if err := dnsServer.Start(ctx, "0.0.0.0", port); err != nil {
 				logger.Panic("Failed starting DNS server", zap.Error(err))
 			}
 		}()
--- a/packages/api/internal/orchestrator/orchestrator.go
+++ b/packages/api/internal/orchestrator/orchestrator.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 
-	"github.com/e2b-dev/infra/packages/shared/pkg/utils"
 	"github.com/go-redis/redis/v8"
 	nomadapi "github.com/hashicorp/nomad/api"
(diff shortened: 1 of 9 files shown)


packages/api/internal/orchestrator/orchestrator.go
```diff
@@ -39,15 +41,20 @@ func New(
 		logger.Errorf("Error initializing Analytics client\n: %v", err)
 	}
 
+	port := utils.RequiredEnv("DNS_PORT", "Local DNS server resolving IPs for sandboxes")
 	dnsServer := dns.New()
+	intPort, convErr := strconv.Atoi(port)
+	if convErr != nil {
+		log.Fatalf("Failed to convert DNS_PORT to int: %v\n", convErr)
```
tychoish: I was going to go through and change these `log.Fatalf` calls, but they call `os.Exit` which means defers don't run. (I'm generally of the opinion that `log.Fatal` should really only be used in the body of `main()` , and otherwise you should either panic or return an error). In any case, this function returns an error, so I think we should just do that here. 

I'd also say it's _probably_ fine to just default to 53, I'm not sure the case where this wouldn't be correct.

packages/api/internal/orchestrator/orchestrator.go
```diff
@@ -39,15 +41,20 @@ func New(
 		logger.Errorf("Error initializing Analytics client\n: %v", err)
 	}
 
+	port := utils.RequiredEnv("DNS_PORT", "Local DNS server resolving IPs for sandboxes")
 	dnsServer := dns.New()
+	intPort, convErr := strconv.Atoi(port)
```
tychoish: in the start function (which takes an `int`) [here](https://github.com/e2b-dev/infra/blob/main/packages/api/internal/dns/server.go#L87-L100) we just convert it _back_ to string (and if it's not a valid port number it'll error there.

Admittedly the Start method doesn't return in the foreground which means we either have to panic, or continue without a functioning DNS server (this has implications for #261). 

While we're (I'm) being pedantic, a valid `int` isn't neccessarily a valid port number: port numbers have to be valid `uint16` numbers and `int` is basically always an `int64` (but sometimes an `int32` though in practice we may be confident that we'll never need to run this software from a 32-bit build). So we could pass this _and still_ have an invalid port number.

packages/nomad/api.hcl
```diff
@@ -73,6 +73,9 @@ job "api" {
         REDIS_URL                     = "${redis_url}"
         # This is here just because it is required in some part of our code which is transitively imported
         TEMPLATE_BUCKET_NAME          = "skip"
+        ADMIN_TOKEN                   = var.admin_token
+        REDIS_URL                     = var.redis_url
```
tychoish: this duplicates line 72-73, which is either unnecessary or confusing 😅

packages/nomad/api.hcl
```diff
@@ -73,6 +73,9 @@ job "api" {
...
+        DNS_PORT                      = 5353
```
tychoish: is there a particular reason to use a non-default DNS port?

jakubno: yes, there already running another DNS server on 53
```
Failed starting DNS server{error 26 0  failed to start DNS server: listen udp 0.0.0.0:53: bind: address already in use}
```

packages/nomad/api.hcl
jakubno: sorry, some mistake in merge conflict

packages/api/internal/orchestrator/orchestrator.go
jakubno: Changed to string

PR #271, opened 2025-01-31
Use runtime logger timestamp instead of log-collector timestamp in sandbox logs and metrics

0div:
### Description 
~~Log the updated `real_timestamp` (to avoid name collision with `timestamp` for `orchestrator` and `envd` loggers runtimes and subsequently pick it up in the API handlers while building the payloads for logs and metrics instead of Loki timestamp.~~

- remap native loki timestamp to log message timestamp in log-collector vector config.
- parse entry line timestamp in sandbox logs instead of using entry timestamp in sandbox logs & metrics api handlers
- filter out metrics from sandbox logs 

### Test
- build and upload orchestrator, envd and api
- make plan & make apply
- build a new base template
- spawn sandbox in cli
- read logs with sandbox id in cli

you should get something like this:
```sh
Logs for sandbox i7b1xzxibrlx663y2bte8-036c7f39:

[2025-02-06 23:46:46.258Z] INFO  { logger: 'orchestrator', message: 'Sandbox created with - end time: 2025-02-06 23:51:46 +00:00', timestamp: '2025-02-06T23:46:46.258460232Z' }
[2025-02-06 23:46:46.609Z] INFO  { command: '/bin/bash -i -l', event_type: 'process_start', logger: 'process', operation_id: '3', pid: 328, timestamp: '2025-02-06T23:46:46.609501347Z' }
```


ValentaTomas:
As this will require changes to the envd and would be adding another field (that we might want to hide), I'm thinking if there is not some way to do this through Loki directly (which we would be able to redeploy better).
Something like https://community.grafana.com/t/loki-unable-to-parse-timestamp-sent-what-timestamp-format-does-loki-expect/74582

packages/nomad/logs-collector.hcl
```diff
@@ -98,7 +98,7 @@ job "logs-collector" {
         destination   = "local/vector.toml"
         change_mode   = "signal"
         change_signal = "SIGHUP"
-        # overriding the delimiters to [[ ]] to avoid conflicts with Vector's native templating, which also uses {{ }}
+        # overriding the delimiters to [ ] to avoid conflicts with Vector's native templating, which also uses {{ }}
```
jakubno:
```suggestion
        # overriding the delimiters to [[ ]] to avoid conflicts with Vector's native templating, which also uses {{ }}
```

jakubno (approved):

PR #272, opened 2025-01-31
fix: narrow scope of signal handler context

tychoish:


jakubno (commented):
How does the server shut down now? I don't see any way it would get cancel.

tychoish:
> How does the server shut down now? I don't see any way it would get cancel.

This is unchanged from what's in `main`: https://github.com/e2b-dev/infra/pull/272/files#diff-67d1b8117fb66e8f1be66e37b0bf6d7f6b84981d730c7b115f90bafc6c9241a8R271 we have a goroutine waiting on the signal handler's context to get canceled, which then calls Shutdown(). Shutdown, stops listening for new connections immediately, and waits for all currently executing handlers to return. We have a wait group that ensures that the server has fully returned before running the cleanup operations. Context's passed to handlers are never canceled until after the HTTP service has returned.

jakubno (approved):

PR #273, opened 2025-02-03
Add a way to delete paused sandboxes

mishushakov:
Adds back the ability to delete templates - only if the there are no snapshots using that template as a base_env.
Ps. I stole some of your stuff from previous commits.


mishushakov:
I still have to make env_id and base_env_id a foreign key so it can be cascaded

mishushakov:
Haven't figured out still why env_id and is not foreign key, still

ValentaTomas:
Making base_env_id foreign key is not for the cascade, right? That will be a normal template which we don't want to delete on snapshot delete

mishushakov:
you cannot delete a template if any snapshot depends on it, so no need for cascade delete in this case

packages/template-manager/internal/template/main.go
```diff
@@ -0,0 +1,52 @@
...
+) error {
+	childCtx, childSpan := tracer.Start(ctx, "delete-template")
+	defer childSpan.End()
+
+	err := templateStorage.Remove(ctx, buildId)
+	if err != nil {
+		return fmt.Errorf("error when deleting template objects: %w", err)
+	}
+
+	op, artifactRegistryDeleteErr := artifactRegistry.DeletePackage(ctx, &artifactregistrypb.DeletePackageRequest{Name: GetDockerImageURL(buildId)})
```
jakubno, line 38: if it's only snapshot there won't be any image, so notFound is kinda expected here (I would just mention it here as comment or inside the log)

packages/shared/pkg/schema/snapshots.go
```diff
@@ -31,7 +31,7 @@ func (Snapshot) Fields() []ent.Field {
 
 func (Snapshot) Edges() []ent.Edge {
 	return []ent.Edge{
-		edge.From("env", Env.Type).Ref("snapshots").Unique().Field("env_id").Required(),
+		edge.From("env", Env.Type).Ref("snapshots").Unique().Field("env_id").Required().Annotations(entsql.OnDelete(entsql.Cascade)),
```
jakubno: I think this isn't correct, it should be on `Env` side

packages/api/internal/handlers/template_delete.go
```diff
@@ -88,7 +89,39 @@ func (a *APIStore) DeleteTemplatesTemplateID(c *gin.Context, aliasOrTemplateID a
...
+	if err != nil {
+		telemetry.ReportError(ctx, fmt.Errorf("error when checking if base env has snapshots: %w", err))
+		a.sendAPIStoreError(c, http.StatusInternalServerError, "Error when checking if base env has snapshots")
+
+		return
+	}
+
+	if hasSnapshots {
+		telemetry.ReportError(ctx, fmt.Errorf("base env '%s' has snapshots", template.ID))
+		a.sendAPIStoreError(c, http.StatusBadRequest, fmt.Sprintf("cannot delete env '%s' because snapshots exist", template.ID))
```
jakubno: this is user facing error, it should be something like
```suggestion
		a.sendAPIStoreError(c, http.StatusBadRequest, fmt.Sprintf("cannot delete template '%s' because there are derived paused sandboxes", template.ID))
```

packages/api/internal/handlers/template_delete.go
```diff
@@ -88,7 +89,39 @@ func (a *APIStore) DeleteTemplatesTemplateID(c *gin.Context, aliasOrTemplateID a
 		attribute.String("env.id", template.ID),
 	)
 
-	deleteJobErr := a.templateManager.DeleteInstance(ctx, template.ID)
+	// check if base env has snapshots
+	hasSnapshots, err := a.db.CheckBaseEnvHasSnapshots(ctx, template.ID)
+	if err != nil {
+		telemetry.ReportError(ctx, fmt.Errorf("error when checking if base env has snapshots: %w", err))
+		a.sendAPIStoreError(c, http.StatusInternalServerError, "Error when checking if base env has snapshots")
```
jakubno, line 96: this is user facing error, it should be something like
```suggestion
		a.sendAPIStoreError(c, http.StatusInternalServerError, "Error when checking if base template has paused sandboxes")
```

packages/api/internal/template-manager/delete_template.go
```diff
@@ -5,17 +5,29 @@ import (
...
 	err = utils.UnwrapGRPCError(err)
 	if err != nil {
-		return fmt.Errorf("failed to delete template '%s': %w", templateID, err)
+		return fmt.Errorf("failed to delete env build '%s': %w", buildId.String(), err)
+	}
+
+	return nil
+}
+
+func (tm *TemplateManager) DeleteInstances(ctx context.Context, buildIds []uuid.UUID) error {
```
jakubno:
```suggestion
func (tm *TemplateManager) DeleteBuilds(ctx context.Context, buildIds []uuid.UUID) error {
```

packages/api/internal/template-manager/delete_template.go
```diff
@@ -5,17 +5,29 @@ import (
...
 	"github.com/e2b-dev/infra/packages/api/internal/utils"
-	"github.com/e2b-dev/infra/packages/shared/pkg/grpc/template-manager"
+	template_manager "github.com/e2b-dev/infra/packages/shared/pkg/grpc/template-manager"
+	"github.com/google/uuid"
 )
 
-func (tm *TemplateManager) DeleteInstance(ctx context.Context, templateID string) error {
-	_, err := tm.grpc.Client.TemplateDelete(ctx, &template_manager.TemplateDeleteRequest{
-		TemplateID: templateID,
+func (tm *TemplateManager) DeleteInstance(ctx context.Context, buildId uuid.UUID) error {
```
jakubno:
```suggestion
func (tm *TemplateManager) DeleteBuild(ctx context.Context, buildId uuid.UUID) error {
```

jakubno (changes requested):

packages/template-manager/internal/template/main.go
```diff
@@ -0,0 +1,60 @@
...
+	"google.golang.org/grpc/codes"
+
+	"github.com/e2b-dev/infra/packages/shared/pkg/consts"
+	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
+	"github.com/gogo/status"
+)
+
+func GetDockerImageURL(templateID string) string {
+	// DockerImagesURL is the URL to the docker images in the artifact registry
+	return fmt.Sprintf("projects/%s/locations/%s/repositories/%s/packages/", consts.GCPProject, consts.GCPRegion, consts.DockerRegistry) + templateID
```
jakubno:
```suggestion
	return fmt.Sprintf("projects/%s/locations/%s/repositories/%s/packages/%s", consts.GCPProject, consts.GCPRegion, consts.DockerRegistry, templateID)
```

packages/template-manager/template-manager.proto
```diff
@@ -37,6 +37,6 @@ message TemplateBuildLog {
 service TemplateService {
   // TemplateCreate is a gRPC service that creates a new template
   rpc TemplateCreate (TemplateCreateRequest) returns (stream TemplateBuildLog);
-  // TemplateDelete is a gRPC service that deletes files associated with a template
-  rpc TemplateDelete (TemplateDeleteRequest) returns (google.protobuf.Empty);
+  // EnvBuildDelete is a gRPC service that deletes files associated with a template build
```
jakubno:
```suggestion
  // TemplateBuildDelete is a gRPC service that deletes files associated with a template build
```

jakubno (approved):

jakubno (approved):

PR #274, opened 2025-02-03
Add proxy config to fix a problem with expecting config on incorrect path

jakubno:


ValentaTomas:
When tested, it this the only thing preventing chrony/init fixing the clock?

jakubno:
After this clocks are synced

ValentaTomas (approved):

PR #276, opened 2025-02-04
Reimplement client proxy in go

jakubno:
Implement proxy in go


packages/client-proxy/main.go
```diff
@@ -0,0 +1,95 @@
...
+
+	// Set the question
+	msg.SetQuestion(host+".", dns.TypeA)
+
+	// Define the DNS server to query
+	dnsServer := "api.service.consul:5353"
+
+	var resp *dns.Msg
+	var err error
+	for range 3 {
```
tychoish: why 3?

packages/client-proxy/main.go
```diff
@@ -0,0 +1,95 @@
...
+		// Send the query to the server
+		resp, _, err = client.Exchange(msg, dnsServer)
+
+		// The api server wasn't found, maybe the API server is rolling and the DNS server is not updated yet
+		if err != nil || len(resp.Answer) == 0 {
+			log.Printf("Host not found: %s\n", host)
```
tychoish: you don't need `\n` in log statements' printf statements, I'd use `%q`.

packages/client-proxy/main.go
```diff
@@ -0,0 +1,95 @@
...
+
+	// Extract sandbox id from the host (<port>-<sandbox id>-<old client id>.e2b.dev)
+	host := strings.Split(r.Host, "-")[1]
+	msg := new(dns.Msg)
+
+	// Set the question
+	msg.SetQuestion(host+".", dns.TypeA)
+
+	// Define the DNS server to query
+	dnsServer := "api.service.consul:5353"
```
tychoish: make this a const at the top

packages/client-proxy/main.go
```diff
@@ -0,0 +1,95 @@
...
+
+func proxy(w http.ResponseWriter, r *http.Request) {
+	log.Printf("Request for %s %s\n", r.Host, r.URL.Path)
+
+	// Extract sandbox id from the host (<port>-<sandbox id>-<old client id>.e2b.dev)
+	host := strings.Split(r.Host, "-")[1]
+	msg := new(dns.Msg)
+
+	// Set the question
+	msg.SetQuestion(host+".", dns.TypeA)
```
tychoish: use `fmt.Sprint()` rather than `+` for string concatenation because `+` reallocates a lot of memory, and `fmt.Sprint` (and friends) have a buffer pool.

packages/client-proxy/main.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/miekg/dns"
+)
+
+// Create a DNS client
+var client = new(dns.Client)
+
+func proxy(w http.ResponseWriter, r *http.Request) {
+	log.Printf("Request for %s %s\n", r.Host, r.URL.Path)
+
+	// Extract sandbox id from the host (<port>-<sandbox id>-<old client id>.e2b.dev)
```
tychoish: I've seen us do this in other cases: probably we should have a utility function do this in only one place. probably future work.

packages/client-proxy/main.go
```diff
@@ -0,0 +1,95 @@
...
+	var resp *dns.Msg
+	var err error
+	for range 3 {
+		// Send the query to the server
+		resp, _, err = client.Exchange(msg, dnsServer)
+
+		// The api server wasn't found, maybe the API server is rolling and the DNS server is not updated yet
+		if err != nil || len(resp.Answer) == 0 {
+			log.Printf("Host not found: %s\n", host)
+			continue
```
tychoish: might be good to log which iteration we're on... My first question was "why 3 retries" and then "shouldn't our chance of getting a good response after an error go up if we have even a little sleep/jitter". Might also be interesting to be able to see what the error is. 

Logging which iteraton we're on and the error, might allow us to spot some latency cause (if everything only succeeds on the second iteration?) or remove some of this code if it turns out that we never get a success on attempts 2 and 3.

packages/client-proxy/main.go
```diff
@@ -0,0 +1,95 @@
...
+		log.Printf("Host not found: %s\n", host)
+		http.Error(w, "Host not found", http.StatusBadGateway)
+		return
+	}
+
+	// We've resolved the node to proxy the request to
+	log.Printf("Proxying request to %s\n", resp.Answer[0].String())
+	targetUrl := &url.URL{
+		Scheme: "http",
+		Host:   resp.Answer[0].(*dns.A).A.String() + ":3003",
```
tychoish: `fmt.Sprint` over direct concatenation.

packages/client-proxy/main.go
```diff
@@ -0,0 +1,95 @@
...
+			w.Write([]byte("Sandbox not found"))
+
+			return
+		}
+
+		break
+	}
+
+	// There's no answer, we can't proxy the request
+	if err != nil || resp == nil || len(resp.Answer) == 0 {
```
tychoish: on line 36 we don't check `resp == nil`. seems like we could omit this (or we need the nil check above.)

packages/client-proxy/main.go
```diff
@@ -0,0 +1,95 @@
...
+		log.Printf("Host not found: %s\n", host)
+		http.Error(w, "Host not found", http.StatusBadGateway)
```
tychoish: This is fine for now, but it might be interesting to in some way distinguish between "not found" (e.g. empty answer) and there was some kind of error with the request (`err != nil`). In many situations, one of these might be retryable, and another might not. 

We also say `sandbox` on line 44 and `host` here. I think we should probably be consistent.

packages/client-proxy/main.go
```diff
@@ -0,0 +1,95 @@
...
+		return
+	}
+
+	// We've resolved the node to proxy the request to
+	log.Printf("Proxying request to %s\n", resp.Answer[0].String())
```
tychoish: in the DNS server implementation we will populate multiple responses in an answer section: seems like we could omit that work (and always take the first) or we ought to do something to pick a response at random.

packages/client-proxy/main.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+func main() {
+	go func() {
+		// Health check
+		server := http.Server{Addr: ":3001"}
+
+		server.Handler = http.HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {
+			writer.WriteHeader(http.StatusOK)
+			writer.Write([]byte("OK"))
```
tychoish: is this response value important? should it be a constant?

packages/client-proxy/main.go
```diff
@@ -0,0 +1,95 @@
...
+	p.ServeHTTP(w, r)
+}
+
+func main() {
+	go func() {
+		// Health check
+		server := http.Server{Addr: ":3001"}
+
+		server.Handler = http.HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {
+			writer.WriteHeader(http.StatusOK)
```
tychoish, line 108: If you write anything to the body, you don't need to set `StatusOK` for the request...

packages/client-proxy/main.go
```diff
@@ -0,0 +1,95 @@
...
+	// We've resolved the node to proxy the request to
+	log.Printf("Proxying request to %s\n", resp.Answer[0].String())
+	targetUrl := &url.URL{
+		Scheme: "http",
+		Host:   resp.Answer[0].(*dns.A).A.String() + ":3003",
+	}
+
+	// Proxy the request
+	p := httputil.NewSingleHostReverseProxy(targetUrl)
+	p.ServeHTTP(w, r)
```
tychoish:
```suggestion
	httputil.NewSingleHostReverseProxy(targetUrl).ServeHTTP(w, r)
```

packages/client-proxy/main.go
```diff
@@ -0,0 +1,95 @@
...
+		server := http.Server{Addr: ":3001"}
+
+		server.Handler = http.HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {
+			writer.WriteHeader(http.StatusOK)
+			writer.Write([]byte("OK"))
+		})
+
+		err := server.ListenAndServe()
+		if err != nil {
+			log.Printf("server error: %v", err)
```
tychoish: I believe that `ListenAndServe` always returns an error: either `ServerClosed` or some problem opening the connection. I think it's probably useful to handle these differently. (if it's not "server closed", which probably only happens when `Shutdown` is called, getting an error here should lead the service to panic. ) 

Also, because this routine isn't checked, or waited for, it's probably the case that the program will exit before this line of code even runs.

packages/client-proxy/main.go
```diff
@@ -0,0 +1,95 @@
...
+	// There's no answer, we can't proxy the request
+	if err != nil || resp == nil || len(resp.Answer) == 0 {
+		log.Printf("Host not found: %s\n", host)
+		http.Error(w, "Host not found", http.StatusBadGateway)
+		return
+	}
+
+	// We've resolved the node to proxy the request to
+	log.Printf("Proxying request to %s\n", resp.Answer[0].String())
+	targetUrl := &url.URL{
```
tychoish: should we cache this in memory? if we're never changing the IP address of a running sandbox (e.g. any valid response will be valid as long as that sandbox is valid.) then caching this for a bit seems like it would help avoid a lot of performance-critical. It means we might proxy a request to a sandbox that isn't still alive but that will return an error just fine.)

probably fine to just add this in as a ticket/note.

tychoish (approved):
this all seems good: I have some weird nit-picks, and a few notes for some incremental improvements.

It seems sort of complicated/fragile that we have this proxy communicating with the API server over DNS. I might be wrong, but only the proxies are going to be making DNS requests, is that correct? 

In the future when there are multiple API servers and the state of the host machines (instances?) is fully managed by the orchestrators, it seems like the API servers and orchestrators could do the proxying directly for a different port/set of routes? 

Because the orchestrator owns the sandboxes, it seems reasonable that combining things in one process would be something that we'd want to delay until some later point when the orchestrators are restartable without impacting sandboxes (e.g. separate process management and the data/management parts, with data persisted (locally) to disk.)

(My instinct says that DNS is maybe the wrong protocol for doing this level of IPC between our components. Moving proxying into the other applications means that we can just have a local API to get data to the proxies. We could use a different transport (e.g. gRPC/etc.) which might be easier to manage/understand) but I'm not sure if my instinct is actually correct: removing the extra communication could be a useful simplification, but I'm less convinced that changing protocols would be important.

tychoish (approved):
this all seems good: I have some weird nit-picks, and a few notes for some incremental improvements.

It seems sort of complicated/fragile that we have this proxy communicating with the API server over DNS. I might be wrong, but only the proxies are going to be making DNS requests, is that correct? 

In the future when there are multiple API servers and the state of the host machines (instances?) is fully managed by the orchestrators, it seems like the API servers and orchestrators could do the proxying directly for a different port/set of routes? 

Because the orchestrator owns the sandboxes, it seems reasonable that combining things in one process would be something that we'd want to delay until some later point when the orchestrators are restartable without impacting sandboxes (e.g. separate process management and the data/management parts, with data persisted (locally) to disk.)

(My instinct says that DNS is maybe the wrong protocol for doing this level of IPC between our components. Moving proxying into the other applications means that we can just have a local API to get data to the proxies. We could use a different transport (e.g. gRPC/etc.) which might be easier to manage/understand) but I'm not sure if my instinct is actually correct: removing the extra communication could be a useful simplification, but I'm less convinced that changing protocols would be important.

packages/client-proxy/main.go
jakubno: I would rather avoid caching for now, we should benchmark this, but the problem is due to sandbox pausing and resuming, the same sandbox id can be in a second on different host. This PR actually fixes this behavior, which is present in nginx by default

packages/client-proxy/main.go
jakubno: No need for value

packages/client-proxy/main.go
jakubno: It isn't the same.. In API it has different format (only `<sandbox-id>-<client-id>`) here it's wirh port (`<port>-<sandbox-id>-<client-id>`)

packages/client-proxy/main.go
tychoish: ah! didn't realize that sandboxes could end up bouncing between different API servers. 

not caching seems good for now, and we can address it long term if we need to.

jakubno:
> It seems sort of complicated/fragile that we have this proxy communicating with the API server over DNS. I might be wrong, but only the proxies are going to be making DNS requests, is that correct?
We didn't want to have all connection over API servers, if API server is proxying, you have to wait till all connections are closed, which is up to 24 hours. Especially if we want to be able to deploy API server often. Also there are probably different requirements for API server and the proxy. 

> (My instinct says that DNS is maybe the wrong protocol for doing this level of IPC between our components. Moving proxying into the other applications means that we can just have a local API to get data to the proxies. We could use a different transport (e.g. gRPC/etc.) which might be easier to manage/understand) but I'm not sure if my instinct is actually correct: removing the extra communication could be a useful simplification, but I'm less convinced that changing protocols would be important.

This is good point, we probably could do that. I was maybe stuck with solution we had already in place, just moved it to the new proxy.

tychoish:
> We didn't want to have all connection over API servers, if API server is proxying, you have to wait till all connections are closed, which is up to 24 hours.

Depends on how resources are allocated, but if we're just talking about a docker container running a go process, it maybe doesn't matter if the old process hangs around for a day without doing very much. Having said that...

The default `http.Server` has, I'm pretty sure, default per-request time-outs. If we need requests to last longer than this, we might need to change that in this PR. 

Also, and correct me if I'm wrong, but it's _very_ unlikely that a _connection_ would last for 24 hours. The sandbox might, but a single (HTTP) connection is _very_ unlikely to. The proxy only needs to live as long as the longest request, not the sandbox.

> This is good point, we probably could do that. I was maybe stuck with solution we had already in place, just moved it to the new proxy.

Seems totally correct.

jakubno:
Yes, isn't it harder to manage if you have basically more stateful resources?

packages/client-proxy/main.go
jakubno: There should be only 1 response

PR #277, opened 2025-02-05
Kill paused sandboxes

mishushakov:
Removes paused sandboxes from the DB

### DIFF
--- a/packages/api/internal/handlers/sandbox_kill.go
+++ b/packages/api/internal/handlers/sandbox_kill.go
@@ -92,5 +92,4 @@ func (a *APIStore) DeleteSandboxesSandboxID(
 	telemetry.ReportEvent(ctx, "deleted paused sandbox from db")
 
 	c.Status(http.StatusNoContent)
-	return
 }
--- a/packages/api/internal/handlers/sandbox_kill.go
+++ b/packages/api/internal/handlers/sandbox_kill.go
@@ -64,6 +64,15 @@ func (a *APIStore) DeleteSandboxesSandboxID(
 		return
 	}
 
+	dbErr := a.db.DeleteEnv(ctx, env.ID)
+	if dbErr != nil {
+		errMsg := fmt.Errorf("error when deleting env from db: %w", dbErr)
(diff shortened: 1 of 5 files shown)


packages/api/internal/handlers/sandbox_kill.go
```diff
@@ -27,30 +28,69 @@ func (a *APIStore) DeleteSandboxesSandboxID(
...
+	deleteJobErr := a.templateManager.DeleteBuilds(ctx, buildIds)
+	if deleteJobErr != nil {
+		errMsg := fmt.Errorf("error when deleting env files from storage: %w", deleteJobErr)
 		telemetry.ReportCriticalError(ctx, errMsg)
+	} else {
+		telemetry.ReportEvent(ctx, "deleted env from storage")
+	}
 
-		a.sendAPIStoreError(c, http.StatusUnauthorized, fmt.Sprintf("Error killing sandbox - sandbox '%s' does not belong to your team '%s'", sandboxID, teamID.String()))
+	dbErr := a.db.DeleteEnv(ctx, env.ID)
```
jakubno: I think you should first delete from db. If this fails you will end up in inconsistent state, template is still in db, but you don't have files

jakubno (changes requested):

packages/api/internal/handlers/sandbox_kill.go
```diff
@@ -27,30 +28,69 @@ func (a *APIStore) DeleteSandboxesSandboxID(
...
+	} else {
+		telemetry.ReportEvent(ctx, "deleted env from storage")
 	}
 
+	a.templateCache.Invalidate(env.ID)
+
+	telemetry.ReportEvent(ctx, "deleted paused sandbox from db")
+
 	c.Status(http.StatusNoContent)
+	return
```
jakubno: No need for this
```suggestion
```

jakubno (approved):

jakubno (changes requested):
Shouldn't you also delete env builds?

jakubno (approved):

PR #279, opened 2025-02-05
Fix clock sync on start of the sandbox

jakubno:
In some distributions there isn't `bash` on `/usr/bin/bash` and `date` on `/usr/bin/date` which resulted in not syncing


ValentaTomas (approved):
I'm not sure if `/usr/sbin/phc_ctl` does not have the same problem, but this is an improvement even without that.

PR #283, opened 2025-02-07
Correct wait group and add exit with appropriate code

jakubno:


packages/client-proxy/main.go
```diff
@@ -125,15 +135,15 @@ func main() {
 			// this probably shouldn't happen...
 			logger.Error("http service exited without error", zap.Int("port", port))
 		}
-		wg.Done()
 	}()
 
 	// Proxy request to the correct node
 	server := http.Server{Addr: fmt.Sprintf(":%d", port)}
```
tychoish: I think we should declare the `http.Server` as pointers (e.g. "&http.Server{}`). I think it works ok  in it's current form, but if we accidentally pass it without it being a pointer, then it ends up being a shallow copy, and that seems bad.

packages/client-proxy/main.go
```diff
@@ -164,4 +173,7 @@ func main() {
 		logger.Error("http service exited without error", zap.Int("port", port))
```
tychoish: I might be inclined to have this also be in a go routine (in the wait group.) 

Particularly since there are two services, I'd also be inclined to cancel the context in (at least) the error case (e.g. line 160). so that if one server crashes or exits before shutdown, you would want the rest of the service to shutdown.

packages/client-proxy/main.go
```diff
@@ -38,6 +39,13 @@ func proxy(logger *zap.SugaredLogger) func(w http.ResponseWriter, r *http.Reques
...
 
 		// Extract sandbox id from the sandboxID (<port>-<sandbox id>-<old client id>.e2b.dev)
+		hostSplit := strings.Split(r.Host, "-")
+		if len(hostSplit) < 2 {
+			logger.Warn("invalid host", zap.String("host", r.Host))
+			http.Error(w, "Invalid host", http.StatusBadRequest)
+			return
+		}
+
 		sandboxID := strings.Split(r.Host, "-")[1]
```
tychoish: we can use `hostSplit` rather than calling `strings.Split` a second time.

packages/client-proxy/main.go
```diff
@@ -112,16 +114,16 @@ func main() {
 		panic(fmt.Errorf("error creating logger: %v", err))
 	}
 
-	healthServer := http.Server{Addr: fmt.Sprintf(":%d", healthCheckPort)}
+	healthServer := &http.Server{Addr: fmt.Sprintf(":%d", healthCheckPort)}
+	healthServer.Handler = http.HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {
+		logger.Debug("Health check")
```
tychoish: this is going to get hit a bunch, and not provide a lot of great insight?

packages/client-proxy/main.go
```diff
@@ -138,9 +140,28 @@ func main() {
...
-	server := http.Server{Addr: fmt.Sprintf(":%d", port)}
+	server := &http.Server{Addr: fmt.Sprintf(":%d", port)}
 	server.Handler = http.HandlerFunc(proxy(logger))
 
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+
+		logger.Info("http service starting", zap.Int("port", port))
+		err = server.ListenAndServe()
```
tychoish:
```suggestion
		err := server.ListenAndServe()
```

packages/client-proxy/main.go
```diff
@@ -138,9 +140,28 @@ func main() {
 	}()
 
 	// Proxy request to the correct node
-	server := http.Server{Addr: fmt.Sprintf(":%d", port)}
+	server := &http.Server{Addr: fmt.Sprintf(":%d", port)}
 	server.Handler = http.HandlerFunc(proxy(logger))
 
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
```
tychoish, line 147: also cancel the root context in a defer here so that we don't block waiting for a sigterm waiting for the health endpoint if there's a crash or problem with this service.

PR #292, opened 2025-02-11
Accept Supabase OAuth token

dobrac:
Allow using Supabase OAuth token instead of Team API Keys or user Access Tokens as authentication method. 

Usage requires following headers for Team API endpoints:
```
X-Supabase-Token: ${Supabase JWT}
X-Supabase-Team: ${Team ID}
```

And following for Access Token endpoints:

```
X-Supabase-Token: ${Supabase JWT}
```

This change requires also infrastructure update and setting up `supabase_jwt_secrets` in Google Secret Manager. Multiple secrets are possible, delimited by `,` in order to support key rotation.

Note: The naming securitySchemes with 1 and 2 is there for having the verification in correct order.


packages/api/internal/auth/middleware.go
```diff
@@ -22,51 +22,74 @@ import (
...
 )
 
-type authenticator[T any] struct {
+type headerKey struct {
+	name         string
+	prefix       string
+	removePrefix string
+}
+
+func newHeaderKey(name, prefix, removePrefix string) headerKey {
```
0div: nit: redundant & doesn't enhance readability if you're only passing properties for a struct w/o methods nor validation and it still doesn't guard against nil values.

packages/api/internal/auth/middleware.go
```diff
@@ -22,51 +22,74 @@ import (
...
 
-// getApiKeyFromRequest extracts an API key from the header.
-func (a *authenticator[T]) getAPIKeyFromRequest(req *http.Request) (string, error) {
-	apiKey := req.Header.Get(a.headerKey)
+type authenticator interface {
+	Authenticate(ctx context.Context, input *openapi3filter.AuthenticationInput) error
+	SecuritySchemeName() string
+}
+
+// getHeaderKeysFromRequest extracts header keys from the header.
```
0div: nit: function name is self-explanatory

packages/api/internal/auth/middleware.go
```diff
@@ -106,51 +133,62 @@ func adminValidationFunction(_ context.Context, token string) (struct{}, *api.AP
...
-		validationFunction: adminValidationFunction,
-		contextKey:         "",
-		errorMessage:       "Invalid Access token.",
+func CreateAuthenticationFunc(
+	tracer trace.Tracer,
+	teamValidationFunction func(context.Context, string) (authcache.AuthTeamInfo, *api.APIError),
+	userValidationFunction func(context.Context, string) (uuid.UUID, *api.APIError),
+	supabaseTokenValidationFunction func(context.Context, string) (uuid.UUID, *api.APIError),
+	supabaseTeamValidationFunction func(context.Context, string) (authcache.AuthTeamInfo, *api.APIError),
+) func(ctx context.Context, input *openapi3filter.AuthenticationInput) error {
```
0div: nit: maybe u could type this function for readability.

`type AuthenticationFunc func(ctx context.Context, input *openapi3filter.AuthenticationInput) error`

packages/shared/pkg/db/apiKeys.go
```diff
@@ -4,13 +4,34 @@ import (
...
+	}
+
+	if team.IsBlocked {
+		if team.BlockedReason == nil {
+			errMsg := fmt.Errorf("team was blocked")
+
+			return errMsg
+		}
+
+		errMsg := fmt.Errorf("team was blocked - %s", *team.BlockedReason)
```
0div:
```suggestion
		errMsg := fmt.Errorf("team was blocked, reason: %s", *team.BlockedReason)
```

packages/api/internal/handlers/store.go
```diff
@@ -35,6 +37,8 @@ const (
 	defaultRequestLimit = 16
 )
 
+var supabaseJWTSecret = os.Getenv("SUPABASE_JWT_SECRET")
```
0div: not for this PR, but we should consider using Hashicorp Vault or something similar bc making this secret readily available in the machine's environment is less safe.

0div (approved):
LGTM, mostly nits, some questions tho:
- possible to write a test since this is now abstracted as an interface?
- if not, how to test this manually?
- what is the mitigation plan if secret key leaks? (and how to better store it?) 
- how will we handle secret key rotation?

packages/api/internal/handlers/store.go
dobrac: I agree

packages/api/internal/auth/middleware.go
dobrac: +1, refactorings relict

packages/api/internal/utils/error.go
```diff
@@ -35,11 +41,57 @@ func ErrorHandler(c *gin.Context, message string, statusCode int) {
...
+
+			err = errW
+			break
+		}
+
+		return fmt.Errorf("%s%s", securityErrPrefix, err.Error())
+	default:
+		// This should never happen today, but if our upstream code changes,
+		// we don't want to crash the server, so handle the unexpected error.
+		return fmt.Errorf("error validating request: %w", err)
```
jakubno, line 95: doesn't it make sense to just to return `err`?

spec/openapi.yml
```diff
@@ -16,6 +16,14 @@ components:
       type: http
       scheme: bearer
       bearerFormat: access_token
+    Supabase1TokenAuth:
+      type: apiKey
+      in: header
+      name: X-Supabase-Token
+    Supabase2TeamAuth:
```
jakubno, line 26: is there a reason for numbers (1,2)?

packages/api/internal/handlers/store.go
```diff
@@ -210,3 +214,74 @@ func (a *APIStore) GetUserFromAccessToken(ctx context.Context, accessToken strin
 
 	return *userID, nil
 }
+
+// supabaseClaims defines the claims we expect from the Supabase JWT.
+type supabaseClaims struct {
+	jwt.RegisteredClaims
+}
+
+func (a *APIStore) GetUserFromSupabaseToken(ctx context.Context, supabaseToken string) (uuid.UUID, *api.APIError) {
```
jakubno:
```suggestion
func (a *APIStore) GetUserIDFromSupabaseToken(ctx context.Context, supabaseToken string) (uuid.UUID, *api.APIError) {
```

packages/api/internal/handlers/store.go
```diff
@@ -210,3 +214,74 @@ func (a *APIStore) GetUserFromAccessToken(ctx context.Context, accessToken strin
...
+
+	return uuid.UUID{}, &api.APIError{
+		Err:       fmt.Errorf("supabase token is invalid: %w", err),
+		ClientMsg: "Backend authentication failed",
+		Code:      http.StatusUnauthorized,
+	}
+}
+
+func (a *APIStore) GetTeamFromSupabaseToken(ctx context.Context, teamID string) (authcache.AuthTeamInfo, *api.APIError) {
+	userID := a.GetUserID(middleware.GetGinContext(ctx))
```
jakubno, line 323: how does this work? this isn't set yet, right?

packages/api/internal/handlers/store.go
```diff
@@ -210,3 +214,74 @@ func (a *APIStore) GetUserFromAccessToken(ctx context.Context, accessToken strin
...
+
+	team, tier, err := a.db.GetTeamByIDAndUserIDAuth(ctx, teamID, userID)
+	if err != nil {
+		return authcache.AuthTeamInfo{}, &api.APIError{
+			Err:       fmt.Errorf("failed getting team: %w", err),
+			ClientMsg: "Backend authentication failed",
```
jakubno, line 338: This will be returned if team is blocked, could you propagate the correct error?

jakubno (changes requested):
I think we should add some caching for the db query, we probably don't want to do DB request on each request

spec/openapi.yml
dobrac: the generator generates validation of the auth methods in the alphabetical order. Because I need the token to be verified first, I've added the numbers there

packages/api/internal/handlers/store.go
dobrac: it is, as the userID verification method is called first

packages/api/internal/utils/error.go
dobrac: it might be difficult to find this place in the code if it eventually happens, otherwise sure

jakubno: make sense, let's keep it like this

spec/openapi.yml
jakubno: Write down this wisdom 🙏🏻

packages/api/internal/handlers/store.go
jakubno: Now I understand, totally missed there's no `-` in oapi spec and though it's just different way how to authenticate.

spec/openapi.yml
dobrac: it's (and always was) in the PR description, I will add it to the code too

dobrac:
Added caching and tests (unit + integration)

packages/shared/pkg/db/auth.go
```diff
@@ -0,0 +1,46 @@
...
+		errMsg := fmt.Errorf("failed to parse team ID: %w", err)
+
+		return nil, nil, errMsg
+	}
+
+	result, err := db.
+		Client.
+		Team.
+		Query().
+		WithUsersTeams().
```
jakubno: do you need user teams? It isn't in `GetTeamAuth` and they are both used in the cache. Or is it purpose, because I see you are using the cache with different keys.

packages/api/internal/cache/auth/cache.go
```diff
@@ -30,43 +29,43 @@ type TeamInfo struct {
...
 		}
 
 		templateInfo = &TeamInfo{team: team, tier: tier, lastRefresh: time.Now()}
-		c.cache.Set(apiKey, templateInfo, authInfoExpiration)
+		c.cache.Set(key, templateInfo, authInfoExpiration)
 
 		return team, tier, nil
 	}
 
 	templateInfo = item.Value()
```
jakubno, line 65: I think this wasn't updated correctly, if you don't mind fixing it here 🙏🏻

jakubno (commented):
This and I will approve

packages/api/internal/cache/auth/cache.go
dobrac: I have reused the same cache, but you can either have apiKey as "key" or teamId as "key" (of the cache entry), that's why I have only renamed it. Yes, you could have there two entries for the same data, but they'll live separately as I'm not able to easily recover teamId from apiKey

packages/shared/pkg/db/auth.go
dobrac: removed WithUsersTeams

jakubno (approved):

PR #295, opened 2025-02-13
Team API Keys & Access Token management with hashing

dobrac:
**Changes**
- Database migration for api keys and access tokens hasing


packages/api/internal/auth/hashing.go
```diff
@@ -0,0 +1,5 @@
+package auth
+
+type Hashing interface {
```
0div: nit: try to use verbiage for interface names, e.g. `Hasher`

packages/api/internal/auth/middleware.go
```diff
@@ -22,51 +22,74 @@ import (
```
0div: i saw these changes here too? 
https://github.com/e2b-dev/infra/pull/292/files#diff-e365902d776242307d149b5817557d68a8824478ade7d07f0686613bf0f7508b

packages/api/internal/handlers/apikey.go
```diff
@@ -0,0 +1,168 @@
...
+
+		errMsg := fmt.Errorf("error when updating team API key name: %w", err)
+		telemetry.ReportCriticalError(ctx, errMsg)
+		return
+	}
+
+	c.Status(http.StatusAccepted)
+}
+
+func (a *APIStore) GetApikeys(c *gin.Context) {
```
0div: To be clear, we're getting the Hashed API Key, correct?

packages/api/internal/handlers/store.go
```diff
@@ -4,12 +4,14 @@ import (
 	"context"
```
0div: saw this in https://github.com/e2b-dev/infra/pull/292/ as well

packages/api/internal/auth/key.go
```diff
@@ -0,0 +1,44 @@
...
+type Key struct {
+	PrefixedRawValue string
+	HashedValue      string
+	MaskedValue      string
+}
+
+func MaskKey(prefix string, value string) string {
+	suffixLength := keySuffixLength
+
+	lastFour := value[len(value)-suffixLength:]
```
0div:
```suggestion
	lastFour := value[len(value)-keySuffixLength:]
```

0div: - is it always "the last four" ? in which case maybe being more explicit about that expected count in the naming of hardcoded values. 
- boring but good idea to apply defensive programming for critical areas like this, e.g. validate the length of `value` (even if expected to be sha sum) beforehand to avoid a negative value after subtraction.

packages/api/internal/handlers/apikey.go
```diff
@@ -0,0 +1,168 @@
...
+			}
+		}
+
+		keyValue := strings.Split(apiKey.APIKey, auth.ApiKeyPrefix)[1]
+
+		teamAPIKeys[i] = api.TeamAPIKey{
+			Id:   apiKey.ID,
+			Name: apiKey.Name,
+			// TODO: remove this once we migrate to hashed API keys
+			KeyMask:   auth.MaskKey(auth.ApiKeyPrefix, keyValue),
```
0div: im not sure i underatand why we do this? maybe add an example or further explanation.

0div (approved):
LGTM, although—not to be that guy—but needs tests.

packages/api/internal/auth/middleware.go
dobrac: >(includes changes from https://github.com/e2b-dev/infra/pull/292)

I needed to build new stuff, but the previous PR was not merged yet

packages/api/internal/handlers/store.go
dobrac: > (includes changes from https://github.com/e2b-dev/infra/pull/292)

packages/api/internal/handlers/apikey.go
dobrac: We need the not-hashed API keys until Dashboard and Fragments are migrated to use the Supabase token. And for them to start implementing it, we need to merge the #292 PR first.

packages/api/internal/handlers/apikey.go
dobrac: No, only masked API keys (e2b_******abcd)

packages/api/internal/auth/key.go
dobrac: > is it always "the last four" ?

Right now that's how we've defined the masked API key to look like. It doesn't have any real effect on the keys though, this is just for the user. What would be your proposal for the renaming?

> boring but good idea to apply defensive programming

Okay, I will add it

0div: ohh, i see where my confusion stemmed from. The term "mask" for me has another meaning in the byte manipulation/crypto world. To be clearer i would probably renamed this as "hiding"

dobrac: do you mean HideKey? I was just following the naming convention we've already had there

0div: yeah `HideKey` is clearer IMO

dobrac: I would keep the MaskKey as I think the mask is actually used to do exactly this action (based on what I've found)

dobrac:
> LGTM, although—not to be that guy—but needs tests.

Added unit and integration tests

packages/shared/pkg/schema/access_token.go
```diff
@@ -15,7 +15,11 @@ type AccessToken struct {
 
 func (AccessToken) Fields() []ent.Field {
 	return []ent.Field{
+		field.UUID("unique_id", uuid.UUID{}).StorageKey("id").Immutable().Unique().Annotations(entsql.Default("gen_random_uuid()")),
 		field.String("id").Unique().StorageKey("access_token").Immutable().SchemaType(map[string]string{dialect.Postgres: "text"}),
```
jakubno: I think you can do it directly like this, otherwise it will be confusing 
```suggestion
		field.UUID("id", uuid.UUID{}).Immutable().Unique().Annotations(entsql.Default("gen_random_uuid()")),
		field.String("access_token").Immutable().SchemaType(map[string]string{dialect.Postgres: "text"}),
```

packages/shared/migrations/20250211160814.sql
```diff
@@ -0,0 +1,15 @@
+-- Add new columns to team_api_keys table
```
jakubno, line 1: I think you need to populate old data

packages/shared/pkg/keys/key.go
```diff
@@ -0,0 +1,56 @@
...
+	"strings"
+)
+
+const (
+	keySuffixLength = 4
+
+	keyLength = 20
+)
+
+var hasher Hasher = NewSHA256Hashing()
```
jakubno, line 16: SHA 256 was chosen because of speed right? But is it considered safe enough?

packages/api/internal/handlers/teams.go
```diff
@@ -34,10 +35,18 @@ func (a *APIStore) GetTeams(c *gin.Context) {
 
 	teams := make([]api.Team, len(teamsDB))
 	for i, teamDB := range teamsDB {
+		apiKey, err := team.CreateAPIKey(ctx, a.db, teamDB.ID, userID, "CLI login/configure")
```
jakubno, line 38: what if reimplement the logic in CLI and then just return here to update CLI? You can't run in pipeline and if you run it manually it's probably okay with reasonable error

packages/api/internal/handlers/apikey.go
```diff
@@ -0,0 +1,181 @@
...
+	for i, apiKey := range apiKeysDB {
+		var createdBy *api.TeamUser
+		if apiKey.Edges.Creator != nil {
+			createdBy = &api.TeamUser{
+				Email: apiKey.Edges.Creator.Email,
+				Id:    apiKey.Edges.Creator.ID,
+			}
+		}
+
+		keyValue := strings.Split(apiKey.APIKey, keys.ApiKeyPrefix)[1]
```
jakubno, line 85: why not to use the mask from team api key directly?

packages/api/internal/handlers/accesstoken.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	accessTokenDB, err := a.db.Client.AccessToken.
+		Create().
+		SetUniqueID(uuid.New()).
+		SetUserID(userID).
+		SetID(accessToken.PrefixedRawValue).
+		SetAccessTokenHash(accessToken.HashedValue).
+		SetAccessTokenMask(accessToken.MaskedValue).
+		SetCreatedAt(time.Now()).
```
jakubno: this should be set automatically

spec/openapi.yml
```diff
@@ -1125,5 +1258,133 @@ paths:
           $ref: "#/components/responses/401"
         "404":
           $ref: "#/components/responses/404"
+        "500":
+          $ref: "#/components/responses/500"
+
+  /accesstokens:
```
jakubno: i would probably prefere `/access-tokens`

spec/openapi.yml
```diff
@@ -1125,5 +1258,133 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/CreatedAccessToken"
+        "401":
+          $ref: "#/components/responses/401"
+        "500":
+          $ref: "#/components/responses/500"
+
+  /accesstokens/{accessTokenID}:
```
jakubno: There's no update for access token name

spec/openapi.yml
```diff
@@ -1125,5 +1258,133 @@ paths:
...
+        "204":
+          description: Access token deleted successfully
+        "401":
+          $ref: "#/components/responses/401"
+        "404":
+          $ref: "#/components/responses/404"
+        "500":
+          $ref: "#/components/responses/500"
+
+  /apikeys:
```
jakubno: `api-keys/`? The generated methods also look better that way

packages/shared/migrations/20250211160814.sql
dobrac: this will be done in a next PR (https://github.com/e2b-dev/infra/pull/298), you have to first start generating the hashes alongside new keys

packages/shared/pkg/keys/key.go
dobrac: it's in the PR description, basically using anything else is a waste of time and doesn't add more security, as the keys are already secure random

packages/api/internal/handlers/apikey.go
dobrac: it might not be populated for old keys, this is solved in the next PR (https://github.com/e2b-dev/infra/pull/298) when this one is deployed

packages/api/internal/handlers/accesstoken.go
dobrac: what do you mean?

spec/openapi.yml
dobrac: it's not necessary rn as it's not directly accessed or used anywhere, even the name is there for internal purpose only

packages/shared/pkg/keys/key.go
jakubno: Make sense, thanks for explanation!

packages/api/internal/handlers/teams.go
dobrac: I didn't want to make this much breaking change as it would immediately break all old cli versions

dobrac:
> I am missing logs, at least for error and info in the end of an successful interaction as deleting an api key

all http endpoints are logged with their results, it's it is an internal error, we log also the reason (sendAPIStoreError)

jakubno (approved):

PR #297, opened 2025-02-13
orchestrator: process initialization and shut down

tychoish:
I'm still running tests on this, but it hardens the initialization and cleanup of the orchestrator process.


packages/orchestrator/main.go
```diff
@@ -18,28 +21,54 @@ func main() {
...
+	defer sigCancel()
 
+	var port uint
+
+	flag.UintVar(&port, "port", defaultPort, "orchestrator server port")
 	flag.Parse()
 
+	wg := &sync.WaitGroup{}
+	exitCode := &atomic.Int32{}
+	telemtrySignal := make(chan struct{})
```
dobrac: nit: telemetry

packages/orchestrator/internal/server/main.go
```diff
@@ -59,22 +68,74 @@ func New() (*grpc.Server, error) {
...
+			grpc.ChainUnaryInterceptor(
+				recovery.UnaryServerInterceptor(),
+			),
+		)
+
+		srv.server = &server{
+			tracer:        otel.Tracer(ServiceName),
+			dns:           srv.dns,
+			sandboxes:     smap.New[*sandbox.Sandbox](),
+			networkPool:   networkPool,
```
dobrac, line 85: I don't see any closing of the networkPool, templateCache, maybe even running sandboxes, is that desired?

packages/orchestrator/internal/server/main.go
```diff
@@ -59,22 +68,74 @@ func New() (*grpc.Server, error) {
...
+	}
+
+	go func() {
+		log.Printf("Starting DNS server")
+		if err := srv.dns.Start("127.0.0.4", 53); err != nil {
+			log.Panic(fmt.Errorf("Failed running DNS server: %w", err))
+		}
+	}()
+
+	// the listener is closed by
```
dobrac: closed by _

packages/orchestrator/internal/server/main.go
tychoish: It's my understanding that when the orchestrator crashes/exits, all of the running sandboxes are orphaned and the machine (essentially) has to start from scratch. While we could clean up these caches, and maybe we should on principle, I don't think that it matters at the moment. I do worry that if we make shut down block on something that doesn't provide us with anything, we just make shutdown take longer which might be more disruptive.

Eventually, I think we'll want to do some changes in the orchestrator so we can restart it/deploy it independently of restarting the sandboxes (potentially) which requires a change in process management.

Anyway, we're not shutting down the caches now, so I think it's safe to merge this without adding it, and it'd be possible to add this in the future.

tychoish (commented):
made some changes in response to comments, and other issues observed in a self-review.

packages/orchestrator/internal/server/main.go
dobrac: I'm not entirely sure if orphaned sandboxes without orchestrator will run correctly. Because, they rely on providing the memory and rootfs data on demand, which not running orchestrator can't supply.

Otherwise I agree that we don't fully care right now about proper clean up as we always throw away the node and recreate it from scratch. It might be nice though to be able to restart orchestrator without having issues with network namespaces etc (for example for development).

I'm fine merging and doing more updates in future PRs. (Not fully ready yet - It would be nice to have also some automated tests)

tychoish: I agree that the goal is to be able to restart the orchestrator without killing sandboxes would be good, but definitely down the road, but that has less to do with cache/data management, and more changing the way that the orchestrator manages the process. I think there are two (or three depending) options:
- we write a very thin "runner" program that runs the sandbox, but the orchestrator starts the process and catches signals such sandboxes outlive the parent process. would need to persist data about the sandboxes to disk and teach the orchestrator to identify and manage processes it didn't start. 
- split the orchestrator into two, a management side (responsible for the lifecycle of sandboxes and the instances they run on) and a process management side (just runs the sandboxes with a very minimal API.)
- as an alternative, use `systemd` (probably user-local) as the process manager, so it can own the lifecycle of the processes and the orchestrator is just a controller for systemd, with its data saved to disk. This could use `systemd-machined` or just conventional (ephemeral) systemd services.

Anyway, I don't think any change is necessary to this part of the code.

dobrac (approved):

packages/orchestrator/internal/server/main.go
```diff
@@ -59,22 +68,78 @@ func New() (*grpc.Server, error) {
...
+		log.Printf("Starting DNS server")
+		if err := srv.dns.Start("127.0.0.4", 53); err != nil {
+			log.Panic(fmt.Errorf("Failed running DNS server: %w", err))
+		}
+	}()
+
+	// the listener is closed by the shutdown operation
+	lis, err := net.Listen("tcp", fmt.Sprintf(":%d", srv.port))
+	if err != nil {
+		return fmt.Errorf("failed to listen on port %d: %w", srv.port, err)
```
jakubno, line 112: if this happens, wouldn't this result in panic? `srv.closer.op` won't be defined, but `Close` will be called

packages/orchestrator/internal/server/main.go
```diff
@@ -59,22 +68,78 @@ func New() (*grpc.Server, error) {
...
+		srv.grpc.GracefulStop()
 
-	grpc_health_v1.RegisterHealthServer(s, health.NewServer())
+		if err := srv.dns.Close(ctx); err != nil {
+			errs = append(errs, err)
+		}
+
+		if err := lis.Close(); err != nil {
+			errs = append(errs, err)
+		}
```
jakubno: I would probably switch these two, not that important. It's similar to API, you can probably serve some traffic if DNS is up, on the other side the server is in this point probably useless

packages/orchestrator/main.go
```diff
@@ -18,28 +23,87 @@ func main() {
...
+	defer sigCancel()
 
+	var port uint
+
+	flag.UintVar(&port, "port", defaultPort, "orchestrator server port")
 	flag.Parse()
 
+	wg := &sync.WaitGroup{}
+	exitCode := &atomic.Int32{}
+	telemetrySignal := make(chan struct{})
```
jakubno, line 36: this isn't used, right? Something is probably missing

jakubno (changes requested):

packages/orchestrator/internal/server/main.go
tychoish: good catch.

packages/orchestrator/internal/server/main.go
tychoish: I think because we call graceful stop on line 126 that this doesn't matter, but I don't think  this order is super meaningful. 

There's the alternate argument that because all of the processes that requests would target are _children_ of the orchestrator, we should stop DNS as early as possible because we want traffic to cleanly halt and stopping DNS before we stop the workload means we we get a (potentially retriable) error rather than have a hanging request or return corrupt/partial data. 

I don't it particularly matters, and I've swapped the order.

packages/orchestrator/main.go
tychoish: we block on it in 47 and close it in 98... just so that telemetry is the last thing to shut down.

tychoish (commented):
pushed change relating to this feedback.

packages/orchestrator/main.go
jakubno: Sorry,  I missed the close

jakubno (approved):

PR #307, opened 2025-02-18
Add support for multiple domains on the same cluster

jakubno:
# Description

This allows us to have multiple domains on the same cluster

### DIFF
--- a/packages/cluster/network/main.tf
+++ b/packages/cluster/network/main.tf
@@ -113,6 +113,7 @@ resource "cloudflare_record" "a_star" {
   name    = "*"
   value   = google_compute_global_forwarding_rule.https.ip_address
   type    = "A"
+  comment = var.gcp_project_id
 }
 
 data "cloudflare_zone" "domains_additional" {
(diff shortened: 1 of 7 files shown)


ValentaTomas:
Looks ok to me, if you tested we can merge.

packages/cluster/network/main.tf
```diff
@@ -16,6 +16,7 @@ provider "cloudflare" {
 }
 
 locals {
+  domain_map = { for d in var.additional_domains : split(".", d)[0] => d }
```
dobrac: What if there are two domains like "zulu.dev" and "zulu.com"?

Maybe using set might be better here.

variables.tf
```diff
@@ -182,6 +182,12 @@ variable "domain_name" {
   description = "The domain name where e2b will run"
 }
 
+variable "additional_domains" {
+  type        = string
+  description = "Additional domains which can be used to access the e2b cluster"
```
dobrac: Could you please add that they are delimited by ","? It might be useful when defining the vars

dobrac (changes requested):

dobrac (approved):

PR #309, opened 2025-02-18
Stop saving empty blocks during build

dobrac:
This PR is a cherry-pick of the implementation from https://github.com/e2b-dev/infra/pull/234

**Changes**:
- Ignore nil diff blocks for memfile (upload and download)
- Ignore nil diff blocks for rootfs (upload and download)

Subtasks:
- [x] _Handle empty block in orchestrator (was already implemented)_
- [x] Start creating empty blocks during template building and upload diff with header instead of the full template
  - [x] Memfile
  - [x] Rootfs
- [x] Handle diffs instead of just full snapshots in orchestrator for normal sandbox start (right now we handle diffs only for resume)

**Testing**
Done manually using newly built base template, comparing the final size and running JS tests.

**Size reduction**
For the base image, the file size differences are following:

|                    | Before  | After (2 GB disk)  |After (50 GB disk)  |
|--------------------|---------|---------|---------|
| memfile            | 512 MB  | 124 MB  |124 MB  |
| memfile.header     | 0 B     | 344 B   |344 B   |
| rootfs.ext4        | 1.9 GB  | 1.3 GB  | 1.3 GB  |
| rootfs.ext4.header | 0 B     | 8.3 KB  | 53.3 KB  |
| snapfile           | 27.2 KB | 27.2 KB | 27.2 KB |


ValentaTomas:
Extra things I would check:
1. Are the "original" templates (that do not have a header) handled properly with the changes to the orchestrator?
2. How much extra time does that transformation take during the build?
3. Snapshotting and then resume is ok?

Also note about the `rootfs.ext4` — it does normally have 5GB free space on top of the used size, so I think you are measuring either without that or before adding that?

dobrac:
> 1. Are the "original" templates (that do not have a header) handled properly with the changes to the orchestrator?
Yes. Tested on this PR infra (old build template, new built template). Tested also pausing by manual testing (and SDK tests).

> 2. How much extra time does that transformation take during the build?
(measured locally using `time` for the 2nd build (1024 MB memory)
Without nil optimization: 0.74s user 0.29s system 1% cpu 59.306 total,  'create full snapshot' API request took 2646396 us.
With nil optimization: 0.80s user 0.31s system 1% cpu 1:01.20 total, 'create full snapshot' API request took 2593117 us.

> 3. Snapshotting and then resume is ok?
This is working (manual testing) as well as currently written tests in the SDK (autopause branch)

> Also note about the `rootfs.ext4` — it does normally have 5GB free space on top of the used size, so I think you are measuring either without that or before adding that?
I think it's set to 2GB for the dev env.

packages/orchestrator/internal/sandbox/template/storage.go
```diff
@@ -23,20 +24,23 @@ func NewStorage(
 	buildId string,
 	fileType build.DiffType,
 	blockSize int64,
-	isSnapshot bool,
 	h *header.Header,
 	bucket *gcs.BucketHandle,
 ) (*Storage, error) {
-	if isSnapshot && h == nil {
+	if h == nil {
```
ValentaTomas, line 30: We should be sure this works for both "old" (not having headers) and "new" (having headers) templates. After we deploy this we might think about rebuilding all templates so there can be one less check and logic.

packages/shared/pkg/storage/header/mapping.go
```diff
@@ -39,9 +38,9 @@ func CreateMapping(
 
 		if blockLength > 0 {
 			m := &BuildMap{
-				Offset:             uint64(int64(startBlock) * int64(metadata.BlockSize)),
+				Offset:             uint64(int64(startBlock) * int64(blockSize)),
 				BuildId:            *buildId,
-				Length:             uint64(blockLength) * uint64(metadata.BlockSize),
+				Length:             uint64(blockLength) * uint64(blockSize),
 				BuildStorageOffset: buildStorageOffset,
```
ValentaTomas, line 45: Was there before) Unnecessary conversion
```suggestion
				Length:             uint64(blockLength) * blockSize,
```

packages/shared/pkg/storage/header/mapping.go
```diff
@@ -39,9 +38,9 @@ func CreateMapping(
 
 		if blockLength > 0 {
 			m := &BuildMap{
-				Offset:             uint64(int64(startBlock) * int64(metadata.BlockSize)),
+				Offset:             uint64(int64(startBlock) * int64(blockSize)),
 				BuildId:            *buildId,
```
ValentaTomas, line 43: (Was there before) Unnecessary conversion
```suggestion
				Offset:             uint64(startBlock) * blockSize,
```

packages/orchestrator/internal/sandbox/template/storage.go
```diff
@@ -23,20 +24,23 @@ func NewStorage(
...
 		}
 
-		h = diffHeader
-	} else if h == nil {
+		if err == nil {
+			h = diffHeader
+		}
+	}
+
+	if h == nil {
```
ValentaTomas, line 43:
```suggestion
		diffHeader, err := header.Deserialize(headerObject)
		if err != nil && !errors.Is(err, gcs.ErrObjectNotExist) {
			return nil, fmt.Errorf("failed to deserialize header: %w", err)
		}

		if err == nil {
			h = diffHeader
		}
	}

  // If we can't find the diff header in storage we try to find the "old" style template without header as a fallback.
	if h == nil {
```

packages/orchestrator/internal/sandbox/template/storage.go
```diff
@@ -23,20 +24,23 @@ func NewStorage(
...
 			return nil, fmt.Errorf("failed to deserialize header: %w", err)
 		}
 
-		h = diffHeader
-	} else if h == nil {
+		if err == nil {
+			h = diffHeader
+		}
+	}
+
```
ValentaTomas, line 42:
```suggestion
  // If we can't find the diff header in storage we try to find the "old" style template without header as a fallback.
```

packages/orchestrator/internal/sandbox/template/storage.go
dobrac: I have already done manual testing (creating a new sandbox from old and new template) and it worked. The only issue is, that after we rebuild, or deploy this optimization, we can't revert this change

ValentaTomas:
I think this is ready—we just need to me careful with deploying the orchestrators with these changes first.

ValentaTomas (approved):

PR #305, opened 2025-02-17
Local grafana

r33drichards:

PR #310, opened 2025-02-18
use secret data instead of referencing secret by name to avoid data races

r33drichards:
okay! so now I have [intermediate step](https://github.com/e2b-dev/infra/pull/305/files)  and [local grafana w/o data races](https://github.com/e2b-dev/infra/pull/305/files) .  The changes should be merged as:

`main` <- `intermediate step`  <- `local grafana w/o races`

in the GitHub pr i'm branching off of `intermediate step` b/c of this.

I have tested locally going back and forth from `main` to `inter-step` and back without failure and `intermediate step` and `local-grafana` by running make plan and make apply w/o errors so i'm feeling pretty confident this will accomplish the task and also not break things in an annoying way

### DIFF
--- a/packages/init/outputs.tf
+++ b/packages/init/outputs.tf
@@ -15,31 +15,31 @@ output "nomad_acl_token_secret" {
 }
 
 output "grafana_api_key_secret_name" {
-  value = google_secret_manager_secret.grafana_api_key.name
+  value = google_secret_manager_secret_version.grafana_api_key.secret_data
 }
 
 output "grafana_logs_username_secret_name" {
-  value = google_secret_manager_secret.grafana_logs_username.name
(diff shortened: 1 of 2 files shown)


ValentaTomas:
The PR has unclear name—I'm not sure what this is doing.

r33drichards:
i updated the description, each PR is still WIP but should be ready for review by EOD 🪵🤛

r33drichards:
the description contains the description of the problem holistically, this change is a intermediary step to prevent errors when upgrading due to data races. It accomplishes this by using secret data instead of referencing it by its name in data definitions, which will create errors when they don't exist

jakubno:
Not needed anymore, fixed in #305

PR #312, opened 2025-02-19
Run periodic tests in github actions to determine liveness

r33drichards:
We want to be able to run periodic liveness tests for our infrastructure across multiple clusters. 

This pr introduces some tests and patterns to accomplish this

### DIFF
--- a/.github/workflows/periodic-test.yml
+++ b/.github/workflows/periodic-test.yml
@@ -4,7 +4,7 @@ on:
   schedule:
     # every hour on hour 38
     # 38 to avoid peak traffic
-    - cron: '38 * * * *'
+    - cron: '*/15 * * * *'
   workflow_dispatch:
   # only run on changes to periodic-test dir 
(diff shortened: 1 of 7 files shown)


ValentaTomas:
Can we finish

Also, can the test be more frequent?
We can do every 15 minutes if we delete the template as a cleanup.

.github/workflows/periodic-test.yml
```diff
@@ -0,0 +1,81 @@
+name: Periodic Tests
+
+on:
+  schedule:
+    # every hour on hour 38
+    # 38 to avoid peak traffic
+    - cron: '38 * * * *'
```
r33drichards:
```suggestion
    - cron: '*/15 * * * *'
```

r33drichards:
@ValentaTomas should be good to merge as is whenever you're ready

ValentaTomas (approved):

PR #313, opened 2025-02-19
Add Integration testing to the CI/CD

dobrac:
**Description**
Add integration tests to the CI/CD. This PR runs PostgreSQL, API, and Orchestrator. It allows to write tests for them, e.g. starting a sandbox.

Before starting services, there is also a build of the template, as our prebuilt template is built for Intel CPU and GH Actions are run on AMD CPUs.

**Writing Tests**
In order to write integration tests, there is a new package `integration-tests` with `internal/tests` folder. For now, there is a pre-generated HTTP client for API, but in the future, there may also be a client for the envd.

In order to add new tests, simply add new GoLang tests to existing or new files.

For running tests against own cluster, you need to have these variables in your `.env` file:
```
# Integration tests variables (only for running integration tests locally)
# your domain name, eg great-innovations.dev
TESTS_API_SERVER_URL=
# your sandbox template ID, eg base
TESTS_SANDBOX_TEMPLATE_ID=base
# your Team API key
TESTS_E2B_API_KEY= 
```

**Notes**
There is required access to bucket `e2b-prod-public-builds`, private `e2b-tests-fc-templates`, docker images `custom-environments-tests`. Access is acquired through the `workload_identity_provider`.

**Additional changes**
- use memory instead of KV for local run
- refactor envd init timeout
- change envd init timeout to 60 seconds (it takes longer than 10 seconds for the first sandbox)
- handle template build errors when envd files don't exist
- check firecracker binary presence before starting a container
- fix missing go mod tidy for template manager
- ability to run orchestrator locally


packages/orchestrator/internal/consul/client.go
```diff
@@ -2,20 +2,22 @@ package consul
...
 	"github.com/hashicorp/consul/api"
 
 	"github.com/e2b-dev/infra/packages/shared/pkg/utils"
 )
 
 var (
-	consulToken = utils.RequiredEnv("CONSUL_TOKEN", "Consul token for authenticating requests to the Consul API")
-	Client      = utils.Must(newClient())
+	Client = sync.OnceValue(func() *api.Client {
+		return utils.Must(newClient())
```
ValentaTomas: I think, we can move the initialization from the top level now that it is conditional—the combination of Must here and onceValue confuses me a little on the first glance.

Having it in the https://github.com/e2b-dev/infra/blob/18dbef244be0be82269e79b8e66da3fb859fb1a9/packages/orchestrator/internal/sandbox/network/storage_kv.go#L17 with the `requiredEnv` also there could work as it should not be ideally used elsewhere.

packages/orchestrator/internal/sandbox/network/pool.go
```diff
@@ -108,7 +111,7 @@ func (p *Pool) Get(ctx context.Context) (Slot, error) {
 	default:
 		select {
 		case <-ctx.Done():
-			return Slot{}, ctx.Err()
+			return *NewSlot("", 0), ctx.Err()
```
ValentaTomas: We create the "nil" slot here this way to be explicit?

dobrac: fixed

packages/orchestrator/internal/consul/client.go
dobrac: I agree, refactored

.github/actions/integration-tests/action.yml
```diff
@@ -0,0 +1,19 @@
...
+  steps:
+    - name: Run Integration Tests
+      env:
+        TESTS_E2B_API_KEY: "e2b_5ec17bd3933af21f80dc10bba686691c4fcd7057"
+        TESTS_API_SERVER_URL: "http://localhost:3000"
+      run: |
+        # Monitor logs of the services
+        ls -l ~/logs
+        tail -f ~/logs/orchestrator.log -n 0 &
+        tail -f ~/logs/api.log -n 0 &
```
r33drichards: neat!

.github/actions/host-init/action.yml
```diff
@@ -0,0 +1,44 @@
...
+description: "Sets up the host environment, including authentication and dependencies."
+
+runs:
+  using: "composite"
+  steps:
+    - name: Setup Service Account
+      uses: google-github-actions/auth@v2
+      with:
+        project_id: 'e2b-prod'
+        workload_identity_provider: 'projects/662973525644/locations/global/workloadIdentityPools/github-actions-e2b-prod-tests/providers/infra-integration-tests'
```
mishushakov, line 12: should we hardcode it like this?

.github/actions/host-init/action.yml
```diff
@@ -0,0 +1,44 @@
...
+
+    - name: Set up Cloud SDK
+      uses: google-github-actions/setup-gcloud@v2
+
+    - name: Setup Environment Variables
+      env:
+        GCP_REGION: "us-central1"
+        GCP_PROJECT_ID: "e2b-prod"
+        POSTGRES_CONNECTION_STRING: "postgresql://postgres:local@localhost:5432/mydatabase?sslmode=disable"
+        TEMPLATE_BUCKET_NAME: "e2b-tests-fc-templates"
```
mishushakov, line 22: maybe not hardcode env vars like this (other people might be using the repo)

.github/workflows/integration_tests.yml
```diff
@@ -0,0 +1,31 @@
...
+        uses: ./.github/actions/build-packages
+
+      - name: Initialize Host
+        uses: ./.github/actions/host-init
+
+      - name: Build Template
+        uses: ./.github/actions/build-sandbox-template
+
+      - name: Start Services
+        uses: ./.github/actions/start-services
```
mishushakov, line 28: I really like this, can we maybe put the 4 steps into one composite "prepare cluster"

.github/workflows/integration_tests.yml
```diff
@@ -0,0 +1,31 @@
...
+
+      - name: Run Integration Tests
+        uses: ./.github/actions/integration-tests
```
mishushakov: I think it would be fine to inline the action in this file, but up to you

.github/actions/host-init/action.yml
dobrac: We can move it to vars in GH, but I don't think other people using the repo should necessarily be able to run all workflows.

.github/actions/host-init/action.yml
dobrac: there is no issue with it being public, the access check is done between the Github, our repo, and GCP

otherwise it's probably the same as the next comment, not sure if we want to necessarily support others to run these actions, there is a little bit of setup needed anyway which might be hidden for users

.github/workflows/integration_tests.yml
dobrac: It would be then quite hidden in the action run, so it won't be clear if it failed on build, template build or anything else

++ the template build is a bit part of the integration testing

.github/workflows/integration_tests.yml
dobrac: I was thinking about it too, I will move it

mishushakov:
Can you also document somewhere which variables are required for the actions to run?

dobrac:
> Can you also document somewhere which variables are required for the actions to run?

All required variables are purposely defined in the actions. I've tried to introduce them only when necessary for that script. The only hidden one is your GCP login (because you can do it multiple ways). Do you want to just extract them somewhere else?

.github/actions/host-init/action.yml
jakubno: We have setup between github and GCP in terraform actually (`terraform/github`) and it's used in release pipeline. The string is saved in `secrets.workload_identity_provider`

packages/template-manager/internal/test/build.go
```diff
@@ -38,8 +38,8 @@ func Build(templateID, buildID string) {
 		TemplateFiles: storage.NewTemplateFiles(
 			templateID,
 			buildID,
-			"vmlinux-5.10.186",
-			"v1.7.0-dev_8bb88311",
+			"vmlinux-6.1.102",
+			"v1.10.1_1fcdaec",
```
jakubno: maybe use constants DefaultFirecrackerVersion and DefaultKernelVersion?

packages/orchestrator/internal/sandbox/network/storage_memory.go
```diff
@@ -0,0 +1,52 @@
...
+func (s *StorageMemory) Release(ips *Slot) error {
+	s.freeSlotsMu.Lock()
+	defer s.freeSlotsMu.Unlock()
+
+	s.freeSlots[ips.Idx] = false
+
+	return nil
+}
+
+func getMemoryKey(slotIdx int) string {
```
jakubno, line 49: do we need it here, it's client scoped

packages/orchestrator/internal/sandbox/fc/process.go
```diff
@@ -106,6 +106,11 @@ func NewProcess(
 		attribute.String("sandbox.cmd", fcStartScript.String()),
 	)
 
+	_, err = os.Stat(files.FirecrackerPath())
+	if err != nil {
+		return nil, fmt.Errorf("error stating firecracker binary: %w", err)
+	}
+
```
jakubno, line 113: does it make sense to do the same for kernel?

packages/integration-tests/tools/tools.go
jakubno, line 1: is this used?

packages/integration-tests/seed.go
```diff
@@ -0,0 +1,61 @@
...
+	defer db.Close()
+
+	// Execute the seed
+	tmpl, err := template.ParseFiles(*fileName)
+	if err != nil {
+		log.Fatalf("Failed to parse seed file: %v", err)
+	}
+
+	var parsed bytes.Buffer
+	err = tmpl.Execute(&parsed, SeedData{
```
jakubno: If you do some of them, you could probably do API_KEY and there's already env variable as well as it is used in github action

packages/integration-tests/internal/tests/api/sandbox_test.go
```diff
@@ -0,0 +1,62 @@
...
+	"net/http"
+	"testing"
+
+	"github.com/e2b-dev/infra/packages/integration-tests/internal/api"
+	"github.com/e2b-dev/infra/packages/integration-tests/internal/setup"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestSandboxCreate(t *testing.T) {
```
jakubno: those tests looks the same, isn't it better to do it in loop? I guess you probably want to run them in series, not in parallel right (if parallelization is enabled)?

packages/orchestrator/Makefile
```diff
@@ -35,6 +35,15 @@ build-local:
 build-debug:
 	CGO_ENABLED=1 GOOS=linux GOARCH=amd64 go build -race -gcflags=all="-N -l" -o bin/orchestrator .
 
+.PHONY: run-debug
+run-debug:
+	make build-debug
+	sudo -E GOTRACEBACK=crash \
+	GODEBUG=madvdontneed=1 \
```
jakubno, line 42: Any special  reason for `GODEBUG=madvdontneed=1` worth mentioning?

packages/api/internal/orchestrator/orchestrator.go
```diff
@@ -77,8 +79,14 @@ func New(
 
 	o.instanceCache = cache
 
-	if env.IsLocal() {
-		zap.L().Info("Skipping syncing sandboxes, running locally")
+  if env.IsLocal() {
+    zap.L().Info("Skipping syncing sandboxes, running locally")
+		// Add a local node for local development
```
jakubno:
```suggestion
		// Add a local node for local development, if there isn't any, it fails silently
```

.github/workflows/fc-test.yml
```diff
@@ -0,0 +1,195 @@
...
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v3
+
+      - name: Install Dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y build-essential libseccomp-dev libcap-dev qemu-kvm
+
+      - name: Download Firecracker
```
jakubno, line 20: Can we use `make copy-public-builds` or downloading the exact version from our bucket, so we test the exact same version as in prod

.github/workflows/fc-test.yml
```diff
@@ -0,0 +1,195 @@
...
+          for d in bin etc lib root sbin usr; do tar c "/$d" | tar x -C /my-rootfs; done && \
+          for dir in dev proc run sys var; do mkdir /my-rootfs/${dir}; done'
+          sudo umount /tmp/my-rootfs
+          
+          echo Check filesystem
+          e2fsck -n rootfs.ext4
+          echo Get filesystem information
+          tune2fs -l rootfs.ext4
+          echo Get filesystem size
+          du -h rootfs.ext4
```
jakubno, line 131: Can't you just extract the filesystem from some docker image, similarly as we do in template manager?

.github/workflows/fc-test.yml
```diff
@@ -0,0 +1,195 @@
...
+          ls -lh rootfs.ext4
+
+      - name: NBD Check
+        run: |
+          sudo modprobe nbd nbds_max=4096
+
+      - name: Create Firecracker Configuration
+        run: |
+          cat <<EOF > config.json
+          {
```
jakubno, line 142: some of these are different than in orchestrator, is that on purpose?

packages/integration-tests/tools/tools.go
dobrac: this is required (from documentation, and also our own usage) for older versions of golang (which we use) to have the oapi-codegen import, so probably required dependencies are properly requested

packages/integration-tests/seed.go
dobrac: Is that API Key used to access production data? I was trying to avoid having any production related keys in there. But I can add specifying it as env var?

packages/orchestrator/Makefile
dobrac: I've just copied this from the orchestrator, should I remove it?

.github/workflows/fc-test.yml
dobrac: this copies them to your configured GCP bucket (which is not set up at all), but I can download the FC version from the public bucket

dobrac: this fc-test is just for testing if firecracker works at all on that type of a machine, it has nothing to do with the integration tests

if you want, I can completely remove this workflow

packages/integration-tests/seed.go
dobrac: (I have added the API key seed from the TESTS_E2B_API_KEY variable, but I'm not using the API_KEY now)

packages/orchestrator/internal/sandbox/network/storage_memory.go
dobrac: removed the consul

.github/workflows/fc-test.yml
dobrac: (same as the https://github.com/e2b-dev/infra/pull/313#discussion_r1981853814)

.github/workflows/fc-test.yml
dobrac: (same as the https://github.com/e2b-dev/infra/pull/313#discussion_r1981853814)

.github/actions/host-init/action.yml
dobrac: we've discussed to add it to terraform eventually, but for now just add TODO comment

.github/actions/host-init/action.yml
```diff
@@ -0,0 +1,45 @@
...
+description: "Sets up the host environment, including authentication and dependencies."
+
+runs:
+  using: "composite"
+  steps:
+    - name: Setup Service Account
+      uses: google-github-actions/auth@v2
+      with:
+        project_id: 'e2b-prod'
+        # TODO: Move the identity provider construction to Terraform (like secrets.workload_identity_provider)
```
jakubno: Maybe change the todo, to get rid of dependance on gcp storage bucket and to be able to use local storage

packages/api/internal/orchestrator/orchestrator.go
```diff
@@ -77,8 +79,14 @@ func New(
 
 	o.instanceCache = cache
 
-	if env.IsLocal() {
-		zap.L().Info("Skipping syncing sandboxes, running locally")
+  if env.IsLocal() {
```
jakubno: formatting

packages/integration-tests/internal/tests/api/sandbox_test.go
```diff
@@ -0,0 +1,47 @@
...
+	var wg sync.WaitGroup
+	for i := 0; i < 2; i++ {
```
jakubno: Add short explanation why there's 2 🙏🏻

packages/integration-tests/seed.go
jakubno: I just wanted to be passed as variable

packages/orchestrator/internal/sandbox/network/storage_memory.go
```diff
@@ -0,0 +1,50 @@
...
+	return fmt.Sprintf("%d", slotIdx)
```
jakubno:
```suggestion
	return slotIdx
```

jakubno (approved):

packages/orchestrator/Makefile
jakubno: It's okay

PR #318, opened 2025-02-21
Add CI/CD to run infra unit tests on new PRs

mishushakov:
- Adds test scripts to Makefiles using go test
- Adds GitHub Action for running tests on each PR (on main) with test caches (using matrix)


.github/workflows/pr-tests.yml
```diff
@@ -0,0 +1,55 @@
...
+          - package: packages/envd
+            test_path: ./...
+          - package: packages/orchestrator
+            test_path: ./...
+          - package: packages/shared
+            test_path: ./pkg/...
+          - package: packages/template-manager
+            test_path: ./...
+    steps:
+      - uses: actions/checkout@v3
```
dobrac: there is already actions/checkout@v4

dobrac (changes requested):

dobrac:
One QoL improvement could be adding custom names to the matrix jobs, so the list won't look like this:

Screenshot 2025-02-21 at 10 40 04 AM

dobrac (approved):

ValentaTomas:
Looks good to me, we can merge.

PR #319, opened 2025-02-21
Split out and test post processing in template build

r33drichards:
Looking to test `createRootFS`, so splitting out this routine into its own stuct to test. In order to test this locally on mac need to set build flags for linux. also refactors some behavior in `shared` to be able to run tests

### DIFF
--- a/packages/shared/pkg/storage/gcs/bucket.go
+++ b/packages/shared/pkg/storage/gcs/bucket.go
@@ -1,6 +1,9 @@
 package gcs
 
 import (
+	"context"
+	"sync"
+
 	"cloud.google.com/go/storage"
 
 	"github.com/e2b-dev/infra/packages/shared/pkg/utils"
@@ -8,12 +11,21 @@ import (
 
 type BucketHandle = storage.BucketHandle
 
+var clientOnce sync.Once
+var client *storage.Client
+
+var getClient = sync.OnceValue(func() *storage.Client {
+	return utils.Must(newClient(context.Background()))
+})
+
 func newBucket(bucket string) *BucketHandle {
-	return client.Bucket(bucket)
+	return getClient().Bucket(bucket)
 }
 
-var (
-	templateBucketName = utils.RequiredEnv("TEMPLATE_BUCKET_NAME", "bucket for storing template files")
+func getTemplateBucketName() string {
+	return utils.RequiredEnv("TEMPLATE_BUCKET_NAME", "bucket for storing template files")
+}
 
-	TemplateBucket = newBucket(templateBucketName)
-)
+func GetTemplateBucket() *BucketHandle {
+	return newBucket(getTemplateBucketName())
+}
--- a/packages/shared/pkg/storage/gcs/bucket.go
+++ b/packages/shared/pkg/storage/gcs/bucket.go
@@ -11,9 +11,6 @@ import (
 
 type BucketHandle = storage.BucketHandle
 
-var clientOnce sync.Once
-var client *storage.Client
-
 var getClient = sync.OnceValue(func() *storage.Client {
 	return utils.Must(newClient(context.Background()))
 })
--- a/packages/template-manager/internal/build/network_linux.go
+++ b/packages/template-manager/internal/build/network_linux.go
@@ -118,7 +118,7 @@ func (n *FCNetwork) setup(ctx context.Context, tracer trace.Tracer) error {
 	tapAttrs.Namespace = ns
 
 	tap := &netlink.Tuntap{
-		Mode:      netlink.TUNTAP_MODE_TUN,
+		Mode:      netlink.TUNTAP_MODE_TAP,
 		LinkAttrs: tapAttrs,
 	}
 
--- a/packages/template-manager/internal/build/network_linux.go
+++ b/packages/template-manager/internal/build/network_linux.go
@@ -1,3 +1,6 @@
+//go:build linux
+// +build linux
+
 package build
 
 import (
@@ -12,6 +15,7 @@ import (
 
 	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
 )
+
 
 const (
 	fcTapAddress        = "169.254.0.22"
@@ -112,8 +116,9 @@ func (n *FCNetwork) setup(ctx context.Context, tracer trace.Tracer) error {
 	tapAttrs := netlink.NewLinkAttrs()
 	tapAttrs.Name = fcTapName
 	tapAttrs.Namespace = ns
+
 	tap := &netlink.Tuntap{
-		Mode:      netlink.TUNTAP_MODE_TAP,
+		Mode:      netlink.TUNTAP_MODE_TUN,
 		LinkAttrs: tapAttrs,
 	}
 
--- a/packages/template-manager/internal/build/network_other.go
+++ b/packages/template-manager/internal/build/network_other.go
@@ -1,0 +1,24 @@
+//go:build !linux
+// +build !linux
+
+package build
+
+import (
+	"context"
+	"fmt"
+
+	"go.opentelemetry.io/otel/trace"
+)
+
+type FCNetwork struct {
+	namespaceID string
+}
+
+// Cleanup is a no-op for non-Linux systems
+func (n *FCNetwork) Cleanup(ctx context.Context, tracer trace.Tracer) {
+}
+
+// NewFCNetwork returns an error
+func NewFCNetwork(ctx context.Context, tracer trace.Tracer, env *Env) (*FCNetwork, error) {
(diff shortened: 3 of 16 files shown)


packages/shared/pkg/storage/gcs/bucket.go
```diff
@@ -1,19 +1,29 @@
...
 
 	"github.com/e2b-dev/infra/packages/shared/pkg/utils"
 )
 
 type BucketHandle = storage.BucketHandle
 
+var client *storage.Client
+
 func newBucket(bucket string) *BucketHandle {
+	if client == nil {
```
dobrac: This is not thread-safe. Better way of creating singleton would be the use of sync.OnceValue

packages/template-manager/internal/build/network_linux.go
```diff
@@ -6,13 +9,21 @@ import (
 	"net"
 	"runtime"
 
+	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
```
dobrac: nit: wrong import group

packages/template-manager/internal/build/network_linux.go
```diff
@@ -112,9 +123,13 @@ func (n *FCNetwork) setup(ctx context.Context, tracer trace.Tracer) error {
 	tapAttrs := netlink.NewLinkAttrs()
 	tapAttrs.Name = fcTapName
 	tapAttrs.Namespace = ns
-	tap := &netlink.Tuntap{
-		Mode:      netlink.TUNTAP_MODE_TAP,
-		LinkAttrs: tapAttrs,
+
+	tap := getTap(tapAttrs)
+	if tap == nil {
```
dobrac: when is the tap nil?

packages/template-manager/internal/build/rootfs.go
```diff
@@ -163,59 +163,61 @@ func (r *Rootfs) dockerTag() string {
...
-
-	var err error
-	postprocessingFinished := make(chan error)
-	defer func() { postprocessingFinished <- err }()
-
-	go func() {
-		now := time.Now()
-		for {
-			msg := []byte(fmt.Sprintf("Postprocessing (%s)       \r", time.Since(now).Round(time.Second)))
+type PostProcesser struct {
```
dobrac: nit: PostProcessor (and all related methods, etc)

packages/template-manager/internal/build/network_other.go
```diff
@@ -0,0 +1,29 @@
...
+	"context"
+	"fmt"
+
+	"go.opentelemetry.io/otel/trace"
+)
+
+const (
+	fcTapAddress = "169.254.0.22"
+	fcTapName    = "30"
+)
```
dobrac: these constants don't seem to be required

packages/template-manager/internal/build/rootfs.go
```diff
@@ -163,59 +163,61 @@ func (r *Rootfs) dockerTag() string {
...
+			return
+		case <-time.After(100 * time.Millisecond):
+			p.writer.Write(msg)
 		}
-	}()
+	}
+
+}
+
+func (p *PostProcesser) sendError(err error) {
```
dobrac: I would rename this method, as it is required to be called to finish the post-processing

packages/template-manager/internal/build/rootfs_test.go
```diff
@@ -0,0 +1,70 @@
...
+				testErr:       nil,
+				shouldContain: "Postprocessing finished.  ",
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+
+			tw := &testWriter{}
+			ctx := context.Background()
```
dobrac: might be better to create a new context for each test run and cancel it after the test finishes

dobrac (changes requested):

packages/template-manager/internal/build/network_linux.go
r33drichards: what does that mean? not familiar with import groups

packages/template-manager/internal/build/network_linux.go
r33drichards: I'm not sure, this is a holdout from an alternate implementation, the error message below is inaccurate but it think its best to check if a pointer is nil regardless

packages/template-manager/internal/build/network_other.go
r33drichards: will double check but i believe i had compiler errors from this

packages/template-manager/internal/build/rootfs.go
r33drichards: what would you rename it to be?

packages/template-manager/internal/build/rootfs_test.go
r33drichards: this does create a new context for each run but does not cancel after test run, can add this

packages/template-manager/internal/build/network_linux.go
dobrac: I think you should return an error instead from the getTap function if the return value would be nil

packages/template-manager/internal/build/network_linux.go
dobrac: just putting the import together with other ones from e2b repository and a new line between the other ones (like "net", "runtime")

packages/template-manager/internal/build/rootfs.go
dobrac: How about just `stop(err error)`?

packages/template-manager/internal/build/rootfs_test.go
r33drichards: how about `context.TODO()` since its not being tested anyways

r33drichards: considering the behavior of context is not being tested, and the docs for each for reference 

[context.Background()](https://pkg.go.dev/context#Background):

`func Background() Context`
Background returns a non-nil, empty Context. It is never canceled, has no values, and has no deadline. It is typically used by the main function, initialization, and tests, and as the top-level Context for incoming requests.

[context.TODO()](https://pkg.go.dev/context#TODO):

`func TODO() Context`
TODO returns a non-nil, empty Context. Code should use context.TODO when it's unclear which Context to use or it is not yet available (because the surrounding function has not yet been extended to accept a Context parameter).

dobrac: Oh, I see. I was thinking using the context as a proper cleanup (if required by the body func), otherwise I'm for the TODO one

packages/template-manager/internal/build/network_linux.go
```diff
@@ -112,8 +116,9 @@ func (n *FCNetwork) setup(ctx context.Context, tracer trace.Tracer) error {
 	tapAttrs := netlink.NewLinkAttrs()
 	tapAttrs.Name = fcTapName
 	tapAttrs.Namespace = ns
+
 	tap := &netlink.Tuntap{
-		Mode:      netlink.TUNTAP_MODE_TAP,
+		Mode:      netlink.TUNTAP_MODE_TUN,
```
dobrac: Is this change desired?

r33drichards:
```suggestion
		Mode:      netlink.TUNTAP_MODE_TAP,
```

packages/shared/pkg/storage/gcs/bucket.go
```diff
@@ -1,19 +1,31 @@
...
 	"cloud.google.com/go/storage"
 
 	"github.com/e2b-dev/infra/packages/shared/pkg/utils"
 )
 
 type BucketHandle = storage.BucketHandle
 
+var clientOnce sync.Once
+var client *storage.Client
+
```
dobrac:
```suggestion
```

r33drichards:
tested with 
```
➜  template-manager git:(post-processing) ✗ go test ./...                                                                                                        ~/infra/packages/template-manager
?       github.com/e2b-dev/infra/packages/template-manager      [no test files]
?       github.com/e2b-dev/infra/packages/template-manager/internal/build/writer        [no test files]
?       github.com/e2b-dev/infra/packages/template-manager/internal/constants   [no test files]
?       github.com/e2b-dev/infra/packages/template-manager/internal/server      [no test files]
?       github.com/e2b-dev/infra/packages/template-manager/internal/template    [no test files]
?       github.com/e2b-dev/infra/packages/template-manager/internal/test        [no test files]
ok      github.com/e2b-dev/infra/packages/template-manager/internal/build       0.324s
➜  template-manager git:(post-processing) ✗     
```

r33drichards:
going to build a template with this change and make sure output looks the same then merging

r33drichards:
Screenshot 2025-02-25 at 11 03 53 AM

PR #320, opened 2025-02-21
Provision gcp logs data source for grafana with terraform

r33drichards:
create connection from gcp logs to grafana using terraform

### DIFF
--- a/terraform/grafana/main.tf
+++ b/terraform/grafana/main.tf
@@ -114,3 +114,89 @@ resource "google_secret_manager_secret_version" "grafana_logs_user" {
   secret      = "projects/${var.gcp_project_id}/secrets/${var.prefix}grafana-logs-user"
   secret_data = grafana_cloud_stack.e2b_stack.logs_user_id
 }
+
+# Enable Cloud Logging API
+resource "google_project_service" "logging_api" {
+  project = var.gcp_project_id
+  service = "logging.googleapis.com"
+}
+
+# Enable Cloud Resource Manager API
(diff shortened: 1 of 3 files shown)


terraform/grafana/main.tf
```diff
@@ -114,3 +114,89 @@ resource "google_secret_manager_secret_version" "grafana_logs_user" {
...
+
+}
+
+resource "grafana_data_source" "gcloud_logs" {
+  provider = grafana.datasource
+
+  name = "gcloud-logs"
+  type = "googlecloud-logging-datasource"
+  json_data_encoded = jsonencode({
+    authenticationType = "jwt"
```
0div, line 192: just making sure: this JWT is setup via the service account?

r33drichards:
> LGTM, although just wanna make sure this fails gracefully for those using clusters w/o grafana cloud set up?

This wont run unless you cd into `terraform/grafana` and apply there so shouldn't have to worry about this case

terraform/grafana/main.tf
r33drichards: correct, the jwt is created by the service account 

```
    clientEmail        = google_service_account.grafana_logging.email
    defaultProject     = var.gcp_project_id
    tokenUri           = "https://oauth2.googleapis.com/token"
  })

  secure_json_data_encoded = jsonencode({
    privateKey = google_service_account_key.grafana_logging_key.private_key
  })
  ```
  these fields in the ui get populated from the decoded jwt if you paste it in the ui

PR #321, opened 2025-02-22
Centralize and structure logging across all infra packages 

0div:
- Create single logger shared lib with HTTP exporter using `zap`
- All packages in infra have a global zap logger initialized with a http exporter to logs-collector (and one with more initial fields and helper methos for sandbox logger)


packages/shared/pkg/logger/exporter.go
```diff
@@ -1,36 +1,43 @@
...
-	address  string
+	ctx       context.Context
+	client    http.Client
+	logQueue  chan []byte
+	debugLogs bool
+	address   string
 }
 
-func NewHTTPLogsExporter(ctx context.Context, address string) *HTTPExporter {
+var logsExporterMU sync.Mutex
```
dobrac: what is the role of this mutex?

packages/orchestrator/internal/sandbox/build/cache.go
```diff
@@ -31,7 +32,7 @@ func NewDiffStore(bucket *gcs.BucketHandle, ctx context.Context) (*DiffStore, er
 
 		err := buildData.Close()
 		if err != nil {
-			fmt.Printf("[build data cache]: failed to cleanup build data for item %s: %v\n", item.Key(), err)
+			zap.L().Warn("failed to cleanup build data for item", zap.String("item_key", item.Key()), zap.Error(err))
```
dobrac: nit: the "[build data cache]:" is lost here

packages/orchestrator/internal/sandbox/rootfs/cow.go
```diff
@@ -126,7 +126,7 @@ func (o *CowDevice) Close() error {
 		err := nbd.Pool.ReleaseDevice(slot)
 		if errors.Is(err, nbd.ErrDeviceInUse{}) {
 			if counter%100 == 0 {
-				log.Printf("[%dth try] error releasing overlay device: %v\n", counter, err)
+				zap.L().Info("error releasing overlay device", zap.Int("counter", counter), zap.Error(err))
```
dobrac: nit: it's probably not "counter", but something like "attempt"

packages/orchestrator/internal/server/main.go
```diff
@@ -41,11 +41,11 @@ func New() (*grpc.Server, error) {
 
 	dnsServer := dns.New()
 	go func() {
-		log.Printf("Starting DNS server")
+		zap.L().Info("Starting DNS server")
 
 		err := dnsServer.Start("127.0.0.4", 53)
 		if err != nil {
-			log.Fatalf("Failed running DNS server: %s\n", err.Error())
+			zap.L().Error("Failed running DNS server", zap.Error(err))
```
dobrac: changed Fatal to Error, Fatal will exit the process

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -38,13 +38,17 @@ func (s *server) Create(ctx context.Context, req *orchestrator.SandboxCreateRequ
...
-		req.Sandbox.TeamId,
-		req.Sandbox.Vcpu,
-		req.Sandbox.RamMb,
-		false,
+	sbxLogger := sbxlogger.NewSandboxLogger(
+		childCtx,
+		sbxlogger.SandboxLoggerConfig{
+			ServiceName:      ServiceName,
+			IsInternal:       true,
+			IsDevelopment:    true,
```
dobrac: is development?

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -38,13 +38,17 @@ func (s *server) Create(ctx context.Context, req *orchestrator.SandboxCreateRequ
...
+			SandboxID:        req.Sandbox.SandboxId,
+			TemplateID:       req.Sandbox.TemplateId,
+			TeamID:           req.Sandbox.TeamId,
+			CollectorAddress: os.Getenv("LOGS_COLLECTOR_ADDRESS"),
```
dobrac: second place of the LOGS_COLLECTOR_ADDRESS env var usage

terraform/grafana/main.tf
```diff
@@ -82,8 +82,8 @@ resource "google_secret_manager_secret_version" "grafana_logs_url" {
 resource "grafana_cloud_access_policy" "logs_collector" {
   provider     = grafana.cloud
   region       = var.gcp_to_grafana_regions[var.gcp_region]
-  name         = "logs-collector-${var.gcp_project_id}"
-  display_name = "Logs Collector for ${var.gcp_project_id}"
+  name         = "logs-collector-workaround"
```
dobrac: workaround?:)

packages/orchestrator/internal/server/sandboxes.go
0div: it's a zap config 

> Development puts the logger in development mode, which changes the behavior of DPanicLevel and takes stacktraces more liberally.

packages/orchestrator/internal/sandbox/build/cache.go
0div: i dislike this type of logging, i added the missing keyword `cache` in the sentence. The rest of the context is handled with the `zap.Caller` context

packages/shared/pkg/logger/exporter.go
0div: legacy from `logs`

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -144,7 +145,7 @@ func NewSandbox(
 	go func() {
 		runErr := rootfsOverlay.Start(childCtx)
 		if runErr != nil {
-			fmt.Fprintf(os.Stderr, "[sandbox %s]: rootfs overlay error: %v\n", config.SandboxId, runErr)
+			sbxLogger.Warn("rootfs overlay error", zap.Error(runErr))
```
dobrac: I think this should stay an error, also, the sandboxId should be included

packages/shared/pkg/logger/sandbox/logger.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	if alwaysReport {
+		if ok {
+			sl.logger.Info(
+				"Control sandbox healthcheck was successful",
+				zap.Bool("healthcheck", ok))
+		} else {
+			sl.logger.Error("",
+				zap.Error(fmt.Errorf("Sandbox healthcheck failed")),
```
dobrac: what is the purpose of the error beying "" and then its sandbox healthcheck failed in zap.Error?

packages/template-manager/internal/server/main.go
```diff
@@ -36,9 +34,9 @@ type serverStore struct {
 
 func New(logger *zap.Logger) *grpc.Server {
 	ctx := context.Background()
-	log.Println("Initializing template manager")
+	logger.Info("Initializing template manager")
 
-	opts := []grpc_zap.Option{logging.WithoutHealthCheck()}
+	opts := []grpc_zap.Option{grpc_zap.WithLevels(grpc_zap.DefaultCodeToLevel)}
```
dobrac: I think you should also exclude the health check (to reduce spam)

packages/api/main.go
```diff
@@ -42,6 +43,8 @@ const (
 	defaultPort          = 80
 )
 
+var logsCollectorAddress = env.GetEnv("LOGS_COLLECTOR_ADDRESS", "")
```
dobrac: in some places the env variable is like this, in some places it's inline

I would unite the approach

packages/api/main.go
```diff
@@ -148,7 +151,17 @@ func main() {
...
 
-	log.Println("Starting API service...")
+	logger := zap.Must(logger.NewLogger(ctx, logger.LoggerConfig{
+		ServiceName:      serviceName,
+		IsInternal:       true,
+		IsDevelopment:    env.IsLocal(),
+		IsDebug:          env.IsDebug(),
+		CollectorAddress: logsCollectorAddress,
+	}))
+	defer logger.Sync()
```
dobrac, line 175: is the sync important to call? just asking because rn we don't have proper shutdown of most services

0div: it's just for flushing queued logs (if possible)

packages/orchestrator/main.go
```diff
@@ -5,7 +5,19 @@ import (
...
+	"log"
+	"net"
+=======
+	"log"
+	"os"
+	"os/signal"
+	"sync"
+	"sync/atomic"
+	"syscall"
+>>>>>>> fc4de370eb2036db95799a3dc8d100c0b9f80650
```
dobrac: ups (whole file)

packages/api/internal/handlers/sandbox.go
```diff
@@ -24,13 +26,35 @@ func (a *APIStore) startSandbox(
...
+
+		return nil, errMsg
+	}
+
+	defer sandboxStartRequestLimit.Release(1)
+	telemetry.ReportEvent(ctx, "create sandbox parallel limit semaphore slot acquired")
+
+	rateSpan.End()
+	telemetry.ReportEvent(ctx, "Reserved team sandbox slot")
+
```
dobrac: this should be removed in main

packages/api/internal/orchestrator/list_instances.go
```diff
@@ -58,32 +62,40 @@ func (o *Orchestrator) getSandboxes(ctx context.Context, node *nNode.NodeInfo) (
...
+			ServiceName:      ServiceName,
+			IsInternal:       true,
+			IsDevelopment:    true,
+			SandboxID:        config.SandboxId,
+			TemplateID:       config.TemplateId,
+			TeamID:           teamID.String(),
+			CollectorAddress: os.Getenv("LOGS_COLLECTOR_ADDRESS"),
+		})
+
+		sandboxesInfo = append(sandboxesInfo, &instance.InstanceInfo{
```
dobrac: this change reverts main, there should be used the new constructor

packages/client-proxy/main.go
```diff
@@ -139,7 +139,13 @@ func main() {
 	signalCtx, sigCancel := signal.NotifyContext(ctx, syscall.SIGTERM, syscall.SIGINT)
 	defer sigCancel()
 
-	logger, err := logging.New(env.IsLocal())
+	logger, err := logger.NewLogger(ctx, logger.LoggerConfig{
+		ServiceName:      "client-proxy",
```
dobrac: extract to a variable as in other places?

packages/shared/pkg/logger/sandbox/logger.go
0div: the first param is message in zap, if u only want structured fields u have to put a nil string

packages/api/internal/cache/instance/instance.go
```diff
@@ -173,7 +169,9 @@ func NewCache(
 	cache.OnEviction(func(ctx context.Context, instanceInfo *InstanceInfo) {
 		err := deleteInstance(instanceInfo)
 		if err != nil {
-			logger.Errorf("Error deleting instance (%v)\n: %v", err)
+			zap.L().Error("Error inserting instance", zap.Error(err))
+
+			return
```
dobrac: this return changes logic, I would remove it to keep it the same

packages/orchestrator/internal/sandbox/build/build.go
```diff
@@ -47,8 +48,8 @@ func (b *File) ReadAt(p []byte, off int64) (n int, err error) {
 		readLength := min(mappedLength, remainingReadLength)
 
 		if readLength <= 0 {
-			fmt.Printf(
-				"(%d bytes left to read, off %d) reading %d bytes from %+v/%+v: [%d:] -> [%d:%d] <> %d (mapped length: %d, remaining read length: %d)\n",
+			zap.L().Debug(fmt.Sprintf(
```
dobrac: this is probably error, not debug

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -178,7 +182,7 @@ func (s *server) Delete(ctx context.Context, in *orchestrator.SandboxDeleteReque
 
 	err := sbx.Stop()
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "error stopping sandbox '%s': %v\n", in.SandboxId, err)
+		sbx.Logger.Warn("error stopping sandbox", zap.String("sandbox_id", in.SandboxId), zap.Error(err))
```
dobrac: isn't it weird to have this (and the ones below) as warn if they say error in the message?

packages/shared/pkg/env/env.go
```diff
@@ -12,6 +12,10 @@ func IsLocal() bool {
 	return environment == "local"
 }
 
+func IsDebug() bool {
+	return GetEnv("E2B_DEBUG", "false") == "true"
```
dobrac, line 16: meaning of this env var might have different meaning than you want 

```
You can use E2B SDKs with env var `E2B_DEBUG=true` or with a debug parameter set to true when creating or connecting to a sandbox to connect to the envd started with this command locally.
```

dobrac (approved):

dobrac:
image

packages/api/internal/cache/instance/instance.go
```diff
@@ -173,7 +169,7 @@ func NewCache(
 	cache.OnEviction(func(ctx context.Context, instanceInfo *InstanceInfo) {
 		err := deleteInstance(instanceInfo)
 		if err != nil {
-			logger.Errorf("Error deleting instance (%v)\n: %v", err)
+			zap.L().Error("Error inserting instance", zap.Error(err))
```
jakubno:
```suggestion
			zap.L().Error("Error deleting instance", zap.Error(err))
```

packages/nomad/logs-collector.hcl
```diff
@@ -133,10 +137,6 @@ auth.password = "${grafana_api_key}"
 
 [sinks.grafana.labels]
 source = "logs-collector"
-service = "{{ service }}"
```
jakubno: service should be always present right?

packages/shared/pkg/logger/exporter.go
```diff
@@ -1,36 +1,43 @@
...
+
+func NewHTTPLogsExporter(ctx context.Context, address string, debugLogs bool) LogsExporter {
+	logsExporterMU.Lock()
+	defer logsExporterMU.Unlock()
```
jakubno: why's there mutex? you can still create more exporters

packages/shared/pkg/logger/sandbox/logger.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/e2b-dev/infra/packages/shared/pkg/logger"
+	"go.uber.org/zap"
+)
+
+type SandboxLogger struct {
+	logger                *zap.Logger
+	healthCheckWasFailing atomic.Bool
+}
+
+type SandboxLoggerConfig struct {
```
jakubno, line 12: thinking if we could have service name as globals instead of passing it, also for collector address and is development

packages/shared/pkg/logger/sandbox/logger.go
```diff
@@ -0,0 +1,117 @@
...
+
+	return
+}
+
+func (sl *SandboxLogger) Healthcheck(ok bool, alwaysReport bool) {
+	if !ok && !sl.healthCheckWasFailing.Load() {
+		sl.healthCheckWasFailing.Store(true)
+
+		sl.logger.Error("",
+			zap.Error(fmt.Errorf("Sandbox healthcheck started failing")),
```
jakubno: What's advantage of having it as error? I think this will be more readable
```suggestion
		sl.logger.Error("Sandbox healthcheck started failing",
```

packages/shared/pkg/logger/sandbox/logger.go
```diff
@@ -0,0 +1,117 @@
...
+			zap.Bool("healthcheck", ok))
+
+		return
+	}
+	if ok && sl.healthCheckWasFailing.Load() {
+		sl.healthCheckWasFailing.Store(false)
+
+		sl.logger.Info("",
+			zap.Error(fmt.Errorf("Sandbox healthcheck recovered")),
+			zap.Bool("healthcheck", ok))
```
jakubno: This isn't even error
```suggestion
		sl.logger.Info("Sandbox healthcheck recovered",
			zap.Bool("healthcheck", ok))
```

packages/shared/pkg/logger/sandbox/logger.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	if alwaysReport {
+		if ok {
+			sl.logger.Info(
+				"Control sandbox healthcheck was successful",
+				zap.Bool("healthcheck", ok))
+		} else {
+			sl.logger.Error("Control sandbox healthcheck was unsuccessful",
+				zap.Error(fmt.Errorf("Sandbox healthcheck failed")),
```
jakubno: Imho no need for that
```suggestion
```

packages/shared/pkg/logger/logger.go
```diff
@@ -0,0 +1,89 @@
...
+	exporters := []io.Writer{}
+	if loggerConfig.CollectorAddress != "" {
+		exporter := NewHTTPLogsExporter(
+			ctx, loggerConfig.CollectorAddress, loggerConfig.IsDevelopment)
+		exporters = append(exporters, exporter)
+	}
+
+	config := zap.Config{
+		Level:             level,
+		Development:       loggerConfig.IsDevelopment,
```
jakubno: I think this could be set by based on `IsProduction()`

packages/nomad/logs-collector.hcl
```diff
@@ -49,7 +49,8 @@ job "logs-collector" {
       env {
         VECTOR_CONFIG          = "local/vector.toml"
         VECTOR_REQUIRE_HEALTHY = "true"
-        VECTOR_LOG             = "warn"
+        VECTOR_LOG             = "info"
+        VECTOR_INTERNAL_LOG_RATE_LIMIT = "1000"
```
jakubno: I would go with 600 (10 minutes),
```suggestion
        # 10 minutes
        VECTOR_INTERNAL_LOG_RATE_LIMIT = "600"   
```

packages/api/internal/orchestrator/list_instances.go
```diff
@@ -13,9 +14,11 @@ import (
 	"github.com/e2b-dev/infra/packages/api/internal/cache/instance"
 	nNode "github.com/e2b-dev/infra/packages/api/internal/node"
 	"github.com/e2b-dev/infra/packages/api/internal/utils"
-	"github.com/e2b-dev/infra/packages/shared/pkg/logs"
+	sbxlogger "github.com/e2b-dev/infra/packages/shared/pkg/logger/sandbox"
 )
 
+const ServiceName = "orchestration-api"
```
jakubno: I think this could be global/env var. Also this probably doesn't have to capitalized

packages/api/internal/handlers/sandbox_create.go
```diff
@@ -15,11 +16,14 @@ import (
...
 	"github.com/e2b-dev/infra/packages/shared/pkg/id"
-	"github.com/e2b-dev/infra/packages/shared/pkg/logs"
+	sbxlogger "github.com/e2b-dev/infra/packages/shared/pkg/logger/sandbox"
 	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
 )
 
-const InstanceIDPrefix = "i"
+const (
+	InstanceIDPrefix = "i"
+	ServiceName      = "orchestration-api"
```
jakubno: As above

packages/nomad/logs-collector.hcl
```diff
@@ -84,10 +85,7 @@ inputs = ["envd"]
 source = """
 del(."_path")
 .service = "envd"
-.sandboxID = .instanceID
```
jakubno: it's dropping all logs, because the label is missing. Whenever you change these configs, test it. It's annoying, it's easy to break, but that's why we have to make sure, it really works. Feel free to come up with some testing mechanism

jakubno: Also there are logs in logs collector with missing teamID label

jakubno (changes requested):
Thanks for this! 

There are some issues.
I don't know why, even though I reverted logs collector config back. There are no logs from API. I didn't have any more errors in logs collector or loki after that, but I don't get any logs or metrics via CLI.

packages/nomad/logs-collector.hcl
0div: yeah i think i committed this accidentally, was testing something

packages/shared/pkg/logger/exporter.go
0div: legacy stuff, lemme remove

packages/shared/pkg/logger/sandbox/logger.go
0div: this is a shared pkg so i think i need to pass it as arg

0div:
> Thanks for this!
> 
> There are some issues. I don't know why, even though I reverted logs collector config back. There are no logs from API. I didn't have any more errors in logs collector or loki after that, but I don't get any logs or metrics via CLI.

Yeah i'm exactly at that stage, no errors but no logs. I'm gonna try to run this locally to get more visibility.

packages/api/internal/orchestrator/list_instances.go
0div: Agree, would be better to set as an env var

packages/nomad/logs-collector.hcl
0div: Now that we're including logs from everywhere, there will be missing labels for sandbox ID. I've tried a combination of label transforms with fallback default values, to no avail so far :/

jakubno: Is the PR ready or not then?

packages/shared/pkg/logger/sandbox/logger.go
jakubno: I though you would set some global variable in init method, something like in stripe you can set global api key and then you don't care about it, no need for that now

packages/nomad/logs-collector.hcl
0div: Gonna test some changes first, if all goes well, we merge.

dobrac:
It should be ready for a review now 🙏 

Notes:
- Regular logs are now sent to the OTEL logs exporter (and effectively Loki - Grafana Cloud)
- Sandbox logs are sent either to the OTEL (internal: true) or to the user Loki (internal: false)
    - There are now also two new global sandbox loggers (internal, external), that can be accessed using sbxlogger.I(c) or sbxlogger.E(c)
    - This should avoid passing sandbox logger instance between many components, plus it should allow to easily change between internal and external log messages
    - Minimize HTTP Exporter buffered processing to only one exporter (and not one per sandbox)
    - Reduce requirement to initialize the sandbox logger many time across the codebase (e.g. each http handle in the API)
- Part of the PR is also a new HTTPExporter implementation, which is buffered (from zapcore) and logs sending is processed by a goroutine

packages/shared/pkg/logger/logger.go
```diff
@@ -0,0 +1,86 @@
...
+	"go.opentelemetry.io/contrib/bridges/otelzap"
+	"go.opentelemetry.io/otel/log/global"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+	"os"
+)
+
+type LoggerConfig struct {
+	ServiceName   string
+	IsInternal    bool
```
jakubno: Can we add comments explaining this 🙏🏻

packages/shared/pkg/logger/sandbox/global.go
jakubno, line 1: does it make sense to name it sandbox logger? Is there anything blocking it to be used for e.g. template build logs?

dobrac: it's kinda the way it was, do you have any better name? For template build, I don't think we necessary have the sandboxID, so some generalization would be required

Right now it requires to specify vars:
- sandboxID
- templateID
- teamID

packages/api/internal/handlers/sandbox_create.go
```diff
@@ -15,11 +15,14 @@ import (
...
```
jakubno: This isn't used

packages/api/internal/orchestrator/cache.go
```diff
@@ -30,12 +29,17 @@ func (o *Orchestrator) GetSandbox(sandboxID string) (*instance.InstanceInfo, err
 
 // keepInSync the cache with the actual instances in Orchestrator to handle instances that died.
 func (o *Orchestrator) keepInSync(ctx context.Context, instanceCache *instance.InstanceCache) {
+	firstSync := make(chan struct{}, 1)
```
jakubno: I think this isn't relevant to logging

packages/api/internal/orchestrator/list_instances.go
jakubno: Also not used

packages/api/main.go
```diff
@@ -64,7 +69,7 @@ func NewGinServer(ctx context.Context, apiStore *handlers.APIStore, swagger *ope
 			"/templates/:templateID/builds/:buildID/status",
 		),
 		customMiddleware.IncludeRoutes(metricsMiddleware.Middleware(serviceName), "/sandboxes"),
-		customMiddleware.ExcludeRoutes(gin.LoggerWithWriter(gin.DefaultWriter),
+		customMiddleware.ExcludeRoutes(ginzap.Ginzap(zap.L(), time.RFC3339, true),
```
jakubno: Isn't this resolved before the our logger is set to global?

packages/api/main.go
```diff
@@ -153,22 +158,59 @@ func main() {
...
-	if debug != "true" {
-		gin.SetMode(gin.ReleaseMode)
-	}
-
 	swagger, err := api.GetSwagger()
 	if err != nil {
 		// this will call os.Exit: defers won't run, but none
 		// need to yet. Change this if this is called later.
-		log.Fatalf("Error loading swagger spec:\n%v", err)
+		fmt.Fprintf(os.Stderr, "error loading swagger spec: %v\n", err)
```
jakubno: is there reason to prefer fmt.Fprint? I prefer log, because of timestamp, also fmt.Fprint is returning error and IDE is complaining

packages/api/main.go
```diff
@@ -153,22 +158,59 @@ func main() {
...
+		return
 	}
 
 	var cleanupFns []func(context.Context) error
+
+	if !env.IsLocal() {
```
jakubno, line 164: why was this moved? there was a reason why `defer cleanup()` was before this block, if we add some code, which can error, cleanup won't be called for this

packages/api/main.go
```diff
@@ -153,22 +158,59 @@ func main() {
...
+		cleanupFns = append(cleanupFns, telemetry.InitOTLPExporter(ctx, serviceName, swagger.Info.Version))
```
jakubno: can we use commit instead of swagger version? swagger version never changes

packages/client-proxy/main.go
```diff
@@ -36,26 +37,18 @@ const (
...
+var (
+	client = new(dns.Client)
+)
 
-func proxyHandler(logger *zap.SugaredLogger) func(w http.ResponseWriter, r *http.Request) {
-	activeConnections, err := meters.GetUpDownCounter(meters.ActiveConnectionsCounterMeterName)
-	if err != nil {
-		logger.Error("failed to create active connections counter", zap.Error(err))
-	}
+func proxyHandler() func(w http.ResponseWriter, r *http.Request) {
```
jakubno, line 46: looks like incorrectly solved merge conflict, but feel free to ignore, I will have to fix few things for this anyway - otel is closed too early

packages/cluster/build-cluster/main.tf
```diff
@@ -65,7 +65,6 @@ resource "google_compute_instance_template" "build" {
 
   instance_description = var.cluster_description
   machine_type         = var.machine_type
-  min_cpu_platform     = "Intel Skylake"
```
jakubno: Please return this

packages/envd/internal/host/sync.go
```diff
@@ -16,12 +16,10 @@ func updateClock() error {
...
 
-	cmd := exec.CommandContext(ctx, "/bin/bash", "-l", "-c", "date -s @$(/usr/sbin/phc_ctl /dev/ptp0 get | cut -d' ' -f5)")
-
-	// Capture both stdout and stderr
-	output, err := cmd.CombinedOutput()
+	// The chronyc -a makestep is not immediately stepping the clock
+	err := exec.CommandContext(ctx, "/bin/bash", "-l", "-c", "date -s @$(/usr/sbin/phc_ctl /dev/ptp0 get | cut -d' ' -f5)").Run()
 	if err != nil {
-		return fmt.Errorf("failed to update clock: %w\nCommand output: %s", err, string(output))
+		return fmt.Errorf("failed to update clock: %w", err)
```
jakubno: also return this

packages/nomad/otel-collector.hcl
jakubno, line 1: return this

packages/orchestrator/internal/sandbox/checks.go
```diff
@@ -50,7 +53,27 @@ func (s *Sandbox) logHeathAndUsage(ctx *utils.LockableCancelableContext) {
...
+			sbxlogger.I(s).Error("healthcheck failed", zap.Error(err))
+			return
+		}
+
+		if ok && s.healthy.CompareAndSwap(false, true) {
+			sbxlogger.E(s).Healthcheck(sbxlogger.Success)
+			return
+		}
+
+		if alwaysReport {
```
jakubno, line 70: if alwaysReport is true and there was a change, there will be doubled logs, is that desired behavior?

packages/orchestrator/internal/sandbox/checks.go
```diff
@@ -50,7 +53,27 @@ func (s *Sandbox) logHeathAndUsage(ctx *utils.LockableCancelableContext) {
...
+			if ok {
+				sbxlogger.E(s).Healthcheck(sbxlogger.ReportSuccess)
+			} else {
+				sbxlogger.E(s).Healthcheck(sbxlogger.ReportFail)
+				sbxlogger.I(s).Error("control healthcheck failed", zap.Error(err))
+			}
+		}
 	}()
```
jakubno, line 79: there's weird check at the bottom of this function
```
	if err != nil {
		return
	}
}
```

I think we don't want to use that error? Can we use `_` or add some log + add comment there explaining what's happening? 🙏🏻

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -198,7 +209,7 @@ func NewSandbox(
 		&fc.MmdsMetadata{
 			SandboxId:            config.SandboxId,
 			TemplateId:           config.TemplateId,
-			LogsCollectorAddress: logs.CollectorPublicIP,
+			LogsCollectorAddress: os.Getenv("LOGS_COLLECTOR_PUBLIC_IP"),
```
jakubno, line 212: I think this is quite error prone, setting the same env variable all over the code base, if you set it incorrectly somewhere, it will be nightmare to debug or even realize

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -232,9 +242,9 @@ func NewSandbox(
 		StartedAt:      startedAt,
 		EndAt:          endAt,
 		rootfs:         rootfsOverlay,
-		Logger:         logger,
 		cleanup:        cleanup,
 		healthcheckCtx: healthcheckCtx,
+		healthy:        atomic.Bool{},
```
jakubno: Can we add default explicitly or add comment saying it's `false`?
```suggestion
		healthy:        atomic.Bool{}, // defaults to `false`
```

packages/orchestrator/internal/server/main.go
```diff
@@ -72,27 +71,29 @@ func New(ctx context.Context, port uint) (*Service, error) {
...
-
 	// BLOCK: initialize services
 	{
 		srv.dns = dns.New()
 
-		opts := []grpc_zap.Option{e2blogging.WithoutHealthCheck()}
+		opts := []logging.Option{
+			logging.WithLogOnEvents(logging.StartCall, logging.PayloadReceived, logging.PayloadSent, logging.FinishCall),
+			logging.WithLevels(logging.DefaultServerCodeToLevel),
+			logging.WithFieldsFromContext(logging.ExtractFields),
```
jakubno, line 81: Are we using this somewhere?

packages/api/main.go
dobrac: it is resolved after, but I can pass the logger to args if that would be clearer

packages/api/main.go
dobrac: fixed by moving the logger initialization up

packages/shared/pkg/telemetry/otel.go
```diff
@@ -57,7 +60,7 @@ func InitOTLPExporter(ctx context.Context, serviceName, serviceVersion string) f
 
 	var otelClient client
 
-	go func() {
+	{
```
jakubno: this will be blocking now, so if otel isn't present the service can't start. I think it would make sense to try once and if that failed, start it in go routine, that would probably solve some of our issues, when this takes too long and there are already some items in counters, which breaks the correct total count

packages/api/main.go
dobrac: OTEL needs to be initialized before logger, so logger can properly connect to it. I didn't notice the whole cleanup func, I will move it there too

packages/api/main.go
dobrac: just realized, this is like "otel is dependent on swagger, swagger is dependent on otel". I think removing the swagger dependency on the version might solve also other issues

packages/shared/pkg/logger/sandbox/metadata.go
```diff
@@ -0,0 +1,29 @@
...
+}
+
+func (sm SandboxMetadata) Fields() []zap.Field {
+	return []zap.Field{
+		zap.String("sandboxID", sm.SandboxID),
+		zap.String("templateID", sm.TemplateID),
+		zap.String("teamID", sm.TeamID),
+
+		// Fields for Vector
+		zap.String("instanceID", sm.SandboxID),
```
jakubno, line 26: I think we could check instanceID in vector and if present then populate sandboxID, but no need to do it here

packages/api/main.go
dobrac: actually, cleanup is dependent on logger, logger is on cleanup, any ideas how to untangle them? Do I need to use the cleanup instead of defer?

packages/shared/pkg/logger/grpc.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/selector"
+	"go.uber.org/zap"
+)
+
+func GRPCLogger(l *zap.Logger) logging.Logger {
+	return logging.LoggerFunc(func(ctx context.Context, lvl logging.Level, msg string, fields ...any) {
+		f := make([]zap.Field, 0, len(fields)/2)
+
+		methodFullNameMap := map[string]string{
+			"grpc.service":     "...",
```
jakubno, line 18: is this some specific syntax or is it just fallback value? If only fallback can we do something more explicit?

packages/shared/pkg/logger/grpc.go
```diff
@@ -0,0 +1,77 @@
...
+			"grpc.method":      "...",
+			"grpc.method_type": "...",
+			"grpc.code":        "-",
+		}
+
+		for i := 0; i < len(fields); i += 2 {
```
jakubno: there should be check fields length is divisible by 2

packages/shared/pkg/logger/exporter.go
```diff
@@ -0,0 +1,113 @@
...
+		}
+	}()
+
+	return httpWriter
+}
+
+func (h *HTTPWriter) Write(source []byte) (n int, err error) {
+	h.wg.Add(1)
+
+	p := make([]byte, len(source))
```
jakubno, line 56:
```suggestion
        // zap is reusing the buffer, so since we're processing it in a Go routine, we need to make a copy.
	p := make([]byte, len(source))
```

packages/shared/pkg/logger/exporter.go
```diff
@@ -0,0 +1,113 @@
...
+	go func() {
+		defer h.wg.Done()
+
+		start := 0
+		for i, b := range p {
+			if b == '\n' {
+				if start < i { // Ignore empty lines
+					line := p[start:i]
+					if err := h.sendLogLine(line); err != nil {
+						fmt.Fprintf(os.Stderr, "Failed to send a log line: %s\n", line)
```
jakubno: Use [log](https://pkg.go.dev/log) instead of fmt.Fprint

packages/orchestrator/internal/sandbox/checks.go
dobrac: there shouldn't be double logs, because the other ifs have return

packages/orchestrator/internal/sandbox/checks.go
dobrac: Screenshot 2025-03-04 at 5 06 23 PM

can you please explain whats happening 👀

packages/orchestrator/internal/sandbox/sandbox.go
dobrac: unfortunately the default can't be easily passed, that's how all atomics are defined 🥲  (it's like this all over the codebase, it uses the default GoLang value)

but I can add the comment if you want, should I?

packages/orchestrator/internal/server/main.go
dobrac: it's used in the grpc interceptor

packages/shared/pkg/telemetry/otel.go
dobrac: I've made it sync, because otel needs to be initialized before the global logger. Could you please expand why otel might not be present?

packages/shared/pkg/logger/sandbox/metadata.go
dobrac: 👀 vector 👀

I agree the changes could be done in vector, I just really didn't want to touch the config to not break it

packages/shared/pkg/logger/grpc.go
dobrac: it's a fallback value, any proposals?

packages/shared/pkg/logger/grpc.go
dobrac: honestly, this impl is from the documentation. There could be, but is that necessary to check correct lib implementation?

packages/shared/pkg/telemetry/otel.go
jakubno: E.g. self hosting without running otel collector, not sure if we want to make it hard requirement. If so, we should make sure otel is running even if grafana isn't present

packages/shared/pkg/logger/sandbox/metadata.go
jakubno: I will create issue, let's not do it here

packages/api/main.go
dobrac: moved program start to run() int function, use defer for loggers cleanup

packages/shared/pkg/telemetry/otel.go
dobrac: how about deciding it based on the definition of:
var otelCollectorGRPCEndpoint = os.Getenv("OTEL_COLLECTOR_GRPC_ENDPOINT")

packages/orchestrator/internal/sandbox/sandbox.go
dobrac: this IP is used only here, the address is then used only in the top main file for logger init

packages/shared/pkg/telemetry/otel.go
dobrac: returned the goroutine, as OTEL will properly initialize it even after init

packages/shared/pkg/logger/grpc.go
dobrac: added len(fields) - 1

packages/shared/pkg/logger/exporter.go
jakubno: Any reason why to not add it?

packages/shared/pkg/logger/grpc.go
```diff
@@ -0,0 +1,77 @@
...
+		f := make([]zap.Field, 0, len(fields)/2)
+
+		methodFullNameMap := map[string]string{
+			"grpc.service":     "...",
+			"grpc.method":      "...",
+			"grpc.method_type": "...",
+			"grpc.code":        "-",
+		}
+
+		for i := 0; i < len(fields)-1; i += 2 {
```
jakubno, line 24: Really small NIT: thinking about it more, wouldn't it be probably better to use quotient (whole number division: `n = q *k + m`, where `k` would be `2` and `q` would the number of iterations)

packages/client-proxy/main.go
```diff
@@ -238,19 +237,19 @@ func main() {
...
-
+		// TODO: Add timeout for ctx
 		if err := server.Shutdown(ctx); err != nil {
 			exitCode.Add(1)
 			logger.Error("http service shutdown error", zap.Int("port", port), zap.Error(err))
 		}
+
+		// TODO: Add timeout for ctx
+		logger.Info("shutting down telemetry")
+		err := stopOtlp(ctx)
```
dobrac: There is now double stopOtlp call

packages/shared/pkg/logger/grpc.go
dobrac: > Any reason why to not add it?
add what please?

packages/client-proxy/main.go
jakubno: the one is called in defer and it will be called only if logger init throws panic, because `defer` isn't called on `os.Exit`. I will move it to `run()` as in `api`, so it's easier to understand

packages/shared/pkg/logger/grpc.go
dobrac: > Really small NIT: thinking about it more

Feels a bit like overthinking :D What would the loop look like? Something like this?

```
k := 2
q := len(fields)/2

for i := 0; i < q; i++ {
   n := i * k
   key := fields[n]
   value := fields[n+1]
}
```

PR #325, opened 2025-02-25
Add create sandbox getLeastBusyNode retries

dobrac:
Adds retries to get the least busy node when starting sandbox. It will first wait for any node to be available, then it will have 3 tries to run the sandbox there.


ValentaTomas (approved):

dobrac:
Added sandbox create fails counter to the nodes admin api endpoints

ValentaTomas (approved):

PR #328, opened 2025-02-25
Add custom cache for sandbox lifecycle

ValentaTomas:
Because of problems with the current TTL cache implementation we are adding a custom one that can be simpler.


dobrac:
Includes also getLeastBusyNode improvement (wait for node availability when getting least busy node)

dobrac (approved):

PR #330, opened 2025-02-25
Share transport among reverse proxies

jakubno:
This enables reusing connections and also lowers memory footprint


packages/client-proxy/main.go
```diff
@@ -116,17 +116,8 @@ func proxyHandler(logger *zap.SugaredLogger) func(w http.ResponseWriter, r *http
...
-			MaxIdleConns:          1024,              // Matches worker_connections
-			MaxIdleConnsPerHost:   8192,              // Matches keepalive_requests
-			IdleConnTimeout:       620 * time.Second, // Matches keepalive_timeout
-			TLSHandshakeTimeout:   10 * time.Second,  // Similar to client_header_timeout
-			ResponseHeaderTimeout: 24 * time.Hour,    // Matches proxy_read_timeout
-			DisableKeepAlives:     false,             // Allow keep-alives
-		}
-
+		// Set the transport
+		proxy.Transport = transport
```
0div, line 138: unless you need those specific transport configs have you tried [Director](https://cs.opensource.google/go/go/+/refs/tags/go1.24.0:src/net/http/httputil/reverseproxy.go;l=155) to only modify the Request? I'm curious to know if it's more efficient. Probably not if we're gonna modify the request url each time.

0div (approved):

packages/client-proxy/main.go
jakubno: I think you are right, we will need only probably need only one ReverseProxy

PR #337, opened 2025-02-26
Move template manager to separate node

jakubno:
# Description

Create a new node pool for template manager and docker reverse proxy, so it doesn't interfere with sandbox orchestration

### DIFF
--- a/packages/cluster/scripts/start-build-cluster.sh
+++ b/packages/cluster/scripts/start-build-cluster.sh
@@ -31,6 +31,55 @@ net.ipv4.tcp_max_syn_backlog = 65535
 EOF
 sudo sysctl -p
 
+echo "Disabling inotify for NBD devices"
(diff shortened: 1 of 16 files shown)


packages/cluster/scripts/start-build-cluster.sh
```diff
@@ -0,0 +1,194 @@
...
+cat <<EOH >/etc/udev/rules.d/97-nbd-device.rules
+# Disable inotify watching of change events for NBD devices
+ACTION=="add|change", KERNEL=="nbd*", OPTIONS:="nowatch"
+EOH
+
+sudo udevadm control --reload-rules
+sudo udevadm trigger
+
+# Load the nbd module with 4096 devices
+sudo modprobe nbd nbds_max=4096
```
dobrac: do we need nbd for template build?

dobrac (approved):
LGTM, if you've tested it and it works

PR #339, opened 2025-02-26
Fix leaks in orchestator

jakubno:
Not tested yet

### DIFF
--- a/packages/orchestrator/internal/sandbox/fc/process.go
+++ b/packages/orchestrator/internal/sandbox/fc/process.go
@@ -39,8 +39,8 @@ type Process struct {
 
 	cmd *exec.Cmd
 
-	stdout *io.PipeReader
-	stderr *io.PipeReader
+	stdout io.ReadCloser
+	stderr io.ReadCloser
 
 	metadata *MmdsMetadata
 
@@ -53,14 +53,11 @@ type Process struct {
 	Exit chan error
 
 	client *apiClient
-
-	logger *logs.SandboxLogger
 }
 
 func NewProcess(
 	ctx context.Context,
 	tracer trace.Tracer,
-	logger *logs.SandboxLogger,
 	slot network.Slot,
 	files *storage.SandboxFiles,
 	mmdsMetadata *MmdsMetadata,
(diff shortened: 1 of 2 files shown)


dobrac:
could you please explain a bit the cause + fix?

jakubno:
These go routines are accumulating and aren't closed. I just wanted to point you in this direction, as I said, I didn't test. But I think the problem is with close being in defer, resulting in pipe not being closed and the reader waitong for the next line.

jakubno:
I will share data once i will get home

dobrac:
Change the implementation to be using cmd.StdoutPipe and cmd.StderrPipe. These should be closed automatically when the process exists. Their implementation is basically the same as ours before:

```
func (c *Cmd) StdoutPipe() (io.ReadCloser, error) {
	if c.Stdout != nil {
		return nil, errors.New("exec: Stdout already set")
	}
	if c.Process != nil {
		return nil, errors.New("exec: StdoutPipe after process started")
	}
	pr, pw, err := os.Pipe()
	if err != nil {
		return nil, err
	}
	c.Stdout = pw
	c.childIOFiles = append(c.childIOFiles, pw)
	c.parentIOPipes = append(c.parentIOPipes, pr)
	return pr, nil
}
```

There is a note from the docs about their usage, but our usage was basically the same now, so we should probably address that separately. (If it even makes sense, because we kill the FC process anyway.) 
[Cmd.Wait](https://pkg.go.dev/os/exec#Cmd.Wait) will close the pipe after seeing the command exit, so most callers need not close the pipe themselves. It is thus incorrect to call Wait before all reads from the pipe have completed. For the same reason, it is incorrect to call [Cmd.Run](https://pkg.go.dev/os/exec#Cmd.Run) when using StdoutPipe. See the example for idiomatic usage.

packages/orchestrator/internal/sandbox/fc/process.go
```diff
@@ -117,11 +117,15 @@ func NewProcess(
 		Pgid:    0,
 	}
 
-	cmdStdoutReader, cmdStdoutWriter := io.Pipe()
-	cmd.Stdout = cmdStdoutWriter
+	cmdStdoutReader, err := cmd.StdoutPipe()
```
ValentaTomas: I think we can create the pipes in the `.Start` method and remove the fields from the struct.

ValentaTomas (approved):

PR #347, opened 2025-02-28
Don't block on writing logs

jakubno:
# Description

Not sure about initial cause yet, but if there's a lot of logs, I think it can happen the buffer in `packages/shared/pkg/logs/exporter/exporter.go` gets full, blocking everything.

This exporter also sends each line separately, which can maybe result in some rate limiting? I explains why we see the start times incremented by 2 seconds, the timeout for http client is 2 seconds. I think we should be rewrite sandbox exporter, not sure if we want to do it before #321


ValentaTomas:
@jakubno and @0div are reworking this as the multiline/batching in vector was more complicated.

jakubno:
part of #321

PR #354, opened 2025-03-03
Change update policy for dev clusters

jakubno:
Improves DX: When you change machine types, start script or something similar, you expect the change to happen right away. It can be confusing and annoying for development.

### DIFF
--- a/packages/cluster/api-cluster/main.tf
+++ b/packages/cluster/api-cluster/main.tf
@@ -48,7 +48,7 @@ resource "google_compute_instance_group_manager" "api_cluster" {
   # Server is a stateful cluster, so the update strategy used to roll out a new GCE Instance Template must be
   # a rolling update.
   update_policy {
-    type                    = var.instance_group_update_policy_type
+    type                    = var.environment == "prod" ? "OPPORUNISTIC" : "PROACTIVE"
     minimal_action          = var.instance_group_update_policy_minimal_action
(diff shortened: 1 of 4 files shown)


ValentaTomas:
Shouldn't the "production" config be default and the development be the one you need to explicitly specify?

packages/cluster/api-cluster/main.tf
```diff
@@ -48,7 +48,7 @@ resource "google_compute_instance_group_manager" "api_cluster" {
   # Server is a stateful cluster, so the update strategy used to roll out a new GCE Instance Template must be
   # a rolling update.
   update_policy {
-    type                    = var.instance_group_update_policy_type
+    type                    = var.environment == "prod" ? "OPPORUNISTIC" : "PROACTIVE"
```
dobrac: I agree with having the "prod" behaviour as default and the "dev" behaviour as set -> you won't accidentally break something

dobrac: Also, we have used previously the instance_group_update_policy_type var. Can't we use it anymore? If not, may it be deleted?

dobrac (changes requested):

packages/cluster/api-cluster/main.tf
jakubno: removed

jakubno: Changed to safer behavior

PR #355, opened 2025-03-03
Remove template logs from API cache

sitole:
Template build logs are pushed to Loki and API is using Loki directly to query them for user instead of local API cache.

**[Deploy]**
First template manager needs to be rolled so it will push logs to logger and also for back compatibility via grcp stream. Later api can be rolled later, there are no user-facing changes here or come breaking one.

### DIFF
--- a/packages/api/internal/handlers/template_build_logs.go
+++ b/packages/api/internal/handlers/template_build_logs.go
@@ -84,6 +84,7 @@ func (a *APIStore) GetTemplatesTemplateIDBuildsBuildIDStatus(c *gin.Context, tem
 	// https://grafana.com/blog/2021/01/05/how-to-escape-special-characters-with-lokis-logql/
 	templateIdSanitized := strings.ReplaceAll(templateID, "`", "")
 	query := fmt.Sprintf("{source=\"logs-collector\", service=\"template-manager\", buildID=\"%s\", envID=`%s`}", buildUUID.String(), templateIdSanitized)
+
 	end := time.Now()
 	start := end.Add(-templateBuildOldestLogsLimit)
 
@@ -98,7 +99,7 @@ func (a *APIStore) GetTemplatesTemplateIDBuildsBuildIDStatus(c *gin.Context, tem
 	}
 
 	logs := make([]string, 0)
-	logsSkippedOffset := 0
+	logsCrawled := 0
 
 	if res.Data.Result.Type() != loghttp.ResultTypeStream {
 		zap.L().Error("unexpected value type received from loki query fetch", zap.String("type", string(res.Data.Result.Type())))
@@ -108,10 +109,10 @@ func (a *APIStore) GetTemplatesTemplateIDBuildsBuildIDStatus(c *gin.Context, tem
 
 	for _, stream := range res.Data.Result.(loghttp.Streams) {
 		for _, entry := range stream.Entries {
-			logsSkippedOffset++
+			logsCrawled++
 
 			// loki does not support offset pagination, so we need to skip logs manually
-			if params.LogsOffset != nil && logsSkippedOffset < int(*params.LogsOffset) {
+			if logsCrawled <= int(*params.LogsOffset) {
 				continue
 			}
 
--- a/packages/api/internal/handlers/template_build_logs.go
+++ b/packages/api/internal/handlers/template_build_logs.go
@@ -1,15 +1,25 @@
 package handlers
 
 import (
+	"encoding/json"
 	"fmt"
+	"github.com/e2b-dev/infra/packages/shared/pkg/models"
+	"github.com/grafana/loki/pkg/loghttp"
+	"github.com/grafana/loki/pkg/logproto"
+	"go.uber.org/zap"
 	"net/http"
+	"time"
 
 	"github.com/e2b-dev/infra/packages/api/internal/api"
 	"github.com/e2b-dev/infra/packages/api/internal/auth"
-	"github.com/e2b-dev/infra/packages/shared/pkg/models"
 	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
 	"github.com/gin-gonic/gin"
 	"github.com/google/uuid"
+)
+
+const (
+	templateBuildLogsLimit       = 100
+	templateBuildOldestLogsLimit = 24 * time.Hour // 1 day
 )
 
 // GetTemplatesTemplateIDBuildsBuildIDStatus serves to get a template build status (e.g. to CLI)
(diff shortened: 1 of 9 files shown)


ValentaTomas:
Two notes here:
- You don't have to mention the linear issue in the name
- Add at least one label to the PR

packages/template-manager/main.go
```diff
@@ -59,17 +57,31 @@ func main() {
...
+	logger := sbxlogger.NewLogger(
+		ctx,
+		sbxlogger.SandboxLoggerConfig{
+			ServiceName:      constants.ServiceName,
+			IsInternal:       true,
+			CollectorAddress: os.Getenv("LOGS_COLLECTOR_ADDRESS"),
+		},
+	)
 	defer logger.Sync()
-	zap.ReplaceGlobals(logger)
```
jakubno, line 69: I think this should stay

packages/template-manager/main.go
```diff
@@ -4,14 +4,12 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	sbxlogger "github.com/e2b-dev/infra/packages/shared/pkg/logger/sandbox"
```
jakubno: NIT: this should be in the other import group

packages/template-manager/internal/server/main.go
```diff
@@ -30,13 +30,15 @@ type serverStore struct {
...
+	logger             *zap.Logger
+	buildLogger        *zap.Logger
 	dockerClient       *client.Client
 	legacyDockerClient *docker.Client
 	artifactRegistry   *artifactregistry.Client
 	templateStorage    *template.Storage
 }
 
-func New(logger *zap.Logger) *grpc.Server {
+func New(logger *zap.Logger, buildLogger *zap.Logger) *grpc.Server {
```
jakubno, line 41: can you use `sbxlogger.E()`

packages/template-manager/internal/build/writer/writer.go
```diff
@@ -2,13 +2,16 @@ package writer
...
 )
 
 type BuildLogsWriter struct {
 	stream template_manager.TemplateService_TemplateCreateServer
+	logger *zap.Logger
 }
 
 func (w BuildLogsWriter) Write(p []byte) (n int, err error) {
+	w.logger.Info(string(p))
 	err = w.stream.Send(&template_manager.TemplateBuildLog{Log: string(p)})
```
jakubno:
```suggestion
        log := string(p)
	w.logger.Info(log)
	err = w.stream.Send(&template_manager.TemplateBuildLog{Log: log})
```

packages/api/internal/handlers/template_build_logs.go
```diff
@@ -68,10 +78,48 @@ func (a *APIStore) GetTemplatesTemplateIDBuildsBuildIDStatus(c *gin.Context, tem
 	}
 
 	status := dockerBuild.GetStatus()
-	logs := dockerBuild.GetLogs()
+
+	query := fmt.Sprintf("{source=\"logs-collector\", service=\"template-manager\", buildID=\"%s\", envID=\"%s\"}", buildUUID.String(), templateID)
```
jakubno: Sanitize IDs as in other handlers (logs, metrics)
https://grafana.com/blog/2021/01/05/how-to-escape-special-characters-with-lokis-logql/

packages/api/internal/handlers/template_build_logs.go
```diff
@@ -68,10 +78,48 @@ func (a *APIStore) GetTemplatesTemplateIDBuildsBuildIDStatus(c *gin.Context, tem
...
+	end := time.Now()
+	start := end.Add(-templateBuildOldestLogsLimit)
+
+	res, err := a.lokiClient.QueryRange(query, templateBuildLogsLimit, start, end, logproto.FORWARD, time.Duration(0), time.Duration(0), true)
+	if err != nil {
+		errMsg := fmt.Errorf("error when returning logs for template build: %w", err)
```
jakubno, line 93: log error

packages/api/internal/handlers/template_build_logs.go
```diff
@@ -68,10 +78,48 @@ func (a *APIStore) GetTemplatesTemplateIDBuildsBuildIDStatus(c *gin.Context, tem
...
+	switch res.Data.Result.Type() {
+	case loghttp.ResultTypeStream:
+		value := res.Data.Result.(loghttp.Streams)
+
+		for _, stream := range value {
+			for _, entry := range stream.Entries {
+				logsSkippedOffset++
+
+				// loki does not support offset pagination, so we need to skip logs manually
+				if logsSkippedOffset < int(*params.LogsOffset) {
```
jakubno: what if the logs offset is not set

packages/api/internal/handlers/template_build_logs.go
```diff
@@ -68,10 +78,48 @@ func (a *APIStore) GetTemplatesTemplateIDBuildsBuildIDStatus(c *gin.Context, tem
...
+		errMsg := fmt.Errorf("error when returning logs for template build: %w", err)
+		telemetry.ReportCriticalError(ctx, errMsg)
+		a.sendAPIStoreError(c, http.StatusInternalServerError, fmt.Sprintf("Error returning logs for template build '%s'", buildUUID.String()))
+		return
+	}
+
+	logs := make([]string, 0)
+	logsSkippedOffset := 0
+
+	switch res.Data.Result.Type() {
```
jakubno: I think it shouldn't happen, but can you add some check as in https://github.com/e2b-dev/infra/blob/main/packages/api/internal/handlers/sandbox_metrics.go#L53C1-L55C3

you don't have to switch then

packages/template-manager/internal/server/main.go
sitole: After discussion we decided to use direct propagation. Will be changed in future tooling logging will be adjusted.

sitole:
> Feel free to lower the interval, maybe print every 5 seconds "Still processing" with timestamp and no need to add /r. If you have a better idea, please go ahead and do it your way. :)

Interval lowered to 5 seconds. Now we are sending ANSIC formatted time instead of seconds since start of build.

packages/template-manager/internal/build/rootfs.go
```diff
@@ -171,16 +171,13 @@ type PostProcessor struct {
 
 // Start starts the post-processing.
 func (p *PostProcessor) Start() {
-
-	now := time.Now()
 	for {
```
jakubno, line 180:
```suggestion
        p.writer.Write(fmt.Sprintf("[%s] Start postprocessing\n", time.Now().Format(time.ANSIC)))

	for {
```

packages/template-manager/internal/build/rootfs.go
```diff
@@ -171,16 +171,13 @@ type PostProcessor struct {
...
-		msg := []byte(fmt.Sprintf("Postprocessing (%s)       \r", time.Since(now).Round(time.Second)))
+		msg := []byte(fmt.Sprintf("Postprocessing (%s)       \n", time.Now().Format(time.ANSIC)))
```
jakubno:
```suggestion
		msg := []byte(fmt.Sprintf("[%s] Postprocessing\n", time.Now().Format(time.ANSIC)))
```

packages/template-manager/internal/build/rootfs.go
```diff
@@ -171,16 +171,13 @@ type PostProcessor struct {
...
 
 		select {
 		case postprocessingErr := <-p.errChan:
 			if postprocessingErr != nil {
 				p.writer.Write([]byte(fmt.Sprintf("Postprocessing failed: %s\n", postprocessingErr)))
```
jakubno:
```suggestion
				p.writer.Write([]byte(fmt.Sprintf("[%s] Postprocessing failed: %s\n", time.Now().Format(time.ANSIC), postprocessingErr)))
```

packages/template-manager/internal/build/rootfs.go
```diff
@@ -171,16 +171,13 @@ type PostProcessor struct {
...
-
 				return
 			}
```
jakubno, line 189: Do the same for postprocessing finished 🙏🏻

jakubno (commented):
Only small improvements for strings, I will do quick test in the mean time and we can merge

jakubno:
Please update description and if you can add if there's some required order for deploy 🙏🏻

jakubno (commented):
One more thing, can we use ISO format?

jakubno (changes requested):
I've this log twice and little bit differently each time 
image
image

sitole:
> I've this log twice and little bit differently each time
Fixed, there was forgotten off-by-one error that you mentioned yesterday and i remove fix it with some other changes

> One more thing, can we use ISO format?
Done
image

sitole:
First template manager needs to be rolled so it will push logs to logger and also for back compatibility via grcp stream. Later api can be rolled later, there are no user-facing changes here or come breaking one.

packages/api/internal/handlers/template_build_logs.go
```diff
@@ -68,10 +79,55 @@ func (a *APIStore) GetTemplatesTemplateIDBuildsBuildIDStatus(c *gin.Context, tem
...
+		telemetry.ReportCriticalError(ctx, errMsg)
+		zap.L().Error("error when returning logs for template build", zap.Error(err), zap.String("buildID", buildID))
+
+		a.sendAPIStoreError(c, http.StatusInternalServerError, fmt.Sprintf("Error returning logs for template build '%s'", buildUUID.String()))
+		return
+	}
+
+	logs := make([]string, 0)
+	logsCrawled := 0
+
```
jakubno, line 103: There's still nil dereference, if you don't set LogsOffset which is optional
```suggestion
	offset := 0
	if params.LogsOffset != nil {
		offset = int(*params.LogsOffset)
	}
```

packages/api/internal/handlers/template_build_logs.go
```diff
@@ -68,10 +79,55 @@ func (a *APIStore) GetTemplatesTemplateIDBuildsBuildIDStatus(c *gin.Context, tem
...
+		a.sendAPIStoreError(c, http.StatusInternalServerError, "Unexpected error during fetching logs")
+		return
+	}
+
+	for _, stream := range res.Data.Result.(loghttp.Streams) {
+		for _, entry := range stream.Entries {
+			logsCrawled++
+
+			// loki does not support offset pagination, so we need to skip logs manually
+			if logsCrawled <= int(*params.LogsOffset) {
```
jakubno:
```suggestion
			if logsCrawled <= offset {
```

jakubno (changes requested):

jakubno (approved):

PR #357, opened 2025-03-03
Increase template ram allocation for select templates

ValentaTomas:
- Needed to create a new PR because the previous one was merged to foxtrot.


r33drichards (approved):
lgtm

PR #358, opened 2025-03-03
Test time synchronization on built templates

r33drichards:
Sometimes our snapshots can be snapshotted with incorrect times. Although we have business logic to sync the firecracker instance's clock, sometimes this can fail. This test verifies that our time is working


r33drichards:
https://github.com/e2b-dev/infra/actions/runs/13641665262/job/38132697132

r33drichards:
https://linear.app/e2b/issue/E2B-1665/periodic-tests-for-clock-syncing

ValentaTomas:
I think this could be part of the normal build test.

r33drichards:
> I think this could be part of the normal build test.

it could, but that would violate Single Responsibility Principle, is there something you are trying to avoid keeping this seperate?

r33drichards:
right, so why not decouple the tests? wouldn't we want to know the state of these things independently of each other?

r33drichards:
after taking another look it does seem like this mostly the same test as https://github.com/e2b-dev/infra/blob/424b88bf8543af4358901b008c4b67e11ce7eeaf/tests/periodic-test/build-template/index.ts so maybe it makes sense to eliminate that one in favor of this. still think that makes sense as a seperate pr

dobrac:
With connection to this discussion, wouldn't the best case be building the template once, and then run many different tests on the already built template? (And those tests might fail independently, maybe having playwright or some other testing framework for doing that.) I don't think it makes sense to create new build for each additional tests what we would like to do, as it doesn't scale the best imo.

r33drichards:
@dobrac @ValentaTomas updated and tested this here https://github.com/e2b-dev/infra/actions/runs/13774996892

ValentaTomas (approved):

ValentaTomas (approved):

ValentaTomas:
@r33drichards Let's resolve conflicts and merge.

PR #359, opened 2025-03-03
don't use footer for link so its clickable on mobile

r33drichards:


ValentaTomas:
I think `SLACK_FOOTER: ""` should now remove the default footer there.

r33drichards:
> I think `SLACK_FOOTER: ""` should now remove the default footer there.

should be gtg

jakubno (approved):

PR #360, opened 2025-03-03
Test logging from e2b cli is operating normally

r33drichards:
test that our logging infrastructure is operational

### DIFF
--- a/tests/periodic-test/cli-logs.ts
+++ b/tests/periodic-test/cli-logs.ts
@@ -11,11 +11,8 @@ try {
     sandbox = await Sandbox.create();
     console.log('Sandbox created with ID:', sandbox.sandboxId);
 
-    let strippedId = sandbox.sandboxId.split('-')[0]
-    console.log('strippedId:', strippedId)
-
     const command = new Deno.Command("npx", {
-        args: ["@e2b/cli", "sandbox", "logs", "-f", strippedId],
+        args: ["@e2b/cli", "sandbox", "logs", "-f", sandbox.sandboxId],
         stdout: "piped",
         stderr: "piped",
     });
--- a/tests/periodic-test/cli-logs.ts
+++ b/tests/periodic-test/cli-logs.ts
@@ -23,7 +23,6 @@ try {
     const child = command.spawn();
 
     // Start collecting logs in background
-    cliProcess = runCliLogsCommand(strippedId);
     console.log('Started CLI logs collection');
 
     // Kill the sandbox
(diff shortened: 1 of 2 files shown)


r33drichards:
https://github.com/e2b-dev/infra/actions/runs/13642949038/job/38136476768#step:9:46

tests/periodic-test/cli-logs.ts
```diff
@@ -0,0 +1,57 @@
...
+    console.log('Sandbox created with ID:', sandbox.sandboxId);
+
+    let strippedId = sandbox.sandboxId.split('-')[0]
+    console.log('strippedId:', strippedId)
+
+    const command = new Deno.Command("npx", {
+        args: ["@e2b/cli", "sandbox", "logs", "-f", strippedId],
+        stdout: "piped",
+        stderr: "piped",
+    });
```
jakubno: I think you don't have to do this, but it doesn't really matter

```suggestion
    const command = new Deno.Command("npx", {
        args: ["@e2b/cli", "sandbox", "logs", "-f", sandbox.sandboxId],
        stdout: "piped",
        stderr: "piped",
    });
```

tests/periodic-test/cli-logs.ts
```diff
@@ -0,0 +1,57 @@
...
+    await sandbox.kill();
+
+    const output = await child.output();
+    const decoder = new TextDecoder();
+    // Wait for CLI process to complete and get its output
+    const stdout = decoder.decode(output.stdout);
+    console.log('CLI process completed');
+
+    // Assert that we got some logs
+    if (!stdout.trim()) {
```
jakubno: Did you test it with some sandbox id which doesn't exist? It will always print 
```
Logs for sandbox <sandbox id>:

Stopped printing logs — sandbox not found
```

This would be worse than having no test at all, giving us false sense of safety

jakubno (changes requested):

tests/periodic-test/cli-logs.ts
r33drichards: added
```
    stdout = stdout.split('\n').filter(line => !line.includes(`Logs for sandbox`)).join('\n');
    stdout = stdout.split('\n').filter(line => !line.includes('Stopped printing logs — sandbox not found')).join('\n');
```
to handle this case

jakubno:
Sorry, I rewrote the test a little. 
- Using the pretty format was flaky, if we change some text in CLI, it would break
- Using `follow` flag was flaky, because it takes some time to propagate the logs
- Added check for having both type of logs - from API and from sandbox

PR #363, opened 2025-03-04
Speed up orchestrator nodes discovery in API

jakubno:
# Description
- Change period for check from 1 min to 20s
- Wait till some node is loaded on start of API
- Add health check for connected nodes
- Increase time for API to start

### DIFF
--- a/packages/api/internal/handlers/store.go
+++ b/packages/api/internal/handlers/store.go
@@ -114,8 +114,8 @@ func NewAPIStore(ctx context.Context) *APIStore {
 	authCache := authcache.NewTeamAuthCache(dbClient)
 	templateSpawnCounter := utils.NewTemplateSpawnCounter(time.Minute, dbClient)
 
-	return &APIStore{
-		Healthy:              true,
+	a := &APIStore{
+		Healthy:              false,
 		orchestrator:         orch,
 		templateManager:      templateManager,
 		db:                   dbClient,
@@ -128,6 +128,19 @@ func NewAPIStore(ctx context.Context) *APIStore {
 		authCache:            authCache,
 		templateSpawnCounter: templateSpawnCounter,
 	}
+
+	go func() {
+		for {
+			if orch.NodeCount() != 0 {
+				a.Healthy = true
+				break
+			}
+
+			time.Sleep(time.Millisecond)
+		}
+	}()
+
+	return a
 }
 
 func (a *APIStore) Close(ctx context.Context) error {
--- a/packages/api/internal/handlers/store.go
+++ b/packages/api/internal/handlers/store.go
@@ -129,6 +129,7 @@ func NewAPIStore(ctx context.Context) *APIStore {
 		templateSpawnCounter: templateSpawnCounter,
 	}
 
+	// Wait till there's at least one, otherwise we can't create sandboxes yet
 	go func() {
 		for {
 			if orch.NodeCount() != 0 {
--- a/packages/api/internal/orchestrator/cache.go
+++ b/packages/api/internal/orchestrator/cache.go
@@ -19,6 +19,8 @@ import (
 	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
 )
 
+const cacheSyncTime = 20 * time.Second
+
 func (o *Orchestrator) GetSandbox(sandboxID string) (*instance.InstanceInfo, error) {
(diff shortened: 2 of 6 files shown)


packages/api/internal/handlers/store.go
```diff
@@ -128,6 +128,20 @@ func NewAPIStore(ctx context.Context) *APIStore {
 		authCache:            authCache,
 		templateSpawnCounter: templateSpawnCounter,
 	}
+
+	// Wait till there's at least one, otherwise we can't create sandboxes yet
+	go func() {
+		for {
```
dobrac, line 135: can this goroutine exit on context end, so we have proper cleanup if required?

packages/api/internal/orchestrator/cache.go
```diff
@@ -70,7 +74,21 @@ func (o *Orchestrator) syncNodes(ctx context.Context, instanceCache *instance.In
...
+			if err != nil {
+				zap.L().Warn("Error checking node health:", zap.String("node_id", n.ID), zap.Error(err))
+				zap.L().Info("Removing node from list:", zap.String("node_id", n.ID))
+				o.nodes.Remove(n.ID)
+			}
+			if health.Status != grpc_health_v1.HealthCheckResponse_SERVING {
+				zap.L().Warn("Node is not healthy:", zap.String("node_id", n.ID), zap.String("status", health.Status.String()))
+				zap.L().Info("Removing node from list:", zap.String("node_id", n.ID))
+				o.nodes.Remove(n.ID)
+			}
```
dobrac: Is this healthcheck necessary here? When we remove node from the list just because it's not healthy, we are losing track of it and resetting all internal temporary state

Non responsive node should return non Ok status, so no scheduling there should be happening


(also, do we need two log message for the same event?)

packages/api/internal/orchestrator/cache.go
```diff
@@ -36,7 +39,7 @@ func (o *Orchestrator) keepInSync(ctx context.Context, instanceCache *instance.I
 			o.logger.Info("Stopping keepInSync")
 
 			return
-		case <-time.After(instance.CacheSyncTime):
+		case <-time.After(cacheSyncTime):
```
dobrac: The reason why the first sync is slow is that this is triggered after cacheSyncTime. To solve this, you could either call the o.syncNodes, or just add something like this:

```
firstSync := make(chan struct{}, 1)
firstSync <- struct{}{}

for {
  select {
    case <-ctx.Done():
	    zap.L().Info("Stopping keepInSync")
    
	    return
    case <-firstSync:
	    o.syncNodes(ctx, instanceCache)
    case <-time.After(instance.CacheSyncTime):
            //...
```

packages/api/internal/orchestrator/cache.go
```diff
@@ -70,7 +81,18 @@ func (o *Orchestrator) syncNodes(ctx context.Context, instanceCache *instance.In
 					o.logger.Errorf("Error connecting to node: %v", err)
 				}
 			}(n)
+		} else {
+			// Check if the node is healthy
+			health, err := orchNode.Client.Health.Check(ctx, &grpc_health_v1.HealthCheckRequest{})
```
dobrac: use ctxTimeout instead of ctx

packages/api/internal/orchestrator/cache.go
```diff
@@ -70,7 +81,18 @@ func (o *Orchestrator) syncNodes(ctx context.Context, instanceCache *instance.In
...
+			if err != nil || health.Status != grpc_health_v1.HealthCheckResponse_SERVING {
+				orchNode.SetStatus(api.NodeStatusUnhealthy)
```
dobrac: can't this reset the draining status?

jakubno: Good point! We probably want this only for healthy/unhealthy nodes

dobrac: one more thing related, would it be possible to somehow simplify all this logic/make it more transparent what it does and how it behaves when anyone else will find this code? (Maybe extract it to a separate function too)

dobrac: or maybe take this check outside of the sync logic, as it might be actually helpful in general for reporting (and managing) API health

dobrac: -> will be done in another PR

dobrac (approved):

PR #366, opened 2025-03-04
Set sandbox healthy status to true by default

jakubno:
# Description

- If not we would report the first health check as a change in the status
- Runs the initial sandbox health check

### DIFF
--- a/packages/orchestrator/internal/sandbox/checks.go
+++ b/packages/orchestrator/internal/sandbox/checks.go
@@ -69,12 +69,9 @@ func (s *Sandbox) Healthcheck(ctx context.Context, alwaysReport bool) {
 		}
 
 		if alwaysReport {
-			fmt.Println("This was called alwaysReport")
 			if ok {
-				fmt.Println("This was called alwaysReport ok")
 				sbxlogger.E(s).Healthcheck(sbxlogger.ReportSuccess)
 			} else {
-				fmt.Println("This was called alwaysReport not ok")
 				sbxlogger.E(s).Healthcheck(sbxlogger.ReportFail)
 				sbxlogger.I(s).Error("control healthcheck failed", zap.Error(err))
 			}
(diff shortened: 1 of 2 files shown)


packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -57,7 +57,7 @@ type Sandbox struct {
 	template template.Template
 
 	healthcheckCtx *utils.LockableCancelableContext
-	healthy        atomic.Bool
+	healthy        *atomic.Bool
```
dobrac: instead of making it pointer, you can initialize it using store after the instance is created

packages/orchestrator/internal/sandbox/checks.go
```diff
@@ -68,9 +69,12 @@ func (s *Sandbox) Healthcheck(ctx context.Context, alwaysReport bool) {
 		}
 
 		if alwaysReport {
+			fmt.Println("This was called alwaysReport")
```
```suggestion
```

packages/orchestrator/internal/sandbox/checks.go
```diff
@@ -68,9 +69,12 @@ func (s *Sandbox) Healthcheck(ctx context.Context, alwaysReport bool) {
...
 			if ok {
+				fmt.Println("This was called alwaysReport ok")
```
```suggestion
```

packages/orchestrator/internal/sandbox/checks.go
```diff
@@ -68,9 +69,12 @@ func (s *Sandbox) Healthcheck(ctx context.Context, alwaysReport bool) {
...
 				sbxlogger.E(s).Healthcheck(sbxlogger.ReportSuccess)
 			} else {
+				fmt.Println("This was called alwaysReport not ok")
```
```suggestion
```

packages/orchestrator/internal/sandbox/sandbox.go
jakubno: Thanks for this

dobrac (approved):

PR #369, opened 2025-03-04
remove unused compiler option from deno.json

r33drichards:


r33drichards:
https://github.com/e2b-dev/infra/actions/runs/13664309143

PR #370, opened 2025-03-05
Deploy clickhouse on nomad

r33drichards:
This change adds clickhouse and corresponding gcs bucket and permissions for clickhouse to access the bucket to our nomad cluster

### DIFF
--- a/packages/nomad/clickhouse.hcl
+++ b/packages/nomad/clickhouse.hcl
@@ -44,8 +44,10 @@ job "clickhouse" {
           nofile = "262144:262144"
         }
 
+
         volumes = [
           "local/config.xml:/etc/clickhouse-server/config.d/gcs.xml",
+          "local/users.xml:/etc/clickhouse-server/users.d/users.xml",
         ]
       }
 
@@ -83,6 +85,26 @@ job "clickhouse" {
 </clickhouse>
 EOF
         destination = "local/config.xml"
+      }
+
+      template {
+        data = <<EOF
+<?xml version="1.0"?>
+<clickhouse>
+    <users>
+        <${username}>
+            <password_sha256_hex>${password_sha256_hex}</password_sha256_hex>
+            <networks>
+                <ip>::/0</ip>
+            </networks>
+            <profile>default</profile>
+            <quota>default</quota>
+            <access_management>1</access_management>
+        </${username}>
+    </users>
+</clickhouse>
+EOF
+        destination = "local/users.xml"
       }
 
       resources {
--- a/packages/nomad/clickhouse.hcl
+++ b/packages/nomad/clickhouse.hcl
@@ -47,7 +47,8 @@ job "clickhouse" {
 
         volumes = [
           "local/config.xml:/etc/clickhouse-server/config.d/gcs.xml",
-          "local/users.xml:/etc/clickhouse-server/users.d/users.xml",
+          # disabled while testing but will pass password to orchestrator in the future
+          # "local/users.xml:/etc/clickhouse-server/users.d/users.xml",
         ]
       }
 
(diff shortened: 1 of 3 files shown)


r33drichards:
https://linear.app/e2b/issue/E2B-1583/add-the-terraform-vars-and-secrets-needed-to-run-clickhouse-client-on

r33drichards:
image
image

packages/nomad/clickhouse.hcl
```diff
@@ -0,0 +1,94 @@
...
+                    <main>
+                        <disk>gcs_cache</disk>
+                    </main>
+                </volumes>
+            </gcs_main>
+        </policies>
+    </storage_configuration>
+</clickhouse>
+EOF
+        destination = "local/config.xml"
```
0div, line 96: i think u mentioned you were gonna add a [users.xml](https://clickhouse.com/docs/operations/settings/settings-users) ?

packages/nomad/clickhouse.hcl
```diff
@@ -0,0 +1,94 @@
...
+        <policies>
+            <gcs_main>
+                <volumes>
+                    <main>
+                        <disk>gcs_cache</disk>
+                    </main>
+                </volumes>
+            </gcs_main>
+        </policies>
+    </storage_configuration>
```
0div, line 93: i'd advise 70% of ram for [max_memory_usage](https://github.com/ClickHouse/ClickHouse/blob/master/programs/server/config.xml#L386)

0div, line 93: this [system dashboard ](https://clickhouse.com/docs/operations/system-tables/dashboards) config can be useful if we're managing this ourselves

r33drichards: do you have any heuristics for how much ram and cpu to allocate for around 9 million daily writes?

0div: i do not! we'll have to find out empirically. In the context of my comment i meant the system's RAM in case that wasn't clear

r33drichards:
> * How would the rolling update look like?

clickhouse is stateless and storage is going to gcp bucket so normal nomad deployments should "just work". The only thing that isn't certain for me is the behavior around a in flight write. 

> * This should probably be in a separate instance group/in the API instance group

makes sense

> * The bucket setup should be in the buckets directory/module with other buckets

can do this! it made sense to me to keep the dependencies  close to each other to make things easier to reason about. Grouping similar resources together in theory makes sense but then all of the dependencies for resources become scattered and its harder to get a sense of what's going on at times. Secrets suffer from this as well.

packages/nomad/clickhouse.hcl
r33drichards: nah makes sense!! just was wondering if you had any guidance or experience for resource allocation as a whole for this as well :)

jakubno:
There still should be some `update` stanza in node, so you have can roll update it without downtime, also check if there's any expected signal for termination and behavior around that

packages/nomad/clickhouse.hcl
r33drichards: looks like there is also [max_server_memory_usage_to_ram_ratio](https://clickhouse.com/docs/operations/server-configuration-parameters/settings#max_server_memory_usage_to_ram_ratio)

> Same as [max_server_memory_usage](https://clickhouse.com/docs/operations/server-configuration-parameters/settings#max_server_memory_usage) but in a ratio to physical RAM. Allows lowering the memory usage on low-memory systems.

ValentaTomas:
Two questions here:
- Are the questions from @0div and @jakubno resolved?
- Can you rerequest review from @0div if so?

ValentaTomas (approved):

PR #377, opened 2025-03-05
Fix break in initial health loop in API

jakubno:
# Description

Break was "caught" by select, resulting in infinite cycle


dobrac (commented):
Nice

dobrac (approved):

PR #378, opened 2025-03-05
Move template build state away from api

sitole:


sitole:
Currently to tbd list:
- [ ] Implement build state pulling on api side
- [ ] Store metadata (root fs size, envd version) received in build state pulling
- [ ] Background checks periodically/when api starts if there are some builds in process and check status/fail them if too old
- [ ] Fully remove build cache from api

packages/api/internal/template-manager/create_template.go
```diff
@@ -64,49 +52,9 @@ func (tm *TemplateManager) CreateTemplate(
 			StartCommand:       startCommand,
 		},
 	})
-	err = utils.UnwrapGRPCError(err)
-	if err != nil {
-		return fmt.Errorf("failed to create template '%s': %w", templateID, err)
-	}
 
-	// Wait for the build to finish and save logs
```
sitole, line 72: `createTemplate` grpc request is no more stream. It will request build and receive result instantly. Build is done in goroutine and checked with `templateBuildStatus` method.

packages/template-manager/internal/server/create_template.go
```diff
@@ -63,79 +52,34 @@ func (s *serverStore) TemplateCreate(templateRequest *template_manager.TemplateC
 		BuildLogsWriter: logsWriter,
 	}
 
-	buildStorage := s.templateStorage.NewBuild(template.TemplateFiles)
```
sitole, line 66: logic for template build extracted from grpc handler to template builder service

sitole:
testing that i done on new flow:
- [X] check of build template fails on when template-manager async run pushes fail status to cache
- [X] background sync should fetch and check waiting and cancel them if takes too long (10 minutes to build and push)
- [X] background sync checked if takes automatically processing builds after restart of api
- [X] try to start api with some builds that are marked as in process (building, waiting) and check for resolve
- [X] check if there is not some issue with global background runner (too fast fails)

packages/api/internal/handlers/template_start_build.go
```diff
@@ -88,9 +88,9 @@ func (a *APIStore) PostTemplatesTemplateIDBuildsBuildID(c *gin.Context, template
 		attribute.String("template.id", templateID),
 	)
 
-	// Create a new build cache for storing logs
-	err = a.buildCache.Create(templateID, buildUUID, team.ID)
-	if err != nil {
+	// Check if some other build from same template is in process, excludes currently created build
+	envBuildDB, err := a.db.GetRunningEnvBuild(ctx, envDB.ID, []uuid.UUID{buildUUID})
+	if err == nil && envBuildDB != nil {
```
jakubno: what if there's an error

jakubno: Also I think you should check and create a new build in one transaction if we want to make sure there are no 2 concurrent builds.

packages/template-manager/template-manager.proto
```diff
@@ -23,20 +23,37 @@ message TemplateCreateRequest {
...
+}
+
 // Data required for deleting a template.
 message TemplateBuildDeleteRequest {
   string buildID = 1;
 }
 
+message TemplateBuildMetadata {
+  int32 rootfsSizeKey = 3;
+  string envdVersionKey = 4;
```
jakubno: I think this can be 1 and 2 , right?

packages/template-manager/internal/server/template_status.go
```diff
@@ -0,0 +1,66 @@
...
+import (
+	"context"
+	"fmt"
+	"github.com/e2b-dev/infra/packages/shared/pkg/grpc/template-manager"
+	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
+	"go.uber.org/zap"
+	"time"
+)
+
+const statusTimeout = time.Minute * 5
```
jakubno: that's basically the limit for the whole build isn't it? I think we may need to increase it. I think 15 minutes should be reasonable based on duration of some builds

packages/template-manager/internal/server/template_status.go
```diff
@@ -0,0 +1,66 @@
...
+
+func (s *serverStore) TemplateBuildStatus(in *template_manager.TemplateStatusRequest, stream template_manager.TemplateService_TemplateBuildStatusServer) error {
+	ctx, cancel := context.WithTimeout(stream.Context(), statusTimeout)
+	defer cancel()
+
+	childCtx, childSpan := s.tracer.Start(ctx, "template-build-status-request")
```
jakubno: There's probably no reason why to use original `ctx` and it's easier to not get it wrong
```suggestion
	ctx, childSpan := s.tracer.Start(ctx, "template-build-status-request")
```

packages/template-manager/template-manager.proto
```diff
@@ -23,20 +23,37 @@ message TemplateCreateRequest {
...
+}
+
 // Logs from template build
-message TemplateBuildLog {
-  string log = 1;
+message TemplateBuildStatusResponse {
+  bool done = 1;
+  bool failed = 2;
```
jakubno: Does it make sense to have to separate fields, there are 4 possible and one of them is probably invalid, I think enum would make better sense here

jakubno: I think failed is not even really used

packages/shared/pkg/db/envs.go
```diff
@@ -161,6 +161,63 @@ func (db *DB) GetEnv(ctx context.Context, aliasOrEnvID string) (result *Template
...
+		Query().
+		Where(
+			envbuild.And(
+				envbuild.EnvID(envID),
+				envbuild.StatusIn(envbuild.StatusWaiting, envbuild.StatusBuilding),
+				envbuild.IDNotIn(excludedBuildUUIDs...),
+			),
+		).
+		Order(models.Desc(envbuild.FieldCreatedAt)).
+		Limit(10).
```
jakubno: `First` already limits the query 
```suggestion
```

packages/shared/pkg/db/envs.go
```diff
@@ -161,6 +161,63 @@ func (db *DB) GetEnv(ctx context.Context, aliasOrEnvID string) (result *Template
...
+	return dbBuild, nil
+}
+
+func (db *DB) GetEnvBuild(ctx context.Context, buildID uuid.UUID) (build *models.EnvBuild, err error) {
+	dbBuild, err := db.
+		Client.
+		EnvBuild.
+		Query().
+		Where(envbuild.ID(buildID)).
+		Limit(1).
```
jakubno:
```suggestion
```

packages/shared/pkg/db/envs.go
```diff
@@ -161,6 +161,63 @@ func (db *DB) GetEnv(ctx context.Context, aliasOrEnvID string) (result *Template
...
+		First(ctx)
+
+	notFound := models.IsNotFound(err)
```
jakubno, line 215: you should return different errors, so you can return proper status code (you are returning 404 for db error)

packages/api/internal/template-manager/template_manager.go
```diff
@@ -1,20 +1,172 @@
...
+	db         *db.DB
+	lock       sync.Mutex
+	processing map[uuid.UUID]struct{ templateID string }
 }
 
-func New() (*TemplateManager, error) {
+var (
+	syncInterval             = time.Minute * 1
+	syncTimeout              = time.Minute * 5
+	syncWaitingStateDeadline = time.Minute * 10
```
jakubno: do 20, the build + push can take sometimes longer

packages/api/internal/handlers/template_start_build.go
```diff
@@ -99,82 +99,72 @@ func (a *APIStore) PostTemplatesTemplateIDBuildsBuildID(c *gin.Context, template
...
 
-	// Set the build status to building
-	err = a.db.EnvBuildSetStatus(ctx, envDB.ID, buildUUID, envbuild.StatusBuilding)
-	if err != nil {
-		err = fmt.Errorf("error when setting build status: %w", err)
-		telemetry.ReportCriticalError(ctx, err)
+	// Trigger the build in the background
+	buildContext, childSpan := a.Tracer.Start(
+		trace.ContextWithSpanContext(context.Background(), span.SpanContext()),
+		"background-build-env",
```
jakubno: this isn't in background anymore, i think this should be in BuildStatusSync routine

packages/api/internal/handlers/template_start_build.go
```diff
@@ -99,82 +99,72 @@ func (a *APIStore) PostTemplatesTemplateIDBuildsBuildID(c *gin.Context, template
...
-		defer childSpan.End()
-
-		startTime := time.Now()
-		build := envDB.Edges.Builds[0]
-		startCmd := ""
-		if build.StartCmd != nil {
-			startCmd = *build.StartCmd
-		}
+	// status building must be set after build is triggered because then
+	// it's possible build status job will be triggered before build cache on template manager is created and build will fail
```
jakubno, line 159: Not sure if I understand this correctly, is there a chance a build will fail now? also read the comment above

packages/api/internal/template-manager/template_manager.go
```diff
@@ -1,20 +1,172 @@
...
+	tm.lock.Lock()
+	defer tm.lock.Unlock()
+
+	_, exists := tm.processing[buildID]
+	if exists {
+		// already in processing queue, skip
+		return true
+	}
+
+	tm.processing[buildID] = struct{ templateID string }{templateID: templateID}
```
jakubno: could you do actual type from the struct

packages/template-manager/internal/build/builder.go
```diff
@@ -0,0 +1,139 @@
...
+func (b *TemplateBuilder) Builder(ctx context.Context, template *Env, envID string, buildID string) error {
+	buildStorage := b.templateStorage.NewBuild(template.TemplateFiles)
+	buildEntry, err := b.buildCache.Get(buildID)
+	if err != nil {
+		return err
+	}
+
+	logsWriter := template.BuildLogsWriter
+
+	// Remove local template files if build fails
```
jakubno: this will remove then even when the build succeeds right? I know it's just copied, but please correct it 🙏🏻

packages/template-manager/internal/build/builder.go
```diff
@@ -0,0 +1,139 @@
...
+		telemetry.ReportCriticalError(ctx, err)
+
+		buildStateErr := b.buildCache.SetFailed(envID, buildID)
+		if buildStateErr != nil {
+			b.logger.Error("Error while setting build state to failed", zap.Error(buildStateErr))
+			telemetry.ReportError(ctx, buildStateErr)
+		}
+	}
+
+	// Remove build files if build fails or times out
```
jakubno: also here

packages/template-manager/internal/build/builder.go
```diff
@@ -0,0 +1,139 @@
...
+		tracer:             tracer,
+		buildCache:         buildCache,
+		buildLogger:        buildLogger,
+		dockerClient:       dockerClient,
+		legacyDockerClient: legacyDockerClient,
+		templateStorage:    templateStorage,
+	}
+}
+
+func (b *TemplateBuilder) Builder(ctx context.Context, template *Env, envID string, buildID string) error {
```
jakubno: It's a build, it doesn't return any Builder or anything right?

packages/template-manager/internal/build/builder.go
```diff
@@ -0,0 +1,139 @@
...
+		if buildStateErr != nil {
+			b.logger.Error("Error while setting build state to failed", zap.Error(buildStateErr))
+			telemetry.ReportError(ctx, buildStateErr)
+		}
+	}
+
+	buildEntry.SetEnvdVersionKey(strings.TrimSpace(string(out)))
+	buildEntry.SetRootFsSizeKey(int32(template.RootfsSizeMB()))
+
+	err = b.buildCache.SetSucceeded(envID, buildID)
```
jakubno: I think it would make sense to add envd version and size here, it doesn't make sense to set it to succeeded without it

jakubno (changes requested):
Could you add contexts instead of background on relevant places to handle the shutdown

packages/template-manager/internal/server/template_status.go
sitole: It's timeout for status check call handler. Not sure if we should increase limit for call here.
Maybe just softly return and end stream so client can re-subscribe?

packages/template-manager/internal/build/builder.go
sitole: first part with disk cleanup is okay, changed typo. But here, we really want to remove templates from bucket ONLY when build failed, right? When succeed should stay there.

packages/api/internal/handlers/template_start_build.go
sitole: > what if there's an error

Implemented custom err, checked for it and returns proper http error

> Also I think you should check and create a new build in one transaction if we want to make sure there are no 2 concurrent builds.

Actually build is here already created. It's in previous request that initializes requests new template build.

packages/api/internal/handlers/template_start_build.go
sitole: Can fail because there is background sync checking for statuses of build in  `building` pase.

This ensures pipeline will be following:
- trigger template-manager and initialize build cache here
- set phase to `building`
- async periodical sync will find it and try to check status (cache already exists in template-manager)

When we will set it to `building` phase before `a.templateManager.CreateTemplate` is called:
- set phase to `building`
- async periodical sync can run here, try to check status of building and fail because state is not template-manager yet  

not sure if this is your concern?

packages/template-manager/internal/server/template_status.go
jakubno: Yes, but the status is running during the build (probably more correct would be call it postprocessing). I think it should be increased. I just want to make sure, if there's a long build, it will succeed and user won't see any error

packages/template-manager/internal/build/builder.go
jakubno: yes, you are right!

packages/api/internal/handlers/template_build_logs.go
```diff
@@ -46,27 +48,21 @@ func (a *APIStore) GetTemplatesTemplateIDBuildsBuildIDStatus(c *gin.Context, tem
...
 		errMsg := fmt.Errorf("error when parsing build id: %w", err)
 		telemetry.ReportError(ctx, errMsg)
-
 		a.sendAPIStoreError(c, http.StatusBadRequest, "Invalid build id")
-
 		return
 	}
 
-	dockerBuild, err := a.buildCache.Get(templateID, buildUUID)
+	templateDB, _, err := a.db.GetEnv(ctx, templateID)
```
jakubno: sorry for spotting it now, but i think it could make sense to keep the cache (as an actual cache), probably with lower TTL. There's 2 db request every call, it's probably unnecessary. Also once the build existed, there's no reason why you should be able to query logs, so it's fairly safe (I would still keep some mechanism to recheck DB as we do e.g. for auth)

packages/api/internal/handlers/template_start_build.go
```diff
@@ -88,93 +90,88 @@ func (a *APIStore) PostTemplatesTemplateIDBuildsBuildID(c *gin.Context, template
...
 
-	// Create a new build cache for storing logs
-	err = a.buildCache.Create(templateID, buildUUID, team.ID)
-	if err != nil {
-		a.sendAPIStoreError(c, http.StatusConflict, fmt.Sprintf("there's already running build for %s", templateID))
+	// Check if some other build from same template is in process, excludes currently created build
+	envBuildDB, err := a.db.GetRunningEnvBuild(ctx, envDB.ID, []uuid.UUID{buildUUID})
+
+	// error is present but different from not found
+	if err != nil && !errors.Is(err, db.TemplateBuildNotFound{}) {
```
jakubno: This is hard to read and I think you even confused yourself, the second if branch is only satisfied if
```
errors.Is(err, db.TemplateBuildNotFound{})
```
and the comment is `we found conflicting build for same env`

packages/api/internal/handlers/template_start_build.go
jakubno: Let's forget about this, there small issue, nothing serious, but we should add check that the status is `waiting`, otherwise you could trigger multiple concurrent, start build for failed / already succeed

packages/api/internal/handlers/template_start_build.go
```diff
@@ -88,93 +90,88 @@ func (a *APIStore) PostTemplatesTemplateIDBuildsBuildID(c *gin.Context, template
...
-			cacheErr := a.buildCache.SetDone(templateID, buildUUID, api.TemplateBuildStatusError)
-			if cacheErr != nil {
-				telemetry.ReportCriticalError(buildContext, fmt.Errorf("error when setting build done in logs: %w", cacheErr))
-			}
-
-			return
-		}
+	// Do not wait for global build sync trigger it immediately
+	go func() {
+		a.templateManager.BuildStatusSync(context.Background(), buildUUID, templateID)
```
jakubno: i am still missing the background context we used before (I mentioned it on the other place you were using it)

packages/api/internal/template-manager/create_template.go
```diff
@@ -35,12 +25,14 @@ func (tm *TemplateManager) CreateTemplate(
 	diskSizeMB,
 	memoryMB int64,
 ) error {
-	childCtx, childSpan := t.Start(ctx, "create-template",
+	ctx, ctxSpan := t.Start(
+		trace.ContextWithSpanContext(context.Background(), span),
+		"background-build-env",
```
jakubno: I don't think this is the background task either, the original code was correct

packages/api/internal/template-manager/create_template.go
```diff
@@ -64,52 +54,12 @@ func (tm *TemplateManager) CreateTemplate(
...
-	}
-
-	telemetry.ReportEvent(childCtx, "created new environment", attribute.String("env.id", templateID))
-
-	cacheErr := buildCache.SetDone(templateID, buildID, api.TemplateBuildStatusReady)
-	if cacheErr != nil {
-		err = fmt.Errorf("error when setting build done in logs: %w", cacheErr)
-		telemetry.ReportCriticalError(childCtx, cacheErr)
+	if utils.UnwrapGRPCError(err) != nil {
+		return fmt.Errorf("failed to create template '%s': %w", templateID, err)
```
jakubno, line 58: we are using the unwrapped error at the other places, is there a reason why not to used it here

packages/api/internal/template-manager/template_manager.go
```diff
@@ -1,20 +1,177 @@
...
+func (tm *TemplateManager) BuildsStatusPeriodicalSync(ctx context.Context) {
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-time.After(syncInterval):
+			buildsRunning, err := tm.db.GetRunningEnvBuilds(ctx)
+			if err != nil {
+				zap.L().Error("Error getting running builds for periodical sync", zap.Error(err))
+				time.Sleep(time.Second) // just not hammer database if there is an error
```
jakubno: How could it hammer database? Wouldn't it be waiting for `syncInterval`? Also is there a reason why to use time.After over ticker?

Could you also add timeout for database query?

packages/api/internal/template-manager/template_manager.go
```diff
@@ -1,20 +1,177 @@
...
+
+			zap.L().Info("Running periodical sync of builds statuses", zap.Int("count", len(buildsRunning)))
+			for _, buildDB := range buildsRunning {
+				go tm.BuildStatusSync(ctx, buildDB.ID, *buildDB.EnvID)
+			}
+		}
+	}
+}
+
+// BuildStatusSync todo: implement failing build when there is build in some state for too long
```
jakubno: There's still a TODO

packages/shared/pkg/db/envs.go
```diff
@@ -161,6 +161,64 @@ func (db *DB) GetEnv(ctx context.Context, aliasOrEnvID string) (result *Template
...
+	return envBuilds, nil
+}
+
+func (db *DB) GetRunningEnvBuild(ctx context.Context, envID string, excludedBuildUUIDs []uuid.UUID) (build *models.EnvBuild, err error) {
+	dbBuild, err := db.
+		Client.
+		EnvBuild.
+		Query().
+		Where(
+			envbuild.And(
```
jakubno: note: it's AND by default

packages/template-manager/internal/cache/build_cache.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+func (c *BuildCache) updateCounter(envID string, buildID string, value int64) {
+	c.counter.Add(context.Background(), value,
+		metric.WithAttributes(attribute.String("env_id", envID)),
+		metric.WithAttributes(attribute.String("build_id", buildID)),
+	)
+}
+
+func (c *BuildCache) Delete(envID string, buildID string) {
```
jakubno, line 157: i think delete should be locking

packages/template-manager/internal/server/create_template.go
```diff
@@ -20,16 +15,11 @@ import (
...
 )
 
-const cleanupTimeout = time.Second * 10
-
-func (s *serverStore) TemplateCreate(templateRequest *template_manager.TemplateCreateRequest, stream template_manager.TemplateService_TemplateCreateServer) error {
-	ctx := stream.Context()
-
-	childCtx, childSpan := s.tracer.Start(ctx, "template-create")
+func (s *serverStore) TemplateCreate(ctx context.Context, templateRequest *template_manager.TemplateCreateRequest) (*emptypb.Empty, error) {
+	childCtx, childSpan := s.tracer.Start(context.Background(), "template-create")
```
jakubno: why background?

jakubno (changes requested):
Have you tried to roll api while building? You can roll api easily by adding env variable in `nomad/api.hcl` definition

packages/api/internal/handlers/template_start_build.go
sitole: as discussed, i added check that build must be in waiting state.

packages/template-manager/internal/server/template_status.go
sitole: ok makes sense. Increased to 10 minutes, same timeout for whole template-manager to build. Its triggered after build is confirmed so should be enough.

packages/api/internal/handlers/template_start_build.go
sitole: yea, i changed it and make it more readable

packages/api/internal/template-manager/template_manager.go
sitole: removed todo. Building stat will fail when requesting template manager. If build is too old it will be missed in cache so it means will return. Also there is timeout for build on time-manager side along with ttl cache.

packages/api/internal/template-manager/template_manager.go
sitole: > How could it hammer database?

Its not possible with this type of calls. Before there was just inifinite `for` with sleep but i re-implement it to support context cancel so its no more needed.

> Could you also add timeout for database query?

Done. Not sure about right amount for explicit value for timeout here but 5s should work for now.

packages/api/internal/template-manager/create_template.go
sitole: dont know why i used it here. fixed. ctx is propagated from parent, `create-template` span is started here

packages/api/internal/handlers/template_start_build.go
sitole: done, so span is extended into child

packages/api/internal/cache/templates/cache.go
```diff
@@ -119,3 +120,70 @@ func (c *TemplateCache) Get(ctx context.Context, aliasOrEnvID string, teamID uui
...
+	return &TemplatesBuildCache{
+		cache: cache,
+		db:    db,
+	}
+}
+
+func (c *TemplatesBuildCache) Get(ctx context.Context, buildID uuid.UUID, templateID string) (*TemplateBuildInfo, error) {
+	item := c.cache.Get(buildID)
+	if item == nil {
+		print("build cache: not present in cache")
```
jakubno: use proper logging

packages/api/internal/cache/templates/cache.go
```diff
@@ -119,3 +120,70 @@ func (c *TemplateCache) Get(ctx context.Context, aliasOrEnvID string, teamID uui
...
+
+		envDB, _, envDBErr := c.db.GetEnv(ctx, templateID)
```
jakubno: Can you create a new function, there will be overhead because of build which you are not using, also this won't work for a newly created templates, because there isn't any build with status `Uploaded`

packages/api/internal/cache/templates/cache.go
```diff
@@ -119,3 +120,70 @@ func (c *TemplateCache) Get(ctx context.Context, aliasOrEnvID string, teamID uui
...
+				TeamID:     envDB.TeamID,
+				TemplateID: envDB.TemplateID,
+			},
+			templateInfoExpiration,
+		)
+
+		return item.Value(), nil
+	}
+
+	print("build cache: hitting")
```
jakubno: also here

packages/api/internal/cache/templates/cache.go
```diff
@@ -119,3 +120,70 @@ func (c *TemplateCache) Get(ctx context.Context, aliasOrEnvID string, teamID uui
...
+		if envDBErr != nil {
+			if errors.Is(envDBErr, db.TemplateNotFound{}) {
+				return nil, TemplateBuildInfoNotFound{}
+			}
+
+			return nil, fmt.Errorf("failed to get template '%s': %w", buildID, envDBErr)
+		}
+
+		// making sure associated template build really exists
+		_, envBuildDBErr := c.db.GetEnvBuild(ctx, buildID)
```
jakubno: NIT: you can do `Exists()`

packages/template-manager/internal/server/template_status.go
```diff
@@ -0,0 +1,57 @@
...
+			telemetry.ReportError(ctx, err)
+			return err
+		}
+
+		if !buildInfo.IsRunning() {
+			logger.Info("Template build finished")
+			return nil
+		}
+
+		time.Sleep(time.Second * 5)
```
jakubno: I think we can send the status every second

packages/api/internal/handlers/template_build_logs.go
```diff
@@ -75,13 +81,23 @@ func (a *APIStore) GetTemplatesTemplateIDBuildsBuildIDStatus(c *gin.Context, tem
...
 		msg := fmt.Errorf("user doesn't have access to env '%s'", templateID)
 		telemetry.ReportError(ctx, msg)
-
 		a.sendAPIStoreError(c, http.StatusForbidden, fmt.Sprintf("You don't have access to this sandbox template (%s)", templateID))
-
 		return
 	}
 
-	status := dockerBuild.GetStatus()
+	buildDB, err := a.db.GetEnvBuild(ctx, buildUUID)
```
jakubno, line 88: why is there still db query, if you do the same in the cache

jakubno: I think this may not be needed, you could invalidate cache from `BuildStatusSync`, but let's keep it simple for now

sitole:
@jakubno done. I tried to re-deploy api service in middle of template build and works nice, logs still streams until finish.

packages/api/internal/cache/templates/cache.go
```diff
@@ -119,3 +123,87 @@ func (c *TemplateCache) Get(ctx context.Context, aliasOrEnvID string, teamID uui
...
+type TemplateBuildInfoNotFound struct{ error }
+
+func (TemplateBuildInfoNotFound) Error() string {
+	return "Template build info not found"
+}
+
+type TemplatesBuildCache struct {
+	cache *ttlcache.Cache[uuid.UUID, *TemplateBuildInfo]
+	db    *db.DB
+	mx    sync.Mutex
```
ValentaTomas, line 142: Nitpick: I think we are using `mu` for mutexes everywhere in the codebase. But this might have been just copied.

packages/template-manager/template-manager.proto
```diff
@@ -23,20 +23,41 @@ message TemplateCreateRequest {
...
+  TemplateBuildMetadata metadata = 2;
 }
 
 // Interface exported by the server.
 service TemplateService {
   // TemplateCreate is a gRPC service that creates a new template
-  rpc TemplateCreate (TemplateCreateRequest) returns (stream TemplateBuildLog);
+  rpc TemplateCreate (TemplateCreateRequest) returns (google.protobuf.Empty);
+
+  // TemplateStatus is a gRPC service that streams the status of a template build
```
ValentaTomas, line 58: If the retrieval of build status/logs works the same way between CLI<->API as right now (polling) won't this break if we have multiple APIs running as the streamed info will be in only one of the API caches?

packages/template-manager/internal/server/template_status.go
```diff
@@ -0,0 +1,57 @@
...
+
+		if !buildInfo.IsRunning() {
+			logger.Info("Template build finished")
+			return nil
+		}
+
+		time.Sleep(time.Second)
+	}
+
+	return nil
```
ValentaTomas: Unreachable

packages/template-manager/internal/server/template_status.go
```diff
@@ -0,0 +1,57 @@
...
+	ctx, cancel := context.WithTimeout(stream.Context(), statusTimeout)
+	defer cancel()
+
+	ctx, ctxSpan := s.tracer.Start(ctx, "template-build-status-request")
+	defer ctxSpan.End()
+
+	logger := s.logger.With(zap.String("buildID", in.BuildID), zap.String("envID", in.TemplateID))
+
+	// stream updated status until build is done or timeout will kick in
+	for {
```
ValentaTomas: This loop could be checking for context done to exit in case the request ended, but:
- This is probably already covered by the call to `stream.Send` implicitly
- Not sure if the stream makes sense here (problem with multiple APIs mentioned in the other commend)

packages/template-manager/template-manager.proto
jakubno: You can connect with multiple APIs

packages/template-manager/internal/server/template_status.go
jakubno: But also true, there probably isn't a reason to have it as a stream if we have 5 second sleep in the stream

packages/template-manager/internal/server/template_status.go
```diff
@@ -0,0 +1,34 @@
...
+	defer ctxSpan.End()
+
+	logger := s.logger.With(zap.String("buildID", in.BuildID), zap.String("envID", in.TemplateID))
+
+	buildInfo, err := s.buildCache.Get(in.BuildID)
+	if err != nil {
+		return nil, fmt.Errorf("error while getting build info, maybe already expired")
+	}
+
+	if buildInfo.IsFailed() {
```
jakubno: does it make sense to return an error here? you have status error for that right?

packages/template-manager/internal/server/template_status.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"context"
+	"fmt"
+	"github.com/e2b-dev/infra/packages/shared/pkg/grpc/template-manager"
+	"go.uber.org/zap"
+)
+
+func (s *serverStore) TemplateBuildStatus(ctx context.Context, in *template_manager.TemplateStatusRequest) (*template_manager.TemplateBuildStatusResponse, error) {
+
+	// todo
```
jakubno: todo what?

jakubno (changes requested):

packages/api/internal/template-manager/template_manager.go
```diff
@@ -1,20 +1,187 @@
...
+				logger.Error("Error when setting build status", zap.Error(dbErr))
+			}
+			return
+		}
+	}
+
+	ticker := time.NewTicker(time.Second)
+	defer ticker.Stop()
+
+	// stream build status
```
jakubno: not stream anymore
```suggestion
	// poll build status
```

jakubno (approved):

PR #381, opened 2025-03-06
Kill to handle subsequent resumes

mishushakov:
_see Linear_


mishushakov:
Let me know what changes are required for logging/tracing?

packages/api/internal/handlers/sandbox_kill.go
```diff
@@ -36,60 +73,51 @@ func (a *APIStore) DeleteSandboxesSandboxID(
...
 			return
 		}
 
-		found := a.orchestrator.DeleteInstance(ctx, sandboxID, false)
-		if !found {
+		// remove running sandbox from the orchestrator
+		sandboxExists := a.orchestrator.DeleteInstance(ctx, sandboxID, false)
+		if !sandboxExists {
+      telemetry.ReportError(ctx, fmt.Errorf("sandbox '%s' not found", sandboxID))
 			a.sendAPIStoreError(c, http.StatusNotFound, fmt.Sprintf("Error killing sandbox - sandbox '%s' was not found", sandboxID))
```
dobrac:
```suggestion
			a.sendAPIStoreError(c, http.StatusNotFound, fmt.Sprintf("Error deleting sandbox - sandbox '%s' was not found", sandboxID))
```

r33drichards:
lgtm! as @jakubno said an integration test would be helpful and then be happy to ✅

mishushakov:
Keep in mind: tests for paused will pass currently, as Sandboxes.list does not include paused sandboxes (wip)

packages/api/internal/handlers/sandbox_kill.go
```diff
@@ -14,6 +15,42 @@ import (
...
+	}
+
+	dbErr := a.db.DeleteEnv(ctx, env.ID)
+	if dbErr != nil {
+		return nil, fmt.Errorf("error deleting env from db: %w", dbErr)
+	}
+
+	a.templateCache.Invalidate(env.ID)
+
+	deleteJobErr := a.templateManager.DeleteBuilds(ctx, envBuildIDs)
```
dobrac: is the order of these two changed for any reason?

packages/api/internal/handlers/sandbox_kill.go
```diff
@@ -14,6 +15,42 @@ import (
...
+	teamID uuid.UUID,
+) (*bool, error) {
+	env, builds, err := a.db.GetSnapshotBuilds(ctx, sandboxID, teamID)
+	if err != nil {
+		return nil, fmt.Errorf("error getting env builds for sandbox '%s': %w", sandboxID, err)
+	}
+
+	var found bool = true
+	if env == nil {
+		found = false
```
dobrac: is this variable necessary?

packages/api/internal/handlers/sandbox_kill.go
```diff
@@ -14,6 +15,42 @@ import (
 	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
 )
 
+func (a *APIStore) deleteSnapshot(
+	ctx context.Context,
+	sandboxID string,
+	teamID uuid.UUID,
+) (*bool, error) {
```
dobrac: why to return *bool ?

packages/api/internal/handlers/sandbox_kill.go
```diff
@@ -36,60 +73,51 @@ func (a *APIStore) DeleteSandboxesSandboxID(
...
 		return
 	}
 
-	dbErr := a.db.DeleteEnv(ctx, env.ID)
-	if dbErr != nil {
-		errMsg := fmt.Errorf("error when deleting env from db: %w", dbErr)
-		telemetry.ReportCriticalError(ctx, errMsg)
+	// remove any snapshots when the sandbox is not running
+	snapshotExists, deleteSnapshotErr := a.deleteSnapshot(ctx, sandboxID, teamID)
+	if snapshotExists != nil && !*snapshotExists {
```
dobrac: if you need to distinguish between error codes, you should return custom error codes and them check them with error.Is

packages/api/internal/handlers/sandbox_kill.go
```diff
@@ -36,60 +73,51 @@ func (a *APIStore) DeleteSandboxesSandboxID(
...
+		telemetry.ReportEvent(ctx, "deleted sandbox from orchestrator")
 
-	env, builds, err := a.db.GetSnapshotBuilds(ctx, sandboxID, teamID)
-	if err != nil {
-		telemetry.ReportError(ctx, fmt.Errorf("error when getting paused sandbox from db: %w", err))
-		a.sendAPIStoreError(c, http.StatusNotFound, fmt.Sprintf("Error killing sandbox - sandbox '%s' was not found", sandboxID))
+		c.Status(http.StatusNoContent)
 
 		return
 	}
```
dobrac, line 101: can the logic of those two branches be combined to one deleteSnapshots call?

packages/api/internal/handlers/sandbox_kill.go
mishushakov: The order doesn't matter here I think, but have moved it to match template_delete.go

packages/api/internal/handlers/sandbox_kill.go
mishushakov: not really, there are 4 possible cases:

1. Deleting a running sandbox with no snapshots (should not cause any error)
2. Deleting a running sandbox with past snapshots (should delete the sandbox and the snapshots and report if errored)
3. Deleting a paused sandbox with past snapshots (should delete snapshots and report if errored)
4. Deleting a non-existing sandbox with no past snapshots (should return 404)

packages/api/internal/handlers/sandbox_kill.go
```diff
@@ -11,9 +13,44 @@ import (
...
+	sandboxID string,
+	teamID uuid.UUID,
+) (bool, error) {
+	env, builds, err := a.db.GetSnapshotBuilds(ctx, sandboxID, teamID)
+	if err != nil {
+		if errors.Is(err, db.EnvNotFound{}) {
+			return false, nil
+		}
+
+		return true, fmt.Errorf("error getting env builds for sandbox '%s': %w", sandboxID, err)
```
dobrac: shouldn't the default here be false to emphasize the operation failed?

packages/api/internal/handlers/sandbox_kill.go
```diff
@@ -36,60 +73,51 @@ func (a *APIStore) DeleteSandboxesSandboxID(
...
 		return
 	}
 
-	dbErr := a.db.DeleteEnv(ctx, env.ID)
-	if dbErr != nil {
-		errMsg := fmt.Errorf("error when deleting env from db: %w", dbErr)
-		telemetry.ReportCriticalError(ctx, errMsg)
+	// remove any snapshots when the sandbox is not running
+	snapshotExists, deleteSnapshotErr := a.deleteSnapshot(ctx, sandboxID, teamID)
+	if !snapshotExists {
```
dobrac: I think it's confusing to check the error as second. Most of the methods I saw are done in a way that when error is returned, the other value is "undefined" (or returns some state when the output of the function is actually in the args)

if you want to distinguish between different type of errors, I would still propose the errors.Is for that. You can check the error type here before returning other error to the user.

packages/api/internal/handlers/sandbox_kill.go
dobrac: I see, the whole IF branch is quite confusing imo

packages/shared/pkg/db/snapshot.go
```diff
@@ -182,20 +182,17 @@ func (db *DB) GetSnapshotBuilds(ctx context.Context, sandboxID string, teamID uu
...
-	if err != nil {
-		return nil, nil, fmt.Errorf("failed to get snapshot build for '%s': %w", sandboxID, err)
+	if notFound {
+		return nil, nil, EnvNotFound{}
 	}
 
-	if notFound {
-		return nil, nil, fmt.Errorf("no snapshot build found for '%s'", sandboxID)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to get snapshot build for '%s': %w", teamID, err)
```
jakubno: why is team id here?

packages/orchestrator/cmd/mock-snapshot/mock.go
```diff
@@ -154,7 +154,7 @@ func mockSnapshot(
 		time.Now(),
 		false,
 		templateId,
-		"test-client",
+		"testclient",
```
jakubno, line 160: what was the reason for renaming?

mishushakov: because the the way it would split the ClientID:

```
Invalid sandbox ID — missing client ID part: i8t9idy79c8bg1axqoc5o-test-client
```

https://github.com/e2b-dev/infra/actions/runs/13767868208/job/38498819765#step:7:307

packages/shared/pkg/db/snapshot.go
mishushakov: got Cusor'd again

mishushakov: it's second time it's happening this week, it overwrites my error messages for some reason

dobrac (approved):

PR #382, opened 2025-03-06
Add media bucket and dashboard service account infrastructure

ben-fornefeld:
This PR adds the necessary GCP infrastructure for media storage and dashboard operations:

### Changes
- Created a new media bucket with CORS configuration and controlled public access
- Added a custom IAM role for limited object access (get only)
- Created a dashboard service account with appropriate permissions
- Set up secret management for the dashboard service account key
- Added necessary IAM bindings for service accounts

### Why
These changes enable secure media storage with controlled public access and establish the service account infrastructure needed for dashboard operations. The media bucket will be used for storing user-accessible content, while the dashboard service account provides the necessary permissions for the dashboard to interact with GCP resources.

In the upcoming version of the dashboard, the media bucket will be used for storing team profile pictures.

### Testing
- Verified bucket creation and permissions in GCP console
- Confirmed service account creation and key management

### DIFF
--- a/packages/buckets/main.tf
+++ b/packages/buckets/main.tf
@@ -100,16 +100,92 @@ resource "google_storage_bucket" "fc_template_bucket" {
   labels = var.labels
 }
 
+resource "google_storage_bucket" "media_bucket" {
+  location = var.gcp_region
+  name     = "${var.gcp_project_id}-media"
+
+  public_access_prevention    = "inherited"
+  storage_class               = "STANDARD"
+  uniform_bucket_level_access = true
+
+  cors {
+    origin          = ["*"]
+    method          = ["GET", "HEAD", "OPTIONS"]
+    response_header = ["Content-Type", "Access-Control-Allow-Origin"]
+    max_age_seconds = 3600
+  }
+
+  labels = var.labels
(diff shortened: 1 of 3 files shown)


ValentaTomas:
Quick question here:
- Isn't the supabase storage enough?
- I'm not sure this should be in infra repo if this concerns just dashboard

ben-fornefeld:
> Quick question here:
> 
> - Isn't the supabase storage enough?
> 
> - I'm not sure this should be in infra repo if this concerns just dashboard

I implemented the feature via supabase bucket at first, then consulted with @jakubno & we came to the conclusion that since we had plans to move away from supabase eventually, and this is storage, it makes sense to have this bucket defined in our infrastructure instead. 

@jakubno What are other reasons you preferred it this way?

r33drichards:
> I'm not sure this should be in infra repo if this concerns just dashboard

+1 this is a great contribution but might be extra overhead for someone looking to self host, but i think this would work well in its own repo!

ben-fornefeld:
i had a talk with @jakubno & we concluded that having the bucket in supabase makes more sense at the moment

PR #384, opened 2025-03-06
Fix cache eviction and sync node synchronization

dobrac:
Fixes synchronization issue between orchestrator sandboxes state and API cache state. This change will check in the exists method also for items that are being evicted, so they won't be added while the eviction is happening.

### DIFF
--- a/packages/api/internal/cache/instance/lifecycle_cache.go
+++ b/packages/api/internal/cache/instance/lifecycle_cache.go
@@ -18,15 +18,19 @@ type lifecycleCacheMetrics struct {
 
 type lifecycleCache[T LifecycleCacheItem] struct {
 	running cmap.ConcurrentMap[string, T]
-	onEvict func(ctx context.Context, instance T)
+	// evicting is used to track items in the process of evicting.
+	// This is to allow checking if an item is still in the process as that might take some time.
+	evicting cmap.ConcurrentMap[string, bool]
+	onEvict  func(ctx context.Context, instance T)
 
 	metrics lifecycleCacheMetrics
 }
 
 func newLifecycleCache[T LifecycleCacheItem]() *lifecycleCache[T] {
 	return &lifecycleCache[T]{
-		running: cmap.New[T](),
-		onEvict: func(ctx context.Context, instance T) {},
+		running:  cmap.New[T](),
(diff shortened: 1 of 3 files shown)


packages/api/internal/cache/instance/lifecycle_cache.go
```diff
@@ -58,7 +62,11 @@ func (c *lifecycleCache[T]) Start(ctx context.Context) {
 							return false
 						}
 
-						go c.onEvict(ctx, value)
+						c.evicting.Set(key, false)
```
jakubno: maybe use `struct{}` instead of `bool`, setting evicting to `false` while it's evicting is little bit confusing

packages/api/internal/cache/instance/lifecycle_cache.go
```diff
@@ -74,6 +82,21 @@ func (c *lifecycleCache[T]) SetIfAbsent(key string, value T) bool {
 	return c.running.SetIfAbsent(key, value)
 }
 
+func (c *lifecycleCache[T]) Has(key string, includeExpired bool) bool {
```
jakubno, line 85: what's `includeExpired` good for, it's used only in `Exists` with `false`

dobrac: it should be used in Exists with true

packages/api/internal/cache/instance/lifecycle_cache_test.go
```diff
@@ -125,6 +125,72 @@ func TestLifecycleCacheLen(t *testing.T) {
...
+
+	// Wait for the eviction process to start but not complete
+	time.Sleep(200 * time.Millisecond)
+
+	assert.True(t, cache.Has("test", true))
+	assert.False(t, cache.Has("test", false))
+
+	// Wait for eviction to complete
+	<-evictCalled
+	time.Sleep(100 * time.Millisecond)
```
jakubno: is the sleep here needed?

jakubno (approved):

packages/api/internal/cache/instance/lifecycle_cache_test.go
dobrac: removed

PR #385, opened 2025-03-06
Add indexes to DB for frequently used columns

jakubno:


packages/shared/migrations/20250306105106.sql
```diff
@@ -0,0 +1,7 @@
+CREATE INDEX CONCURRENTLY idx_envs_builds_envs ON public.env_builds (env_id);
+CREATE INDEX CONCURRENTLY idx_envs_envs_aliases ON public.env_aliases (env_id);
+CREATE INDEX CONCURRENTLY ids_users_access_tokens ON public.access_tokens (user_id);
```
dobrac:
```suggestion
CREATE INDEX CONCURRENTLY idx_users_access_tokens ON public.access_tokens (user_id);
```

dobrac (approved):

PR #392, opened 2025-03-07
support for envd cli to print out source commit sha

sitole:
Exact commit sha of sourced version will help with better debugging


jakubno (approved):
Thanks, this will be super useful!

PR #395, opened 2025-03-07
fixed empty reader buffer after second try of calling envd init endpoint

sitole:
there was bug when reader is provided to request loop that is trying to initialize envd but because its not running yet, some requests can fails. Reader buffer provided with JSON request body was read in first round that failed and later it was empty so envd service initialized with empty body, that means without environments.

### DIFF
--- a/packages/orchestrator/internal/sandbox/envd.go
+++ b/packages/orchestrator/internal/sandbox/envd.go
@@ -24,8 +24,7 @@ const (
 func doRequestWithInfiniteRetries(ctx context.Context, method, address string, requestBody []byte) (*http.Response, error) {
 	for {
 		reqCtx, cancel := context.WithTimeout(ctx, requestTimeout)
-		reqBody := bytes.NewReader(requestBody)
-		request, err := http.NewRequestWithContext(reqCtx, method, address, reqBody)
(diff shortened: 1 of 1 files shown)


packages/orchestrator/internal/sandbox/envd.go
```diff
@@ -21,10 +21,12 @@ const (
 
 // doRequestWithInfiniteRetries does a request with infinite retries until the context is done.
 // The parent context should have a deadline or a timeout.
-func doRequestWithInfiniteRetries(ctx context.Context, method, address string, body io.Reader) (*http.Response, error) {
+func doRequestWithInfiniteRetries(ctx context.Context, method, address string, requestBody []byte) (*http.Response, error) {
 	for {
 		reqCtx, cancel := context.WithTimeout(ctx, requestTimeout)
-		request, err := http.NewRequestWithContext(reqCtx, method, address, body)
+		reqBody := bytes.NewReader(requestBody)
+		request, err := http.NewRequestWithContext(reqCtx, method, address, reqBody)
```
jakubno:
```suggestion
		request, err := http.NewRequestWithContext(reqCtx, method, address, bytes.NewReader(requestBody))
```

jakubno (approved):

PR #399, opened 2025-03-08
Use wget instead of ping to test internet access to avoid installing package in test, also include timeout

r33drichards:


r33drichards:
working https://github.com/e2b-dev/infra/actions/runs/13774978216

ValentaTomas (approved):

PR #406, opened 2025-03-10
Add random service instance id to distinguish between new deploys

jakubno:
# Description

This should be fix issues in grafana, when there is rolling update 
image


dobrac (approved):
would be nicer to be commitHash or something

jakubno:
I can also add commit hash, but commit hash won't be unique for example for restarts

PR #408, opened 2025-03-10
Improve integration tests error when orchestrator or api fails during start

dobrac:
Error out when orchestrator or api fails during start

### DIFF
--- a/.github/actions/start-services/action.yml
+++ b/.github/actions/start-services/action.yml
@@ -34,11 +34,14 @@ runs:
         echo "ENVD_TIMEOUT=${ENVD_TIMEOUT}" >> .env.test
         
         mkdir -p ~/logs
+
         make -C packages/orchestrator run-debug 2>&1 | tee ~/logs/orchestrator.log &
+        ORCH_PID=$!
         sleep 30
-        echo "Orchestrator started, starting API now..."
+        if ps -p $ORCH_PID > /dev/null; then echo "Orchestrator started, starting API now..."; else exit 1; fi
 
         make -C packages/api run 2>&1 | tee ~/logs/api.log &
+        API_PID=$!
         sleep 30
-        echo "API service started"
+        if ps -p $API_PID > /dev/null; then echo "API service started"; else exit 1; fi
       shell: bash


ValentaTomas (approved):

.github/actions/start-services/action.yml
```diff
@@ -34,11 +34,14 @@ runs:
...
+        ORCH_PID=$!
         sleep 30
-        echo "Orchestrator started, starting API now..."
+        if ps -p $ORCH_PID > /dev/null; then echo "Orchestrator started, starting API now..."; else exit 1; fi
 
         make -C packages/api run 2>&1 | tee ~/logs/api.log &
+        API_PID=$!
         sleep 30
-        echo "API service started"
+        if ps -p $API_PID > /dev/null; then echo "API service started"; else exit 1; fi
```
0div, line 46: it could save some time if we did only one `sleep 30` after running the two `make` commands and doing the two `if ps -p ${ORCH,API}_PID` afterwards.

.github/actions/start-services/action.yml
```diff
@@ -34,11 +34,14 @@ runs:
         echo "ENVD_TIMEOUT=${ENVD_TIMEOUT}" >> .env.test
         
         mkdir -p ~/logs
+
         make -C packages/orchestrator run-debug 2>&1 | tee ~/logs/orchestrator.log &
+        ORCH_PID=$!
         sleep 30
-        echo "Orchestrator started, starting API now..."
+        if ps -p $ORCH_PID > /dev/null; then echo "Orchestrator started, starting API now..."; else exit 1; fi
```
0div, line 41: i would echo that "Orchestrator is not responsive after 30secs" or something to make it clear to someone debugging.

0div, line 41: you could also `wait` the PID to get the exact exit code

0div (changes requested):

.github/actions/start-services/action.yml
dobrac: I need to be sure that orchestrator is healthy before I start API, otherwise they won't connect together

0div: either way we non-zero exit which is the intent?

dobrac: I think so. Basically the sleep 30 is there instead of proper health check that the api/orchestrator booted successfully. So you need some sleep for the API and Orchestrator. Why there is not one, but two sleeps, is that for proper boot, I need to have the orchestrator ready before the API starts starting, and need to have the API ready before moving to the next step in the GH pipeline. Does it make sense?

0div: I see, makes sense.

PR #410, opened 2025-03-10
rm FlameGraph folder, it was added for debugging purposes

r33drichards:


r33drichards:
i deleted this locally so it shouldn't happen again

PR #411, opened 2025-03-10
Prevent removing resumed sandbox if the cleanup for the paused sandbox takes too long

ValentaTomas:
https://e2b-team.slack.com/archives/C05AGT4UFMJ/p1741625029438799
This should be a general improvement that might fix the outlined problem—still testing.


ValentaTomas:
https://linear.app/e2b/issue/E2B-1751/prevent-removing-resumed-sandbox-if-the-cleanup-for-the-paused-sandbox

r33drichards (approved):

PR #412, opened 2025-03-11
Track sandbox firecracker process errors and return them when requesting sandbox 

0div:
# Description 

- [x] Start using redis client in ochestrator server and 
- [x] Storing sandbox errors in it
- [ ] Use redis client in "session proxy" DNS server ?


0div:
@jakubno is it safe for me to make changes to the orchestrator [DNS server](https://github.com/e2b-dev/infra/blob/573e408f160235648911901678ac178a37eff23c/packages/orchestrator/internal/dns/server.go) for it to return sandbox errors? Or will you remove it soon?

ValentaTomas:
I would wait for the changes to the orchestrator state management by @tychoish and also after we merge the session proxy (https://github.com/e2b-dev/infra/pull/417). Then we can open a new PR.

PR #415, opened 2025-03-11
Add basic orchestrator healthcheck

dobrac:
Adds:
- HTTP Endpoint "/health" reporting 
```
{
    "status": "healthy",
    "version": "6fe34854"
}
```
- Orchestrator count metric (async updown counter) "orchestrator.env.sandbox.running"
- OTEL InstanceID

### DIFF
--- a/packages/orchestrator/internal/server/healthcheck.go
+++ b/packages/orchestrator/internal/server/healthcheck.go
@@ -1,0 +1,150 @@
+package server
+
+import (
+	"context"
+	"encoding/json"
+	"log"
+	"net"
+	"net/http"
+	"sync"
+	"time"
+
+	"go.opentelemetry.io/otel/metric"
+	"go.uber.org/zap"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/health"
+	healthpb "google.golang.org/grpc/health/grpc_health_v1"
+
+	"github.com/e2b-dev/infra/packages/shared/pkg/meters"
+)
+
+const healthcheckFrequency = 5 * time.Second
+const healthcheckTimeout = 30 * time.Second
+
+type Status string
+
+const (
+	Healthy   Status = "healthy"
+	Unhealthy Status = "unhealthy"
+)
+
+type Healthcheck struct {
+	version    string
+	server     *server
+	grpc       *grpc.Server
+	grpcHealth *health.Server
+
+	// TODO: Replace with status from SQL Lite
+	status  Status
+	lastRun time.Time
+	mu      sync.RWMutex
+
+	sandboxCounter metric.Int64Gauge
+}
+
+func NewHealthcheck(server *server, grpc *grpc.Server, grpcHealth *health.Server, version string) (*Healthcheck, error) {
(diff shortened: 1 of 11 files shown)


jakubno (commented):
Do I understand it correctly that orchestrator its own health endpoint to determine its health?

packages/orchestrator/internal/server/healthcheck.go
```diff
@@ -0,0 +1,135 @@
...
+	defer h.mu.Unlock()
+
+	childCtx, cancel := context.WithTimeout(ctx, healthcheckTimeout)
+	defer cancel()
+
+	// Update last run on report
+	h.lastRun = time.Now()
+
+	// Report health
+	c, err := h.grpcHealth.Check(childCtx, &healthpb.HealthCheckRequest{
```
jakubno: Add some comment explaining the "self" check

dobrac: refactored to a separate function + added comment explaining the report + grpchealth

packages/orchestrator/internal/server/healthcheck.go
```diff
@@ -0,0 +1,146 @@
...
+	"time"
+
+	"go.uber.org/zap"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/health"
+	healthpb "google.golang.org/grpc/health/grpc_health_v1"
+)
+
+const healthcheckFrequency = 5 * time.Second
+const healthcheckTimeout = 30 * time.Second
```
jakubno, line 19: does this make sense (timeout > frequency)?

packages/orchestrator/internal/server/healthcheck.go
```diff
@@ -0,0 +1,146 @@
...
+			if err != nil {
+				zap.L().Error("Error reporting healthcheck", zap.Error(err))
+			}
+		}
+	}
+}
+
+// report updates the health status.
+// This function is run in a goroutine every healthcheckFrequency for the reason of having
+// longer running tasks that might me too slow or resource intensive to be run
```
jakubno:
```suggestion
// longer running tasks that might be too slow or resource intensive to be run
```

jakubno (approved):

packages/orchestrator/internal/server/healthcheck.go
dobrac: The timeout is 30 seconds for a reason, when we'll try to start sandbox in the healthcheck. The 5 seconds is there that 30 seconds might be unnecessary too long when everything is okay.

jakubno: Usually the health check interval should be longer than timeout, otherwise something like this can happen
there is some issue and health checks will timeout, so it will set unhealthy after 30s, but in the meantime the service started to work again and for some period of time there are changes from healthy to unhealthy and back, because the successful health checks will be mixed with failing ones

dobrac: Let's make it 10 seconds for both then?

jakubno: It's running sync, so this can't happen, sorry

PR #417, opened 2025-03-11
Replace Nginx based session proxy with orchestrator proxy

sitole:
Implements http proxy in orchestrator that replaces current nginx-based session proxy that is using dns server from orchestrator to route traffic.

- Legacy Nginx session-proxy is intact and can still be used as it is, next step will be remove it when traffic migrated
- Orchestrator proxy for now runs on port `5007` allowing smooth deployment
- Support for pretty (for browsers) and json error messages when port in sandbox not listening
- Proxy settings is taken from currently used client-proxy, so should be already battle-tested

### DIFF
--- a/packages/client-proxy/main.go
+++ b/packages/client-proxy/main.go
@@ -112,7 +112,7 @@ func proxyHandler(transport *http.Transport) func(w http.ResponseWriter, r *http
 		zap.L().Debug("Proxying request", zap.String("sandbox_id", sandboxID), zap.String("node", node))
 		targetUrl := &url.URL{
 			Scheme: "http",
-			Host:   fmt.Sprintf("%s:%d", node, orchestratorProxyPort),
+			Host:   fmt.Sprintf("%s:%d", node, sandboxPort),
 		}
 
 		// Proxy the request
--- a/packages/client-proxy/main.go
+++ b/packages/client-proxy/main.go
@@ -28,13 +28,13 @@ import (
 )
 
 const (
-	ServiceName      = "client-proxy"
-	dnsServer        = "api.service.consul:5353"
-	healthCheckPort  = 3001
-	port             = 3002
-	sandboxPort      = 3003 // legacy session proxy port
-	sandboxProxyPort = 5007 // orchestrator proxy port
-	maxRetries       = 3
+	ServiceName           = "client-proxy"
+	dnsServer             = "api.service.consul:5353"
+	healthCheckPort       = 3001
+	port                  = 3002
+	sandboxPort           = 3003 // legacy session proxy port
+	orchestratorProxyPort = 5007 // orchestrator proxy port
+	maxRetries            = 3
 )
 
 var commitSHA string
@@ -112,7 +112,7 @@ func proxyHandler(transport *http.Transport) func(w http.ResponseWriter, r *http
 		zap.L().Debug("Proxying request", zap.String("sandbox_id", sandboxID), zap.String("node", node))
 		targetUrl := &url.URL{
 			Scheme: "http",
-			Host:   fmt.Sprintf("%s:%d", node, sandboxProxyPort),
+			Host:   fmt.Sprintf("%s:%d", node, orchestratorProxyPort),
 		}
 
 		// Proxy the request
--- a/packages/client-proxy/main.go
+++ b/packages/client-proxy/main.go
@@ -28,12 +28,13 @@ import (
 )
 
 const (
-	ServiceName     = "client-proxy"
-	dnsServer       = "api.service.consul:5353"
-	healthCheckPort = 3001
-	port            = 3002
-	sandboxPort     = 3003
-	maxRetries      = 3
+	ServiceName      = "client-proxy"
+	dnsServer        = "api.service.consul:5353"
+	healthCheckPort  = 3001
+	port             = 3002
+	sandboxPort      = 3003 // legacy session proxy port
+	sandboxProxyPort = 5007 // orchestrator proxy port
+	maxRetries       = 3
 )
 
 var commitSHA string
@@ -111,7 +112,7 @@ func proxyHandler(transport *http.Transport) func(w http.ResponseWriter, r *http
 		zap.L().Debug("Proxying request", zap.String("sandbox_id", sandboxID), zap.String("node", node))
 		targetUrl := &url.URL{
 			Scheme: "http",
-			Host:   fmt.Sprintf("%s:%d", node, sandboxPort),
+			Host:   fmt.Sprintf("%s:%d", node, sandboxProxyPort),
 		}
 
 		// Proxy the request
--- a/packages/orchestrator/cmd/mock-snapshot/mock.go
+++ b/packages/orchestrator/cmd/mock-snapshot/mock.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"github.com/e2b-dev/infra/packages/orchestrator/internal/proxy"
 	"log"
 	"os"
 	"os/signal"
@@ -113,6 +114,7 @@ func mockSnapshot(
 	buildId,
 	sandboxId string,
 	dns *dns.DNS,
+	proxy *proxy.SessionProxy,
 	keepAlive time.Duration,
 	networkPool *network.Pool,
 	templateCache *template.Cache,
@@ -137,6 +139,7 @@ func mockSnapshot(
 		childCtx,
 		tracer,
 		dns,
+		proxy,
 		networkPool,
 		templateCache,
 		&orchestrator.SandboxConfig{
@@ -234,7 +237,7 @@ func mockSnapshot(
 		return fmt.Errorf("failed to add snapshot to template cache: %w", err)
 	}
 
-	fmt.Println("Add snapshot to template cache time: ", time.Since(snapshotTime).Milliseconds())
+	fmt.Println("AddSandbox snapshot to template cache time: ", time.Since(snapshotTime).Milliseconds())
 
 	start = time.Now()
 
@@ -242,6 +245,7 @@ func mockSnapshot(
 		childCtx,
 		tracer,
 		dns,
+		proxy,
 		networkPool,
 		templateCache,
 		&orchestrator.SandboxConfig{
--- a/packages/orchestrator/cmd/mock-snapshot/mock.go
+++ b/packages/orchestrator/cmd/mock-snapshot/mock.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"flag"
 	"fmt"
-	"github.com/e2b-dev/infra/packages/orchestrator/internal/proxy"
 	"log"
 	"os"
 	"os/signal"
@@ -15,6 +14,7 @@ import (
 	"golang.org/x/sync/errgroup"
 
 	"github.com/e2b-dev/infra/packages/orchestrator/internal/dns"
+	"github.com/e2b-dev/infra/packages/orchestrator/internal/proxy"
 	"github.com/e2b-dev/infra/packages/orchestrator/internal/sandbox"
 	"github.com/e2b-dev/infra/packages/orchestrator/internal/sandbox/nbd"
 	"github.com/e2b-dev/infra/packages/orchestrator/internal/sandbox/network"
@@ -52,6 +52,7 @@ func main() {
 		cancel()
 	}()
 
+	proxyServer := proxy.New(3333)
 	dnsServer := dns.New()
 	go func() {
 		log.Printf("Starting DNS server")
@@ -91,6 +92,7 @@ func main() {
 			*buildId,
 			*sandboxId+"-"+strconv.Itoa(v),
 			dnsServer,
+			proxyServer,
 			time.Duration(*keepAlive)*time.Second,
 			networkPool,
 			templateCache,
--- a/packages/orchestrator/cmd/mock-snapshot/mock.go
+++ b/packages/orchestrator/cmd/mock-snapshot/mock.go
@@ -239,7 +239,7 @@ func mockSnapshot(
 		return fmt.Errorf("failed to add snapshot to template cache: %w", err)
 	}
 
-	fmt.Println("AddSandbox snapshot to template cache time: ", time.Since(snapshotTime).Milliseconds())
+	fmt.Println("Add snapshot to template cache time: ", time.Since(snapshotTime).Milliseconds())
 
 	start = time.Now()
 
--- a/packages/orchestrator/cmd/mock-snapshot/mock.go
+++ b/packages/orchestrator/cmd/mock-snapshot/mock.go
@@ -116,7 +116,7 @@ func mockSnapshot(
 	buildId,
 	sandboxId string,
 	dns *dns.DNS,
-	proxy *proxy.SessionProxy,
+	proxy *proxy.SandboxProxy,
 	keepAlive time.Duration,
 	networkPool *network.Pool,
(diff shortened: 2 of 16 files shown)


packages/client-proxy/main.go
```diff
@@ -111,7 +112,7 @@ func proxyHandler(transport *http.Transport) func(w http.ResponseWriter, r *http
 		zap.L().Debug("Proxying request", zap.String("sandbox_id", sandboxID), zap.String("node", node))
 		targetUrl := &url.URL{
 			Scheme: "http",
-			Host:   fmt.Sprintf("%s:%d", node, sandboxPort),
+			Host:   fmt.Sprintf("%s:%d", node, sandboxProxyPort),
```
ValentaTomas: Would probably name it orchestratorProxy as it is named so everywhere else

packages/orchestrator/internal/proxy/proxy.go
```diff
@@ -0,0 +1,184 @@
...
+)
+
+//go:embed proxy_browser_502.html
+var proxyBrowser502PageHtml string
+
+var browserIdentityKeywords = []string{
+	"mozilla", "chrome", "safari", "firefox", "edge", "opera", "msie",
+}
+
+type SessionProxy struct {
```
ValentaTomas: "Session" was named so because originally we had sessions and not sandboxes. We can rename these to something more fitting—the SandboxProxy might make sense here (inside orchestrator).

packages/orchestrator/internal/sandbox/sandbox_linux.go
```diff
@@ -314,11 +316,13 @@ func NewSandbox(
 	sbx.StartedAt = time.Now()
 
 	dns.Add(config.SandboxId, ips.HostIP())
+	proxy.AddSandbox(config.SandboxId, ips.HostIP())
 
 	telemetry.ReportEvent(childCtx, "added DNS record", attribute.String("ip", ips.HostIP()), attribute.String("hostname", config.SandboxId))
 
 	cleanup.Add(func() error {
 		dns.Remove(config.SandboxId, ips.HostIP())
+		proxy.RemoveSandbox(config.SandboxId)
```
ValentaTomas: The danger here—and the reason we used the hostIP to specify the remove in DNS is that it might be possible to accidentally remove a new entry in the DNS/proxy (for example for a resumed sandbox with he same sandboxID) because the cleanup might run asynchronously in some cases.
I would add a way to make it so the removal is again specific to sandboxID+hostIP.

ValentaTomas:
Right now (for the deployment) we will have both proxies running at the same time?

packages/orchestrator/internal/sandbox/sandbox_linux.go
ValentaTomas: I'm thinking we may later just use the already existing sandbox map in orchestrator that we pass to the proxy part to handle the proxying without explicitly add/remove to the DNS.

sitole:
> Right now (for the deployment) we will have both proxies running at the same time?

Yea, both are running but for traffic new one is used. Actually it's just change of the port in client-proxy.

Comes to my mind, if you think its somehow invasive we can for example route traffic to now one just for some teams / enable it with some feature flag to enable quick revert. Request proxy is done in context where we know about exact client so gives us big space.

packages/orchestrator/internal/sandbox/sandbox_linux.go
sitole: Yes, was my first idea. Was not sure if its okay to without context know-how just use list directly, because there can be sandbox in map but it should not be routed (paused, not healthy etc etc) and i taught will be more transparent do it as now explicitly add/remove sandboxed that should be router. But sure, can be one map theoretically.

packages/orchestrator/internal/proxy/proxy.go
sitole: agree, will rename

packages/orchestrator/internal/sandbox/sandbox_linux.go
ValentaTomas: Yes, I think that was correct thinking, because there are dependencies and the order matters. So let's switch it as it is and then we can refactor to map later.

packages/orchestrator/main.go
```diff
@@ -149,6 +155,32 @@ func main() {
...
 	}()
 
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+
+		errChan := make(chan error, 1)
+		go func() {
+			err := sessionProxy.Start()
+			errChan <- err
```
0div:
```suggestion
			errChan <- sessionProxy.Start()
```

packages/orchestrator/main.go
```diff
@@ -5,12 +5,14 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"github.com/e2b-dev/infra/packages/orchestrator/internal/proxy"
```
0div: nit: put this in the bottom internal import sublist

packages/orchestrator/internal/proxy/proxy.go
```diff
@@ -0,0 +1,184 @@
...
+
+//go:embed proxy_browser_502.html
+var proxyBrowser502PageHtml string
+
+var browserIdentityKeywords = []string{
+	"mozilla", "chrome", "safari", "firefox", "edge", "opera", "msie",
+}
+
+type SandboxProxy struct {
+	sandboxes *smap.Map[string]
```
0div, line 42: do we plan on eventually using the redis client here? I'm asking bc i will likely be storing sandbox errors with it and return them at this step and/or api.

packages/orchestrator/main.go
0div: (nit)

packages/orchestrator/main.go
```diff
@@ -149,6 +155,32 @@ func main() {
...
+
+		select {
+		case <-ctx.Done():
+		case <-errChan:
+			if err != nil {
+				zap.L().Error("session proxy failed", zap.Error(err))
+				exitCode.Add(1)
+				cancel()
+			}
+		}
```
0div:
```suggestion
		case err := <-errChan:
			if err != nil {
				zap.L().Error("session proxy failed", zap.Error(err))
				exitCode.Add(1)
				cancel()
			}
		}
```

packages/orchestrator/internal/proxy/proxy.go
```diff
@@ -0,0 +1,184 @@
...
+func (p *SandboxProxy) buildHtmlClosedPortError(sandboxId string, host string, port uint64) []byte {
+	replacements := map[string]string{
+		"{{sandbox_id}}":   sandboxId,
+		"{{sandbox_port}}": strconv.FormatUint(port, 10),
+		"{{sandbox_host}}": host,
+	}
+
+	adjustedErrTemplate := proxyBrowser502PageHtml
+	for placeholder, value := range replacements {
+		adjustedErrTemplate = strings.ReplaceAll(adjustedErrTemplate, placeholder, value)
```
0div: https://pkg.go.dev/html/template probably more efficient and appropriate here if we're gonna return more detailed error pages down the line.

0div (changes requested):

packages/orchestrator/internal/proxy/proxy.go
sitole: not sure. [discussed with tomas](https://github.com/e2b-dev/infra/pull/417#discussion_r1990092471) that in future we should use some orchestrator-wide map.

sitole:
> Commit https://github.com/e2b-dev/infra/pull/417/commits/96c3425ed26d98bab989b76db3e7e56bc68472b9

Reverts change that routed traffic to orchestrator proxy. After internal discussion we decided that we want to implement orchestrator proxy and then try to add feature-flags to enable it, but its dedicated to separated request to not be stack on it.

Just if you want to test changes its needed to revert this commit, without it traffic will float via session-proxy instead.

packages/orchestrator/internal/sandbox/sandbox_linux.go
sitole: implemented checking if sandboxid and ip are matching

packages/orchestrator/main.go
```diff
@@ -149,6 +155,31 @@ func main() {
...
+		case err = <-errChan:
+			if err != nil {
+				zap.L().Error("session proxy failed", zap.Error(err))
+				exitCode.Add(1)
+				cancel()
+			}
+		}
+
+		// close session proxy, wait 5 seconds until all connections are closed
+		shutdownCtx, shutdownCtxCancel := context.WithTimeout(context.Background(), 5*time.Second)
```
jakubno: this can be up to 24 hours (sandbox max lifetime)

packages/orchestrator/cmd/mock-snapshot/mock.go
```diff
@@ -234,14 +239,15 @@ func mockSnapshot(
 		return fmt.Errorf("failed to add snapshot to template cache: %w", err)
 	}
 
-	fmt.Println("Add snapshot to template cache time: ", time.Since(snapshotTime).Milliseconds())
+	fmt.Println("AddSandbox snapshot to template cache time: ", time.Since(snapshotTime).Milliseconds())
```
jakubno: mistake?

packages/orchestrator/internal/proxy/proxy.go
```diff
@@ -0,0 +1,193 @@
...
+func (p *SandboxProxy) isBrowser(userAgent string) bool {
+	userAgent = strings.ToLower(userAgent)
+	for _, keyword := range browserIdentityKeywords {
+		if strings.Contains(userAgent, keyword) {
+			return true
+		}
+	}
+
+	return false
+}
```
jakubno: I did some benchmark
```
BenchmarkIsBrowserSlice-11       7179894               144.4 ns/op
BenchmarkIsBrowserRegex-11      15990450                74.71 ns/op
```
```suggestion
func (p *SandboxProxy) isBrowser(userAgent string) bool {
	return browserRegex.MatchString(userAgent)
}
```

packages/orchestrator/internal/proxy/proxy.go
```diff
@@ -0,0 +1,193 @@
...
+	"strings"
+	"time"
+)
+
+//go:embed proxy_browser_502.html
+var proxyBrowser502PageHtml string
+
+var browserIdentityKeywords = []string{
+	"mozilla", "chrome", "safari", "firefox", "edge", "opera", "msie",
+}
```
jakubno: feel free to generate the regex based on the current slice, it's easier to read
```suggestion
var browserRegex = regexp.MustCompile(`(?i)mozilla|chrome|safari|firefox|edge|opera|msie`)
```

packages/orchestrator/internal/proxy/proxy.go
jakubno: What is the reason to ot use the sandboxes `sandboxes` smap directly?

Discussed in person and it's e.g. to prevent some broken connection to already closing sandbox

packages/orchestrator/internal/proxy/proxy.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+}
+
+func (p *SandboxProxy) buildHtmlClosedPortError(sandboxId string, host string, port uint64) ([]byte, error) {
+	htmlResponse := new(bytes.Buffer)
+	htmlTmpl, err := template.New("template").Parse(proxyBrowser502PageHtml)
+	if err != nil {
+		return nil, err
+	}
+
```
jakubno: You can initialize the template only once

packages/orchestrator/internal/proxy/proxy.go
```diff
@@ -0,0 +1,193 @@
...
+	err = htmlTmpl.Execute(htmlResponse, map[string]string{"sandboxId": sandboxId, "sandboxHost": host, "sandboxPort": fmt.Sprintf("%d", port)})
+	if err != nil {
+		return nil, err
+	}
+
+	return htmlResponse.Bytes(), nil
+}
+
+func (p *SandboxProxy) buildJsonClosedPortError(sandboxId string, port uint64) []byte {
+	response := map[string]interface{}{
```
jakubno: please define the type for improving the performance

packages/orchestrator/internal/proxy/proxy.go
```diff
@@ -0,0 +1,193 @@
...
+}
+
+func (p *SandboxProxy) buildHtmlClosedPortError(sandboxId string, host string, port uint64) ([]byte, error) {
+	htmlResponse := new(bytes.Buffer)
+	htmlTmpl, err := template.New("template").Parse(proxyBrowser502PageHtml)
+	if err != nil {
+		return nil, err
+	}
+
+	err = htmlTmpl.Execute(htmlResponse, map[string]string{"sandboxId": sandboxId, "sandboxHost": host, "sandboxPort": fmt.Sprintf("%d", port)})
```
jakubno: please define the type for improving the performance and this could be better `strconv.FormatUint(port, 10)` than `Sprintf`

jakubno (changes requested):

packages/orchestrator/internal/proxy/proxy.go
sitole: https://github.com/e2b-dev/infra/pull/417#discussion_r1990106663

packages/orchestrator/main.go
sitole: in client-proxy there is for server shutdown function used `context.Background()` should we use that? Like it maybe more than just explicitly define 24 hours timeout.

Will also add log telling server shutdown started to make it more clear what is happening during shutdown.

jakubno (changes requested):
I think there's some problem with RPC after this change (I changed the port in client proxy). I can write / read files from sandbox (uses HTTP server). But I can't start command etc..

ValentaTomas:
@sitole Building up to what @jakubno encountered—running the E2B SDK test suite (and also potentially the code interpreter tests) against your staging with might be a good early test for similar functionality.

sitole:
With commit 3330b9fc26d6fbbdfae78d23109dd74376df5583 i fixed issue were envd randomly responds with `connection
reset by peer`.
Disabling connections keep-alive on new proxy resolved issue.

I noticed current nginx session-proxy is missing `keepalive` setup (https://nginx.org/en/docs/http/ngx_http_upstream_module.html#keepalive) and it looks like default value for missing keep-alive statement is empty, so keep-alive is disabled. That makes sense why current solution worked until now.

I don't like idea to disabling it just because envd complains. Iam mostly concerned about how this can affect traffic. Sure, `keepalive` was already disabled, but iam not sure if this behave exactly matches nginx vs golang http-proxy one.

I will continue with finding some relevant docs about golang http.server and handling idle connections.

cc @ValentaTomas @jakubno

jakubno (approved):

packages/orchestrator/internal/proxy/proxy.go
```diff
@@ -0,0 +1,196 @@
...
+
+type htmlTemplateData struct {
+	SandboxId   string
+	SandboxHost string
+	SandboxPort string
+}
+
+type jsonTemplateData struct {
+	Error     string `json:"error"`
+	SandboxId string `json:"sandbox_id"`
```
jakubno: it was `sandboxId` before, there are already tests in sdk, if there isn't a strong argument to rename it, please revert to the old naming

jakubno (changes requested):

0div (approved):

jakubno (approved):

PR #420, opened 2025-03-11
Use google.com instead of e2b.dev to test internet connectivity

r33drichards:
We can change the endpoint we are checking so we are not reporting sandbox * [e2b.dev](http://e2b.dev/) availability this way? :D


r33drichards:
https://github.com/e2b-dev/infra/actions/runs/13799295385/job/38598199250

mlejva (approved):

PR #421, opened 2025-03-11
Connect ClickHouse client to clickhouse deployment in terraform

0div:
- generate clickhouse password with random
- store it in google secrets
- get that secret in api


r33drichards (approved):

main.tf
```diff
@@ -173,7 +173,7 @@ module "nomad" {
   # Clickhouse
   clickhouse_connection_string = var.clickhouse_connection_string
   clickhouse_username          = var.clickhouse_username
-  clickhouse_password          = var.clickhouse_password
+  clickhouse_password          = module.init.clickhouse_password_secret_data
   clickhouse_database          = var.clickhouse_database
```
jakubno, line 177: If you are handling password this way, imho everything should be in GCP secrets. It would be the the aligned how we handle e.g. database.

Also is there a reason to have username, password, database and then connection string? aren't those already in the connection string?

jakubno (changes requested):

main.tf
0div: Makes sense for the username, as for the connection string, the native CH go client config doesn't seem well suited to accept a DSN like that.

r33drichards: > If you are handling password this way, imho everything should be in GCP secrets. It would be the the aligned how we handle e.g. database.

password is only a secret for developer convenience, if we need to get that value for our dev cluster we can find them in our gcp secrets. otherwise no reason to put it there at all and we could leave it in terraform state. 

as for username, password, and database, these are all static non secret values so no reason to include them in our secrets, just hard coded strings should be sufficient

PR #425, opened 2025-03-12
Cache base template in integration tests

dobrac:
Cache base template in integration tests before the first test run by creating first sandbox. This prevents timing out or setting high timeout on unrelevant tests.


dobrac:
found an issue, fixing

jakubno (changes requested):
I think it didn't work
image

dobrac:
fixed

PR #427, opened 2025-03-12
Fix reference to correct go file for seeding database on prep-cluster

ben-fornefeld:
under `/packages/shared/Makefile` we have a command `prep-cluster` which is also mentioned in `self-host.md`. the referenced go file was missing, which is fixed by this pr.

### DIFF
--- a/packages/orchestrator/internal/sandbox/build/cache.go
+++ b/packages/orchestrator/internal/sandbox/build/cache.go
@@ -21,13 +21,13 @@ type deleteDiff struct {
 
 type DiffStore struct {
 	cachePath string
-	cache     *ttlcache.Cache[string, Diff]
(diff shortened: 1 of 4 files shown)


packages/shared/scripts/e2b.toml
```diff
@@ -10,6 +10,7 @@
 # import { Sandbox } from 'e2b'
 # const sandbox = await Sandbox.create('base')
 
+team_id = "58ce6562-fb73-476f-aadb-2232e6ae35ef"
```
mishushakov: should this be added?

jakubno: No

jakubno (changes requested):

PR #428, opened 2025-03-12
Create config DB if not exists in migrator

0div:
* explicitly specify db name in migration files
* create config DB is not exist upon starting migrator

### DIFF
--- a/packages/shared/pkg/chdb/migrator.go
+++ b/packages/shared/pkg/chdb/migrator.go
@@ -86,6 +86,13 @@ func NewMigrator(config ClickHouseConfig) (*ClickhouseMigrator, error) {
 		},
 	})
 
+	_, err = db.Exec(fmt.Sprintf(`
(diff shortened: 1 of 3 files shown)


packages/shared/pkg/chdb/migrator.go
```diff
@@ -86,15 +86,22 @@ func NewMigrator(config ClickHouseConfig) (*ClickhouseMigrator, error) {
 		},
 	})
 
-	_, err = db.Exec(`
-		CREATE TABLE IF NOT EXISTS schema_migrations (
+	_, err = db.Exec(fmt.Sprintf(`
```
r33drichards, line 89: this makes me a little squeamish b/c of sql injection here

0div:
let's use default for now, there are some implications to using custom db name that we don't need to worry about for now, let's ship this quicker.

PR #431, opened 2025-03-12
Connect clickhouse client to clickhouse deployment

r33drichards:
Clickhouse needs some environment variables to be set in order to connect to the database. We need to set them but they are static and not secret. The password is also generated by terraform. This removes the top level variables so that the programmer does not have to include configuration to deploy clickhouse

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -171,10 +171,10 @@ module "nomad" {
   otel_tracing_print      = var.otel_tracing_print
 
   # Clickhouse
-  clickhouse_connection_string = var.clickhouse_connection_string
-  clickhouse_username          = var.clickhouse_username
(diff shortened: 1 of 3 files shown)


main.tf
```diff
@@ -171,10 +171,10 @@ module "nomad" {
   otel_tracing_print      = var.otel_tracing_print
 
   # Clickhouse
-  clickhouse_connection_string = var.clickhouse_connection_string
-  clickhouse_username          = var.clickhouse_username
+  clickhouse_connection_string = "clickhouse.service.consul:9000"
```
0div, line 174: perhaps you could use an output here, at least for the port?

0div (approved):
LGTM 🚀

PR #433, opened 2025-03-12
Fix build cache key for snapshots

dobrac:
The https://github.com/e2b-dev/infra/pull/253 introduced a bug by using wrong build cache key when creating snapshots. This PR fixes the cache key and adds a bit of refactor to prevent the same bug in the future.

The issue could be simulated using the following steps:
1) start sandbox
2) pause sandbox
3) ...sandbox files are uploading to GCP, but not in the build cache (because of the wrong key)
4) resume sandbox
5) cache miss, trying to download from GCP, but nothing is yet in GCP -> error

### DIFF
--- a/packages/orchestrator/internal/sandbox/build/local_diff.go
+++ b/packages/orchestrator/internal/sandbox/build/local_diff.go
@@ -13,7 +13,7 @@ import (
 type LocalDiffFile struct {
 	*os.File
 	cachePath string
-	cacheKey  string
+	cacheKey  DiffStoreKey
 }
 
 func NewLocalDiffFile(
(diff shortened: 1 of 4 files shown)


packages/orchestrator/internal/sandbox/build/local_diff.go
```diff
@@ -103,8 +113,8 @@ func (b *localDiff) FileSize() (int64, error) {
 	return b.cache.FileSize()
 }
 
-func (b *localDiff) CacheKey() string {
-	return b.cachePath
```
dobrac, line 107: this is the bug, it shouldn't have been cachePath, but cacheKey (which differs)

ValentaTomas (approved):

PR #436, opened 2025-03-13
Prevent API and Loki colocation on nodes

ValentaTomas:
Because the Loki can consume a lot of RAM/CPU we might want to prevent it being placed on the same node as API. Also in case of problems with the node because of Loki resource this would minimize impact.

This PR prevent the colocation by using a Nomad scheduling trick with defining the same port for both jobs.

This though prevent the scheduling when there is only one node (dev) which might cause some problems.


ValentaTomas:
It might be worth refactoring the shared port.

ValentaTomas:
These changes are not deployed on foxtrot, because the Loki restart when placed on the same node as API seemed it could down the whole node.

jakubno (approved):

PR #440, opened 2025-03-13
Fix gRPC idle status reporting

dobrac:
Fixes status reporting when grpc is in IDLE status. Right now, if the node goes to the IDLE state (no requests for longer time), no new sandboxes could be created and the state doesn't change back.

### DIFF
--- a/packages/api/internal/orchestrator/node.go
+++ b/packages/api/internal/orchestrator/node.go
@@ -51,8 +51,15 @@ func (n *Node) Status() api.NodeStatus {
 		return n.status
 	}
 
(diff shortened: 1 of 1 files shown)


jakubno (changes requested):
I think you can't recover from unhealthy/connecting status

jakubno:
API handling, once you set the node status to unhealthy it can't go back

dobrac:
Ok, I've changed it to connecting

jakubno:
Sorry, it was okay, i've just got bamboozled with `Status()` and `.status` 🙈

jakubno (approved):

PR #441, opened 2025-03-13
Adds sandbox startedAt to snapshot

mishushakov:
Added sandbox startedAt to snapshot

- would require db migration
- you'd need to pull sandbox logs from billing to set when they were started

more info:
https://e2b-team.slack.com/archives/C07VCHSF9NK/p1739207493629409


ValentaTomas:
@mish Can this move from draft? Is this necessary for the sandbox paginated listing (of paused)?

mishushakov:
@ValentaTomas it is, but it's already integrated in the Sandbox.list PR

ValentaTomas:
Got it, I saw the DB migration there and was confused.

ValentaTomas:
@mishushakov Can we close as it is in Sandbox.list?

PR #443, opened 2025-03-13
Add timestamp to periodic checks

jakubno:
Add timestamp to improve debugging


dobrac (approved):
utility function would be better

PR #448, opened 2025-03-13
Lower Loki chunk retention and ttls

ValentaTomas:
From the Loki Grafana it seems that part of the memory problems with the current Loki setup is because of the number of retained chunks. These can be failed chunks in several cases though, so I think we first need to find stable Loki configuration.
The next step would be adding scaling/moving the data to other DB (clickhouse?).

We also disabled `wal` which normally helps with replaying the previous unprocessed traffic, but with our job placement it does not make sense now.

### DIFF
--- a/packages/nomad/loki.hcl
+++ b/packages/nomad/loki.hcl
@@ -92,7 +92,7 @@ chunk_store_config:
   chunk_cache_config:
     embedded_cache:
       enabled: true
-      max_size_mb: 512
+      max_size_mb: 2048
       ttl: 30m
 
 query_range:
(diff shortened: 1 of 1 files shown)


ValentaTomas:
This is deployed on foxtrot, because the Loki there kept restarting.

r33drichards (approved):

ValentaTomas:
It seems like the big memory consumption might be the actual querying as opposed to processing logs. I increased the caches for now, but the monitoring we have does not provide insights into this—without that we are a little blind.

PR #449, opened 2025-03-14
Fix bug when template has multiple builds

jakubno:
# Description

`.Only()` overrides `Limit`, so it will change `Limit` to 2 and then raise an error if there's more builds present

### DIFF
--- a/packages/shared/pkg/db/envs.go
+++ b/packages/shared/pkg/db/envs.go
@@ -122,7 +122,6 @@ func (db *DB) GetEnv(ctx context.Context, aliasOrEnvID string) (result *Template
 				env.HasEnvAliasesWith(envalias.ID(aliasOrEnvID)),
 				env.ID(aliasOrEnvID),
 			),
-			env.HasBuildsWith(envbuild.StatusEQ(envbuild.StatusUploaded)),
 		).
 		WithEnvAliases(func(query *models.EnvAliasQuery) {
 			query.Order(models.Asc(envalias.FieldID)) // TODO: remove once we have only 1 alias per env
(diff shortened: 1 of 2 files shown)


packages/shared/pkg/db/envs.go
```diff
@@ -122,7 +122,6 @@ func (db *DB) GetEnv(ctx context.Context, aliasOrEnvID string) (result *Template
 				env.HasEnvAliasesWith(envalias.ID(aliasOrEnvID)),
 				env.ID(aliasOrEnvID),
 			),
-			env.HasBuildsWith(envbuild.StatusEQ(envbuild.StatusUploaded)),
```
jakubno, line 125: not needed, the is done by the second query

packages/shared/pkg/db/envs.go
```diff
@@ -132,15 +131,15 @@ func (db *DB) GetEnv(ctx context.Context, aliasOrEnvID string) (result *Template
 	if notFound {
 		return nil, nil, fmt.Errorf("template '%s' not found: %w", aliasOrEnvID, err)
 	} else if err != nil {
-		return nil, nil, fmt.Errorf("failed to get env '%s': %w", aliasOrEnvID, err)
+		return nil, nil, fmt.Errorf("failed to get template '%s': %w", aliasOrEnvID, err)
 	}
 
-	build, err = db.Client.EnvBuild.Query().Where(envbuild.EnvID(template.ID), envbuild.StatusEQ(envbuild.StatusUploaded)).Order(models.Desc(envbuild.FieldFinishedAt)).Limit(1).Only(ctx)
+	build, err = db.Client.EnvBuild.Query().Where(envbuild.EnvID(template.ID), envbuild.StatusEQ(envbuild.StatusUploaded)).Order(models.Desc(envbuild.FieldFinishedAt)).First(ctx)
```
jakubno, line 137:
```
// First returns the first EnvBuild entity from the query.
// Returns a *NotFoundError when no EnvBuild was found.
```

mishushakov: I am curious why two queries instead of join?

jakubno: Because entgo doesn't do joins, it does multiple queries underhood and then do the join in memory. This produces better query and also we have better observability and better errors

tests/integration/seed.go
```diff
@@ -119,6 +121,25 @@ func seed(db *db.DB, data SeedData) error {
...
+		SetTotalDiskSizeMB(1982).
+		SetKernelVersion("vmlinux-6.1.102").
+		SetFirecrackerVersion("v1.10.1_1fcdaec").
+		SetEnvdVersion("0.1.5").
+		SetCreatedAt(time.Now().Add(-time.Hour)).
+		Save(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to create env build: %w", err)
+	}
+
```
mishushakov: more elegant?

```go
builds := []map[string]interface{}{
	{
		"id":        data.BuildID,
		"createdAt": nil,
	},
	{
		"id":        uuid.New(),
		"createdAt": time.Now().Add(-time.Hour),
	},
}

for _, build := range builds {
	creator := db.Client.EnvBuild.Create().
		SetID(build["id"].(uuid.UUID)).
		SetEnvID(data.EnvID).
		SetDockerfile("FROM e2bdev/base:latest").
		SetStatus(envbuild.StatusUploaded).
		SetVcpu(2).
		SetRAMMB(512).
		SetFreeDiskSizeMB(512).
		SetTotalDiskSizeMB(1982).
		SetKernelVersion("vmlinux-6.1.102").
		SetFirecrackerVersion("v1.10.1_1fcdaec").
		SetEnvdVersion("0.1.5")
	if build["createdAt"] != nil {
		creator.SetCreatedAt(build["createdAt"].(time.Time))
	}
	_, err = creator.Save(ctx)
	if err != nil {
		return fmt.Errorf("failed to create env build: %w", err)
	}
}
```

mishushakov (approved):

sitole (approved):

PR #453, opened 2025-03-14
Added air live-reload on dev

mishushakov:
- adds make dev-init and make dev command for live auto-reload
- removes cleanup process on local


jakubno:
Imho clean up isn't really needed for local dev

packages/api/main.go
```diff
@@ -218,6 +218,11 @@ func run() int {
 	var cleanupFns []func(context.Context) error
 	exitCode := &atomic.Int32{}
 	cleanupOp := func() {
+		if env.IsLocal() {
```
jakubno: is this really needed, the cleanup should be pretty fast

mishushakov: i will test time diff

mishushakov: ok, there was no significant difference, the most difference was the 15 second delay (L330)

r33drichards (approved):
nice!

packages/api/main.go
ValentaTomas: One last note here—how about making the delay a flag that we pass (can default to 15 sec) instead of depending on local? Do you think that makes sense? Then you would either  add `--no-shutdown-delay` or `--shutdown-delay=0`.

PR #455, opened 2025-03-14
Add sqlc setup

jakubno:
# Description

Minimal configuration for sqlc, the idea is to start minimal and progressively convert to sqlc. 

This way we will have time to properly evaluate without huge investment on refactoring whole codebase. Also the style we want to use it will probably evolve.

Also started separating `shared` and `db` packages


packages/db/db/client.go
```diff
@@ -0,0 +1,65 @@
...
+	database "github.com/e2b-dev/infra/packages/db/db/queries"
+)
+
+type Client struct {
+	*database.Queries
+	ctx  context.Context
+	conn *pgxpool.Pool
+}
+
+var databaseURL = os.Getenv("POSTGRES_CONNECTION_STRING")
```
ValentaTomas: Wouldn't it make sense to load the env vars here via the `utils.RequiredEnv`?

packages/db/db/client.go
```diff
@@ -0,0 +1,65 @@
...
+
+func NewClient(ctx context.Context) (*Client, error) {
+	if databaseURL == "" {
+		return nil, fmt.Errorf("database URL is empty")
+	}
+
+	// Parse the connection pool configuration
+	config, err := pgxpool.ParseConfig(databaseURL)
+	if err != nil {
+		log.Fatalf("Unable to parse database URL: %v", err)
```
ValentaTomas: Does it make sense to do `log.Fatal` here that does `os.Exit` vs returning error so we can handle that (also by quitting probably).

packages/db/db/client.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	// Set the maximum number of connections
+	config.MaxConns = 100
+
+	// Create the connection pool
+	pool, err := pgxpool.NewWithConfig(ctx, config)
+	if err != nil {
+		log.Fatalf("Unable to create connection pool: %v", err)
```
ValentaTomas: Same as the comment above about returning error.

packages/db/queries/example.sql
```diff
@@ -0,0 +1,3 @@
+-- name: Example :one
+SELECT *
```
ValentaTomas: Is the example here intentional?

packages/db/db/client.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+	defer pool.Close()
```
ValentaTomas: Why is the pool closed on defer, before we return the client?

jakubno:
It isn't ready yet, sorry

packages/db/db/migrator.go
```diff
@@ -12,9 +12,6 @@ import (
 
 // Thin wrapper around the migrate package to make it easier to use.
 
-//go:embed migrations/*.sql
```
jakubno: This required to have migrations inside the code folder, I think having it in top level folder in the package is much better

packages/db/scripts/migrate/postgres/main.go
jakubno, line 1: this was only moved

packages/db/queries/teams.sql
```diff
@@ -0,0 +1,5 @@
+-- name: TeamsListForUser :many
+SELECT sqlc.embed(t), sqlc.embed(ut)
```
0div: cool feature! didn't know about this one 🤩

packages/db/pkg/migrator/migrator.go
```diff
@@ -72,7 +70,7 @@ func (dm *DatabaseMigrator) SetLogger(logger migrate.Logger) {
 }
 
 func NewMigrator(connectionString string) (*DatabaseMigrator, error) {
-	d, err := iofs.New(migrationsFS, "migrations")
+	d, err := iofs.New(os.DirFS("."), "migrations")
```
0div: curious what made u decide to move away from go-embedded migration files?

Makefile
```diff
@@ -122,9 +122,15 @@ copy-public-builds:
 	gsutil cp -r gs://e2b-prod-public-builds/kernels/* gs://$(GCP_PROJECT_ID)-fc-kernels/
 	gsutil cp -r gs://e2b-prod-public-builds/firecrackers/* gs://$(GCP_PROJECT_ID)-fc-versions/
 
+
+.PHONY: generate
+generate:
+	rm -rf packages/db/pkg/queries/*
```
0div: im a bit confused by this directory structure, on the one hand we have `packages/db/queries/*.sql` for the raw defined queries and then `packages/db/pkg/queries/*.go` for the generated sqlc go code—the latter naming scheme, especially `.../pkg/...`, doesn't make sense to me.

packages/db/queries/teams.sql
```diff
@@ -0,0 +1,5 @@
...
+FROM "public"."teams" t
+JOIN "public"."users_teams" ut ON ut.team_id = t.id
+WHERE ut.user_id = $1;
```
0div: highly recommend `{table1}_{table2}_join.sql` for any join queries and and singular/plural table names to express `{many,one}-to-{many,one}` at a glance

packages/db/queries/teams.sql
```diff
@@ -0,0 +1,5 @@
+-- name: TeamsListForUser :many
```
0div: highly recommend dumb naming these queries `Get{Thing}With{Thingy}`, or at least verbiage as first word e.g _Get, Update, Delete_

0div (approved):
SQLC FTW

packages/db/pkg/migrator/migrator.go
jakubno: https://github.com/e2b-dev/infra/pull/455#discussion_r2012373462

Makefile
jakubno: Moved it

packages/db/queries/teams.sql
jakubno: Changed the name

packages/db/queries/teams.sql
jakubno: I wanted to start with the model so it's easier to get the relevant stuff for your model, but the search in IDE is fulltext so it doesn't matter that much and you probably want to see WHAT happens first when you read the code, so it will be better as you say :) 

Do you distinguish between Get / List?

0div: No strong opinion on Get vs List although for a query returning single rows i'd prob use Get

0div (approved):

PR #460, opened 2025-03-17
Update loki package to 2.9.13

jakubno:


jakubno:
No need to update it, the vulnerability isn't relevant to us and we will be moving from loki anyway

PR #462, opened 2025-03-17
Add script to enable VPC flow

jakubno:
# Description

Because we are using default network it ain't really possible to use terraform without some imports. (We should use separate non-default network).


jakubno:
https://linear.app/e2b/issue/E2B-1671/gcp-enable-vpc-flow-logs

PR #463, opened 2025-03-17
Limit public access to our VM ports

jakubno:
# Description

- Client proxy health endpoint is the only service with different port for health check
- You can still SSH via GCP console, but you can't connect directly from command line (you would need to add your personal IP)

**Allow**:
- For internal IPs for 22 and 3389 (remote desktop access) - prio 900 (all ips if dev)

**Deny**: 
- All access for 22 and 3389 (remote desktop access) - prio 1000
- Access for not needed ports

### DIFF
--- a/packages/cluster/network/main.tf
+++ b/packages/cluster/network/main.tf
@@ -543,19 +543,20 @@ resource "google_compute_firewall" "internal_remote_connection_firewall_ingress"
 
   priority = 900
 
-  direction     = "INGRESS"
-  target_tags   = [var.cluster_tag_name]
-  source_ranges = var.environment == "dev" ? ["0.0.0.0/0"] : ["10.128.0.0/9"]
+  direction   = "INGRESS"
+  target_tags = [var.cluster_tag_name]
+  # https://googlecloudplatform.github.io/iap-desktop/setup-iap/
+  source_ranges = var.environment == "dev" ? ["0.0.0.0/0"] : ["35.235.240.0/20"]
 }
 
 resource "google_compute_firewall" "remote_connection_firewall_ingress" {
   name    = "${var.prefix}${var.cluster_tag_name}-remote-connection-firewall-ingress"
   network = var.network_name
 
-    deny {
-      protocol = "tcp"
-      ports    = ["22", "3389"]
-    }
+  deny {
+    protocol = "tcp"
+    ports    = ["22", "3389"]
+  }
 
 
 
--- a/packages/cluster/network/main.tf
+++ b/packages/cluster/network/main.tf
@@ -532,17 +532,31 @@ resource "google_compute_firewall" "client_proxy_firewall_ingress" {
 }
 
 
+resource "google_compute_firewall" "internal_remote_connection_firewall_ingress" {
+  name    = "${var.prefix}${var.cluster_tag_name}-internal-remote-connection-firewall-ingress"
+  network = var.network_name
+
+  allow {
+    protocol = "tcp"
+    ports    = ["22", "3389"]
+  }
+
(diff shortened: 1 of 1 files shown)


packages/cluster/network/main.tf
```diff
@@ -531,6 +531,43 @@ resource "google_compute_firewall" "orch_firewall_ingress" {
...
+    ports    = ["22", "3389"]
+  }
+
+
+
+  priority = 1000
+
+  direction     = "INGRESS"
+  target_tags   = [var.cluster_tag_name]
+  source_ranges = ["0.0.0.0/0"]
```
0div, line 567: this is pretty broad, are we eventually considering using VPN to narrow it down?

packages/cluster/network/main.tf
```diff
@@ -531,6 +531,43 @@ resource "google_compute_firewall" "orch_firewall_ingress" {
...
+  source_ranges = var.environment == "dev" ? ["0.0.0.0/0"] : ["35.235.240.0/20"]
+}
+
+resource "google_compute_firewall" "remote_connection_firewall_ingress" {
+  name    = "${var.prefix}${var.cluster_tag_name}-remote-connection-firewall-ingress"
+  network = var.network_name
+
+  deny {
+    protocol = "tcp"
+    ports    = ["22", "3389"]
```
0div, line 558: i don't think we'll use RDP port ever since this has no GUI? i would remove bc a lot of bots on the www scan for these open ports and try to find vulnerabilities, even despite being blocked, it's annoying to get hammered by them

packages/cluster/network/main.tf
jakubno: This is `deny` block

packages/cluster/network/main.tf
jakubno: it's `deny`

packages/cluster/network/main.tf
jakubno: it's allowed by default by gcp

mlejva:
> I'm thinking how to keep the debugging (port forward, list FC processes) functionality we have now with this. Of course we will merge it, but would like to not lose tools.

You can create a secure tunnel with something like tailscale

dobrac:
> > I'm thinking how to keep the debugging (port forward, list FC processes) functionality we have now with this. Of course we will merge it, but would like to not lose tools.
> 
> You can create a secure tunnel with something like tailscale

you create a secure tunnel to the GCP VM through the gcloud cli, which is now used in integration tests to connect to the orchestrator

ValentaTomas:
Oh, I though gcloud used raw SSH. Then everything should be ok as that is also used in internal-cli.<br>Btw, why is there GCP VM in intergration tests? I might be missing some context.

ValentaTomas:
> I'm thinking how to keep the debugging (port forward, list FC processes) functionality we have now with this. Of course we will merge it, but would like to not lose tools.
> 
> VPN?

> I'm thinking how to keep the debugging (port forward, list FC processes) functionality we have now with this. Of course we will merge it, but would like to not lose tools.
> 
> You can create a secure tunnel with something like tailscale


Yes, these are of course options, I just didn't want to be left dry right after merging this because we just lost the tools.

dobrac:
> Oh, I though gcloud used raw SSH. Then everything should be ok as that is also used in internal-cli.Btw, why is there GCP VM in intergration tests? I might be missing some context.

Oh, the GCP tunnel might require the port 22 to be accessible. 

For the integration tests, it’s used when you develop tests locally, so you can directly call the orchestrator grpc endpoints (but having orchestrator hosted in GCP).

PR #465, opened 2025-03-17
Enable IPv6 inside of sandboxes

dobrac:
Enable usage of IPv6 inside of sandboxes for internal use. This is to allow running e.g. [Supabase](https://supabase.com/docs/guides/self-hosting/docker), which is ipv6 only, inside of the sandbox using Docker. 

Changes the kernel parameters from:
`ipv6.disable=1`
to:
`ipv6.disable=0 ipv6.autoconf=1`

### DIFF
--- a/packages/template-manager/internal/build/snapshot_linux.go
+++ b/packages/template-manager/internal/build/snapshot_linux.go
@@ -26,10 +26,12 @@ import (
 )
 
 const (
-	fcMaskLong   = "255.255.255.252"
 	fcMacAddress = "02:FC:00:00:00:05"
-	fcAddr       = "169.254.0.21"
-	fcMask       = "/30"
+
+	// IPv4 configuration
+	fcAddr     = "169.254.0.21"
+	fcMaskLong = "255.255.255.252"
+	fcDNS      = "8.8.8.8" // Google DNS
 
 	fcIfaceID  = "eth0"
 	tmpDirPath = "/tmp"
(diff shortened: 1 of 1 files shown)


packages/template-manager/internal/build/snapshot_linux.go
```diff
@@ -26,10 +26,12 @@ import (
 )
 
 const (
-	fcMaskLong   = "255.255.255.252"
 	fcMacAddress = "02:FC:00:00:00:05"
-	fcAddr       = "169.254.0.21"
-	fcMask       = "/30"
```
r33drichards, line 32: where did this mask go?

dobrac: it wasn't used

r33drichards: it looks like it unused in this pr but it must be unused anywhere?

packages/template-manager/internal/build/snapshot_linux.go
```diff
@@ -279,8 +281,9 @@ func (s *Snapshot) configureFC(ctx context.Context, tracer trace.Tracer) error {
 	childCtx, childSpan := tracer.Start(ctx, "configure-fc")
 	defer childSpan.End()
 
-	ip := fmt.Sprintf("%s::%s:%s:instance:eth0:off:8.8.8.8", fcAddr, fcTapAddress, fcMaskLong)
-	kernelArgs := fmt.Sprintf("quiet loglevel=1 ip=%s reboot=k panic=1 pci=off nomodules i8042.nokbd i8042.noaux ipv6.disable=1 random.trust_cpu=on", ip)
+	// IPv4 configuration - format: [local_ip]::[gateway_ip]:[netmask]:hostname:iface:dhcp_option:[dns]
```
r33drichards, line 284: ❤️

packages/template-manager/internal/build/snapshot_linux.go
dobrac: it wasn't used anywhere in the codebase, if it was, the build would fail in the checks

ValentaTomas: Both were moved/renamed and it should be ok. Thanks for catching the unuse.

r33drichards: > it looks like it unused in this pr but it must be unused anywhere?


didn't see your reply when i sent this!

jakubno (approved):

PR #471, opened 2025-03-18
Store clickhouse data on nfs

r33drichards:
The issue is for how to do migrations for clickhouse, and currently i'm experiencing behavior where clickhouse will lose its system.table metadata when the process is restarted

Storing this on a nfs solves this problem and also make sure that no matter where clickhouse is allocated the data migrates to that node


packages/cluster/main.tf
```diff
@@ -154,6 +154,29 @@ module "client_cluster" {
...
 
+# Add Filestore instance for NFS
+resource "google_filestore_instance" "api_nfs" {
+  name     = "${var.prefix}-api-nfs"
+  tier     = "BASIC_HDD"
+  location = var.gcp_zone
+
+  file_shares {
+    name        = "api_share"
+    capacity_gb = 1024
```
jakubno, line 165: does it make sense to have 1TB disk for metadata?

packages/cluster/main.tf
```diff
@@ -154,6 +154,29 @@ module "client_cluster" {
...
+    nfs_export_options {
+      ip_ranges   = ["10.0.0.0/8"] # VPC network range - adjust as needed
```
jakubno, line 167: should this be adjusted?

jakubno (commented):
Why do we need 1TB NFS for storing table metadata? Can't we use fuse and mount bucket instead? 

Also more than half of files are relevant to the PR...

packages/cluster/main.tf
r33drichards: it does not, but that's the default minimum for google filestore

packages/cluster/main.tf
r33drichards: Not sure

r33drichards:
> 
>
> In more advanced cases, the directory backing the host volume could be a mounted network filesystem, like NFS, or cluster-aware filesystem, like glusterFS. This can enable more complex, automatic failure-recovery scenarios in the event of a node failure.

##

jakubno:
> we could but that doesn't mean we should, there's a few technical reason outlined in this reply https://www.reddit.com/r/aws/comments/dplfoa/comment/f5wht6a/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

That's very general, I think that table metadata and please correct me I may be wrong, I don't know anything about clickhouse, will be rather small and not changing that often

In the nomad document I wouldn't take that as a recommendation (they say "like NFS", that's just an example). Recommending using S3 as a filesystem doesn't make sense, so recommending S3 probably isn't good idea, but again the context matters.

PR #472, opened 2025-03-19
Communicate process kill signals without being blocked by stdout/err reads

0div:
# Description
- [x] use go channels to communicate signals to the handler read loops in envd and broadcast shutdown w/o being blocked by blocked reads
- [x] make local `envd` more similar to what [provision.sh](https://github.com/e2b-dev/infra/blob/f90d8fcd33dee98ace53b22f1310b4d6f2f3a401/packages/template-manager/internal/build/provision.sh#L4) does in template-manager

# Test
I will commit an integration test in E2B js-sdk that goes like this:
```js
test.skipIf(!isIntegrationTest)(
  'kill running nextjs app and make sure process is gone',
  async () => {
    const sandbox = await Sandbox.create(integrationTestTemplate, {
      timeoutMs: 60_000,
    })

    const cmd = await sandbox.commands.run('cd /test-app && next dev', {
      background: true,
    })

    await wait(30_000)

    const pid = cmd.pid
    console.log('PID', pid)

    await sandbox.commands.kill(pid)
    const processes = await sandbox.commands.list()
    console.log('processes:', processes)
    assert.equal(processes.length, 0)
  }
)
```


packages/envd/debug.Dockerfile
```diff
@@ -3,11 +3,22 @@ FROM golang:1.23
...
+
+# Create default user.
 RUN useradd -ms /bin/bash user
+RUN usermod -aG sudo user
+RUN passwd -d user
+RUN echo "user ALL=(ALL:ALL) NOPASSWD: ALL" >>/etc/sudoers
+
+RUN mkdir -p /home/user
+RUN chmod 777 -R /home/user
+RUN chmod 777 -R /usr/local
```
r33drichards, line 21: chmod 777 seems excessive, but i'm low context, you might be better of with `chown -R user /home/user` and `chown -R user /user/local`

0div: I'm just mirroring what is done in provision script used in template manager, doesn't invalidate your comment tho—here's it's solely for local dev, however in template manager it's used for templates

dobrac:
>  await sandbox.commands.kill(pid)
    await wait(10_000)

is it necessary to wait after the kill? Shouldn’t the await be enough?

mlejva:
> await sandbox.commands.kill(pid)
> await wait(10_000)

Agree with @dobrac here. Why to have wait here if we're already await `kill`?

0div:
> > await sandbox.commands.kill(pid)
> > await wait(10_000)
> 
> Agree with @dobrac here. Why to have wait here if we're already await `kill`?

@dobrac @mlejva correct it is not necessary, this is a remnant of older test iterations, i haven't committed this code yet.

mlejva:
@0div just a quick question, would it make sense to add tests for this? Or it makes more sense to have the tests in the SDK repo?

dobrac:
@0div there is not setup for envd client in the integration tests (running command, killing command inside of the sandbox), I can try to set it up so you can take the test to the infra

0div:
> @0div there is no setup for envd client in the integration tests (running command, killing command inside of the sandbox), I can try to set it up so you can take the test to the infra

@dobrac it would be cool to be able to move/adapt these [SDK integration-y tests ](https://github.com/e2b-dev/E2B/tree/main/packages/js-sdk/tests/integration)(at least the setup and template) into this!

packages/envd/debug.Dockerfile
ValentaTomas: For now we leave the permissions as they are, but we will need to go over this when changing build system.

ValentaTomas:
Added an [alternative implementation](https://github.com/e2b-dev/infra/tree/context-kill) building on top of https://github.com/e2b-dev/infra/pull/480 that we should probably use here.

0div:
> Added an [alternative implementation](https://github.com/e2b-dev/infra/tree/context-kill) building on top of #480 that we should probably use here.

Should I merge it into here?

packages/envd/internal/services/process/handler/handler.go
```diff
@@ -104,8 +113,14 @@ func New(ctx context.Context, user *user.User, req *rpc.StartRequest, logger *ze
 	cmd.Env = formattedVars
 
 	outMultiplex := NewMultiplexedChannel[rpc.ProcessEvent_Data](outputBufferSize)
+
 	var outWg sync.WaitGroup
 
+	go func() {
+		outWg.Wait()
+		cancel()
```
ValentaTomas: We want to move the `close(p.DataEvent.Source)` to after the `.Wait` here.

ValentaTomas (approved):

PR #474, opened 2025-03-19
template manager with health endpoint and graceful shutdown

sitole:
Another part needed to make template manager fully rollable. 

Template manager (tm) now offers health endpoint returning 3 states (healthy, draining, unhealthy).
Server implements close method in tm (not used in current version), with following flow:
- mark server draining
- wait for all build jobs to be done
- wait fixed time of 15 seconds so api can fetch state of finished job
- mark service as unhealthy
- graceful shutdown of grcp server

API server template manager grcp client contains new method `healthCheckSync` that runs in background every 5s 
and checks healthy status of template manager. For now connection is not closed if health check fails for too long (will be implemented in next round when support for multiple template managers will be added so we can fallback).

Template creation / deletion checks if template manager is operation and if not will fast fail.


packages/template-manager/internal/server/main.go
```diff
@@ -40,9 +42,11 @@ type serverStore struct {
 	buildLogger      *zap.Logger
 	templateStorage  *template.Storage
 	artifactRegistry *artifactregistry.Client
+	healthyStatus    templatemanager.HealthState
```
jakubno:
```suggestion
	healthStatus    templatemanager.HealthState
```

packages/template-manager/internal/server/health_status.go
```diff
@@ -0,0 +1,18 @@
...
+	"context"
+	"github.com/e2b-dev/infra/packages/shared/pkg/grpc/template-manager"
+	"google.golang.org/protobuf/types/known/emptypb"
+)
+
+func (s *ServerStore) HealthStatus(ctx context.Context, req *emptypb.Empty) (*template_manager.HealthStatusResponse, error) {
+	ctx, ctxSpan := s.tracer.Start(ctx, "health-status-request")
+	defer ctxSpan.End()
+
+	s.logger.Debug("Template manager health status request")
```
jakubno: log the current status

packages/template-manager/internal/server/main.go
```diff
@@ -88,17 +92,41 @@ func New(logger *zap.Logger, buildLogger *zap.Logger) *grpc.Server {
...
+
+		// wait for all builds to finish
+		s.wg.Wait()
+
+		// give some time so all connected services can check build status
+		time.Sleep(15 * time.Second)
+
+		// mark service as unhealthy so now new request will be accepted
+		s.healthyStatus = templatemanager.HealthState_Unhealthy
+		s.server.GracefulStop()
```
jakubno: the `unhealthy` state won't be ever propagated, because you won't be able to do any request here

packages/template-manager/internal/server/main.go
```diff
@@ -88,17 +92,41 @@ func New(logger *zap.Logger, buildLogger *zap.Logger) *grpc.Server {
...
-	})
+		healthyStatus:    templatemanager.HealthState_Healthy,
+		wg:               &sync.WaitGroup{},
+	}
 
+	templatemanager.RegisterTemplateServiceServer(s, store)
 	grpc_health_v1.RegisterHealthServer(s, health.NewServer())
-	return s
+
+	return s, store
```
jakubno, line 111: what's the reason to return store

packages/api/internal/template-manager/client.go
```diff
@@ -34,16 +41,66 @@ func NewClient() (*GRPCClient, error) {
...
+}
+
+func (a *GRPCClient) healthCheckSync() {
+	ticker := time.NewTicker(healthCheckInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+			ctx, ctxCancel := context.WithTimeout(context.Background(), 2*time.Second)
```
jakubno: wouldn't it be better to use context instead of `healthSyncStop`, you could use the context here and cancel this request, because it isn't needed anymore

jakubno (changes requested):

packages/template-manager/internal/server/main.go
sitole: preparation for future step where in main logic for shutdown will be done and will include `store.Close()` call.

packages/template-manager/internal/server/main.go
sitole: understood, but i thing its good to propagate it, or we should just use healthy and draining? Maybe make sense, there will be less logic and less states to handle in api (dont need to check status AND actually handle failing connection because server is not receiving new connections).

jakubno: I think it's cleaner to have it there, but you need to let API to get the info

packages/api/internal/template-manager/client.go
```diff
@@ -14,15 +16,19 @@ import (
...
+	host                = os.Getenv("TEMPLATE_MANAGER_ADDRESS")
+	healthCheckInterval = 5 * time.Second
 )
 
 type GRPCClient struct {
 	Client     template_manager.TemplateServiceClient
 	connection e2bgrpc.ClientConnInterface
+
+	lastHealthCheckAt *time.Time
+	healthy           template_manager.HealthState
```
jakubno: should be probably `health`, `status` or something like that

packages/api/internal/template-manager/client.go
```diff
@@ -34,9 +40,47 @@ func NewClient() (*GRPCClient, error) {
...
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			reqCtx, reqCtxCancel := context.WithTimeout(ctx, 2*time.Second)
+			healthStatus, err := a.Client.HealthStatus(reqCtx, nil)
+			healthCheckAt := time.Now()
```
jakubno, line 69: you can do this right here
```suggestion
			healthCheckAt := time.Now()
			a.lastHealthCheckAt = &healthCheckAt
```

packages/template-manager/internal/server/main.go
```diff
@@ -88,17 +92,40 @@ func New(logger *zap.Logger, buildLogger *zap.Logger) *grpc.Server {
...
+}
+
+func (s *ServerStore) Close(ctx context.Context) error {
+	select {
+	case <-ctx.Done():
+		return errors.New("context cancelled during server graceful shutdown")
+	default:
+		// no new jobs should be started
+		s.healthStatus = templatemanager.HealthState_Draining
+
```
jakubno, line 124: add wait for few seconds here as well to give some time to propagate the status

jakubno (approved):

PR #478, opened 2025-03-19
Add logging around state transitions for TemplatesBuildCache.SetStatus

r33drichards:


packages/api/internal/template-manager/template_manager.go
```diff
@@ -11,7 +11,7 @@ import (
 	templatecache "github.com/e2b-dev/infra/packages/api/internal/cache/templates"
 	"github.com/e2b-dev/infra/packages/api/internal/utils"
 	"github.com/e2b-dev/infra/packages/shared/pkg/db"
-	"github.com/e2b-dev/infra/packages/shared/pkg/grpc/template-manager"
+	template_manager "github.com/e2b-dev/infra/packages/shared/pkg/grpc/template-manager"
```
r33drichards, line 14: this was added by auto formatter, we should probably be linting this somewhere

PR #480, opened 2025-03-19
Pass SDK timeout to envd command timeout 

0div:
# Description 
Get the timeout from SDK in the GRPC header and use it as a `context.WithTimeout` in `exec.CommandContext` 

This is still WIP bc we need to test some edge cases mentioned 

> Some concerns from Tomas:

> * We should handle the 0 timeout that can be passed from SDK properly, as no timeout

We handle no timeout with 0 value by default from `InitializeStartProcess` 

> * As the timeout is also enforced by the ctx passed to the request (which has the same value as the header we use to parse the timeout from), creating a context with timeout might result in a race condition between the request ctx created from the header by connectRpc and command ctx created from header by us. This means we could potentially not return process end event as the connection would end right before the command is killed by the timeout

I've tried various combinations of command duration, `timeoutMs` and `requestTimeoutMs` as follows
```js
    const cmd = await sandbox.commands.run('sleep {command_duration} && echo "hello"', {
      timeoutMs: {timeoutMs},
      requestTimeoutMs: {requestTimeoutMs}
    })
```
with same values, slightly shorter request timeout vs method timeout and the command seems to complete when given enough time and be killed as expected depending on the test cases. 

> * If we use the exec.CommandContext the command will be killed by signal—we should make sure this is handled properly as there could be some deadlock as we experienced with .kill

Kill method works as expected if command times out before or not; additionally checking `ps` inside the sandbox shows that the process gets killed at the OS level—either by the method or the timeout—and not only in the envd  `processes` map


packages/envd/internal/services/process/start.go
```diff
@@ -14,6 +17,8 @@ import (
 	"connectrpc.com/connect"
 )
 
+const DefaultProcessTimeout = 60 * time.Second
```
ValentaTomas: I think the default timeout should stay in the SDK. If the header is not present assume no timeout.

packages/envd/internal/services/process/start.go
```diff
@@ -204,3 +213,18 @@ func (s *Service) handleStart(ctx context.Context, req *connect.Request[rpc.Star
...
+	if timeoutHeader == "" {
+		return DefaultProcessTimeout, nil
+	}
+
+	timeout, err := strconv.Atoi(timeoutHeader)
+	if err != nil {
+		return DefaultProcessTimeout, err
+	}
+
+	return time.Duration(timeout-500) * time.Millisecond, nil
```
ValentaTomas: I think this change was for testing.

packages/envd/internal/services/process/start.go
```diff
@@ -204,3 +213,18 @@ func (s *Service) handleStart(ctx context.Context, req *connect.Request[rpc.Star
 		return nil
 	}
 }
+
+func determineTimeoutFromHeader(header http.Header) (time.Duration, error) {
+	timeoutHeader := header.Get("Connect-Timeout-Ms")
+
+	if timeoutHeader == "" {
+		return DefaultProcessTimeout, nil
```
ValentaTomas: Let's not assume timeout if no header.

packages/envd/internal/services/process/start.go
```diff
@@ -204,3 +213,18 @@ func (s *Service) handleStart(ctx context.Context, req *connect.Request[rpc.Star
...
+	}
+
+	timeout, err := strconv.Atoi(timeoutHeader)
+	if err != nil {
+		return DefaultProcessTimeout, err
```
ValentaTomas: If header cannot be parsed properly we should just error with a proper ConnectRpc error.

ValentaTomas:
Pinning the issue notes here:

> As the timeout is also enforced by the ctx passed to the request (which has the same value as the header we use to parse the timeout from), creating a context with timeout might result in a race condition between the request ctx created from the header by connectRpc and command ctx created from header by us. This means we could potentially not return process end event as the connection would end right before the command is killed by the timeout
> 
> If we use the exec.CommandContext the command will be killed by signal—we should make sure this is handled properly as there could be some deadlock as we experienced with .kill

ValentaTomas (approved):

PR #485, opened 2025-03-20
Setup client for envd in integration tests

dobrac:
Setup client for envd (http and grpc) in the integration tests. 

**Integration tests**
Uses the new `session-proxy` in the orchestrator running on port `5007`. 

**Locally**
You can use client proxy with "dummy" sandbox url (like `https://dummy.domain.dev`), it will be replaced by correct sandbox url in runtime.


packages/envd/go.mod
```diff
@@ -6,7 +6,7 @@ toolchain go1.23.7
 
 require (
 	connectrpc.com/authn v0.1.0
-	connectrpc.com/connect v1.16.2
+	connectrpc.com/connect v1.18.1
```
0div: despite this being a minor upgrade, were all the SDK tested with this version? asking bc their client gRPC implementations can be fragile.

0div (changes requested):
This is neat! 

I was however hoping we could also run integration tests with the SDK themselves? 
with a custom template and everything, like here: https://github.com/e2b-dev/E2B/tree/main/packages/js-sdk/tests/integration. That way we don't have to write these tests in separate E2B/ repo

packages/envd/go.mod
dobrac: reverted to v1.16.2, just in case

dobrac:
> I was however hoping we could also run integration tests with the SDK themselves?
with a custom template and everything, like here: https://github.com/e2b-dev/E2B/tree/main/packages/js-sdk/tests/integration. That way we don't have to write these tests in separate E2B/ repo

I think we can migrate those here. The issue with using SDK is that in order for you to write tests for infrastructure, you need to have the SDK updated first, but in order to have the SDK updated, you need to have the infrastructure ready

0div:
> > I was however hoping we could also run integration tests with the SDK themselves?
> > with a custom template and everything, like here: https://github.com/e2b-dev/E2B/tree/main/packages/js-sdk/tests/integration. That way we don't have to write these tests in separate E2B/ repo
> 
> I think we can migrate those here. The issue with using SDK is that in order for you to write tests for infrastructure, you need to have the SDK updated first, but in order to have the SDK updated, you need to have the infrastructure ready

@dobrac 

> but in order to have the SDK updated, you need to have the infrastructure ready

The integration tests i would want to run here would be for changes in infra that don't require SDK updates. 
This would be to make sure that latest SDK still works as expected with newly introduced infra changes.

dobrac:
> The integration tests i would want to run here would be for changes in infra that don't require SDK updates.
This would be to make sure that latest SDK still works as expected with newly introduced infra changes.

Unfortunately that is not supported right now as it would require having the infra completely set up (domain, etc) - which is not the case here. What we can do right now (or in the near future) is to run these SDK tests against staging. But, every new feature would require SDK change anyway.

I think we can iterate on it when anyone has better idea how to do it

tests/integration/Makefile
```diff
@@ -1,14 +1,25 @@
...
+	# Copy the envd grpc clients
+	rm -rf internal/envd/
+	mkdir -p internal/envd/
+	cp -r ../../packages/envd/internal/services/spec/ internal/envd/.
+	find internal/. -type f -exec perl -pi -e 's|github.com/e2b-dev/infra/packages/envd/internal/services/spec/|github.com/e2b-dev/infra/tests/integration/internal/envd/|g' {} +
+
+	# Create envd HTTP client
+	mkdir -p internal/envd/api/
+	$(codegen) -generate client --package api $(openapi_envd) > internal/envd/api/client.gen.go
+	$(codegen) -generate models --package api $(openapi_envd) > internal/envd/api/models.gen.go
```
0div, line 22: are these only used for the integration test? in which case i would probably make it a separate directive (something like "setup-test" or something) to not add too much unused files for dev/prod

0div (changes requested):
Could you add a description somewhere (README?) on how to use this new feature? thanks

tests/integration/Makefile
dobrac: yes, they are used only for integration tests, and they are already only in the integration tests folder. This is part of the integration tests Makefile. Not sure what do you mean by adding unused files for dev/prod

dobrac:
> Could you add a description somewhere (README?) on how to use this new feature? thanks

done

jakubno (approved):

PR #488, opened 2025-03-21
Use cache for dependencies in golang deps check workflow

sitole:


sitole:
I think, is right decision to let cache configuration be here, but now as iam checking workflow runs before and after this change it didn't help much.

| Cache | Set up Go + Post Setup | Integrity check | Total time |
|-------|------------------------|-----------------|------------|
| No    | 11s + 10s              | 22s             | 43s        |
| Yes   | 29s + 7s               | 1s              | 37s        |

Cache improved integrity run itself but now time is spend on downloading and pushing cache.

dobrac:
Based on a brief discussion with @sitole, an improved workflow which requires depencencies check to finish sucessfully before running tests: https://github.com/e2b-dev/infra/pull/489

ValentaTomas (approved):

sitole:
Closed based on https://github.com/e2b-dev/infra/pull/489 that already contains deps cache

PR #489, opened 2025-03-21
Improve workflows caching

dobrac:
Improves GitHub workflows on pull requests by first running the go integrity test, then unit and integration tests. Tests reuse mod cache (replaces https://github.com/e2b-dev/infra/pull/488) from the integrity test and are not run if integrity test fails.

**This PR includes update to Go version 1.23.7 (from 1.23.0).** It's there because version 1.23.0 has a bug where caches don't work properly.


sitole:
Like it, whole flow is now much more clear.

dobrac:
One more thing, I think we should also add here build cache

mishushakov (approved):

PR #490, opened 2025-03-21
Shutdown logic for template manager

sitole:
Adds safe shutdown for template manager that waits for all jobs to be finished. Adds kill timeout for template manager service in nomad.

Currently depends on https://github.com/e2b-dev/infra/pull/508 to remove cherry picks from this PR.

### DIFF
--- a/packages/nomad/template-manager.hcl
+++ b/packages/nomad/template-manager.hcl
@@ -5,7 +5,9 @@ job "template-manager" {
 
 %{ if update_stanza }
   update {
-     progress_deadline = "20m"
+      auto_promote      = true # Whether to promote the canary if the rest of the group is not healthy
+      canary            = 1 # Allows to spawn new version of the service before killing the old one
+      progress_deadline = "20m" # Deadline for the update to be completed
   }
 %{ endif }
 
--- a/packages/nomad/template-manager.hcl
+++ b/packages/nomad/template-manager.hcl
@@ -3,6 +3,12 @@ job "template-manager" {
   node_pool  = "build"
   priority = 70
 
+%{ if update_stanza }
+  update {
+     progress_deadline = "20m"
+  }
+%{ endif }
+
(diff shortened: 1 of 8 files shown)


packages/nomad/template-manager.hcl
```diff
@@ -1,126 +1,66 @@
...
       }
     }
 
     task "start" {
       driver = "raw_exec"
 
+%{ if update_stanza }
+      # If we need more than 30s we will need to update the max_kill_timeout in nomad
+      # https://developer.hashicorp.com/nomad/docs/configuration/client#max_kill_timeout
+      kill_timeout = "10m"
```
sitole: there is maximum kill timeout 10m for cluster now. I guess in some cases for build finish we need more

sitole:
Cherry picks fixes for template builder failing state https://github.com/e2b-dev/infra/pull/508.

packages/nomad/template-manager.hcl
jakubno: increase it then 🙏🏻

packages/nomad/template-manager.hcl
```diff
@@ -1,126 +1,66 @@
...
-        port         = var.port
+        port         = "${port}"
       }
     }
 
     task "start" {
       driver = "raw_exec"
 
+%{ if update_stanza }
+      # If we need more than 10m we will need to update the max_kill_timeout in nomad
```
jakubno: don't you need `canary = 1`? also i think we want `auto_promote=True`
https://developer.hashicorp.com/nomad/docs/job-specification/update#auto_promote

sitole: fixed

packages/nomad/template-manager.hcl
sitole: done

jakubno (approved):

PR #492, opened 2025-03-21
Track applied migrations

dobrac:
Track applied migrations in PostgreSQL using the https://github.com/golang-migrate/migrate library (already used in ClickHouse db migrations).

If used on already migrated DB, the migrations will fail (because they are not properly idempotent). There will be created table public.schema_migrations. You need to set the latest version to the latest applied migration timestamp and dirty to false.

The down migration will migrate only one version down. There is an empty down migration for the latest one, so you won't go to "no version" from the latest one by accident.

This change is necessary for automated deployment.


dobrac:
The down migration is basically not implemented right now (so it doesn't do anything). It would have to be defined by the .down.sql files. I can remove it from the command completely so noone can run it

dobrac:
Changed to go down only by one step

jakubno (approved):

PR #500, opened 2025-03-25
add keeper to clickhouse for fault tolerance and availability

r33drichards:
This pull request includes significant changes to the infrastructure and configuration of the API cluster, particularly focusing on the integration of NFS storage and the deployment of ClickHouse. The most important changes include the addition of a Filestore instance, updates to the API cluster environment variables, and modifications to the Nomad configuration scripts.

### Infrastructure Enhancements:
* Added a Filestore instance for NFS in `packages/cluster/main.tf` to provide shared storage for the API cluster.
* Updated environment variables in `packages/cluster/main.tf` to include NFS-related settings, such as `NFS_IP`, `NFS_SHARE`, and `NFS_MOUNT_PATH`.

### Configuration Updates:
* Modified `packages/cluster/scripts/run-api-nomad.sh` to create directories for ClickHouse data and configure host volumes for ClickHouse server and keeper. [[1]](diffhunk://#diff-d26b132d85970e632a7e5e9bf041c62cb8d20b7bc8ff164deb8ac5188e59412dR139-R143) [[2]](diffhunk://#diff-d26b132d85970e632a7e5e9bf041c62cb8d20b7bc8ff164deb8ac5188e59412dR162-R172)
* Added NFS client installation and NFS share mounting steps in `packages/cluster/scripts/start-api.sh` to ensure the NFS storage is properly mounted on the API instances.

### ClickHouse Deployment:
* add feature flag to clickhouse deployment to not confuse our users
* Introduced a new Nomad job configuration for ClickHouse in `packages/nomad/clickhouse.hcl`, including settings for both ClickHouse servers and keepers, with detailed configurations for storage, logging, and replication. [[1]](diffhunk://#diff-70fdcd05684c7d2e46c882c06fbf8a95ea1c14315311614af6d9946eef495a7bR1-R300) [[2]](diffhunk://#diff-70fdcd05684c7d2e46c882c06fbf8a95ea1c14315311614af6d9946eef495a7bL132-R334)
* Updated `packages/nomad/main.tf` to include parameters for the number of ClickHouse keepers and servers.

### DIFF
--- a/packages/cluster/scripts/run-api-nomad.sh
+++ b/packages/cluster/scripts/run-api-nomad.sh
@@ -136,11 +136,6 @@ function generate_nomad_config {
   instance_region=$(get_instance_region)
   zone=$(get_instance_zone)
 
-  # Create a directory for ClickHouse data if it doesn't exist
-  # since this config is dependent on the volume being present
-  clickhouse_data_dir="/mnt/nfs/clickhouse/data"
-  mkdir -p "$clickhouse_data_dir/clickhouse-server"
-  mkdir -p "$clickhouse_data_dir/clickhouse-keeper"
 
   log_info "Creating default Nomad config file in $config_path"
   cat >"$config_path" <<EOF
@@ -160,15 +155,7 @@ leave_on_terminate = true
 
 client {
 
-  host_volume "clickhouse-server" {
-    path      = "$clickhouse_data_dir/clickhouse-server"
-    read_only = false
-  }
 
-  host_volume "clickhouse-keeper" {
-    path      = "$clickhouse_data_dir/clickhouse-keeper"
-    read_only = false
-  }
   
   enabled = true
   node_pool = "api"
--- a/packages/cluster/scripts/run-api-nomad.sh
+++ b/packages/cluster/scripts/run-api-nomad.sh
@@ -299,7 +299,11 @@ function run {
     shift
   done
 
-
+  export ARCH_CNI=$([ $(uname -m) = aarch64 ] && echo arm64 || echo amd64)
+  export CNI_PLUGIN_VERSION=v1.6.2
+  curl -L -o cni-plugins.tgz "https://github.com/containernetworking/plugins/releases/download/${CNI_PLUGIN_VERSION}/cni-plugins-linux-${ARCH_CNI}-${CNI_PLUGIN_VERSION}".tgz &&
+    sudo mkdir -p /opt/cni/bin &&
+    sudo tar -C /opt/cni/bin -xzf cni-plugins.tgz
 
   use_sudo="true"
 
--- a/packages/cluster/scripts/run-api-nomad.sh
+++ b/packages/cluster/scripts/run-api-nomad.sh
@@ -136,6 +136,11 @@ function generate_nomad_config {
   instance_region=$(get_instance_region)
   zone=$(get_instance_zone)
 
+  # Create a directory for ClickHouse data if it doesn't exist
+  # since this config is dependent on the volume being present
+  clickhouse_data_dir="/mnt/nfs/clickhouse/data"
+  mkdir -p "$clickhouse_data_dir/clickhouse-server"
+  mkdir -p "$clickhouse_data_dir/clickhouse-keeper"
 
   log_info "Creating default Nomad config file in $config_path"
   cat >"$config_path" <<EOF
@@ -155,7 +160,15 @@ leave_on_terminate = true
 
 client {
 
+  host_volume "clickhouse-server" {
+    path      = "$clickhouse_data_dir/clickhouse-server"
+    read_only = false
+  }
 
+  host_volume "clickhouse-keeper" {
+    path      = "$clickhouse_data_dir/clickhouse-keeper"
+    read_only = false
+  }
   
   enabled = true
   node_pool = "api"
--- a/packages/cluster/scripts/run-api-nomad.sh
+++ b/packages/cluster/scripts/run-api-nomad.sh
@@ -139,7 +139,8 @@ function generate_nomad_config {
   # Create a directory for ClickHouse data if it doesn't exist
   # since this config is dependent on the volume being present
   clickhouse_data_dir="/mnt/nfs/clickhouse/data"
-  mkdir -p "$clickhouse_data_dir"
+  mkdir -p "$clickhouse_data_dir/clickhouse-server"
+  mkdir -p "$clickhouse_data_dir/clickhouse-keeper"
 
   log_info "Creating default Nomad config file in $config_path"
   cat >"$config_path" <<EOF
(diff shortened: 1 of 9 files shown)


r33drichards:
E2B-1787

ValentaTomas:
A reason for not merging this right now to `main` could be that with the API script modification, this would be mounting the NFS in the API pool (and also adding the deps, etc) and then we would be just removing it later. It might be ok with correct upgrade strategy, but not sure what is there now.

packages/cluster/scripts/run-api-nomad.sh
```diff
@@ -136,6 +136,11 @@ function generate_nomad_config {
   instance_region=$(get_instance_region)
   zone=$(get_instance_zone)
 
+  # Create a directory for ClickHouse data if it doesn't exist
+  # since this config is dependent on the volume being present
+  clickhouse_data_dir="/mnt/nfs/clickhouse/data"
+  mkdir -p "$clickhouse_data_dir/clickhouse-server"
```
ValentaTomas: These might not be necessary as the direct volume mounts might be creating their own dirs. I observed something the implied this, but have not tested specifically.

packages/cluster/scripts/run-api-nomad.sh
```diff
@@ -154,6 +159,17 @@ leave_on_interrupt = true
 leave_on_terminate = true
 
 client {
+
```
ValentaTomas, line 157: This is no longer used with the direct volume mounts.

packages/cluster/scripts/run-api-nomad.sh
r33drichards: better to be safe than sorry?

packages/cluster/scripts/run-api-nomad.sh
```diff
@@ -283,7 +299,11 @@ function run {
     shift
   done
 
-
+  export ARCH_CNI=$([ $(uname -m) = aarch64 ] && echo arm64 || echo amd64)
```
ValentaTomas, line 289: We might be able to get rid of "bridge" in the job and then remove this.

packages/cluster/scripts/start-api.sh
```diff
@@ -19,6 +18,19 @@ exec > >(tee /var/log/user-data.log | logger -t user-data -s 2>/dev/console) 2>&
 ulimit -n 1048576
 export GOMAXPROCS='nproc'
 
+# Install NFS client
+apt-get update
```
ValentaTomas, line 22: Ideally move this to the build step of the cluster image.

packages/cluster/scripts/start-api.sh
```diff
@@ -19,6 +18,19 @@ exec > >(tee /var/log/user-data.log | logger -t user-data -s 2>/dev/console) 2>&
...
+apt-get install -y nfs-common
+
+# Create mount directory
+mkdir -p "${NFS_MOUNT_PATH}"
+
+# Mount NFS share
+mount -t nfs -o rw,hard,intr "${NFS_IP}:${NFS_SHARE}" "${NFS_MOUNT_PATH}"
```
ValentaTomas, line 29: Will this script fail here on restart if we added the fstab for persistence?

packages/nomad/clickhouse.hcl
```diff
@@ -1,119 +1,304 @@
...
 
     update {
-      max_parallel     = 2
-      min_healthy_time = "30s"
-      healthy_deadline = "4m"
+      max_parallel = 0
+    }
+
+    network {
+      mode = "bridge"
```
ValentaTomas, line 16: As mentioned before, this might not be needed.

packages/nomad/clickhouse.hcl
```diff
@@ -1,119 +1,304 @@
...
+      }
 
-
-        volumes = [
-          "local/config.xml:/etc/clickhouse-server/config.d/gcs.xml",
-          # disabled while testing but will pass password to orchestrator in the future
-          "local/users.xml:/etc/clickhouse-server/users.d/users.xml",
-        ]
+      resources {
+        cpu    = 500
```
ValentaTomas, line 215: Current resource constraints might not be recommended—how to allow it to grow to maximum on the node it is on?

r33drichards: it seems to do that by default

packages/nomad/clickhouse.hcl
```diff
@@ -1,119 +1,304 @@
...
+        ]
+
+        volumes = [
+          "/mnt/nfs/clickhouse/data/clickhouse-keeper-${i + 1}:/var/lib/clickhouse",
+          "local/keeper.xml:/etc/clickhouse-server/config.d/keeper_config.xml",
+        ]
+      }
+
+      resources {
+        cpu    = 400
```
ValentaTomas, line 62: What are the recommended resources for keepers?

ValentaTomas:
Thinking about the NFS mount in the API cluster—ideally though we would still have it there, because the clickhouse keepers would be in there too. Otherwise we will need 3 nodes in the new pool for the keepers spread.

packages/nomad/main.tf
```diff
@@ -419,6 +419,7 @@ resource "google_storage_hmac_key" "clickhouse_hmac_key" {
 
 # Add this with your other Nomad jobs
 resource "nomad_job" "clickhouse" {
+  count = var.clickhouse_enabled ? 1 : 0
   jobspec = templatefile("${path.module}/clickhouse.hcl", {
     zone                = var.gcp_zone
     clickhouse_version  = "25.1.5.31" # Or make this a variable
```
ValentaTomas, line 425: Unused right now (not propagated to the job)

r33drichards: wdym?

packages/nomad/clickhouse.hcl
r33drichards: there are no general recommendations for this afaik, there docs largely mention "it depends" https://clickhouse.com/docs/guides/sizing-and-hardware-recommendations#how-many-cpu-cores-should-i-use

r33drichards: > The number of CPUs you should use depends on your workload. However, we generally recommend the following memory to CPU core ratios based on your CPU type:

[M-type](https://aws.amazon.com/ec2/instance-types/) (general purpose use cases): 4:1 memory to CPU core ratio
[R-type](https://aws.amazon.com/ec2/instance-types/#Memory_Optimized) (data warehousing use cases): 8:1 memory to CPU core ratio
[C-type](https://aws.amazon.com/ec2/instance-types/#Compute_Optimized) (compute-optimized use cases): 2:1 memory to CPU core ratio
As an example, when using M-type CPUs, we recommend provisioning 100GB of memory per 25 CPU cores. To determine the amount of memory appropriate for your application, profiling your memory usage is necessary. You can read [this guide on debugging memory issues](https://clickhouse.com/docs/guides/developer/debugging-memory-issues) or use the [built-in observability dashboard](https://clickhouse.com/docs/operations/monitoring) to monitor ClickHouse.

packages/cluster/scripts/run-api-nomad.sh
r33drichards: i'll try this out

packages/cluster/scripts/start-api.sh
r33drichards: where does this happen?

PR #503, opened 2025-03-26
Fix multiple requests sharing same trace ID

jakubno:
# Description

If the header `traceparent` was set (probably automatically from some of our users), the context was used on multiple requests.


packages/shared/pkg/gin_utils/middleware/otel/tracing/middleware.go
```diff
@@ -55,11 +55,18 @@ func Middleware(service string) gin.HandlerFunc {
...
+			c.Request = c.Request.WithContext(ctx)
 		}()
-		ctx := cfg.Propagators.Extract(savedCtx, propagation.HeaderCarrier(c.Request.Header))
+
+		// Remove traceparent (it's coming from our users and it can cause multiple calls share the same trace ID)
+		if c.Request.Header.Get("traceparent") != "" {
+			c.Request.Header.Del("traceparent")
+		}
+		// No need for calling Extract, as we are not expecting any incoming trace
+		// ctx := cfg.Propagators.Extract(savedCtx, propagation.HeaderCarrier(c.Request.Header))
```
dobrac, line 68: are we not expecting the trace header when having trace on the API and then the same one on the orchestrator?

dobrac (changes requested):

packages/shared/pkg/gin_utils/middleware/otel/tracing/middleware.go
dobrac: not relevant as it's used in the API only -> maybe moving the logic to the API directly, or adding a parameter, so it's transparent for new usages

jakubno: Moved middleware to API

PR #508, opened 2025-03-27
Fixed slow waiting for template build status to be fetched

sitole:


sitole:
In current implementation it misses both retries when build is not success/failed. When its in building phase it will return and 
another round will wait for one minute (we are doing fetch for not processed builds every minutes). This is not wanted, because user in SDK will be stuck with minute of waiting before knowing proper build status.

New implementation is replacing upper layer with ticket and for that can now decide between error 
I was fixing error with two layers of retries (one used for pulling and one for error retry), i replaced it with just for with ticker and inner req error retry. When there was error in previous implementation it caused for to be returned

PR #511, opened 2025-03-27
Change client instance group from zonal to regional e2b 1704

jakubno:
# Description 

Create a regional managed instance group instead of zonal one


ValentaTomas:
Will we be able to smoothly roll this?

jakubno:
If we are okay with different name for instance group (only in GCP), that transition should be easy

the problem is with orchestrator job definition...

ValentaTomas:
:| I see

ValentaTomas:
Pinging this as now we should have the orchestrator jobs handled.

.env.template
```diff
@@ -19,6 +19,8 @@ CLIENT_MACHINE_TYPE=
 CLIENT_CLUSTER_SIZE=
 # Max number of additional instances if the CPU usage is above 80%, e.g. 0
 CLIENT_CLUSTER_AUTO_SCALING_MAX=
+# e.g.
```
dobrac:
```suggestion
# e.g. 1
```

packages/cluster/client/main.tf
```diff
@@ -151,9 +215,5 @@ resource "google_compute_instance_template" "client" {
   # which this Terraform resource depends will also need this lifecycle statement.
   lifecycle {
     create_before_destroy = true
-
-    # TODO: Temporary workaround to avoid unnecessary updates to the instance template.
-    #  This should be removed once cluster size is removed from the metadata
-    ignore_changes = [metadata]
```
dobrac, line 157: is it removed from the metadata?

packages/cluster/variables.tf
```diff
@@ -7,11 +7,6 @@ variable "environment" {
   type        = string
 }
 
-variable "notification_email_secret_version" {
```
dobrac, line 10: is this related to this change?

packages/nomad/template-manager.hcl
```diff
@@ -1,5 +1,4 @@
 job "template-manager" {
-  datacenters = ["${gcp_zone}"]
```
dobrac: why remove this only in the template-manager? Also, shouldn't we change it to the region?

dobrac (changes requested):
Also, please fix the `connect-orchestrator` command

packages/cluster/client/main.tf
jakubno: yes

packages/cluster/variables.tf
jakubno: not really, but I am adding new variable and this isn't used anymore

packages/nomad/template-manager.hcl
jakubno: Reverted as template manager runs on builder node, which is still in the same zone

tests/integration/Makefile
```diff
@@ -56,5 +56,7 @@ connect-orchestrator:
 		--filter="name~'^.*client.*'" \
 		--format='value(name)' \
 		--zones=$(GCP_ZONE) | head -n1) && \
-	INSTANCE_ID=$$(gcloud compute instance-groups list-instances "$$CLIENT_IG" --project=$(GCP_PROJECT_ID) --zone=$(GCP_ZONE) --format='value(instance)' | head -n1) && \
+	INSTANCE_INFO=$$(gcloud compute instance-groups list-instances "$$CLIENT_IG" --project=$(GCP_PROJECT_ID) --zone=$(GCP_ZONE) --format='value(instance,zone)' | head -n1) && \
```
dobrac: doesn't this list only in the specified gcp_zone?

.env.template
```diff
@@ -19,6 +19,8 @@ CLIENT_MACHINE_TYPE=
 CLIENT_CLUSTER_SIZE=
 # Max number of additional instances if the CPU usage is above 80%, e.g. 0
 CLIENT_CLUSTER_AUTO_SCALING_MAX=
+# e.g. 1
+CLIENT_REGIONAL_CLUSTER_SIZE=
```
dobrac, line 23: can this (and later on also CLIENT_CLUSTER_SIZE) be just undefined? Would that result to default 0?

jakubno: I had to refactor how we passed the env variables. With the old setup, we set the variable with an empty string, which was used instead of the default value.

tests/integration/Makefile
dobrac: the zone is still in the INSTANCE_INFO

.env.template
dobrac: nice, could you also set default 0 for the CLIENT_CLUSTER_SIZE?

dobrac (approved):

PR #512, opened 2025-03-28
Use chown of chmod for user home dir to fix permission and ownership issues

r33drichards:
if a user includes the following line in their  Dockerfile 

```
RUN chmod 666 -R /home/user/app
```

these  permissions are overwritten by the following line later in our provision.sh script 
```
chmod 777 -R /home/user
```

This change prevents this from happening while retaining user ownership of files copied into that directory using a flow similar to the following 

```
FROM python:3.12-slim 

RUN mkdir -p /home/user 

COPY startup.sh /home/user/startup.sh

```

### DIFF
--- a/packages/template-manager/internal/build/provision.sh
+++ b/packages/template-manager/internal/build/provision.sh
@@ -117,7 +117,15 @@ EOF
 passwd -d root
 
 # Create default user.
-adduser --disabled-password --gecos "" user
+# if the /home/user directory exists, we copy the skeleton files to it because the adduser command
+# will ignore the directory if it exists, but we want to include the skeleton files in the home directory
+# in our case.
+ADDUSER_OUTPUT=$(adduser -disabled-password --gecos "" user 2>&1)
+if echo "$ADDUSER_OUTPUT" | grep -q "The home directory \`/home/user' already exists"; then
+    # Copy skeleton files if they don't exist in the home directory
+    cp -rn /etc/skel/. /home/user/
+fi
+
 usermod -aG sudo user
 passwd -d user
 echo "user ALL=(ALL:ALL) NOPASSWD: ALL" >>/etc/sudoers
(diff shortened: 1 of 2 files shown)


ValentaTomas:
I think this should work—the `user` will still have permissions to the `/home/user` directory so all current interactions should be the same, and the `root` should have access there regardless. I thought there was some problem for using `chown` during the provisioning (@jakubno can you remember?) but it looks like it is working.

Before we deploy this we should finish rolling the orchestrator nodes in prod so the previous change to template manager (ignore empty blocks) won't break orchestrators.

mlejva:
Good by me if this doesn't break anything for current users. 

@r33drichards it's not clear from the title of this PR and the Linear ticket isn't public for non team people - please write a short example what this PR will change for users (seems just chmod but would like to be sure)

dobrac:
Is it possible to add some integration tests so we know that it doesn’t break the permissions for users and it folder permissions are correctly set?

mlejva:
> Is it possible to add some integration tests so we know that it doesn’t break the permissions for users and the folder permissions are correctly set?

Good point. Please add integration tests @r33drichards !

r33drichards:
> doesn’t break the permissions for users
this pr by design breaks that backwards compatibility, so its not worth testing this. There was a changeset for not breaking backwards compatibility but @ValentaTomas mentioned that this was not necessary. I'm happy to add some for the behavior in this fix though

r33drichards:
> what happens with /home/user if the user didn't run chmod themselves? Will they still be able to write, read, execute there?

yes

ValentaTomas:
The write/read should be more of less the same, the only situation that I think could reasonably happen that would break would be if there was a script/bin you copy to the /home/user and it was not chmod +x at any point before or during the build it would not have execution priviledges, which makes sense, but it is possible that somebody depends on that.

ValentaTomas:
Note—the orchestrator nodes are redeployed, so we should be able to update template manager now.

r33drichards:
@ValentaTomas @dobrac @mlejva added integration test

jakubno (approved):

PR #517, opened 2025-03-31
Improve NBD handling

ValentaTomas:
We are experiencing a kernel NULL dereference that might be caused by incorrect NBD handling.

### DIFF
--- a/packages/orchestrator/internal/sandbox/nbd/dispatch.go
+++ b/packages/orchestrator/internal/sandbox/nbd/dispatch.go
@@ -76,7 +76,7 @@ func NewDispatch(ctx context.Context, fp io.ReadWriteCloser, prov Provider) *Dis
 	return d
 }
 
-func (d *Dispatch) Wait() {
+func (d *Dispatch) Drain() {
 	d.shuttingDownLock.Lock()
 	d.shuttingDown = true
 	defer d.shuttingDownLock.Unlock()
--- a/packages/orchestrator/internal/sandbox/nbd/dispatch.go
+++ b/packages/orchestrator/internal/sandbox/nbd/dispatch.go
@@ -3,12 +3,15 @@ package nbd
 import (
 	"context"
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"io"
 	"sync"
 
 	"go.uber.org/zap"
 )
+
+var ErrShuttingDown = errors.New("shutting down. Cannot serve any new requests")
 
 type Provider interface {
 	io.ReaderAt
@@ -55,7 +58,9 @@ type Dispatch struct {
 	writeLock        sync.Mutex
 	prov             Provider
 	pendingResponses sync.WaitGroup
-	pendingMu        sync.Mutex
+	shuttingDown     bool
+	shuttingDownLock sync.Mutex
+	fatal            chan error
 }
 
 func NewDispatch(ctx context.Context, fp io.ReadWriteCloser, prov Provider) *Dispatch {
@@ -64,6 +69,7 @@ func NewDispatch(ctx context.Context, fp io.ReadWriteCloser, prov Provider) *Dis
 		fp:             fp,
 		prov:           prov,
 		ctx:            ctx,
+		fatal:          make(chan error, 1),
 	}
 
 	binary.BigEndian.PutUint32(d.responseHeader, NBDResponseMagic)
@@ -71,8 +77,9 @@ func NewDispatch(ctx context.Context, fp io.ReadWriteCloser, prov Provider) *Dis
 }
 
 func (d *Dispatch) Wait() {
-	d.pendingMu.Lock()
-	defer d.pendingMu.Unlock()
+	d.shuttingDownLock.Lock()
+	d.shuttingDown = true
+	defer d.shuttingDownLock.Unlock()
 
 	// Wait for any pending responses
 	d.pendingResponses.Wait()
@@ -124,9 +131,10 @@ func (d *Dispatch) Handle() error {
 		rp := 0
 	process:
 		for {
-
-			// If the context has been cancelled, quit
+			// Check if there is a fatal error from an async read/write to return
 			select {
+			case err := <-d.fatal:
+				return err
 			case <-d.ctx.Done():
 				return d.ctx.Err()
 			default:
@@ -173,14 +181,13 @@ func (d *Dispatch) Handle() error {
 					}
 				case NBDCmdTrim:
 					rp += 28
-					err = d.cmdTrim(request.Handle, request.From, request.Length)
+					err := d.cmdTrim(request.Handle, request.From, request.Length)
 					if err != nil {
 						return err
 					}
 				default:
 					return fmt.Errorf("nbd not implemented %d", request.Type)
 				}
-
 			} else {
 				break // Try again when we have more data...
 			}
@@ -194,9 +201,14 @@ func (d *Dispatch) Handle() error {
 }
 
 func (d *Dispatch) cmdRead(cmdHandle uint64, cmdFrom uint64, cmdLength uint32) error {
-	d.pendingMu.Lock()
-	d.pendingResponses.Add(1)
-	d.pendingMu.Unlock()
+	d.shuttingDownLock.Lock()
+	if !d.shuttingDown {
+		d.pendingResponses.Add(1)
+	} else {
+		d.shuttingDownLock.Unlock()
+		return ErrShuttingDown
+	}
+	d.shuttingDownLock.Unlock()
 
 	performRead := func(handle uint64, from uint64, length uint32) error {
 		errchan := make(chan error)
@@ -215,20 +227,21 @@ func (d *Dispatch) cmdRead(cmdHandle uint64, cmdFrom uint64, cmdLength uint32) e
 		case e = <-errchan:
 		}
 
-		errorValue := uint32(0)
 		if e != nil {
-			errorValue = 1
-			data = make([]byte, 0) // If there was an error, don't send data
+			return d.writeResponse(1, handle, []byte{})
 		}
-		return d.writeResponse(errorValue, handle, data)
+		return d.writeResponse(0, handle, data)
 	}
 
 	go func() {
 		err := performRead(cmdHandle, cmdFrom, cmdLength)
 		if err != nil {
-			zap.L().Error("nbd error cmd read", zap.Error(err))
+			select {
+			case d.fatal <- err:
+			default:
+				zap.L().Error("nbd error cmd read", zap.Error(err))
+			}
 		}
-
 		d.pendingResponses.Done()
 	}()
 
@@ -236,14 +249,19 @@ func (d *Dispatch) cmdRead(cmdHandle uint64, cmdFrom uint64, cmdLength uint32) e
 }
 
 func (d *Dispatch) cmdWrite(cmdHandle uint64, cmdFrom uint64, cmdData []byte) error {
-	d.pendingMu.Lock()
-	d.pendingResponses.Add(1)
-	d.pendingMu.Unlock()
+	d.shuttingDownLock.Lock()
+	if !d.shuttingDown {
+		d.pendingResponses.Add(1)
+	} else {
+		d.shuttingDownLock.Unlock()
+		return ErrShuttingDown
+	}
+	d.shuttingDownLock.Unlock()
 
-	go func() {
+	performWrite := func(handle uint64, from uint64, data []byte) error {
 		errchan := make(chan error)
 		go func() {
-			_, e := d.prov.WriteAt(cmdData, int64(cmdFrom))
+			_, e := d.prov.WriteAt(data, int64(from))
 			errchan <- e
 		}()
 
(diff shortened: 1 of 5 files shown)


ValentaTomas (commented):
Important changes:

### `dispatch.go`
- Reintroduce fatal channel for closing `Handle` on error
- Refactor pending lock/wg (refuse new cmds after close starts) 

### `path_direct_linux.go`
- Increase timeout for operations (4s -> 30s)
- Fix cancel call in `Close`
- Allow multiple socket connections
- Refactor device release on failure
- When retrying `Open` retry just connection establishment, without getting new device (not 100% sure this won't be a regression)

dobrac:
Ready for review. I've tested it using simulating traffic with pause/resume 3x for each sandbox for about a half an hour. I've not encountered any errors or undesired behavior.

> When retrying Open retry just connection establishment, without getting new device (not 100% sure this won't be a regression)

I'm not sure about this either, by default there is nbdnl.IndexAny so they are effectively changing the device index for each retry. If you are not sure either, I would remove it from this PR.

packages/orchestrator/internal/sandbox/nbd/path_direct_linux.go
```diff
@@ -156,16 +197,9 @@ func (d *DirectPathMount) Close(ctx context.Context) error {
...
-	telemetry.ReportEvent(childCtx, "closing socket pairs")
-	err = d.conn.Close()
-	if err != nil {
-		return err
-	}
-
 	// Wait until it's completely disconnected...
 	telemetry.ReportEvent(childCtx, "waiting for complete disconnection")
-	ctxTimeout, cancel := context.WithTimeout(childCtx, 4*time.Second)
+	ctxTimeout, cancel := context.WithTimeout(childCtx, timeout)
```
ValentaTomas: Let's not reuse the timeout that is also used for the NBD connection timeout.

Also I think ideally we would not use timeout here and the file would be 100% closed. If we can implement it reliably that way I think we should go for it.

dobrac: Changed to separate timeout. I agree it shouldn't be necessary, but I would leave it here for now, so we won't break it. Also with the increased time, we should see it in stats that the requests are timing out more clearly, as well as giving the disconnect more time.

ValentaTomas: Agreed

ValentaTomas:
Before we merge, can we do a simple benchmark to see how the number of connections affects the FS speed?
If it is significant, we will announce this.

dobrac:
> Before we merge, can we do a simple benchmark to see how the number of connections affects the FS speed? If it is significant, we will announce this.

Simple test using the command `curl -sL https://yabs.sh | bash -s -- -ig` doesn't show any significant difference in speed when changing the number of socket connections.

ValentaTomas:
Because of the layers (FC->kernel->NBD) I think this might affect only specific tests like *reading* enough data to profit from parallelized connections, where the data must not be cached on the system level already (so data previously not accessed by this FC). The improvements to writes might be mostly hidden as that will be flushing in the background.

ValentaTomas:
That is related, because it can be a bottleneck, but this could probably allow additional parallelism of the download too and the read should be affected even if the template is downloaded.

packages/orchestrator/internal/sandbox/nbd/dispatch.go
```diff
@@ -64,15 +69,17 @@ func NewDispatch(ctx context.Context, fp io.ReadWriteCloser, prov Provider) *Dis
...
 		prov:           prov,
 		ctx:            ctx,
+		fatal:          make(chan error, 1),
 	}
 
 	binary.BigEndian.PutUint32(d.responseHeader, NBDResponseMagic)
 	return d
 }
 
 func (d *Dispatch) Wait() {
```
jakubno: I would rename the method to Close/Shutdown

packages/orchestrator/internal/sandbox/nbd/path_direct_linux.go
```diff
@@ -41,74 +58,85 @@ func NewDirectPathMount(tracer trace.Tracer, b block.Device, devicePool *DeviceP
...
-		if err != nil {
-			return 0, err
-		}
-
-		// Create the socket pairs
-		sockPair, err := syscall.Socketpair(syscall.AF_UNIX, syscall.SOCK_STREAM, 0)
-		if err != nil {
-			return 0, err
-		}
+	// TODO: Do we need retry getting the device or it is enough to retry the dispatchers?
```
jakubno: What should be done?

packages/orchestrator/internal/sandbox/nbd/path_direct_linux.go
```diff
@@ -41,74 +58,85 @@ func NewDirectPathMount(tracer trace.Tracer, b block.Device, devicePool *DeviceP
...
+	d.deviceIndex, err = d.devicePool.GetDevice(ctx)
+	if err != nil {
+		return 0, err
+	}
 
-		client := os.NewFile(uintptr(sockPair[0]), "client")
-		server := os.NewFile(uintptr(sockPair[1]), "server")
-		d.conn, err = net.FileConn(server)
+	for {
```
jakubno, line 83: if you retry, you are not closing `d.dispatchers` and not cleaning and closing `d.socks`

packages/orchestrator/internal/sandbox/nbd/path_direct_linux.go
```diff
@@ -41,74 +58,85 @@ func NewDirectPathMount(tracer trace.Tracer, b block.Device, devicePool *DeviceP
...
-		_ = client.Close()
-
-		connErr := d.conn.Close()
-		if connErr != nil {
-			zap.L().Error("error closing conn", zap.Error(connErr))
+		for _, sock := range socks {
+			sock.Close()
 		}
 
-		releaseErr := d.devicePool.ReleaseDevice(d.deviceIndex)
```
jakubno, line 105: there's no need to release because we didn't connect, do i understand it correctly right?

packages/orchestrator/internal/sandbox/nbd/dispatch.go
dobrac: renamed to Drain

packages/orchestrator/internal/sandbox/nbd/path_direct_linux.go
dobrac: socks should be closed, there is no drain await on dispatchers, but it's probably not necessary as it doesn't release any resources 
```
for _, sock := range socks {
	sock.Close()
}
```

packages/orchestrator/internal/sandbox/nbd/path_direct_linux.go
dobrac: That goes to the TODO `// TODO: Do we need retry getting the device or it is enough to retry the dispatchers?`, so there is now an implementation, where the device is acquired at the beginning (Open) and released only on Close

packages/orchestrator/internal/sandbox/nbd/path_direct_linux.go
dobrac: @ValentaTomas asked here
> I'm not sure. @jakubno Do you remember if the device is unusable after fail or if just the connection was a problem?

I have no idea if it could cause any issue atm

packages/orchestrator/internal/sandbox/nbd/path_direct_linux.go
jakubno: fair for dispatchers
for socket [this](https://github.com/e2b-dev/infra/pull/517/files/0820688b5c8b4b0d17f10246e9e44e47ace61c8d#diff-f8e9ba3bee1cac243166476cd3fad1b2d83c48a0ebe6699d5193c98839a8fb1dR90) isn't closed 

```go
serverc, err := net.FileConn(server)
```

dobrac: added closing server and client sockets and retry and on nbd close

packages/orchestrator/internal/sandbox/nbd/path_direct_linux.go
```diff
@@ -143,10 +180,23 @@ func (d *DirectPathMount) Close(ctx context.Context) error {
 	telemetry.ReportEvent(childCtx, "canceling context")
 	d.cancelfn()
 
+	// Close all server socket pairs...
+	telemetry.ReportEvent(childCtx, "closing socket pairs server")
+	for _, v := range d.socksServer {
+		err := v.Close()
+		if err != nil {
+			return err
```
ValentaTomas: Should we return error here or report error and continue so we can finish the device release?
The error return was the previous behavior, so it should be ok to do this now, I just want to mention this as it might be problem in some cases.

packages/orchestrator/internal/sandbox/nbd/path_direct_linux.go
```diff
@@ -178,9 +221,40 @@ func (d *DirectPathMount) Close(ctx context.Context) error {
...
 			break
 		}
-
 		time.Sleep(100 * time.Nanosecond)
 	}
 
+	// Close all client socket pairs...
+	telemetry.ReportEvent(childCtx, "closing socket pairs client")
+	for _, v := range d.socksClient {
+		err := v.Close()
```
ValentaTomas: I don't think we were previously closing the client side—could there be a chance of this returning error if the client side was closed automatically because the server side was closed?

packages/orchestrator/internal/sandbox/nbd/path_direct_linux.go
ValentaTomas: @jakubno In the previous implementation when the NBD connection was not working we retried by trying to use a new NBD and new connection. 
Was using a new NBD necessary or could just a new connection work? I don't remember how many experiments we did with this.

ValentaTomas:
Let's resolve the questions and merge—we won't be doing more detailed benchmarking now.

packages/orchestrator/internal/sandbox/nbd/path_direct_linux.go
dobrac: There wasn't, I've added it as the client sockets were closed only on retry and I haven't found them to be automatically closed. Based on testing, it doesn't throw any errors. But, we can just log the errors and don't return the function

packages/orchestrator/internal/sandbox/nbd/path_direct_linux.go
dobrac: I'm not sure that if one of the Close errors out, if this will correctly finish and won't deadlock the Stop:
```
for _, d := range d.dispatchers {
	d.Drain()
}
```

packages/orchestrator/internal/sandbox/nbd/path_direct_linux.go
dobrac: Added "soft" error log here and removed the return statement

packages/orchestrator/internal/sandbox/nbd/path_direct_linux.go
dobrac: changed to soft errors with additional checks

dobrac:
Done few additional changes:
- Added soft errors on close instead of hard return
- return errors from the cow close
- get new device for each try on NBD open for now (there is added logging, so we can monitor if it even happens)
- return math.MaxUint32 for errors in nbd to not confuse the deviceIndex 0 with error state (as we might still call Close on errored NBD connection)

packages/orchestrator/internal/sandbox/nbd/path_direct_linux.go
jakubno: I don't remember anymore, I would return the logic that was there. I would add some logging we can monitor it and if there's no problem then we can move it to only socket pairs or delete it completely

ValentaTomas:
Let's update and merge this then (@jakubno review? :D)

packages/orchestrator/internal/sandbox/nbd/path_direct_linux.go
```diff
@@ -19,153 +24,237 @@ import (
...
-		releaseErr := d.devicePool.ReleaseDevice(d.deviceIndex)
-		if releaseErr != nil {
-			zap.L().Error("error releasing device", zap.Error(releaseErr))
+		for _, sock := range d.socksServer {
+			sock.Close()
 		}
-
-		d.deviceIndex = 0
+		// Release the device back to the pool
+		d.devicePool.ReleaseDevice(idx)
```
jakubno: shoudln't we at least log the error

packages/orchestrator/internal/sandbox/nbd/path_direct_linux.go
```diff
@@ -19,153 +24,237 @@ import (
...
+		blockSize:   4096,
+		devicePool:  devicePool,
+		socksClient: make([]*os.File, 0),
+		socksServer: make([]io.Closer, 0),
+		deviceIndex: math.MaxUint32,
 	}
 }
 
-func (d *DirectPathMount) Open(ctx context.Context) (uint32, error) {
+func (d *DirectPathMount) Open(ctx context.Context) (retDeviceIndex uint32, err error) {
```
jakubno, line 69: why not just like this?
```suggestion
func (d *DirectPathMount) Open(ctx context.Context) (deviceIndex uint32, err error) {
```

jakubno (approved):

packages/orchestrator/internal/sandbox/nbd/path_direct_linux.go
dobrac: it would be in conflict with local variable name deviceIndex

jakubno (approved):

PR #519, opened 2025-03-31
Fix sandbox pause slowdown

dobrac:


packages/orchestrator/internal/sandbox/nbd/path_direct_other.go
```diff
@@ -7,21 +7,23 @@ import (
...
+
 	"github.com/e2b-dev/infra/packages/orchestrator/internal/sandbox/block"
 )
 
 type DirectPathMount struct {
 	Backend block.Device
 }
 
-func NewDirectPathMount(b block.Device, devicePool *DevicePool) *DirectPathMount {
+func NewDirectPathMount(tracer trace.Tracer, b block.Device, devicePool *DevicePool) *DirectPathMount {
```
dobrac, line 19: this is only "path_direct_other" which is not used anywhere

jakubno (approved):
lgtm

PR #524, opened 2025-04-01
Add optional depth parameter to envd's ListDir 

0div:
# Description

This `envd` change allows us to add an optional `depth` parameter to `files.list` method on the SDK. 

The DX on the SDK side would look like this:
```js
const files = await sandbox.files.list(dirName, { depth: 3 })
```

- [x] update pb spec
- [x] generate pb go files
- [x] update ListDir method 

# Test 
- [x] infra integration test
- [x] [SDK](https://github.com/e2b-dev/E2B/pull/649) tests


mishushakov (commented):
I think this looks good to me, but keep in mind that "filepath.WalkDir reads an entire directory into memory before proceeding to walk that directory"

0div:
> I think this looks good to me, but keep in mind that "filepath.WalkDir reads an entire directory into memory before proceeding to walk that directory"

oh wow really ? where did you see this?

jakubno:
>I think this looks good to me, but keep in mind that "filepath.WalkDir reads an entire directory into memory before proceeding to walk that directory"

it reads stores names of the files / directories, which should be fine

tests/integration/internal/tests/envd/filesystem_test.go
```diff
@@ -123,3 +134,20 @@ func createTextFile(tb testing.TB, path string, content string) (*bytes.Buffer,
...
+	client := setup.GetEnvdClient(tb, ctx)
+	req := connect.NewRequest(&filesystem.MakeDirRequest{
+		Path: path,
+	})
+	setup.SetUserHeader(req.Header(), "user")
+	_, err := client.FilesystemClient.MakeDir(ctx, req)
+	if err != nil {
+		tb.Fatal(err)
+	}
+}
```
mishushakov, line 193: maybe move this to the top?

tests/integration/internal/tests/envd/filesystem_test.go
```diff
@@ -25,16 +25,27 @@ func TestListDir(t *testing.T) {
...
+	createTextFile(t, "/test-dir/sub-dir/file.txt", "Hello, World!")
+
 	envdClient := setup.GetEnvdClient(t, ctx)
+
+	// List all files in the root directory with depth 1
 	req := connect.NewRequest(&filesystem.ListDirRequest{
-		Path: "/",
+		Path:  "/",
+		Depth: 1,
 	})
```
mishushakov: can you also add tests for different depth and when no depth parameter is supplied (if that's even possible)?

tests/integration/README.md
```diff
@@ -1,16 +1,22 @@
 # Integration Tests
+
 Package for defining integration tests. Currently, there is a setup for API and Orchestrator testing.
 
 ## Run locally
-1) Setup env variables in the .env file
-2) If necessary, run `make connect-orchestrator` to create a tunnel to one orchestrator client VM in GCP
-3) Run `make test`
+
+1. Setup env variables in the `infra/.env` file
```
mishushakov: .env file (infra does not exist in current path)

tests/integration/README.md
```diff
@@ -1,16 +1,22 @@
...
+2. If you made changes to the `api` or `envd` protobuf spec, run `make generate` from this folder (and don't forget to generate it in `envd` if changes apply there too).
```
mishushakov, line 8: If you made changes to the `api` or `envd` protobuf spec, run `make generate` in their respective directories (`packages/api`, `packages/envd`).

packages/envd/spec/filesystem/filesystem.proto
```diff
@@ -62,6 +62,7 @@ enum FileType {
 
 message ListDirRequest {
     string path = 1;
+    uint32 depth = 2;
```
mishushakov, line 65: is it possible to set a default parameter here?

mishushakov (changes requested):
Looks good, minor nits

packages/envd/spec/filesystem/filesystem.proto
0div: in `proto3` custom default values are disallowed

tests/integration/internal/tests/envd/filesystem_test.go
0div: i did in SDKs, can also add to integration yes

tests/integration/internal/tests/envd/filesystem_test.go
0div: im just used to grouping helper un-exported functions at the bottom of files since order of def doesn't matter in Go but i don't have a strong opinion about them being at the bottom or the top, why would you prefer this?

mishushakov: I don't have a concrete opinion, but will ask in engineering

mishushakov: Tomas said prefer at the bottom

tests/integration/internal/tests/envd/filesystem_test.go
```diff
@@ -25,16 +25,67 @@ func TestListDir(t *testing.T) {
...
+			name:  "depth 2 lists all directories and files",
+			depth: 2,
+			expectedPaths: []string{
+				"test-dir",
+				"test-dir/sub-dir-1",
+				"test-dir/sub-dir-2",
+				"test-dir/sub-dir/file.txt",
+			},
+		},
+	}
```
mishushakov, line 66: my bad, make it explicit, it's now more difficult to digest

mishushakov (changes requested):
Make tests explicit and the rest looks good

PR #525, opened 2025-04-01
Fix connection issue for packer build

jakubno:
# Description

As we don't allow SSH connection by default, this change uses IAP to use google internal network

### DIFF
--- a/packages/cluster-disk-image/main.pkr.hcl
+++ b/packages/cluster-disk-image/main.pkr.hcl
@@ -24,6 +24,11 @@ source "googlecompute" "orch" {
 
   # Enable nested virtualization
   image_licenses = ["projects/vm-options/global/licenses/enable-vmx"]
+
+  # Enable IAP for SSH
+  network    = var.network_name
+  subnetwork = "${var.network_name}-subnetwork"
+  use_iap    = true
 }
 
 build {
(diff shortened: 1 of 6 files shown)


packages/cluster-disk-image/main.pkr.hcl
```diff
@@ -24,6 +24,10 @@ source "googlecompute" "orch" {
 
   # Enable nested virtualization
   image_licenses = ["projects/vm-options/global/licenses/enable-vmx"]
+
+  # Enable IAP for SSH
+  use_iap = true
+  tags = ["packer-build"]
```
dobrac: is it not possible to use the `packer_build_tag_name` here?

dobrac (approved):
Is it possible to merge the two variables to be defined only in one place instead of two of them?

packages/cluster/network/variables.tf
```diff
@@ -19,6 +19,11 @@ variable "cluster_tag_name" {
   type = string
 }
 
+variable "packer_build_tag_name" {
```
ValentaTomas: Why do we need this? It is because the default network we use would be also used for the packer build?

Should we create a separate network for this?

jakubno: Updated

packages/cluster-disk-image/main.tf
```diff
@@ -0,0 +1,45 @@
...
+}
+
+
+resource "google_compute_firewall" "internal_remote_connection_firewall_ingress" {
+  name    = "${var.network_name}-firewall-ingress"
+  network = google_compute_network.packer_network.name
+
+  allow {
+    protocol = "tcp"
+    ports    = ["22", "3389"]
```
dobrac: do we need port 3389?

PR #527, opened 2025-04-01
Envd with http auth support

sitole:
Envd can now receive optional attribute access token to authorize all traffic coming to envd itself.

When auth is enabled envd is expecting `X-Access-Token` for all incoming requests.
There is exception for requests coming to `/health` and `/files`. Files upload and download is little bit special and can be authorized not just with header but also with query parameter `signing` that can be generated locally just from request parameters and access token. This is special case when you need to share URL for file upload/download with some env where it's not easy to provide access token as header or you don't want to share access token with 3rd party.

If authorization is not enabled envd will behave exactly same us until this change.

Signing key generation is following.
We are providing prefix with version so there is space for future changed in algorithm used for signing. 
```go
hasher := keys.NewSHA256Hashing()
signatureTmp := fmt.Sprintf("%s:%s:%s:%s:%s", path, operation, username, apiToken, strconv.FormatInt(timestamp, 10))
signature := fmt.Sprintf("v1_%s", hasher.HashWithoutPrefix([]byte(localSignatureTmp)))
```


packages/envd/internal/api/store.go
```diff
@@ -2,23 +2,134 @@ package api
...
+func (a *API) WithAuthorization(handler http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		if a.accessToken != nil {
+			authHeader := req.Header.Get(accessTokenHeader)
+
+			// check if this path is allowed without authentication (e.g., health check, endpoints supporting signing)
+			allowedPath := slices.Contains(allowedPaths, req.Method+req.URL.Path)
+
+			if authHeader != *a.accessToken && !allowedPath {
+				response := jsonErrorResponse{Error: "Unauthorized access, please provide a valid access token or method signing if supported"}
```
jakubno: Can't we use the same error type and in `packages/envd/internal/api/error.go`?

packages/envd/internal/api/store.go
```diff
@@ -2,23 +2,134 @@ package api
...
+
+type jsonErrorResponse struct {
+	Error string `json:"error"`
 }
 
 func New(l *zerolog.Logger, envVars *utils.Map[string, string]) *API {
 	return &API{logger: l, envVars: envVars}
 }
 
+func (a *API) WithAuthorization(handler http.Handler) http.Handler {
```
jakubno: I would probably move all these to separate file to keep it clean

packages/envd/internal/api/store.go
```diff
@@ -2,23 +2,134 @@ package api
...
+			// check if this path is allowed without authentication (e.g., health check, endpoints supporting signing)
+			allowedPath := slices.Contains(allowedPaths, req.Method+req.URL.Path)
+
+			if authHeader != *a.accessToken && !allowedPath {
+				response := jsonErrorResponse{Error: "Unauthorized access, please provide a valid access token or method signing if supported"}
+				responseBytes, _ := json.Marshal(response)
+
+				a.logger.Error().Msg("Trying to access secured envd without correct access token")
+
+				w.WriteHeader(http.StatusUnauthorized)
```
jakubno: How does this look like in SDK?

packages/envd/internal/api/store.go
```diff
@@ -2,23 +2,134 @@ package api
...
+	}
+
+	// check if access token is sent in the header
+	tokenFromHeader := r.Header.Get(accessTokenHeader)
+	if tokenFromHeader != "" && tokenFromHeader != *a.accessToken {
+		errMsg = fmt.Errorf("access token present in header but does not match")
+		errorCode = http.StatusUnauthorized
+
+		a.logger.Err(errMsg)
+		jsonError(w, errorCode, errMsg)
```
jakubno: not sure if such side effect should be expected here, I would probably change the func return type only `*Error` and handle the rest in the actual handler

jakubno (commented):
Do we want to allow to set / change access token during runtime / after it was set? Not sure if it's a problem, but you can "lock" sandbox from the owner if the access token wasn't set. E.g somebody shared link to some server running inside the sandbox.

Also should the signed urls work forever?

packages/envd/internal/api/store.go
sitole: For this feature to be used user will need to have new version of SDK with parameter supporting that so we can handle it properly i suppose?

sitole:
> Do we want to allow to set / change access token during runtime / after it was set? Not sure if it's a problem, but you can "lock" sandbox from the owner if the access token wasn't set. E.g somebody shared link to some server running inside the sandbox.

Yes, thats true. Let's add attribute declaring that envd was already initialized and auth token cannot be changed then.

> Also should the signed urls work forever?

Sure. We can add expiration with timestamp. This will require to change format of the signing to be not hashed but encrypted by access token. But sure, will help us in future to support multiple attributes here.

packages/envd/internal/api/store.go
sitole: done

packages/envd/internal/api/store.go
sitole: done

packages/envd/internal/api/store.go
sitole: done

sitole:
Added option for scoping file download/upload with expiration that is send in URL and also its part of signature.

sitole:
> > Do we want to allow to set / change access token during runtime / after it was set? Not sure if it's a problem, but you can "lock" sandbox from the owner if the access token wasn't set. E.g somebody shared link to some server running inside the sandbox.
> 
> Yes, thats true. Let's add attribute declaring that envd was already initialized and auth token cannot be changed then.

We cannot do it easily now because first init will be automatically done by orchestrator. So lets wait until access token will be propagated to api and orchestrator, then we can do it.

packages/envd/internal/api/auth.go
```diff
@@ -0,0 +1,111 @@
...
+	"time"
+)
+
+const (
+	SigningReadOperation  = "read"
+	SigningWriteOperation = "write"
+)
+
+var (
+	accessTokenHeader = "X-Access-Token"
```
jakubno: this can be const

packages/envd/internal/api/init.go
```diff
@@ -35,6 +35,11 @@ func (a *API) PostInit(w http.ResponseWriter, r *http.Request) {
 				a.envVars.Store(key, value)
 			}
 		}
+
+		if initRequest.AccessToken != nil {
+			a.logger.Debug().Str(string(logs.OperationIDKey), operationID).Msg("Setting access token")
+			a.accessToken = initRequest.AccessToken
+		}
```
jakubno, line 42: what if the access token is already set? should you be able to change it?

jakubno: This will be adjusted when we call this from orchestrator

packages/envd/internal/api/auth.go
sitole: done, thx

PR #528, opened 2025-04-01
Fix unit tests on other platforms

dobrac:
fixes unit tests for orchestrator on other platforms


mishushakov (commented):
Can you write a small summary what the change is about? What other platforms?

dobrac:
> Can you write a small summary what the change is about? What other platforms?

It's just a fix for feature parity. 

Some files are using linux specific libraries, so in order for you to be able to compile them on !linux platforms (hence other), we've defined alternative (usually just mocked for compilation) implementation. These other files are not used anywhere in production.

PR #530, opened 2025-04-02
Allow orchestrator deployment updates without restarting running deployments

ValentaTomas:
This PR outlines how the orchestrator job updates could work, if we need to "opportunistically" deploy orchestrator job/envs changes.

It moves the changing parts of the job to a separate script while providing this script via Nomad variable—as opposed to a normal variable. We are able to control the lifecycle of Nomad variable if we use it in a template stanza and set the behavior for changes to "noop" if the change should not restart tasks.
Later, we can try the signal/script change behavior to trigger the orchestrator reload.

This change moves the logic (artifact download, setting envs, starting orchestrator) to a separate script, but I did not find any other way to achieve the functionality we wanted.

Before this is ready, we need to at least:
- [ ] Make the orchestrator (previously in artifact stanza) downloading/version check more robust
- [ ] Test in all situations
- [ ] Have a clear path on how we will move from the current system to this one
- [ ] Check if we should pass a port variable via the job's env stanza, to prevent a situation where a port change in the job changes just services (not reloading allocation), but the ports in the allocations are not changed. It is possible that the service/port var change would trigger a whole job reload nevertheless.

### DIFF
--- a/packages/nomad/main.tf
+++ b/packages/nomad/main.tf
@@ -306,7 +306,6 @@ data "google_storage_bucket_object" "orchestrator" {
   bucket = var.fc_env_pipeline_bucket_name
 }
 
-
 data "google_compute_machine_types" "client" {
   zone   = var.gcp_zone
   filter = "name = \"${var.client_machine_type}\""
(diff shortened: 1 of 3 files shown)


jakubno:
You still can't change e.g. datacenter, ports and other job definitions

ValentaTomas:
For datacenter, we might do:

```
datacenters (array<string>: ["*"]) - A list of datacenters in the region which are eligible for task placement. This field allows wildcard globbing through the use of * for multi-character matching. The default value is ["*"], which allows the job to be placed in any available datacenter.
```

[https://developer.hashicorp.com/nomad/docs/v1.6.x/job-specification/job#datacenters](https://developer.hashicorp.com/nomad/docs/v1.6.x/job-specification/job#datacenters)<br><br>For the ports—it might be possible to use nomad vars with templating here too to dynamically add the service/ports stanzas, but ***really*** not sure about rerender/reloads then. Also as it is a raw exec system job we technically don't need port/service stanzas to have access to the endpoints (won't have nomad service healthchecks though)<br><br>For other job definition—I deliberately moved as much possible to the to the scripts/outside of the job.<br>Might be worth considering when it is still the same job then, if we change everything about it, but kinda hitting limits of the templates with this draft. :D<br><br>For the solution with essentially creating new jobs, I might have a few ideas too. It might be easier in the end in some ways, just trying to find the right angle.

ValentaTomas:
Also, we might want to use the `restart` stanza regardless and the nomadVars for future triggering of the orchestrator updates (`SIGHUP`?) even if we scrap this, as this provides z way to do that.

ValentaTomas:
I'm leaving this open for reference until we solve the orchestrator deployments, but I don't think we will be using this solution.

PR #531, opened 2025-04-02
Fix potential nil dereference

jakubno:
# Description

If `Limit` was `nil` and you would try to list Paused sandboxes (not present in the code right now), it would result in nil dereference. 

I think it make sense to use default limit here, because listing all paused sandboxes could be super expensive

### DIFF
--- a/packages/api/internal/handlers/sandboxes_list.go
+++ b/packages/api/internal/handlers/sandboxes_list.go
@@ -24,6 +24,8 @@ import (
 	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
 )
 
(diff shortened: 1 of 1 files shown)


ValentaTomas:
There also should be a hard max limit.

mishushakov:
can you update it for running as well? nil dereference here would in practice not occur, because we have a default limit in the spec when not specified

mishushakov:
@ValentaTomas the default amount/max limit is 1000 in the spec, but it's not explicitly in the code

ValentaTomas:
Also, be please more specific with the PR naming (nil dereference where, during what).

PR #535, opened 2025-04-03
Re-added previous sandboxes list logic (v1)

mishushakov:
- Moved current/new Sandbox.list logic under /v2
- Added tests for previous (v1) version


ValentaTomas:
@mishushakov Please assign and add labels.

jakubno (approved):

PR #536, opened 2025-04-03
Add possibility to switch to managed Redis 

jakubno:
# Description

- Add possibility to deploy managed redis by setting env variable `MANAGER_REDIS=true`
- If you don't add managed redis, you will still use the locally deployed one

## Deployment strategy
- This PR only adds managed redis and in API server it will send data to the managed instance. This will allow us to migrate without any problem
- There will be second PR which will remove the second redis after at least 24 hours, so we are sure all data are in the managed redis


packages/api/internal/dns/server.go
```diff
@@ -38,11 +39,15 @@ type DNS struct {
 	}
 }
 
-func New(ctx context.Context, rc *redis.Client) *DNS {
+func New(ctx context.Context, rc *redis.Client, rc2 *redis.Client) *DNS {
```
mishushakov, line 42: why are there two Redis clients at the same time now? I am confused

jakubno: As described in description, this way we can populate data without causing downtime, if the redis restarts (or if we replace it with a new one, it would lost all routing information to sandboxes, all sandboxes started prior to the change would be inaccessible).

After 24 hours we will be sure that all data are in redis and we can remove this and use only the new redis without any issue

ValentaTomas (approved):
We could have used LaunchDarkly and write it in a way that allows for both redis conditions to be evaluated independendly.

PR #538, opened 2025-04-03
Feature flag for client proxy target traffic switching

sitole:
Implemented Launchdarkly SDK for feature flags.
This allows us to remotely control rolling features and possibly reverting it safely.

Implementation is that in way that when ENV with API key is not defined, 
SDK is switched to offline mode and feature flags returns default values that are defined in function calls.

All feature flags are stored locally and there are hard default so in case there is some issue with networking, accidentally deleted feature flag in remote platform there will be always safe fallback present.

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -236,7 +236,3 @@ variable "redis_managed" {
   default = false
   type    = bool
 }
-
-variable "launch_darkly_api_key" {
-  type = string
-}


packages/shared/pkg/feature-flags/client.go
```diff
@@ -0,0 +1,56 @@
...
+	ldClient, err = ldclient.MakeClient(launchDarklyApiKey, waitForInitialize)
+	if err != nil {
+		return nil, err
+	}
+
+	return &Client{Ld: ldClient}, nil
+}
+
+func (c *Client) Close() error {
+	if c.Ld != nil {
```
dobrac: nit for code improvement (imo):

```
func (c *Client) Close() error {
  if c.Ld == nil {
      return nil
  }
   
  err := c.Ld.Close()
  if err != nil {
      zap.L().Error("Error during launch-darkly client shutdown", zap.Error(err))
      return err
  }
  return nil
}
```

packages/shared/pkg/feature-flags/client.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+var launchDarklyApiKey = os.Getenv("LAUNCH_DARKLY_API_KEY")
+
+const (
+	// User by client proxy to route traffic between nginx sandbox proxy and orchestrator proxy
+	// https://app.launchdarkly.com/projects/default/flags/sandbox-proxy-traffic-target
+	ClientProxyTrafficTargetFeatureFlag = "sandbox-proxy-traffic-target"
+
+	ClientProxyTrafficToNginx        = "nginx-proxy"
```
dobrac: can this be somehow scoped to the specific feature flag (ClientProxyTrafficTargetFeatureFlag)?

packages/client-proxy/main.go
```diff
@@ -152,9 +156,25 @@ func proxyHandler(transport *http.Transport) func(w http.ResponseWriter, r *http
 
 		// We've resolved the node to proxy the request to
 		zap.L().Debug("Proxying request", zap.String("sandbox_id", sandboxID), zap.String("node", node))
+
+		flagCtx := ldcontext.NewBuilder("client-proxy-context").SetString("sandbox_id", sandboxID).Build()
```
dobrac: I suppose this `client-proxy-context` might be const

packages/client-proxy/main.go
```diff
@@ -152,9 +156,25 @@ func proxyHandler(transport *http.Transport) func(w http.ResponseWriter, r *http
...
+		flag, flagErr := featureFlags.Ld.StringVariation(featureflags.ClientProxyTrafficTargetFeatureFlag, flagCtx, featureflags.ClientProxyTrafficToNginx)
```
dobrac: Maybe also worth setting `featureflags.ClientProxyTrafficToNginx` as a const in file saying like defaultFallback

dobrac (changes requested):
lgmt, a question about testing, can we have something here so the client proxy could be used in integration tests?

variables.tf
```diff
@@ -231,3 +231,7 @@ variable "template_bucket_name" {
   type        = string
   description = "The name of the FC template bucket"
 }
+
+variable "launch_darkly_api_key" {
+  type = string
```
jakubno: not sure if this shouldn't be part of secrets as we do it for all other services like this
```suggestion
  type     = string
  default = ""
```

packages/shared/pkg/feature-flags/client.go
sitole: exported to sub package to gain different import prefix to make it cleaner when there will be more of them.

sitole:
> Question about testing, can we have something here so the client proxy could be used in integration tests?

Added accessible global store that can be used for mocking. There is also example test.
Will work in integration tests, but i'am not sure how we should write tooling for integration tests. 
Launchdarkly officially says we should use dedicated environment for it.

packages/shared/pkg/feature-flags/client.go
```diff
@@ -18,7 +22,7 @@ func NewClient(waitForInitialize time.Duration) (*Client, error) {
 	var err error
 
 	if launchDarklyApiKey == "" {
-		ldClient, err = ldclient.MakeCustomClient("", ldclient.Config{Offline: true}, waitForInitialize)
```
sitole: Offline mode is not needed with testing data source

dobrac (approved):

variables.tf
sitole: done

packages/nomad/client-proxy.hcl
```diff
@@ -66,6 +66,10 @@ job "client-proxy" {
       env {
         OTEL_COLLECTOR_GRPC_ENDPOINT  = "${otel_collector_grpc_endpoint}"
         LOGS_COLLECTOR_ADDRESS        = "${logs_collector_address}"
+
+%{ if launch_darkly_api_key != " " && launch_darkly_api_key != "" }
```
jakubno:
```suggestion
%{ if launch_darkly_api_key != "" }
```

packages/nomad/main.tf
```diff
@@ -104,6 +108,7 @@ resource "nomad_job" "client_proxy" {
 
       otel_collector_grpc_endpoint = "localhost:4317"
       logs_collector_address       = "http://localhost:${var.logs_proxy_port.port}"
+      launch_darkly_api_key        = data.google_secret_manager_secret_version.launch_darkly_api_key.secret_data
```
jakubno:
```suggestion
      launch_darkly_api_key        = trimspace(data.google_secret_manager_secret_version.launch_darkly_api_key.secret_data)
```

packages/shared/pkg/feature-flags/client.go
```diff
@@ -0,0 +1,53 @@
...
+	ldClient, err = ldclient.MakeClient(launchDarklyApiKey, waitForInitialize)
+	if err != nil {
+		return nil, err
+	}
+
+	return &Client{Ld: ldClient}, nil
+}
+
+func (c *Client) Close() error {
+	if c.Ld != nil {
```
jakubno:
```suggestion
	if c.Ld == nil {
```

packages/shared/pkg/feature-flags/client_test.go
```diff
@@ -0,0 +1,31 @@
...
+	flagValue, _ := client.Ld.BoolVariation(flagName, clientCtx, false)
+	assert.False(t, flagValue)
+
+	LaunchDarklyOfflineStore.Update(
+		LaunchDarklyOfflineStore.Flag(flagName).VariationForAll(true),
+	)
+
+	// value is set manually in datastore and should be taken from there
+	flagValue, _ = client.Ld.BoolVariation(flagName, clientCtx, false)
+	assert.Equal(t, true, flagValue)
```
jakubno:
```suggestion
	assert.True(t, flagValue)
```

jakubno (approved):

PR #539, opened 2025-04-03
Sandbox list PR Refactor

jakubno:
# Description

Refactor to simplify the logic
- split the function so the required parameter are really required
-  return better errors to user

### DIFF
--- a/packages/api/internal/handlers/sandboxes_list.go
+++ b/packages/api/internal/handlers/sandboxes_list.go
@@ -20,6 +20,7 @@ import (
 	"github.com/e2b-dev/infra/packages/api/internal/auth"
 	authcache "github.com/e2b-dev/infra/packages/api/internal/cache/auth"
 	"github.com/e2b-dev/infra/packages/api/internal/cache/instance"
+	"github.com/e2b-dev/infra/packages/api/internal/orchestrator"
 	"github.com/e2b-dev/infra/packages/api/internal/utils"
 	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
 )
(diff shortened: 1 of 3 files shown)


ValentaTomas:
Be more decriptive in the PR name

jakubno:
It isn't towards main, only to Mish's PR instead of explaining this in comments, but you are right, I will update the name and description

packages/api/internal/handlers/sandboxes_list.go
```diff
@@ -64,23 +73,11 @@ func parseCursor(cursor string) (time.Time, string, error) {
 	return cursorTime, parts[1], nil
 }
 
-func (a *APIStore) getRunningSandboxes(runningSandboxes []*instance.InstanceInfo, metadataFilter *map[string]string, cursorTime time.Time, cursorID string, limit *int32) ([]PaginatedSandbox, error) {
+func transformToSandboxAPI(runningSandboxes []*instance.InstanceInfo) []PaginatedSandbox {
```
mishushakov: I got confused by this, why is this necessary?

packages/api/internal/handlers/sandboxes_list_metrics.go
```diff
@@ -191,21 +191,17 @@ func (a *APIStore) GetSandboxesMetrics(c *gin.Context, params api.GetSandboxesMe
...
-		zap.L().Error("Error fetching sandboxes", zap.Error(err))
-		a.sendAPIStoreError(c, http.StatusInternalServerError, fmt.Sprintf("Error returning sandboxes for team '%s': %s", team.ID, err))
+		zap.L().Error("Error parsing metadata", zap.Error(err))
+		a.sendAPIStoreError(c, http.StatusBadRequest, fmt.Sprintf("Error parsing metadata: %s", err))
 
 		return
 	}
 
+	// Get relevant running sandboxes
+	sandboxes := a.getRunningSandboxes(ctx, team.ID, metadataFilter)
```
mishushakov: I was bit confused by this, in the code there's now `getSandboxes`, `getRunningSandboxes` and `a.orchestrator.GetSandboxes`

packages/api/internal/handlers/sandboxes_list.go
jakubno: It isn't strictly necessary, but I wanted to make the function smaller, so they don't do everything. This way we can reuse some of the functions for multiple endpoints without any weird logic for making some parameters optional, but then actually requiring them later

jakubno: Feel free to propose better name for this function though, I struggled with that :D

packages/api/internal/handlers/sandboxes_list_metrics.go
jakubno: I didn't add none of these, they've been there already, we can get rid of getSandboxes though..

jakubno: Removed `getSandboxes`

PR #543, opened 2025-04-04
Create merge conflict for migrations out of order

dobrac:
## Description
When creating database migrations in PRs, they could be eventually out of order when merged, causing  potential skip of the migration altogether.

This PR also fixes one skipped migration adding sandbox_started_at field to sandboxes.

## Solution
This PR tries to solve the issue using a single file which should always contain the latest version in the repository. If two PRs are having migrations, the first PR merged will cause the other one a merge conflict on this file. It's not ideal solution, but it's better than silently failing. The best solution would be if the library itself would solve this. More description of the solution is here: https://github.com/golang-migrate/migrate/issues/510#issuecomment-2560182894

## Alternatives
Another option would be to change the migration tool from golang-migrate to e.g. goose (https://github.com/pressly/goose)


jakubno:
I think we should use `goose` instead

ValentaTomas:
@jakubno @dobrac Do you want to switch to `goose` do you think it is worth it? I don't have an opinion here.

jakubno:
Superseded by #589. Closing this PR.

PR #545, opened 2025-04-04
Redis cluster improvements

jakubno:
# Description

Redis Cluster has to use ClusterClient instead of the standard Client

After we migrate, we can probably abstract the RedisClient vs RedisClientCluster, so there will be only 1 redisCache independant on the client


ValentaTomas:
Briefly checked and I think it should be ok.

packages/api/internal/dns/server.go
```diff
@@ -9,14 +9,24 @@ import (
...
-	"github.com/go-redis/redis/v8"
+	"github.com/go-redis/cache/v9"
 	resolver "github.com/miekg/dns"
+	"github.com/redis/go-redis/v9"
 	"go.uber.org/zap"
 
 	"github.com/e2b-dev/infra/packages/shared/pkg/smap"
 )
 
+// Rediser generalize RedisClient and RedisClusterClient, taken from github.com/go-redis/cache/v9
```
jakubno, line 21: After migration period, we will be using only one of them, this is prep for smoother transition. There will be only one client afterwards

dobrac (approved):
lgtm, better would be use of a list, but I suppose it's quite temporary here, so it doesn't matter that much

jakubno:
Would be better, especially after the generalization, but as it is temporary, it would be reverted anyway

PR #553, opened 2025-04-09
Add Grafana dashboard to terraform

jakubno:
# Description

Makes dashboard sync easier

## How to update dashboard/panel

Update your Grafana, click three dots on the updated panel, copy JSON, update the corresponding JSON in `./terraform/grafana/panels`  and make sure to not overwrite templated fields

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ OTEL_TRACING_PRINT ?= false
 TEMPLATE_BUCKET_LOCATION ?= $(GCP_REGION)
 CLIENT_CLUSTER_AUTO_SCALING_MAX ?= 0
 REDIS_MANAGED ?= false
+GRAFANA_MANAGED ?= false
 
 tf_vars := TF_VAR_client_machine_type=$(CLIENT_MACHINE_TYPE) \
 	TF_VAR_client_cluster_size=$(CLIENT_CLUSTER_SIZE) \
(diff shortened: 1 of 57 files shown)


terraform/grafana/variables.tf
```diff
@@ -11,6 +11,16 @@ variable "gcp_region" {
   type = string
 }
 
+variable "domain_name" {
+  type = string
+}
+
+variable "directory_path" {
+  description = "Path to the directory containing files"
```
dobrac:
```suggestion
  description = "Path to the directory containing panel files"
```

terraform/grafana/main.tf
```diff
@@ -235,6 +235,23 @@ resource "grafana_data_source" "gcloud_monitoring" {
...
+      stackdriver_uid = grafana_data_source.gcloud_monitoring.uid
+      prefix          = var.prefix
+    })
+  }
 }
 
+resource "grafana_dashboard" "dashboard" {
+  provider = grafana.datasource
+
+  config_json = templatefile("${path.module}/dashboard.json", merge({
```
dobrac: can there be also a folder for dashboards?

dobrac (approved):

Makefile
```diff
@@ -77,6 +79,7 @@ update-prod-env:
 plan:
 	@ printf "Planning Terraform for env: `tput setaf 2``tput bold`$(ENV)`tput sgr0`\n\n"
 	$(TF) fmt -recursive
+	@ if [ "$(GRAFANA_MANAGED)" = "true" ]; then make -C terraform/grafana plan; fi
```
dobrac: shouldn't the Grafana be setup after the plan, because when you don't have any cluster, how do you set the data sources? (same applies to apply)

jakubno: Refactored

terraform/grafana/main.tf
jakubno: Added

terraform/grafana/variables.tf
jakubno: Fixed

dobrac:
Just a note, production configs (and staging envs with Grafana) will need to be updated with `GRAFANA_MANAGED`

dobrac (approved):

PR #554, opened 2025-04-09
Add template alias attribute to gin middleware metrics

dobrac:
### Description
Improves observability of startup times of specific templates.

#### Monitored templates
- base
- code-interpreter-v1
- code-interpreter-beta
- desktop

### DIFF
--- a/packages/api/internal/handlers/sandbox_create.go
+++ b/packages/api/internal/handlers/sandbox_create.go
@@ -14,6 +14,7 @@ import (
 	"github.com/e2b-dev/infra/packages/api/internal/auth"
 	authcache "github.com/e2b-dev/infra/packages/api/internal/cache/auth"
 	"github.com/e2b-dev/infra/packages/api/internal/cache/instance"
+	"github.com/e2b-dev/infra/packages/api/internal/middleware/otel/metrics"
 	"github.com/e2b-dev/infra/packages/api/internal/utils"
 	"github.com/e2b-dev/infra/packages/shared/pkg/id"
 	sbxlogger "github.com/e2b-dev/infra/packages/shared/pkg/logger/sandbox"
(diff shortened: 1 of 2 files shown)


mishushakov:
keep in mind there's also some stats about template spawn count in the db already
Screenshot 2025-04-09 at 12 47 51

dobrac:
> keep in mind there's also some stats about template spawn count in the db
> 
> Screenshot 2025-04-09 at 12 47 51

I know (see the linked Linear issue). Do you propose to get these automatically? I think that is unnecessary for now. Ideally we could track all of them, but Prometheus&Grafana pricing is against it.

mishushakov:
I propose finding out if it's used anywhere and removing that to reduce the load on the db

packages/api/internal/handlers/sandbox_create.go
```diff
@@ -14,14 +14,27 @@ import (
...
+)
+
+var (
+	// mostUsedTemplates is a map of the most used template aliases.
+	// It is used for monitoring and to reduce metric cardinality.
+	mostUsedTemplates = map[string]struct{}{
+		"base":                  {},
+		"code-interpreter-v1":   {},
+		"code-interpreter-beta": {},
+		"desktop-dev1":          {},
```
jakubno:
```suggestion
		"desktop":          {},
```

packages/api/internal/handlers/sandbox_create.go
```diff
@@ -152,3 +168,21 @@ func (a *APIStore) PostSandboxes(c *gin.Context) {
...
+
+	if _, ok := mostUsedTemplates[alias]; !ok {
+		// alias is not most used, set the name to "other"
+		alias = "other"
+	}
+
+	c.Set(metricTemplateAlias, alias)
+}
+
+func firstAlias(aliases []string) (string, bool) {
```
jakubno: is there reason to return `bool`, it isn't used?

Also there's already something similar on line 106, maybe you could merge it

jakubno (changes requested):

packages/api/internal/handlers/sandbox_create.go
dobrac: Nice catch!

dobrac:
Updated + improved the setTemplateNameMetric for template with multiple aliases

jakubno (approved):

PR #558, opened 2025-04-09
Added API support for sandbox envd auth

sitole:
Depends on https://github.com/e2b-dev/infra/pull/527


tests/integration/internal/tests/envd/auth_test.go
```diff
@@ -40,46 +43,31 @@ func TestAccessToAuthorizedPathWithoutToken(t *testing.T) {
...
-	// first init call
-	sandboxEnvdInitCall(t, ctx, envdInitCall{
-		sbx:                   sbx,
-		client:                envdClient,
-		body:                  envdapi.PostInitJSONRequestBody{AccessToken: &envdAuthToken},
-		expectedResErr:        nil,
-		expectedResHttpStatus: http.StatusNoContent,
-	})
 
 	// second call without authorization
```
jakubno: outdated comment

tests/integration/internal/tests/envd/auth_test.go
```diff
@@ -123,7 +104,7 @@ func TestAccessAuthorizedPathWithOutdatedAccessToken(t *testing.T) {
 	sandboxEnvdInitCall(t, ctx, envdInitCall{
```
jakubno, line 106: this shouldn't be possible anymore, right?

tests/integration/internal/tests/api/sandbox_test.go
```diff
@@ -40,3 +40,26 @@ func TestSandboxCreate(t *testing.T) {
 
 	assert.Equal(t, http.StatusCreated, resp.StatusCode())
 }
+
+func TestSandboxResumeUnknownSandbox(t *testing.T) {
```
jakubno, line 45: Can you also check for resumed sandbox, it has the non `nil`/ same `accessToken`?

spec/openapi.yml
```diff
@@ -281,6 +284,12 @@ components:
           type: string
           format: date-time
           description: Time when the sandbox will expire
+        envdVersion:
```
jakubno, line 287: do we want this when listing sandboxes?

packages/api/internal/sandbox/sandbox_envd_secret.go
```diff
@@ -0,0 +1,31 @@
...
+	return &EnvdAccessTokenGenerator{}
+}
+
+func (g *EnvdAccessTokenGenerator) GenerateAccessToken(id api.SandboxID) (string, error) {
+	if saltKey == "" {
+		return "", errors.New("salt key is not set")
+	}
+
+	raw := []byte(fmt.Sprintf("%s%s", id, saltKey))
+	hasher := keys.NewSHA256Hashing()
```
jakubno: shouldn't we use hmac-sha256 instead of sha256?

packages/db/queries/get_env_aliases.sql
```diff
@@ -0,0 +1,4 @@
+-- name: GetEnvAliases :many
+SELECT sqlc.embed(ea)
```
jakubno: why should we use embed in here?

packages/orchestrator/internal/sandbox/envd.go
```diff
@@ -31,6 +31,12 @@ func doRequestWithInfiniteRetries(ctx context.Context, method, address string, r
 			return nil, err
 		}
 
+		// make sure request to already authorized envd will not fail
+		// this can happen in sandbox resume and in some edge cases when previous request was success, but we continued
```
jakubno, line 35: why should it fail?

packages/api/internal/cache/templates/cache.go
```diff
@@ -69,55 +73,70 @@ func NewTemplateCache(db *db.DB) *TemplateCache {
...
 
-		c.aliasCache.cache.Set(envDB.TemplateID, envDB.TemplateID, templateInfoExpiration)
-		if envDB.Aliases != nil {
-			for _, alias := range *envDB.Aliases {
-				c.aliasCache.cache.Set(alias, envDB.TemplateID, templateInfoExpiration)
-			}
+		build = &result.EnvBuild
+		template := &result.Env
+
+		aliasesDB, err := c.db.GetEnvAliases(ctx, result.Env.ID)
```
jakubno: Can you get aliases already in the first query?

packages/api/internal/handlers/sandbox_get.go
```diff
@@ -57,35 +61,32 @@ func (a *APIStore) GetSandboxesSandboxID(c *gin.Context, id string) {
...
+		ClientID:        "00000000", // for backwards compatibility we need to return a client id
+		TemplateID:      lastSnapshot.Snapshot.EnvID,
+		SandboxID:       lastSnapshot.Snapshot.SandboxID,
+		StartedAt:       lastSnapshot.Snapshot.SandboxStartedAt.Time,
+		CpuCount:        cpuCount,
+		MemoryMB:        memoryMB,
+		EndAt:           info.GetEndTime(),
+		State:           api.Paused,
+		EnvdVersion:     lastSnapshot.EnvBuild.EnvdVersion,
+		EnvdAccessToken: nil,
```
jakubno: why should it be `nil` for paused sandbox?

packages/db/queries/get_env_with_build.sql
```diff
@@ -0,0 +1,8 @@
+-- name: GetEnvWithBuild :one
+SELECT sqlc.embed(e), sqlc.embed(eb)
+FROM "public"."envs" e
+LEFT JOIN "public"."env_aliases" ea ON ea.env_id = e.id
```
jakubno: if you want aliases, you will have to do it little bit differently. I think you can do subquery and aggregate them to array

jakubno (changes requested):
In env variables and terraform the hash is called seed, can you unify it? :pray:

tests/integration/internal/tests/envd/auth_test.go
sitole: You can do it now, because request is authorized with valid token. I think it's okay to do it. We can check envd if token is already set and you are trying to change it do different one and fail, that is possible

Problem is that sometimes envd will be initialized multiple times and if some of the requests in orchestration envd init call will fail it will block orchestrator and try it again until context is reached. This during testing efficiently blocked orchestrator from spawning new sandboxes.

packages/orchestrator/internal/sandbox/envd.go
sitole: Not sure exactly but it happen randomly. This was added to fix that. Definitely it will fail when envd is resumed.

spec/openapi.yml
sitole: I think there is not much reason for access token to be present in list, but envd version can be pretty interesting to have there.

packages/db/queries/get_env_with_build.sql
sitole: I merged env, env build and aliases into one query.

tests/integration/internal/tests/envd/auth_test.go
```diff
@@ -65,70 +60,64 @@ func TestAccessToAuthorizedPathWithoutToken(t *testing.T) {
...
-	// second call without authorization
 	sandboxEnvdInitCall(t, ctx, envdInitCall{
 		sbx:                   sbx,
 		client:                envdClient,
 		body:                  envdapi.PostInitJSONRequestBody{},
 		expectedResErr:        nil,
 		expectedResHttpStatus: http.StatusUnauthorized,
 	})
+
+	t.Cleanup(func() {
```
jakubno: NIT: it could be already on line 79

tests/integration/internal/tests/envd/auth_test.go
jakubno: I would allow to run it multiple times, but I wouldn't allow to change the token.

If user changes the token and pauses the sandbox, it will fail on resume with the same problem as you mentined

spec/openapi.yml
jakubno: yea, probably doesn't hurt, on the other side, we will probably drop it in SDK, so why to send it, i would keep this information only for sandbox detail probably

packages/orchestrator/internal/sandbox/envd.go
jakubno: I meant the comment `when previous request was success, but we continued`

tests/integration/internal/tests/envd/signatures_test.go
```diff
@@ -260,7 +259,13 @@ func TestAccessWithNotCorrespondingSignatureAndSignatureExpiration(t *testing.T)
 		setup.WithSandbox(sbx.JSON201.SandboxID, sbx.JSON201.ClientID),
 	)
 
-	assert.NoError(t, readErr)
+	require.Nil(t, readErr)
+
+	t.Cleanup(func() {
```
jakubno: couldn't this be part of `createSandbox`? also it can happen you won't clean up the sandbox, because the above require

tests/integration/internal/tests/envd/signatures_test.go
```diff
@@ -222,7 +220,13 @@ func TestDownloadWithHealthyToken(t *testing.T) {
...
```
jakubno: same here

tests/integration/internal/tests/envd/auth_test.go
```diff
@@ -138,13 +127,163 @@ func TestAccessAuthorizedPathWithOutdatedAccessToken(t *testing.T) {
 	req = connect.NewRequest(&filesystem.ListDirRequest{Path: "/"})
 	setup.SetSandboxHeader(req.Header(), sbx.JSON201.SandboxID, sbx.JSON201.ClientID)
 	setup.SetUserHeader(req.Header(), "user")
-	setup.SetAccessTokenHeader(req.Header(), envdAuthTokenA)
+	setup.SetAccessTokenHeader(req.Header(), *envdAuthTokenA)
 
 	_, err = envdClient.FilesystemClient.ListDir(ctx, req)
 	require.Error(t, err)
+
+	t.Cleanup(func() {
```
jakubno: and here

tests/integration/internal/tests/envd/auth_test.go
```diff
@@ -138,13 +127,163 @@ func TestAccessAuthorizedPathWithOutdatedAccessToken(t *testing.T) {
...
+		&envdapi.GetFilesParams{Path: &filePath, Username: "user"},
+		setup.WithSandbox(sbx.JSON201.SandboxID, sbx.JSON201.ClientID),
+		setup.WithAccessToken(*sbxMeta.EnvdAccessToken),
+	)
+
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	t.Cleanup(func() {
```
jakubno: here

tests/integration/internal/tests/envd/auth_test.go
```diff
@@ -138,13 +127,163 @@ func TestAccessAuthorizedPathWithOutdatedAccessToken(t *testing.T) {
...
+		ctx,
+		&envdapi.GetFilesParams{Path: &filePath, Username: "user"},
+		setup.WithSandbox(sbx.JSON201.SandboxID, sbx.JSON201.ClientID),
+	)
+
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	t.Cleanup(func() {
```
jakubno: here

tests/integration/internal/tests/api/sandbox_test.go
```diff
@@ -40,3 +41,68 @@ func TestSandboxCreate(t *testing.T) {
...
+
+	sbxResume, err := c.PostSandboxesSandboxIDResumeWithResponse(ctx, sbxIdWithClient, api.PostSandboxesSandboxIDResumeJSONRequestBody{}, setup.WithAPIKey())
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	assert.Equal(t, http.StatusCreated, sbxResume.StatusCode())
+	require.NotNil(t, sbxResume.JSON201)
+
+	t.Cleanup(func() {
```
jakubno, line 102: here

tests/integration/internal/tests/api/sandbox_test.go
```diff
@@ -40,3 +41,68 @@ func TestSandboxCreate(t *testing.T) {
...
+
+	// try to generate non-existing sandbox id but with real client part
+	unknownSbxId := "xxx" + sbxCreate.JSON201.SandboxID[3:] + "-" + sbxCreate.JSON201.ClientID
+
+	sbxResume, err := c.PostSandboxesSandboxIDResumeWithResponse(ctx, unknownSbxId, api.PostSandboxesSandboxIDResumeJSONRequestBody{}, setup.WithAPIKey())
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	t.Cleanup(func() {
```
jakubno, line 64: here

packages/db/migrations/20250409113306_add_envd_secured_to_snapshot.up.sql
jakubno, line 1: could you add down migration as well 🙏🏻

packages/db/queries/get_env_with_build.sql
jakubno, line 1: and this is why we have sqlc <3

packages/api/internal/sandbox/sandbox_envd_secret.go
```diff
@@ -0,0 +1,28 @@
...
+func NewEnvdAccessTokenGenerator() *EnvdAccessTokenGenerator {
+	return &EnvdAccessTokenGenerator{}
+}
+
+func (g *EnvdAccessTokenGenerator) GenerateAccessToken(id api.SandboxID) (string, error) {
+	if saltKey == "" {
+		return "", errors.New("salt key is not set")
+	}
+
+	hasher := keys.NewHMACSHA256Hashing([]byte(saltKey))
```
jakubno: wouldn't it be better to create this in init? instead of initializing hasher for each sandbox?

packages/shared/pkg/keys/hmac_sha256.go
```diff
@@ -0,0 +1,21 @@
...
+	key []byte
+}
+
+func NewHMACSHA256Hashing(key []byte) *HMACSha256Hashing {
+	return &HMACSha256Hashing{key: key}
+}
+
+func (h *HMACSha256Hashing) Hash(content []byte) string {
+	mac := hmac.New(sha256.New, h.key)
+	mac.Write(content)
```
jakubno: shouldn't you check for errors?

packages/shared/pkg/keys/hmac_sha256_test.go
```diff
@@ -0,0 +1,73 @@
...
+	key := []byte("test-key")
+	hasher := NewHMACSHA256Hashing(key)
+	content := []byte("hello world")
+	unexpectedHash := "0000000000000000000000000000000000000000000000000000000000000000"
+	actualHash := hasher.Hash(content)
+
+	if actualHash == unexpectedHash {
+		t.Errorf("hash should not match %s", unexpectedHash)
+	}
+}
```
jakubno: this test is kinda unnecessary with `TestHMACSha256Hashing_ValidHash`, isn't it?

packages/api/internal/handlers/sandbox_get.go
```diff
@@ -57,35 +62,45 @@ func (a *APIStore) GetSandboxesSandboxID(c *gin.Context, id string) {
...
 	}
 
-	memoryMB := int32(-1)
-	cpuCount := int32(-1)
+	memoryMB := int32(lastSnapshot.EnvBuild.RamMb)
+	cpuCount := int32(lastSnapshot.EnvBuild.Vcpu)
+
+	var sbxAccessToken *string = nil
+	if lastSnapshot.Snapshot.EnvSecure {
+		hashed := sandbox.NewEnvdAccessTokenGenerator()
```
jakubno: is it necessary to init the generator for each sandbox, can't there be a global generator?

jakubno (changes requested):
Could you split get and list sandbox types then? It can be confused for people using directly our API, if you don't return correct information

tests/integration/internal/tests/envd/signatures_test.go
sitole: Nice, much cleaner now!

spec/openapi.yml
sitole: new response `SandboxDetail` created and used specially for `GET /sandboxes/<sbx-id>` endpoint.
Envd version and access token removed on `ListedSandbox`.

tests/integration/internal/tests/envd/auth_test.go
sitole: Agree, fixed

spec/openapi.yml
```diff
@@ -249,6 +249,55 @@ components:
         envdVersion:
           type: string
           description: Version of the envd running in the sandbox
+        envdAccessToken:
+          type: string
+          description: Access token used for envd communication
+
+    SandboxDetailed:
```
jakubno: Can we rather name it
```suggestion
    SandboxDetail:
```

tests/integration/internal/tests/envd/auth_test.go
```diff
@@ -47,19 +54,11 @@ func TestAccessToAuthorizedPathWithoutToken(t *testing.T) {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	sbx := createSandbox(t, setup.WithAPIKey())
+	sbx := createSandbox(t, true, setup.WithAPIKey())
+	assert.NotNil(t, sbx.JSON201)
```
jakubno:
```suggestion
	require.NotNil(t, sbx.JSON201)
```

tests/integration/internal/tests/envd/auth_test.go
```diff
@@ -68,28 +67,20 @@ func TestAccessToAuthorizedPathWithoutToken(t *testing.T) {
...
 	assert.Equal(t, err.Error(), "unauthenticated: 401 Unauthorized")
 }
 
 func TestRunUnauthorizedInitWithAlreadySecuredEnvd(t *testing.T) {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	sbx := createSandbox(t, setup.WithAPIKey())
+	sbx := createSandbox(t, true, setup.WithAPIKey())
+	assert.NotNil(t, sbx.JSON201)
```
jakubno:
```suggestion
	require.NotNil(t, sbx.JSON201)
```

tests/integration/internal/tests/envd/auth_test.go
```diff
@@ -99,53 +90,162 @@ func TestRunUnauthorizedInitWithAlreadySecuredEnvd(t *testing.T) {
...
 
-func TestAccessAuthorizedPathWithOutdatedAccessToken(t *testing.T) {
+func TestChangeAccessAuthorizedToken(t *testing.T) {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	sbx := createSandbox(t, setup.WithAPIKey())
-	envdClient := setup.GetEnvdClient(t, ctx)
+	sbx := createSandbox(t, true, setup.WithAPIKey())
+	assert.NotNil(t, sbx.JSON201)
```
jakubno:
```suggestion
	require.NotNil(t, sbx.JSON201)
```

packages/api/internal/sandbox/sandbox_envd_secret.go
```diff
@@ -0,0 +1,28 @@
...
+
+type EnvdAccessTokenGenerator struct {
+}
+
+func NewEnvdAccessTokenGenerator() *EnvdAccessTokenGenerator {
+	return &EnvdAccessTokenGenerator{}
+}
+
+func (g *EnvdAccessTokenGenerator) GenerateAccessToken(id api.SandboxID) (string, error) {
+	if saltKey == "" {
```
jakubno: I would move thsi to `NewEnvdAccessTokenGenerator`, it doesn't make sense to check for every call, also I would probably pass it as a parameter, so you can more easily testable

packages/db/migrations/20250409113306_add_envd_secured_to_snapshot.down.sql
jakubno, line 1: I am really sorry, but the migration formats changed, could you update it 🙏🏻

tests/integration/internal/tests/envd/auth_test.go
```diff
@@ -3,6 +3,7 @@ package envd
 import (
 	"context"
 	"fmt"
+	"github.com/e2b-dev/infra/tests/integration/internal/utils"
```
jakubno: sort imports 🙏🏻

jakubno:
And there are still 2 unresolved issues:
https://github.com/e2b-dev/infra/pull/558#discussion_r2060210051
https://github.com/e2b-dev/infra/pull/558#pullrequestreview-2784577526

packages/api/internal/sandbox/sandbox_envd_secret.go
sitole: done

sitole:
> And there are still 2 unresolved issues: [#558 (comment)](https://github.com/e2b-dev/infra/pull/558#discussion_r2060210051) [#558 (review)](https://github.com/e2b-dev/infra/pull/558#pullrequestreview-2784577526)

Both are resolved. Thx!

jakubno (approved):

PR #559, opened 2025-04-09
Simplify passing envs in makefile

jakubno:
# Description

Instead of passing each env variable one by one, load them dynamically

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -43,6 +43,16 @@ DOMAIN_NAME=
 # for Supabase format postgresql://postgres.<username>:<password>.<host>@<your-full-url-domain.com>:<port-number>/postgres
 POSTGRES_CONNECTION_STRING=
 
+# ------------------------------------------- Optional block -----------------------------------------------------------
+# Managed Redis (default: false)
+REDIS_MANAGED=false
+# Managed Grafana (default: false), read more in ./terraform/grafana/README.md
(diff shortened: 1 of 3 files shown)


.env.template
```diff
@@ -56,17 +66,3 @@ TESTS_ENVD_PROXY=
 TESTS_SANDBOX_TEMPLATE_ID=base
 # your Team API key
 TESTS_E2B_API_KEY=
-
-# <host>:<port> no protocol
-CLICKHOUSE_CONNECTION_STRING=
```
dobrac, line 61: the clickhouse env removal should be done probably in a different PR

dobrac (changes requested):
I know it's annoying to define the variables, but we are losing with this track of which variables should be defined, etc.

Also, for the vars with the default value, we're losing ability to define these (not every env has to have correctly setup env vars in the .env file)

jakubno:
> I know it's annoying to define the variables, but we are losing with this track of which variables should be defined, etc.
>
>Also, for the vars with the default value, we're losing ability to define these (not every env has to have correctly setup env vars in the .env file)

- We can set defaults in `variables.tf`, which I consider cleaner solution than in `Makefile`
- I get the argument for explicitly defining the envs, but either the variable is required and it will fail or it is optional and it will behave the same way as it is now

PR #560, opened 2025-04-10
Don't fail build on missing logs

jakubno:
# Description

If loki crashes, it shouldn't fail build, especially if the build is still running in the background.

### DIFF
--- a/packages/api/internal/handlers/template_build_status.go
+++ b/packages/api/internal/handlers/template_build_status.go
@@ -8,8 +8,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/e2b-dev/infra/packages/shared/pkg/db"
-
 	"github.com/grafana/loki/pkg/loghttp"
 	"github.com/grafana/loki/pkg/logproto"
 	"go.uber.org/zap"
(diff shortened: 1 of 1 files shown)


packages/api/internal/handlers/template_build_status.go
```diff
@@ -4,11 +4,12 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/e2b-dev/infra/packages/shared/pkg/db"
 	"net/http"
 	"strings"
 	"time"
 
+	"github.com/e2b-dev/infra/packages/shared/pkg/db"
```
dobrac: nit group with others

packages/api/internal/handlers/template_build_status.go
```diff
@@ -105,48 +106,45 @@ func (a *APIStore) GetTemplatesTemplateIDBuildsBuildIDStatus(c *gin.Context, tem
...
-				zap.L().Error("error parsing log line", zap.Error(err), zap.String("buildID", buildID), zap.String("line", entry.Line))
+				logs = append(logs, line["message"].(string))
 			}
-
-			logs = append(logs, line["message"].(string))
 		}
+	} else {
+		errMsg := fmt.Errorf("error when returning logs for template build: %w", err)
+		telemetry.ReportError(ctx, errMsg)
+		zap.L().Error("error when returning logs for template build", zap.Error(err), zap.String("buildID", buildID))
```
dobrac, line 146: If the logs error out, don't we want to send to the user logs at least some message? E.g. that the build succeeded.

dobrac (changes requested):

packages/api/internal/handlers/template_build_status.go
jakubno: > don't we want to send to the user logs at least some message? E.g. that the build succeeded.
That's in `Status`, so that should be fine. I wanted to add some warning, but we can't add it to logs, because the offset is counted on client side. So we would need to add extra field.

dobrac: Okay

PR #563, opened 2025-04-11
Add longterm statistics for sandbox CREATE and KILL speed

dobrac:
add longterm (14 days) statistics monitoring for sandbox CREATE and KILL

(note: the Grafana gridPos.y values have to be after the inserted row all changed, it's rn somehow working, though it might not be 100% correct. We should move this to a better solution)


terraform/grafana/stack/panels/sandbox-create-latency-old.json
jakubno, line 1: is the only change the grid position or anything else

jakubno (approved):
The charts are the same as we already have above but with fixed 14 days?

terraform/grafana/stack/panels/sandbox-create-latency-old.json
dobrac: It should be just formatting change + the gridPos.y

PR #564, opened 2025-04-11
Orchestrator rolling updates with job definiton

jakubno:
# Description

- This will allow us to update orchestrator job definition without causing downtime
- Also it improves observability how many orchestrators are running for the current version

## How it works

We generate new unique id if either job definition, secret or orchestrator binary changed
We use this ID to generate a new job, which has the new job definition
We save the ID to nomad as a variable and theres's a `prestart` check, which compares the ID of the job with the latest ID (the one saved in nomad), if they don't match the orchestrator is not started

This means there will be multiple jobs for orchestrator, but new orchestrator will start only for the latest job

### DIFF
--- a/packages/nomad/orchestrator.hcl
+++ b/packages/nomad/orchestrator.hcl
@@ -1,16 +1,8 @@
-job "orchestrator-${random_id}" {
+job "orchestrator-${latest_orchestrator_job_id}" {
   type = "system"
-  datacenters = ["${gcp_zone}"]
+  node_pool = "default"
 
   priority = 90
-
-  // This constraint ensures that only one orchestrator is running at a time,
-  // now that we use differently names jobs.
-  // TODO: We cannot use this on group level, so if we want to run the script check we will need to use colliding port.
-  constraint {
-    distinct_property = "$${node.unique.id}"
-    value             = "1"
-  }
 
   group "client-orchestrator" {
     network {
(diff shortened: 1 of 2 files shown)


packages/nomad/orchestrator.hcl
```diff
@@ -30,9 +30,42 @@ job "orchestrator" {
...
+      restart {
+        attempts = 0
+      }
+
+      template {
+        destination = "local/check-placement.sh"
+        data = <<EOT
+#!/bin/bash
+
+if [ "{{with nomadVar "nomad/jobs" }}{{ .latest_orchestrator_job_id }}{{ end }}" != "${latest_orchestrator_job_id}" ]; then
```
dobrac, line 50: can you please explain how does this work? Especially when there are multiple jobs trying to place on one node?

dobrac (commented):
Nice!

Could you please describe how would a migration from the current solution look like? Is there any migration necessary?

ValentaTomas:
I think the migration could be the following:
1. Adjust the priority of the new job so it is evaluated before the old orchestrator job—it will block the old orchestrator job being deployed
2. Deploy the new job once and roll all orchestrators
3. Remove the old orchestrator job

The only question left is how to delete the old jobs that are unused.

ValentaTomas:
@jakubno The `priority` on job cannot make it so that the new job is evaluated before the currently running orchestrator? Thinking if we even need the `wait` at all.

ValentaTomas:
> The only question left is how to delete the old jobs that are unused.

Also we need to solve this before merging.

jakubno:
>> The only question left is how to delete the old jobs that are unused.
>
> Also we need to solve this before merging

There won't be that many of them, deploying new orchestrator version is now rather slow process

packages/nomad/orchestrator.hcl
```diff
@@ -1,15 +1,16 @@
-job "orchestrator" {
+job "orchestrator-${latest_orchestrator_job_id}" {
   type = "system"
-  datacenters = ["${gcp_zone}"]
+  node_pool = "default"
 
   priority = 90
```
ValentaTomas, line 5: @jakubno Shouldn't increasing this (to be bigger than the current orch job) ensure this is evaluated first?

jakubno: It just prioritize if there's resource shortage, it doesn't work for port collision

jakubno: That's why there's the `--wait` "hack"

jakubno:
Wait for #647 is deployed

PR #565, opened 2025-04-11
Report orchestrator node commit

jakubno:
# Description

Add commit sha to information reported from API, so we have better overview, what versions are running in the cluster

```json
  {
    "allocatedCPU": 0,
    "allocatedMemoryMiB": 0,
    "createFails": 0,
    "nodeID": "9c1a05eb",
    "sandboxCount": 0,
    "sandboxStartingCount": 0,
    "status": "ready",
    "version": "5a1e6ee50"
  }
```


packages/api/internal/orchestrator/client.go
```diff
@@ -61,15 +66,36 @@ func (o *Orchestrator) connectToNode(ctx context.Context, node *node.NodeInfo) e
 	buildCache := ttlcache.New[string, interface{}]()
 	go buildCache.Start()
 
+	resp, err := o.httpClient.Get(fmt.Sprintf("http://%s/health", node.OrchestratorAddress))
```
dobrac: Might be worth extracting all the /health client call to a separate function

packages/api/internal/orchestrator/client.go
```diff
@@ -61,15 +66,36 @@ func (o *Orchestrator) connectToNode(ctx context.Context, node *node.NodeInfo) e
...
+
+	// Check if the node is healthy
+	var healthResp e2bHealth.Response
+	err = json.NewDecoder(resp.Body).Decode(&healthResp)
+	if err != nil {
+		return fmt.Errorf("failed to decode health response: %w", err)
+	}
+
+	if healthResp.Status != e2bHealth.Healthy {
+		return fmt.Errorf("node is not healthy: %s", healthResp.Status)
```
dobrac: We should keep track of unhealthy nodes. I would propose rather setting different status than not including the node at all

dobrac (changes requested):
lgtm, just minor comments

packages/api/internal/orchestrator/client.go
```diff
@@ -61,15 +67,33 @@ func (o *Orchestrator) connectToNode(ctx context.Context, node *node.NodeInfo) e
...
+
+		version = nodeStatus.Version
+	}
+
 	n := &Node{
 		Client:         client,
 		buildCache:     buildCache,
 		sbxsInProgress: smap.New[*sbxInProgress](),
-		status:         api.NodeStatusReady,
+		status:         status,
```
dobrac: if the status is set here as unhealthy, I don't see if it can ever recover

dobrac (changes requested):

dobrac (approved):

PR #566, opened 2025-04-11
Show long term sandbox creation stats by template alias

dobrac:
Improves Grafana panel for long term sandbox creation time by splitting by the template id


ValentaTomas:
Ready for merge @dobrac?

dobrac:
> Ready for merge @dobrac?

Let's keep it as a draft for now until the required stats are deployed, latest Grafana updates applied and maybe update the way we define dashboards (if possible).

Currently this draft contains only 95th percentile in one graph. Maybe we would like to show all the percentiles, but have many charts. Maybe even moving this to a separate dashboard.

ValentaTomas:
This is no longer relevant in this repo, right?

PR #568, opened 2025-04-15
Always log team id for authorized requests

sitole:


packages/api/main.go
```diff
@@ -136,6 +135,33 @@ func NewGinServer(ctx context.Context, logger *zap.Logger, apiStore *handlers.AP
 			}),
 	)
 
+	r.Use(
+		// Request logging must be executed after authorization (if required) is done,
+		// so that we can log team ID.
+		customMiddleware.ExcludeRoutes(
+			func(c *gin.Context) {
```
sitole, line 142: we need to call gin zap logger manually because in we need to extract team in-request

PR #573, opened 2025-04-22
Fix integration tests

jakubno:
# Description

- Fixes file path when using list dir with depth
- Fixes integration test for listing dir


0div (approved):
thx!

PR #578, opened 2025-04-22
Reconfigure reverse proxies timeouts and keepalive

ValentaTomas:
The previous configuration of proxies might be affecting parts of traffic to sandboxes and creates unexpected errors.

### DIFF
--- a/packages/orchestrator/internal/proxy/proxy.go
+++ b/packages/orchestrator/internal/proxy/proxy.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"html/template"
+	"net"
 	"net/http"
 	"net/http/httputil"
 	"net/url"
@@ -26,6 +27,12 @@ var proxyBrowser502PageHtml string
 
 var browserRegex = regexp.MustCompile(`(?i)mozilla|chrome|safari|firefox|edge|opera|msie`)
 var browserTemplate = template.Must(template.New("template").Parse(proxyBrowser502PageHtml))
+
+const (
+	MaxIdleConnections    = 32768
+	MaxConnectionDuration = 24 * time.Hour
+	IdleTimeout           = 630 * time.Second
+)
 
 type htmlTemplateData struct {
 	SandboxId   string
@@ -50,7 +57,13 @@ type SandboxProxy struct {
 }
 
 func New(port uint) *SandboxProxy {
-	server := &http.Server{Addr: fmt.Sprintf(":%d", port)}
+	server := &http.Server{
+		Addr:              fmt.Sprintf(":%d", port),
+		ReadTimeout:       MaxConnectionDuration,
+		WriteTimeout:      MaxConnectionDuration,
+		IdleTimeout:       IdleTimeout,
+		ReadHeaderTimeout: 20 * time.Second,
+	}
 
 	return &SandboxProxy{
 		server:    server,
@@ -70,12 +83,15 @@ func (p *SandboxProxy) Start() error {
 	// similar values to our old the nginx configuration
 	serverTransport := &http.Transport{
 		Proxy:                 http.ProxyFromEnvironment,
-		MaxIdleConns:          1024,              // Matches worker_connections
-		MaxIdleConnsPerHost:   8192,              // Matches keepalive_requests
-		IdleConnTimeout:       620 * time.Second, // Matches keepalive_timeout
-		TLSHandshakeTimeout:   10 * time.Second,  // Similar to client_header_timeout
-		ResponseHeaderTimeout: 24 * time.Hour,    // Matches proxy_read_timeout
-		DisableKeepAlives:     true,              // Disable keep-alives, envd doesn't support idle connections
+		MaxIdleConnsPerHost:   MaxIdleConnections,
+		IdleConnTimeout:       IdleTimeout,
+		TLSHandshakeTimeout:   20 * time.Second,
+		ResponseHeaderTimeout: 20 * time.Second,
+		DialContext: (&net.Dialer{
+			Timeout:   30 * time.Second, // Connect timeout (no timeout by default)
+			KeepAlive: 30 * time.Second, // Lower than our http keepalives (50 seconds)
+		}).DialContext,
+		DisableCompression: true, // No need to request or manipulate compression
 	}
 
 	p.server.Handler = http.HandlerFunc(p.proxyHandler(serverTransport))
--- a/packages/client-proxy/main.go
+++ b/packages/client-proxy/main.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"log"
 	"math/rand"
+	"net"
 	"net/http"
 	"net/http/httputil"
 	"net/url"
@@ -41,6 +42,9 @@ const (
 	port                  = 3002
 	orchestratorProxyPort = 5007 // orchestrator proxy port
 	maxRetries            = 3
+	maxConnectionDuration = 24 * time.Hour    // The same as the current max sandbox duration
+	maxIdleConnections    = 32768             // Reasonably big number that is lower than the number of available ports
+	maxIdleTimeout        = 620 * time.Second // This should be ideally bigger than the previous downstream and lower than then next upstream timeout so the closing is not from the most upstream server.
 )
 
 var (
@@ -245,17 +249,26 @@ func run() int {
 	}()
 
 	// Proxy request to the correct node
-	server := &http.Server{Addr: fmt.Sprintf(":%d", port)}
+	server := &http.Server{
+		Addr:              fmt.Sprintf(":%d", port),
+		ReadTimeout:       maxConnectionDuration,
+		WriteTimeout:      maxConnectionDuration,
+		IdleTimeout:       maxIdleTimeout,
+		ReadHeaderTimeout: 20 * time.Second,
+	}
 
 	// similar values to our old the nginx configuration
 	transport := &http.Transport{
 		Proxy:                 http.ProxyFromEnvironment,
-		MaxIdleConns:          1024,              // Matches worker_connections
-		MaxIdleConnsPerHost:   8192,              // Matches keepalive_requests
-		IdleConnTimeout:       620 * time.Second, // Matches keepalive_timeout
-		TLSHandshakeTimeout:   10 * time.Second,  // Similar to client_header_timeout
-		ResponseHeaderTimeout: 24 * time.Hour,    // Matches proxy_read_timeout
-		DisableKeepAlives:     true,              // Disable keep-alive, not supported
+		MaxIdleConnsPerHost:   maxIdleConnections,
+		IdleConnTimeout:       maxIdleTimeout,
+		TLSHandshakeTimeout:   20 * time.Second,
+		ResponseHeaderTimeout: 20 * time.Second,
+		DialContext: (&net.Dialer{
+			Timeout:   10 * time.Second, // Connect timeout—can be lower between proxies
+			KeepAlive: 30 * time.Second, // Lower than our http keepalives (50 seconds)
+		}).DialContext,
+		DisableCompression: true, // No need to request or manipulate compression
 	}
 
 	server.Handler = http.HandlerFunc(proxyHandler(transport))


ValentaTomas:
I'll revert the last commit, removing some additional configuration, if the integration test problems are because of flakiness.

ValentaTomas:
https://linear.app/e2b/issue/E2B-1739/command-in-sdk-stops-with-[unknown]-terminated-error

packages/shared/pkg/reverse-proxy/proxy.go
```diff
@@ -0,0 +1,230 @@
...
+)
+
+type routingTargetContextKey struct{}
+
+// RoutingTarget contains information about where to route the request.
+type RoutingTarget struct {
+	Url       *url.URL
+	SandboxId string
+	Logger    *zap.Logger
+	// ConnectionKey is used for identifying which keepalive connections are not the same so we can prevent uninted reuse.
```
jakubno:
```suggestion
	// ConnectionKey is used for identifying which keepalive connections are not the same so we can prevent unintended reuse.
```

packages/orchestrator/internal/proxy/proxy.go
```diff
@@ -1,211 +1,73 @@
...
+				Host:   fmt.Sprintf("%s:%d", sbx.Slot.HostIP(), port),
+			}
 
-func (p *SandboxProxy) isBrowser(userAgent string) bool {
-	return browserRegex.MatchString(userAgent)
+			return &reverse_proxy.RoutingTarget{
+				Url:       url,
+				SandboxId: sbx.Config.SandboxId,
+				// We need to include sandboxId to prevent reuse of connection to the same IP:port pair by different sandboxes reusing the network slot.
+				ConnectionKey: fmt.Sprintf("%s", sbx.Config.SandboxId),
```
jakubno:
```suggestion
				ConnectionKey: sbx.Config.SandboxId,
```

packages/client-proxy/internal/proxy/proxy.go
```diff
@@ -0,0 +1,114 @@
...
+		Scheme: "http",
+		Host:   fmt.Sprintf("%s:%d", node, orchestratorProxyPort),
+	}
+
+	return &reverse_proxy.RoutingTarget{
+		Url:       url,
+		SandboxId: sandboxId,
+		Logger:    logger,
+		// We need to include sandboxId to prevent reuse of connection to the same IP:port pair by different sandboxes reusing the network slot.
+		ConnectionKey: fmt.Sprintf("%s", sandboxId),
```
jakubno:
```suggestion
		ConnectionKey: sandboxId,
```

packages/shared/pkg/reverse-proxy/error.go
```diff
@@ -0,0 +1,54 @@
...
+	err *template.TemplatedError[T],
+	logger *zap.Logger,
+) error {
+	if isBrowser(r) {
+		body, buildErr := err.BuildHtml()
+		if buildErr != nil {
+			return buildErr
+		}
+
+		w.WriteHeader(http.StatusBadGateway)
```
jakubno: shouldn't the status code be taken from error?

packages/shared/pkg/reverse-proxy/proxy.go
```diff
@@ -0,0 +1,230 @@
...
+				} else {
+					t.Logger.Debug("Reverse proxy response",
+						zap.Int("status_code", r.StatusCode),
+					)
+				}
+
+				return nil
+			},
+			// Ideally we would add info about sandbox to each error log, but there is no easy way right now.
+			ErrorLog: proxyLogger,
```
jakubno: you could do something like custom logger, which would do `fmt. Sprintf("sandbox %s: %s", sandboxId, message)` right?

packages/shared/pkg/reverse-proxy/proxy.go
```diff
@@ -0,0 +1,230 @@
...
+				}).DialContext(ctx, network, addr)
+				if err == nil {
+					counter.Add(1)
+				}
+				return conn, err
+			},
+			DisableCompression: true, // No need to request or manipulate compression
+		}
+
+		proxyLogger, _ := zap.NewStdLogAt(zap.L(), zap.ErrorLevel)
```
jakubno: why are you not checking the error?

packages/shared/pkg/reverse-proxy/proxy.go
```diff
@@ -0,0 +1,230 @@
...
+		proxies.Add(connectionKey, proxy)
+
+		return proxy
+	}
+
+	return func(w http.ResponseWriter, r *http.Request) {
+		if activeConnections != nil {
+			// TODO: Won't cancellation of the request make adding/removing from the counter unpredictable?
+			// We should probably use observable gauge and separate counter without context.
+			// Also not 100% sure if this handled idle connections/streaming correctly.
```
jakubno: In looked okay in grafana, but I think we should use some global context

packages/shared/pkg/reverse-proxy/proxy.go
```diff
@@ -0,0 +1,230 @@
...
+	getProxy := func(connectionKey string) *httputil.ReverseProxy {
+		proxiesMu.Lock()
+		defer proxiesMu.Unlock()
+
+		proxy, ok := proxies.Get(connectionKey)
+		if ok {
+			return proxy
+		}
+
+		transport := &http.Transport{
```
jakubno: By creating more transports this way, RAM consumption will go up a lot. In the initial setup, we didn't reuse the transport. 

Reusing them for each sandbox should work well in the orchestrator, but it might be an issue in the client proxy. Could you check the impact on memory before and after this change by simulating increased traffic?

jakubno (changes requested):

packages/shared/pkg/reverse-proxy/proxy_test.go
```diff
@@ -0,0 +1,418 @@
...
+		getRoutingTarget,
+	)
+
+	// Start the proxy server
+	go func() {
+		_ = proxy.ListenAndServe()
+	}()
+
+	// Wait for the proxy to start
+	time.Sleep(100 * time.Millisecond)
```
jakubno: isn't there a better way to tell that the proxy has started ,like using ConnState when starting the server?

packages/shared/pkg/reverse-proxy/proxy_test.go
```diff
@@ -0,0 +1,418 @@
...
+
+	return proxy, uint(port), nil
+}
+
+func TestProxyRoutesToTargetServer(t *testing.T) {
+	listener, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		t.Fatalf("failed to create listener: %v", err)
+	}
```
jakubno: just FYI you can use `require.Nil`

packages/shared/pkg/reverse-proxy/proxy.go
ValentaTomas: Global context could make it so that the reqeust gets stuck.
I'm really thinking that we should not otel meter for draining.

ValentaTomas: We can skip for now, but will keep the comment.

packages/shared/pkg/reverse-proxy/proxy.go
ValentaTomas: Yes, when I switched the caching this became viable again. Will add.

packages/shared/pkg/reverse-proxy/proxy_test.go
ValentaTomas: ConnState would be for the connections. The other way to do this would probably be checking if the port is listening, but with the tests here that brings complications because:
- Keeping track of number of connections is important for the test
- We are reusing the port for the reuse connection tests so it might cause more confusion there

ValentaTomas: Resolved by reusing listener

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -101,9 +99,12 @@ func (s *server) Create(ctxConn context.Context, req *orchestrator.SandboxCreate
 				return false
 			}
 
-			return sbx.CleanupID == v.CleanupID
+			return sbx.StartID == v.StartID
 		})
 
+		// Remove the proxies assigned to the sandbox from the pool to prevent them from being reused.
+		s.proxy.RemoveFromPool(sbx.StartID)
```
dobrac, line 106: Is this called at the correct time in terms of that the resumed sandbox can have different ip? (Because the removal callback can already fail at this point as there might be resumed sandbox running)

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -240,7 +238,6 @@ func (s *server) Pause(ctx context.Context, in *orchestrator.SandboxPauseRequest
 		return nil, status.New(codes.NotFound, errMsg.Error()).Err()
 	}
 
-	s.dns.Remove(in.SandboxId, sbx.Slot.HostIP())
```
dobrac, line 243: Are we completely moved away from using the NGINX proxy on the orchestrator?

ValentaTomas: Yes, it should be so. I think it is still running as a job @sitole, but can be removed, right?

packages/orchestrator/internal/server/sandboxes.go
ValentaTomas: Proxying for the specific sandbox is disabled immediatelly when removed from the sandboxes map, then the connections that are keyed to the sandbox `StartID` (unique to each create call in orchstrator) are removed right after the cleanup is done. They do not share anything with the possible new connections to any new sandbox or a resumed sandbox. If the removeCb fails nothing changes as eventhough the sandbox is "still" in the map the connection the sandbox uses in the orchestrator proxy depend on the `StartID`.

packages/orchestrator/internal/server/sandboxes.go
jakubno: yes, we are completely moved

sitole: @ValentaTomas will prepare PR with removal.

sitole: Its actually already merged https://github.com/e2b-dev/infra/pull/544

ValentaTomas (commented):
In Grafana the metrics from proxy counters are exposed as:
- `client_proxy_proxy_server_connections_open`
- `client_proxy_proxy_pool_size`
- `client_proxy_proxy_pool_connections_open`

- `orchestrator_proxy_server_connections_open`
- `orchestrator_proxy_pool_size`
- `orchestrator_proxy_pool_connections_open`

packages/shared/pkg/proxy/pool/destination.go
```diff
@@ -0,0 +1,20 @@
...
+	"go.uber.org/zap"
+)
+
+type DestinationContextKey struct{}
+
+// Destination contains information about where to route the request.
+type Destination struct {
+	Url           *url.URL
+	SandboxId     string
+	SandboxPort   *uint64
```
jakubno: why isn't this just a `unit64`?

packages/shared/pkg/proxy/handler.go
```diff
@@ -0,0 +1,92 @@
...
+
+			return
+		}
+
+		if err != nil {
+			zap.L().Error("failed to route request", zap.Error(err), zap.String("host", r.Host))
+			http.Error(w, fmt.Sprintf("Unexpected error when routing request: %s", err), http.StatusInternalServerError)
+
+			return
+		}
```
jakubno, line 83: wouldn't it make sense to have it like this? (error handling only if `err != nil`)
```suggestion
		if err != nil {
		        var invalidHostErr *ErrInvalidHost
		        if errors.As(err, &invalidHostErr) {
			        zap.L().Warn("invalid host", zap.String("host", r.Host))
			        http.Error(w, "Invalid host", http.StatusBadRequest)
        
			        return
		        }
        
		        var invalidPortErr *ErrInvalidSandboxPort
		        if errors.As(err, &invalidPortErr) {
			        zap.L().Warn("invalid sandbox port", zap.String("host", r.Host))
			        http.Error(w, "Invalid sandbox port", http.StatusBadRequest)
        
			        return
		        }
        
		        var notFoundErr *ErrSandboxNotFound
		        if errors.As(err, &notFoundErr) {
			        zap.L().Warn("sandbox not found", zap.String("host", r.Host))
        
			        err := template.
				        NewSandboxNotFoundError(notFoundErr.SandboxId, r.Host).
				        HandleError(w, r)
			        if err != nil {
				        zap.L().Error("failed to handle sandbox not found error", zap.Error(err))
				        http.Error(w, "Failed to handle sandbox not found error", http.StatusInternalServerError)
        
				        return
			        }
        
			        return
		        }

			zap.L().Error("failed to route request", zap.Error(err), zap.String("host", r.Host))
			http.Error(w, fmt.Sprintf("Unexpected error when routing request: %s", err), http.StatusInternalServerError)

			return
		}
```

packages/client-proxy/internal/proxy/proxy.go
```diff
@@ -0,0 +1,132 @@
...
+
+			// We've resolved the node to proxy the request to
+			logger.Debug("Proxying request")
+
+			url := &url.URL{
+				Scheme: "http",
+				Host:   fmt.Sprintf("%s:%d", node, orchestratorProxyPort),
+			}
+
+			return &pool.Destination{
```
jakubno, line 101: Don't you want to set explicitly a constant as a `ConnectionKey` with a comment explaining it uses url and then if they match `ConnectionKey` is used, so different nodes won't try to reuse the connection

packages/shared/pkg/proxy/pool/client.go
```diff
@@ -0,0 +1,141 @@
...
+		transport: transport,
+		ReverseProxy: httputil.ReverseProxy{
+			Transport: transport,
+			Rewrite: func(r *httputil.ProxyRequest) {
+				t, ok := r.In.Context().Value(DestinationContextKey{}).(*Destination)
+				if !ok {
+					zap.L().Error("failed to get routing destination from context")
+
+					// Error from this will be later caught as r.Host == "" in the ErrorHandler
+					r.SetURL(r.In.URL)
```
jakubno, line 66: is this a hack? 👀

packages/shared/pkg/proxy/pool/client.go
```diff
@@ -0,0 +1,141 @@
...
+				t, ok := r.Context().Value(DestinationContextKey{}).(*Destination)
+				if !ok {
+					zap.L().Error("failed to get routing destination from context")
+
+					http.Error(w, "Failed to route request to sandbox", http.StatusInternalServerError)
+
+					return
+				}
+
+				if t.SandboxPort == nil {
```
jakubno: how can this happen?

packages/shared/pkg/proxy/pool/client.go
```diff
@@ -0,0 +1,141 @@
...
+
+					return
+				}
+
+				r.SetURL(t.Url)
+				r.SetXForwarded()
+				r.Out.Host = r.In.Host
+			},
+			ErrorHandler: func(w http.ResponseWriter, r *http.Request, err error) {
+				if r.Host == "" {
```
jakubno, line 76: how can this happen?

packages/shared/pkg/proxy/pool/pool.go
```diff
@@ -0,0 +1,120 @@
...
+
+			logger = l
+		} else {
+			l, err := zap.NewStdLogAt(zap.L(), zap.ErrorLevel)
+			if err != nil {
+				zap.L().Warn("failed to create logger", zap.Error(err))
+			}
+
+			logger = l
+		}
```
jakubno: couldn't you check if `d.SandboxId != ""`, it would make more sense?

also this could be probably simplified something like this
```suggestion
		withFields := make([]zap.Filed, 0)
		if d.ConnectionKey != "" {
		         withFields = append(withFields, zap.String("sandbox_id", d.SandboxId))
		}
		
		l, err := zap.NewStdLogAt(zap.L().With(withFields...), zap.ErrorLevel)
			if err != nil {
				zap.L().Warn("failed to create logger", zap.Error(err))
			}

			logger = l
```

packages/shared/pkg/proxy/pool/pool.go
```diff
@@ -0,0 +1,120 @@
...
+			logger,
+		)
+	})
+}
+
+func (p *ProxyPool) Close(connectionKey string) {
+	for _, key := range p.keys(connectionKey) {
+		p.pool.RemoveCb(key, func(key string, proxy *proxyClient, exists bool) bool {
+			if proxy != nil {
+				proxy.closeIdleConnections()
```
jakubno, line 94: is this enough? If you kill the sandbox, will all connections be marked as idle already?

ValentaTomas: There is no way to effectively kill them without hijacking the transport more.
But should cover all possible situations:
1. They are idle and killed
2. They are active and they will break as the sandbox is killed
3. ("fallback") They will 100% be killed after the TCP probe keepalive check fails

packages/shared/pkg/proxy/pool/client.go
ValentaTomas: Maybe? This was the only way to propagate the error from the rewrite that I found without panic.

packages/shared/pkg/proxy/pool/client.go
ValentaTomas: It should not happen when normally using the proxy, but as the error from passing context needs to be handled (otherwise it *could* cause panic then) I added this mechanism that propagates it.

packages/shared/pkg/proxy/proxy_test.go
```diff
@@ -0,0 +1,407 @@
...
+		}, nil
+	}
+
+	proxy, port, err := newTestProxy(getDestination)
+	if err != nil {
+		t.Fatalf("failed to create proxy: %v", err)
+	}
+	defer proxy.Close()
+
+	assert.Equal(t, proxy.TotalPoolConnections(), uint64(0))
```
jakubno, line 160:
```suggestion
	assert.Equal(t, proxy.TotalPoolConnections(), uint64(0))
	assert.Equal(t, backend.RequestCount(), uint64(0))
```

packages/shared/pkg/proxy/proxy_test.go
```diff
@@ -0,0 +1,407 @@
...
+	backend2, err := newTestBackend(listener, "backend-2")
+	if err != nil {
+		t.Fatalf("failed to create second backend: %v", err)
+	}
+	defer backend2.Close()
+
+	// Make request to second backend
+	resp2, err := http.Get(proxyURL)
+	if err != nil {
+		t.Fatalf("unexpectedly got a response from the second backend: %v", resp2.StatusCode)
```
jakubno: the error message doesn't sound right

packages/client-proxy/internal/proxy/proxy.go
ValentaTomas: Added constant key with a description. The reuse prevention is explained in detail in the orchestrator proxy connection key passing.

packages/shared/pkg/proxy/pool/destination.go
ValentaTomas: Changed handling to allow this change.

packages/shared/pkg/proxy/proxy_test.go
ValentaTomas: This should be a "failing" test (// This is a test that verify that the proxy reuse fails when the backend changes.)
Does it still sound strange?

packages/shared/pkg/proxy/pool/pool.go
ValentaTomas: Refactord to more clearly express what should happen.

packages/shared/pkg/proxy/handler.go
ValentaTomas: Not really a fan of nesting more than necessary.
I was thinking about short-circuity with `if err == nil`, but that if also confusing.

packages/shared/pkg/proxy/proxy_test.go
jakubno: but you fatal if you get an error and then the message says got a response

jakubno (approved):

ValentaTomas:
Example of the dashboard for the proxies here—https://e2bstagingtomas.grafana.net/d/0c658054-d707-403b-9957-cab464479a9c

PR #579, opened 2025-04-22
Add proxying on `sandbox.*` subdomain

ValentaTomas:
To improve connection reuse and speed to sandbox request we want to add proxy that uses path segments instead of subdomain to get port and sandbox ID.


ValentaTomas:
Closing for now—will reopen when we return to this.

PR #588, opened 2025-04-25
Fix Docker Reverse Proxy requiring insecureRepositories

dobrac:


packages/docker-reverse-proxy/internal/handlers/login.go
jakubno, line 1: this isn't used anymore

jakubno (changes requested):
Actually now I am getting 
```
There was an issue during Docker authentication. Please follow the workaround steps from https://e2b.dev/docs/troubleshooting/templates/build-authentication-error and then continue.
```

image
image

dobrac:
> Actually now I am getting
> 
> ```
> There was an issue during Docker authentication. Please follow the workaround steps from https://e2b.dev/docs/troubleshooting/templates/build-authentication-error and then continue.
> ```
> 
> image image

Should be fixed now by:
- adding back authorization to the /v2/ endpoint
- not require authorization for PATCH request to the artifacts-upload path (Some registries (especially Artifact Registry / GCR / ECR) return a pre-signed URL during the initial POST. If the URL is presigned, the Docker client does not need Authorization on subsequent PATCH)

packages/docker-reverse-proxy/internal/handlers/login.go
dobrac: it's again

packages/docker-reverse-proxy/internal/handlers/token.go
```diff
@@ -18,47 +19,43 @@ type DockerToken struct {
...
 
-			return fmt.Errorf("invalid access token")
-		}
+		return fmt.Errorf("invalid access token")
+	}
 
-		w.Header().Set("Content-Type", "application/json")
-		w.WriteHeader(http.StatusOK)
+	scope := r.URL.Query().Get("scope")
+	templateID := "not-yet-known"
```
jakubno: NIT: would it make sense to have this inside the if statement

jakubno (approved):

PR #602, opened 2025-04-30
Setup clickhouse in cluster mode

jakubno:
# Description

- Setup clickhouse in cluster mode
- Use dedicated machines for clickhouse servers
- Deploy it only for dev cluster until it will be production ready
- Use persistent disk for clickhouse metadata (needed for persisting tables after replacing the machine)


packages/api/main.tf
```diff
@@ -111,14 +111,6 @@ resource "google_secret_manager_secret_version" "api_secret_value" {
   secret_data = random_password.api_secret.result
 }
 
-resource "google_secret_manager_secret" "clickhouse_password" {
```
jakubno, line 114: wasn't used

packages/cluster/scripts/run-clickhouse-nomad.sh
jakubno, line 1: We could generalize this and have only one run nomad template, but I would do it in separate PR

packages/cluster/scripts/start-clickhouse.sh
jakubno, line 1: Same here, this could be generalized and we could have only 1 template for starting any VM

packages/init/main.tf
```diff
@@ -223,29 +223,6 @@ resource "google_artifact_registry_repository_iam_member" "orchestration_reposit
   depends_on = [time_sleep.artifact_registry_api_wait_90_seconds]
 }
 
-resource "random_password" "clickhouse_password" {
```
jakubno, line 226: moved closer to the jobs as we will probably want to have multiple users

packages/nomad/clickhouse.hcl
```diff
@@ -1,126 +1,153 @@
...
       config {
         image = "clickhouse/clickhouse-server:${clickhouse_version}"
-        ports = ["clickhouse", "clickhouse_http"]
+        ports = ["clickhouse-server"]
 
         ulimit {
           nofile = "262144:262144"
         }
 
+        extra_hosts = [
```
jakubno, line 66: needed for self reference

.env.template
```diff
@@ -36,6 +36,11 @@ BUILD_MACHINE_TYPE=
 # e.g. 1
 BUILD_CLUSTER_SIZE=
 
+# eg e2-standard-4
+CLICKHOUSE_MACHINE_TYPE
```
dobrac:
```suggestion
# e.g. e2-standard-4
CLICKHOUSE_MACHINE_TYPE=
```

main.tf
```diff
@@ -95,10 +95,12 @@ module "cluster" {
...
-  server_machine_type = var.server_machine_type
-  client_machine_type = var.client_machine_type
-  api_machine_type    = var.api_machine_type
-  build_machine_type  = var.build_machine_type
+  server_machine_type     = var.server_machine_type
+  client_machine_type     = var.client_machine_type
+  api_machine_type        = var.api_machine_type
+  build_machine_type      = var.build_machine_type
+  clickhouse_machine_type = var.clickhouse_machine_type
+  clickhouse_cluster_size = var.clickhouse_cluster_size
```
dobrac: nit: how about moving the cluster size to the group up with sizes definitions

packages/buckets/main.tf
```diff
@@ -79,6 +79,22 @@ resource "google_storage_bucket" "fc_env_pipeline_bucket" {
...
+
+  public_access_prevention    = "enforced"
+  storage_class               = "STANDARD"
+  uniform_bucket_level_access = true
+
+  soft_delete_policy {
+    retention_duration_seconds = 0
+  }
+
+  force_destroy = true
```
dobrac: do we want this?

packages/cluster/scripts/start-clickhouse.sh
```diff
@@ -0,0 +1,87 @@
...
+/opt/consul/bin/run-consul.sh --client \
+    --consul-token "${CONSUL_TOKEN}" \
+    --cluster-tag-name "${CLUSTER_TAG_NAME}" \
+    --enable-gossip-encryption \
+    --gossip-encryption-key "${CONSUL_GOSSIP_ENCRYPTION_KEY}" \
+    --dns-request-token "${CONSUL_DNS_REQUEST_TOKEN}" &
+
+/opt/nomad/bin/run-nomad.sh --consul-token "${CONSUL_TOKEN}" &
+
+# Install clickhouse client
```
jakubno: this isn't required, but it makes it easier to debug or check clickhouse, because it isn't accessible outside the cluster

packages/cluster/clickhouse-cluster/main.tf
```diff
@@ -0,0 +1,146 @@
...
+  auto_healing_policies {
+    health_check      = google_compute_health_check.nomad_check.id
+    initial_delay_sec = 600
+  }
+
+  # Server is a stateful cluster, so the update strategy used to roll out a new GCE Instance Template must be
+  # a rolling update.
+  update_policy {
+    type                  = var.environment == "dev" ? "PROACTIVE" : "OPPORTUNISTIC"
+    minimal_action        = "REPLACE"
```
jakubno: To prevent having stale data on machine. The problem is with data from old startup script, they are there, which can cause unexpected behavior

packages/cluster/scripts/start-clickhouse.sh
```diff
@@ -0,0 +1,124 @@
...
+touch /etc/systemd/resolved.conf.d/consul.conf
+cat <<EOF >/etc/systemd/resolved.conf.d/consul.conf
+[Resolve]
+DNS=127.0.0.1:8600
+DNSSEC=false
+Domains=~consul
+EOF
+
+# Expose systemd-resolved’s DNS stub on the Docker bridge so that
+# containers can resolve *.consul names.
```
dobrac, line 82: How is this working in the API? I think it resolves consul names as well

packages/nomad/clickhouse.hcl
```diff
@@ -1,126 +1,175 @@
...
+          "clickhouse-server-${i + 1}.service.consul:127.0.0.1",
+        ]
 
         volumes = [
-          "local/config.xml:/etc/clickhouse-server/config.d/gcs.xml",
-          # disabled while testing but will pass password to orchestrator in the future
+          "/clickhouse/data/clickhouse-server-${i + 1}:/var/lib/clickhouse",
```
dobrac, line 71: is this safe to do with the server "id"? Could there be a possibility of the servers interchanging the data -> causing issues with routing, etc?

packages/nomad/clickhouse.hcl
```diff
@@ -1,126 +1,175 @@
...
         </policies>
     </storage_configuration>
+    <remote_servers replace="true">
+      <cluster>
+        <!-- a secret for servers to use to communicate to each other  -->
+        <secret>${server_secret}</secret>
+        %{ for j in range("${server_count}") }
+        <shard>
+          <replica>
+            <host>clickhouse-server-${j + 1}.service.consul</host>
```
dobrac, line 144: is the server id stable enough?

packages/cluster/clickhouse-cluster/main.tf
```diff
@@ -0,0 +1,146 @@
...
+
+  instance_group_manager = google_compute_instance_group_manager.cluster.name
+
+  name = "${var.cluster_name}-${each.key}"
+
+  preserved_state {
+    disk {
+      device_name = google_compute_disk.stateful_disk[each.key].name
+      mode        = "READ_WRITE"
+      delete_rule = "ON_PERMANENT_INSTANCE_DELETION"
```
dobrac: what happens when you downscale and upscale again?

packages/nomad/clickhouse.hcl
```diff
@@ -1,126 +1,175 @@
...
+          "clickhouse-server-${i + 1}.service.consul:127.0.0.1",
+        ]
 
         volumes = [
-          "local/config.xml:/etc/clickhouse-server/config.d/gcs.xml",
-          # disabled while testing but will pass password to orchestrator in the future
+          "/clickhouse/data/clickhouse-server-${i + 1}:/var/lib/clickhouse",
+          "local/config.xml:/etc/clickhouse-server/config.d/config.xml",
           "local/users.xml:/etc/clickhouse-server/users.d/users.xml",
+          "local/macros.xml:/etc/clickhouse-server/config.d/macros.xml",
```
dobrac: why are these needed?

packages/cluster/scripts/start-clickhouse.sh
jakubno: The network is set to `host`

packages/nomad/clickhouse.hcl
jakubno: Because of this constraint it's safe. The i-th VM will always run i-th clickhouse server 
```
    constraint {
      attribute = "$${meta.job_constraint}"
      value     = "${job_constraint_prefix}-${i + 1}"
    }
```

packages/nomad/clickhouse.hcl
jakubno: It's from `range`, so it's always from `1` to `server_count` (because of the +1)

packages/nomad/clickhouse.hcl
jakubno: right, macros aren't needed
Users config is needed to create users, set passwords, etc : https://clickhouse.com/docs/operations/settings/settings-users

packages/buckets/main.tf
jakubno: Removed

packages/cluster/clickhouse-cluster/main.tf
jakubno: Made it permanent for prod environment

PR #603, opened 2025-05-02
Log a warning in envd indicating that the sandbox exceeded the threshold cpu/memory

0div:
This would happen upon getting metrics at set interval already set in orchestrator checks.

### DIFF
--- a/packages/envd/internal/api/store.go
+++ b/packages/envd/internal/api/store.go
@@ -10,6 +10,11 @@ import (
 	"github.com/rs/zerolog"
 )
 
+const (
+	memThresholdPct = 80
+	cpuThresholdPct = 80
+)
+
 type API struct {
 	logger      *zerolog.Logger
 	accessToken *string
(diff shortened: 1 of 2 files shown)


jakubno (changes requested):
Can we have the logic in orchestrator? It won't require the templates to be rebuilt and when we want to move to metrics, we can prevent doubling the logs. We already have most of the logic there as well

0div:
> Can we have the logic in orchestrator? It won't require the templates to be rebuilt and when we want to move to metrics, we can prevent doubling the logs. We already have most of the logic there as well

We want this warning in sandbox logs, unless there is a non-convoluted way to do that outside of the sandbox, im all for it.

jakubno:
It is, in `LogMetrics` instead of `sbxlogger.E(s).Metrics` you would call `sbxlogger.E(s).Warn()`, please comment out the current code

0div:
> ```shell
> stress-ng --vm 1 --vm-bytes 512M --vm-method all --verify --timeout 60s --metrics-brief
> ```

ok I got it to show in `e2b sbx lg` when logged from orchestrator: 

```
[2025-05-05 22:21:29.064Z] WARN  { buildID: 'unknown', category: 'default', logger: 'orchestrator', mem_threshold_percent: 80, mem_used_percent: 87.80992, message: 'Memory usage threshold exceeded', pid: 19608, stacktrace: 'github.com/e2b-dev/infra/packages/orchestrator/internal/sandbox.(*Sandbox).LogMetrics\n\t/build/orchestrator/internal/sandbox/metrics.go:74\ngithub.com/e2b-dev/infra/packages/orchestrator/internal/sandbox.(*Sandbox).logMetricsBasedOnConfig\n\t/build/orchestrator/internal/sandbox/checks.go:38', templateID: 'rki5dems9wqfm4r03t7g' }
```

packages/orchestrator/internal/sandbox/checks.go
```diff
@@ -18,7 +18,7 @@ import (
 
 const (
 	healthCheckInterval      = 20 * time.Second
-	metricsCheckInterval     = 10 * time.Second
+	metricsCheckInterval     = 1 * time.Second
```
jakubno: I would do at least 5 seconds

0div: this is a remnant of when i was testing, wasn't meant to be commited, but ok, ill update to 5s

jakubno (approved):

PR #612, opened 2025-05-05
Merge template manager with the orchestrator

dobrac:
Merge template manager with the orchestrator in the codebase

Done steps:
- Move files from the template-manager to the orchestrator/internal/template
- Extracted grpc server outside of orchestrator to share a common one with the template-manager
- Added cleanup of network slots
- Change the build cluster script to a parity with the orchestrators (disk, nbd init, etc)
- Ability to define env var `ORCHESTRATOR_SERVICES`, currently used only for the ServiceName

Todo:
To merge this PR, required check for "unit-tests / Run tests for packages/template-manager" has to be removed


packages/cluster/build-cluster/main.tf
```diff
@@ -94,6 +94,14 @@ resource "google_compute_instance_template" "build" {
     disk_type    = var.root_volume_disk_type
   }
 
+  disk {
+    auto_delete  = true
+    boot         = false
+    type         = "PERSISTENT"
+    disk_size_gb = 500
```
dobrac: do we need 500 Gb disk for the template manager right now? (I'm not using the orchestrator parts yet, but will soon)

packages/nomad/template-manager.hcl
```diff
@@ -38,6 +38,8 @@ job "template-manager" {
 %{ if update_stanza }
       # https://developer.hashicorp.com/nomad/docs/configuration/client#max_kill_timeout
       kill_timeout      = "20m"
+%{ else }
+      kill_timeout      = "1m"
```
dobrac, line 42: how long should the timeout be when not awaiting close, but still want to close nbd/network slots?

packages/cluster/build-cluster/main.tf
jakubno: I would do 200 for now

packages/orchestrator/upload-orchestrator.sh
jakubno, line 1: i think this file should stay executable

packages/orchestrator/upload-envs.sh
jakubno, line 1: i think this file should stay executable

packages/orchestrator/main.go
```diff
@@ -6,114 +6,200 @@ import (
...
 
-func run() int32 {
+func main() {
+	port := flag.Uint("port", defaultPort, "orchestrator server port")
+	proxyPort := flag.Uint("proxy-port", defaultProxyPort, "orchestrator proxy port")
+	flag.Parse()
+	if *port > math.MaxUint16 {
+		log.Fatalf("%d is larger than maximum possible port %d", port, math.MaxInt16)
+		os.Exit(1)
+	}
```
jakubno, line 56: shouldn't you do the same for proxy port?

packages/orchestrator/main.go
```diff
@@ -6,114 +6,200 @@ import (
...
+
+	os.Exit(run(*port, *proxyPort))
+}
+
+func run(port, proxyPort uint) (result int) {
+	result = 0
```
jakubno: this isn't the same, we had a way to count how many errors happened, now we only know there was some error

dobrac: I don't think the count holds any additional value, you should already see all the error logs and status code 1

I'll change the result to boolean instead

packages/nomad/template-manager.hcl
jakubno: Minute should be fine, how many NBDs will be in use 1, maybe 2? If there will be errors, we can increase

dobrac: yes, I think so, ok 👍

jakubno (approved):

PR #614, opened 2025-05-05
Check the database version before starting API

jakubno:
# Description

Check if the latest expected migration is applied, fail if not

### DIFF
--- a/packages/api/Makefile
+++ b/packages/api/Makefile
@@ -5,7 +5,7 @@ IMAGE := e2b-orchestration/api
 
 openapi := ../../spec/openapi.yml
 codegen := go run github.com/oapi-codegen/oapi-codegen/v2/cmd/oapi-codegen@latest
-expectedMigration := $(shell git ls-tree --name-only HEAD ../db/migrations/* | sed 's|../db/migrations/||' | sed 's/_.*//' | sort | tail -n 1)
+expectedMigration := $(shell ./../../scripts/get-latest-migration.sh)
 
 .PHONY: generate
 generate:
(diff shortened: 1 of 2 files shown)


packages/api/Makefile
```diff
@@ -17,11 +17,14 @@ generate:
 build:
 	# Allow for passing commit sha directly for docker builds
 	$(eval COMMIT_SHA ?= $(shell git rev-parse --short HEAD))
-	CGO_ENABLED=0 go build -o bin/api -ldflags "-X=main.commitSHA=$(COMMIT_SHA)" .
+	$(eval EXPECTED_MIGRATION_TIMESTAMP ?= $(shell git ls-tree --name-only HEAD packages/db/migrations/* | sed 's|packages/db/migrations/||' | sed 's/_.*//' | sort | tail -n 1))
```
dobrac: is it possible to unite this with the out of order migrations CI, and maybe move to the db package?

packages/api/internal/utils/db.go
```diff
@@ -0,0 +1,42 @@
...
+	"fmt"
+	"os"
+
+	"github.com/pressly/goose/v3"
+	"go.uber.org/zap"
+)
+
+const trackingTable = "_migrations"
+
+func CheckMigrationVersion(expectedMigration int64) error {
```
dobrac, line 14: can we use time.Time instead?

dobrac (changes requested):

packages/api/internal/utils/db.go
jakubno: Goose is returning int, so you would need to convert both.
Also you could theoretically use version only for order not the actual timestamp, so int is more general

packages/api/Makefile
jakubno: It's little bit different, in CI we are getting latest version from `main` and versions of all migrations in `HEAD`

dobrac (approved):

PR #616, opened 2025-05-05
Reconfigure timeout in envd

ValentaTomas:
The lower read header timeout and too big idle timeout can cause problems with our setup.

- https://github.com/golang/go/issues/47007

### DIFF
--- a/packages/envd/main.go
+++ b/packages/envd/main.go
@@ -157,7 +157,6 @@ func main() {
 		),
 		Addr: fmt.Sprintf("0.0.0.0:%d", port),
 		// We remove the timeouts as the connection is terminated by closing of the sandbox and keepalive close.
-		// https://github.com/golang/go/issues/47007
 		ReadTimeout:  0,
 		WriteTimeout: 0,
 		IdleTimeout:  idleTimeout,
--- a/packages/envd/main.go
+++ b/packages/envd/main.go
@@ -23,9 +23,9 @@ import (
 )
 
 const (
-	// We limit the timeout more in proxies.
-	maxTimeout = 24 * time.Hour
-	maxAge     = 2 * time.Hour
+	// Greater than in orchestrator proxy.
+	idleTimeout = 640 * time.Second
+	maxAge      = 2 * time.Hour
 
 	defaultPort = 49983
 )
@@ -155,11 +155,12 @@ func main() {
 				middleware.Wrap(handler),
 			),
 		),
-		Addr:              fmt.Sprintf("0.0.0.0:%d", port),
-		ReadHeaderTimeout: 5 * time.Second,
-		ReadTimeout:       maxTimeout,
-		WriteTimeout:      maxTimeout,
-		IdleTimeout:       maxTimeout,
+		Addr: fmt.Sprintf("0.0.0.0:%d", port),
+		// We remove the timeouts as the connection is terminated by closing of the sandbox and keepalive close.
+		// https://github.com/golang/go/issues/47007
+		ReadTimeout:  0,
+		WriteTimeout: 0,
+		IdleTimeout:  idleTimeout,
 	}
 
 	if startCmdFlag != "" {


jakubno:
From the issue it seems it's relevant only if the ReadHeaderTimeout is present.

Is there a bug you are trying to fix? Do you have reproducible example?

packages/envd/main.go
```diff
@@ -155,11 +155,12 @@ func main() {
...
-		ReadHeaderTimeout: 5 * time.Second,
-		ReadTimeout:       maxTimeout,
-		WriteTimeout:      maxTimeout,
-		IdleTimeout:       maxTimeout,
+		Addr: fmt.Sprintf("0.0.0.0:%d", port),
+		// We remove the timeouts as the connection is terminated by closing of the sandbox and keepalive close.
+		// https://github.com/golang/go/issues/47007
+		ReadTimeout:  0,
+		WriteTimeout: 0,
+		IdleTimeout:  idleTimeout,
```
ValentaTomas, line 162: The previous Idle timeout was 24 hours, which should not be causing problems because it is bigger than the orchestrator proxy timeout, but it will lead to connections being active for too long. 640s is still bigger than the orchestrator proxy timeout, but reasonable so idle connections will get closed if inactive.

packages/envd/main.go
```diff
@@ -155,11 +155,12 @@ func main() {
...
 		),
-		Addr:              fmt.Sprintf("0.0.0.0:%d", port),
-		ReadHeaderTimeout: 5 * time.Second,
-		ReadTimeout:       maxTimeout,
-		WriteTimeout:      maxTimeout,
-		IdleTimeout:       maxTimeout,
+		Addr: fmt.Sprintf("0.0.0.0:%d", port),
+		// We remove the timeouts as the connection is terminated by closing of the sandbox and keepalive close.
+		// https://github.com/golang/go/issues/47007
+		ReadTimeout:  0,
```
ValentaTomas, line 160: There was no point in having a fixed 24 hour timeout here as the sandbox will get closed after 24 hours.
Because of the streams the timeout would also be ineffective/limiting and could lead to error that are hard to explain if this timeout closes connection.

ValentaTomas: At the same time this still needs to be at least 24 hours so we don't accidentally interrupt streams.

packages/envd/main.go
```diff
@@ -155,11 +155,12 @@ func main() {
 				middleware.Wrap(handler),
 			),
 		),
-		Addr:              fmt.Sprintf("0.0.0.0:%d", port),
-		ReadHeaderTimeout: 5 * time.Second,
```
ValentaTomas, line 159: The error mentioned in the issue could be triggered by this as it is lower than the idle timeout.

jakubno: But only if idle timeout isn't specified which weren't our case

ValentaTomas: I think it was specified by the fact we defined the readTimeout, because it defaults to that if it is not specified.

ValentaTomas: Also, the case where we just define read+write timeout and they should [include the header already](https://mariocarrion.com/img/2021/09/17/0.png) might be just simpler overall, and consistent with all the other places where we set timeouts.

ValentaTomas: Linking https://github.com/e2b-dev/infra/pull/617/files#r2078354536

jakubno: I think the linked issue, isn't relevant. But agree with the new values make more sense

jakubno (approved):

packages/envd/main.go
```diff
@@ -155,11 +155,12 @@ func main() {
...
-		ReadTimeout:       maxTimeout,
-		WriteTimeout:      maxTimeout,
-		IdleTimeout:       maxTimeout,
+		Addr: fmt.Sprintf("0.0.0.0:%d", port),
+		// We remove the timeouts as the connection is terminated by closing of the sandbox and keepalive close.
+		// https://github.com/golang/go/issues/47007
```
jakubno: It isn't relevant and specifying only ReadHeaderTimeout is kinda weird anyway
```suggestion
```

PR #617, opened 2025-05-05
Reconfigure API timeouts

ValentaTomas:
The read header timeout lower than the backend timeout and the idle timeout lower than the LB idle timeout can be causing connection problems.

- https://github.com/golang/go/issues/47007
- https://cloud.google.com/load-balancing/docs/https/troubleshooting-ext-https-lbs#troubleshoot_general_connectivity_issues

### DIFF
--- a/packages/api/main.go
+++ b/packages/api/main.go
@@ -45,9 +45,9 @@ const (
 	maxMultipartMemory = 1 << 27 // 128 MiB
 	maxUploadLimit     = 1 << 28 // 256 MiB
 
-	maxReadHeaderTimeout = 60 * time.Second
-	maxReadTimeout       = 75 * time.Second
-	maxWriteTimeout      = 75 * time.Second
+	maxReadTimeout  = 75 * time.Second
+	maxWriteTimeout = 75 * time.Second
+	idleTimeout     = 620 * time.Second
 
 	defaultPort = 80
 )
@@ -171,12 +171,14 @@ func NewGinServer(ctx context.Context, logger *zap.Logger, apiStore *handlers.AP
 	r.MaxMultipartMemory = maxMultipartMemory
 
 	s := &http.Server{
-		Handler:           r,
-		Addr:              fmt.Sprintf("0.0.0.0:%d", port),
-		ReadHeaderTimeout: maxReadHeaderTimeout,
-		ReadTimeout:       maxReadTimeout,
-		WriteTimeout:      maxWriteTimeout,
-		BaseContext:       func(net.Listener) context.Context { return ctx },
+		Handler: r,
+		Addr:    fmt.Sprintf("0.0.0.0:%d", port),
+		// Configure timeouts to be greater than the proxy timeouts.
+		// https://github.com/golang/go/issues/47007
+		ReadTimeout:  maxReadTimeout,
+		WriteTimeout: maxWriteTimeout,
+		IdleTimeout:  idleTimeout,
+		BaseContext:  func(net.Listener) context.Context { return ctx },
 	}
 
 	return s
--- a/packages/api/main.go
+++ b/packages/api/main.go
@@ -176,7 +176,6 @@ func NewGinServer(ctx context.Context, logger *zap.Logger, apiStore *handlers.AP
 		Handler: r,
 		Addr:    fmt.Sprintf("0.0.0.0:%d", port),
 		// Configure timeouts to be greater than the proxy timeouts.
-		// https://github.com/golang/go/issues/47007
 		ReadTimeout:  maxReadTimeout,
 		WriteTimeout: maxWriteTimeout,
 		IdleTimeout:  idleTimeout,
--- a/packages/api/main.go
+++ b/packages/api/main.go
@@ -47,7 +47,9 @@ const (
 
 	maxReadTimeout  = 75 * time.Second
 	maxWriteTimeout = 75 * time.Second
-	idleTimeout     = 620 * time.Second
+	// This timeout should be > 600 (GCP LB upstream idle timeout) to prevent race condition
+	// https://cloud.google.com/load-balancing/docs/https#timeouts_and_retries%23:~:text=The%20load%20balancer%27s%20backend%20keepalive,is%20greater%20than%20600%20seconds
+	idleTimeout = 620 * time.Second
 
 	defaultPort = 80
 )
--- a/packages/cluster/network/main.tf
+++ b/packages/cluster/network/main.tf
@@ -310,8 +310,9 @@ resource "google_compute_url_map" "orch_map" {
 
 ### IPv4 block ###
 resource "google_compute_target_https_proxy" "default" {
-  name    = "${var.prefix}https-proxy"
-  url_map = google_compute_url_map.orch_map.self_link
+  name                        = "${var.prefix}https-proxy"
+  url_map                     = google_compute_url_map.orch_map.self_link
+  http_keep_alive_timeout_sec = 540
 
   certificate_map = "//certificatemanager.googleapis.com/${google_certificate_manager_certificate_map.certificate_map.id}"
 }
--- a/packages/cluster/network/main.tf
+++ b/packages/cluster/network/main.tf
@@ -310,9 +310,8 @@ resource "google_compute_url_map" "orch_map" {
 
 ### IPv4 block ###
 resource "google_compute_target_https_proxy" "default" {
-  name                        = "${var.prefix}https-proxy"
-  url_map                     = google_compute_url_map.orch_map.self_link
-  http_keep_alive_timeout_sec = 540
+  name    = "${var.prefix}https-proxy"
+  url_map = google_compute_url_map.orch_map.self_link
 
   certificate_map = "//certificatemanager.googleapis.com/${google_certificate_manager_certificate_map.certificate_map.id}"
 }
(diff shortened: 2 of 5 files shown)


packages/cluster/network/main.tf
```diff
@@ -310,8 +310,9 @@ resource "google_compute_url_map" "orch_map" {
 
 ### IPv4 block ###
 resource "google_compute_target_https_proxy" "default" {
-  name    = "${var.prefix}https-proxy"
-  url_map = google_compute_url_map.orch_map.self_link
+  name                        = "${var.prefix}https-proxy"
+  url_map                     = google_compute_url_map.orch_map.self_link
+  http_keep_alive_timeout_sec = 540
```
jakubno: do we need to change this? can this be changed without any disruption?

packages/api/main.go
```diff
@@ -45,9 +45,9 @@ const (
 	maxMultipartMemory = 1 << 27 // 128 MiB
 	maxUploadLimit     = 1 << 28 // 256 MiB
 
-	maxReadHeaderTimeout = 60 * time.Second
-	maxReadTimeout       = 75 * time.Second
-	maxWriteTimeout      = 75 * time.Second
+	maxReadTimeout  = 75 * time.Second
+	maxWriteTimeout = 75 * time.Second
+	idleTimeout     = 620 * time.Second
```
ValentaTomas: Because we previously did not define idle timeout it defaulted to the readTimeout—75 sec. This was lower than the idle timeout on the Load balancer (600s) so the connection could be interrupted between the LB and API server, while the client still thinking the connection was there.

packages/cluster/network/main.tf
ValentaTomas: Based on [this](https://cloud.google.com/load-balancing/docs/https#global-classic-connections) the LB client facing part has a default timeout of 610s (that can be configured), but the backend has reconfigurable timeout of 600s. I don't know why these are the defaults, because of the issue this can cause:
The reason to set the timeout this way (rising going upstream) are [here](https://cloud.google.com/load-balancing/docs/https/troubleshooting-ext-https-lbs#troubleshoot_general_connectivity_issues).
This applies to both API and sandbox traffic.

This could be even lower (300 might be ok too), but 540 seems reasonable.

As for the reload without disruption—it should be just part of Envoy config, so most likely yes, but I have not tested this.

packages/api/main.go
```diff
@@ -171,12 +171,14 @@ func NewGinServer(ctx context.Context, logger *zap.Logger, apiStore *handlers.AP
 	r.MaxMultipartMemory = maxMultipartMemory
 
 	s := &http.Server{
-		Handler:           r,
-		Addr:              fmt.Sprintf("0.0.0.0:%d", port),
-		ReadHeaderTimeout: maxReadHeaderTimeout,
```
ValentaTomas, line 176: Related issues
- https://github.com/golang/go/issues/35626
- https://github.com/golang/go/issues/47007

Also linking the similar case here https://github.com/e2b-dev/infra/pull/616/files#r2078148106

packages/cluster/network/main.tf
jakubno: I think there's misunderstanding, I think the default Load Balancer timeout is higher than the one to service on purpose, so it can close the connection to backend gracefully and to maximize the connection length to client, this should reduce latency

```
Time →──────────────────────────────────────────────────────────────────────────────▶
  • Client–GFE Idle Keep-Alive Timeout = 610 s
  • GFE–Backend Idle Keep-Alive Timeout = 600 s (fixed)
  • Your Server Idle Keep-Alive Timeout > 620 s

─── At 0 s ────────────────────────────────────────────────────────────────────────  
   Client opens a TCP connection to GFE  
   GFE opens a TCP connection to your server  
   (Both sides handshake, HTTP keep-alive enabled)

─── Idle, no requests for 600 s ─────────────────────────────────────────────────  
   At **600 s**, GFE notices its **backend socket** has been idle for 600 s → GFE sends a **FIN** to your server and closes that backend TCP connection.  
   Your server (with 620 s idle config) has not yet closed its side, so it does a **graceful close** when it sees GFE’s FIN.

─── Idle from 600 s to 605 s ──────────────────────────────────────────────────  
   Client ↔ GFE connection still open (it only times out at 610 s)  
   Backend connection is now closed on both ends (GFE closed it at 600 s; server followed suit)

─── At **605 s**, client issues a new HTTP request over the **existing** TCP connection to GFE  
   1. **GFE receives** the request on the client socket (still alive).  
   2. GFE looks for an **idle backend socket** to forward on—but it closed that at 600 s.  
   3. GFE immediately **opens a brand-new TCP connection** to your server (SYN → SYN-ACK → ACK).  
   4. GFE **forwards** the client’s request over that new connection.  
   5. Your server processes and replies; GFE relays the response back to the client.  

─── Connection now active again; both sides may keep it alive for their configured idle timeouts ────▶


```

packages/api/main.go
jakubno: Yea, this looks good

packages/api/main.go
jakubno: The first one isn't relevant as ReadHeaderTimeout < ReadTimeout and not ReadHeaderTimeout > ReadTimeout

jakubno: The second also isn't relevant for us, as there was `ReadTimeout` set

ValentaTomas: You are right, I mislooked, still would replace the values to be saner.

packages/cluster/network/main.tf
ValentaTomas: You might be right, the fact that they did set it up this way by default is at least suspicious.

ValentaTomas: Let's scrap this for now.

ValentaTomas:
I also wanted to mention a few relevant things to the setup
- Previously, in the code interpreter, the idle timeout was set to a small default value: https://github.com/e2b-dev/code-interpreter/pull/102
- We are also missing the manual "ping" in the code interpreter that in [envd](https://github.com/e2b-dev/infra/blob/main/packages/envd/internal/services/process/start.go#L142) keeps the connection open
- I think the default behavior of nodejs fetch changed and in the latest version it might be using keepalive—we should ensure it is configured the same in all our our SDKs (API+envd client, Python+JS)

packages/cluster/network/main.tf
ValentaTomas: Thinking, if we should configure our proxies to also have (620s-610s, 630s-620s) timeouts.

ValentaTomas: I'm really thinking if the LB's 610s-600s (fixed) have any advantage, because the client's (-> LB) timeout should already be <610s, so the connection would still be closed from the client side at that point.
I'm just thinking if there is a point of making the "client" part of each proxy timeout lower than the "server"—wouldn't it work equally well if the "client" part timeout was slightly higher (or the same, doesn't matter in the end) too if the next server timeout was still bigger?

jakubno: There are two idle timeouts, downstream and upstream and the rule for (downstream > upstream) is mainly for the same service.

If you close the downstream connection first, the new downstream connection can try to reuse the TCP connection even though it's already in `CLOSE_WAIT` state, because transport doesn't check the connection state.

Doing this results in error.

2) For different services you are trying to prevent both services sending `FIN` at the same time, it's a good practice to add a cushion towards upstream, so you don't have to recreate these connections. Theoretically you could have it the other way around

So the important idle timeout is between server and client in the same service, I committed it with a comment

jakubno, line 315:
```suggestion
```

packages/shared/pkg/proxy/proxy.go
```diff
@@ -33,10 +34,13 @@ func New(
...
-			Addr:              fmt.Sprintf(":%d", port),
-			ReadTimeout:       0,
-			WriteTimeout:      0,
-			IdleTimeout:       idleTimeout,
+			Addr:         fmt.Sprintf(":%d", port),
+			ReadTimeout:  0,
+			WriteTimeout: 0,
+			// Downstream idle > upstream idle,
+			// otherwise a new downstream connection can try to reuse an open upstream connection
+			// and reuse TCP connection which is already closed
```
jakubno:
```suggestion
			// otherwise a new downstream connection can try to reuse an upstream connection
			// which is already in state `CLOSE_WAIT` resulting in error
```

packages/api/main.go
```diff
@@ -171,12 +171,14 @@ func NewGinServer(ctx context.Context, logger *zap.Logger, apiStore *handlers.AP
...
-		ReadTimeout:       maxReadTimeout,
-		WriteTimeout:      maxWriteTimeout,
-		BaseContext:       func(net.Listener) context.Context { return ctx },
+		Handler: r,
+		Addr:    fmt.Sprintf("0.0.0.0:%d", port),
+		// Configure timeouts to be greater than the proxy timeouts.
+		// https://github.com/golang/go/issues/47007
```
jakubno: Not relevant
```suggestion
```

packages/cluster/network/main.tf
jakubno: Return the previous values, we don't want to change this
```suggestion
  name    = "${var.prefix}https-proxy"
  url_map = google_compute_url_map.orch_map.self_link
```

ValentaTomas:
I lowered the timeout in proxies to take into account the +10s so the timeouts of the orchestrator proxy server is not the same as the changed timeout of the envd server (https://github.com/e2b-dev/infra/pull/616/files).

PR #623, opened 2025-05-07
Abstract sandbox creation for template manager

dobrac:
Abstract sandbox operations (create, pause, resume) and reuse them for the template creation

This PR also includes:
- nil blocks filtering for pause (united with template creation)
- ignore logs for template-manager health checks
- accessible network for start commands

Notes:
- There is a possible improvement of running mappings normalization only when empty mappings are not empty for this pause snapshot
- I'm not 100% sure when the release lock on pause can be done


packages/orchestrator/internal/sandbox/sandbox_linux.go
```diff
@@ -354,223 +445,380 @@ func (s *Sandbox) Stop(ctx context.Context) error {
...
+		memfile.Path(),
+	)
 	if err != nil {
-		return nil, fmt.Errorf("error pausing vm: %w", err)
+		return nil, fmt.Errorf("error creating snapshot: %w", err)
 	}
 
-	err = s.uffd.Disable()
+	// TODO: When is the right time to release the lock?
+	releaseLock()
```
dobrac: I'm not 100% sure when the release lock on pause can be done

packages/orchestrator/internal/sandbox/fc/client_linux.go
```diff
@@ -143,3 +144,124 @@ func (c *apiClient) setMmds(ctx context.Context, metadata *MmdsMetadata) error {
...
+			HostDevName: &tapName,
+		},
+	}
+
+	_, err := c.client.Operations.PutGuestNetworkInterfaceByID(&networkConfig)
+	if err != nil {
+		return fmt.Errorf("error setting fc network config: %w", err)
+	}
+
+	// TODO: does this have to be done after machine config?
```
dobrac: it seems to be working this way based on my testing

packages/orchestrator/internal/sandbox/fc/process.go
```diff
@@ -242,6 +237,127 @@ func (p *Process) Start(
...
+		fcStopErr := p.Stop()
+
+		return errors.Join(fmt.Errorf("error setting fc network config: %w", err), fcStopErr)
+	}
+	telemetry.ReportEvent(childCtx, "set fc network config")
+
+	// Machine Config
+	// hack for 16GB RAM templates
+	// TODO: Remove this in place of Tiers
+	// customer template id raocbwn4f2mtdrjuajsx
```
dobrac: this should be removed after the customer reconfigures their settings

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -54,7 +53,6 @@ func (s *server) Create(ctxConn context.Context, req *orchestrator.SandboxCreate
 		childSpan.SpanContext().TraceID().String(),
 		req.StartTime.AsTime(),
 		req.EndTime.AsTime(),
-		req.Sandbox.Snapshot,
```
jakubno, line 57: This information isn't used anymore, we could probably drop it completely

packages/orchestrator/internal/sandbox/sandbox_linux.go
dobrac: @ValentaTomas

packages/orchestrator/internal/template/build/rootfs.go
```diff
@@ -181,7 +191,8 @@ func (r *Rootfs) createRootfsFile(ctx context.Context, tracer trace.Tracer, post
 		FcAddress   string
 		MemoryLimit int
 	}{
-		FcAddress:   fcAddr,
+		// TODO: use slot instead of a hardcoded value
```
jakubno: Here's one TODO

packages/orchestrator/internal/sandbox/fc/client_linux.go
```diff
@@ -143,3 +144,124 @@ func (c *apiClient) setMmds(ctx context.Context, metadata *MmdsMetadata) error {
...
+
+	_, err := c.client.Operations.PutGuestDriveByID(&driversConfig)
+	if err != nil {
+		return fmt.Errorf("error setting fc drivers config: %w", err)
+	}
+
+	return nil
+}
+
+func (c *apiClient) setNetworkInterface(ctx context.Context, ifaceID string, tapName string, tapMac string) error {
```
jakubno, line 189: isn't this setting up MMDS?

packages/orchestrator/internal/sandbox/fc/client_linux.go
jakubno: did you want to do it in parallel? If you found out it has to be done after, could you add comment stating it 🙏🏻

packages/orchestrator/internal/sandbox/fc/process.go
```diff
@@ -220,10 +214,11 @@ func (p *Process) Start(
 				}
 			}
 
-			errMsg := fmt.Errorf("error waiting for fc process: %w", waitErr)
+			zap.L().Error("error waiting for fc process", zap.Error(waitErr))
 
+			errMsg := fmt.Errorf("error waiting for fc process: %w", waitErr)
 			p.Exit <- errMsg
-
+			// TODO: Handle the error properly
```
jakubno: this is for the future?

packages/orchestrator/internal/sandbox/fc/process.go
jakubno: done

packages/orchestrator/internal/sandbox/fc/process.go
```diff
@@ -242,6 +237,127 @@ func (p *Process) Start(
...
+		return fmt.Errorf("error symlinking rootfs: %w", err)
+	}
+
+	err = p.client.setRootfsDrive(childCtx, p.buildRootfsPath)
+	if err != nil {
+		fcStopErr := p.Stop()
+
+		return errors.Join(fmt.Errorf("error setting fc drivers config: %w", err), fcStopErr)
+	}
+	telemetry.ReportEvent(childCtx, "set fc drivers config")
```
jakubno: could this also be moved to separate method and reused?

packages/orchestrator/internal/sandbox/nbd/pool.go
```diff
@@ -288,20 +302,19 @@ func GetDevicePath(slot DeviceSlot) DevicePath {
...
 func (d *DevicePool) Close(_ context.Context) error {
-	d.mu.Lock()
-	defer d.mu.Unlock()
+	zap.L().Info("Closing device pool", zap.Uint("used_slots", d.usedSlots.Count()))
 
-	if d.slots != nil {
-		close(d.slots)
-	}
+	d.exit <- nil
+	close(d.exit)
```
jakubno: I think we can just close

packages/orchestrator/internal/sandbox/sandbox_linux.go
```diff
@@ -187,145 +302,120 @@ func NewSandbox(
...
 		sandboxFiles,
+		rootfsOverlay,
+		baseTemplateID,
+		readonlyRootfs.Header().Metadata.BaseBuildId.String(),
+	)
+	if fcErr != nil {
+		return nil, cleanup, fmt.Errorf("failed to create FC: %w", fcErr)
+	}
+
+	// todo: check if kernel, firecracker, and envd versions exist
```
jakubno, line 354: would be nice, but we can keep it for separate PR

packages/orchestrator/internal/sandbox/sandbox_linux.go
```diff
@@ -187,145 +302,120 @@ func NewSandbox(
...
-		if uffdStopErr != nil {
-			errs = append(errs, fmt.Errorf("failed to stop uffd: %w", uffdStopErr))
-		}
+		ClickhouseStore: clickhouseStore,
+	}
 
-		healthcheckCtx.Lock()
-		healthcheckCtx.Cancel()
-		healthcheckCtx.Unlock()
+	// TODO: Part of the sandbox as we need to stop Checks before pausing the sandbox (why?)
```
jakubno: maybe there was an issue the sandbox wasn't responding anymore causing sandbox to be incorrectly reported as unhealthy?

packages/orchestrator/internal/sandbox/sandbox_linux.go
```diff
@@ -354,223 +445,380 @@ func (s *Sandbox) Stop(ctx context.Context) error {
 	return nil
 }
 
-func (s *Sandbox) Snapshot(
+// Close cleans up the sandbox and stops all resources.
+func (s *Sandbox) Close(ctx context.Context, tracer trace.Tracer) error {
+	_, span := tracer.Start(ctx, "fc-uffd-stop")
+	defer span.End()
```
jakubno: shouldn't this be something like 
```suggestion
	_, span := tracer.Start(ctx, "sandbox-close")
	defer span.End()
```

packages/orchestrator/internal/sandbox/sandbox_linux.go
jakubno: I though it was to limit number of concurrently uploading sandboxes, but that can't be true based on this, maybe taking snapshot is resource intensive, but @ValentaTomas will have to answer here

packages/orchestrator/internal/sandbox/template/template.go
```diff
@@ -3,13 +3,14 @@ package template
...
 
 type Template interface {
 	Files() *storage.TemplateCacheFiles
-	Memfile() (*Storage, error)
-	Rootfs() (*Storage, error)
+	Memfile() (block.ReadonlyDevice, error)
+	Rootfs() (block.ReadonlyDevice, error)
 	Snapfile() (File, error)
 	Close() error
 }
```
jakubno, line 14: pls fix `closeTemplate` so it returns if `t.Snapfile()` if error, the error correctly

packages/orchestrator/internal/template/build/rootfs.go
```diff
@@ -48,7 +50,7 @@ type Rootfs struct {
 	client       *client.Client
 	legacyClient *docker.Client
 
-	env *Env
+	env *TemplateConfig
```
jakubno: can we rename it from env?

packages/orchestrator/internal/template/build/template_builder.go
```diff
@@ -0,0 +1,266 @@
...
+		dockerClient:       dockerClient,
+		legacyDockerClient: legacyDockerClient,
+		templateStorage:    templateStorage,
+		storage:            storage,
+		devicePool:         devicePool,
+		networkPool:        networkPool,
+	}
+}
+
+func (b *TemplateBuilder) Build(ctx context.Context, env *TemplateConfig, envID string, buildID string) error {
```
jakubno: also here

packages/orchestrator/internal/template/build/template_config.go
```diff
@@ -0,0 +1,73 @@
...
+		FirecrackerVersion: e.FirecrackerVersion,
+		HugePages:          e.HugePages,
+		SandboxId:          uuid.New().String(),
+
+		EnvdVersion: envdVersion,
+		Vcpu:        e.VCpuCount,
+		RamMb:       e.MemoryMB,
+
+		// TODO: set proper teamID
+		TeamId: uuid.New().String(),
```
jakubno: we just need it from api right?

packages/orchestrator/internal/sandbox/sandbox_linux.go
dobrac: The scenario should be:

1. Pause FC via API
2. Snapshot FC via API—memory dump to “file on disk” that is actually tmpfs, because othewise it was too slow
3. Create the diff, and at the same time copy the diff pages from tmpfs to normal disk file
4. Delete the tmpfs file
5. Unlock so another snapshot can use tmpfs space

I'll update it appropriately

packages/orchestrator/internal/server/sandboxes.go
dobrac: let's do this in another PR, it is defined only in the proto file

packages/orchestrator/internal/template/build/rootfs.go
dobrac: This will be removed in the future PR with removing docker container (and updating provision script as it's not used anyway)

packages/orchestrator/internal/sandbox/fc/client_linux.go
dobrac: It does both, why?

packages/orchestrator/internal/sandbox/fc/process.go
dobrac: removed, not sure what it means

packages/orchestrator/internal/sandbox/sandbox_linux.go
dobrac: ok, update the comment

packages/orchestrator/internal/template/build/template_config.go
dobrac: I'm not sure why the teamId is necessary there, it's passed to MMDS, but it's probably not used for the templates right now

jakubno: I think MMDS isn't used in build at all

packages/orchestrator/internal/sandbox/fc/client_linux.go
jakubno: just naming wise, wouldn't it be better to call it `setMMDSNetworkInterface`?

packages/orchestrator/internal/sandbox/fc/process.go
jakubno: I meant we can remove this

packages/orchestrator/internal/template/build/template_config.go
dobrac: ok, I'll remove it

packages/orchestrator/internal/sandbox/fc/client_linux.go
dobrac: It does two actions:
_, err := c.client.Operations.PutGuestNetworkInterfaceByID(&networkConfig)
_, err = c.client.Operations.PutMmds(&mmdsConfig)

packages/orchestrator/internal/template/build/template_config.go
dobrac: removed

packages/shared/pkg/storage/temporary_memfile.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	releaseOnce := sync.OnceFunc(func() {
+		snapshotCacheQueue.Release(1)
+	})
+
+	randomID, err := uuid.NewRandom()
+	if err != nil {
+		releaseOnce()
+		return nil, fmt.Errorf("failed to generate identifier: %w", err)
+	}
```
jakubno:
```suggestion
	randomID, err := uuid.NewRandom()
	if err != nil {
		return nil, fmt.Errorf("failed to generate identifier: %w", err)
	}

	err = snapshotCacheQueue.Acquire(ctx, 1)
	if err != nil {
		return nil, fmt.Errorf("failed to acquire cache: %w", err)
	}

	releaseOnce := sync.OnceFunc(func() {
		snapshotCacheQueue.Release(1)
	})
```

packages/orchestrator/internal/sandbox/fc/process.go
jakubno: The same thing is in here https://github.com/e2b-dev/infra/pull/623/files/4f151c3d2821ed7a037c1245125fd85210c15c0a#diff-b30556f7a2c843402c15f698349130a89aaecafc5721d65517622d54c0d1d666R361

jakubno (approved):

PR #628, opened 2025-05-08
Remove proxy pool min connection enforcement

ValentaTomas:
Because the HTTP pipelining is disabled, there is not much reason to split pools and enforce minimum connections.
This PR also sets Docker ulimit for the client-proxy via Nomad.

### DIFF
--- a/packages/nomad/client-proxy.hcl
+++ b/packages/nomad/client-proxy.hcl
@@ -77,6 +77,10 @@ job "client-proxy" {
         image        = "${image_name}"
         ports        = ["${port_name}"]
         args         = ["--port", "${port_number}"]
+
+        ulimit {
+          nofile = "262144:262144"
+        }
       }
     }
   }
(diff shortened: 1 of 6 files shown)


packages/nomad/client-proxy.hcl
```diff
@@ -77,6 +77,10 @@ job "client-proxy" {
         image        = "${image_name}"
         ports        = ["${port_name}"]
         args         = ["--port", "${port_number}"]
+
+        ulimit {
+          nofile = "262144:262144"
```
jakubno: do you want to limit it? Now it is `1048576`

ValentaTomas: It is propagated to the docker by default?

jakubno: If no limits are set, it should use the system settings

jakubno (approved):

PR #630, opened 2025-05-08
Improve comments for idle timeouts

jakubno:
# Description

Clarifies timeouts

### DIFF
--- a/packages/envd/main.go
+++ b/packages/envd/main.go
@@ -23,7 +23,7 @@ import (
 )
 
 const (
-	// Greater than in orchestrator proxy.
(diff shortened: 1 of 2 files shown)


ValentaTomas:
Can this be merged with the current deployment? Won't we break the downstream < upstream that is currently deployed if we deploy this?

jakubno:
Good point, we can either deploy client proxy > orchestrator > envd (I can split in different PRs) or I can just fix the comments, so it's obvious what's and why is going on

ValentaTomas (approved):

PR #632, opened 2025-05-09
Use separate status for snapshots

jakubno:
# Description

There is a race condition with template manager checking old builds, which can mark the snapshot as failed

### DIFF
--- a/packages/shared/pkg/models/migrate/schema.go
+++ b/packages/shared/pkg/models/migrate/schema.go
@@ -91,7 +91,7 @@ var (
 		{Name: "created_at", Type: field.TypeTime, Default: "CURRENT_TIMESTAMP"},
 		{Name: "updated_at", Type: field.TypeTime},
 		{Name: "finished_at", Type: field.TypeTime, Nullable: true},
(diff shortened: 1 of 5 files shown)


dobrac (approved):
One more reason to split them somehow differently

packages/shared/pkg/models/migrate/schema.go
```diff
@@ -91,7 +91,7 @@ var (
 		{Name: "created_at", Type: field.TypeTime, Default: "CURRENT_TIMESTAMP"},
 		{Name: "updated_at", Type: field.TypeTime},
 		{Name: "finished_at", Type: field.TypeTime, Nullable: true},
-		{Name: "status", Type: field.TypeEnum, Enums: []string{"waiting", "building", "failed", "success", "uploaded"}, Default: "waiting", SchemaType: map[string]string{"postgres": "text"}},
+		{Name: "status", Type: field.TypeEnum, Enums: []string{"waiting", "building", "snapshotting", "failed", "success", "uploaded"}, Default: "waiting", SchemaType: map[string]string{"postgres": "text"}},
```
sitole, line 94: Maybe we should switch that alter to Postgres enum to make it little bit more strong typed?

sitole (approved):

sitole:
> One more reason to split snapshots from builds somehow differently

Maybe separated flag such as `is_snapshot` will be enough.

PR #633, opened 2025-05-09
Better error for blocked teams

jakubno:
# Description

Return better error when team is blocked or banned.

```
{
  "code": 403,
  "message": "Forbidden: team was blocked, reason: didn't use e2b enough"
}
```

instead of

```
{
  "code": 401,
  "message": "Invalid API key, please visit https://e2b.dev/docs/quickstart/api-key for more information."
}
```

### DIFF
--- a/packages/api/internal/auth/middleware.go
+++ b/packages/api/internal/auth/middleware.go
@@ -93,7 +93,7 @@ func (a *commonAuthenticator[T]) Authenticate(ctx context.Context, input *openap
 
 		var forbiddenError *db.TeamForbiddenError
 		if errors.As(validationError.Err, &forbiddenError) {
-			return validationError.Err
+			return fmt.Errorf("Forbidden: %w", validationError.Err)
 		}
 
 		return fmt.Errorf("%s\n%s (%w)", a.errorMessage, validationError.ClientMsg, validationError.Err)
--- a/packages/api/internal/auth/middleware.go
+++ b/packages/api/internal/auth/middleware.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"log"
 	"net/http"
 	"os"
 	"strings"
@@ -14,9 +13,11 @@ import (
 	"github.com/google/uuid"
 	middleware "github.com/oapi-codegen/gin-middleware"
 	"go.opentelemetry.io/otel/trace"
+	"go.uber.org/zap"
 
 	"github.com/e2b-dev/infra/packages/api/internal/api"
 	authcache "github.com/e2b-dev/infra/packages/api/internal/cache/auth"
+	"github.com/e2b-dev/infra/packages/shared/pkg/db"
 	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
 )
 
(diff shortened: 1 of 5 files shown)


packages/api/internal/auth/middleware.go
```diff
@@ -74,11 +75,6 @@ func (a *commonAuthenticator[T]) getHeaderKeysFromRequest(req *http.Request) (st
 
 // Authenticate uses the specified validator to ensure an API key is valid.
 func (a *commonAuthenticator[T]) Authenticate(ctx context.Context, input *openapi3filter.AuthenticationInput) error {
-	// Our security scheme is named ApiKeyAuth, ensure this is the case
-	if input.SecuritySchemeName != a.securitySchemeName {
```
dobrac, line 78: was this completely unnecessary? Why was it there in a first place?

packages/api/internal/cache/auth/cache.go
```diff
@@ -53,7 +52,7 @@ func (c *TeamAuthCache) GetOrSet(ctx context.Context, key string, dataCallback D
 	if item == nil {
 		team, tier, err = dataCallback(ctx, key)
 		if err != nil {
-			return nil, nil, fmt.Errorf("failed to get the team from db for a key: %w", err)
```
dobrac, line 56: is there a reason for direct err and not named?

packages/api/internal/utils/error.go
```diff
@@ -41,6 +45,19 @@ func ErrorHandler(c *gin.Context, message string, statusCode int) {
 
 	c.Error(errMsg)
 
+	// Handle forbidden errors
+	if strings.HasPrefix(message, forbiddenErrPrefix) {
```
dobrac, line 49: can these check be done a bit more universally so we can distinguish more errors in the future?

dobrac (changes requested):
Is it possible to say also the reason why the team is blocked? I'm not sure how intuitive it is to see just `Forbidden: team was blocked` when you hit the billing limit and it's not a mistake on our side

packages/api/internal/auth/middleware.go
jakubno: Yes, it was unnecessary, based on the same condition we decide, if the Authenticate function is called

jakubno:
> Is it possible to say also the reason why the team is blocked? I'm not sure how intuitive it is to see just Forbidden: team was blocked when you hit the billing limit and it's not a mistake on our side

If there's reason it will be there as well, improved the example

packages/api/internal/utils/error.go
jakubno: I don;t know, we could set the error on `Context`?

This is what's called from the library
`ErrorHandler(c, err.Error(), http.StatusBadRequest)`

You have error string and hardcoded `400`

packages/api/internal/cache/auth/cache.go
jakubno: I didn't like this format:
```
  "message": "Forbidden: failed to get the team from db for a key: team was blocked, reason: didn't use e2b enough"

```

jakubno: I change it to 
```
  "message": "Forbidden: error while getting the team: team was blocked, reason: didn't use e2b enough"
```

packages/api/internal/utils/error.go
dobrac: I've meant code wise, otherwise ok for these two cases

dobrac (approved):

PR #634, opened 2025-05-10
Disable `SetXForwarded` to prevent servers returning unexpected `content-location`

ValentaTomas:
It seems that the `SetXForwarded` setting of `X-Forwarded-Proto` header for `http` might be affecting user functionality—some services like Expo are returning unexpected `content-location` starting with `http://` then.

We need to check if the `X-Forwarded-Proto` really does this and if so deploy this fix in client-proxy and orchestrators.

### DIFF
--- a/packages/shared/pkg/proxy/pool/client.go
+++ b/packages/shared/pkg/proxy/pool/client.go
@@ -69,7 +69,7 @@ func newProxyClient(
 				}
 
 				r.SetURL(t.Url)
-				r.SetXForwarded()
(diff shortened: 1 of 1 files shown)


mlejva (approved):

jakubno:
We can probably set this up manually, but from the `SetXForwarded` implementation, it seems this is indeed the case:

```go
	if request.In.TLS == nil {
		request.Out.Header.Set("X-Forwarded-Proto", "http")
	} else {
		request.Out.Header.Set("X-Forwarded-Proto", "https")
	}
```

PR #635, opened 2025-05-10
Add template cache proxy to improve speed of new template starts

ValentaTomas:


packages/buckets/main.tf
```diff
@@ -114,6 +114,18 @@ resource "google_storage_bucket" "fc_template_bucket" {
...
 }
 
+resource "time_sleep" "destroy_wait_5000_seconds" {
+  depends_on       = [google_storage_bucket.fc_template_bucket]
+  destroy_duration = "5000s"
+}
+
+resource "google_storage_anywhere_cache" "cache" {
+  bucket     = google_storage_bucket.fc_template_bucket.name
+  zone       = var.gcp_zone
```
dobrac: How will the zone work, as we are moving to the regional instance groups, when the instances will be in a different zone?

packages/nomad/template-cache.hcl
```diff
@@ -0,0 +1,161 @@
...
+  node_pool = "api"
+
+  priority = 80
+
+  group "template-cache" {
+    count = 1
+
+    network {
+      port "template-cache" {
+        static = 28778
```
dobrac: can this port be a variable? (and properly propagated to other places)

packages/nomad/template-cache.hcl
```diff
@@ -0,0 +1,161 @@
...
+      }
+    
+      port "status" {
+        static = 3004
```
dobrac: can this port be a variable?

packages/nomad/template-cache.hcl
```diff
@@ -0,0 +1,161 @@
...
+    gzip_buffers 16 8k;
+    gzip_types text/plain text/css application/json application/javascript text/xml application/xml application/xml+rss text/javascript;
+
+    # Caching zone and path
+    proxy_cache_path /mycache levels=1:2 keys_zone=mycache:10m max_size=10g 
+             inactive=60m use_temp_path=off;
+    proxy_cache mycache;
+
+    server {
+        listen 28778;  # Change this if another service is on 8080
```
dobrac: can this be a variable?

packages/nomad/template-cache.hcl
```diff
@@ -0,0 +1,161 @@
...
+        client_body_buffer_size 2M;
+        client_max_body_size 2M;
+        client_header_buffer_size 1k;
+        large_client_header_buffers 2 2M;
+        client_body_timeout 60s;
+        keepalive_timeout 60;
+    }
+
+    server {
+      listen 3004;
```
dobrac: can this be a variable?

packages/nomad/template-cache.hcl
```diff
@@ -0,0 +1,161 @@
...
+        image        = "nginx:1.27.0"
+        network_mode = "host"
+        ports        = ["template-cache", "status"]
+        volumes = [
+          "local:/etc/nginx/",
+          "/var/log/template-cache:/var/log/nginx"
+        ]
+      }
+
+      // TODO: Saner resources
```
dobrac, line 63: todo

packages/shared/pkg/storage/storage.go
```diff
@@ -58,7 +58,7 @@ func GetTemplateStorageProvider(ctx context.Context) (StorageProvider, error) {
 	case AWSStorageProvider:
 		return NewAWSBucketStorageProvider(ctx, bucketName)
 	case GCPStorageProvider:
-		return NewGCPBucketStorageProvider(ctx, bucketName)
+		return NewGCPBucketStorageProvider(ctx, bucketName, WithProxy("template-cache.service.consul:28778"))
```
dobrac: Pass it as env variable please

packages/shared/pkg/storage/proxy_transport.go
```diff
@@ -0,0 +1,29 @@
...
+
+	"go.uber.org/zap"
+)
+
+type proxyTransport struct {
+	baseTransport *http.Transport
+	proxyHost     string
+}
+
+func newProxyTransport(proxyHost string) *proxyTransport {
```
dobrac, line 14: I think we should also do proper close cleanup for the `http.Transport` (CloseIdleConnections)

packages/shared/pkg/storage/storage_google.go
```diff
@@ -24,27 +29,68 @@ const (
...
+}
+
+func NewGCPBucketStorageProvider(ctx context.Context, bucketName string, opts ...GCPBucketStorageProviderOption) (*GCPBucketStorageProvider, error) {
+	var providerOptions GCPBucketStorageProviderOptions
+
+	for _, opt := range opts {
+		opt(&providerOptions)
+	}
+
+	// Create the legact GCP storage client with proper credentials
```
dobrac:
```suggestion
	// Create the legacy GCP storage client with proper credentials
```

dobrac: and, is it legacy?

packages/shared/pkg/storage/storage_google.go
```diff
@@ -24,27 +29,68 @@ const (
...
+		ht, err := htransport.NewTransport(ctx, providerOptions.transportProxy)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create GCS client: %w", err)
+		}
+
+		httpClient := &http.Client{
+			Transport: ht,
+		}
+
+		proxiedClient, err := storage.NewClient(ctx, option.WithHTTPClient(httpClient))
```
dobrac, line 70: is all of this properly cleanedup/closed?

packages/shared/pkg/storage/storage_google.go
```diff
@@ -74,6 +120,8 @@ func (g *GCPBucketStorageProvider) GetDetails() string {
 }
 
 func (g *GCPBucketStorageProvider) OpenObject(ctx context.Context, path string) (StorageObjectProvider, error) {
+	fmt.Fprintf(os.Stderr, "OpenObject: %s\n", path)
```
dobrac: forgotten debug

packages/shared/pkg/storage/storage_google.go
```diff
@@ -112,13 +164,32 @@ func (g *GCPBucketStorageObjectProvider) Size() (int64, error) {
 
 	return attrs.Size, nil
 }
-
 func (g *GCPBucketStorageObjectProvider) ReadAt(buff []byte, off int64) (n int, err error) {
+	if g.proxiedHandle != nil {
```
dobrac: nit: the if could be flipped to have smaller if body

packages/shared/pkg/storage/storage_google.go
```diff
@@ -112,13 +164,32 @@ func (g *GCPBucketStorageObjectProvider) Size() (int64, error) {
...
+			return g.readAt(buff, off, true)
+		}
+		return n, nil
+	}
+
+	return g.readAt(buff, off, false)
+}
+
+func (g *GCPBucketStorageObjectProvider) readAt(buff []byte, off int64, proxied bool) (n int, err error) {
+	zap.L().Debug("Reading at GCS", zap.String("path", g.path), zap.Int64("off", off), zap.Int("len(buff)", len(buff)), zap.Bool("proxied", proxied))
```
dobrac: is this debug necessary? Looks to be a bit spammy

packages/shared/pkg/storage/storage_google.go
```diff
@@ -144,7 +215,25 @@ func (g *GCPBucketStorageObjectProvider) ReadAt(buff []byte, off int64) (n int,
 }
 
 func (g *GCPBucketStorageObjectProvider) ReadFrom(src io.Reader) (int64, error) {
-	w := g.handle.NewWriter(g.ctx)
+	if g.proxiedHandle != nil {
```
dobrac: nit: the if can be flipped to have smaller if body

packages/shared/pkg/storage/storage_google.go
```diff
@@ -112,13 +164,32 @@ func (g *GCPBucketStorageObjectProvider) Size() (int64, error) {
...
+func (g *GCPBucketStorageObjectProvider) readAt(buff []byte, off int64, proxied bool) (n int, err error) {
+	zap.L().Debug("Reading at GCS", zap.String("path", g.path), zap.Int64("off", off), zap.Int("len(buff)", len(buff)), zap.Bool("proxied", proxied))
 	ctx, cancel := context.WithTimeout(g.ctx, googleReadTimeout)
 	defer cancel()
 
+	var reader *storage.Reader
 	// The file should not be gzip compressed
-	reader, err := g.handle.NewRangeReader(ctx, off, int64(len(buff)))
+	if proxied {
+		reader, err = g.proxiedHandle.NewRangeReader(ctx, off, int64(len(buff)))
```
dobrac, line 174: how about passing the reader directly instead of proxied bool?

packages/shared/pkg/storage/storage_google.go
```diff
@@ -144,7 +215,25 @@ func (g *GCPBucketStorageObjectProvider) ReadAt(buff []byte, off int64) (n int,
...
+		return n, nil
+	}
+
+	return g.readFrom(src, false)
+}
+
+func (g *GCPBucketStorageObjectProvider) readFrom(src io.Reader, proxied bool) (int64, error) {
+	var w *storage.Writer
+	if proxied {
+		w = g.proxiedHandle.NewWriter(g.ctx)
```
dobrac, line 220: how about passing the writer directly instead of proxied bool?

packages/nomad/template-cache.hcl
```diff
@@ -0,0 +1,161 @@
...
+        listen 28778;  # Change this if another service is on 8080
+
+        slice              20m;
+        proxy_cache_key    $host$uri$is_args$args$slice_range;
+        proxy_set_header   Range $slice_range;
+        proxy_http_version 1.1;
+        proxy_cache_valid  200 206 1h;
+
+        location / {
+            # Use mTLS?
```
dobrac: is this resolved?

dobrac (changes requested):
Could you please also measure some of the improvements?

packages/nomad/template-cache.hcl
```diff
@@ -0,0 +1,161 @@
...
+
+    task "template-cache" {
+      driver = "docker"
+
+      config {
+        image        = "nginx:1.27.0"
+        network_mode = "host"
+        ports        = ["template-cache", "status"]
+        volumes = [
+          "local:/etc/nginx/",
```
dobrac, line 58: I think it would make sense to mount additional drive to the API instance for this cache purpose

packages/shared/pkg/storage/storage_google.go
0div: Right, I forgot we were logging debug by default

packages/shared/pkg/storage/storage_google.go
0div: I don't see much of a difference in readability if that's what you meant.

packages/shared/pkg/storage/storage_google.go
0div: same comment as `readAt`

packages/buckets/main.tf
0div: Didn't know we were moving to regional instance groups, not ideal but https://cloud.google.com/compute/docs/instance-groups/regional-migs#selectingzones. Or else pull the plug all together on the _aptly_ named Anywhere Cache

dobrac: Do you mean that we would specify all zones manually? I was thinking there could be:<br>1) Either it doesn't matter it's in a different zone, it would function fine<br>2) We could deploy it to all zones (not sure how expensive it is)

0div: 1) yes it will fallback to regular cloud storage read calls if Anywhere Cache not in the same zone as client, also we 
2) that's a total of `us-west1-{a,b,c}` and per doc: 
> Read operations served from Anywhere Cache are priced lower than Class B operations served from a bucket in Standard storage.

packages/shared/pkg/storage/storage_google.go
```diff
@@ -24,27 +28,54 @@ const (
...
-func NewGCPBucketStorageProvider(ctx context.Context, bucketName string) (*GCPBucketStorageProvider, error) {
-	client, err := storage.NewClient(ctx)
+// Add possible proxy option
+type GCPBucketStorageProviderOptions struct {
+	transportProxy *proxyTransport
+}
+
+func NewGCPBucketStorageProvider(ctx context.Context, bucketName string, proxyURL string) (*GCPBucketStorageProvider, error) {
+	// Create the legacy GCP storage client with proper credentials
+	client, err := storage.NewClient(ctx, option.WithEndpoint("private.googleapis.com"))
```
sitole, line 50: Can we ensure this will always work, not just for in-VPC requests?

ValentaTomas:
We will be most likely using a different mechanism for this, combining read/write, because we need to optimize the upload too.

PR #639, opened 2025-05-12
Return snapshot-base template alias in list of the sandboxes

sitole:
For v2 sandboxes list endpoints adds alias of source template id for each paused sandbox.


packages/api/internal/handlers/sandboxes_list.go
```diff
@@ -179,35 +198,37 @@ func (a *APIStore) GetV2Sandboxes(c *gin.Context, params api.GetV2SandboxesParam
 	if nextToken != nil {
 		c.Header("X-Next-Token", *nextToken)
 	}
-	c.Header("X-Total-Items", strconv.Itoa(len(sandboxes)))
 
+	c.Header("X-Total-Items", strconv.Itoa(len(sandboxes)))
```
jakubno, line 202: Not related to this PR, but this doesn't seem as total count, could you create a Linear issue 🙏🏻

packages/api/internal/handlers/sandboxes_list.go
```diff
@@ -179,35 +198,37 @@ func (a *APIStore) GetV2Sandboxes(c *gin.Context, params api.GetV2SandboxesParam
...
-	for _, snapshot := range snapshots {
-		env := snapshot.Edges.Env
+	for _, record := range snapshots {
+		snapshot := record.Snapshot
+		build := record.EnvBuild
 
-		snapshotBuilds := env.Edges.Builds
-		if len(snapshotBuilds) == 0 {
-			continue
+		var alias *string = nil
```
jakubno:
```suggestion
		var alias *string
```

packages/db/queries/get_snapshots_with_cursor.sql.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/google/uuid"
+	"github.com/jackc/pgx/v5/pgtype"
+)
+
+const getSnapshotsWithCursor = `-- name: GetSnapshotsWithCursor :many
+SELECT COALESCE(ea.alias, '') as alias, s.created_at, s.env_id, s.sandbox_id, s.id, s.metadata, s.base_env_id, s.sandbox_started_at, s.env_secure, eb.id, eb.created_at, eb.updated_at, eb.finished_at, eb.status, eb.dockerfile, eb.start_cmd, eb.vcpu, eb.ram_mb, eb.free_disk_size_mb, eb.total_disk_size_mb, eb.kernel_version, eb.firecracker_version, eb.env_id, eb.envd_version
+FROM "public"."snapshots" s
+JOIN "public"."envs" e ON e.id = s.env_id
+LEFT JOIN LATERAL (
+    SELECT alias FROM "public"."env_aliases" ea
```
dobrac: have you considered there being multiple aliases?

packages/shared/pkg/db/snapshot.go
mishushakov, line 1: can you also remove my predicate in packages/shared/pkg/models/snapshot/snapshot_predicates.go

packages/db/queries/get_snapshots_with_cursor.sql
```diff
@@ -0,0 +1,30 @@
...
+        eb.env_id = s.env_id
+        AND eb.status = 'success'
+    ORDER BY eb.created_at DESC
+    LIMIT 1
+) eb ON TRUE
+WHERE
+    e.team_id = @team_id
+    AND s.metadata @> @metadata
+    AND (
+        (s.created_at < @cursor_time)
```
jakubno: aren't these brackets unnecessary?

jakubno (commented):
Could you add a test with 2 sandboxes and only one matching the metadata filter? 🙏🏻 
`tests/integration/internal/tests/api/sandbox_list_test.go`

packages/db/queries/get_snapshots_with_cursor.sql.go
sitole: Yes, there is currently not place where to return array of aliases or something, so we will take first one and use it.

sitole: Later we should return all of them and maybe have some marked as primary one for this cases i guess.

packages/api/internal/handlers/sandboxes_list.go
sitole: https://linear.app/e2b/issue/E2B-2269/wrong-api-response-header-with-sandboxes-count

mishushakov: This the count returned on the page, not the total count

mishushakov:
Don't forget to also run the SDK tests to make sure nothing is breaking.

sitole:
> Don't forget to also run the SDK tests to make sure nothing is breaking.

Run, passed both JS and Python (current main).

packages/db/queries/get_snapshots_with_cursor.sql.go
jakubno: Ideally we would get rid of multiple aliases, I think there's only 1 template, which has multiple aliases at the moment and is used

packages/api/internal/handlers/sandboxes_list.go
jakubno: That's the problem, It's called `Total-items`. This header is used to designate how many Total items are available, you can easily know many items are returned

tests/integration/internal/tests/api/sandbox_list_test.go
```diff
@@ -47,9 +47,13 @@ func TestSandboxList(t *testing.T) {
 func TestSandboxListWithFilter(t *testing.T) {
 	c := setup.GetAPIClient()
 
-	sbx := utils.SetupSandboxWithCleanup(t, c)
+	// standard sandbox
+	_ = utils.SetupSandboxWithCleanup(t, c)
 
-	metadataString := "sandboxType=test"
+	// sandbox with custom metadata
+	sbx := utils.SetupSandboxWithCleanupWithTimeout(t, c, 30, &api.SandboxMetadata{"favouriteColor": "blue"})
```
jakubno, line 54: is there a reason you use `WithTimeout`?

tests/integration/internal/utils/sandbox.go
```diff
@@ -13,19 +13,26 @@ import (
...
+func SetupSandboxWithCleanupWithTimeout(t *testing.T, c *api.ClientWithResponses, sbxTimeout int32, sbxMetadata *api.SandboxMetadata) *api.Sandbox {
 	t.Helper()
 
 	// t.Context() doesn't work with go vet, so we use our own context
 	ctx, cancel := context.WithCancel(context.Background())
 	t.Cleanup(cancel)
 
+	var metadata *api.SandboxMetadata
+	if sbxMetadata != nil {
+		metadata = sbxMetadata
```
jakubno: Do you think it's better like this or if we also add `sandboxType: test`

jakubno (changes requested):
Also I don't see any changes to running sandboxes

packages/db/queries/get_snapshots_with_cursor.sql
```diff
@@ -0,0 +1,30 @@
+-- name: GetSnapshotsWithCursor :many
+-- We must convert NULL alias into empty string, because sqlc will not generate string pointer to it and it will crash when NULL is received
+SELECT COALESCE(ea.alias, '') as alias, sqlc.embed(s), sqlc.embed(eb)
+FROM "public"."snapshots" s
+JOIN "public"."envs" e ON e.id = s.env_id
+LEFT JOIN LATERAL (
+    SELECT alias FROM "public"."env_aliases" ea
```
jakubno: maybe aggregate them to array and return the first in API, it would be better prepared if we decide to support multiple aliases publicly, also it maybe solves the problem with `null`?

packages/db/queries/get_snapshots_with_cursor.sql
```diff
@@ -1,12 +1,12 @@
 -- name: GetSnapshotsWithCursor :many
 -- We must convert NULL alias into empty string, because sqlc will not generate string pointer to it and it will crash when NULL is received
```
jakubno: comment not needed

tests/integration/internal/tests/api/sandbox_list_test.go
sitole: `SetupSandboxWithCleanup` just wraps `SetupSandboxWithCleanupWithTimeout`. I did not want to expose metadata param directly into `SetupSandboxWithCleanup` because it will cause nill for every use of it.

tests/integration/internal/utils/sandbox.go
sitole: Okay, added append with sandbox type even for override

packages/db/queries/get_snapshots_with_cursor.sql
sitole: Added. I aggregate them directly inside lateral because otherwise i would need to group by all fields manually

jakubno (changes requested):
Still missing the changes for running sandboxes

sitole:
> Still missing the changes for running sandboxes

Sorry, missed.

Base template id is now added into instance cache when triggered by sbx resume endpoint. Resume also fetches aliases for base template and first one is pushed into instance cache.

jakubno (approved):

PR #641, opened 2025-05-12
Return after port closed error handling in proxy

ValentaTomas:


packages/shared/pkg/proxy/handler.go
```diff
@@ -67,6 +67,7 @@ func handler(p *pool.ProxyPool, getDestination func(r *http.Request) (*pool.Dest
 				HandleError(w, r)
 			if err != nil {
 				zap.L().Error("failed to handle sandbox not found error", zap.Error(err))
+
```
jakubno:
```suggestion
```

packages/shared/pkg/proxy/pool/client.go
```diff
@@ -101,9 +101,13 @@ func newProxyClient(
 
 						return
```
jakubno, line 102: this return is kinda useless

jakubno (approved):

PR #644, opened 2025-05-14
fix hang in handle_userfault

mengqiqi-github:
Before closing uffd, you need to unregister the memory area. When the huge page memory is insufficient, closing uffd without following the standard shutdown process may cause the kernel thread to enter the D state.

Reproduction process:
Create a large number of sandboxes until the large page memory is exhausted, then wait for the sandbox to close and you will find that many D processes are generated.

image

### DIFF
--- a/packages/orchestrator/internal/sandbox/uffd/handler.go
+++ b/packages/orchestrator/internal/sandbox/uffd/handler.go
@@ -9,6 +9,7 @@ import (
 	"sync"
 	"syscall"
 	"time"
+	"unsafe"
 
 	"github.com/bits-and-blooms/bitset"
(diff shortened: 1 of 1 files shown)


ValentaTomas:
@mengqiqi-github Out of curiosity, why you closed this?

mengqiqi-github:
> @mengqiqi-github Out of curiosity, why you closed this?

Sorry for taking so long to reply. 

I found that the reason for this problem is that our colleague restarted the orchestrator service, but these firecracker processes were not closed. The orchestrator did not reconnect to continue processing the pagefault requests of these firecrackers, which caused the kernel thread to enter the D state.

PR #645, opened 2025-05-14
Orchestrator gRCP service info method

sitole:
- Orchestrator HTTP health check no longer reposts service version
- Added RPC info service that returns basic service info (version, metrics, detailed health status), and allows to call fo service status change
- API is now using new service info method for query of service version, health status etc
- API now registers only healthy orchestrators. Previously API registered all discovered one and filled info struct with dummy data.


packages/api/internal/orchestrator/cache.go
```diff
@@ -131,21 +133,44 @@ func (o *Orchestrator) syncNode(ctx context.Context, node *Node, nodes []*node.N
 		return
 	}
 
-	activeInstances, instancesErr := o.getSandboxes(ctx, node.Info)
-	if instancesErr != nil {
-		zap.L().Error("Error getting instances", zap.Error(instancesErr))
-		return
-	}
+	syncMaxRetries := 4
```
dobrac: make it const at the top of the file

packages/api/internal/orchestrator/client.go
```diff
@@ -67,25 +67,33 @@ func (o *Orchestrator) connectToNode(ctx context.Context, node *node.NodeInfo) e
...
-	} else {
+	ok, err := o.getNodeHealth(node)
+	if err != nil {
 		zap.L().Error("Failed to get node health", zap.Error(err))
+		return err
+	}
+
+	if !ok {
+		zap.L().Error("Node is not healthy", zap.String("node_id", node.ID))
+		return fmt.Errorf("node is not healthy")
```
dobrac: we should still keep track of unhealthy nodes, to be able to know about them

packages/orchestrator/internal/grpcserver/healthcheck.go
```diff
@@ -116,10 +116,7 @@ func (h *Healthcheck) healthHandler(w http.ResponseWriter, r *http.Request) {
 	h.mu.RLock()
 	defer h.mu.RUnlock()
 
-	response := e2bHealth.Response{
-		Status:  h.status,
-		Version: h.version,
-	}
+	response := e2bHealth.Response{Status: h.status}
```
dobrac: how about the version here? I think we are using the version already

packages/orchestrator/internal/server/main.go
```diff
@@ -41,8 +43,39 @@ type server struct {
...
 }
 
+type ServiceInfo struct {
+	ClientId             string
+	ServiceId            string
+	ServiceSourceVersion string
+	ServiceSourceCommit  string
+	ServiceStartup       time.Time
+
+	CanWorkAsOrchestrator    bool
```
dobrac: Do you need the explicit booleans like this or can it be an array?

packages/orchestrator/orchestrator.proto
```diff
@@ -99,3 +126,8 @@ service SandboxService {
 
   rpc ListCachedBuilds(google.protobuf.Empty) returns (SandboxListCachedBuildsResponse);
 }
+
+service InfoService {
```
dobrac: Is this orchestrator only related, or separate service reused also for the template manager?

packages/orchestrator/internal/server/main.go
```diff
@@ -137,6 +165,7 @@ func New(
 	}
 
 	orchestrator.RegisterSandboxServiceServer(grpc.GRPCServer(), srv.server)
+	orchestrator.RegisterInfoServiceServer(grpc.GRPCServer(), srv.server)
```
dobrac: Is the info service orchestrator only?

packages/orchestrator/main.go
```diff
@@ -178,7 +186,19 @@ func run(port, proxyPort uint) (success bool) {
...
 
-	_, err = server.New(ctx, grpcSrv, networkPool, devicePool, tracer, clientID, commitSHA, sandboxProxy, sandboxes)
+	serviceInfo := &server.ServiceInfo{
+		ClientId:             clientID,
+		ServiceId:            uuid.NewString(),
+		ServiceStartup:       time.Now(),
+		ServiceSourceVersion: version,
+		ServiceSourceCommit:  commitSHA,
+
+		CanWorkAsOrchestrator:    true,
```
dobrac:
```suggestion
		CanWorkAsOrchestrator:    true,
		CanWorkAsOrchestrator: slices.Contains(services, servicetype.Orchestrator),
```

packages/orchestrator/internal/grpcserver/healthcheck.go
sitole: It was used by API and its already migrated to service info.

packages/orchestrator/internal/server/main.go
sitole: added array of roles instead of booleans. Good point!

packages/api/internal/orchestrator/client.go
sitole: Added support for that now. I worried about filling dummy values when we don't know them from info request, but should be okay. Now we are filling just node service version.

packages/api/internal/orchestrator/client.go
```diff
@@ -67,25 +67,46 @@ func (o *Orchestrator) connectToNode(ctx context.Context, node *node.NodeInfo) e
...
+			node.ID, &Node{
+				Client:         client,
+				Info:           node,
+				buildCache:     buildCache,
+				status:         api.NodeStatusUnhealthy,
+				version:        "unknown",
+				sbxsInProgress: smap.New[*sbxInProgress](),
+				createFails:    atomic.Uint64{},
+			},
+		)
```
dobrac: I think you can merge the nodes.Insert to logic to one statement and just construct the Node struct based on the received data

packages/orchestrator/main.go
```diff
@@ -178,7 +187,28 @@ func run(port, proxyPort uint) (success bool) {
...
 
-	_, err = server.New(ctx, grpcSrv, networkPool, devicePool, tracer, clientID, commitSHA, sandboxProxy, sandboxes)
+	serviceRoles := make([]orchestrator.ServiceInfoRole, 0)
+	if slices.Contains(services, servicetype.Orchestrator) {
+		serviceRoles = append(serviceRoles, orchestrator.ServiceInfoRole_Orchestrator)
+	}
+
+	if slices.Contains(services, servicetype.TemplateManager) {
+		serviceRoles = append(serviceRoles, orchestrator.ServiceInfoRole_TemplateManager)
+	}
```
dobrac: creating a mapping function might be better

packages/orchestrator/main.go
```diff
@@ -178,7 +187,28 @@ func run(port, proxyPort uint) (success bool) {
...
+
+	serviceInfo := &server.ServiceInfo{
```
dobrac: can we construct this in a separate function (file)

packages/orchestrator/main.go
```diff
@@ -215,6 +245,9 @@ func run(port, proxyPort uint) (success bool) {
 		closers = append([]Closeable{tmpl}, closers...)
 	}
 
+	// Initialize info service as last to make sure template manager/orchestrator are already initialized
+	info.NewInfoService(ctx, grpcSrv, serviceInfo, sandboxes)
```
dobrac, line 226: > // Initialize info service as last to make sure template manager/orchestrator are already initialized

why?

packages/orchestrator/orchestrator.proto
```diff
@@ -88,7 +88,37 @@ message SandboxListCachedBuildsResponse {
...
+  Orchestrator = 1;
+}
+
+message ServiceInfoResponse {
+  string node_id = 1;
+  string service_id = 2;
+  string service_version = 3;
+  string service_commit = 4;
+
+  ServiceInfoStatus service_status = 5;
```
dobrac: should we leave some ids gap for the first group too?

packages/orchestrator/orchestrator.proto
```diff
@@ -99,3 +129,8 @@ service SandboxService {
 
   rpc ListCachedBuilds(google.protobuf.Empty) returns (SandboxListCachedBuildsResponse);
 }
+
+service InfoService {
```
dobrac: I would put this service to a separate file

packages/api/internal/orchestrator/cache.go
```diff
@@ -131,21 +135,43 @@ func (o *Orchestrator) syncNode(ctx context.Context, node *Node, nodes []*node.N
...
+		node.SetStatus(o.getNodeStatusConverted(nodeInfo.ServiceStatus))
 
-	builds, buildsErr := o.listCachedBuilds(ctx, node.Info.ID)
-	if buildsErr != nil {
-		zap.L().Error("Error listing cached builds", zap.Error(buildsErr))
-		return
+		activeInstances, instancesErr := o.getSandboxes(ctx, node.Info)
+		if instancesErr != nil {
+			zap.L().Error("Error getting instances", zap.Error(instancesErr))
+			continue
```
dobrac, line 153: Should the retry logic be separate for syncing instanceCache and for syncing builds? Otherwise if build sync fails, it will resync instanceCache again

dobrac (changes requested):

packages/orchestrator/main.go
sitole: Ou, its just server struct and listen is done below all that configuration. They comment does not make sense. Will remove.

packages/api/internal/orchestrator/cache.go
```diff
@@ -131,18 +135,39 @@ func (o *Orchestrator) syncNode(ctx context.Context, node *Node, nodes []*node.N
...
+	if !syncRetrySuccess {
+		zap.L().Error("Failed to sync node after max retries, temporarily marking as draining", zap.String("node_id", node.Info.ID))
+		node.SetStatus(api.NodeStatusDraining)
+		return
+	}
 
 	builds, buildsErr := o.listCachedBuilds(ctx, node.Info.ID)
 	if buildsErr != nil {
 		zap.L().Error("Error listing cached builds", zap.Error(buildsErr))
-		return
```
dobrac: should we sync build even when the error is returned?

packages/orchestrator/info.proto
```diff
@@ -0,0 +1,43 @@
+syntax = "proto3";
+
+import "google/protobuf/empty.proto";
+import "google/protobuf/timestamp.proto";
+
+option go_package = "https://github.com/e2b-dev/infra/orchestrator";
+
+// needs to be different from the enumeration in the template manager
```
dobrac, line 8: would it make sense to remove it in the template manager then and replace it with this?

packages/orchestrator/internal/grpcserver/healthcheck.go
dobrac: we're using it in the internal-cli

packages/orchestrator/internal/server/sandboxes_test.go
```diff
@@ -66,6 +66,7 @@ func Test_server_List(t *testing.T) {
 			s := &server{
 				sandboxes: smap.New[*sandbox.Sandbox](),
 				tracer:    noop.NewTracerProvider().Tracer(""),
+				info:      &ServiceInfo{ClientId: ""},
```
dobrac:
```suggestion
				info:      &ServiceInfo{},
```

dobrac (changes requested):

packages/orchestrator/info.proto
sitole: Make sense to use info service here. Added new status check into API with backup to template manager status that can be removed in future where all API and template managers will contain new info service logic.

packages/api/internal/orchestrator/cache.go
sitole: true, removed

dobrac (approved):

PR #646, opened 2025-05-14
Remove running Docker container in template build

dobrac:
### Description
This PR removes Docker container run during the template build and replaces it with Firecracker run.

The process looks now:

1. Get Docker image from the remote repository
2. Inject new file layers with the required setup for hostname, dns, envd service configuration, basic provisioning script that is run before most of the VM services
3. Extract the ext4 filesystem
4. Start FC VM with BusyBox init that runs just the provisioning script, wait for exit. This will install systemd, that is later used for proper VM boot.
5. Start the FC VM (using systemd) and wait for Envd
5. Run two additional commands:
     - configuration script (enable swap, create user, change folder permissions, etc)
     - start command (if defined)
6. Snapshot
7. Upload template

This PR should also fix sometimes missing hostname from /etc/hosts as this is now completely in our scope and not docker container run. The hostname is set to `e2b.local`.

### Systemd
Because we are using systemd as init, it needs to be supplied before the first start of the FC. Because this is an operation, that is required for all images, it doesn't make much sense to spend time on each template build. For this reason, the systemd is installed using provisioning script.

### User speed optimizations
If the user wants the build to be faster for iterations, they can preinstall the following packages in their Dockerfile `systemd systemd-sysv openssh-server sudo socat chrony linuxptp`. Their installation will be correctly detected by the provisioning script and installation will be skipped. This speeds up provisioning by 10s.

Another optimization is with running `chmod 777 -R /usr/local`. In the Docker container, this command was taking minutes in the code interpreter template. After this change, it takes around 1 second.

### Old Envd
This version doesn't include old (websockets) env version. Further removal will be done in future PR.


packages/orchestrator/internal/sandbox/fc/process.go
```diff
@@ -268,7 +269,22 @@ func (p *Process) Create(
 
 	// IPv4 configuration - format: [local_ip]::[gateway_ip]:[netmask]:hostname:iface:dhcp_option:[dns]
 	ipv4 := fmt.Sprintf("%s::%s:%s:instance:%s:off:%s", p.slot.NamespaceIP(), p.slot.TapIP(), p.slot.TapMaskString(), p.slot.VpeerName(), p.slot.TapName())
-	kernelArgs := fmt.Sprintf("quiet loglevel=1 ip=%s ipv6.disable=0 ipv6.autoconf=1 reboot=k panic=1 pci=off nomodules i8042.nokbd i8042.noaux random.trust_cpu=on", ipv4)
```
dobrac, line 271: the `nomodules` had no effect, as the correct kernel arg name is `nomodule`

https://github.com/firecracker-microvm/firecracker/blame/25480f41921eae4e3486f5a339b91e484da166ad/CHANGELOG.md#L603C12-L603C12

packages/orchestrator/internal/sandbox/fc/process.go
```diff
@@ -268,7 +270,22 @@ func (p *Process) Create(
...
+	args := KernelArgs{
+		"console":                             "ttyS0",
+		"loglevel":                            "5", //KERN_NOTICE
+		"systemd.journald.forward_to_console": "",
```
dobrac: Should we keep this for prod?

packages/orchestrator/internal/sandbox/block/cache.go
```diff
@@ -158,10 +158,30 @@ func (m *Cache) Close() error {
...
+
+	f, err := os.OpenFile(m.filePath, os.O_RDWR, 0o644)
+	if err != nil {
+		e = errors.Join(e, fmt.Errorf("error opening file: %w", err))
+	}
+
+	err = f.Sync()
+	if err != nil {
+		e = errors.Join(e, fmt.Errorf("error syncing file: %w", err))
+	}
```
jakubno: what is the idea behind calling sync here?

packages/orchestrator/internal/sandbox/block/cache.go
```diff
@@ -158,10 +158,30 @@ func (m *Cache) Close() error {
...
-	)
+	err := m.mmap.Flush()
+	if err != nil {
+		e = errors.Join(e, fmt.Errorf("error flushing mmap: %w", err))
+	}
+
+	err = m.mmap.Unmap()
+	if err != nil {
+		e = errors.Join(e, fmt.Errorf("error unmapping mmap: %w", err))
+	}
```
jakubno: // Unmap deletes the memory mapped region, flushes any remaining changes, and sets
// m to nil.

Is it neeeded to call flush?

packages/orchestrator/internal/sandbox/fc/process.go
dobrac: Let's disable this for prod (https://github.com/firecracker-microvm/firecracker/blob/main/docs/prod-host-setup.md#logging-and-performance)

packages/orchestrator/internal/sandbox/diffcreator.go
```diff
@@ -20,15 +20,11 @@ type DiffCreator interface {
 }
 
 type RootfsDiffCreator struct {
-	rootfs   *rootfs.CowDevice
+	rootfs   rootfs.Provider
 	stopHook func(context.Context) error
 }
 
 func (r *RootfsDiffCreator) process(ctx context.Context, out io.Writer) (*header.DiffMetadata, error) {
-	if err := r.rootfs.Flush(ctx); err != nil {
```
jakubno, line 28: this isn't needed anymore?

packages/orchestrator/internal/template/build/configure.sh
```diff
@@ -0,0 +1,42 @@
...
+chmod 777 -R /usr/local
+
+echo "Create /code directory"
+mkdir -p /code
+echo "Give 777 permission to /code"
+chmod 777 -R /code
+
+echo "Finished configuration script"
+
+exit 0
```
jakubno: is `exit 0` neeeded?

packages/orchestrator/internal/template/build/template_builder.go
```diff
@@ -123,22 +140,66 @@ func (b *TemplateBuilder) Build(ctx context.Context, template *TemplateConfig, e
...
-		return err
+		return nil, fmt.Errorf("error building environment: %w", err)
 	}
 
+	localTemplate, err := templatelocal.NewLocalTemplate(templateCacheFiles, rootfs, memfile)
+	if err != nil {
+		postProcessor.WriteMsg(fmt.Sprintf("Error creating local template: %v", err))
+		return nil, fmt.Errorf("error creating local template: %w", err)
+	}
+	defer localTemplate.Close()
```
jakubno: If there was an error for `NewLocalTemplate` (which it currently won't), you would leak rootfs and memory

packages/orchestrator/internal/sandbox/sandbox_linux.go
```diff
@@ -214,17 +215,6 @@ func CreateSandbox(
 		return sbx.Close(ctx, tracer)
 	})
 
-	err = sbx.waitForStart(
-		ctx,
-		tracer,
-	)
-	if err != nil {
-		return nil, cleanup, fmt.Errorf("failed to wait for sandbox start: %w", err)
-	}
```
jakubno, line 223: why did you move this? isn't it more unintuitive now? In resume we wait, in create we don't. Maybe making it an option?

packages/orchestrator/internal/template/build/oci/oci.go
```diff
@@ -0,0 +1,127 @@
...
+			return 0, fmt.Errorf("error getting layer (%d) size: %w", index, err)
+		}
+		imageSize += layerSize
+	}
+
+	return imageSize, nil
+}
+
+func ToExt4(ctx context.Context, img v1.Image, rootfsPath string, sizeLimit int64) error {
+	pr := mutate.Extract(img)
```
jakubno: what `pr` stands for?

packages/orchestrator/internal/template/build/tar.go
```diff
@@ -2,58 +2,82 @@ package build
...
+}
+
+// LayerFile creates a layer from a single file map. These layers are reproducible and consistent.
+// A filemap is a path -> file content map representing a file system.
+func LayerFile(filemap map[string]layerFile) (v1.Layer, error) {
+	b := &bytes.Buffer{}
+	w := tar.NewWriter(b)
 
-		return errMsg
+	fn := []string{}
```
jakubno: could you improve the naming?

packages/orchestrator/internal/sandbox/block/cache.go
dobrac: no, removed

packages/orchestrator/internal/sandbox/block/cache.go
dobrac: removed

packages/orchestrator/internal/template/build/configure.sh
dobrac: removed

packages/orchestrator/internal/template/build/oci/oci.go
dobrac: renamed to `r` as reader

packages/orchestrator/internal/template/build/tar.go
dobrac: renamed to `names`

packages/orchestrator/internal/sandbox/sandbox_linux.go
dobrac: renamed to waitForEnvd so it's more explicit

packages/orchestrator/internal/template/build/template_builder.go
dobrac: Removed impossible error case

packages/orchestrator/internal/sandbox/diffcreator.go
dobrac: It was moved to the close method to properly flush nbd even on clean shutdown of sandbox

packages/orchestrator/internal/sandbox/fc/process.go
jakubno: https://linear.app/e2b/issue/E2B-2343/try-to-use-nomodule-as-kernel-arg

packages/orchestrator/internal/sandbox/fc/process.go
```diff
@@ -270,20 +281,37 @@ func (p *Process) Create(
...
-		"ip":                                  ipv4,
-		"ipv6.disable":                        "0",
-		"ipv6.autoconf":                       "1",
-		"reboot":                              "k",
-		"panic":                               "1",
-		"pci":                                 "off",
-		"i8042.nokbd":                         "",
-		"i8042.noaux":                         "",
-		"random.trust_cpu":                    "on",
+		// Disable kernel logs for production to speed the the FC operations
```
jakubno:
```suggestion
		// Disable kernel logs for production to speed the the FC operations
		// https://github.com/firecracker-microvm/firecracker/blob/main/docs/prod-host-setup.md#logging-and-performance
```

PR #653, opened 2025-05-16
Use service info status to report orchestrator health check

sitole:
- moved service info struct into grpc server (dependency cycle)
- health check now reports status from shared service info
- this will allow us to remotely change orchestrator / template manager status


packages/orchestrator/internal/healthcheck/healthcheck.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+	}()
+}
+
+func (h *Healthcheck) getStatus() e2bHealth.Status {
+	switch h.info.GetStatus() {
+	case e2borchestratorinfo.ServiceInfoStatus_OrchestratorHealthy:
+		return e2bHealth.Healthy
+	case e2borchestratorinfo.ServiceInfoStatus_OrchestratorDraining:
+		return e2bHealth.Healthy // orchestration is currently not supporting draining and its manager by api/edge api
```
dobrac: why is it not supporting draining?

packages/orchestrator/internal/grpcserver/server_info.go
```diff
@@ -1,4 +1,4 @@
-package info
+package grpcserver
```
dobrac: is this grpcserver related or separate service?

packages/orchestrator/internal/service/info.go
```diff
@@ -0,0 +1,74 @@
...
+
+func (s *ServiceInfo) SetStatus(status orchestratorinfo.ServiceInfoStatus) {
+	s.statusMu.Lock()
+	defer s.statusMu.Unlock()
+
+	if s.status != status {
+		zap.L().Info("Service status changed", zap.String("status", status.String()))
+		s.status = status
+	}
+}
```
dobrac, line 49: isn't this in conflict with the other PR?

dobrac (changes requested):

packages/orchestrator/internal/service/info.go
sitole: this is not related with https://github.com/e2b-dev/infra/pull/654 if you mean this one.

packages/orchestrator/internal/healthcheck/healthcheck.go
sitole: Added state to our shared health response struct and adjusted API to handle it

packages/orchestrator/internal/grpcserver/server_info.go
sitole: moved into service package more related to this stuff

dobrac (approved):

PR #654, opened 2025-05-16
API: Request to node status will send status change to the orchestrator

sitole:
- currently there was just per-api health status adjustment
- this new thing now requests status change on orchestrator itself and waits for propagation via node sync

### DIFF
--- a/packages/api/internal/orchestrator/node.go
+++ b/packages/api/internal/orchestrator/node.go
@@ -76,10 +76,10 @@ func (n *Node) setStatus(status api.NodeStatus) {
 	}
 }
 
-func (n *Node) SendStatusChange(ctx context.Context, nodeId string, s api.NodeStatus) error {
+func (n *Node) SendStatusChange(ctx context.Context, s api.NodeStatus) error {
 	nodeStatus, ok := ApiNodeToOrchestratorStateMapper[s]
 	if !ok {
-		zap.L().Error("Unknown service info status", zap.Any("status", s), zap.String("node_id", nodeId))
+		zap.L().Error("Unknown service info status", zap.Any("status", s), zap.String("node_id", n.Info.ID))
 		return fmt.Errorf("unknown service info status: %s", s)
 	}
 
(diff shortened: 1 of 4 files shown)


packages/api/internal/orchestrator/node.go
```diff
@@ -65,7 +66,7 @@ func (n *Node) Status() api.NodeStatus {
 	return n.status
 }
 
-func (n *Node) SetStatus(status api.NodeStatus) {
+func (n *Node) setStatus(status api.NodeStatus) {
```
dobrac, line 69: do we need this method? Shouldn't it use SendStatusChange?

packages/api/internal/orchestrator/node.go
```diff
@@ -75,6 +76,22 @@ func (n *Node) SetStatus(status api.NodeStatus) {
 	}
 }
 
+func (n *Node) SendStatusChange(ctx context.Context, nodeId string, s api.NodeStatus) error {
```
dobrac: Why do you need to specify the nodeId? It should be already in Node

sitole: done, missed

packages/api/internal/orchestrator/node.go
sitole: `SetState` and `SendStatusChange` are different. Set state sets state into pool node during health check. Status change calls orchestrator and forces it to change state.

dobrac: But the usage is only here in the REST method:

```
func (a *APIStore) PostNodesNodeID(c *gin.Context, nodeId api.NodeID) {
	ctx := c.Request.Context()

	body, err := utils.ParseBody[api.PostNodesNodeIDJSONRequestBody](ctx, c)
	if err != nil {
		a.sendAPIStoreError(c, http.StatusBadRequest, fmt.Sprintf("Error when parsing request: %s", err))

		errMsg := fmt.Errorf("error when parsing request: %w", err)

		return
	}

	node := a.orchestrator.GetNode(nodeId)
	if node == nil {
		c.Status(http.StatusNotFound)
		return
	}

	node.SetStatus(body.Status)

}
```

sitole: Its also used by `syncNode`. I want to have it in separated method to make sure status mutex will be called every time properly

dobrac: my bad, old branch

dobrac (approved):

PR #655, opened 2025-05-16
Added fields to envd stat method response

mishushakov:
- Added fields such as size, mode, permissions, owner, group, modified time to file stat and list response
- Added test for stat method


tests/integration/internal/tests/envd/filesystem_test.go
```diff
@@ -179,7 +179,61 @@ func TestFilePermissions(t *testing.T) {
 			assert.Contains(t, line, "user user")
 		}
 	}
+}
+
+func TestStat(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
```
jakubno, line 188: the cancel is useless if you don't really use it (defer is called as the last thing, so there's nothing to cancel)

tests/integration/internal/tests/envd/filesystem_test.go
```diff
@@ -179,7 +179,61 @@ func TestFilePermissions(t *testing.T) {
...
+	assert.NoError(t, err)
+	assert.Equal(t, http.StatusOK, createFileResp.StatusCode())
+
+	req := connect.NewRequest(&filesystem.StatRequest{
+		Path: filePath,
+	})
+	setup.SetSandboxHeader(req.Header(), sbx.SandboxID, sbx.ClientID)
+	setup.SetUserHeader(req.Header(), "user")
+	statResp, err := envdClient.FilesystemClient.Stat(ctx, req)
+	assert.NoError(t, err)
```
jakubno:
```suggestion
	require.NoError(t, err)
```

tests/integration/internal/tests/envd/filesystem_test.go
```diff
@@ -179,7 +179,61 @@ func TestFilePermissions(t *testing.T) {
...
+
+	// Verify the stat response
+	assert.NotNil(t, statResp.Msg)
```
jakubno:
```suggestion
	require.NotNil(t, statResp.Msg)
```

tests/integration/internal/tests/envd/filesystem_test.go
```diff
@@ -179,7 +179,61 @@ func TestFilePermissions(t *testing.T) {
...
+	assert.NotNil(t, statResp.Msg.Entry)
```
jakubno:
```suggestion
	require.NotNil(t, statResp.Msg.Entry)
```

jakubno (changes requested):
Can you make sure this works with old SDK, there's for some reason a lot of changes in `packages/envd/internal/services/spec/filesystem/filesystem.pb.go`

tests/integration/internal/tests/envd/filesystem_test.go
mishushakov: can you expand more on this? is it that only that particular instance that is concerning? I might be missing something - this is how all the other tests are structured, so I just took it as a starting point

mishushakov:
> Can you make sure this works with old SDK

The only "breaking" change in that PR is that I've expanded the return type for files.list(), but the new type should be compatible with the old type (only added extra fields). I will make sure old SDKs are passing tests, though.

> there's for some reason a lot of changes in `packages/envd/internal/services/spec/filesystem/filesystem.pb.go`

Might be due to the fact that the version of protobuf generators that I am using differs from the one that was used to create the original file.

mishushakov:
I have verified that past versions of the SDK still work as expected.

tests/integration/internal/tests/envd/filesystem_test.go
jakubno: it's just NIT, it's fine to leave it as it is. I wasn't sure if you didn't create the cancel for some purpose and then forgot

jakubno (commented):
LGTM. Can you just increase a patch version in envd?

mishushakov:
Okay this is surprising, it worked for me (on current main with E2B debug). But, I believe in that case extending the type will cause the same issue? The client will try to parse message and it will have extra attributes it's not supposed to receive. Might also be because it's an array.

I have removed the changes to the type for ListDirResponse

```diff
message ListDirResponse {
-    repeated EntryInfoExtended entries = 1;
}
``` 

```diff
message ListDirResponse {
+    repeated EntryInfo entries = 1;
}
```

mishushakov:
I think the error above might be specific to Python? Did you have the same error message with JS SDK? Let me see, maybe there's a setting that controls how strict it is parsed.

mishushakov:
If we still want the additional fields in files.list(), I might actually create a new version of the ListDir rpc after the PR is merged

mishushakov:
Can I send back a different type which has the same fields of the expected type? I will then extend EntryInfo with some fields for Stat and make current EntryInfo > EntryInfoBasic with only the current 3 fields

jakubno:
I don't think so, you can try but I doubt that, based on how grpc works. If you want to introduce a new type and still have backwards compatibility you will need add a new method

packages/envd/internal/services/filesystem/stat.go
```diff
@@ -31,12 +32,21 @@ func (Service) Stat(ctx context.Context, req *connect.Request[rpc.StatRequest])
 		return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("error statting file: %w", err))
 	}
 
+	owner, group := getFileOwnership(fileInfo)
+	fileMode := fileInfo.Mode()
+
 	return connect.NewResponse(
 		&rpc.StatResponse{
 			Entry: &rpc.EntryInfo{
```
jakubno: could you simplify by some util function which would create `EntryInfo` from `fileInfo`? So you don't have to double the logic?

jakubno (changes requested):
Add test for list as well

packages/envd/internal/services/filesystem/utils.go
```diff
@@ -15,3 +22,45 @@ func getEntryType(entry osEntry) rpc.FileType {
...
+	if g, err := user.LookupGroupId(fmt.Sprintf("%d", sys.Gid)); err == nil {
+		group = g.Name
+	} else {
+		group = fmt.Sprintf("%d", sys.Gid)
+	}
+
+	return owner, group
+}
+
+func EntryInfoFromFileInfo(fileInfo os.FileInfo, path string) *rpc.EntryInfo {
```
jakubno: could you change it to internal? it isn't used elsewhere
```suggestion
func entryInfoFromFileInfo(fileInfo os.FileInfo, path string) *rpc.EntryInfo {
```

jakubno (approved):

PR #656, opened 2025-05-16
Orchestrator node id is now instance name

sitole:
This will make node id compatible with future non-gcp deployments.


jakubno (approved):

tests/integration/internal/setup/envd_client.go
```diff
@@ -41,7 +41,7 @@ func GetEnvdClient(tb testing.TB, _ context.Context) *EnvdClient {
 
 func WithSandbox(sandboxID string, clientID string) func(ctx context.Context, req *http.Request) error {
```
jakubno: clientID isn't used anymore, can we also simplify

PR #658, opened 2025-05-17
Use the FC api /entropy endpoint to configure the virtio-rng device

0div:
this is equivalent to running
```sh
curl --unix-socket $socket_location -i \
    -X PUT 'http://localhost/entropy' \
    -H 'Accept: application/json' \
    -H 'Content-Type: application/json' \
    -d "{
        \"rate_limiter\": {
            \"bandwidth\": {
                \"size\": 1000,
                \"one_time_burst\": 0,
                \"refill_time\": 100
            }
        }
    }"
  ```

### DIFF
--- a/packages/orchestrator/internal/sandbox/fc/process.go
+++ b/packages/orchestrator/internal/sandbox/fc/process.go
@@ -312,6 +312,14 @@ func (p *Process) Create(
 	}
 	telemetry.ReportEvent(childCtx, "set fc machine config")
 
+	err = p.client.setEntropyDevice(childCtx, 1000, 0, 100)
+	if err != nil {
+		fcStopErr := p.Stop()
+
+		return errors.Join(fmt.Errorf("error setting fc entropy config: %w", err), fcStopErr)
+	}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/orchestrator/internal/sandbox/fc/process.go [hunk 1]
```diff
@@ -312,6 +312,14 @@ func (p *Process) Create(
 	}
 	telemetry.ReportEvent(childCtx, "set fc machine config")
 
+	err = p.client.setEntropyDevice(childCtx, 1000, 0, 100)
```
jakubno: Can you make these constants?

dobrac (approved):
Nice! Could you please add some integration tests for this or it's not possible?

jakubno (changes requested):
> > Nice! Could you please add some integration tests for this or it's not possible?

> I'm currently working on solving a code-interpreter same seed problem that will validate this or not.

The test should be here, otherwise it's possible to deploy broken version and find out too late

PR #668, opened 2025-05-19
Fix nil pointer for old templates

dobrac:
Fix nil pointer for old templates


packages/orchestrator/internal/sandbox/template/storage.go
```diff
@@ -12,6 +12,8 @@ import (
 	"github.com/e2b-dev/infra/packages/shared/pkg/storage/header"
 )
 
+const oldRootfsBlockSize = 2 << 11
```
jakubno:
```suggestion
const oldRootfsBlockSize = 2 << 11  // 4 KiB
```

jakubno (approved):

PR #671, opened 2025-05-20
Orchestrators network namespaces safe-acquire

sitole:
- Local-only implementation of network namespaces acquire that is safe to run on not clean environments.
- During startup checks all already create namespaces and adds them into foreign "block list" to skip them.
- When available namespace is found it will test if its really not existing (was not created during program lifetime).

Notes:
- Acquire is infinite loop with context timeout set to 5 seconds. We should lower it, will check that.
- When acquire function will find empty namespaces but then its marked as available maybe we should also put it into map and fast-skip it later.

### DIFF
--- a/packages/orchestrator/internal/sandbox/network/storage.go
+++ b/packages/orchestrator/internal/sandbox/network/storage.go
@@ -1,6 +1,6 @@
 package network
 
-import "github.com/e2b-dev/infra/packages/shared/pkg/env"
+import "go.opentelemetry.io/otel/trace"
 
 type Storage interface {
 	Acquire() (*Slot, error)
@@ -8,10 +8,6 @@ type Storage interface {
 }
 
 // NewStorage creates a new slot storage based on the environment, we are ok with using a memory storage for local
-func NewStorage(slotsSize int, clientID string) (Storage, error) {
-	if env.IsLocal() {
-		return NewStorageMemory(slotsSize)
-	} else {
-		return NewStorageKV(slotsSize, clientID)
-	}
+func NewStorage(slotsSize int, clientID string, tracer trace.Tracer) (Storage, error) {
+	return NewStorageLocal(slotsSize, tracer)
 }
--- a/packages/orchestrator/internal/sandbox/network/storage.go
+++ b/packages/orchestrator/internal/sandbox/network/storage.go
@@ -1,13 +1,15 @@
 package network
 
 import (
+	"context"
+
 	"go.opentelemetry.io/otel/trace"
 
 	"github.com/e2b-dev/infra/packages/shared/pkg/env"
 )
 
 type Storage interface {
-	Acquire() (*Slot, error)
+	Acquire(ctx context.Context) (*Slot, error)
 	Release(*Slot) error
 }
 
--- a/packages/orchestrator/internal/sandbox/network/storage.go
+++ b/packages/orchestrator/internal/sandbox/network/storage.go
@@ -1,6 +1,10 @@
 package network
 
-import "go.opentelemetry.io/otel/trace"
+import (
+	"go.opentelemetry.io/otel/trace"
+
+	"github.com/e2b-dev/infra/packages/shared/pkg/env"
+)
 
 type Storage interface {
 	Acquire() (*Slot, error)
@@ -9,5 +13,9 @@ type Storage interface {
 
 // NewStorage creates a new slot storage based on the environment, we are ok with using a memory storage for local
 func NewStorage(slotsSize int, clientID string, tracer trace.Tracer) (Storage, error) {
-	return NewStorageLocal(slotsSize, tracer)
+	if env.IsDevelopment() {
+		return NewStorageLocal(slotsSize, tracer)
+	}
+
+	return NewStorageKV(slotsSize, clientID)
 }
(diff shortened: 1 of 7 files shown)


packages/orchestrator/internal/sandbox/network/storage.go
```diff
@@ -1,17 +1,13 @@
...
+import "go.opentelemetry.io/otel/trace"
 
 type Storage interface {
 	Acquire() (*Slot, error)
 	Release(*Slot) error
 }
 
 // NewStorage creates a new slot storage based on the environment, we are ok with using a memory storage for local
-func NewStorage(slotsSize int, clientID string) (Storage, error) {
-	if env.IsLocal() {
```
sitole, line 12: Maybe for now we should implement env-based switch and Memory/KV storage should still be preferred.

packages/orchestrator/internal/sandbox/network/storage_local.go
```diff
@@ -0,0 +1,167 @@
...
+		foreignNs:    foreignNsMap,
+		slotsSize:    slotsSize,
+		acquiredNs:   make(map[string]struct{}, slotsSize),
+		acquiredNsMu: sync.Mutex{},
+		tracer:       tracer,
+	}, nil
+}
+
+func (s *StorageLocal) Acquire() (*Slot, error) {
+	acquireTimeoutCtx, acquireCancel := context.WithTimeout(context.Background(), time.Millisecond*500)
```
sitole: Want to see some heavily used cluster. For me its mostly around 806 µs.
Times should be pretty consistent when there is not much newly create foreign namespaces during orchestrator run.

packages/orchestrator/internal/sandbox/network/storage_memory.go
```diff
@@ -45,7 +44,3 @@ func (s *StorageMemory) Release(ips *Slot) error {
 
 	return nil
 }
-
-func getMemoryKey(slotIdx int) string {
```
dobrac: why is it removed here?

packages/orchestrator/internal/sandbox/network/storage_local.go
```diff
@@ -0,0 +1,167 @@
...
+
+	return ns, nil
+}
+
+func getSlotName(slotIdx int) string {
+	slotIdxStr := strconv.Itoa(slotIdx)
+	return fmt.Sprintf("ns-%s", slotIdxStr)
+}
+
+func getMemoryKey(slotIdx int) string {
```
dobrac:
```suggestion
func getLocalKey(slotIdx int) string {
```

packages/orchestrator/internal/sandbox/network/storage_local.go
dobrac: Can we pass the `context.Background()` from up the call stack?

packages/orchestrator/internal/sandbox/network/storage_local.go
```diff
@@ -0,0 +1,167 @@
...
+
+	s.acquiredNsMu.Lock()
+	defer s.acquiredNsMu.Unlock()
+
+	// we skip the first slot because it's the host slot
+	slotIdx := 1
+
+	for {
+		select {
+		case <-acquireTimeoutCtx.Done():
```
dobrac, line 63: I think you should use the tracer context after it's creation

packages/orchestrator/internal/sandbox/network/storage_local.go
```diff
@@ -0,0 +1,167 @@
...
+			return nil, fmt.Errorf("failed to acquire IP slot: timeout")
+		default:
+			if len(s.acquiredNs) >= s.slotsSize {
```
dobrac: shouldn't this be > instead of >=

dobrac (changes requested):

packages/orchestrator/internal/sandbox/network/storage_local.go
```diff
@@ -0,0 +1,167 @@
...
+				return nil, fmt.Errorf("failed to acquire IP slot: no empty slots found")
+			}
+
+			slotIdx++
```
dobrac, line 70: could we store the free slots somewhere so we wouldn't need to scan through all of them each time?

packages/orchestrator/internal/sandbox/network/storage_memory.go
sitole: just moved to local storage, that should be permanently here.

packages/orchestrator/internal/sandbox/network/storage_local.go
sitole: If we will store them during init phase then we still need to check if they are not used in acquire method, because some other program can use them in meantime, right?

If it will not be available we need to fallback to iteration again, right?

packages/orchestrator/internal/sandbox/network/storage_memory.go
sitole: moved back and create `getLocalKey` for new impl.

packages/orchestrator/internal/sandbox/network/storage_local.go
dobrac: Yes, but the iteration will go only through the potential candidates. If any of them is unusable, it will exclude it forever from the iteration. 

The slot might become available, but I'm not sure if you handle it now and if you can reuse it.

sitole: If i will save list of potential candidates and then some of them is not available i need to adjust list of candidates to match at least size of node pool, right? Just making sure we want to implement it same way.

dobrac: Actually, you might not need to worry about this, because this optimization might be already implemented in the upper class

dobrac (approved):

PR #672, opened 2025-05-20
Fix memfile header for old templates

dobrac:
Fix memfile header for old templates


packages/orchestrator/internal/sandbox/template/storage.go
```diff
@@ -64,14 +65,24 @@ func NewStorage(
 			return nil, fmt.Errorf("failed to parse build id: %w", err)
 		}
 
+		// TODO: This is a workaround for the old style template without a header.
+		// We don't know the block size of the old style template, so we set it manually.
+		var blockSize uint64
```
jakubno, line 70: Wouldn't switch be better

PR #675, opened 2025-05-22
make the variables readonly

djeebus:
'local readonly' makes a variable called "readonly" local, which isn't actually what we want to have happen here.

To be clear, my IDE warned me of this (thanks to shellchecks' [SC2316](https://github.com/koalaman/shellcheck/wiki/SC2316) warning). I did throw together a demo to prove that it was correct:

```bash
#!/usr/bin/env bash

function test1() {
    local readonly one=$1
    local readonly two=$2
    one=$two
    echo "test1: that shouldn't have worked"
}

function test2() {
    local -r one=$1
    local -r two=$2
    one=$two    # line 13
    echo "test2: that shouldn't have worked"
}

test1
test2
```

When running bash 5.2.37(1) locally, it results in the following output:

```
$ ./test.sh 
test1: that shouldn't have worked
./test.sh: line 13: one: readonly variable
```

Not particularly important in the places it's being used here, but if it's going to be committed, it may as well work :grin: 

Really interesting work here, I'm enjoying digging through it. Thanks!

### DIFF
--- a/packages/cluster-disk-image/setup/install-nomad.sh
+++ b/packages/cluster-disk-image/setup/install-nomad.sh
@@ -31,30 +31,30 @@ function print_usage {
 }
 
 function log {
-  local readonly level="$1"
-  local readonly message="$2"
-  local readonly timestamp=$(date +"%Y-%m-%d %H:%M:%S")
+  local -r level="$1"
+  local -r message="$2"
+  local -r timestamp=$(date +"%Y-%m-%d %H:%M:%S")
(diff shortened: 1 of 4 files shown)


dobrac:
Hi @djeebus, thank you for your contribution! Indeed, the behavior of `local readonly` is not as it looks like and `local -r` looks to be correct. Accepting you PR, good work!

dobrac (approved):

PR #680, opened 2025-05-22
Return a correct status code for resume conflict

jakubno:
# Description

If you tried to resume a sandbox while it's already resuming it failed and returned 429 incorrectly. Now it returns 409

### DIFF
--- a/packages/api/internal/cache/instance/reservation.go
+++ b/packages/api/internal/cache/instance/reservation.go
@@ -23,16 +23,11 @@ func NewReservationCache() *ReservationCache {
 	}
 }
 
-func (r *ReservationCache) reserve(instanceID string, team uuid.UUID) error {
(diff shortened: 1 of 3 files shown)


packages/api/internal/cache/instance/reservation.go
```diff
@@ -79,16 +82,18 @@ func (c *InstanceCache) Reserve(instanceID string, team uuid.UUID, limit int64)
...
-		return fmt.Errorf("error when reserving instance: %w", err), nil
+	inserted := c.reservations.insertIfAbsent(instanceID, team)
+	if !inserted {
+		return false, nil, &ErrAlreadyBeingStarted{
+			sandboxID: instanceID,
+		}
 	}
 
-	return nil, func() {
+	return false, func() {
```
dobrac: instead of returning true/false, how about returning also an error here?

dobrac (approved):

PR #688, opened 2025-05-23
Use Go Tools for defining tools versions

dobrac:
Use Go Tools for defining tools versions

Updates the connectrpc.com/connect to v1.18.1 for compatibility reasons with the tools (doing regenerate doesn't seem to produce different results).

### TODO
- [ ] Test on not set up environment
- [x] Run SDK tests


jakubno:
> Updates the connectrpc.com/connect to v1.18.1 for compatibility reasons with the tools (doing regenerate doesn't seem to produce different results).

We use `connectrpc.com/connect` as a library, could you also test with SDK tests if everything works?

jakubno (approved):

PR #689, opened 2025-05-23
Add execution ID

jakubno:
# Description

Adds a unique ID per execution (this will be different when sandbox is paused and resumed)


packages/api/internal/cache/instance/instance.go
```diff
@@ -50,8 +51,14 @@ func NewInstanceInfo(
 	EnvdAccessToken *string,
 	BaseTemplateID string,
 ) *InstanceInfo {
+	// TODO: Temporary workaround, until all orchestrators report this
```
jakubno: I could also separate the PR and add the fallback only to one of the services (api/orchestrator). This is caused by the circular dependency

packages/api/internal/cache/instance/instance.go
```diff
@@ -50,8 +51,14 @@ func NewInstanceInfo(
...
+	if ExecutionID == "" {
+		ExecutionID = uuid.New().String()
```
dobrac: I would move this to the getSandboxes function when you receive the data from the orchestrator

dobrac (approved):

PR #695, opened 2025-05-24
Make sandbox internet access configurable

dobrac:


packages/orchestrator/internal/config/config.go
```diff
@@ -0,0 +1,5 @@
+package config
+
+const (
+	AllowSandboxInternet = true
```
jakubno: can this be a env var or how should we configure it?

packages/orchestrator/internal/sandbox/network/slot.go
```diff
@@ -116,3 +124,87 @@ func (s *Slot) TapCIDR() string {
...
+			if err != nil {
+				return fmt.Errorf("error setting firewall rules: %w", err)
+			}
+		}
+
+		return nil
+	})
+	if err != nil {
+		return fmt.Errorf("failed execution in network namespace '%s': %w", s.NamespaceID(), err)
+	}
```
jakubno: you don't need to do anything from this if not `allowInternet`, right?

packages/orchestrator/internal/sandbox/network/storage_local.go
```diff
@@ -21,7 +21,7 @@ type StorageLocal struct {
 	tracer       trace.Tracer
 }
 
-const netNamespacesDir = "/var/run/netns"
+const NetNamespacesDir = "/var/run/netns"
```
jakubno: why uppercase it?

packages/orchestrator/internal/sandbox/sandbox_linux.go
```diff
@@ -77,26 +77,33 @@ func (m *Metadata) LoggerMetadata() sbxlogger.SandboxMetadata {
...
+	slot *network.Slot
+	err  error
+}
+
 func CreateSandbox(
 	ctx context.Context,
 	tracer trace.Tracer,
 	networkPool *network.Pool,
-	devicePool *nbd.DevicePool,
+	_ *nbd.DevicePool,
```
jakubno, line 89: do we need to pass it at all then?

packages/orchestrator/internal/sandbox/sandbox_linux.go
```diff
@@ -77,26 +77,33 @@ func (m *Metadata) LoggerMetadata() sbxlogger.SandboxMetadata {
...
 
 	cleanup := NewCleanup()
 
-	ips, err := getNetworkSlot(childCtx, tracer, networkPool, cleanup)
-	if err != nil {
-		return nil, cleanup, fmt.Errorf("failed to get network slot: %w", err)
-	}
+	ipsCh := getNetworkSlotAsync(childCtx, tracer, networkPool, cleanup, allowInternet)
+	defer func() {
+		// Ensure the slot is returned so it's cleaned up properly
```
jakubno:
```suggestion
		// Ensure the slot is received from chan so the chan is cleaned up properly
```

packages/orchestrator/internal/sandbox/sandbox_linux.go
dobrac: not really right now, we might though in the future when the disk resize would be live and we would like to properly track the dirty blocks

jakubno (approved):

jakubno (changes requested):
please add the env variable to job definition and terraform

jakubno (approved):

PR #697, opened 2025-05-26
Set reasonable DB connection limits

jakubno:
# Description

Set limits for number of connections to DB


packages/shared/pkg/db/client.go
```diff
@@ -3,10 +3,11 @@ package db
 import (
 	"fmt"
 	"os"
+	"time"
 
 	"entgo.io/ent/dialect"
 	"entgo.io/ent/dialect/sql"
-	_ "github.com/lib/pq"
+	_ "github.com/jackc/pgx/v5/stdlib"
```
dobrac: why this change?

dobrac (approved):

PR #699, opened 2025-05-26
Improve golangci-lint install

dobrac:
Improve golangci-lint install to check also the installed version. If it's missing, install it for you.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -215,9 +215,5 @@ connect-orchestrator:
 
 .PHONY: fmt
 fmt:
-	@command -v golangci-lint >/dev/null 2>&1 || { \
-		echo >&2 "golangci-lint not found, please install it first. Run following command please:\n"; \
-		echo >&2 'curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/HEAD/install.sh | sh -s -- -b $(go env GOPATH)/bin v2.1.6\n'; \
-		exit 1; \
-	}
+	@./scripts/golangci-lint-install.sh
 	golangci-lint fmt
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -215,9 +215,5 @@ connect-orchestrator:
 
 .PHONY: fmt
 fmt:
-	@command -v golangci-lint >/dev/null 2>&1 || { \
-		echo >&2 "golangci-lint not found, please install it first. Run following command please:\n"; \
-		echo >&2 'curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/HEAD/install.sh | sh -s -- -b $(go env GOPATH)/bin v2.1.6\n'; \
-		exit 1; \
-	}
+	@./scripts/golangci-lint-install.sh
```
jakubno: wouldn't it be better to pass the version as an argument from here?

scripts/golangci-lint-install.sh
```diff
@@ -0,0 +1,17 @@
...
+    "$BIN_PATH" --version 2>/dev/null | awk '{print $4}'
+}
+
+if ! command -v golangci-lint >/dev/null 2>&1 || [[ "$(get_linter_version)" != "$REQUIRED_VERSION" ]]; then
+    echo >&2 "golangci-lint not found or incorrect version. Installing v${REQUIRED_VERSION}..."
+    curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/HEAD/install.sh | sh -s -- -b "$(go env GOPATH)/bin" "v${REQUIRED_VERSION}" || {
+        echo >&2 "Installation failed."; exit 1;
+    }
+else
+    echo "golangci-lint v${REQUIRED_VERSION} is already installed."
```
jakubno: Do we want to `echo` something each time you run `make fmt`?

PR #702, opened 2025-05-26
Local network namespaces storage adjustments

sitole:
- Env based switch to run local namespace storage in non dev envs
- Forgotten err catch in rootfs overlay that can cause nil pointer panic


sitole:
I'am thinking about following peace of code. It will print debug log in case of foreign network namespace exists.
In environments like us production with dozens of re-deployments it can print hundreds of several thousand times, so I'm not sure it should stay.


```golang
// skip the slot if it's already in use by foreign program
if _, found := s.foreignNs[slotName]; found {
	zap.L().Debug("Skipping slot because already use by foreign program", zap.String("slot", slotName))
	continue
}
```

packages/orchestrator/internal/sandbox/sandbox_linux.go
```diff
@@ -280,6 +280,11 @@ func ResumeSandbox(
 		return nil, cleanup, fmt.Errorf("failed to create rootfs overlay: %w", err)
 	}
 
+	if err != nil {
```
dobrac: why is it here twice now?

sitole: Oh, @jakubno already fix this yesterday. Sorry, did not notice during rebase.

dobrac:
> I'am thinking about following peace of code. It will print debug log in case of foreign network namespace exists. In environments like us production with dozens of re-deployments it can print hundreds of several thousand times, so I'm not sure it should stay.
> 
> ```go
> // skip the slot if it's already in use by foreign program
> if _, found := s.foreignNs[slotName]; found {
> 	zap.L().Debug("Skipping slot because already use by foreign program", zap.String("slot", slotName))
> 	continue
> }
> ```

I’m okay with removing this as it should have been logged at least one time at this point.

PR #713, opened 2025-05-28
Network slot allocation with configurable CIDR and with incremental IPs

sitole:
- Changed default host and vrt newtork CIDRs to occupy less space
- Network CIDRs are now configurable for envs where sbx allocation can create collision with subnet/vpc IPs.
- Vrt (vpeer and veth) ips are now in /31 CIDR and they are allocated next to each other in vrt network CIDRs so there is removed wasting of last 8 bits.

Following IPs will be generated from corresponding slot idx. This mechanism allow us to occupy 16 bits corresponding to 65,536 addressed. Every veth needs to addresses so then it falls to 32,768 that can be allocated. Also we don't want to run into last subnet address because of broadcast so we are subtracting number of hosts per slot (2).

```
Math for 10.12.0.0/16 CIDR

Total IPs in subnet: 65536
Total slots in subnet 32768

First slot index: 1
First vEth IP: 10.12.0.2, vPeer IP: 10.12.0.3

Last slot index: 32766
Last vEth IP: 10.12.255.252, vPeer IP: 10.12.255.253
```

### DIFF
--- a/packages/orchestrator/internal/sandbox/network/network_linux.go
+++ b/packages/orchestrator/internal/sandbox/network/network_linux.go
@@ -193,11 +193,8 @@ func (s *Slot) CreateNetwork() error {
 
 	// Add routing from host to FC namespace
 	err = netlink.RouteAdd(&netlink.Route{
-		Gw: s.VpeerIP(),
-		Dst: &net.IPNet{
-			IP:   s.HostIP(),
-			Mask: s.HostMask(),
-		},
+		Gw:  s.VpeerIP(),
+		Dst: s.HostNet(),
 	})
 	if err != nil {
 		return fmt.Errorf("error adding route from host to FC: %w", err)
--- a/packages/orchestrator/internal/sandbox/network/slot.go
+++ b/packages/orchestrator/internal/sandbox/network/slot.go
@@ -21,8 +21,8 @@ const (
 	defaultVrtNetworkCIDR  = "10.12.0.0/16"
 
 	hostMask          = 32
-	vrtMask           = 31 // 2 usable ips per block (vpeer and veth)
-	vrtAddressPerSlot = 2  // vrt addresses per slot (vpeer and veth)
+	vrtMask           = 31                  // 2 usable ips per block (vpeer and veth)
+	vrtAddressPerSlot = 1 << (32 - vrtMask) // vrt addresses per slot (vpeer and veth)
 
 	tapMask          = 30
 	tapInterfaceName = "tap0"


dobrac:
Is there a reason why for slot 1 it's:
10.11.0.1/32 | 10.12.0.2/31 | 10.12.0.3/31 
(1, 2, 3)

but for the last slot its:
10.11.255.253/32 | 10.12.255.253/31 | 10.12.255.254/31
(253, 253, 254)

Also, why not use 10.12.0.1?

packages/orchestrator/internal/sandbox/network/slot.go
```diff
@@ -10,20 +10,50 @@ import (
...
 )
 
-// We are using a more debuggable IP address allocation for now that only covers 255 addresses.
 const (
-	octetSize = 256
-	octetMax  = octetSize - 1
 	// This is the maximum number of IP addresses that can be allocated.
-	slotsSize = octetSize * octetSize
+	// Last address is reserved for broadcast, we need to subtract with, second address reserved for Veth.
+	slotsSize = (256 * 256 / vrtAddressPerSlot) - vrtAddressPerSlot
```
dobrac: nit: can we compute the size from the mask?

packages/orchestrator/internal/sandbox/network/slot.go
```diff
@@ -44,55 +74,47 @@ func (s *Slot) VpeerName() string {
...
+	if err != nil {
+		zap.L().Error("Failed to get indexed IP", zap.Int("index", s.Idx), zap.Error(err))
+	}
+	return ip
 }
 
-func (s *Slot) VMask() int {
-	return vMask
+func (s *Slot) VethIP() net.IP {
+	ip, err := netutils.GetIndexedIP(vrtNetworkCIDR, s.Idx*vrtAddressPerSlot+1)
```
dobrac: nit: didn't you switch the vpeer with veth in terms of 1st, 2nd? (I guess it doesn't matter that much though)

packages/orchestrator/internal/sandbox/network/slot.go
```diff
@@ -44,55 +74,47 @@ func (s *Slot) VpeerName() string {
...
+	if err != nil {
+		zap.L().Error("Failed to get indexed IP", zap.Int("index", s.Idx), zap.Error(err))
+	}
+	return ip
```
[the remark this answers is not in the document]
dobrac: is it possible to propagate the error correctly upstream, or would that require too much work?

(same applies to all others)

packages/orchestrator/internal/sandbox/network/network_linux.go
```diff
@@ -69,15 +69,16 @@ func (s *Slot) CreateNetwork() error {
 		return fmt.Errorf("error setting vpeer device up: %w", err)
 	}
 
-	ip, ipNet, err := net.ParseCIDR(s.VpeerCIDR())
+	vPeerIp := s.VpeerIP()
+	_, vrtIpNet, err := net.ParseCIDR(s.VrtCIDR())
```
dobrac: Why having this split and not returning the `net.IPNet` from VpeerIP directly?

The following line is in this function a bit too many times + it looks like a potential future place for a mistake (as the IP and Mask are not together):
`_, vrtIpNet, err := net.ParseCIDR(s.VrtCIDR())`

(same applies to the veth)

dobrac (changes requested):
Looks good! Just few nits

packages/orchestrator/internal/sandbox/network/network_linux.go
sitole: Done

packages/orchestrator/internal/sandbox/network/slot.go
sitole: true, should be okay, but I move it back so its more consistent.

packages/orchestrator/internal/sandbox/network/slot.go
sitole: done

dobrac:
> > Only last IP is allocated to subnet broadcast so we can allocate 255.254 too.
>

yes and no, as the mask 31 for 252/253 includes 252 + 253, but there is no mask to include only 254 and not 255

packages/orchestrator/internal/sandbox/network/network_linux.go
```diff
@@ -207,14 +192,12 @@ func (s *Slot) CreateNetwork() error {
...
-	_, ipNet, err = net.ParseCIDR(s.HostCIDR())
-	if err != nil {
-		return fmt.Errorf("error parsing host snapshot CIDR: %w", err)
-	}
-
 	err = netlink.RouteAdd(&netlink.Route{
-		Gw:  net.ParseIP(s.VpeerIP()),
-		Dst: ipNet,
+		Gw: s.VpeerIP(),
+		Dst: &net.IPNet{
```
dobrac: use HostNet instead

packages/orchestrator/internal/sandbox/network/slot.go
```diff
@@ -21,8 +21,8 @@ const (
 	defaultVrtNetworkCIDR  = "10.12.0.0/16"
 
 	hostMask          = 32
-	vrtMask           = 31 // 2 usable Ips per block (Vpeer and Veth ips)
-	vrtAddressPerSlot = 1 << (32 - vrtMask)
+	vrtMask           = 31 // 2 usable ips per block (vpeer and veth)
+	vrtAddressPerSlot = 2  // vrt addresses per slot (vpeer and veth)
```
dobrac: why not have this calculated anymore?

sitole: can be there, but seem liter over kill imho

dobrac: ok

dobrac (approved):

PR #715, opened 2025-05-28
Support different container registry for template manager

sitole:
- Added adapter for AWS ECR and local Docker repository backends. Default repository is set to GCP Artifacts so there no need for any configuration changes in current deployments.
- Different providers can be selected with `ARTEFACTS_REGISTRY_PROVIDER` env variable. For AWS ECR there must be machine setup that supports fetching AWS credentials.
- Fixed issue in template manager delete endpoint that silently fall into error during template docker delete, because of mismatch between templateId and buildId.
- Integration tests are now using local Docker registry instead of GCP


.github/actions/build-sandbox-template/action.yml
```diff
@@ -9,36 +9,22 @@ runs:
...
-        echo "Docker image template path: ${DOCKER_IMAGE_PATH}"
-        
-        gcloud auth configure-docker "${GCP_REGION}-docker.pkg.dev" --quiet
-        docker pull "${DOCKER_IMAGE_PATH}:98130b63-ca75-431d-956a-9d829d14d317"
-        docker tag "${DOCKER_IMAGE_PATH}:98130b63-ca75-431d-956a-9d829d14d317" "${DOCKER_IMAGE_PATH}:${BUILD_ID}"
-        docker push "${DOCKER_IMAGE_PATH}:${BUILD_ID}"
-        
+
         make -C packages/orchestrator build-template \
+          ARTEFACTS_REGISTRY_PROVIDER=Local \
```
dobrac: nit: It's `ARTIFACTS`, not ARTEFACTS (same everywhere else)

packages/orchestrator/cmd/build-template/main.go
```diff
@@ -116,6 +117,11 @@ func buildTemplate(parentCtx context.Context, kernelVersion, fcVersion, template
 		}
 	}()
 
+	artefactsRegistry, err := artefactsregistry.GetArtefactsRegistryProvider()
```
dobrac: artifacts

packages/orchestrator/internal/template/build/build.go
```diff
@@ -26,14 +27,15 @@ func Build(
...
+	artifactRegistry artefactsregistry.ArtefactsRegistry,
 	templateBuildDir string,
 	rootfsPath string,
 ) (r *block.Local, m *block.Local, e error) {
 	childCtx, childSpan := tracer.Start(ctx, "template-build")
 	defer childSpan.End()
 
 	// Create a rootfs file
-	rtfs := NewRootfs(templateConfig)
+	rtfs := NewRootfs(templateConfig, artifactRegistry)
```
dobrac: nit: I would switch the order of args: 1st dependencies, 2nd function specific

packages/api/internal/handlers/template_delete.go
```diff
@@ -120,9 +120,12 @@ func (a *APIStore) DeleteTemplatesTemplateID(c *gin.Context, aliasOrTemplateID a
 	}
 
 	// get all build ids
-	buildIds := make([]uuid.UUID, len(template.Edges.Builds))
+	buildIds := make([]template_manager.DeleteBuild, len(template.Edges.Builds))
 	for i, build := range template.Edges.Builds {
-		buildIds[i] = build.ID
+		buildIds[i] = template_manager.DeleteBuild{
+			BuildID:    build.ID,
+			TemplateId: *build.EnvID,
```
dobrac, line 127: why do we need not the template ID?

packages/shared/pkg/artefacts-registry/registry.go
```diff
@@ -0,0 +1,52 @@
...
+
+	storageProviderEnv = "ARTEFACTS_REGISTRY_PROVIDER"
+)
+
+var (
+	ErrImageNotExists = errors.New("image does not exist")
+)
+
+type ArtefactsRegistry interface {
+	GetUrl(ctx context.Context, templateId string, buildId string) (string, error)
```
dobrac: this should be GetTag

packages/shared/pkg/artefacts-registry/registry_aws.go
```diff
@@ -0,0 +1,136 @@
...
+		client:         client,
+	}, nil
+}
+
+func (g *AWSArtefactsRegistry) Delete(ctx context.Context, templateId string, buildId string) error {
+	imageIds := []types.ImageIdentifier{
+		{ImageTag: &buildId},
+	}
+
+	// for AWS implementation we are using only build id as image tag
```
dobrac: why have different implementation for the AWS in terms of not using template ID?

packages/shared/pkg/artefacts-registry/registry_gcp.go
```diff
@@ -0,0 +1,116 @@
...
+		return nil, fmt.Errorf("error creating artifact registry client: %v", err)
+	}
+
+	return &GCPArtefactsRegistry{registry: registry}, nil
+}
+
+func (g *GCPArtefactsRegistry) Delete(ctx context.Context, templateId string, buildId string) error {
+	packageName := g.getDockerImagePath(templateId)
+
+	op, err := g.registry.DeletePackage(ctx, &artifactregistrypb.DeletePackageRequest{Name: packageName})
```
dobrac: this deletes the whole template, not only the build?

dobrac (changes requested):
LGTM, just please rename all `artEfact` to `artIfact` 🙏

.github/actions/build-sandbox-template/action.yml
```diff
@@ -9,36 +9,22 @@ runs:
         TEMPLATE_ID: "2j6ly824owf4awgai1xo"
         KERNEL_VERSION: "vmlinux-6.1.102"
         FIRECRACKER_VERSION: "v1.10.1_1fcdaec"
-        # Rootfs Docker image repository
-        GCP_DOCKER_REPOSITORY_NAME: "custom-environments-tests"
-        # Not used, but required to be defined
-        GOOGLE_SERVICE_ACCOUNT_BASE64: "required-but-not-needed"
```
dobrac, line 15: nice 👍

packages/shared/pkg/artefacts-registry/registry_aws.go
sitole: AFIK template id is used only in gcp because we are creating repository (template id) and then using build id as tags. In AWS we have one shared repository (you cannot just push to repository and it will be created such as in gcp) and we are using tags that are build id. We can use template id + build id but will not help us with anything. Filtration will be slow in point where there will be thousands of tags.

packages/shared/pkg/artefacts-registry/registry_gcp.go
sitole: Looks like current implementation of deleting never worked, because there was request to delete package that is named as build id, but in reality packages are named with template id (https://github.com/e2b-dev/infra/blob/0a44d0a169d4e136af791c94f59690d3bfd2260a/packages/orchestrator/internal/template/template/main.go#L38).

Will re-work so we are removing only corresponding tag and not whole package.

packages/orchestrator/main.go
```diff
@@ -42,7 +42,7 @@ type Closeable interface {
 const (
 	defaultPort      = 5008
 	defaultProxyPort = 5007
-	defaultWait      = 30
+	defaultWait      = 0
```
dobrac: not part of this PR

packages/shared/pkg/artefacts-registry/registry_aws.go
dobrac: Will this change be compatible with docker reverse proxy when used with AWS then?

dobrac (changes requested):

packages/shared/pkg/artefacts-registry/registry_aws.go
sitole: Should work, ideally image tag should be returned from api and not computed locally. This will help us to support byoc (where docker is pushed directly to repository and not via our proxy).

dobrac (approved):

PR #719, opened 2025-05-29
Add option for a template ready check

dobrac:
Add option for a template ready check. You can specify `ready_cmd` which will be run repeatedly (with 2 second delay after exit) until it exits with 0. The template will be then marked as ready. Start command and ready command are run simultaneously.

If no ready command is defined, `sleep 20` is assumed when the start command is present, and `sleep 0` when none is defined. It is also possible to define ready command without any start command. This might be helpful if the filesystem contains self-booted services (e.g. using systemd).

This adds possibility to configure the default 20 seconds wait time for a start command.


packages/orchestrator/internal/template/build/ready_command.go
```diff
@@ -0,0 +1,97 @@
...
+
+	"github.com/e2b-dev/infra/packages/orchestrator/internal/template/build/writer"
+	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
+)
+
+const (
+	defaultReadyWait = 20 * time.Second
+
+	readyCommandRetryInterval = 2 * time.Second
+	readyCommandTimeout       = 5 * time.Minute
```
dobrac, line 16: maximum ready command time

packages/orchestrator/internal/template/build/ready_command.go
```diff
@@ -0,0 +1,97 @@
...
+			"ready",
+			sandboxID,
+			template.ReadyCmd,
+			"root",
+			&cwd,
+		)
+
+		if err == nil {
+			// Template is ready
+			cancel()
```
jakubno: do you need to call cancel here?

packages/orchestrator/internal/template/build/template_builder.go
```diff
@@ -238,46 +238,68 @@ func (b *TemplateBuilder) Build(ctx context.Context, template *TemplateConfig) (
...
+				ctx,
+				postProcessor,
+				"start",
+				sbx.Metadata.Config.SandboxId,
+				template.StartCmd,
+				"root",
+				&cwd,
+			)
+			// If the ctx is canceled, the ready command succeeded and no start command await is necessary.
+			if err != nil && !errors.Is(err, context.Canceled) {
```
jakubno, line 278: Why are you checking for errors.Is(err, context.Canceled)? You can safely call readyCancel even if the context is already cancelled

packages/orchestrator/internal/template/build/ready_command.go
```diff
@@ -0,0 +1,97 @@
...
+	ctx context.Context,
+	postProcessor *writer.PostProcessor,
+	template *TemplateConfig,
+	sandboxID string,
+) error {
+	startTime := time.Now()
+	defer func() {
+		telemetry.ReportEvent(ctx, "waited for template ready", attribute.Float64("seconds", time.Since(startTime).Seconds()))
+	}()
+
```
jakubno: wouldn't this be nicer as a span?

packages/orchestrator/internal/template/build/ready_command.go
```diff
@@ -0,0 +1,97 @@
...
+			}
+			// Template is ready, the start command finished before the ready command
+			break wait
+		case <-time.After(readyCommandRetryInterval):
+			// Wait for readyCommandRetryInterval time before retrying the ready command
+		}
+	}
+
+	// Cancel the context for the ready command if it is still running
+	cancel()
```
jakubno: this isn't needed either, you are already calling it in defer

packages/orchestrator/internal/template/build/template_builder.go
```diff
@@ -238,46 +238,68 @@ func (b *TemplateBuilder) Build(ctx context.Context, template *TemplateConfig) (
...
+		template,
+		sbx.Metadata.Config.SandboxId,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("error running ready command: %w", err)
+	}
+
+	// Cancel the start command context (it's running in the background anyway).
+	// If it has already finished, check the error.
+	readyCancel()
```
jakubno: you are cancelling incorrect context

packages/orchestrator/internal/template/build/template_builder.go
dobrac: I don't want to consider an error when the context is canceled (and returning error, failing the build)

packages/orchestrator/internal/template/build/ready_command.go
dobrac: not anymore

packages/orchestrator/internal/template/build/template_builder.go
dobrac: I've passed incorrect context to the start command. This cancel should be correct

packages/orchestrator/internal/template/build/template_builder.go
dobrac: Also, I don't want early finish of the start command to bypass the ready command, as the ready command doesn't need to depend only on the start command

packages/orchestrator/internal/template/build/command.go
```diff
@@ -59,15 +59,12 @@ func (b *TemplateBuilder) runCommand(
 		createAppStream.Close()
 	}()
 
-	cmdCtx, cancel := context.WithTimeout(ctx, cmdWait)
-	defer cancel()
-
-	msgCh, msgErrCh := grpc.StreamToChannel(cmdCtx, createAppStream)
+	msgCh, msgErrCh := grpc.StreamToChannel(ctx, createAppStream)
```
jakubno: nit: can you rename the variable, i think it was copied from a test

packages/orchestrator/internal/template/build/ready_command.go
```diff
@@ -55,30 +49,24 @@ wait:
...
+			return nil
 		} else {
 			postProcessor.WriteMsg(fmt.Sprintf("Template is not ready: %v", err))
 		}
 
 		select {
 		case <-ctx.Done():
 			if errors.Is(ctx.Err(), context.DeadlineExceeded) {
-				postProcessor.WriteMsg(fmt.Sprintf("Ready command timed out, exceeding %s", readyCommandTimeout))
+				return fmt.Errorf("ready command timed out after %s: %w", readyCommandTimeout, ctx.Err())
```
jakubno: Is there a reason why adding `ctx.Err()`? It will always be `context deadline exceeded`. Also, if the parent context would time out, this may not be necessarily true, but I think it's okay to leave it like this.

packages/orchestrator/internal/template/build/template_builder.go
```diff
@@ -296,7 +296,7 @@ func (b *TemplateBuilder) Build(ctx context.Context, template *TemplateConfig) (
 
 	// Cancel the start command context (it's running in the background anyway).
 	// If it has already finished, check the error.
-	readyCancel()
+	commandsCancel()
```
jakubno, line 310: there's a race condition, you need to make sure that the request for the start command was sent

jakubno (changes requested):

packages/orchestrator/internal/template/build/ready_command.go
dobrac: removed the err and used actually spent time

packages/orchestrator/internal/template/build/template_builder.go
dobrac: fixed

packages/orchestrator/internal/template/build/command.go
```diff
@@ -50,16 +73,18 @@ func (b *TemplateBuilder) runCommand(
 
 	processCtx, processCancel := context.WithCancel(ctx)
 	defer processCancel()
-	createAppStream, err := processC.Start(processCtx, createAppReq)
+	commandStream, err := processC.Start(processCtx, createAppReq)
```
jakubno:
```suggestion
	commandStream, err := processC.Start(processCtx, commandReq)
```

packages/orchestrator/internal/template/build/ready_command.go
jakubno: Is there a reason why adding ctx.Err()? It will always be `context deadline exceeded`

dobrac: I've removed the ctx.Err, maybe outdated code state?

PR #722, opened 2025-05-29
Implement logic for sending metrics to Clickhouse

jakubno:
# Description

- Implement simple Clickhouse client
- Send metrics from orchestrator in batches to Clickhouse


packages/clickhouse/pkg/metrics.go
```diff
@@ -0,0 +1,134 @@
...
+    SELECT *
+    FROM metrics
+    WHERE sandbox_id = ?
+      AND team_id    = ?
+      AND timestamp  >= ?
+    ORDER BY timestamp ASC
+    LIMIT ?`
+
+// InsertMetrics queues the record for the background goroutine.
+func (c *Client) InsertMetrics(_ context.Context, m Metrics) error {
```
dobrac, line 36: is there a reason why not use OTEL collector and just export it to ClickHouse?

packages/clickhouse/pkg/metrics.go
```diff
@@ -0,0 +1,134 @@
...
+		if err := rows.ScanStruct(&m); err != nil {
+			return nil, fmt.Errorf("error scaning metrics: %w", err)
+		}
+		out = append(out, m)
+	}
+
+	return out, rows.Err()
+}
+
+func (c *Client) metricsBatchLoop() {
```
dobrac, line 65: would it make sense to pass context here, so you can then use it instead of `context.Background()`?

packages/clickhouse/pkg/mock.go
```diff
@@ -0,0 +1,24 @@
+package clickhouse
+
+import (
+	"context"
+	"time"
+)
+
+type MockClient struct{}
```
dobrac, line 8: I think this is more of a noop client than mock, as it's used outside of tests

packages/nomad/api.hcl
```diff
@@ -70,12 +70,9 @@ job "api" {
       env {
         ORCHESTRATOR_PORT              = "${orchestrator_port}"
         TEMPLATE_MANAGER_ADDRESS       = "${template_manager_address}"
+        CLICKHOUSE_CONNECTION_STRING   = "${clickhouse_connection_string}"
```
dobrac, line 73: don't you need to set also "WRITE_CLICKHOUSE_METRICS"?

jakubno: api doesn't write to clickhouse and reading will be in separate PR

dobrac: I see, so `READ_METRICS_FROM_CLICKHOUSE` usage will be in a next PR?

PR #728, opened 2025-06-02
Delete snapshots in the background

jakubno:
# Description

Delete the snapshots in background, so it isn't blocking kill
Also we don't want to return an error for kill if there's some issue with deleting snapshots

### DIFF
--- a/packages/api/internal/handlers/sandbox_kill.go
+++ b/packages/api/internal/handlers/sandbox_kill.go
@@ -9,6 +9,7 @@ import (
 	"github.com/gin-gonic/gin"
 	"github.com/google/uuid"
 	"go.opentelemetry.io/otel/attribute"
+	"go.uber.org/zap"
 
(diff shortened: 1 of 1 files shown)


packages/api/internal/handlers/sandbox_kill.go
```diff
@@ -28,26 +29,39 @@ func (a *APIStore) deleteSnapshot(
...
+		}
+
+		if len(envBuildIDs) == 0 {
+			return
+		}
+
+		deleteJobErr := a.templateManager.DeleteBuilds(deleteCtx, envBuildIDs)
+		if deleteJobErr != nil {
+			zap.L().Warn("Error deleting snapshot builds", zap.Error(deleteJobErr), zap.String("sandboxID", sandboxID))
+			telemetry.ReportError(ctx, deleteJobErr)
```
dobrac: do we want to report the error to the parent ctx or deleteCtx?

PR #729, opened 2025-06-02
Add ability to cancel running build

dobrac:
Add ability to cancel running build. This change will cancel any already running build for the same template ID on a new build trigger. This should allow faster iterations when you need to retry the build before it has completed (failed or succeeded).

Includes also:
- Improved error handling for postprocessing
- Fix OTEL running builds counter


packages/orchestrator/internal/template/cache/build_cache.go
```diff
@@ -61,63 +63,93 @@ func (b *BuildInfo) GetStatus() template_manager.TemplateBuildState {
...
-	counter metric.Int64UpDownCounter
+	cache *ttlcache.Cache[string, *BuildInfo]
 
 	mu sync.Mutex
 }
 
 func NewBuildCache() *BuildCache {
 	cache := ttlcache.New(ttlcache.WithTTL[string, *BuildInfo](buildInfoExpiration))
-	counter, err := meters.GetUpDownCounter(meters.BuildCounterMeterName)
+	_, err := meters.GetObservableUpDownCounter(meters.BuildCounterMeterName, func(ctx context.Context, observer metric.Int64Observer) error {
```
dobrac, line 80: the previous metric was not working at all, so I've reused the name for proper implementation

packages/api/internal/handlers/template_start_build.go
```diff
@@ -98,19 +100,33 @@ func (a *APIStore) PostTemplatesTemplateIDBuildsBuildID(c *gin.Context, template
 			envbuild.StatusIn(envbuild.StatusWaiting, envbuild.StatusBuilding),
 			envbuild.IDNotIn(buildUUID),
 		).
-		Count(ctx)
+		All(ctx)
 	if err != nil {
 		zap.L().Error("Error when getting count of running builds", zap.Error(err))
```
jakubno:
```suggestion
		zap.L().Error("Error when getting running builds", zap.Error(err))
```

packages/orchestrator/internal/template/cache/build_cache.go
```diff
@@ -61,63 +63,97 @@ func (b *BuildInfo) GetStatus() template_manager.TemplateBuildState {
...
+
+	return b.ctx
+}
+
+func (b *BuildInfo) Cancel() {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+
+	b.ctxCancel()
+}
```
jakubno: Do the locks server any purpose here?

packages/orchestrator/internal/template/cache/build_cache.go
jakubno, line 1: I think we can get rid of the templateID(envID) completely, the only place it's used is in delete, but there's already buildID required

packages/shared/pkg/utils/filter.go
```diff
@@ -0,0 +1,11 @@
+package utils
+
+func Filter[T any](input []T, f func(T) bool) []T {
```
jakubno, line 6: could you add comment, at least state true means include 🙏🏻

packages/shared/pkg/utils/map.go
```diff
@@ -0,0 +1,9 @@
+package utils
+
+func Map[T any, U any](input []T, f func(T) U) []U {
```
jakubno, line 5: also here, I think `Map` is kinda overloaded word. Just we don't have to scan the code

jakubno (changes requested):

packages/orchestrator/internal/template/cache/build_cache.go
dobrac: no idea, but they were there in all other getters, so I've kept the logic

packages/shared/pkg/utils/filter.go
dobrac: sure, I've just added helpers that are standard in most high level languages

PR #737, opened 2025-06-05
Report only long-running sandboxes to prevent unnecessary updates

jakubno:
# Description

- Report only long running sandboxes (running for more than 30 mins to prevent spamming analytics)
- Report the execution to distinguish between repeatedly paused sandboxes

### DIFF
--- a/packages/api/internal/orchestrator/analytics.go
+++ b/packages/api/internal/orchestrator/analytics.go
@@ -27,8 +27,6 @@ func (o *Orchestrator) reportLongRunningSandboxes(ctx context.Context) {
 			zap.L().Info("Stopping node analytics reporting due to context cancellation")
 			return
 		case <-ticker.C:
-			childCtx, cancel := context.WithTimeout(ctx, syncAnalyticsTime)
-
 			sandboxes := o.instanceCache.Items()
 			longRunningSandboxes := make([]*instance.InstanceInfo, 0, len(sandboxes))
 			for _, sandbox := range sandboxes {
@@ -37,8 +35,7 @@ func (o *Orchestrator) reportLongRunningSandboxes(ctx context.Context) {
 				}
 			}
 
-			sendAnalyticsForLongRunningSandboxes(childCtx, o.analytics, longRunningSandboxes)
-			cancel()
+			sendAnalyticsForLongRunningSandboxes(ctx, o.analytics, longRunningSandboxes)
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


packages/api/internal/orchestrator/analytics.go
```diff
@@ -14,56 +13,46 @@ import (
...
+
+			sandboxes := o.instanceCache.Items()
+			longRunningSandboxes := make([]*instance.InstanceInfo, 0, len(sandboxes))
+			for _, sandbox := range sandboxes {
+				if time.Since(sandbox.StartTime) > oldSandboxThreshold {
+					longRunningSandboxes = append(longRunningSandboxes, sandbox)
+				}
+			}
+
+			sendAnalyticsForLongRunningSandboxes(ctx, o.analytics, longRunningSandboxes)
```
0div, line 38: should we send analytics if `longRunningSandboxes` is empty?

packages/api/internal/orchestrator/analytics.go
```diff
@@ -14,56 +13,46 @@ import (
...
 
 	for {
 		select {
-		case <-ctx.Done():
+		case <-parentCtx.Done():
+			zap.L().Info("Stopping node analytics reporting due to context cancellation")
 			return
 		case <-ticker.C:
-			o.reportNodesAnalytics(ctx)
+			ctx, cancel := context.WithTimeout(parentCtx, syncAnalyticsTime)
```
0div: nit: for readability i'd keep the `ctx` func param name convention and name this one `childCtx`

packages/api/internal/orchestrator/analytics.go
```diff
@@ -14,56 +13,46 @@ import (
...
+			sandboxes := o.instanceCache.Items()
+			longRunningSandboxes := make([]*instance.InstanceInfo, 0, len(sandboxes))
+			for _, sandbox := range sandboxes {
+				if time.Since(sandbox.StartTime) > oldSandboxThreshold {
+					longRunningSandboxes = append(longRunningSandboxes, sandbox)
+				}
+			}
+
+			sendAnalyticsForLongRunningSandboxes(ctx, o.analytics, longRunningSandboxes)
+			cancel()
```
0div: i'd prob `defer` the `cancel` inside `sendAnalyticsForLongRunningSandboxes` for a sure way of cleaning up the timers and goroutines no matter what happens.

0div (approved):

packages/api/internal/orchestrator/analytics.go
jakubno: Moved to the function

packages/api/internal/orchestrator/analytics.go
jakubno: added handling of this case

PR #740, opened 2025-06-06
Optimize team_api_keys schema

dobrac:
The planned team_api_keys schema currently has one column keyMask, which holds information about a keyPrefix, the key length and the last 4 chars of the key. It also contains n obscured chars with *, which wastes database disk space and is unideal for future developments for api keys.

Hence we can optimize the schema as follows:
- `*_prefix` - identifies the key type (currently `e2b_`)
- `*_length` - string length of the token
- `*_mask_prefix` - prefix to identify masked tokens
- `*_mask_suffix` - suffix to identify masked tokens


0div:
What's the reason we need to store this information about the keys and tokens? If leaking them is a concern why not hash them like we would passwords? also wrt to saving space it doesn't seem significant? 

```
>>> select count(id) from team_api_keys;
64486

>>> select count(api_key_mask) from team_api_keys;
3120
```

0div:
What's the reason we need to store this information about the keys and tokens? If leaking them is a concern why not hash them like we would passwords? also wrt to saving space it doesn't seem significant?

dobrac:
We're moving to hashing them, so we want to store the masked values for users to be able to reference them in the dashboard

> also wrt to saving space it doesn't seem significant

yeah, probably not that important, but we were for some time saving like this "e2b\_\*\*\*\*\*\*\*\*\*\*\*\*\*\*\*\*\*\*\*\*\*\*\*\*\*\*\*\*\*\*\*\*\*\*\*\*6240"

packages/db/migrations/20250606204750_optimize_hashed_key_schema.sql
jakubno, line 1: Could you also handle the data, this should be possible to do in both directions

Why are there different types in access token and in team API keys?

dobrac: Do you mean to do the data migration? I'll do that in the following PR for all existing keys, when this one is deployed. (So I can be sure all new incoming data will populate the columns)

The different types are there only for the parity with the previous ones, I could optimize all of them to be e.g. varchar(5) only, as the prefix + suffix have fixed lengths

dobrac: Updated the string types in DB

packages/api/internal/handlers/accesstoken.go
```diff
@@ -56,24 +58,14 @@ func (a *APIStore) PostAccessTokens(c *gin.Context) {
...
-
 	c.JSON(http.StatusCreated, api.CreatedAccessToken{
 		Id:    accessTokenDB.ID,
 		Token: accessToken.PrefixedRawValue,
 		Mask: api.IdentifierMaskingDetails{
-			Prefix:            maskedToken.Prefix,
-			ValueLength:       maskedToken.ValueLength,
-			MaskedValuePrefix: maskedToken.MaskedValuePrefix,
-			MaskedValueSuffix: maskedToken.MaskedValueSuffix,
+			Prefix:            accessTokenDB.AccessTokenPrefix,
```
jakubno, line 65: nit: do you need the version from DB, you could use .Exec() and `accessToken` to prevent extra database call

jakubno (approved):
Can you add task for removing masked value?

packages/api/internal/handlers/accesstoken.go
dobrac: would this be ok then?
`CreatedAt: time.Now()`

jakubno: Actually, it doesn't do any new query, sorry for misinformation

PR #745, opened 2025-06-07
Added fields to envd stat method response

mishushakov:
- Added fields such as size, mode, permissions, owner, group, modified time to file stat and list response
- Added test for stat method

This is an updated PR in favour of https://github.com/e2b-dev/infra/pull/655 which was reversed due to breaking Python SDK


packages/envd/internal/services/filesystem/utils.go
```diff
@@ -15,3 +22,50 @@ func getEntryType(entry osEntry) rpc.FileType {
...
+	owner, group := getFileOwnership(fileInfo)
+	fileMode := fileInfo.Mode()
+
+	entry := &rpc.EntryInfo{
+		Name: fileInfo.Name(),
+		Type: getEntryType(fileInfo),
+		Path: path,
+	}
+
+	if sdkVersion != "" && sdkVersion >= "1.5.2" {
```
dobrac: this is incorrect comparison of versions. Here is the util implementation to handle it in the API/Orchestrator:
```
func IsGTEVersion(curVersion, minVersion string) bool {
	if len(curVersion) > 0 && curVersion[0] != 'v' {
		curVersion = "v" + curVersion
	}

	if !semver.IsValid(curVersion) {
		return false
	}

	return semver.Compare(curVersion, minVersion) >= 0
}
```

jakubno: why are we checking only version? The problem is python specific. Isn't the first check useless? Also `1.5.10` is less than `1.5.2` when compared as strings. Add comment with explanation

mishushakov:
I have changed the comparison logic: it will include all fields by default, expect when sdkVersion is provided: python and higher than or equal to supported version

PR #752, opened 2025-06-11
Add CICD workflow for generated files

jakubno:
# Description

Check if all generated files are up today


.github/workflows/generated-files.yml
```diff
@@ -0,0 +1,47 @@
...
+          export PATH=$PATH:/usr/local/bin
+          
+          # Install protoc-gen-go
+          go install google.golang.org/protobuf/cmd/protoc-gen-go@v${PROTOC_GEN_GO_VER}
+          
+          # Install protoc-gen-go-grpc
+          go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@v${PROTOC_GEN_GO_GRPC_VER}
+          
+          # Install protoc-gen-connect-go 
+          go install connectrpc.com/connect/cmd/protoc-gen-connect-go@${PROTOC_GEN_CONNECT_GO_VER}
```
dobrac, line 36: All of these should be installed as tools, do we need to install them manually?

dobrac (changes requested):
Nice! Small comment to the installed versions

.github/workflows/generated-files.yml
jakubno: it's used in this command 
```
@protoc --go_out=../shared/pkg/grpc/orchestrator/ --go_opt=paths=source_relative --go-grpc_out=../shared/pkg/grpc/orchestrator/ --go-grpc_opt=paths=source_relative orchestrator.proto
```

and I didn't find `protoc`  would support `go tools` (yet?)

dobrac: I think protoc doesn't, but the cmds somehow might. If not at all, I would move all of this to a Docker container, so we have parity between dev and this CI/CD

ValentaTomas: I think @0div created a command that does the generation inside a Docker already.

ValentaTomas:
I left a comment, but otherwise looks good.

.github/workflows/generated-files.yml
dobrac: I'm not sure, but I think it was in the SDKs only

dobrac: (yes, the Docker is SDK only)

dobrac: you may now use the docker generators

PR #755, opened 2025-06-11
Allow customizing some resources based on the config

jakubno:
# Description

Allow configuring some of the resources based on the config


dobrac (approved):
Nice 🫶

PR #757, opened 2025-06-11
Use otel collector for collecting sandbox metrics

jakubno:
# Description

Use Otel Collector for sending the metrics to clickhouse

### DIFF
--- a/packages/clickhouse/migrations/20250521131546_add_metrics.sql
+++ b/packages/clickhouse/migrations/20250521131546_add_metrics.sql
@@ -1,0 +1,48 @@
+-- +goose Up
+-- +goose StatementBegin
+CREATE TABLE metrics_gauge_local(
+    ResourceAttributes    Map(LowCardinality(String), String) CODEC (ZSTD(1)),
+    ResourceSchemaUrl     String CODEC (ZSTD(1)),
+    ScopeName             String CODEC (ZSTD(1)),
+    ScopeVersion          String CODEC (ZSTD(1)),
+    ScopeAttributes       Map(LowCardinality(String), String) CODEC (ZSTD(1)),
+    ScopeDroppedAttrCount UInt32 CODEC (ZSTD(1)),
+    ScopeSchemaUrl        String CODEC (ZSTD(1)),
+    ServiceName           LowCardinality(String) CODEC (ZSTD(1)),
+    MetricName            LowCardinality(String) CODEC (ZSTD(1)),
+    MetricDescription     String CODEC (ZSTD(1)),
+    MetricUnit            String CODEC (ZSTD(1)),
+    Attributes            Map(LowCardinality(String), String) CODEC (ZSTD(1)),
+    StartTimeUnix         DateTime64(9) CODEC (Delta, ZSTD(1)),
+    TimeUnix              DateTime64(9) CODEC (Delta, ZSTD(1)),
+    Value                 Float64 CODEC (ZSTD(1)),
+    Flags                 UInt32 CODEC (ZSTD(1)),
+    Exemplars             Nested(
+        FilteredAttributes Map(LowCardinality(String), String),
+        TimeUnix              DateTime64(9),
+        Value                 Float64,
+        SpanId                String,
+        TraceId String) CODEC (ZSTD(1)
+    ),
+    INDEX                 idx_res_attr_key mapKeys(ResourceAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
+    INDEX                 idx_res_attr_value mapValues(ResourceAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
+    INDEX                 idx_scope_attr_key mapKeys(ScopeAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
+    INDEX                 idx_scope_attr_value mapValues(ScopeAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
+    INDEX                 idx_attr_key mapKeys(Attributes) TYPE bloom_filter(0.01) GRANULARITY 1,
+    INDEX                 idx_attr_value mapValues(Attributes) TYPE bloom_filter(0.01) GRANULARITY 1
+) ENGINE = MergeTree
+    PARTITION BY toDate(TimeUnix)
+    ORDER BY (ServiceName, MetricName, Attributes, toUnixTimestamp64Nano(TimeUnix))
+    TTL toDateTime(TimeUnix) + INTERVAL 7 DAY
+    SETTINGS index_granularity = 8192, ttl_only_drop_parts = 1;
+
+CREATE TABLE metrics_gauge as metrics_gauge_local
+    ENGINE = Distributed('cluster', currentDatabase(), 'metrics_gauge_local', xxHash64(Attributes));
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
+DROP TABLE IF EXISTS metrics_gauge;
+DROP TABLE IF EXISTS metrics_gauge_local;
+-- +goose StatementEnd
+
--- a/packages/nomad/main.tf
+++ b/packages/nomad/main.tf
@@ -204,6 +204,12 @@ resource "nomad_job" "otel_collector" {
     grafana_username             = data.google_secret_manager_secret_version.grafana_username.secret_data
     consul_token                 = var.consul_acl_token_secret
 
+    clickhouse_username = var.clickhouse_username
+    clickhouse_password = random_password.clickhouse_password.result
+    clickhouse_port     = var.clickhouse_server_port.port
+    clickhouse_host     = "clickhouse.service.consul"
+    clickhouse_database = var.clickhouse_database
+
     gcp_zone = var.gcp_zone
   })
 }
@@ -326,10 +332,8 @@ locals {
     otel_tracing_print           = var.otel_tracing_print
     template_bucket_name         = var.template_bucket_name
     otel_collector_grpc_endpoint = "localhost:4317"
-    clickhouse_connection_string = "clickhouse.service.consul:9000"
-    clickhouse_username          = var.clickhouse_username
-    clickhouse_password          = random_password.clickhouse_password.result
-    clickhouse_database          = var.clickhouse_database
+    write_to_loki                = var.write_loki_metrics
+    write_to_clickhouse          = var.environment == "dev" ? true : var.write_clickhouse_metrics
   }
 
   orchestrator_job_check = templatefile("${path.module}/orchestrator.hcl", merge(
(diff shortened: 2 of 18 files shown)


packages/clickhouse/migrations/20250521131546_add_metrics.sql
jakubno, line 1: It's based on [default DDL](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/clickhouseexporter/example/default_ddl/metrics_gauge.sql)

The schema has to stay like this, based on [readme](
 https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/clickhouseexporter/README.md)

packages/nomad/otel-collector.hcl
```diff
@@ -44,7 +44,7 @@ job "otel-collector" {
 
       config {
         network_mode = "host"
-        image        = "otel/opentelemetry-collector-contrib:0.99.0"
+        image        = "otel/opentelemetry-collector-contrib:0.123.0"
```
jakubno, line 47: I had to update because of some clickhouse exporter functionalities

packages/orchestrator/internal/metrics/otel.go
jakubno, line 1: created a separate otel meter provider, because I wanted:
- remove Exemplar, to minimize size
- set different export period
- set Delta temporality

packages/orchestrator/internal/metrics/meters.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+var gaugeIntDesc = map[GaugeIntType]string{
+	SandboxRamUsedGaugeName:  "Amount of RAM used by the sandbox.",
+	SandboxRamTotalGaugeName: "Amount of RAM available to the sandbox.",
+	SandboxCpuTotalGaugeName: "Amount of CPU available to the sandbox.",
+}
+
+var gaugeIntUnits = map[GaugeIntType]string{
+	// TODO: is mebibyte small enough? shouldn't we just log bytes?
```
jakubno: The annoying part is, we will have to change envd...

packages/orchestrator/internal/sandbox/checks.go
```diff
@@ -62,22 +79,20 @@ func (c *Checks) Start() {
...
-		logger.LogMetricsLoki(ctx)
-	}
-
-	if c.useClickhouseMetrics == "true" {
-		logger.LogMetricsClickhouse(ctx)
-	}
-
-	// ensure backward compatibility if neither are set
-	if c.useClickhouseMetrics != "true" && c.useLokiMetrics != "true" {
+	if c.useLokiMetrics {
```
jakubno, line 95: removed the fallback, it's handled via env variables and it doesn't make sense to use loki if you explicitly pass `false`

packages/nomad/main.tf
```diff
@@ -326,10 +332,8 @@ locals {
     otel_tracing_print           = var.otel_tracing_print
     template_bucket_name         = var.template_bucket_name
     otel_collector_grpc_endpoint = "localhost:4317"
-    clickhouse_connection_string = "clickhouse.service.consul:9000"
-    clickhouse_username          = var.clickhouse_username
-    clickhouse_password          = random_password.clickhouse_password.result
-    clickhouse_database          = var.clickhouse_database
+    write_to_loki                = var.write_loki_metrics
+    write_to_clickhouse          = var.environment == "dev" ? true : var.write_clickhouse_metrics
```
dobrac, line 343: should these be mutually exclusive in the defaults?

packages/nomad/otel-collector.hcl
```diff
@@ -101,11 +101,16 @@ receivers:
 processors:
   batch:
     timeout: 5s
+
+  batch/clickhouse:
+    timeout: 5s
+    send_batch_size: 100000
```
dobrac, line 122: what is the idea behind the batch size?

packages/orchestrator/internal/config/config.go
```diff
@@ -3,5 +3,7 @@ package config
 import "github.com/e2b-dev/infra/packages/shared/pkg/env"
 
 var (
-	AllowSandboxInternet = env.GetEnv("ALLOW_SANDBOX_INTERNET", "true") != "false"
+	AllowSandboxInternet   = env.GetEnv("ALLOW_SANDBOX_INTERNET", "true") != "false"
+	WriteLokiMetrics       = env.GetEnv("WRITE_LOKI_METRICS", "true") != "false"
+	WriteClickhouseMetrics = env.GetEnv("WRITE_CLICKHOUSE_METRICS", "true") != "false"
```
dobrac: are we by default write to both of them?

packages/orchestrator/internal/metrics/meters.go
dobrac: could you elaborate on what do you mean by changing envd?

packages/orchestrator/internal/metrics/meters.go
```diff
@@ -0,0 +1,75 @@
+package metrics
```
dobrac: what is the difference between this meters.go file and the one in shared/pkg? Should we somehow unite them?

packages/orchestrator/internal/metrics/otel.go
```diff
@@ -0,0 +1,96 @@
...
+type MeterProvider struct {
+	meterLock   sync.Mutex
+	gaugesInt   map[GaugeIntType]metric.Int64ObservableGauge
+	gaugesFloat map[GaugeFloatType]metric.Float64ObservableGauge
+
+	meter         metric.Meter
+	meterProvider *sdkmetric.MeterProvider
+}
+
+func NewSandboxMetricProvider(ctx context.Context, serviceVersion string, instanceID string, exportPeriod time.Duration) (*MeterProvider, error) {
```
dobrac: How does this play with `InitOTLPExporter`? Can we maybe reuse/abstract the logic in the `InitOTLPExporter` or merge them together?

packages/orchestrator/internal/sandbox/checks.go
```diff
@@ -62,22 +79,20 @@ func (c *Checks) Start() {
...
 	c.ctx.Lock()
 	c.ctx.Cancel()
+
+	if c.unregisterMetrics != nil {
+		if err := c.unregisterMetrics(); err != nil {
+			sbxlogger.I(c.sandbox).Error("failed to unregister metrics", zap.Error(err))
+		}
+	}
+
 	c.ctx.Unlock()
```
dobrac: maybe put this to defer

packages/orchestrator/internal/sandbox/checks.go
```diff
@@ -18,37 +19,53 @@ const (
 	minEnvdVersionForMetrcis = "0.1.5"
```
```suggestion
	minEnvdVersionForMetrics = "0.1.5"
```

packages/orchestrator/internal/sandbox/checks.go
```diff
@@ -18,37 +19,53 @@ const (
...
+		}
+
+		h.unregisterMetrics = unregister.Unregister
+	}
+
+	return h, nil
+}
+
+func (c *Checks) getMetrics(ctx context.Context) (*metrics.SandboxMetrics, error) {
+	if isGTEVersion(c.sandbox.Config.EnvdVersion, minEnvdVersionForMetrcis) {
```
dobrac: nit: flip the if, return early

packages/orchestrator/internal/sandbox/sandbox_linux.go
```diff
@@ -213,11 +211,13 @@ func CreateSandbox(
 		process:  fcHandle,
 
 		cleanup: cleanup,
-
-		ClickhouseStore: nil,
 	}
 
-	sbx.Checks = NewChecks(sbx, "", "")
+	checks, err := NewChecks(nil, sbx, false, false)
```
dobrac, line 215: is the metrics provider not required here?

packages/nomad/main.tf
jakubno: It doesn't matter that much here. I want to use LaunchDarkly, I already have some minimal setup where I want to tu support all 4 possibilities (None, one of them, both)

packages/nomad/otel-collector.hcl
jakubno: I copied it from otel collector readme: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/clickhouseexporter/README.md#example

Clickhouse works the best with batches around 10k-100k. In our case i think it won't be ever over 100k, it will be always triggered by timeout

dobrac: ok, good reference

packages/orchestrator/internal/config/config.go
jakubno: If you deploy via nomad then the default is false for clickhouse. But I will change it in here as well

packages/orchestrator/internal/metrics/meters.go
jakubno: We are reporting `mem_total_mib` as `int64` from `envd`. If we want something more precise, we have to change envd

packages/orchestrator/internal/sandbox/sandbox_linux.go
jakubno: It isn't, there's check in `NewChecks`

packages/orchestrator/internal/metrics/otel.go
jakubno: Explained in here https://github.com/e2b-dev/infra/pull/757#discussion_r2140924479

packages/orchestrator/internal/sandbox/sandbox_linux.go
dobrac: yes, but that means we don't export metrics for "create sandbox", effectively the build, right?

jakubno: Yes, the metrics are scoped for a sandbox

packages/shared/pkg/telemetry/sandbox.go
```diff
@@ -0,0 +1,105 @@
...
+		func(ctx context.Context, o metric.Observer) error {
+			sbxMetrics, err := getMetrics(ctx)
+			if err != nil {
+				return err
+			}
+
+			o.ObserveInt64(mp.cpuTotal, sbxMetrics.CPUCount, attributes)
+			o.ObserveFloat64(mp.cpuUsed, sbxMetrics.CPUUsedPercent, attributes)
+			// Save as KiB for future, so we can return more accurate values
+			o.ObserveInt64(mp.memoryTotal, sbxMetrics.MemTotalMiB*1024, attributes)
```
dobrac: can't we use bytes directly?

dobrac (approved):
Great simplification!

PR #758, opened 2025-06-11
Edge service implementation

sitole:
Implementation of Edge API that extends and replaces current Client Proxy service.

- This service is preparation for E2B extended deployment options where we can manage remote compute clusters. Edge provides us with almost exact copy of Template Manager and Orchestrator APIs for managing sandboxes and builds on remote infrastructure.
- Edge provides also remote cluster service discovery, that will help with instances management. You can use 3 different service discovery implementations (DNS, AWS EC2 that filters instances by tags, static one mainly used for tests or on-premise deployment). Service discovery query orchestrators (both build and sandbox spawn capable) and edge services (there can be multiple of them for rolling deployments and HA). You can query basic instances details and there are options to make node draining and mark it as killed (both with use of service status reporting).
- Edge also provides access to remote Loki logs implementation that will help us to see sandbox and build logs easily.
- Service works also as edge proxy for routing sandboxes traffic (such as current client proxy).
- Cross-cluster consistency of running sandboxes is done via syncing sandboxes with Redis cluster


packages/client-proxy/Makefile
```diff
@@ -18,6 +18,16 @@ build-and-upload:
 	@docker build --platform linux/amd64 --tag "$(GCP_REGION)-docker.pkg.dev/$(GCP_PROJECT_ID)/$(IMAGE)" --push --build-arg COMMIT_SHA="$(COMMIT_SHA)" .
 	@rm -rf .shared/
 
+openapi := ../../spec/openapi-edge.yml
+codegen := go run github.com/oapi-codegen/oapi-codegen/v2/cmd/oapi-codegen@latest
```
dobrac: use go tool

packages/client-proxy/internal/edge/handlers/sandbox-create.go
```diff
@@ -0,0 +1,103 @@
...
+			SandboxId:          body.Sandbox.SandboxId,
+			Alias:              body.Sandbox.Alias,
+			TeamId:             body.Sandbox.TeamId,
+			KernelVersion:      body.Sandbox.KernelVersion,
+			FirecrackerVersion: body.Sandbox.FirecrackerVersion,
+			EnvdVersion:        body.Sandbox.EnvdVersion,
+			EnvdAccessToken:    body.Sandbox.EnvdAccessToken,
+			Metadata:           sbxMetadata,
+			EnvVars:            sbxEnvVars,
+			MaxSandboxLength:   24 * 60 * 60,
```
dobrac: can we try to unite it with the API const?

spec/openapi-edge.yml
```diff
@@ -0,0 +1,783 @@
...
+          type: string
+          format: date-time
+          description: Time when the node was registered
+        roles:
+          type: array
+          items:
+            $ref: '#/components/schemas/ClusterOrchestratorRole'
+        status:
+          $ref: "#/components/schemas/ClusterNodeStatus"
+        metricVCpuUsed:
```
jakubno, line 167: Don't we need also total cpu, memory, disk?

spec/openapi-edge.yml
```diff
@@ -0,0 +1,783 @@
...
+        sandboxes:
+          type: array
+          items:
+            $ref: '#/components/schemas/RunningSandbox'
+      required: [sandboxes]
+
+    SandboxConfig:
+      type: object
+      properties:
+        orchestratorId:     { type: string }
```
jakubno: why do you use different syntax here?

spec/openapi-edge.yml
```diff
@@ -0,0 +1,783 @@
...
+          $ref: "#/components/responses/401"
+        "404":
+          $ref: "#/components/responses/404"
+        "500":
+          $ref: "#/components/responses/500"
+
+  /v1/service-discovery/nodes/{node_id}/kill:
+    post:
+      operationId: v1ServiceDiscoveryNodeKill
+      description: Mark node as unhealthy
```
jakubno: This can be little bit confusing, I would simply put Kills the node, not needed to specify the implementation details here

spec/openapi-edge.yml
```diff
@@ -0,0 +1,783 @@
...
+      parameters:
+        - name: node_id
+          in: path
+          required: true
+          description: Node ID
+          schema:
+            type: string
+      responses:
+        "200":
+          description: Successfully updated the node
```
jakubno: Do you wait till the node is killed?

packages/nomad/main.tf
```diff
@@ -103,20 +103,27 @@ resource "nomad_job" "docker_reverse_proxy" {
...
 
-      gcp_zone           = var.gcp_zone
-      port_name          = var.client_proxy_port.name
-      port_number        = var.client_proxy_port.port
-      health_port_number = var.client_proxy_health_port.port
-      environment        = var.environment
+      gcp_zone    = var.gcp_zone
+      environment = var.environment
+
+      redis_url = "redis://redis.service.consul:${var.redis_port.port}"
```
jakubno: can you use the same as in API here?
https://github.com/e2b-dev/infra/pull/769/files#diff-f62e160572dc6b7009a9fa9b3747244ce623cb87a60938252340641f02f317b7R61

packages/nomad/edge.hcl
```diff
@@ -13,27 +13,40 @@ job "client-proxy" {
...
 
     service {
       name = "proxy"
-      port = "${port_name}"
+      port = "${proxy_port_name}"
 
+      check {
+        type     = "http"
+        name     = "health"
+        path     = "/health"
```
jakubno: there's different endpoint for this right?

packages/client-proxy/internal/edge/handlers/sandbox-create.go
```diff
@@ -0,0 +1,106 @@
...
```
jakubno: copilot is right, make it a constant, something like `maxSandboxLength`

packages/client-proxy/internal/edge/handlers/sandbox-delete.go
```diff
@@ -0,0 +1,52 @@
...
+
+	_, err = orchestrator.Client.Sandbox.Delete(ctx, &grpcorchestrator.SandboxDeleteRequest{SandboxId: sandboxId})
+	if err != nil {
+		zap.L().Error("Error when deleting sandbox", zap.Error(err))
+		a.sendAPIStoreError(c, http.StatusInternalServerError, "Error when deleting sandbox")
+		telemetry.ReportCriticalError(ctx, "error when deleting sandbox", err)
+		return
+	}
+
+	zap.L().Info("Sandbox deleted", zap.String("sandbox_id", sandboxId))
```
jakubno: use the new `WithSandboxID` method (everywhere) 🙏🏻

packages/client-proxy/internal/edge/handlers/sandboxes-list.go
```diff
@@ -0,0 +1,92 @@
...
+
+	sandboxes := make([]api.RunningSandbox, 0, len(sandboxesRaw.Sandboxes))
+	for _, sbx := range sandboxesRaw.Sandboxes {
+		startTime := sbx.StartTime.AsTime()
+		endTime := sbx.EndTime.AsTime()
+
+		envVars := make(map[string]string)
+		for k, v := range sbx.Config.EnvVars {
+			envVars[k] = v
+		}
```
jakubno: do we want to return secret values here?

packages/client-proxy/internal/edge/handlers/sandboxes-list.go
```diff
@@ -0,0 +1,92 @@
...
+			EnvVars:         &envVars,
+			EnvdAccessToken: sbx.Config.EnvdAccessToken,
+			EnvdVersion:     sbx.Config.EnvdVersion,
+
+			HugePages:        sbx.Config.HugePages,
+			MaxSandboxLength: sbx.Config.MaxSandboxLength,
+			AutoPause:        sbx.Config.AutoPause,
+
+			RamMB:           sbx.Config.RamMb,
+			Snapshot:        sbx.Config.Snapshot,
```
jakubno: why is snapshot in this "group"?

packages/client-proxy/internal/edge/handlers/service-discovery-node-drain.go
```diff
@@ -0,0 +1,103 @@
...
+	logger := a.logger.With(zap.String("service_id", serviceId))
+
+	// try to find orchestrator node first
+	o, ok := a.orchestratorPool.GetOrchestrator(serviceId)
+	if !ok {
+		return errors.New("orchestrator node not found")
+	}
+
+	if o != nil {
+		logger.Info("found orchestrator node, calling drain request")
```
jakubno: inaccurate log

packages/client-proxy/internal/edge/handlers/service-discovery-node-drain.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+
+		return nil
+	}
+
+	// try to find edge node
+	e, err := a.edgePool.GetNode(serviceId)
+	if err != nil {
+		logger.Error("failed to get edge node", zap.Error(err))
+		return errors.New("failed to get edge node")
```
jakubno, line 81: it isn't quite correct, maybe you were trying to get orch

packages/client-proxy/internal/edge/handlers/store.go
```diff
@@ -0,0 +1,158 @@
...
+	a.info.SetStatus(api.Draining)
+}
+
+func (a *APIStore) SetUnhealthy() {
+	a.info.SetStatus(api.Unhealthy)
+}
+
+func (a *APIStore) GracefullyShutdown() {
+	a.SetUnhealthy()
+}
```
jakubno: is this graceful shutdown?

packages/client-proxy/internal/edge/handlers/store.go
```diff
@@ -0,0 +1,158 @@
...
+	if err != nil {
+		bodyErr := fmt.Errorf("error when parsing request: %w", err)
+		telemetry.ReportCriticalError(ctx, "error when parsing request", err)
+		return body, bodyErr
+	}
+
+	return body, nil
+}
+
+func (a *APIStore) getOrchestratorNode(orchestratorId string) (*e2borchestrators.OrchestratorNode, *APIUserFacingError) {
```
jakubno: i would move these helper functions to separate file

packages/client-proxy/internal/edge/handlers/template-build-create.go
```diff
@@ -0,0 +1,60 @@
...
+	defer templateSpan.End()
+
+	body, err := parseBody[api.TemplateBuildCreateRequest](ctx, c)
+	if err != nil {
+		a.sendAPIStoreError(c, http.StatusBadRequest, fmt.Sprintf("Error when parsing request: %s", err))
+		telemetry.ReportCriticalError(ctx, "error when parsing request", err)
+		return
+	}
+
+	orchestrator, findErr := a.getTemplateManagerNode(body.OrchestratorId)
```
jakubno:
```suggestion
	templateManager, findErr := a.getTemplateManagerNode(body.OrchestratorId)
```

packages/client-proxy/internal/edge/pool/edge-pool.go
```diff
@@ -0,0 +1,217 @@
...
+					break
+				}
+			}
+
+			// orchestrator is no longer in the list coming from service discovery
+			if !found {
+				err := p.syncNode(spanCtx, node, false)
+				if err != nil {
+					p.logger.Error("Error during edge node sync", zap.Error(err))
+				}
```
jakubno: why sync, if not found? the logic is little bit weird

packages/client-proxy/internal/edge/pool/edge-pool.go
```diff
@@ -0,0 +1,217 @@
...
+			p.logger.Error("Error closing connection to node", zap.Error(err))
+		}
+
+		// stop background sync and close everything
+		err = node.Kill()
+		if err != nil {
+			p.logger.Error("Error closing connection to node", zap.Error(err))
+		}
+
+		p.nodes.Remove(node.ServiceId) // remove from pool
```
jakubno: isn't there some possible race condition? the nodes are locked

packages/client-proxy/internal/edge/pool/orchestrator-pool.go
```diff
@@ -0,0 +1,198 @@
...
+		case <-ctx.Done():
+			p.logger.Info("Stopping keepInSync")
+			return
+		case <-ticker.C:
+			p.syncNodes(ctx)
+		}
+	}
+}
+
+func (p *OrchestratorsPool) syncNodes(ctx context.Context) {
```
jakubno, line 101: is this the same logic as for edge? Can it be abstracted?

packages/client-proxy/internal/proxy/proxy.go
```diff
@@ -34,106 +36,153 @@ const (
...
-			sandboxId, port, err := reverse_proxy.ParseHost(r.Host)
+			sandboxId, port, err := reverseproxy.ParseHost(r.Host)
 			if err != nil {
 				return nil, err
 			}
 
 			logger := zap.L().With(
 				zap.String("host", r.Host),
-				l.WithSandboxID(sandboxId),
+				zap.String("sandbox_id", sandboxId),
```
jakubno: 👮🏻

packages/client-proxy/internal/service-discovery/dns-service-discovery.go
```diff
@@ -0,0 +1,119 @@
...
+			sd.sync(ctx)
+		}
+	}
+}
+
+func (sd *DnsServiceDiscovery) sync(ctx context.Context) {
+	ctxTimeout, ctxCancel := context.WithTimeout(ctx, cacheRefreshInterval)
+	defer ctxCancel()
+
+	ips := make(map[string]string)
```
jakubno:
```suggestion
	ips := make(map[string]struct{})
```

packages/client-proxy/main.go
```diff
@@ -52,103 +62,186 @@ func run() int {
...
+	signalCtx, sigCancel := signal.NotifyContext(ctx, syscall.SIGTERM, syscall.SIGINT)
+	defer sigCancel()
+
 	logger.Info("Starting client proxy", zap.String("commit", commitSHA), zap.String("instance_id", instanceID))
 
-	healthServer := &http.Server{Addr: fmt.Sprintf(":%d", healthCheckPort)}
-	healthServer.Handler = http.HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {
-		writer.WriteHeader(http.StatusOK)
-	})
+	tracer := otel.Tracer(serviceName)
```
jakubno:
```suggestion
	tracer := tel.TracerProvider.Tracer(serviceName)
```

packages/client-proxy/main.go
```diff
@@ -52,103 +62,186 @@ func run() int {
...
-		}
-	}()
+	redisUrl := utils.RequiredEnv("REDIS_URL", "Redis cluster URL")
+	opts, err := redis.ParseURL(redisUrl)
+	if err != nil {
+		zap.L().Fatal("invalid redis URL", zap.String("url", redisUrl), zap.Error(err))
+	}
+
+	redisClient := redis.NewClient(opts)
+	redisSync := redsync.New(goredis.NewPool(redisClient))
```
jakubno: does it make sense to initiliaze if you aren't using redis resolutiion?

spec/openapi-edge.yml
sitole: It can take minutes, because it will be killed by cloud scheduler. In aws it can take like 10 minutes. Or you mean just wait for service to agree on unhealthy status?

packages/nomad/edge.hcl
sitole: true, thx!

sitole:
Add memory-based catalog provider so we don't need to use redis cluster for development or single-node deployments that don't heed ha.

sitole:
For client proxy add env that will optionally enable redis resolution.

spec/openapi-edge.yml
sitole: Orchestrator does not support that, right? But yes, we may want to add it.

packages/client-proxy/main.go
sitole: its still used, because of catalog. Will abstract this away and add in-memory catalog for dev and single-node deployments.

packages/client-proxy/internal/edge/handlers/sandbox-pause.go
```diff
@@ -0,0 +1,74 @@
...
+		zap.L().Error("Error when pausing sandbox", zap.Error(err))
+		a.sendAPIStoreError(c, http.StatusInternalServerError, "Error when pausing sandbox")
+		telemetry.ReportCriticalError(ctx, "error when pausing sandbox", err)
+		return
+	}
+
+	err = a.sandboxes.DeleteSandbox(sandboxId)
+	if err != nil {
+		zap.L().Error("Error when deleting sandbox from cache", zap.Error(err))
+	}
```
sitole: I'am deleting sandbox from cache here, same as in sandbox delete handler. In both cases it's used during request to fetch it for cache so will not work when I will pause and then delete (if this is even valid case).

On API we are removing it from cache after calling pause, so its just replicated same flow.

sitole:
> Add memory-based catalog provider so we don't need to use redis cluster for development or single-node deployments that don't heed ha.

Implemented

packages/client-proxy/internal/edge/pool/edge-pool.go
sitole: will use dedicated `removeNode` method. Previously `syncNode` handled both remove and node data sync, sync is not moved directly into edge node struct.

packages/client-proxy/internal/edge/pool/orchestrator-pool.go
sitole: yes, basically it is. Now I simplified it little bit and theoretically we can merge it, but I would not do it right now, because at some point we can do things differently and then we will need to split it or IF it and that is something I don't want to do.

packages/client-proxy/internal/edge/handlers/store.go
sitole: removed. main function will mark api unhealthy, optionally wait for dependent services to realize that, then it will gracefully shutdown http server and cancel context (that will kill all pools and background work running there).

spec/openapi-edge.yml
jakubno: Not right now, could you create an issue to add it? 🙏🏻

spec/openapi-edge.yml
jakubno: Then we should probably return 202 and write better description, like "Accepted a request to kill the node`

spec/openapi-edge.yml
sitole: https://linear.app/e2b/issue/E2B-2488/orchestrator-and-edge-should-return-memory-and-cpu-stats

packages/client-proxy/internal/edge/handlers/sandbox-pause.go
```diff
@@ -60,6 +64,11 @@ func (a *APIStore) V1PauseSandbox(c *gin.Context, sandboxId api.SandboxId) {
 		return
 	}
 
-	zap.L().Info("Sandbox paused", zap.String("sandbox_id", sandboxId))
+	err = a.sandboxes.DeleteSandbox(sandboxId)
+	if err != nil {
+		zap.L().Error("Error when deleting sandbox from cache", zap.Error(err))
```
jakubno:
```suggestion
		zap.L().Error("Error when deleting sandbox from cache",  l.WithSandboxID(sandboxId), zap.Error(err))
```

packages/client-proxy/internal/edge/handlers/store.go
```diff
@@ -93,7 +90,7 @@ func (a *APIStore) SetUnhealthy() {
 }
 
 func (a *APIStore) GracefullyShutdown() {
```
jakubno: not used

packages/client-proxy/internal/edge/handlers/store.go
```diff
@@ -65,7 +62,7 @@ func NewStore(ctx context.Context, logger *zap.Logger, tracer trace.Tracer, info
 			return
 		}
 
-		ticker := time.NewTicker(5 * time.Millisecond)
+		ticker := time.NewTicker(100 * time.Millisecond)
```
jakubno: please make constant

packages/client-proxy/internal/edge/pool/edge-pool.go
```diff
@@ -159,9 +159,9 @@ func (p *EdgePool) syncNodes(ctx context.Context) {
 
 			// orchestrator is no longer in the list coming from service discovery
 			if !found {
-				err := p.syncNode(spanCtx, node, false)
+				err := p.removeNode(spanCtx, node)
```
jakubno, line 162: shouldn't remove node also lock?

packages/client-proxy/internal/edge/pool/edge-pool.go
```diff
@@ -182,36 +182,36 @@ func (p *EdgePool) connectNode(ctx context.Context, node *sd.ServiceDiscoveryIte
...
-			p.logger.Error("Error closing connection to node", zap.Error(err))
-		}
+	// stop background sync and close everything
+	err = node.Kill()
+	if err != nil {
+		p.logger.Error("Error closing connection to node", zap.Error(err))
+	}
 
-		p.nodes.Remove(node.ServiceId) // remove from pool
+	p.mutex.Lock()
```
jakubno, line 205: does it make sense to lock here?  You probably don't want to use node which have already closed connection?

packages/client-proxy/internal/edge/pool/orchestrator-pool.go
jakubno, line 1: probably same as for edge pool

packages/client-proxy/internal/edge/sandboxes/catalog_memory.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+func (c *MemorySandboxCatalog) StoreSandbox(sandboxId string, sandboxInfo *SandboxInfo) error {
+	_, span := c.tracer.Start(c.ctx, "sandbox-catalog-store")
+	defer span.End()
+
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+
+	c.cache.Set(sandboxId, sandboxInfo, catalogCacheExpiration)
```
jakubno: wouldn't it be better to set it to max sandbox length (+ some buffer). We may allow longer sandboxes in future

packages/client-proxy/internal/edge/handlers/store.go
```diff
@@ -0,0 +1,128 @@
...
+			return
+		}
+
+		// we don't want to skip it entirely, and we want to wait few seconds in case of cluster already contains orchestrators
+		// so we are not propagating API without not yet registered orchestrators
+		if skipInitialOrchestratorCheck {
+			time.Sleep(10 * time.Second)
+			store.info.SetStatus(api.Healthy)
+			return
+		}
```
sitole, line 80: @jakubno just added this. It optionally allows to run edge without orchestrators.

Its usefully when creating new cluster, because edge instance will spawn in less than minute, but compute takes few minutes, between compute nodes is healthy, edge will be killed because not yet healthy etc.

packages/client-proxy/internal/edge/sandboxes/catalog_redis.go
```diff
@@ -0,0 +1,121 @@
...
+	"go.opentelemetry.io/otel/trace"
+)
+
+const (
+	catalogClusterLockName = "sandboxes-catalog-cluster-lock"
+	catalogRedisTimeout    = time.Second * 5
+
+	// this is just how long we are keeping sandbox in local cache so we don't have to query redis every time
+	// we don't want to go too high because then sbx can be run on different orchestrator, and we will not be able to find it
+	catalogRedisLocalCacheTtl = time.Second * 10
```
jakubno: I think even 10 seconds is too much, also does it get updated in background or if I will be pinging it every 5 seconds, it will stay in the cache forever?

packages/client-proxy/internal/edge/handlers/store.go
sitole: 10 second wait is here just for cases when we still want to wait some time until edge will catch up service discovery and register already existing nodes so we are not allowing traffic to not yet ready (empty) edge.

packages/client-proxy/internal/edge/sandboxes/catalog_redis.go
```diff
@@ -0,0 +1,121 @@
...
+		return nil, ErrSandboxNotFound
+	}
+
+	var info *SandboxInfo
+	err = json.Unmarshal(data, &info)
+	if err != nil {
+		return nil, fmt.Errorf("failed to unmarshal sandbox info: %w", err)
+	}
+
+	err = c.StoreSandbox(sandboxId, info)
```
jakubno: why do you call store here?

packages/client-proxy/internal/edge/sandboxes/catalog_redis.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	return info, nil
+}
+
+func (c *RedisSandboxCatalog) StoreSandbox(sandboxId string, sandboxInfo *SandboxInfo) error {
+	spanCtx, span := c.tracer.Start(c.ctx, "sandbox-catalog-store")
+	defer span.End()
+
+	err := c.clusterMutex.Lock()
```
jakubno, line 88: probably okay for now, but we may need to change this to per sandbox lock, because this is clearly a potential bottleneck

packages/client-proxy/internal/edge/handlers/sandbox-pause.go
jakubno: makes sense, once the orchestrator returns pause it doesn't make sense to route anything there

packages/client-proxy/internal/edge/pool/edge-pool.go
sitole: You want to lock, because what if at the same time you will request it from map with listing/get function?

packages/client-proxy/internal/edge/sandboxes/catalog_redis.go
sitole: yes, we can improve in future, agree

packages/client-proxy/main.go
```diff
@@ -52,103 +64,196 @@ func run() int {
...
-		case err != nil:
-			exitCode.Add(1)
-			logger.Error("http service encountered error", zap.Int("port", healthCheckPort), zap.Error(err))
-		default:
-			// this probably shouldn't happen...
-			logger.Error("http service exited without error", zap.Int("port", healthCheckPort))
+	var redisClient *redis.Client
+	var catalog sandboxes.SandboxesCatalog
+
+	redisUrl := os.Getenv("REDIS_URL")
```
jakubno: can you support redis cluster as here:
https://github.com/e2b-dev/infra/blob/main/packages/api/internal/handlers/store.go#L119

packages/client-proxy/internal/edge/sandboxes/catalog_redis.go
sitole: Lowered to 5 seconds. It ttl, so if during 5 seconds interval someone will ask entry it will not agin gain 5 seconds, but will continue from counting until it will be removed.

packages/client-proxy/internal/edge/sandboxes/catalog_redis.go
sitole: when it is not in local cache, we will change redis, if its there we will store it in local one and return it.

packages/client-proxy/internal/edge/pool/edge-pool.go
sitole: Iam locking it when calling removal from map inside `removeNode` method

```golang
	p.mutex.Lock()
	defer p.mutex.Unlock()

	p.nodes.Remove(node.ServiceId)

	p.logger.Info("Edge node node connection has been closed.", zap.String("node_id", node.ServiceId))
	return nil
```

packages/client-proxy/internal/edge/sandboxes/catalog_redis.go
jakubno: Add a comment there please 🙏🏻

sitole: done, thx

packages/client-proxy/internal/edge/handlers/sandbox-create.go
```diff
@@ -83,14 +84,14 @@ func (a *APIStore) V1CreateSandbox(c *gin.Context) {
...
 	}
 
-	err = a.sandboxes.StoreSandbox(
-		body.Sandbox.SandboxId,
-		&sandboxes.SandboxInfo{
-			OrchestratorId: body.Sandbox.OrchestratorId,
-			TemplateId:     body.Sandbox.TemplateId,
-		},
-	)
+	sbxMaxLifetime := time.Duration(body.Sandbox.MaxSandboxLength) * time.Second
```
jakubno:
```
        maxSandboxLength:
          type: integer
          format: int64
          description: Maximum duration in hours
```
```suggestion
	sbxMaxLifetime := time.Duration(body.Sandbox.MaxSandboxLength) * time.Hour
```

packages/client-proxy/internal/edge/sandboxes/catalog_redis.go
```diff
@@ -68,15 +69,16 @@ func (c *RedisSandboxCatalog) GetSandbox(sandboxId string) (*SandboxInfo, error)
 		return nil, fmt.Errorf("failed to unmarshal sandbox info: %w", err)
 	}
 
-	err = c.StoreSandbox(sandboxId, info)
+	lifetime := time.Duration(info.MaxSandboxLength * int64(time.Second))
```
jakubno: also hours here

jakubno: maybe rename the field so it's less confusing

packages/client-proxy/main.go
```diff
@@ -251,8 +248,6 @@ func run() int {
 			time.Sleep(shutdownUnhealthyWait)
 		}
 
-		edgeApiStore.GracefullyShutdown()
```
jakubno: why did you rewrite this logic?

packages/client-proxy/internal/edge/sandboxes/catalog_memory.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func (c *MemorySandboxCatalog) GetSandbox(sandboxId string) (*SandboxInfo, error) {
+	_, span := c.tracer.Start(c.ctx, "sandbox-catalog-get")
+	defer span.End()
+
+	c.mtx.RLock()
+	defer c.mtx.RUnlock()
+
+	sandboxInfo := c.cache.Get(sandboxId)
```
jakubno, line 37: you probably want get without touch otherwise you will reset the TTL

packages/client-proxy/internal/edge/pool/orchestrator-pool.go
```diff
@@ -24,7 +24,8 @@ type OrchestratorsPool struct {
 }
 
 const (
-	orchestratorsCacheRefreshInterval = 10 * time.Second
+	orchestratorsCacheRefreshInterval    = 10 * time.Second
+	orchestratorsAnalyticsReportInterval = 1 * time.Minute
```
jakubno: the name is off

jakubno: I would name same as API - `statusLogInterval`, same for the loop

jakubno (changes requested):

packages/client-proxy/internal/edge/pool/orchestrator-pool.go
```diff
@@ -79,7 +79,7 @@ func (p *OrchestratorsPool) keepInSync(ctx context.Context) {
 }
 
 func (p *OrchestratorsPool) analyticsSync(ctx context.Context) {
```
jakubno: Rename please

packages/client-proxy/internal/service-discovery/builder.go
```diff
@@ -52,14 +52,23 @@ func resolveServiceDiscoveryConfig(ctx context.Context, prefix string, port int,
...
-		return nil, fmt.Errorf("missing %s environment variable", env)
+		return nil, fmt.Errorf("missing %s environment variable", dnsHostsEnv)
+	}
+
+	var dnsResolverAddress string
+	dnsResolverEnv := fmt.Sprintf("%s_DNS_RESOLVER_ADDRESS", prefix)
+	dnsResolverRaw := os.Getenv(dnsResolverEnv)
+	if dnsResolverRaw == "" {
+		return nil, fmt.Errorf("missing %s environment variable", dnsResolverEnv)
+	} else {
```
jakubno: does it need to be in else clause?

jakubno (changes requested):

jakubno (approved):

PR #759, opened 2025-06-12
Replace global telemetry usage with the local one

jakubno:
# Description

Instead of using a global otel provider, pass it so we can customize.


packages/api/internal/orchestrator/orchestrator.go
```diff
@@ -68,17 +71,20 @@ func New(
...
+		analytics:       analyticsInstance,
+		nomadClient:     nomadClient,
+		tracer:          tracer,
+		nodes:           smap.New[*Node](),
+		dns:             dnsServer,
+		dbClient:        dbClient,
+		telemetryClient: telemetryClient,
 	}
 
+	meter := telemetryClient.MeterProvider.Meter("api[")
```
dobrac: `api[` ?

packages/api/main.go
```diff
@@ -69,14 +71,14 @@ func NewGinServer(ctx context.Context, logger *zap.Logger, apiStore *handlers.AP
...
-			tracingMiddleware.Middleware(serviceName),
+			tracingMiddleware.Middleware(telemetryClient.TracerProvider, serviceName),
 			"/health",
 			"/sandboxes/:sandboxID/refreshes",
 			"/templates/:templateID/builds/:buildID/logs",
 			"/templates/:templateID/builds/:buildID/status",
 		),
 		customMiddleware.IncludeRoutes(
-			metricsMiddleware.Middleware(serviceName),
+			metricsMiddleware.Middleware(telemetryClient.MeterProvider.Meter("api-metrics"), serviceName),
```
dobrac: this is a diff from `metric.WithInstrumentationVersion(SemVersion())`, is it better?

packages/api/main.go
```diff
@@ -206,10 +208,17 @@ func run() int {
...
+	if err != nil {
+		zap.L().Fatal("failed to create metrics exporter", zap.Error(err))
 	}
+	defer func() {
+		err := telemetryClient.Shutdown(ctx)
+		if err != nil {
+			log.Printf("telemetry shutdown:%v\n", err)
+		}
+	}()
+	global.SetLoggerProvider(telemetryClient.LogsProvider)
```
dobrac: you can pass it to `GetOTELCore` instead of setting it as global

packages/api/main.go
```diff
@@ -315,11 +324,11 @@ func run() int {
 	// Create an instance of our handler which satisfies the generated interface
 	//  (use the outer context rather than the signal handling
 	//   context so it doesn't exit first.)
-	apiStore := handlers.NewAPIStore(ctx)
+	apiStore := handlers.NewAPIStore(ctx, telemetryClient)
 	cleanupFns = append(cleanupFns, apiStore.Close)
 
 	// pass the signal context so that handlers know when shutdown is happening.
-	s := NewGinServer(ctx, logger, apiStore, swagger, port)
+	s := NewGinServer(ctx, telemetryClient, logger, apiStore, swagger, port)
```
dobrac: nit: how about renaming all "telemetryClient" vars to something much shorter?

packages/shared/pkg/telemetry/logs.go
```diff
@@ -0,0 +1,59 @@
...
+
+func (noopLogExporter) Export(context.Context, []sdklog.Record) error { return nil }
+
+func (noopLogExporter) Shutdown(context.Context) error { return nil }
+
+func (noopLogExporter) ForceFlush(context.Context) error { return nil }
+
+func NewLogExporter(ctx context.Context, extraOption ...otlploggrpc.Option) (sdklog.Exporter, error) {
+	if env.IsLocal() {
+		return &noopLogExporter{}, nil
```
dobrac: I'm not a fan to having this inside of the logic, I would prefer it back being up the stream, so it's more transparent when you initialize it

packages/shared/pkg/telemetry/metrics.go
```diff
@@ -0,0 +1,78 @@
...
+func (noopMetricExporter) ForceFlush(context.Context) error {
+	return nil
+}
+func (noopMetricExporter) Shutdown(context.Context) error {
+	return nil
+}
+
+func NewMeterExporter(ctx context.Context, extraOption ...otlpmetricgrpc.Option) (sdkmetric.Exporter, error) {
+	if env.IsLocal() {
+		return &noopMetricExporter{}, nil
```
dobrac: same here

dobrac (changes requested):

packages/api/main.go
jakubno: It was some copied code, it was set to always return "1.0.0"

packages/api/internal/middleware/otel/metrics/otelrecorder.go
```diff
@@ -25,8 +24,7 @@ func GetRecorder(metricsPrefix string) Recorder {
 		return metricName
 	}
 
-	meter := otel.Meter("api-metrics", metric.WithInstrumentationVersion(SemVersion()))
-
+	meter := meterProvider.Meter("middleware")
```
dobrac: is it fine to rename it?

packages/shared/pkg/telemetry/meters.go
```diff
@@ -88,56 +87,69 @@ var observableUpDownCounterUnits = map[ObservableUpDownCounterType]string{
...
-	upDownCounters[name] = counter
+var gaugeIntDesc = map[GaugeIntType]string{
+	SandboxRamUsedGaugeName:  "Amount of RAM used by the sandbox.",
+	SandboxRamTotalGaugeName: "Amount of RAM available to the sandbox.",
+	SandboxCpuTotalGaugeName: "Amount of CPU available to the sandbox.",
+}
 
-	return counter, nil
+var gaugeIntUnits = map[GaugeIntType]string{
+	// TODO: is mebibyte small enough? shouldn't we just log bytes?
```
dobrac: could you please add the context with envd here please?

packages/api/internal/middleware/otel/metrics/otelrecorder.go
jakubno: Currently it isn't used anywhere, you can use it e.g. to filter by it in otel collector. I will improve the names with some logical grouping

dobrac (approved):
Nice job! 🚀

PR #762, opened 2025-06-12
Remove zonal instances

jakubno:
# Description

Last step of migrating to regional instances


dobrac (approved):
Just to mark it, this is a breaking change that will cause a downtime without proper transition care

PR #765, opened 2025-06-12
Add nf_tables support to the FC kernel

dobrac:
Add nf_tables support to the FC kernel. Contains all vars from here: https://wiki.nftables.org/wiki-nftables/index.php/Building_and_installing_nftables_from_sources + few extra I've encountered when testing.


ValentaTomas:
Any possible performance implications of this?

PR #766, opened 2025-06-13
BYOC: Support template build for team clusters

sitole:
Implements abstraction for template managers in API to support edge clusters.

### DIFF
--- a/packages/shared/pkg/synchronization/synchronization.go
+++ b/packages/shared/pkg/synchronization/synchronization.go
@@ -30,7 +30,8 @@ type Synchronize[SourceItem any, PoolItem any] struct {
 	tracerSpanPrefix string
 	logsPrefix       string
 
-	cancel chan struct{} // channel for cancellation of synchronization
+	cancel     chan struct{} // channel for cancellation of synchronization
+	cancelOnce sync.Once
 }
 
 func NewSynchronize[SourceItem any, PoolItem any](tracer trace.Tracer, spanPrefix string, logsPrefix string, store Store[SourceItem, PoolItem]) *Synchronize[SourceItem, PoolItem] {
@@ -75,7 +76,9 @@ func (s *Synchronize[SourceItem, PoolItem]) Start(syncInterval time.Duration, sy
 }
 
 func (s *Synchronize[SourceItem, PoolItem]) Close() {
-	close(s.cancel)
+	s.cancelOnce.Do(
+		func() { close(s.cancel) },
+	)
 }
 
 func (s *Synchronize[SourceItem, PoolItem]) sync(ctx context.Context) error {


packages/api/internal/template-manager/template_manager.go
```diff
@@ -71,247 +82,105 @@ func (tm *TemplateManager) BuildsStatusPeriodicalSync(ctx context.Context) {
...
-			for _, buildDB := range buildsRunning {
-				go tm.BuildStatusSync(ctx, buildDB.ID, *buildDB.EnvID)
-			}
+			//for _, buildDB := range buildsRunning {
+			//
+			//	buildDB.
+			//
+			//
+			//	go tm.BuildStatusSync(ctx, buildDB.ID, *buildDB.EnvID, nil, buildDB.ClusterNodeID)
+			//}
```
sitole: This needs to be adjusted

packages/api/internal/template-manager/template_manager_test.go
```diff
@@ -82,43 +70,35 @@ func TestPollBuildStatus_setStatus(t *testing.T) {
...
 			status:  nil,
 		},
-		// if status client and err are nil, return err
-		{
-			name: "should return error if status client and err are nil",
-			fields: fields{
-				statusClient: nil,
-			},
-			wantErr: true,
-		},
```
sitole, line 99: Testing if the status client is nil. No more relevant because I removed status clients, and it's handled by template client, same as different template build-related requests.

packages/api/internal/grpc/client.go
```diff
@@ -0,0 +1,28 @@
...
+}
+
+func (a *GRPCClient) Close() error {
+	err := a.Connection.Close()
+	if err != nil {
+		return fmt.Errorf("failed to close connection: %w", err)
+	}
+
+	return nil
+}
```
sitole: Shared GRPC client for template manager and orchestrator so we can use them in unified way across codebase

packages/shared/pkg/schema/build.go
```diff
@@ -43,6 +43,7 @@ func (EnvBuild) Fields() []ent.Field {
 		field.String("kernel_version").Default(DefaultKernelVersion).SchemaType(map[string]string{dialect.Postgres: "text"}),
 		field.String("firecracker_version").Default(DefaultFirecrackerVersion).SchemaType(map[string]string{dialect.Postgres: "text"}),
 		field.String("envd_version").SchemaType(map[string]string{dialect.Postgres: "text"}).Nillable().Optional(),
+		field.String("cluster_node_id").SchemaType(map[string]string{dialect.Postgres: "text"}).Optional().Nillable(),
```
dobrac, line 46: does this mean the cluster_node_id has to be globally unique?

packages/api/internal/grpc/client.go
```diff
@@ -0,0 +1,28 @@
...
+
+	orchestratorgrpc "github.com/e2b-dev/infra/packages/shared/pkg/grpc/orchestrator"
+	infogrpc "github.com/e2b-dev/infra/packages/shared/pkg/grpc/orchestrator-info"
+	tempaltemanagergrpc "github.com/e2b-dev/infra/packages/shared/pkg/grpc/template-manager"
+)
+
+type GRPCClient struct {
+	Info     infogrpc.InfoServiceClient
+	Sandbox  orchestratorgrpc.SandboxServiceClient
+	Template tempaltemanagergrpc.TemplateServiceClient
```
dobrac:
```suggestion
	Template templatemanagergrpc.TemplateServiceClient
```

packages/api/internal/handlers/sandbox_kill.go
```diff
@@ -18,11 +18,7 @@ import (
 	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
 )
 
-func (a *APIStore) deleteSnapshot(
-	ctx context.Context,
-	sandboxID string,
-	teamID uuid.UUID,
-) error {
+func (a *APIStore) deleteSnapshot(ctx context.Context, sandboxID string, teamID uuid.UUID, teamClusterId *uuid.UUID) error {
```
dobrac:
```suggestion
func (a *APIStore) deleteSnapshot(ctx context.Context, sandboxID string, teamID uuid.UUID, teamClusterID *uuid.UUID) error {
```

packages/api/internal/edge/cluster_nodes.go
```diff
@@ -0,0 +1,168 @@
...
+
+	"github.com/e2b-dev/infra/packages/api/internal/utils"
+	infogrpc "github.com/e2b-dev/infra/packages/shared/pkg/grpc/orchestrator-info"
+	l "github.com/e2b-dev/infra/packages/shared/pkg/logger"
+	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
+)
+
+type ClusterNode struct {
+	Id     string // service id (will change on restart)
+	NodeId string // machine id
```
dobrac:
```suggestion
	ID     string // service id (will change on restart)
	NodeID string // machine id
```

packages/api/internal/handlers/sandbox_kill.go
```diff
@@ -102,7 +101,7 @@ func (a *APIStore) DeleteSandboxesSandboxID(
 		}
 
 		// remove any snapshots of the sandbox
-		err := a.deleteSnapshot(ctx, sandboxID, teamID)
+		err := a.deleteSnapshot(ctx, sandboxID, teamID, a.GetTeamInfo(c).Team.ClusterID)
```
dobrac: the team is already accessed on line 75 (`teamID := c.Value(auth.TeamContextKey).(authcache.AuthTeamInfo).Team.ID`)

packages/api/internal/handlers/sandbox_kill.go
```diff
@@ -118,7 +117,7 @@ func (a *APIStore) DeleteSandboxesSandboxID(
 	}
 
 	// remove any snapshots when the sandbox is not running
-	deleteSnapshotErr := a.deleteSnapshot(ctx, sandboxID, teamID)
+	deleteSnapshotErr := a.deleteSnapshot(ctx, sandboxID, teamID, a.GetTeamInfo(c).Team.ClusterID)
```
dobrac: the team is already accessed on line 75 (teamID := c.Value(auth.TeamContextKey).(authcache.AuthTeamInfo).Team.ID)

packages/api/internal/handlers/template_start_build.go
```diff
@@ -141,6 +141,13 @@ func (a *APIStore) PostTemplatesTemplateIDBuildsBuildID(c *gin.Context, template
 		return
 	}
 
+	// team is part of the cluster but template build is not assigned to a cluster node so its invalid stats
+	if team.ClusterID != nil && build.ClusterNodeID == nil {
+		a.sendAPIStoreError(c, http.StatusInternalServerError, "build is not assigned to a cluster node")
+		telemetry.ReportCriticalError(ctx, "build is not assigned to a cluster node", fmt.Errorf("build is not assigned to a cluster node"), telemetry.WithTemplateID(templateID))
```
dobrac:
```suggestion
		telemetry.ReportCriticalError(ctx, "build is not assigned to a cluster node", nil, telemetry.WithTemplateID(templateID))
```

packages/api/internal/utils/db.go
```diff
@@ -20,7 +20,7 @@ func CheckMigrationVersion(expectedMigration int64) error {
 	defer func() {
 		dbErr := db.Close()
 		if dbErr != nil {
-			zap.L().Error("failed to close database connection", zap.Error(dbErr))
+			zap.L().Error("Failed to close database connection", zap.Error(dbErr))
```
dobrac, line 23: just curious, why the capitalization changes?

packages/api/internal/edge/cluster.go
```diff
@@ -0,0 +1,193 @@
...
+	grpclient "github.com/e2b-dev/infra/packages/api/internal/grpc"
+	infogrpc "github.com/e2b-dev/infra/packages/shared/pkg/grpc/orchestrator-info"
+	tempaltemanagergrpc "github.com/e2b-dev/infra/packages/shared/pkg/grpc/template-manager"
+	api "github.com/e2b-dev/infra/packages/shared/pkg/http/edge"
+	"github.com/e2b-dev/infra/packages/shared/pkg/smap"
+	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
+)
+
+type Cluster struct {
+	Id uuid.UUID
```
dobrac:
```suggestion
	ID uuid.UUID
```

packages/shared/pkg/schema/build.go
sitole: This is just to identify the node inside the cluster where the build runs. When we are taking a node, we are already scoped just to a specific cluster.

packages/api/internal/utils/db.go
sitole: Looks like a bigger part of our codebase is using logger messages with first upper, so want to go with it.

packages/db/migrations/20250624001048_cluster_for_templates.sql
```diff
@@ -0,0 +1,17 @@
+-- +goose Up
+-- +goose StatementBegin
+ALTER TABLE envs
+    ADD COLUMN IF NOT EXISTS cluster_id UUID NULL
+    REFERENCES clusters(id);
+
+CREATE INDEX IF NOT EXISTS envs_cluster_id_uq
```
dobrac: does the _uq mean that it should be unique?

packages/api/internal/template-manager/logs.go
```diff
@@ -0,0 +1,114 @@
...
+	client *api.ClientWithResponses
+	nodeId string
+}
+
+func NewClusterPlacementLogsProvider(client *api.ClientWithResponses, nodeId string) PlacementLogsProvider {
+	return &ClusterPlacementLogsProvider{client: client, nodeId: nodeId}
+}
+
+func (l *ClusterPlacementLogsProvider) GetLogs(ctx context.Context, buildId string, templateId string, offset *int32) (*[]string, error) {
+	res, err := l.client.V1TemplateBuildLogsWithResponse(
```
dobrac: this should be a call to the edge node, correct? If so, can we name the "client" bit more explicit?

packages/api/internal/template-manager/template_manager.go
```diff
@@ -2,35 +2,62 @@ package template_manager
...
+	grpclient "github.com/e2b-dev/infra/packages/api/internal/grpc"
+	"github.com/e2b-dev/infra/packages/api/internal/sandbox"
+	"github.com/e2b-dev/infra/packages/api/internal/utils"
+	sqlcdb "github.com/e2b-dev/infra/packages/db/client"
 	"github.com/e2b-dev/infra/packages/shared/pkg/db"
-	template_manager "github.com/e2b-dev/infra/packages/shared/pkg/grpc/template-manager"
-	"github.com/e2b-dev/infra/packages/shared/pkg/logger"
-	"github.com/e2b-dev/infra/packages/shared/pkg/models/envbuild"
+	infogrpc "github.com/e2b-dev/infra/packages/shared/pkg/grpc/orchestrator-info"
+	tempaltemanagergrpc "github.com/e2b-dev/infra/packages/shared/pkg/grpc/template-manager"
```
dobrac:
```suggestion
	templatemanagergrpc "github.com/e2b-dev/infra/packages/shared/pkg/grpc/template-manager"
```

packages/api/internal/edge/cluster.go
```diff
@@ -0,0 +1,193 @@
...
+	"go.opentelemetry.io/otel/trace"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/keepalive"
+	"google.golang.org/grpc/metadata"
+
+	grpclient "github.com/e2b-dev/infra/packages/api/internal/grpc"
+	infogrpc "github.com/e2b-dev/infra/packages/shared/pkg/grpc/orchestrator-info"
+	tempaltemanagergrpc "github.com/e2b-dev/infra/packages/shared/pkg/grpc/template-manager"
```
dobrac:
```suggestion
	templatemanagergrpc "github.com/e2b-dev/infra/packages/shared/pkg/grpc/template-manager"
```

packages/api/internal/edge/cluster.go
```diff
@@ -0,0 +1,193 @@
...
+	tracer trace.Tracer
+
+	ctx       context.Context
+	ctxCancel context.CancelFunc
+}
+
+const (
+	edgeApiAuthHeader = "X-API-Key"
+	edgeRpcAuthHeader = "authorization"
+	edgeRpcNodeHeader = "node-id"
```
dobrac: I would assume this is defined twice, as these variables are used only once. Can we unite it?

packages/api/internal/edge/cluster.go
```diff
@@ -0,0 +1,193 @@
...
+)
+
+var (
+	ErrTemplateBuilderNotFound          = errors.New("template builder not found")
+	ErrAvailableTemplateBuilderNotFound = errors.New("available template builder not found")
+)
+
+func NewCluster(ctx context.Context, tracer trace.Tracer, tel *telemetry.Client, endpoint string, endpointTls bool, secret string, id uuid.UUID) (*Cluster, error) {
+	// so we during cluster disconnect we don't cancel the upper context
```
dobrac: clarify meaning of the comment

packages/api/internal/edge/cluster.go
```diff
@@ -0,0 +1,193 @@
...
+			keepalive.ClientParameters{
+				Time:                10 * time.Second, // Send ping every 10s
+				Timeout:             2 * time.Second,  // Wait 2s for response
+				PermitWithoutStream: true,
+			},
+		),
+	}
+
+	if endpointTls {
+		cred := credentials.NewTLS(&tls.Config{MinVersion: tls.VersionTLS12})
```
dobrac:
```suggestion
		cred := credentials.NewTLS(&tls.Config{MinVersion: tls.VersionTLS13})
```

(we use VersionTLS13 else where already)

packages/api/internal/edge/cluster.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+
+	conn, err := grpc.NewClient(endpoint, grpcOptions...)
+	if err != nil {
+		ctxCancel()
+		return nil, fmt.Errorf("failed to create grpc client: %w", err)
+	}
+
+	grpcClient := &grpclient.GRPCClient{
+		Sandbox:    nil,
```
dobrac: why is the sandbox client nil?

packages/api/internal/edge/cluster.go
```diff
@@ -0,0 +1,193 @@
...
+		grpcClient: grpcClient,
+	}
+
+	// periodically sync cluster nodes
+	go c.syncBackground()
+
+	return c, nil
+}
+
+func (c *Cluster) Disconnect() error {
```
dobrac: Why not Close?

packages/api/internal/edge/cluster.go
```diff
@@ -0,0 +1,193 @@
...
+	node, found := c.nodes.Get(nodeID)
+	if !found {
+		return nil, ErrTemplateBuilderNotFound
+	}
+
+	// protect node state with mutex
+	node.mutex.RLock()
+	defer node.mutex.RUnlock()
+
+	if node.Status == infogrpc.ServiceInfoStatus_OrchestratorUnhealthy || !slices.Contains(node.Roles, infogrpc.ServiceInfoRole_TemplateManager) {
```
dobrac: out of topic: this is quite a miss in the enum naming imo

packages/api/internal/edge/cluster.go
```diff
@@ -0,0 +1,193 @@
...
+		return nil, ErrTemplateBuilderNotFound
+	}
+
+	return node, nil
+}
+
+func (c *Cluster) GetAvailableTemplateBuilder() (*ClusterNode, error) {
+	_, span := c.tracer.Start(c.ctx, "template-builder-get-available-node")
```
dobrac: is the span relevant to the whole lifecycle context of the Cluster? Isn't it more related to one specific execution?

packages/api/internal/edge/cluster.go
```diff
@@ -0,0 +1,193 @@
...
+	span.SetAttributes(telemetry.WithClusterID(c.Id))
+	defer span.End()
+
+	for _, node := range c.nodes.Items() {
+		node.mutex.RLock()
+
+		// we don't want to place new builds to not healthy nodes
+		if node.Status != infogrpc.ServiceInfoStatus_OrchestratorHealthy {
```
dobrac: can you create getter/setter for the status, that will include the locks? I think it will be less error prone then

packages/api/internal/edge/cluster.go
```diff
@@ -0,0 +1,193 @@
...
+			node.mutex.RUnlock()
+			continue
+		}
+		node.mutex.RUnlock()
+
+		// we want to use only template builders
+		if !slices.Contains(node.Roles, infogrpc.ServiceInfoRole_TemplateManager) {
```
dobrac: I would create a node function that will accept the desired type, instead calling everywhere `slices.Contains` (to not need the comment then)

packages/api/internal/edge/cluster_nodes.go
```diff
@@ -0,0 +1,168 @@
...
+
+	Status infogrpc.ServiceInfoStatus
+	Roles  []infogrpc.ServiceInfoRole
+
+	Version       string
+	VersionCommit string
+
+	mutex sync.RWMutex // mutex to protect the node state
```
dobrac: can you rename it, so the comment is not needed?

packages/api/internal/edge/cluster_nodes.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+}
+
+func (c *Cluster) sync(ctx context.Context) error {
+	spanCtx, span := c.tracer.Start(ctx, "keep-in-sync-cluster-nodes")
+	span.SetAttributes(telemetry.WithClusterID(c.Id))
+	defer span.End()
+
+	// fetch cluster nodes with use of service discovery
+	res, err := c.httpClient.V1ServiceDiscoveryGetOrchestratorsWithResponse(c.ctx)
```
dobrac:
```suggestion
	res, err := c.httpClient.V1ServiceDiscoveryGetOrchestratorsWithResponse(spanCtx)
```

packages/api/internal/edge/cluster_nodes.go
```diff
@@ -0,0 +1,168 @@
...
+	timer := time.NewTicker(clusterNodesSyncInterval)
+	defer timer.Stop()
+
+	for {
+		select {
+		case <-c.ctx.Done():
+			zap.L().Info("Cluster nodes sync ended", l.WithClusterID(c.Id))
+			return
+		case <-timer.C:
+			syncTimeout, syncCancel := context.WithTimeout(c.ctx, clusterNodesSyncInterval)
```
dobrac: nit: I would prefer passing the context as argument and not rely on inner var, same applies to many other places that use `c.ctx`, but don't have to

packages/api/internal/edge/cluster_nodes.go
```diff
@@ -0,0 +1,168 @@
...
+
+func (c *Cluster) sync(ctx context.Context) error {
+	spanCtx, span := c.tracer.Start(ctx, "keep-in-sync-cluster-nodes")
+	span.SetAttributes(telemetry.WithClusterID(c.Id))
+	defer span.End()
+
+	// fetch cluster nodes with use of service discovery
+	res, err := c.httpClient.V1ServiceDiscoveryGetOrchestratorsWithResponse(c.ctx)
+	if err != nil {
+		zap.L().Error("Failed to get cluster nodes", zap.Error(err), l.WithClusterID(c.Id))
```
dobrac: this error is already logged, no need to do it twice

packages/api/internal/edge/cluster_nodes.go
```diff
@@ -0,0 +1,168 @@
...
+		return err
```
dobrac:
```suggestion
		return fmt.Errorf("failed to get cluster nodes: %w", err)
```

packages/api/internal/edge/cluster_nodes.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	if res.JSON200 == nil {
+		return errors.New("request to get builders returned nil response")
+	}
+
+	var wg sync.WaitGroup
+
+	// register newly discovered nodes
+	_, spanNewlyDiscovered := c.tracer.Start(spanCtx, "cluster-nodes-sync-newly-discovered")
```
dobrac: make this a separate function, where you can then defer spanNewlyDiscovered.end()

packages/api/internal/edge/cluster_nodes.go
```diff
@@ -0,0 +1,168 @@
...
+			continue
+		}
+
+		zap.L().Info("Adding new node into cluster nodes pool", l.WithClusterID(c.Id), l.WithClusterNodeID(sdNode.Id))
+		node := &ClusterNode{
+			Id:     sdNode.Id,
+			NodeId: sdNode.NodeId,
+
+			// initial values before first sync
+			Status: infogrpc.ServiceInfoStatus_OrchestratorDraining,
```
dobrac: Why draining and not unhealthy?

packages/api/internal/edge/cluster_nodes.go
```diff
@@ -0,0 +1,168 @@
...
+			wg.Done()
+		}()
+	}
+
+	// wait for all new nodes to be added
+	wg.Wait()
+	spanNewlyDiscovered.End()
+
+	// remove nodes that are no longer present in the service discovery
+	_, spanOutdatedNodes := c.tracer.Start(spanCtx, "cluster-nodes-sync-outdated-nodes")
```
dobrac: extract this to a separate function too

packages/api/internal/edge/cluster_nodes.go
```diff
@@ -0,0 +1,168 @@
...
+			c.syncNode(ctx, node)
+			wg.Done()
+		}()
+	}
+
+	// wait for all new nodes to be added
+	wg.Wait()
+	spanNewlyDiscovered.End()
+
+	// remove nodes that are no longer present in the service discovery
```
dobrac: is it possible that the service discovery will return empty list in some occasions even though there should be some nodes? (e.g. unvoluntary nil or something)

packages/api/internal/edge/cluster_nodes.go
```diff
@@ -0,0 +1,168 @@
...
+			}
+		}
+
+		// synchronize node state
+		if found {
+			wg.Add(1)
+			go func() {
+				c.syncNode(ctx, node)
+				wg.Done()
+			}()
```
dobrac: aren't the recently added nodes synchronized twice?

packages/api/internal/edge/pool.go
```diff
@@ -0,0 +1,163 @@
...
+		tracer: tracer,
+		pool:   smap.New[*Cluster](),
+	}
+
+	syncTimeout, syncCancel := context.WithTimeout(p.ctx, poolSyncInterval)
+	defer syncCancel()
+
+	err := p.sync(syncTimeout)
+	if err != nil {
+		zap.L().Error("Failed to initialize edge pool", zap.Error(err))
```
dobrac: do we need to log the errors at all like this? The error should be logged at the top level imo, we can here just pass:
fmt.Errorf("failed to initialize edge pool: %w", err)


(this applies to all other places, I don't want to add the comment everywhere)

packages/api/internal/edge/pool.go
```diff
@@ -0,0 +1,163 @@
...
+			wg.Done()
+		}(cluster)
+	}
+	spanForOutdated.End()
+
+	// wait for all not wanted clusters to be disconnected
+	wg.Wait()
+
+	return nil
+}
```
dobrac: this whole logic is really similar to the cluster sync. Can you make some abstraction and reuse it?

packages/api/internal/template-manager/template_manager_local.go
```diff
@@ -0,0 +1,43 @@
...
+		case <-ticker.C:
+			reqCtx, reqCtxCancel := context.WithTimeout(ctx, 5*time.Second)
+			res, err := tm.localClient.Info.ServiceInfo(reqCtx, &emptypb.Empty{})
+			reqCtxCancel()
+
+			err = utils.UnwrapGRPCError(err)
+			if err != nil {
+				zap.L().Error("Failed to get health status of template manager", zap.Error(err))
+				tm.localClientMutex.Lock()
+				tm.localClientStatus = orchestratorinfo.ServiceInfoStatus_OrchestratorDraining
```
dobrac: same locking comment as in the clusters, create a getter/setter for the status

packages/api/internal/template-manager/template_manager_local.go
```diff
@@ -0,0 +1,43 @@
...
+				tm.localClientMutex.Unlock()
+			}
+
+			tm.localClientMutex.Lock()
+			tm.localClientStatus = res.ServiceStatus
+			zap.L().Debug("Template manager health status", zap.String("status", tm.localClientStatus.String()))
```
dobrac: this log is probably unnecessary too spammy, I would remove it

packages/api/internal/template-manager/template_status.go
```diff
@@ -0,0 +1,249 @@
...
+	"github.com/pkg/errors"
+	"go.uber.org/zap"
+
+	templatemanagergrpc "github.com/e2b-dev/infra/packages/shared/pkg/grpc/template-manager"
+	"github.com/e2b-dev/infra/packages/shared/pkg/logger"
+	"github.com/e2b-dev/infra/packages/shared/pkg/models/envbuild"
+)
+
+func (tm *TemplateManager) BuildStatusSync(ctx context.Context, buildID uuid.UUID, templateID string, clusterId *uuid.UUID, clusterNodeId *string) {
+	childCtx, childCtxCancel := context.WithTimeout(ctx, syncTimeout)
```
dobrac, line 24: colocate the syncTimeout

packages/api/internal/template-manager/template_status.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+
+	// remove from processing queue when done
+	defer tm.removeFromProcessingQueue(buildID)
+
+	logger := zap.L().With(logger.WithBuildID(buildID.String()), logger.WithTemplateID(templateID))
+
+	envBuildDb, err := tm.db.GetEnvBuild(childCtx, buildID)
+	if err != nil {
+		logger.Error("Error when fetching env build for background sync", zap.Error(err))
```
dobrac: instead of logging all error codes, I would change the method to return "error" and then log the error somewhere top the call stack

packages/api/internal/template-manager/template_manager.go
```diff
@@ -39,22 +66,63 @@ const (
...
-func New(ctx context.Context, tracerProvider trace.TracerProvider, meterProvider metric.MeterProvider, db *db.DB, buildCache *templatecache.TemplatesBuildCache) (*TemplateManager, error) {
-	client, err := NewClient(ctx, tracerProvider, meterProvider)
+var (
+	templateManagerHost = os.Getenv("TEMPLATE_MANAGER_HOST")
+
+	ErrLocalTemplateManagerNotAvailable = errors.New("local template manager is not available")
+)
+
+func New(ctx context.Context, tracer trace.Tracer, tracerProvider trace.TracerProvider, meterProvider metric.MeterProvider, db *db.DB, sqlcDB *sqlcdb.Client, edgePool *edge.Pool, lokiClient *loki.DefaultClient, buildCache *templatecache.TemplatesBuildCache) (*TemplateManager, error) {
+	conn, err := grpc.NewClient(templateManagerHost,
```
dobrac: why move the grpc client creation from separate function here?

packages/api/internal/template-manager/template_manager.go
```diff
@@ -39,22 +66,63 @@ const (
...
+	}
+
+	// periodically check for local template manager health status
+	go tm.localBuilderHealthCheckSync(ctx)
+
+	return tm, nil
 }
 
 func (tm *TemplateManager) Close() error {
+	// todo: we should track all in-progress requests and wait for them to finish
```
dobrac: doesn't grpc.Close() already do that?

packages/api/internal/template-manager/template_manager.go
```diff
@@ -66,252 +134,179 @@ func (tm *TemplateManager) BuildsStatusPeriodicalSync(ctx context.Context) {
...
-	defer tm.lock.Unlock()
+	telemetry.ReportEvent(ctx, "Template build started")
+	return nil
+}
 
-	_, exists := tm.processing[buildID]
-	if exists {
-		// already in processing queue, skip
-		return true
+func (tm *TemplateManager) GetStatus(ctx context.Context, buildId uuid.UUID, templateId string, clusterId *uuid.UUID, clusterNodeId *string) (*tempaltemanagergrpc.TemplateBuildStatusResponse, error) {
```
dobrac:
```suggestion
func (tm *TemplateManager) GetStatus(ctx context.Context, buildID uuid.UUID, templateID string, clusterID *uuid.UUID, clusterNodeID *string) (*tempaltemanagergrpc.TemplateBuildStatusResponse, error) {
```

packages/api/internal/template-manager/template_manager.go
```diff
@@ -66,252 +134,179 @@ func (tm *TemplateManager) BuildsStatusPeriodicalSync(ctx context.Context) {
...
-func (c *PollBuildStatus) checkBuildStatus(ctx context.Context) (error, bool) {
-	c.logger.Info("Checking template build status")
+	return nil
+}
 
-	retrier := retry.NewRetrier(
-		10,
-		100*time.Millisecond,
-		time.Second,
+func (tm *TemplateManager) CreateTemplate(t trace.Tracer, ctx context.Context, templateID string, buildID uuid.UUID, kernelVersion, firecrackerVersion, startCommand string, vCpuCount, diskSizeMB, memoryMB int64, readyCommand string, clusterId *uuid.UUID, clusterNodeId *string) error {
```
dobrac:
```suggestion
func (tm *TemplateManager) CreateTemplate(t trace.Tracer, ctx context.Context, templateID string, buildID uuid.UUID, kernelVersion, firecrackerVersion, startCommand string, vCpuCount, diskSizeMB, memoryMB int64, readyCommand string, clusterID *uuid.UUID, clusterNodeId *string) error {
```

packages/api/internal/template-manager/template_manager.go
```diff
@@ -66,252 +134,179 @@ func (tm *TemplateManager) BuildsStatusPeriodicalSync(ctx context.Context) {
...
-	return e.err.Error()
+	return client, clientMetadata, logs, nil
 }
 
-func newTerminalError(err error) error {
-	return terminalError{
-		err: retry.Stop(errors.WithStack(err)),
-	}
-}
+func (tm *TemplateManager) DeleteBuild(ctx context.Context, t trace.Tracer, buildId uuid.UUID, templateId string, clusterId *uuid.UUID, clusterNodeId *string) error {
```
dobrac:
```suggestion
func (tm *TemplateManager) DeleteBuild(ctx context.Context, t trace.Tracer, buildID uuid.UUID, templateID string, clusterID *uuid.UUID, clusterNodeID *string) error {
```

packages/api/internal/template-manager/template_manager.go
```diff
@@ -66,252 +134,179 @@ func (tm *TemplateManager) BuildsStatusPeriodicalSync(ctx context.Context) {
...
+		&tempaltemanagergrpc.TemplateStatusRequest{
+			BuildID: buildId.String(), TemplateID: templateId,
+		},
+	)
 }
 
-func (tm *TemplateManager) SetFinished(ctx context.Context, templateID string, buildID uuid.UUID, rootfsSize int64, envdVersion string) error {
-	// first do database update to prevent race condition while calling status
-	err := tm.db.FinishEnvBuild(ctx, templateID, buildID, rootfsSize, envdVersion)
+func (tm *TemplateManager) GetLogs(ctx context.Context, buildId uuid.UUID, templateId string, clusterId *uuid.UUID, clusterNodeId *string, offset *int32) (*[]string, error) {
```
dobrac:
```suggestion
func (tm *TemplateManager) GetLogs(ctx context.Context, buildID uuid.UUID, templateID string, clusterID *uuid.UUID, clusterNodeID *string, offset *int32) (*[]string, error) {
```

dobrac: why returning *[]string? You can return []string instead, []string can be nil

dobrac (changes requested):
I would mainly unite the syncing logic, as it's there now twice, but doing basically the same thing

packages/api/internal/edge/cluster.go
sitole: By default, AWS load balancer with TLS is using version 12. I can manually adjust it, just important to keep it in mind

packages/api/internal/edge/cluster.go
sitole: Node used (we were using just the template manager and info clients. It's initialized here (https://github.com/e2b-dev/infra/pull/831).

packages/api/internal/edge/cluster.go
dobrac: Oh I see, thats fine probably then, just add a comment there if thats the dependency

packages/api/internal/edge/cluster.go
sitole: You mean `Orchestrator` be part of enum name?

packages/api/internal/edge/cluster_nodes.go
sitole: Where's it logged? Propably we should change it to something like `Failed to get cluster nodes from service discovery`

sitole: okay, changed message and returning err without adding another log

dobrac: on line 53 in this file

packages/api/internal/template-manager/template_status.go
sitole: This method is expected to run in a goroutine, so during initial we can add an anonymous go function and catch the returned error, but not sure if it will be cleaner. In this exact case, where function is called in the background.

packages/api/internal/edge/cluster.go
sitole: You want to ideally know how long it takes to get an available builder per cluster.

dobrac: but this will track all spans to the parent context/span, not span per `GetAvailableTemplateBuilder` invocation

packages/api/internal/edge/cluster_nodes.go
sitole: true, removing

packages/api/internal/edge/cluster.go
sitole: Maybe the comment is just confusing. We are creating a ctx with a cancel for each sandbox. Then, during cluster close, ctx cancel is called, so all in-progress requests and synchronizations are shut down.

packages/api/internal/edge/cluster.go
dobrac: yeah

packages/api/internal/edge/cluster_nodes.go
sitole: For internal methods, it makes sense, but for public methods, we want to have our per-cluster or per-node ctx so during cluster/node close, it will be cancelled properly.

packages/api/internal/template-manager/template_status.go
dobrac: yes, but it would be ideally up to the caller, not the implementation how it's called

packages/api/internal/edge/cluster.go
dobrac: I understand the reason, it's just the comment sentence doesn't make much sense to me, nor explain fully the reason behind it

packages/api/internal/edge/cluster_nodes.go
sitole: Theoretically, if the service discovery provider returns an empty list because of some internal issue, then yes, but then we don't know the difference between there are no nodes vs it's just an error that returns an empty list.

packages/api/internal/template-manager/template_manager.go
sitole: can move it back, but moved as part of rebasing from global removal of our grpc client creation

packages/api/internal/template-manager/template_status.go
dobrac: I think if you try to test this method would clarify it a bit

packages/api/internal/edge/cluster.go
sitole: okay, will remove it

packages/api/internal/template-manager/template_manager.go
dobrac: as the init of grpc is not just a single line/function call, I would split it to at least some helper function. Another approach might be to do it all in one/two line/function calls

packages/api/internal/edge/cluster.go
sitole: If ctx provided does not have span then it should be just per function call right?

packages/api/internal/template-manager/template_manager.go
sitole: done, make sense

packages/api/internal/edge/cluster_nodes.go
sitole: `c.ctx` removed and replaced with channel

packages/api/internal/edge/cluster_nodes.go
sitole: resolved with a new utility for synchronization

packages/api/internal/edge/cluster_nodes.go
sitole: resolved with a new utility for synchronization

packages/api/internal/edge/pool.go
sitole: introduces new utility for synchronization, solving this

packages/api/internal/edge/cluster_nodes.go
dobrac: thats probably ok

packages/api/internal/handlers/template_build_status.go
```diff
@@ -93,51 +83,12 @@ func (a *APIStore) GetTemplatesTemplateIDBuildsBuildIDStatus(c *gin.Context, tem
...
-				line := make(map[string]interface{})
-				err := json.Unmarshal([]byte(entry.Line), &line)
-				if err != nil {
-					zap.L().Error("error parsing log line", zap.Error(err), logger.WithBuildID(buildID), zap.String("line", entry.Line))
-				}
-
-				logs = append(logs, line["message"].(string))
-			}
-		}
+	l, err := a.templateManager.GetLogs(ctx, buildUUID, templateID, team.ClusterID, buildInfo.ClusterNodeId, params.LogsOffset)
```
dobrac:
```suggestion
	l, err := a.templateManager.GetLogs(ctx, buildUUID, templateID, team.ClusterID, buildInfo.ClusterNodeID, params.LogsOffset)
```

packages/api/internal/template-manager/logs.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
+)
+
+const (
+	lokiTemplateBuildLogsLimit       = 1_000
+	lokiTemplateBuildOldestLogsLimit = 24 * time.Hour // 1 day
+)
+
+type PlacementLogsProvider interface {
+	GetLogs(ctx context.Context, buildId string, templateId string, offset *int32) ([]string, error)
```
dobrac, line 27:
```suggestion
	GetLogs(ctx context.Context, templateID string, buildID string, offset *int32) ([]string, error)
```

packages/api/internal/template-manager/template_manager.go
```diff
@@ -2,56 +2,93 @@ package template_manager
...
+	localClientMutex  sync.RWMutex
+	localClientStatus infogrpc.ServiceInfoStatus
+}
+
+type DeleteBuild struct {
+	BuildID    uuid.UUID
+	TemplateId string
+
+	ClusterId     *uuid.UUID
+	ClusterNodeId *string
```
dobrac:
```suggestion
	TemplateID string

	ClusterID     *uuid.UUID
	ClusterNodeID *string
```

packages/api/internal/template-manager/template_manager.go
```diff
@@ -2,56 +2,93 @@ package template_manager
...
 		db:         db,
+		sqlcDB:     sqlcDB,
+		tracer:     tracer,
+		buildCache: buildCache,
+		edgePool:   edgePool,
+		lokiClient: lokiClient,
+
+		localClient:       client,
+		localClientMutex:  sync.RWMutex{},
+		localClientStatus: infogrpc.ServiceInfoStatus_OrchestratorHealthy,
```
dobrac: is it always healthy when you add it?

packages/api/internal/template-manager/template_manager.go
```diff
@@ -66,252 +103,184 @@ func (tm *TemplateManager) BuildsStatusPeriodicalSync(ctx context.Context) {
...
-	logger := zap.L().With(logger.WithBuildID(buildID.String()), logger.WithTemplateID(templateID))
-
-	envBuildDb, err := tm.db.GetEnvBuild(childCtx, buildID)
-	if err != nil {
-		logger.Error("Error when fetching env build for background sync", zap.Error(err))
-		return
-	}
+func (tm *TemplateManager) getBuilderClient(clusterID *uuid.UUID, nodeID *string, placement bool) (*grpclient.GRPCClient, metadata.MD, PlacementLogsProvider, error) {
+	if clusterID == nil || nodeID == nil {
+		tm.localClientMutex.RLock()
```
dobrac, line 130: maybe put the RUnlock to defer here?

packages/api/internal/template-manager/template_manager.go
```diff
@@ -66,252 +103,184 @@ func (tm *TemplateManager) BuildsStatusPeriodicalSync(ctx context.Context) {
...
+		),
+	)
+	defer span.End()
+
+	_, _, logs, err := tm.getBuilderClient(clusterID, clusterNodeID, false)
 	if err != nil {
-		tm.buildCache.SetStatus(buildID, envbuild.StatusFailed, "error when finishing build")
-		return err
+		emptyLogs := make([]string, 0)
+		return emptyLogs, fmt.Errorf("failed to get builder edgeHttpClient: %w", err)
```
dobrac:
```suggestion
		return nil, fmt.Errorf("failed to get builder edgeHttpClient: %w", err)
```

packages/api/internal/template-manager/template_manager_local.go
```diff
@@ -0,0 +1,50 @@
...
+			return
+		case <-ticker.C:
+			reqCtx, reqCtxCancel := context.WithTimeout(ctx, 5*time.Second)
+			res, err := tm.localClient.Info.ServiceInfo(reqCtx, &emptypb.Empty{})
+			reqCtxCancel()
+
+			err = utils.UnwrapGRPCError(err)
+			if err != nil {
+				zap.L().Error("Failed to get health status of template manager", zap.Error(err))
+				tm.setLocalClientStatus(orchestratorinfo.ServiceInfoStatus_OrchestratorDraining)
```
dobrac: draining or unhealthy?

packages/shared/pkg/synchronization/synchronization_test.go
```diff
@@ -0,0 +1,123 @@
+package synchronization
```
dobrac, line 1: nice ❤️

packages/api/internal/edge/pool.go
```diff
@@ -0,0 +1,145 @@
...
+	// Shutdown function to gracefully close the pool
+	go func() {
+		<-ctx.Done()
+		p.Close()
+	}()
+
+	return p, nil
+}
+
+func (p *Pool) syncBackground() {
```
dobrac: it should be probably something like start, etc. The function doesn't (and shouldn't) know if it's gonna be run in the background or not

packages/api/internal/edge/pool.go
```diff
@@ -0,0 +1,145 @@
...
+	synchronize := synchronization.Synchronize[queries.GetActiveClustersRow, string, *Cluster]{
+		Tracer:           p.tracer,
+		TracerSpanPrefix: "clusters-pool",
+		LogsPrefix:       "Clusters pool",
+		Store:            poolSynchronizationStore{pool: p},
+	}
+
+	synchronize.SyncInBackground(p.close, poolSyncInterval, poolSyncTimeout, true)
```
dobrac: same for this name, the SyncInBackground runs it actually in the foreground

packages/shared/pkg/synchronization/synchronization.go
```diff
@@ -0,0 +1,145 @@
...
+		return err
+	}
+
+	s.runSyncDiscovered(spanCtx, sourceItems)
+	s.runSyncOutdated(spanCtx, sourceItems)
+
+	return nil
+}
+
+func (s *Synchronize[SourceItem, SourceKey, PoolItem]) SyncInBackground(cancel chan struct{}, syncInterval time.Duration, syncRoundTimeout time.Duration, runInitialSync bool) {
```
dobrac: I don't like the cancel channel passing here. I would prefer separate method to close/stop the loop, so the cancel will be inner property of the Synchronize type

packages/api/internal/template-manager/template_manager.go
sitole: added initial health check, default value set to unhealthy

sitole:
- Make consisten variable naming `templateId` -> `templateID` in all newly written logic and logic that we are touching here
- API is now using `NodeID ` that represents VM machine identification instead of `ServiceInstanceID` that represents edge/orchestrator process UUID. This is mainly because API is alrady using `NodeID` for orchestrators pool so we want to stay consistent
- Renamed edge API interface to better communicate differences between node id and service instance ids.
- Changed implementation of synchronization util that does not require map

packages/api/internal/edge/cluster.go
```diff
@@ -24,8 +25,9 @@ type Cluster struct {
 	httpClient *api.ClientWithResponses
 	grpcClient *grpclient.GRPCClient
 
-	nodes  *smap.Map[*ClusterNode]
-	tracer trace.Tracer
+	nodes           *smap.Map[*ClusterNode]
+	synchronization *synchronization.Synchronize[api.ClusterOrchestratorNode, *ClusterNode]
+	tracer          trace.Tracer
 
 	close chan struct{}
```
dobrac: is this channel used after the changes?

packages/api/internal/edge/cluster_nodes.go
```diff
@@ -113,46 +105,56 @@ func (d clusterSynchronizationStore) SourceList(ctx context.Context) ([]api.Clus
 	return *res.JSON200, nil
 }
 
-func (d clusterSynchronizationStore) SourceKey(item api.ClusterOrchestratorNode) string {
-	return item.Id
+func (d clusterSynchronizationStore) SourceExists(ctx context.Context, s []api.ClusterOrchestratorNode, p *ClusterNode) bool {
```
dobrac, line 108: is the clusterSynchronizationStore receiver not a pointer by design?

packages/api/internal/edge/pool.go
```diff
@@ -73,6 +72,8 @@ func (p *Pool) GetClusterById(id uuid.UUID) (*Cluster, bool) {
 }
 
 func (p *Pool) Close() {
+	p.synchronization.Close()
+
 	// Close pool, this needs to be called before closing the clusters
 	// so background jobs syncing cluster will not try to update the pool nodes
 	close(p.close)
```
dobrac: is this still used now?

packages/api/internal/edge/pool.go
```diff
@@ -101,45 +102,56 @@ func (d poolSynchronizationStore) SourceList(ctx context.Context) ([]queries.Get
 	return d.pool.db.GetActiveClusters(ctx)
 }
 
-func (d poolSynchronizationStore) SourceKey(item queries.GetActiveClustersRow) string {
-	return item.Cluster.ID.String()
+func (d poolSynchronizationStore) SourceExists(ctx context.Context, s []queries.GetActiveClustersRow, p *Cluster) bool {
```
dobrac, line 98: is the poolSynchronizationStore receiver not a pointer by design?

packages/shared/pkg/synchronization/synchronization.go
```diff
@@ -75,71 +74,74 @@ func (s *Synchronize[SourceItem, SourceKey, PoolItem]) SyncInBackground(cancel c
 	}
 }
 
-func (s *Synchronize[SourceItem, SourceKey, PoolItem]) runSyncDiscovered(ctx context.Context, sourceItems []SourceItem) {
-	spanCtx, span := s.Tracer.Start(ctx, s.getSpanName("sync-discovered-items"))
+func (s *Synchronize[SourceItem, PoolItem]) Close() {
+	close(s.cancel)
```
dobrac: calling close twice will cause panic, how about using atomic bool to guard it? (we should have already some examples of that in the codebase)

packages/api/internal/edge/cluster_nodes.go
sitole: nope for reason, fixing. thx!

packages/shared/pkg/synchronization/synchronization.go
sitole: will add `sync.Once`

packages/api/internal/edge/pool.go
sitole: removing, thx

packages/api/internal/edge/cluster_nodes.go
sitole: no need for that, structure should be self sustaining and different invocations should not do changes on shared state

dobrac (approved):

PR #769, opened 2025-06-13
Remove doubled Redis

jakubno:
# Description

This was there only for migration purposes, it's safe to remove now

### DIFF
--- a/packages/api/internal/dns/server.go
+++ b/packages/api/internal/dns/server.go
@@ -75,7 +75,7 @@ func (d *DNS) Add(ctx context.Context, sandboxID, ip string) {
 			Value: ip,
 		})
 		if err != nil {
-			zap.L().Warn("error adding DNS item to redis cache", zap.Error(err), logger.WithSandboxID(sandboxID))
+			zap.L().Error("error adding DNS item to redis cache", zap.Error(err), logger.WithSandboxID(sandboxID))
 		}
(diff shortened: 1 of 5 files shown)


packages/api/internal/dns/server.go
```diff
@@ -123,16 +100,6 @@ func (d *DNS) Get(ctx context.Context, sandboxID string) net.IP {
 		if err := d.redisCache.Get(ctx, d.cacheKey(sandboxID), &res); err != nil {
 			if errors.Is(err, cache.ErrCacheMiss) {
 				zap.L().Warn("item missing in redisCache DNS cache", logger.WithSandboxID(sandboxID))
```
dobrac: not fully related to this PR, but do we need to log this? It's quite spammy and I'm not sure if it adds any value

packages/api/internal/orchestrator/orchestrator.go
```diff
@@ -45,16 +46,42 @@ func New(
...
+	}
+
+	if redisClient != nil {
+		_, err := redisClient.Ping(ctx).Result()
+		if err != nil {
+			zap.L().Fatal("could not connect to Redis", zap.Error(err))
+		}
+
+		zap.L().Info("connected to Redis cluster")
+	}
```
dobrac: why move this inside? It should be a dependency imo

PR #771, opened 2025-06-13
Fix nodes sync race condition

dobrac:
Fixes a race condition, where the sandbox could be starting, but not yet returned from the orchestrator, causing immediate eviction and sandbox stop. This will result into 502 (sandbox not running/timeout) for the user.

All sandboxes that are started won't be synchronized from the orchestrator in a 10 second period. This is not optimal fix, but should be ok for now until we resolve the issue in general.

### DIFF
--- a/packages/api/internal/cache/instance/sync.go
+++ b/packages/api/internal/cache/instance/sync.go
@@ -13,9 +13,9 @@ import (
 )
 
 // TODO: this should be removed once we have a better way to handle node sync
-// Don't sync instances that were started in the grace period on node sync
-// This is to prevent add/remove instances that are still being started
-const syncSandboxGracePeriod = 10 * time.Second
+// Don't remove instances that were started in the grace period on node sync
+// This is to prevent remove instances that are still being started
+const syncSandboxRemoveGracePeriod = 10 * time.Second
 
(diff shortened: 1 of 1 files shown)


packages/api/internal/cache/instance/sync.go
```diff
@@ -71,11 +79,15 @@ func (c *InstanceCache) Sync(ctx context.Context, instances []*InstanceInfo, nod
...
 	// Add instances that are not in the cache with the default TTL
 	for _, instance := range instances {
-		if !c.Exists(instance.Instance.SandboxID) {
-			err := c.Add(ctx, instance, false)
-			if err != nil {
-				zap.L().Error("error adding instance to cache", zap.Error(err))
-			}
+		if time.Since(instance.StartTime) <= syncSandboxGracePeriod {
+			continue
+		}
```
jakubno: why this part? It's weird, but this could cause other problems e.g. newly started sandboxes wouldn't get loaded to API after rolling update resulting in problems with connecting to the sandbox (because client proxy asks API for the routing)

dobrac: Reverted

jakubno (approved):

PR #773, opened 2025-06-14
Remove sandbox other

jakubno:
# Description

Clean up the `sandbox_other.go` as it is unused and out of date. We probably want to abstract the linux dependent stuff (network, nbd) and not whole create sandbox anyway


ValentaTomas:
I was also thinking that we could just use devcontainers for this. Does jetbrains support it?

jakubno:
for what exactly you mean, developing on non linux machine? I don't think it's such a problem, you can develop on remove that's fine by me. I don't think abstracting nbd and networking just for allowing to build orchestrator on non linux machine has priority right now

PR #776, opened 2025-06-14
Update confirm script

jakubno:
# Description

Deployments should be now done from main

### DIFF
--- a/scripts/confirm.sh
+++ b/scripts/confirm.sh
@@ -5,9 +5,8 @@ ENV=$1
 BRANCH=$(git rev-parse --abbrev-ref HEAD)
 # Check if the ENV variable is set to "prod"
 if [[ "$ENV" == prod* ]]; then
-  # Replace prod with e2b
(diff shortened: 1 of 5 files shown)


dobrac (changes requested):
Update also the deployment pipeline with some arg like `-y`, so it doesn't fail

scripts/confirm.sh
```diff
@@ -1,17 +1,35 @@
...
-  ENV=$(echo $ENV | sed 's/prod/e2b/')
-  if [ "$ENV" != "$BRANCH" ]; then
+if [[ "$ENV" != "dev" ]]; then
+  # Check if the current branch is "main"
+  if [ "$BRANCH" != "main" ]; then
     echo "You are trying to deploy to $ENV from $BRANCH"
     exit 1
   fi
 
+  if [[ "$AUTO_CONFIRM_DEPLOY" == "true" ]]; then
```
dobrac: this enforces that every pipeline deployment is from main, this could be an issue if quick bugfix is necessary

PR #778, opened 2025-06-14
Setup Launchdarkly for clickhouse

jakubno:
# Description

Allow dynamically setting if the metrics should be sent to ClickHouse.


packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -406,6 +407,8 @@ func ResumeSandbox(
 	if err != nil {
 		return nil, cleanup, fmt.Errorf("failed to create health check: %w", err)
 	}
+
+	zap.L().Info("created health check", logger.WithSandboxID(config.SandboxId), zap.Bool("use_clickhouse_metrics", useClickhouseMetrics))
```
dobrac: move this to NewChecks?

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -61,7 +69,7 @@ func (s *server) Create(ctxConn context.Context, req *orchestrator.SandboxCreate
 		req.Sandbox.BaseTemplateId,
 		s.devicePool,
 		config.AllowSandboxInternet,
-		config.WriteClickhouseMetrics,
+		flag,
```
dobrac: will the flag have a default value if the init soft fails?

packages/orchestrator/main.go
```diff
@@ -236,7 +237,12 @@ func run(port, proxyPort uint) (success bool) {
 
 	grpcSrv := grpcserver.New(tel.TracerProvider, tel.MeterProvider, serviceInfo)
 
-	_, err = server.New(ctx, grpcSrv, tel, networkPool, devicePool, tracer, serviceInfo, sandboxProxy, sandboxes)
+	featureFlags, err := featureflags.NewClient(5 * time.Second)
```
dobrac: move the 5*Second to const

dobrac: shouldn't the client be also closed?

packages/orchestrator/internal/server/sandboxes.go
dobrac: should this flag passing be also done for a sandbox create (and not only resume)?

sitole: yes, work like that

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -61,7 +69,7 @@ func (s *server) Create(ctxConn context.Context, req *orchestrator.SandboxCreate
 		req.Sandbox.BaseTemplateId,
 		s.devicePool,
 		config.AllowSandboxInternet,
-		config.WriteClickhouseMetrics,
```
sitole, line 64: Can we maybe rename var to something more obvious that just `flag`?

packages/orchestrator/main.go
jakubno: it will closed, line 264

packages/orchestrator/internal/server/sandboxes.go
jakubno: we don't support metrics for create as it is tight to sandbox id and build doesn't have sandbox id.

We can add metrics, but not sure if needed and I wouldn't complicate it now

packages/orchestrator/main.go
dobrac: I've meant the featureFlags client

jakubno: image

dobrac: oh I see, thx

PR #782, opened 2025-06-16
Run the template build on the infrastructure

dobrac:
Add basic ability to run template building on the infrastructure directly, instead of using local Docker.

**This PR adds the following endpoints**
- POST /v2/templates
- GET /templates/{templateID}/files/{hash}
- POST /v2/templates/{templateID}/builds/{buildID}

The flow is described here: https://www.notion.so/e2bdev/API-Communication-Spec-20db8c29687380ab802be1a6f23eabdd?source=copy_link

**Required changes**
In order to use the new builder, you need to first run `make build-engine-template` in the `packages/shared` package. The build engine Docker image is available under `e2bdev/build-engine`.

If the engine is not built or no steps are provided, you can still use the previous way where the already built Docker image is supplied.

### DIFF
--- a/packages/shared/pkg/storage/storage_aws.go
+++ b/packages/shared/pkg/storage/storage_aws.go
@@ -78,7 +78,7 @@ func (a *AWSBucketStorageProvider) GetDetails() string {
 	return fmt.Sprintf("[AWS Storage, bucket set to %s]", a.bucketName)
 }
 
-func (a *AWSBucketStorageProvider) SignedURL(ctx context.Context, path string, ttl time.Duration) (string, error) {
+func (a *AWSBucketStorageProvider) UploadSignedURL(ctx context.Context, path string, ttl time.Duration) (string, error) {
 	input := &s3.PutObjectInput{
 		Bucket: aws.String(a.bucketName),
 		Key:    aws.String(path),
--- a/packages/shared/pkg/storage/storage_fs.go
+++ b/packages/shared/pkg/storage/storage_fs.go
@@ -37,7 +37,7 @@ func (fs *FileSystemStorageProvider) GetDetails() string {
 	return fmt.Sprintf("[Local file storage, base path set to %s]", fs.basePath)
 }
 
-func (fs *FileSystemStorageProvider) SignedURL(_ context.Context, _ string, _ time.Duration) (string, error) {
+func (fs *FileSystemStorageProvider) UploadSignedURL(_ context.Context, _ string, _ time.Duration) (string, error) {
 	return "", fmt.Errorf("file system storage does not support signed URLs")
 }
 
--- a/packages/orchestrator/internal/template/server/upload_layer_files_template.go
+++ b/packages/orchestrator/internal/template/server/upload_layer_files_template.go
@@ -27,7 +27,7 @@ func (s *ServerStore) TemplateLayerFilesUpload(ctx context.Context, in *template
 	if err != nil {
 		zap.L().Warn("layer files not found", zap.Error(err))
 
-		signedUrl, err := s.storage.SignedURL(ctx, path, signedUrlExpiration)
+		signedUrl, err := s.storage.UploadSignedURL(ctx, path, signedUrlExpiration)
 		if err != nil {
 			return nil, fmt.Errorf("failed to get signed url: %w", err)
 		}
--- a/packages/shared/pkg/storage/storage.go
+++ b/packages/shared/pkg/storage/storage.go
@@ -27,7 +27,7 @@ const (
 
 type StorageProvider interface {
 	DeleteObjectsWithPrefix(ctx context.Context, prefix string) error
-	SignedURL(ctx context.Context, path string, ttl time.Duration) (string, error)
+	UploadSignedURL(ctx context.Context, path string, ttl time.Duration) (string, error)
 	OpenObject(ctx context.Context, path string) (StorageObjectProvider, error)
 	GetDetails() string
 }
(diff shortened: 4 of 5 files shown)


spec/openapi.yml
```diff
@@ -1177,6 +1250,66 @@ paths:
...
+              schema:
+                $ref: "#/components/schemas/Template"
+        "400":
+          $ref: "#/components/responses/400"
+        "401":
+          $ref: "#/components/responses/401"
+        "500":
+          $ref: "#/components/responses/500"
+
+  /templates/{templateID}/files/{hash}:
```
sitole, line 1281: Maybe it makes sense to prefix this one also with V2, because it's part of the new feature that comes by V2?

packages/shared/pkg/storage/storage_fs.go
```diff
@@ -36,6 +37,10 @@ func (fs *FileSystemStorageProvider) GetDetails() string {
 	return fmt.Sprintf("[Local file storage, base path set to %s]", fs.basePath)
 }
 
+func (fs *FileSystemStorageProvider) SignedURL(_ context.Context, _ string, _ time.Duration) (string, error) {
```
sitole: How will the testing of the new build system work in integration tests/environments without object storage?

spec/openapi.yml
dobrac: I've made v2 only the endpoints, that have version two (we have the same example with sandboxes list), this one is new, so no v2 required here. V2 has nothing to do with the Build System V2, but I understand the question

packages/shared/pkg/storage/storage_aws.go
```diff
@@ -75,6 +78,20 @@ func (a *AWSBucketStorageProvider) GetDetails() string {
 	return fmt.Sprintf("[AWS Storage, bucket set to %s]", a.bucketName)
 }
 
+func (a *AWSBucketStorageProvider) SignedURL(ctx context.Context, path string, ttl time.Duration) (string, error) {
```
sitole: Can we maybe name it with something like UploadSignedURL?

packages/orchestrator/template-manager.proto
```diff
@@ -75,4 +113,7 @@ service TemplateService {
 
   // todo (2025-05): this is deprecated, please use InfoService that is used for both orchestrator and template manager
   rpc HealthStatus (google.protobuf.Empty) returns (HealthStatusResponse);
+
+  // TemplateLayerFilesUpload requests an upload URL for a file to be cached for the template build.
+  rpc TemplateLayerFilesUpload (TemplateLayerFilesUploadRequest) returns (TemplateLayerFilesUploadResponse);
```
sitole, line 118: This will basically run for every file? So if I have a folder with hundreds 10s or 100s of files, it will run this many times? Of course, it will be ideally just for the first build, but still.

Can we not dot it array at least? Maybe it will not be sent in batches now, but just make the interface prepared for that.

packages/shared/scripts/buildengine/e2b.toml
```diff
@@ -0,0 +1,19 @@
...
+# Python SDK
+# from e2b import Sandbox, AsyncSandbox
+# sandbox = Sandbox("e2b-build-engine") # Sync sandbox
+# sandbox = await AsyncSandbox.create("e2b-build-engine") # Async sandbox
+
+# JS SDK
+# import { Sandbox } from 'e2b'
+# const sandbox = await Sandbox.create('e2b-build-engine')
+
+template_id = "p9kw2u9cc1zj1cov2zru"
```
sitole, line 13: How will this work on different clusters?

sitole (commented):
Just some quick look nits for now

packages/orchestrator/template-manager.proto
dobrac: This runs for every layer, every layer is one tar file

packages/shared/scripts/buildengine/e2b.toml
dobrac: the idea is that it's the same as the base image. Then it's references by the alias. Do you think there will be any issues with that?

dobrac: I've added the templateID also to the seed of a new cluster. Maybe some migration path will be necessary there for existing clusters, not sure though

packages/shared/pkg/storage/storage_fs.go
dobrac: Good question. I was thinking that it could return instead of `https://....` something like `file://path..`, and then we would handle just putting the file to that path. I think this is ok for the integration tests, etc.

Another way might be to create some API endpoint of our own, where you could upload the file through url (same as for the buckets) and we'll handle the file save.

As we are not using this yet, I haven't really implemented it, but these are my ideas

packages/orchestrator/template-manager.proto
dobrac: also, the idea of running it in multiple requests is that it can be parallelized

packages/api/internal/handlers/template_layer_files_upload.go
```diff
@@ -0,0 +1,53 @@
...
+	"go.opentelemetry.io/otel/attribute"
+
+	"github.com/e2b-dev/infra/packages/api/internal/api"
+	"github.com/e2b-dev/infra/packages/db/queries"
+	"github.com/e2b-dev/infra/packages/shared/pkg/models/env"
+	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
+	"github.com/e2b-dev/infra/packages/shared/pkg/utils"
+)
+
+func (a *APIStore) GetTemplatesTemplateIDFilesHash(c *gin.Context, templateID api.TemplateID, hash string) {
```
ValentaTomas, line 18: Create a separate bucket for the file uploads.

spec/openapi.yml
```diff
@@ -1177,6 +1250,66 @@ paths:
...
+    get:
+      description: Get an upload link for a file that should be uploaded to the template build cache
```
ValentaTomas, line 1283: File or files? Is this the tar file with files?

PR #783, opened 2025-06-16
Improve Otel Collector - Add node name, drop unused labels

jakubno:
# Description

- Adds **node name** from GCP node 
image

instead of 
image
- Adds clickhouse otel agent to add the info for those metrics as well. It can't be in the main definition, because that gets deployed to all nodes and then it errors out (if you use localhost) and we can't use `*.service.consul`, because you would get incorrect host info.
- Move the config to separate file to improve readability and take advantage of tools for yaml
- Drop unused labels


packages/nomad/otel-collector.hcl
```diff
@@ -63,241 +63,14 @@ job "otel-collector" {
       }
 
       resources {
-        memory_max = ${memory_mb * 1.5}
-        memory     = ${memory_mb}
-        cpu        = ${cpu_count * 1000}
+        memory     = 4096
+        cpu        = 2000
```
dobrac: why is this not a variable anymore?

dobrac (changes requested):

dobrac (approved):

PR #784, opened 2025-06-16
Create internal sandbox event endpoint e2b 2486

0div:


packages/orchestrator/internal/server/main.go
```diff
@@ -61,6 +62,7 @@ func New(
 	tracer trace.Tracer,
 	info *service.ServiceInfo,
 	proxy *proxy.SandboxProxy,
+	eventServer *event.EventServer,
```
dobrac: unused

packages/orchestrator/cmd/build-template/main.go
```diff
@@ -29,7 +29,10 @@ import (
 	"github.com/e2b-dev/infra/packages/shared/pkg/storage"
 )
 
-const proxyPort = 5007
+const (
+	proxyPort          = 5007
+	sbxEventServerPort = 5010
```
dobrac: is this used?

packages/orchestrator/main.go
```diff
@@ -218,6 +222,25 @@ func run(port, proxyPort uint) (success bool) {
 		zap.L().Fatal("failed to create sandbox proxy", zap.Error(err))
 	}
 
+	var redisClient redis.UniversalClient
+	if redisClusterUrl := os.Getenv("REDIS_CLUSTER_URL"); redisClusterUrl != "" {
```
dobrac, line 227: do we want to add Redis dependency to the orchestrators?

dobrac, line 227: the same init is in the API, abstract to shared

packages/orchestrator/main.go
```diff
@@ -218,6 +222,25 @@ func run(port, proxyPort uint) (success bool) {
...
+			Addr:         redisUrl,
+			MinIdleConns: 1,
+		})
+	} else {
+		zap.L().Fatal("REDIS_URL not set")
+	}
+
+	sbxEventHandlers := event.NewEventHandlers(redisClient)
+
+	sbxEventServer := event.NewEventServer(sbxEventServerPort, sbxEventHandlers)
```
dobrac: I don't see proper cleanup for these two

packages/orchestrator/main.go
```diff
@@ -218,6 +222,25 @@ func run(port, proxyPort uint) (success bool) {
 		zap.L().Fatal("failed to create sandbox proxy", zap.Error(err))
 	}
 
+	var redisClient redis.UniversalClient
```
dobrac, line 226: missing client cleanup on stop

packages/orchestrator/internal/template/build/rootfs.go
```diff
@@ -197,6 +197,7 @@ ExecStart=-/sbin/agetty --noissue --autologin root %I 115200,38400,9600 vt102
 `
 
 	hostname := "e2b.local"
+	eventProxyHostname := "event.e2b.dev"
```
dobrac: should it be event.e2b.dev or events.e2b.dev? (description of the RP states different url)

packages/orchestrator/internal/sandbox/network/network_linux.go
```diff
@@ -217,6 +217,12 @@ func (s *Slot) CreateNetwork() error {
 		return fmt.Errorf("error creating postrouting rule: %w", err)
 	}
 
+	// Redirect http://event.e2b.dev traffic destined to event server
+	err = tables.Append("nat", "PREROUTING", "-i", s.VethName(), "-p", "tcp", "-d", "8.8.8.7", "--dport", "80", "-j", "REDIRECT", "--to-port", "5010")
```
dobrac: the port 5010 should be passed from the var

packages/orchestrator/internal/sandbox/event/server.go
```diff
@@ -0,0 +1,65 @@
...
+
+// EventServer handles outbound HTTP requests from sandboxes calling the event.e2b.com endpoint
+type EventServer struct {
+	server *http.Server
+}
+
+func validateHeaders(next http.HandlerFunc) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		sandboxID := r.Header.Get("E2B_SANDBOX_ID")
+		teamID := r.Header.Get("E2B_TEAM_ID")
```
dobrac: this can be spoofed by the sandbox, we shouldn't trust anything the sandbox sends us

packages/orchestrator/internal/sandbox/network/network_linux.go
dobrac: use different IP than 8.8.8.7, I guess any IP from these should be more appropriate:

```
var blockedRanges = []string{
	"10.0.0.0/8",
	"169.254.0.0/16",
	"192.168.0.0/16",
	"172.16.0.0/12",
}
```

0div: ideally yes, i first tried private IPs, they don't leave the sandbox through the network bridge with current setup, i'd have to figure out a way to route one of the private ones this way.

packages/orchestrator/internal/sandbox/event/server.go
0div: indeed, but when these are used for ad-hoc events defined by users that should be fine, for our internal defined ones, they would be sent via envd which would have access to those via mmds stuff ive been working on. the same could be said about envd server. I'm thinking of maybe redirecting the requests to an envd forward proxy but that could be rerouted too with some effort i think.

packages/orchestrator/cmd/build-template/main.go
0div: relic from the past, removing

packages/orchestrator/main.go
0div: why should we be shy about using redis here? i can always add a fallback for local cache.

0div: im working on an eventStore interface atm, it will replace all of this 👍

packages/orchestrator/internal/template/build/rootfs.go
0div: still debating, Tomas was thinking of using something simpler like `event.local`

packages/orchestrator/internal/sandbox/network/network_linux.go
0div: indeed, will make it parametrizable

dobrac: just note, these ranges are blocked by default, you need to enable the target IP address (maybe that might be why you haven't seen them routed)

packages/orchestrator/internal/sandbox/event/server.go
dobrac: the security concern is that the user sandbox would pretend it's another sandbox ID and team ID -> sending the event to completely different target, potentially with malicious desires

0div: it really depends what the handlers do, and if we're to check if the sandbox is running and belongs to the team id that should narrow it down. the same could be said for the potential for mangling the responses from the server running in envd.

sitole: We should probably know the source ip and find a sandbox with it?

packages/orchestrator/internal/sandbox/network/network_linux.go
sitole: Can we maybe not use public IP, but something like metadata IP used by envd for accessing Firecracker VM metadata?

sitole: More details on why Firecracker and cloud providers are using `169.254/16` https://datatracker.ietf.org/doc/html/rfc3927

packages/orchestrator/internal/sandbox/event/server.go
0div: Yep, I'm looking into getting the IP from the namespace and map it to sandboxID

packages/orchestrator/internal/sandbox/network/network_linux.go
0div: > According to RFC 5737 §3, there are specific IP address blocks reserved for documentation and example use. These blocks are: 
> - 192.0.2.0/24 (TEST-NET-1)
> - 198.51.100.0/24 (TEST-NET-2)
> - 203.0.113.0/24 (TEST-NET-3) 
> These addresses are explicitly designated as non-routable and should be filtered by network operators. 

https://en.wikipedia.org/wiki/Reserved_IP_addresses

i found a good candidate for `/etc/hosts`: `203.0.113.0  events.e2b.dev`

dobrac: `203.0.113.0/32` is probably default GW, but otherwise good 👍

sitole: > i found a good candidate for `/etc/hosts`: `203.0.113.0 events.e2b.dev`
Can we consider something more generic than `events.e2b.dev` if we want to use it for different things?  Maybe something like `controller.` or `egress.` or something?

Also, lets try to use some local domain name such as `controller.local` or at least something like `controller.internal.e2b.dev` so its obvious its available only from the sandbox and its not public thing.

packages/orchestrator/internal/sandbox/event/server.go
```diff
@@ -0,0 +1,48 @@
...
+	mux := http.NewServeMux()
+
+	for _, handler := range handlers {
+		mux.HandleFunc(handler.Path(), handler.HandlerFunc)
+	}
+
+	server := &http.Server{
+		Addr:    fmt.Sprintf(":%d", port),
+		Handler: mux,
+	}
```
sitole, line 24: What about running Gin here so we can use a strict Golang client in envd and track all version changes?

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -421,6 +423,12 @@ func ResumeSandbox(
 		return nil, cleanup, fmt.Errorf("failed to wait for sandbox start: %w", err)
 	}
 
+	sandboxIP := ips.slot.HostIPString()
+	eventStore.SetSandboxIP(config.SandboxId, sandboxIP)
+	cleanup.AddPriority(func(ctx context.Context) error {
+		return eventStore.DelSandboxIP(sandboxIP)
+	})
+
```
sitole, line 431: Why can we not directly get the sandbox from the orchestrator cache and get the slot IP on it?

packages/orchestrator/internal/template/build/rootfs.go
```diff
@@ -197,6 +198,9 @@ ExecStart=-/sbin/agetty --noissue --autologin root %I 115200,38400,9600 vt102
 `
 
 	hostname := "e2b.local"
+	eventProxyHostname := "events.e2b.dev"
+
+	eventIP := env.GetEnv("SANDBOX_EVENT_IP", "203.0.113.0")
```
sitole: Lets make IP constant in some shared package

packages/orchestrator/internal/sandbox/network/network_linux.go
0div: @dobrac default GW is: 
```
user@e2b:~$ ip route | grep default
default via 169.254.0.22 dev eth0
```

0div: @sitole 
fine, ill use a `.local` TLD for now. i was projecting a future in which we would use the domain as a two-way bus, thus the need for a public internet domain & TLD

packages/orchestrator/internal/sandbox/event/server.go
0div: sure, we can use gin, but i'm not sure it's more strict than the std lib and also what do u mean by tracking version changes in this case?

packages/orchestrator/internal/sandbox/sandbox.go
0div: Correct me if i'm wrong but that map uses sandbox ID as key? in my use case I need sandbox IP as key to get the ID.

packages/orchestrator/internal/template/build/rootfs.go
0div: Yeah, i was gonna introduce a `consts.go` like in other packages.

packages/orchestrator/internal/sandbox/network/network_linux.go
sitole: Perhaps provide the domain as an environment variable via envd so we are not locked into it for the future? Then does not matter what domain we will use now

dobrac: It does matter in a sense where it will be encoded in all envds

packages/orchestrator/internal/sandbox/event/server.go
sitole: > but i'm not sure it's more strict than the std lib and also what do u mean by tracking version changes in this case?

I was thinking that the "metrics server" can be Gin-backed by the OpenAPI scheme, as we are using for the API, so all breaking changes will be clear (because of the re-generated schema). You will also be able to use the generated Go client from EnvD to call metrics service endpoints.

sitole: The point was just make it strong types on both caller and receiver

packages/orchestrator/internal/sandbox/sandbox.go
sitole: Yes, you are right. But still, using Redis may not be needed here? You don't need to share state to a 3rd party service, because you will always receive requests just from sandboxes that are managed by the orchestrator itself, right? This way on Orchestrator, you should always know about all sandboxes without fetching them from Redis.

Probably there will be <300 sandboxes, which means the sandboxes map is not that big to iterate it every time you need to find an IP in it. The number of sandboxes will probably not grow, because of networking, memory, and CPU limitations. If you feel that it can be issue, then we can do some pattern that will create in-memory map "sbx ip -> sbx id" and make it synced with sandboxes map.

packages/orchestrator/internal/sandbox/network/network_linux.go
sitole: > It does matter in a sense where it will be encoded in all envds

That was my point with `Perhaps provide the domain as an environment variable via envd so we are not locked into it for the future`

dobrac: Oh, you mean through MMDS, not environment variable, correct?

sitole: Yes, sorry for confusion

packages/orchestrator/internal/template/build/rootfs.go
```diff
@@ -197,6 +198,9 @@ ExecStart=-/sbin/agetty --noissue --autologin root %I 115200,38400,9600 vt102
 `
 
 	hostname := "e2b.local"
+	eventProxyHostname := "events.e2b.dev"
+
+	eventIP := internal.GetSandboxEventIP()
```
jakubno, line 203: do we want to bake in into the template? Can't we use a mmds server here?

packages/orchestrator/internal/consts.go
```diff
@@ -0,0 +1,13 @@
+package internal
+
+import (
+	"github.com/e2b-dev/infra/packages/shared/pkg/env"
+)
+
+const (
+	defaultSandboxEventIP = "203.0.113.0"
```
jakubno, line 8: why do you use TEST-NET-3 here?

packages/orchestrator/internal/sandbox/event/server.go
jakubno: Agree with @sitole

jakubno (commented):
Will we use redis or clickhouse here?

packages/orchestrator/internal/sandbox/sandbox.go
jakubno: Agree with @sitole here

sitole:
@0div is there some update?

I would like to use the sandbox -> orchestrator HTTP server you introduced here.
I'm thinking we could split it into separate PRs to make the review quicker?

0div:
> @0div is there some update?
> 
> I would like to use the sandbox -> orchestrator HTTP server you introduced here. I'm thinking we could split it into separate PRs to make the review quicker?

@sitole , i put this back as draft bc i would start using it once i can consume it with the webhooks system i'm implenting. 
What do you need this for exactly and how urgent do you need it?

PR #789, opened 2025-06-17
Make sandbox metadata available as global env vars in envd

0div:


packages/envd/main.go
```diff
@@ -143,17 +146,27 @@ func main() {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	l := logs.NewLogger(ctx, debug)
+	envVars := utils.NewMap[string, string]()
+	isSandboxBoolStr := strconv.FormatBool(!debug)
+	envVars.Store("E2B_SANDBOX", isSandboxBoolStr)
+	if err := os.WriteFile("/etc/.E2B_SANDBOX", []byte(isSandboxBoolStr), 0444); err != nil {
```
dobrac: wouldn't a better path be `/run/.E2B_SANDBOX` or `/run/e2b/.E2B_SANDBOX`?

packages/envd/internal/logs/exporter/exporter.go
```diff
@@ -76,30 +87,8 @@ func (w *HTTPExporter) start() {
 			continue
 		}
 
-		token, err := w.getMMDSToken(w.ctx)
```
dobrac, line 79: This works only because the MMDS is not set now during the build. If that would change any time in the future, this setup will be incorrect.

This is also a change in the behavior, as the triggers called this when the sandbox was resumed, which is not the case anymore

packages/envd/internal/logs/exporter/exporter.go
```diff
@@ -58,8 +60,17 @@ func printLog(logs []byte) {
...
-	w.waitForMMDS(w.ctx)
+func (w *HTTPExporter) start(mmdsChan <-chan *host.MMDSOpts) {
+	mmdsOpts, ok := <-mmdsChan
+	if !ok {
+		fmt.Fprintf(os.Stderr, "channel closed when starting exporter")
+		return
+	}
+
+	if mmdsOpts == nil {
+		fmt.Fprintf(os.Stderr, "nil mmds opts received on channel when starting exporter")
```
dobrac: this will silently fail in our case, we should have a retry logic here

packages/envd/internal/host/mmds.go
```diff
@@ -0,0 +1,149 @@
...
+				continue
+			}
+
+			mmdsOpts, err := getMMDSOpts(ctx, httpClient, token)
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "error getting mmds opts: %v\n", err)
+				continue
+			}
+
+			if mmdsOpts.Address != "" {
```
dobrac, line 155: nit: flip the if so the body is smaller

packages/envd/internal/host/mmds.go
```diff
@@ -0,0 +1,149 @@
...
+				envVars.Store("E2B_SANDBOX_ID", mmdsOpts.SandboxID)
+				envVars.Store("E2B_ENV_ID", mmdsOpts.EnvID)
+				envVars.Store("E2B_TEAM_ID", mmdsOpts.TeamID)
+				if err := os.WriteFile("/etc/.E2B_SANDBOX_ID", []byte(mmdsOpts.SandboxID), 0444); err != nil {
+					fmt.Fprintf(os.Stderr, "error writing sandbox ID file: %v\n", err)
```
dobrac: I don't think these fmt.Fprintf are logged anywhere

packages/envd/internal/host/mmds.go
```diff
@@ -0,0 +1,149 @@
...
+
+func (opts *MMDSOpts) AddOptsToJSON(jsonLogs []byte) ([]byte, error) {
+	parsed := make(map[string]interface{})
+
+	err := json.Unmarshal(jsonLogs, &parsed)
+	if err != nil {
+		return nil, err
+	}
+
+	parsed["sandboxID"] = opts.SandboxID
```
dobrac: you need to keep the `instanceID` for Vector ingest compatibility

dobrac (changes requested):

packages/envd/main.go
0div: similarly to `systemd`, i'll store them in a `/run/e2b/` dir

packages/envd/internal/logs/exporter/exporter.go
0div: tested before and after pause/resume, works fine with the same values being asserted for the first snapshot

packages/envd/internal/host/mmds.go
0div: They are not, but useful for debugging local dev , could add `if debug {`

packages/envd/internal/host/mmds.go
0div: you are unfortunately right

packages/envd/internal/host/mmds.go
```diff
@@ -131,13 +133,13 @@ func PollForMMDSOpts(ctx context.Context, mmdsChan chan<- *MMDSOpts, envVars *ut
 				envVars.Store("E2B_SANDBOX_ID", mmdsOpts.SandboxID)
 				envVars.Store("E2B_ENV_ID", mmdsOpts.EnvID)
 				envVars.Store("E2B_TEAM_ID", mmdsOpts.TeamID)
-				if err := os.WriteFile("/etc/.E2B_SANDBOX_ID", []byte(mmdsOpts.SandboxID), 0444); err != nil {
+				if err := os.WriteFile(E2BRunDir+"/.E2B_SANDBOX_ID", []byte(mmdsOpts.SandboxID), 0444); err != nil {
 					fmt.Fprintf(os.Stderr, "error writing sandbox ID file: %v\n", err)
 				}
-				if err := os.WriteFile("/etc/.E2B_ENV_ID", []byte(mmdsOpts.EnvID), 0444); err != nil {
+				if err := os.WriteFile(E2BRunDir+"/.E2B_ENV_ID", []byte(mmdsOpts.EnvID), 0444); err != nil {
```
dobrac: can this be TEMPLATE_ID?

packages/envd/internal/logs/exporter/exporter.go
```diff
@@ -88,16 +88,19 @@ func (w *HTTPExporter) start(mmdsChan <-chan *host.MMDSOpts) {
 		}
 
 		for _, logLine := range logs {
-			logsWithOpts, jsonErr := mmdsOpts.AddOptsToJSON(logLine)
-			if jsonErr != nil {
-				log.Printf("error adding instance logging options (%+v) to JSON (%+v) with logs : %v\n", mmdsOpts, logLine, jsonErr)
+			if mmdsOpts != nil {
```
dobrac: why to send logs when we don't have correct metadata? They'll get lost anyway

packages/envd/internal/logs/exporter/exporter.go
dobrac: this one is not resolved

packages/envd/internal/logs/exporter/exporter.go
dobrac: this one is not resolved either, we should still handle if the mmds change in the sandbox runtime, for example to not break it for cloning/snapshotting

packages/envd/internal/logs/exporter/exporter.go
0div: forgot to remove that line

packages/envd/internal/logs/exporter/exporter.go
0div: gonna prevent this from making an http request to get token and get mmds for each call to the logger `Write()`, will call it on `init` and make this range get the updates.

packages/envd/Makefile
```diff
@@ -18,7 +18,7 @@ build-debug:
 	CGO_ENABLED=1 go build -race -gcflags=all="-N -l" -o bin/debug/envd ${LDFLAGS}
 
 start-docker:
-	DOCKER_BUILDKIT=1 docker build -t envd-debug . -f debug.Dockerfile
+	DOCKER_BUILDKIT=1 docker build -t envd-debug. -f debug.Dockerfile
```
jakubno: Can you do this?
`envd-debug.` shouldn't there be a space?

packages/envd/internal/logs/exporter/exporter.go
```diff
@@ -58,57 +63,61 @@ func printLog(logs []byte) {
...
+			if !ok {
+				return nil
+			}
+			w.Lock()
+			if mmdsOpts == nil {
+				mmdsOpts = &host.MMDSOpts{
+					InstanceID: "unknown",
+					EnvID:      "unknown",
+					TeamID:     "unknown",
+					Address:    "http://localhost:30006", // default logs collector address
```
jakubno: this won't work as the network inside sandbox won't allow to send it to localhost on host machine. I think adding something like empty string and not sending any request would make more sense?

packages/envd/Makefile
0div: a find and replace gone wrong, thx for the catch, fixed.

packages/envd/internal/logs/exporter/exporter.go
0div: good point, ill add that logic

packages/envd/Makefile
```diff
@@ -27,7 +27,7 @@ start-docker:
 	-p 8000:8000 \
 	-p 8001:8001 \
 	--rm \
-	-i envd-debug \
```
dobrac: revert

packages/envd/README.md
```diff
@@ -10,7 +10,7 @@ Run the following command to (re)build the envd daemon and start a Docker contai
 make build && make start-docker
 ```
 
-You can use E2B SDKs with env var `E2B_DEBUG=true` or with a debug parameter set to true when creating or connecting to a sandbox to connect to the envd started with this command locally.
```
dobrac: revert

packages/envd/internal/host/mmds.go
```diff
@@ -100,51 +106,47 @@ func (w *HTTPExporter) doMmdsRequest(ctx context.Context, token string) (*opts,
...
+	defer httpClient.CloseIdleConnections()
+
 	ticker := time.NewTicker(50 * time.Millisecond)
 	defer ticker.Stop()
 
 	for {
 		select {
 		case <-ctx.Done():
-			return
+			fmt.Fprintf(os.Stderr, "context cancelled while waiting for mmds opts")
```
dobrac, line 127:
```suggestion
			fmt.Fprintf(os.Stderr, "context cancelled while waiting for mmds opts")
			return
```

packages/envd/internal/logs/exporter/exporter.go
```diff
@@ -58,57 +67,61 @@ func printLog(logs []byte) {
...
-			}
-
-			continue
-		}
-
 		for _, logLine := range logs {
-			logsWithOpts, jsonErr := mmdsOpts.addOptsToJSON(logLine)
-			if jsonErr != nil {
-				log.Printf("error adding instance logging options (%+v) to JSON (%+v) with logs : %v\n", mmdsOpts, logLine, jsonErr)
+			logLineWithOpts, err := w.mmdsOpts.AddOptsToJSON(logLine)
```
dobrac, line 119: I think you need to lock here also when you read from the `mmdsOpts` variable (applies to other places too)

packages/envd/internal/host/mmds.go
```diff
@@ -100,51 +106,47 @@ func (w *HTTPExporter) doMmdsRequest(ctx context.Context, token string) (*opts,
...
-			mmdsOpts, err := w.doMmdsRequest(ctx, token)
+			mmdsOpts, err := getMMDSOpts(ctx, httpClient, token)
 			if err != nil {
-				fmt.Printf("error getting mmds opts: %v\n", err)
+				fmt.Fprintf(os.Stderr, "error getting mmds opts: %v\n", err)
 				continue
 			}
 
 			if mmdsOpts.Address != "" {
+				envVars.Store("E2B_SANDBOX_ID", mmdsOpts.InstanceID)
```
dobrac: you might want to store the data even if Address is == ""

packages/envd/internal/host/mmds.go
```diff
@@ -100,51 +106,47 @@ func (w *HTTPExporter) doMmdsRequest(ctx context.Context, token string) (*opts,
...
+				}
+				if err := os.WriteFile(filepath.Join(E2BRunDir, ".E2B_TEMPLATE_ID"), []byte(mmdsOpts.EnvID), 0o666); err != nil {
+					fmt.Fprintf(os.Stderr, "error writing template ID file: %v\n", err)
+				}
+				if err := os.WriteFile(filepath.Join(E2BRunDir, ".E2B_TEAM_ID"), []byte(mmdsOpts.TeamID), 0o666); err != nil {
+					fmt.Fprintf(os.Stderr, "error writing team ID file: %v\n", err)
+				}
+				mmdsChan <- mmdsOpts
+
 				return
```
dobrac: are we reading it only once? So the usecase when the sandboxID changes in the future won't work

0div: forgot to call it in init

packages/envd/internal/host/mmds.go
```diff
@@ -0,0 +1,178 @@
...
+func (opts *MMDSOpts) Update(traceID, instanceID, envID, address, teamID string) {
+	opts.Lock()
+	defer opts.Unlock()
+
+	opts.TraceID = traceID
+	opts.InstanceID = instanceID
+	opts.EnvID = envID
+	opts.Address = address
+	opts.TeamID = teamID
+}
```
dobrac: why do these methods (and the lock on MMDSOpts) exits? Wouldn't it be better to work with MMDSOpts just as data class?

packages/envd/internal/logs/exporter/exporter.go
```diff
@@ -58,57 +70,63 @@ func printLog(logs []byte) {
...
+func (w *HTTPExporter) start() {
 	for range w.triggers {
 		logs := w.getAllLogs()
 
 		if len(logs) == 0 {
 			continue
 		}
 
-		if w.debug {
+		if w.isNotFC {
```
dobrac, line 109: looking at the code again, I think debug name was more explicit in what it actually does

packages/envd/internal/logs/exporter/exporter.go
```diff
@@ -58,57 +70,63 @@ func printLog(logs []byte) {
 	fmt.Fprintf(os.Stdout, "%v", string(logs))
 }
 
-func (w *HTTPExporter) start() {
-	w.waitForMMDS(w.ctx)
+func (w *HTTPExporter) listenForMMDSOptsAndStart(mmdsChan <-chan *host.MMDSOpts) *host.MMDSOpts {
```
dobrac: why does it return *host.MMDSOpts?

packages/envd/internal/logs/exporter/exporter.go
```diff
@@ -58,57 +70,63 @@ func printLog(logs []byte) {
...
+	for {
+		select {
+		case <-w.ctx.Done():
+			return nil
+		case mmdsOpts, ok := <-mmdsChan:
+			if !ok {
+				return nil
+			}
+			w.mmdsLock.Lock()
```
dobrac, line 89: the lock can be only around mmdsOpts assignment. Btw, have you considered doing getter and setter for the mmdsOpts on the HTTPExporter instance, so the lock can be colocated to the methods?

packages/envd/internal/host/mmds.go
0div: what do you mean by data class?

packages/envd/internal/logs/exporter/exporter.go
0div: pls explain why. 
I'm seeing it in code branches to determine if we use local logging or not and if we poll for FC specific MMDS data

packages/envd/internal/host/mmds.go
dobrac: actually, I think these methods and the lock are not even used

packages/envd/internal/logs/exporter/exporter.go
dobrac: I mean extensive log messages mean debug, not FC or not. But I see that there might be stuff decided on the debug field that are more connected with the environment. I think it would make sense in that case to name it relevantly for each usage, so it's appropriate to what it does in that scope. Then just decide if it's used based on the debug value (or some new introduced value and keep debug just for the logging)

packages/envd/internal/host/mmds.go
0div: using them, added commit this time.

packages/envd/internal/logs/exporter/exporter.go
0div: it's implemented in host/mmds.go

packages/envd/internal/logs/exporter/exporter.go
0div: it's extensive logging using data only available in an FC VM, and i believe it's not used for debugging but info level logging.

dobrac: I agree it's better when you register/run specific FC tasks (like the MMDS). On the other hand, look at the linked usage. Based on `if w.isNotFC {` it's decided to logs stuff or not. I don't think it makes semantic sense in this case, for that I would change it or used different var

Also, this is the flag definition rn, could you please fix the naming, so it's not mix of both? Also, It's used not only in the HTTPExporter, but also in envd api, where I'm not sure if it has the same meaning as in the logger

```
	flag.BoolVar(
		&isNotFC,
		"debug",
		false,
		"isNotFCmode prints all logs to stdout",
	)
```

packages/envd/internal/logs/exporter/exporter.go
```diff
@@ -58,57 +76,54 @@ func printLog(logs []byte) {
...
-			}
-
-			continue
-		}
-
 		for _, logLine := range logs {
-			logsWithOpts, jsonErr := mmdsOpts.addOptsToJSON(logLine)
-			if jsonErr != nil {
-				log.Printf("error adding instance logging options (%+v) to JSON (%+v) with logs : %v\n", mmdsOpts, logLine, jsonErr)
+			w.mmdsOpts.RLock()
```
dobrac: because you've moved the lock to the mmdsOpts, this lock is now useless. I would propose to revert moving updating the MMDs to the Update method, removing the mutex from MMDSOpts, and adding helper functions to the HTTPExporter using mmdsLock

packages/shared/pkg/storage/storage_google.go
```diff
@@ -36,7 +37,7 @@ type GCPBucketStorageObjectProvider struct {
 }
 
 func NewGCPBucketStorageProvider(ctx context.Context, bucketName string) (*GCPBucketStorageProvider, error) {
-	client, err := storage.NewClient(ctx)
+	client, err := storage.NewClient(ctx, option.WithUniverseDomain("private.googleapis.com"))
```
dobrac: not relevant to this PR

packages/api/internal/handlers/sandbox_events.go
```diff
@@ -0,0 +1,37 @@
...
+
+	"github.com/gin-gonic/gin"
+
+	"github.com/e2b-dev/infra/packages/api/internal/auth"
+	authcache "github.com/e2b-dev/infra/packages/api/internal/cache/auth"
+	"github.com/e2b-dev/infra/packages/api/internal/utils"
+	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
+)
+
+func (a *APIStore) GetSandboxesSandboxIDEvents(
```
dobrac: not relevant to this PR

packages/envd/internal/logs/exporter/exporter.go
0div: updated flag. 

> it's decided to logs stuff or not.

it logs regardless, decides where to write to, stdout vs http

packages/envd/internal/logs/exporter/exporter.go
0div: the argument to move the synchronization outside of the data struct methods makes sense, updating.

packages/envd/internal/logs/exporter/exporter.go
dobrac: oh, I see, my bad

packages/envd/internal/logs/exporter/exporter.go
```diff
@@ -85,6 +85,9 @@ func (w *HTTPExporter) listenForMMDSOptsAndStart(mmdsChan <-chan *host.MMDSOpts)
 			if !ok {
 				return
 			}
+
+			w.mmdsLock.Lock()
+			defer w.mmdsLock.Unlock()
```
dobrac: are you sure the defer is called after exit from the `case` and not from the whole listen method?

dobrac (changes requested):
one more thing and gtg! 🥳

packages/envd/internal/logs/exporter/exporter.go
0div: yeah good point, i changed it for the loop but i should do it here too

dobrac (approved):

PR #791, opened 2025-06-17
ClickHouse data management

jakubno:
# Description

- Increases disk size
- Cleans up clickhouse config
- Adds a job for periodic backup
- Adds a parametrized job for restoring backup


0div (approved):
Nice you're using `altinity/clickhouse-backup`!  
I strongly recommend calling a slack web-hook to notify when backup succeeded or failed, relied on it a lot in the past. 
What will our retention policy be?

PR #792, opened 2025-06-17
Update self-host.md

tomassrnka:
Small hints on what needs to be set for succesful deployment in GCP and Supabase

- Adjust Quota setting for disk space in GCP, default is 400G, which is not enough.
- By default the comms go via IPv4 from API to Supabase, hence IPv4 must be enabled, otherwise API deployment fails.


jakubno:
Thanks @tomassrnka! Really appreciate updating the guide!

PR #795, opened 2025-06-17
BYOC: Edge cluster pool implementation

sitole:
First part of byoc cluster implementation in api. Adds needed database migrations and introducing edge pool in api.

- Database migrations for storing clusters, attached to team / tempalte
- Edge pool that periodically sync used clusters


packages/db/migrations/20250606213446_deployment_cluster.sql
```diff
@@ -0,0 +1,25 @@
+-- +goose Up
+-- +goose StatementBegin
+CREATE TABLE IF NOT EXISTS clusters (
+    id        UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+    endpoint  TEXT NOT NULL,
+    token     TEXT NOT NULL
```
ValentaTomas: Shouldn't this be just a hash?

sitole: not hash, but should be encrypted.

packages/shared/pkg/schema/cluster.go
```diff
@@ -0,0 +1,49 @@
...
+func (Cluster) Fields() []ent.Field {
+	return []ent.Field{
+		field.UUID("id", uuid.UUID{}).
+			Immutable().
+			Unique().
+			Annotations(entsql.Default("gen_random_uuid()")),
+		field.String("endpoint").
+			NotEmpty().
+			SchemaType(map[string]string{dialect.Postgres: "text"}),
+		field.String("token").
```
jakubno, line 29: is this sensitive?

packages/api/internal/edge/edge.go
```diff
@@ -0,0 +1,252 @@
...
+			}
+		}
+
+		if found {
+			continue
+		}
+
+		// cluster disconnect takes time
+		wg.Add(1)
+		go func(custer *Cluster) {
```
jakubno: it's somehow mixed here
```suggestion
		go func(cluster *Cluster) {
```

packages/api/internal/edge/edge.go
```diff
@@ -0,0 +1,252 @@
...
+		return nil, err
+	}
+
+	// periodically sync clusters with the database
+	go func() { p.syncBackground() }()
+
+	return p, nil
+}
+
+func (p *Pool) GetClusterById(clusterId uuid.UUID) (*Cluster, bool) {
```
jakubno: not used

packages/api/internal/edge/edge.go
```diff
@@ -0,0 +1,252 @@
...
+
+type Cluster struct {
+	ctx context.Context
+
+	Id     uuid.UUID
+	Client *api.ClientWithResponses
+}
+
+func NewCluster(ctx context.Context, endpoint string, secret string, id uuid.UUID) (*Cluster, error) {
+	// so we during cluster disconnect we don't cancel the upper context
```
jakubno, line 167: Context works the other way around, if you cancel child context, nothing happens to the parent context

packages/api/internal/edge/edge.go
```diff
@@ -0,0 +1,252 @@
...
+	return &Cluster{
+		ctx: ctx,
+
+		Id:     id,
+		Client: client,
+	}, nil
+}
+
+func (c *Cluster) Disconnect() {
+	c.ctx.Done()
```
jakubno: This return `chan` - it will just return without waiting

packages/api/internal/edge/edge.go
```diff
@@ -0,0 +1,252 @@
...
+}
+
+func (c *Cluster) getTemplateBuilders(ctx context.Context) ([]*api.ClusterOrchestratorNode, error) {
```
jakubno: passing the context here and using different inside is little confusing

packages/api/internal/edge/edge.go
```diff
@@ -0,0 +1,252 @@
...
+	if err != nil {
+		zap.L().Error("failed to get orchestrators", zap.Error(err), l.WithClusterID(c.Id))
+		return nil, err
+	}
+
+	if res.JSON200 == nil {
+		return nil, errors.New("api request failed")
+	}
+
+	orchestrators := make([]*api.ClusterOrchestratorNode, 0)
```
jakubno: rename please

packages/api/internal/edge/edge.go
```diff
@@ -0,0 +1,252 @@
...
+		}
+
+		orchestrators = append(orchestrators, &o)
+	}
+
+	return orchestrators, nil
+}
+
+func (c *Cluster) GetTemplateBuilderById(nodeId string) (*api.ClusterOrchestratorNode, error) {
+	orchestrators, err := c.getTemplateBuilders(c.ctx)
```
jakubno: rename

packages/api/internal/edge/edge.go
```diff
@@ -0,0 +1,252 @@
...
+		if o.Id == nodeId {
+			return o, nil
+		}
+	}
+
+	return nil, ErrTemplateBuilderNotFound
+}
+
+func (c *Cluster) GetAvailableTemplateBuilder() (*api.ClusterOrchestratorNode, error) {
+	orchestrators, err := c.getTemplateBuilders(c.ctx)
```
jakubno: rename

packages/db/queries/get_active_clusters.sql
```diff
@@ -0,0 +1,4 @@
+-- name: GetActiveClusters :many
+SELECT DISTINCT sqlc.embed(c)
```
jakubno, line 2: why is there distinct if cluster <-> team should be unique?

packages/db/migrations/20250611172307_cluster_node_for_builds.sql
jakubno, line 1: why is this here? I don't see it used anywhere

packages/db/migrations/20250611042354_cluster_for_templates.sql
jakubno, line 1: why is this here? I don't see it used anywhere

packages/api/internal/edge/edge.go
sitole: That's the point. I want to cancel the cluster, not the pool (the upper one).

packages/db/queries/get_active_clusters.sql
sitole: Added migrations to remove the unique index in the team table. So now we can have multiple teams with the same cluster ID, there are no blockers for this.

packages/shared/pkg/schema/cluster.go
sitole: it is, we are in disucssion what exact way of securing this we will use.

packages/db/migrations/20250606213446_deployment_cluster.sql
```diff
@@ -0,0 +1,26 @@
+-- +goose Up
+-- +goose StatementBegin
+CREATE TABLE IF NOT EXISTS clusters (
+    id            UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+    endpoint      TEXT NOT NULL,
+    endpoint_tls  BOOLEAN NOT NULL DEFAULT TRUE,
```
sitole, line 6: Endpoint tls added so we can from endpoint + endpoint tls build both HTTP client URL and host for GRPC connection (used for orchestrator/builder proxy).

packages/shared/pkg/schema/cluster.go
jakubno: can you mark it sensitive?

packages/api/internal/edge/edge.go
jakubno: you want to use WithCancel

packages/api/internal/edge/edge.go
```diff
@@ -0,0 +1,257 @@
...
+	return &Cluster{
+		ctx: ctx,
+
+		Id:     id,
+		Client: endpointClient,
+	}, nil
+}
+
+func (c *Cluster) Disconnect() {
+	<-c.ctx.Done()
```
jakubno: When does the cancel get cancelled?

packages/api/internal/edge/edge.go
```diff
@@ -0,0 +1,257 @@
...
+}
+
+func (c *Cluster) getTemplateBuilders() ([]*api.ClusterOrchestratorNode, error) {
+	res, err := c.Client.V1ServiceDiscoveryGetOrchestratorsWithResponse(c.ctx)
+	if err != nil {
+		zap.L().Error("failed to get builders", zap.Error(err), l.WithClusterID(c.Id))
+		return nil, err
+	}
+
+	if res.JSON200 == nil {
```
jakubno, line 219: can you first change status code ?

packages/api/internal/edge/edge.go
```diff
@@ -0,0 +1,257 @@
...
+
+	return &Cluster{
+		ctx: ctx,
+
+		Id:     id,
+		Client: endpointClient,
+	}, nil
+}
+
+func (c *Cluster) Disconnect() {
```
jakubno, line 204: I think what you want to do is to track ongoing requests, yo would probably have to manually check for each request if it should still serve, if yes add the request to wait group and here wait fro that group to complete

packages/api/internal/edge/edge.go
sitole: fixed, was midded that and created context without cancel instead of WITH cancel.

packages/shared/pkg/schema/cluster.go
sitole: done

packages/api/internal/edge/edge.go
sitole: This should primarily in future hold disconnect and removal until GRPC requests are done.
Currently, I am canceling the context that is used for HTTP client requests, so they will immediately cancel.

This case will happen only when the cluster is removed from the database.

packages/api/internal/edge/edge.go
```diff
@@ -187,18 +188,21 @@ func NewCluster(ctx context.Context, endpoint string, endpointTls bool, secret s
...
+		ctxCancel: ctxCancel,
 
 		Id:     id,
 		Client: endpointClient,
 	}, nil
 }
 
 func (c *Cluster) Disconnect() {
+	c.ctxCancel()
 	<-c.ctx.Done()
```
jakubno: I think you don't have to wait for ctx.Done as it won't do anything (you cancel right before)

jakubno (changes requested):

PR #798, opened 2025-06-18
Add migrator

jakubno:
# Description

- Adds a Clickhouse migrator
- It starts only once - when deployed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -165,6 +165,7 @@ build-and-upload:build-and-upload/docker-reverse-proxy
 build-and-upload:build-and-upload/orchestrator
 build-and-upload:build-and-upload/template-manager
 build-and-upload:build-and-upload/envd
+build-and-upload:build-and-upload/clickhouse
 build-and-upload/template-manager:
 	./scripts/confirm.sh $(TERRAFORM_ENVIRONMENT)
 	GCP_PROJECT_ID=$(GCP_PROJECT_ID) $(MAKE) -C packages/orchestrator build-and-upload/template-manager
--- a/packages/nomad/clickhouse-migrator.hcl
+++ b/packages/nomad/clickhouse-migrator.hcl
@@ -1,0 +1,51 @@
+job "clickhouse-migrator" {
+  type        = "batch"
+  node_pool   = "${node_pool}"
+
+%{ for i in range("${server_count}") }
+  group "migrator-${i + 1}" {
+    count = 1
+
+
+    restart {
+      interval         = "5m"
+      attempts         = 5
+      delay            = "15s"
+      mode             = "delay"
+    }
+
+    network {
+      mode = "host"
+
+      dns {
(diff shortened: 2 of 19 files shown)


packages/nomad/clickhouse.hcl
```diff
@@ -206,6 +206,8 @@ EOF
         <${username}>
             <password>${password}</password>
             <networks>
+              <ip>172.26.64.1/16</ip> <!-- allow Nomad access -->
```
jakubno: Allow access from other docker containers

packages/nomad/main.tf
```diff
@@ -34,6 +43,17 @@ data "google_secret_manager_secret_version" "redis_url" {
   secret = var.redis_url_secret_version.secret
 }
 
+
+data "docker_registry_image" "api_image" {
```
jakubno, line 47: moved the docker images here as passing them over all repo is annoying

Makefile
```diff
@@ -165,12 +165,16 @@ build-and-upload:build-and-upload/docker-reverse-proxy
 build-and-upload:build-and-upload/orchestrator
 build-and-upload:build-and-upload/template-manager
 build-and-upload:build-and-upload/envd
+build-and-upload:build-and-upload/clickhouse-migrations
```
dobrac: should it be `clickhouse-migrations` or `clickhouse-migrator`?

packages/nomad/clickhouse-migrator.hcl
```diff
@@ -0,0 +1,44 @@
...
+      attribute = "$${meta.job_constraint}"
+      value     = "${job_constraint_prefix}-${i + 1}"
+    }
+
+    task "migrator" {
+      driver = "docker"
+
+      env {
+        GOOSE_DBSTRING="clickhouse://${clickhouse_username}:${clickhouse_password}@localhost:${clickhouse_port}/default"
+        GOOSE_DRIVER="clickhouse"
```
dobrac: I would set this in the Docker image directly, as it's required for the tool to run correctly

packages/nomad/clickhouse-migrator.hcl
```diff
@@ -0,0 +1,44 @@
...
+      }
+
+      config {
+        image = "${clickhouse_migrator_version}"
+        network_mode = "host"
+        entrypoint = ["/bin/sh", "-c"]
+        args = ["goose -table _migrations -dir migrations up"]
```
dobrac: why not set these in the docker image directly too?

packages/nomad/clickhouse.hcl
dobrac: Could you add a comment what is the meaning of that IP address CIDR or where you got it from?

dobrac (changes requested):
LGTM, small nits

dobrac (approved):

PR #801, opened 2025-06-18
BYOC: Edge nodes proxy implementation

sitole:


sitole:
TODO:
- Cleanup client proxy service startup
- Manage proper service shutdown (wait or properly close running gRPC connections in edge proxy).
- Rename edge proxy to nodes pass-through.
- Metrics for pass-through usage.

packages/shared/pkg/http/edge/spec.gen.go
jakubno, line 1: isn't this deleted?

packages/shared/pkg/grpc/connection.go
```diff
@@ -25,13 +25,7 @@ type ClientConnInterface interface {
 }
 
 // TODO: Fix Host <-> Url
-func GetConnection(host string, safe bool, options ...grpc.DialOption) (ClientConnInterface, error) {
-	if strings.TrimSpace(host) == "" {
```
jakubno: won't this cause some issues if you provide empty string for analytics collector in API?

packages/client-proxy/internal/edge-pass-through/proxy.go
```diff
@@ -0,0 +1,217 @@
...
+
+// Following code implement a gRPC pass-through proxy that forwards requests to the appropriate node
+// Code is based on following source: https://github.com/siderolabs/grpc-proxy/tree/main
+//
+// Core implementation is just following methods that are handling forwarding, proper closing and propagating of errors from both sides of the stream.
+// The handler is called for every request that is not handled by any other gRPC service.
+func (s *NodePassThrough) handler(srv interface{}, serverStream grpc.ServerStream) error {
+	fullMethodName, ok := grpc.MethodFromServerStream(serverStream)
+	if !ok {
+		return status.Errorf(codes.Internal, "lowLevelServerStream not exists in context")
```
jakubno: what's `lowLevelServerStream`

packages/client-proxy/internal/edge-pass-through/proxy.go
```diff
@@ -0,0 +1,217 @@
...
+				return c2sErr
+			}
+			return nil
+		}
+	}
+
+	return status.Errorf(codes.Internal, "gRPC proxying should never reach this stage.")
+}
+
+func (s *NodePassThrough) forwardClientToServer(src grpc.ClientStream, dst grpc.ServerStream) chan error {
```
jakubno: In case of one2one there's different implementation in the source code
https://github.com/siderolabs/grpc-proxy/blob/8614c7158032488d36285ce2245d06f49d7447c4/proxy/handler_one2one.go#L59

jakubno: maybe we don't even need client streaming right now, i would jsut reference the code for potential later use

packages/client-proxy/internal/edge/handlers/node-utils.go
jakubno, line 1: this isn't used at all

packages/client-proxy/internal/edge/pool/edge-pool.go
```diff
@@ -192,19 +192,19 @@ func (p *EdgePool) removeNode(ctx context.Context, node *EdgeNode) error {
 	_, childSpan := p.tracer.Start(ctx, "remove-edge-node")
 	defer childSpan.End()
 
-	p.logger.Info("Edge node connection is not active anymore, closing.", zap.String("node_id", node.ServiceId))
+	p.logger.Info("Edge node Connection is not active anymore, closing.", zap.String("node_id", node.ServiceId))
```
jakubno: uppercased this by mistake during some refactor?

packages/client-proxy/internal/edge/pool/orchestrator-pool.go
jakubno, line 1: also here

packages/client-proxy/main.go
```diff
@@ -155,8 +173,36 @@ func run() int {
...
+	}
+
+	muxServer := cmux.New(lis)
+	muxRpcListener := muxServer.MatchWithWriters(cmux.HTTP2MatchHeaderFieldSendSettings("content-type", "application/grpc")) // handler requests for gRPC pass through
+	muxRestListener := muxServer.Match(cmux.Any())                                                                           // handler requests for REST API
+
+	go func() {
+		grpcSrv := edgePassThroughHandler.GetServer()
+		if err := grpcSrv.Serve(muxRpcListener); err != nil && !errors.Is(err, http.ErrServerClosed) {
+			log.Fatalf("gRPC Serve: %v", err)
```
jakubno: don't fatal here, handle it gracefully (look at trafficProxy)

packages/client-proxy/main.go
```diff
@@ -155,8 +173,36 @@ func run() int {
...
+		}
+	}()
+
+	go func() {
+		httpSrv := &http.Server{Handler: edgeHttpHandler}
+		if err := httpSrv.Serve(muxRestListener); err != nil && !errors.Is(err, http.ErrServerClosed) {
+			log.Fatalf("HTTP Serve: %v", err)
```
jakubno: don't fatal here, handle it gracefully (look at trafficProxy)

packages/client-proxy/main.go
```diff
@@ -248,11 +294,19 @@ func run() int {
...
+		if err != nil {
+			shutdownLogger.Error("edge rest api service shutdown error", zap.Error(err))
+		}
+
+		err = muxRpcListener.Close()
+		if err != nil {
+			shutdownLogger.Error("edge pass through proxy service shutdown error", zap.Error(err))
 		}
+
+		muxServer.Close()
```
jakubno, line 347: have you tested this stops gracefully?

jakubno (changes requested):

packages/shared/pkg/http/edge/spec.gen.go
sitole: There are still some requests (service discovery, sandbox catalog create)

packages/client-proxy/internal/edge/pool/edge-pool.go
sitole: yes, fixed thx

packages/shared/pkg/grpc/connection.go
sitole: Yep, handled here https://github.com/e2b-dev/infra/pull/809

packages/client-proxy/internal/edge-pass-through/proxy.go
sitole: agree, using more simple version without stream support

packages/client-proxy/internal/edge-pass-through/proxy.go
sitole: Renamed. The function getting method is inside, trying to extract the stream server from the context with `ServerTransportStreamFromContext`.

packages/client-proxy/internal/edge-pass-through/proxy.go
sitole: I checked and iam already using one2one implementation. Yes, there are nits in how checking EOF error or iteration over responses, but functionality is the same. I just simplified it into fewer methods to make it more compact.

packages/client-proxy/main.go
sitole: Issue resolved here: https://github.com/e2b-dev/infra/pull/812
For current implementation, i addeda  graceful shutdown for the grpc proxy and rest api server.

packages/client-proxy/internal/edge-pass-through/proxy.go
```diff
@@ -0,0 +1,215 @@
...
+
+func NewNodePassThrough(ctx context.Context, nodes *e2borchestrators.OrchestratorsPool, info *e2binfo.ServiceInfo, authorization authorization.AuthorizationService) *NodePassThrough {
+	nodePassThrough := &NodePassThrough{
+		authorization: authorization,
+		nodes:         nodes,
+		info:          info,
+		ctx:           ctx,
+	}
+
+	nodePassThrough.server = grpc.NewServer(
```
jakubno: it's little bit weird that the server is defined here, but handled out of the scope of this.

packages/client-proxy/main.go
```diff
@@ -155,8 +173,54 @@ func run() int {
...
+	muxServer := cmux.New(lis)
+
+	grpcListener := muxServer.MatchWithWriters(cmux.HTTP2MatchHeaderFieldSendSettings("content-type", "application/grpc")) // handler requests for gRPC pass through
+	grpcSrv := edgePassThroughHandler.GetServer()
+
+	restListener := muxServer.Match(cmux.Any())
+	restSrv := &http.Server{Handler: edgeHttpHandler} // handler requests for REST API
+
+	go func() {
+		err := grpcSrv.Serve(grpcListener)
```
jakubno:
```suggestion

	wg.Add(1)
	go func() {
		defer wg.Done()
		err := grpcSrv.Serve(grpcListener)
```

packages/client-proxy/main.go
```diff
@@ -155,8 +173,54 @@ func run() int {
...
+		case err != nil:
+			exitCode.Add(1)
+			zap.L().Error("edge grpc service encountered error", zap.Error(err))
+		default:
+			// this probably shouldn't happen...
+			zap.L().Error("edge grpc service exited without error")
+		}
+	}()
+
+	go func() {
```
jakubno, line 212: same here

packages/client-proxy/main.go
```diff
@@ -244,11 +308,20 @@ func run() int {
 		shutdownLogger.Info("waiting for unhealthy state propagation", zap.Float64("wait_in_seconds", shutdownUnhealthyWait.Seconds()))
 		time.Sleep(shutdownUnhealthyWait)
 
-		ginErr := edgeGinServer.Shutdown(ctx)
-		if ginErr != nil {
-			exitCode.Add(1)
-			shutdownLogger.Error("edge api shutdown error", zap.Error(ginErr))
+		// wait for graceful shutdown of the gRPC server with  pass through proxy
+		grpcSrv.GracefulStop()
```
jakubno, line 332: Can you make sure the instance manager won't kill it before this?

jakubno (changes requested):

packages/client-proxy/internal/edge-pass-through/proxy.go
sitole: true, will directly return grpc server so we do not need to deal with pass-through struct just to get grpc server on it

packages/client-proxy/main.go
sitole: - For instance, heal check we have set up that marks an instance unhealthy after 3*5 seconds so this covers 15 seconds of sleep we have here
- We are not using any stream for http/grpc connections so all connections should be safely shutdowned under <1s.
- Theoretically, the health check for auto scaler is separated from load balancer health checks so we can set up bigger numbers such as marking instance unhealthy (so ready for a kill) after 10 times 5 seconds or so.

Also looks like there is an option on the autoscaling group so we can set up a "wait for termination" lifecycle. Will explore this more

sitole: Okay, increased autoscaler health check so it will take UP to 50 seconds until the autoscaler receives info about an unhealthy node. Autoscaler reaction time is not described, but from my experience, it takes more than 10s seconds until the autoscaler will react to this.

packages/client-proxy/internal/edge-pass-through/proxy.go
```diff
@@ -0,0 +1,209 @@
...
+	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/types/known/emptypb"
+
+	"github.com/e2b-dev/infra/packages/proxy/internal/edge/authorization"
+	e2binfo "github.com/e2b-dev/infra/packages/proxy/internal/edge/info"
+	e2borchestrators "github.com/e2b-dev/infra/packages/proxy/internal/edge/pool"
+	api "github.com/e2b-dev/infra/packages/shared/pkg/http/edge"
+)
+
+type NodePassThrough struct {
```
jakubno: Can we rename to
```suggestion
type NodePassThroughServer struct {
```

packages/client-proxy/internal/edge-pass-through/proxy.go
```diff
@@ -0,0 +1,209 @@
...
+
+	grpcMetadataNodeIdKey        = "node-id"
+	grpcMetadataAuthorizationKey = "authorization"
+
+	grpcHealthMethodName = "/EdgePassThrough/healthcheck"
+)
+
+var clientStreamDescForProxying = &grpc.StreamDesc{ServerStreams: true, ClientStreams: true}
+
+func NewNodePassThrough(ctx context.Context, nodes *e2borchestrators.OrchestratorsPool, info *e2binfo.ServiceInfo, authorization authorization.AuthorizationService) *grpc.Server {
```
jakubno:
```suggestion
func NewNodePassThroughServer(ctx context.Context, nodes *e2borchestrators.OrchestratorsPool, info *e2binfo.ServiceInfo, authorization authorization.AuthorizationService) *grpc.Server {
```

packages/client-proxy/internal/edge/info/info.go
```diff
@@ -39,3 +40,16 @@ func (s *ServiceInfo) SetStatus(status api.ClusterNodeStatus) {
...
 	}
 }
+
+func (s *ServiceInfo) SetTerminating() {
+	s.statusMu.Lock()
+	defer s.statusMu.Unlock()
+	s.terminating = true
+}
+
+func (s *ServiceInfo) GetTerminating() bool {
```
jakubno: IsTerminating or only Terminating would be better name

packages/client-proxy/internal/edge/handlers/healthy.go
```diff
@@ -31,3 +31,14 @@ func (a *APIStore) HealthCheckTraffic(c *gin.Context) {
 	c.Status(http.StatusServiceUnavailable)
 	c.Writer.Write([]byte("unhealthy"))
 }
+
+func (a *APIStore) HealthCheckMachine(c *gin.Context) {
```
jakubno, line 38: can you add comments to all healthcheck explaining for what they are used

jakubno (approved):

PR #802, opened 2025-06-19
Increase inodes count

dobrac:
### Description
Increase the inode ratio to `4096`. This is the maximum ration that's possible with the `4096` block size. This should increase the count on the default +5G image from 178K to 1.7M.

This PR **also changes** the implementation of how **ext4 filesystem creation** works. It replaces the tar2ext4 library for the linux native command, and system mounting. Fixes:
- https://github.com/google/go-containerregistry/issues/977
- https://github.com/google/go-containerregistry/issues/2002

### New rootfs flattening
The rootfs unpacking works in this way now:
1. mkfs.ext4 creates a file with max size (15G)
2. the ext4 file is mounted using loopdevice
3. the OCI image is extracted to a temp folder and mounted using `overlayfs`
4. all files are copied from the mounted overlayFS to the mounted ext4 using `rsync`
5. everything is unmounted, cleaned
6. ext4 file is shrinked to a minimal size using `resize2fs -M`

This approach should also enable layer caching in the future, where only file copying step would be required instead of downloading and unpacking the layers (when cached)

_Another tried approach was unpacking the layers one after the other directly to the target ext4 filesystem. This was around 30% slower than doing it in parallel and copying it later on the base image._


packages/orchestrator/internal/template/build/tar.go
```diff
@@ -47,7 +48,7 @@ func LayerFile(filemap map[string]layerFile) (containerregistry.Layer, error) {
 	// Return a new copy of the buffer each time it's opened.
 	return tarball.LayerFromOpener(func() (io.ReadCloser, error) {
 		return io.NopCloser(bytes.NewBuffer(b.Bytes())), nil
-	})
+	}, tarball.WithMediaType(types.OCILayer))
```
jakubno: why is this needed?

packages/orchestrator/internal/template/build/rootfs.go
```diff
@@ -115,7 +115,20 @@ func (r *Rootfs) createExt4Filesystem(ctx context.Context, tracer trace.Tracer,
...
 	// Resize rootfs
-	rootfsFinalSize, err := ext4.Enlarge(ctx, tracer, rootfsPath, r.template.DiskSizeMB<<ToMBShift)
+	rootfsFreeSpace, err := ext4.GetFreeSpace(ctx, tracer, rootfsPath, r.template.RootfsBlockSize())
+	if err != nil {
+		return containerregistry.Config{}, fmt.Errorf("error getting free space: %w", err)
+	}
+	// We need to remove the remaining free space from the ext4 file size
+	// This is a residual space that could not be shrunk when creating the filesystem,
+	// but is still available for use
+	diskAdd := r.template.DiskSizeMB<<ToMBShift - rootfsFreeSpace
```
jakubno, line 126: can we be sure this is non negative?

packages/orchestrator/internal/template/build/oci/oci_test.go
```diff
@@ -0,0 +1,90 @@
...
+		t.Fatal(err)
+	}
+	tw.Close()
+
+	return &buf
+}
+
+func TestCreateExportLayersOrder(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
```
jakubno: NIT you can use only context.Background

packages/orchestrator/internal/template/build/oci/oci.go
```diff
@@ -106,3 +139,149 @@ func ParseEnvs(envs []string) map[string]string {
...
+		return file.Name()
+	}), nil
+}
+
+// copyFiles uses rsync to copy files from the source directory to the destination directory.
+func copyFiles(ctx context.Context, tracer trace.Tracer, src, dest string) error {
+	_, childSpan := tracer.Start(ctx, "copy-files")
+	defer childSpan.End()
+
+	cmd := exec.Command("rsync", "-a", "--whole-file", "--inplace", src+"/", dest)
```
jakubno: why (only) these flags? What exactly do we want to preserve here?

packages/orchestrator/internal/template/build/tar.go
dobrac: not anymore, will remove

packages/orchestrator/internal/template/build/rootfs.go
dobrac: good catch

packages/orchestrator/internal/template/build/oci/oci.go
dobrac: `-a` option preserves all metadata and copies all files
`--whole-file` is an optimization (Normally, rsync uses a delta-transfer algorithm, which breaks files into blocks and transfers only the changed parts to save bandwidth. Disables compression and block-based checksums, making it faster over local or high-speed networks.)
`--inplace` is an optimization (By default, rsync creates a temporary file during transfer, then replaces the destination file once the transfer completes. This changes it.)

jakubno: I think it would be worth it to write down what -a does
```
Recursion into directories
Symlinks
Permissions
Modification times
Group/owner (if possible)
Device files and special files
```

But my question was rather opposite, why did you decide to exclude e.g. hard links?

dobrac: I think it's more that I forgot to preserve the hard links, as that's not the default

dobrac: Do you have any other properties that you think are missing?

dobrac: Added the `-H` option + comment with explanation

jakubno: The docs say using `-H` is an expensive operation, and it seems the performance penalty is there even if you don't have any hard links. Can you check how much slower this becomes?

jakubno: Maybe some of these?
```
 -E, --executability         preserve executability
 -X, --xattrs                preserve extended attributes
 ```

Comment (dobrac) [reply packages/orchestrator/internal/template/build/oci/oci.go]:
```
archive mode; equals -rlptgoD (no -H,-A,-X)
```

executability is included with perms, not sure what are the extended attributes used for

Comment (dobrac) [reply packages/orchestrator/internal/template/build/oci/oci.go]: on the base template, the difference is less than 1s (for the -H flag)

File: packages/orchestrator/internal/template/build/rootfs.go [hunk 5]
```diff
@@ -128,11 +129,13 @@ func (r *Rootfs) createExt4Filesystem(ctx context.Context, tracer trace.Tracer,
...
 	)
-	rootfsFinalSize, err := ext4.Enlarge(ctx, tracer, rootfsPath, diskAdd)
-	if err != nil {
-		return containerregistry.Config{}, fmt.Errorf("error enlarging rootfs: %w", err)
+	if diskAdd > 0 {
+		rootfsFinalSize, err := ext4.Enlarge(ctx, tracer, rootfsPath, diskAdd)
+		if err != nil {
+			return containerregistry.Config{}, fmt.Errorf("error enlarging rootfs: %w", err)
+		}
+		r.template.rootfsSize = rootfsFinalSize
```
jakubno, line 137: why do you assign it here? why not for both cases this if statement? (you would have to save result of Enlarge to ext4Size

dobrac: I would have to use the `=` instead of `:=` in the if statement, which is imo tricky, as there could be future consequences when doing refactoring. This approach handles the logic localized, so you don't need the whole `createExt4Filesystem` function context to understand it

jakubno (approved):

PR #812, opened 2025-06-20
Fixing broken client proxy graceful shutdown

sitole:
- Fixed health check path on load balancer (we have one path for edge service and a second only for proxy traffic).
- Removed use of `env.IsDevelopment()` that was causing to skipp shutdown waits to be skipped, because the default function output is true.
- Adjusted logging to be more clear what server was stopped


sitole:
- Added fix to the `env.IsDevelopment()` handler that by default takes production as the current env.
- Cluster env propagated to Nomad job for edge service.

jakubno (approved):

PR #815, opened 2025-06-20
check integrity after ext4 creation

dobrac:
Check integrity after ext4 creation. Helps with:
```
Please run 'e2fsck -f /tmp/build-templates/.../rootfs.ext4.build first
```


0div (commented):
did u doublecheck if integrity isn't already checked in [Build func](https://github.com/e2b-dev/infra/blob/d537c7fe0a3794a862c3f8f673010c6c8e8bd142/packages/orchestrator/internal/template/build/build.go#L26) called upstream in the callstack?

jakubno (approved):

PR #819, opened 2025-06-23
Update self-host.md

tomassrnka:
Update disk space requirements, one of the last commits added Clickhouse, which increases disk space needs, see the quota status after fresh deployment. 
Screenshot 2025-06-23 at 19 15 02


jakubno:
Thanks @tomassrnka!

PR #820, opened 2025-06-23
Rewrite GRPC connection

jakubno:
# Description

- Don't block on creating client
- Which results in adding failed nodes to the list of all nodes with correct status


packages/api/internal/analytics_collector/client.go
```diff
@@ -2,17 +2,20 @@ package analyticscollector
...
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
 	"google.golang.org/protobuf/types/known/emptypb"
-
-	e2bgrpc "github.com/e2b-dev/infra/packages/shared/pkg/grpc"
 )
 
-var host = os.Getenv("ANALYTICS_COLLECTOR_HOST")
+// This isn't host, but a url. Remove the protocol part
+var host = regexp.MustCompile(`https?://`).ReplaceAllString(os.Getenv("ANALYTICS_COLLECTOR_HOST"), "")
```
dobrac: nit: rename to represent it's url

dobrac (changes requested):
LGTM, just one nit + waiting for passing tests

packages/api/internal/analytics_collector/client.go
```diff
@@ -24,11 +26,28 @@ func NewAnalytics() (*Analytics, error) {
 	var connection *grpc.ClientConn
 
 	if host == "" {
-		zap.L().Warn("Running dummy implementation of analytics collector client, no host provided")
+		zap.L().Warn("Running dummy implementation of analytics collector client, no url provided")
```
dobrac:
```suggestion
		zap.L().Warn("Running dummy implementation of analytics collector client, no host provided")
```

dobrac: just to be consistent now

PR #822, opened 2025-06-24
Refactor sandbox metrics observer

jakubno:
# Description

There were 2 shortcoming for sandbox observer:
- Register callback contains lock (and there's lock when collecting metrics), resulting in prolonged sandbox create times
- Callbacks are called serially

This combination results in long times before acquiring lock and registering the callbacks causing the sandbox start to take long time.

After refactor there's only 1 callback which collects metrics from all the sandboxes, getting metrics concurrently and making sure all the callbacks are done within the timeout.

Screenshots from my dev cluster (I added the trace):
Before: 
image


packages/orchestrator/internal/metrics/sandboxes.go
```diff
@@ -0,0 +1,206 @@
...
+func (so *SandboxObserver) startObserving() (metric.Registration, error) {
+	unregister, err := so.meter.RegisterCallback(
+		func(ctx context.Context, o metric.Observer) error {
+			sbxCount := so.sandboxes.Count()
+
+			wg := errgroup.Group{}
+			// Get all sandbox metrics at most in the export interval (there is a 1-second timeout for each sandbox metrics request)
+			concurrentRequests := math.Ceil(float64(sbxCount) / so.exportInterval.Seconds())
+			// Multiply by 2 to leave some room for other callbacks
+			wg.SetLimit(int(2 * concurrentRequests))
```
dobrac: What could be the other callbacks?

packages/orchestrator/internal/metrics/sandboxes.go
```diff
@@ -0,0 +1,206 @@
...
+					continue
+				}
+
+				wg.Go(func() error {
+					// Make sure the sandbox doesn't change while we are getting metrics (the slot could be assigned to another sandbox)
+					childCtx, cancel := context.WithTimeout(ctx, time.Second)
+					sbx.Checks.Lock()
+					sbxMetrics, err := sbx.GetMetrics(childCtx)
+					cancel()
+					sbx.Checks.Unlock()
```
dobrac: so, this could still lock e.g. pause by 1 second. Why don't we do the request and then verify that we've received the request from the sandbox we expect it from?

packages/orchestrator/internal/sandbox/checks.go
```diff
@@ -2,120 +2,94 @@ package sandbox
...
 	for {
 		select {
 		case <-healthTicker.C:
 			childCtx, cancel := context.WithTimeout(c.ctx, time.Second)
 
-			c.ctx.Lock()
+			c.mu.Lock()
 			c.Healthcheck(childCtx, false)
-			c.ctx.Unlock()
+			c.mu.Unlock()
```
dobrac: does the lock have still any purpose here?

packages/orchestrator/internal/metrics/sandboxes.go
jakubno: Host metrics or any other metrics we will add in the future

packages/orchestrator/internal/sandbox/checks.go
jakubno: Yes, if the lock is not here, the sandbox can be killed before the healthcheck resulting in incorrectly reporting unhealthy healthcheck, which looks like the sandbox died unexpectedly. 

The alternative is not report the state on sandbox kill

packages/orchestrator/internal/metrics/sandboxes.go
jakubno: Decreased to 50ms, when tested the time for getting the metrics was 1ms

packages/orchestrator/cmd/benchmark-metrics/main.go
jakubno, line 1: The benchmark isn't making much sense anymore. Since exporting is done asynchronously, we can't access any information. We can produce the metrics, but there's no feedback on how much was actually ingested or how long the export request took.

packages/orchestrator/internal/sandbox/metrics.go
```diff
@@ -42,36 +44,11 @@ func (s *Sandbox) GetMetrics(ctx context.Context) (*telemetry.SandboxMetrics, er
...
+	var m Metrics
 	err = json.NewDecoder(response.Body).Decode(&m)
 	if err != nil {
 		return nil, err
 	}
 
 	return &m, nil
 }
-
-func (s *Sandbox) LogMetricsThresholdExceeded(ctx context.Context) {
```
jakubno, line 54: moved to observing part

packages/orchestrator/internal/metrics/sandboxes.go
```diff
@@ -0,0 +1,208 @@
...
+	meter       metric.Meter
+	cpuTotal    metric.Int64ObservableGauge
+	cpuUsed     metric.Float64ObservableGauge
+	memoryTotal metric.Int64ObservableGauge
+	memoryUsed  metric.Int64ObservableGauge
+}
+
+const shiftFromMiBToBytes = 20 // Shift to convert MiB to bytes
+
+func NewSandboxObserver(ctx context.Context, commitSHA, clientID string, sandboxMetricsExportPeriod time.Duration, sandboxes *smap.Map[*sandbox.Sandbox]) (*SandboxObserver, error) {
```
jakubno, line 54: this was moved from telemetry/sandbox.go, only `registration, err := so.startObserving()` part was changed

packages/orchestrator/cmd/benchmark-metrics/main.go
```diff
@@ -1,150 +0,0 @@
...
-	fmt.Printf("Total Metrics: %d\n", totalMetrics)
-	fmt.Printf("Metrics emit interval: %.2fs\n", cfg.emitInterval.Seconds())
-	fmt.Printf("Metrics per second: %.2f\n", metricsPerSecond)
-	fmt.Printf("Nodes: %d\n", cfg.nodeCount)
-	fmt.Printf("Metrics per node: %d\n", totalMetrics/cfg.sandboxesPerNodeCount)
-
-	return nil
-}
-
-func main() {
```
dobrac, line 131: why is this removed?

packages/orchestrator/internal/sandbox/checks.go
```diff
@@ -2,128 +2,85 @@ package sandbox
...
-	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
-	"github.com/e2b-dev/infra/packages/shared/pkg/utils"
 )
 
 const (
-	healthCheckInterval      = 20 * time.Second
-	metricsCheckInterval     = 5 * time.Second
-	minEnvdVersionForMetrics = "0.1.5"
+	healthCheckInterval = 20 * time.Second
+	healthCheckTimeout  = 50 * time.Millisecond
```
dobrac: it was a second previously, is 20 millisecond long enough?

packages/orchestrator/internal/sandbox/checks.go
```diff
@@ -2,128 +2,85 @@ package sandbox
...
 		}
 	}
 }
 
-func (c *Checks) Healthcheck(ctx context.Context, alwaysReport bool) {
-	ok, err := c.sandbox.HealthCheck(ctx)
+func (c *Checks) Healthcheck(alwaysReport bool) {
+	ok, err := c.GetHealth(healthCheckTimeout)
+	// Sandbox stopped
+	if errors.Is(err, ErrChecksStopped) {
```
dobrac, line 81: is this actually working? The context will have error "canceled" with cause "ErrChecksStopped"

packages/orchestrator/internal/metrics/sandboxes.go
```diff
@@ -0,0 +1,209 @@
...
+			for _, sbx := range so.sandboxes.Items() {
+				if !utils.IsGTEVersion(sbx.Config.EnvdVersion, minEnvdVersionForMetrics) {
+					continue
+				}
+
+				if !sbx.Checks.UseClickhouseMetrics {
+					continue
+				}
+
+				sbx := sbx // Capture the sandbox in the loop variable to avoid data race
```
dobrac: just a note, the config is a pointer (not copied data), so I'm not 100% sure what was the data race

dobrac (changes requested):

packages/orchestrator/cmd/benchmark-metrics/main.go
jakubno: The benchmark isn't making much sense anymore. Since exporting is done asynchronously, we can't access any information. We can produce the metrics, but there's no feedback on how much was actually ingested or how long the export request took.

packages/orchestrator/internal/sandbox/checks.go
jakubno: it takes less than 1 ms, I can increase it a little, but if the request takes more than e.g. 100ms, than it probably isn't responding to your requests either

dobrac: Ok, I would put there at least 100ms. It isn't only the request timeout, but the whole context timeout

packages/orchestrator/internal/metrics/sandboxes.go
jakubno: the problem is that go reuses

jakubno: This behavior was fixed in go 1.22, not each loop creates a new variable instead of reusing the same one, preventing gotchas like this.

Removed

packages/orchestrator/internal/sandbox/checks.go
jakubno: yes, i've tested it
image
image

PR #823, opened 2025-06-24
Added row lever security for cluster table

sitole:
Enable row-level security for the clusters table. Needed because of the Supabase access model.


packages/db/migrations/20250624001047_deploy_cluster_policy.sql
```diff
@@ -0,0 +1,8 @@
+-- +goose Up
+-- +goose StatementBegin
+ALTER TABLE "public"."clusters" ENABLE ROW LEVEL SECURITY;
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
```
jakubno, line 7: where is the down migration?

jakubno (approved):

PR #824, opened 2025-06-24
Feat: Add Supabase auth headers to api CORS policy

ben-fornefeld:
In order to enable usage of the `e2b` SDK inside a browser environment + authorizing to the api via the user's `Supabase Token`, this PR updates the CORS `AllowedHeaders` policy accordingly

### DIFF
--- a/packages/api/main.go
+++ b/packages/api/main.go
@@ -97,6 +97,9 @@ func NewGinServer(ctx context.Context, tel *telemetry.Client, logger *zap.Logger
 		// API Key header
 		"Authorization",
 		"X-API-Key",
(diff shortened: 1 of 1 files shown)


dobrac (approved):

dobrac (changes requested):
This allows managing the sandboxes create/kill from browser, but not the folders/files, etc.

ben-fornefeld:
> This allows managing the sandboxes create/kill from browser, but not the folders/files, etc.

i might specify - from browser with access to the supabase auth tokens of a user on the same cluster. api keys worked already inside browser environments before this addition

PR #825, opened 2025-06-24
API - Orchestrator status - Return unhealthy if cant connect

jakubno:
# Description

- Report correct status for the node if we can't connect to it
- Add a metric to monitor node status

### DIFF
--- a/packages/api/internal/orchestrator/cache.go
+++ b/packages/api/internal/orchestrator/cache.go
@@ -166,7 +166,7 @@ func (o *Orchestrator) syncNode(ctx context.Context, node *Node, nodes []*node.N
 
 	if !syncRetrySuccess {
 		zap.L().Error("Failed to sync node after max retries, temporarily marking as draining", zap.String("node_id", node.Info.ID))
-		node.setStatus(api.NodeStatusDraining)
+		node.setStatus(api.NodeStatusUnhealthy)
 		return
 	}
 
--- a/packages/api/internal/orchestrator/cache.go
+++ b/packages/api/internal/orchestrator/cache.go
@@ -165,7 +165,7 @@ func (o *Orchestrator) syncNode(ctx context.Context, node *Node, nodes []*node.N
 	}
 
 	if !syncRetrySuccess {
-		zap.L().Error("Failed to sync node after max retries, temporarily marking as draining", zap.String("node_id", node.Info.ID))
+		zap.L().Error("Failed to sync node after max retries, temporarily marking as unhealthy", zap.String("node_id", node.Info.ID))
 		node.setStatus(api.NodeStatusUnhealthy)
 		return
 	}
--- a/packages/api/internal/orchestrator/client.go
+++ b/packages/api/internal/orchestrator/client.go
@@ -95,6 +95,7 @@ func (o *Orchestrator) connectToNode(ctx context.Context, node *node.NodeInfo) e
 	nodeStatus := api.NodeStatusUnhealthy
 	nodeVersion := "unknown"
 	nodeCommit := "unknown"
+	orchestratorID := node.ID
 
 	ok, err := o.getNodeHealth(node)
 	if err != nil {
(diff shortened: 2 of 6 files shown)


packages/api/internal/orchestrator/cache.go
```diff
@@ -166,7 +166,7 @@ func (o *Orchestrator) syncNode(ctx context.Context, node *Node, nodes []*node.N
 
 	if !syncRetrySuccess {
 		zap.L().Error("Failed to sync node after max retries, temporarily marking as draining", zap.String("node_id", node.Info.ID))
```
dobrac:
```suggestion
		zap.L().Error("Failed to sync node after max retries, temporarily marking as unhealthy", zap.String("node_id", node.Info.ID))
```

packages/api/internal/orchestrator/metrics.go
```diff
@@ -0,0 +1,38 @@
...
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
+
+	"github.com/e2b-dev/infra/packages/api/internal/api"
+	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
+)
+
+func (o *Orchestrator) setupMetrics(meterProvider metric.MeterProvider) error {
+	meter := meterProvider.Meter("api.orchestrator")
+	_, err := telemetry.GetObservableUpDownCounter(meter, telemetry.ApiOrchestratorCountMeterName, func(ctx context.Context, observer metric.Int64Observer) error {
```
dobrac: Is the count enough? Wouldn't reporting the status itself be more helpful, so you would be able to see also which nodes have which status?

jakubno: I was worried about number of labels, but we can change later if it will be too much

packages/api/internal/orchestrator/client.go
```diff
@@ -95,6 +95,7 @@ func (o *Orchestrator) connectToNode(ctx context.Context, node *node.NodeInfo) e
 	nodeStatus := api.NodeStatusUnhealthy
 	nodeVersion := "unknown"
 	nodeCommit := "unknown"
+	orchestratorID := node.ID
```
dobrac, line 98: why not call it nodeID?

packages/api/internal/orchestrator/metrics.go
```diff
@@ -7,25 +7,17 @@ import (
...
-			}
-			nodeStatus[status]++
-		}
-
-		for status, count := range nodeStatus {
-			observer.Observe(count, metric.WithAttributes(attribute.String("status", string(status))))
+		for _, node := range o.nodes.Items() {
+			observer.Observe(1, metric.WithAttributes(
+				attribute.String("status", string(node.status)),
+				attribute.String("node_id", node.orchestratorID),
```
dobrac: node_id -> node.id

packages/api/internal/orchestrator/metrics.go
```diff
@@ -7,25 +7,17 @@ import (
...
-			if _, ok := nodeStatus[status]; !ok {
-				nodeStatus[status] = 0
-			}
-			nodeStatus[status]++
-		}
-
-		for status, count := range nodeStatus {
-			observer.Observe(count, metric.WithAttributes(attribute.String("status", string(status))))
+		for _, node := range o.nodes.Items() {
+			observer.Observe(1, metric.WithAttributes(
```
dobrac: doesn't make sense to use any different type of metric than updown counter then?

dobrac (changes requested):

packages/api/internal/orchestrator/client.go
jakubno: Because nodeID in api is nomad node id and it could lead to some problems with a mismatch

packages/api/internal/orchestrator/metrics.go
jakubno: Which one? It doesn't really matter as it is always `1`.

dobrac (approved):

PR #828, opened 2025-06-25
Improve logs retrieval speed

dobrac:


packages/api/internal/handlers/template_build_status.go
```diff
@@ -83,21 +83,43 @@ func (a *APIStore) GetTemplatesTemplateIDBuildsBuildIDStatus(c *gin.Context, tem
...
+	offset := params.LogsOffset
+
+	logsProviders := make([]buildlogs.Provider, 0)
+	logsProviders = append(logsProviders, &buildlogs.TemplateManagerProvider{TemplateManager: a.templateManager})
+	logsProviders = append(logsProviders, &buildlogs.LokiProvider{LokiClient: a.lokiClient})
+	logsProviders = append(logsProviders, &buildlogs.ClusterPlacementProvider{TemplateManager: a.templateManager})
+
+	logsTotal := make([]string, 0)
+	for _, provider := range logsProviders {
+		logs, err := provider.GetLogs(ctx, templateID, buildUUID, team.ClusterID, buildInfo.ClusterNodeID, offset)
```
jakubno: if there are no data and you are running build on our infra, it can result in panic as `ClusterPlacementProvider` http client is `nil`

packages/api/internal/template-manager/template_manager.go
```diff
@@ -125,39 +125,38 @@ func (tm *TemplateManager) BuildsStatusPeriodicalSync(ctx context.Context) {
 	}
 }
 
-func (tm *TemplateManager) getBuilderClient(clusterID *uuid.UUID, nodeID *string, placement bool) (*grpclient.GRPCClient, metadata.MD, PlacementLogsProvider, error) {
+func (tm *TemplateManager) GetBuilderClient(clusterID *uuid.UUID, nodeID *string, placement bool) (*edge.ClusterGRPC, *edge.ClusterHTTP, error) {
```
jakubno: Maybe this function should be split. We are always using either grpc, which is fine for both, but if we use http, it should be only for BYOC, otherwise return an error

packages/orchestrator/internal/template/server/create_template.go
```diff
@@ -78,11 +76,12 @@ func (s *ServerStore) TemplateCreate(ctx context.Context, templateRequest *templ
...
-		// Wait for the CLI to load all the logs
-		// This is a temporary ~fix for the CLI to load most of the logs before finishing the template build
-		// Ideally we should wait in the CLI for the last log message
-		time.Sleep(8 * time.Second)
+		externalWriter := writer.New(logger)
+		defer externalWriter.Close()
+
+		logsWriter := io.MultiWriter(&logs, externalWriter)
+		res, err := s.builder.Build(buildContext, template, logsWriter)
+		externalWriter.Sync()
```
jakubno: if you want to ignore the error, make it explicit

packages/orchestrator/internal/template/server/create_template.go
```diff
@@ -39,30 +40,27 @@ func (s *ServerStore) TemplateCreate(ctx context.Context, templateRequest *templ
...
+		DiskSizeMB: int64(config.DiskSizeMB),
+		HugePages:  config.HugePages,
 	}
 
-	buildInfo, err := s.buildCache.Create(config.BuildID)
+	logger := s.buildLogger.
+		With(zap.Field{Type: zapcore.StringType, Key: "envID", String: config.TemplateID}).
+		With(zap.Field{Type: zapcore.StringType, Key: "buildID", String: config.BuildID})
+
+	var logs bytes.Buffer
```
jakubno: bytes.Buffer isn't thread safe, writing logs can also cause you will miss some logs when asking for status

packages/api/internal/template-manager/logs/template_manager_provider.go
```diff
@@ -0,0 +1,54 @@
...
+	grpc, _, err := t.TemplateManager.GetBuilderClient(clusterID, clusterNodeID, false)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get builder edgeHttpClient: %w", err)
+	}
+
+	logs := make([]string, 0)
+
+	reqCtx := metadata.NewOutgoingContext(ctx, grpc.Metadata)
+	res, err := grpc.Client.Template.TemplateBuildStatus(
+		reqCtx, &templatemanagergrpc.TemplateStatusRequest{
```
jakubno, line 22: do you want to return all logs (no offset)?

jakubno (changes requested):

packages/api/internal/handlers/template_build_status.go
dobrac: nice catch! Refactored to use build client

packages/api/internal/template-manager/template_manager.go
dobrac: refactored

packages/orchestrator/internal/template/server/create_template.go
dobrac: the error is handled on the next line, it shouldn't be ignored, otherwise linter would scream

packages/api/internal/template-manager/logs/template_manager_provider.go
dobrac: what do you mean? The offset is managed on the API level

packages/orchestrator/internal/template/server/create_template.go
dobrac: nice catch, made it synchronized now

packages/api/internal/template-manager/logs/template_manager_provider.go
jakubno: That you will send all logs on each status call, basically asking if it isn't better to propagate the offset

packages/orchestrator/internal/template/server/create_template.go
jakubno: I meant the error from `Sync()`, also wouldn't simple defer work here?

packages/api/internal/template-manager/template_manager.go
```diff
@@ -125,39 +126,64 @@ func (tm *TemplateManager) BuildsStatusPeriodicalSync(ctx context.Context) {
...
+		return tm.GetClusterBuildClient(*clusterID, *nodeID)
+	}
+}
 
-		// for getting build information only not valid state is getting already unhealthy builder
-		if tm.GetLocalClientStatus() == infogrpc.ServiceInfoStatus_OrchestratorUnhealthy {
-			zap.L().Error("Local template manager is unhealthy")
-			return nil, nil, nil, ErrLocalTemplateManagerNotAvailable
-		}
+func (tm *TemplateManager) GetLocalBuildClient(clusterID *uuid.UUID, nodeID *string, placement bool) (*BuildClient, error) {
```
jakubno: this doesn't need cluster nor node id

jakubno: I think the placement logic doesn't make sense here, the logic is different for BYOC and it should be probably handled inside the build function (you have to handle this there anyway as the status can change in the meantime)

jakubno (changes requested):

packages/api/internal/template-manager/logs/template_manager_provider.go
dobrac: maybe, will check

packages/orchestrator/internal/template/server/create_template.go
dobrac: I see, sure, I can ignore it explicitly

defer won't work, I need to have it synced before reporting any status back

jakubno: could you explain why you need call sync before reporting the status?

dobrac: I need to sync the logs before reporting a status because when the CLI will fetch status of the build (with logs), and it will see that the build has finished, no new request for logs will happen. At that point, the logs need to be complete.  If they weren't complete, missing lines would never be fetched again.

packages/api/internal/template-manager/template_manager.go
dobrac: Fixed the ids. I agree with the placement logic, but this is probably change for another PR (cc @sitole )

packages/api/internal/template-manager/logs/template_manager_provider.go
dobrac: refactored

packages/orchestrator/internal/template/server/create_template.go
jakubno: but logs from external writer goes through vector to loki, so you can't really coordinate it only with this, right?

dobrac: Here is the handling in the CLI:

```
case 'error':
        template.logs.forEach((line) =>
          process.stdout.write(asBuildLogs(stripAnsi.default(line)))
        )
        throw new Error(
          `\n❌ Building sandbox template ${asFormattedSandboxTemplate({
            aliases,
            ...template,
          })} failed.\nCheck the logs above for more details or contact us ${asPrimary(
            '(https://e2b.dev/docs/support)'
          )} to get help.\n`
        )
```

dobrac: > but logs from external writer goes through vector to loki, so you can't really coordinate it only with this, right?

Hmm, thats true, removing the sync

packages/api/internal/template-manager/logs/template_manager_provider.go
```diff
@@ -0,0 +1,41 @@
...
+			TemplateID: templateID,
+			BuildID:    buildID,
+			Offset:     offset,
+		},
+	)
+
+	logs := make([]string, 0)
+	if err == nil {
+		for _, entry := range res.GetLogs() {
+			logs = append(logs, fmt.Sprintf("%s\n", entry))
```
jakubno: why are you adding an extra newilne?

packages/api/internal/template-manager/logs/template_manager_provider.go
```diff
@@ -0,0 +1,41 @@
...
+		}
+	} else {
+		telemetry.ReportError(ctx, "error when returning logs for template build", err)
+		zap.L().Error("error when returning logs for template build", zap.Error(err), logger.WithBuildID(buildID))
+	}
+
+	return logs, err
```
jakubno: Why it isn't handled the "classic" way? Also you could probably iterate over the already present slice rather than allocate new one
```suggestion
	if err != nil {
		telemetry.ReportError(ctx, "error when returning logs for template build", err)
		zap.L().Error("error when returning logs for template build", zap.Error(err), logger.WithBuildID(buildID))
		return nil, err
	}

	logs := res.GetLogs()
	// Add an extra newline
	for i := range res.GetLogs() {
		logs[i] += "\n"
	}

	return logs, nil
```

packages/api/internal/template-manager/logs/template_manager_provider.go
dobrac: the CLI expects it, but I didn't want to pollute the template managers too

packages/orchestrator/internal/template/build/writer/writer.go
```diff
@@ -17,3 +17,11 @@ func New(logger *zap.Logger) BuildLogsWriter {
 	writer := BuildLogsWriter{logger: logger}
 	return writer
 }
+
+func (w BuildLogsWriter) Sync() error {
```
jakubno: this is now unused

PR #829, opened 2025-06-25
Report build status reason

dobrac:
Report build result reason. This is used especially on build failure

### DIFF
--- a/packages/api/internal/template-manager/template_status.go
+++ b/packages/api/internal/template-manager/template_status.go
@@ -68,7 +68,7 @@ func (tm *TemplateManager) BuildStatusSync(ctx context.Context, buildID uuid.UUI
 
 type templateManagerClient interface {
 	SetStatus(ctx context.Context, templateID string, buildID uuid.UUID, status envbuild.Status, reason *string) error
-	SetFinished(ctx context.Context, templateID string, buildID uuid.UUID, rootfsSize int64, envdVersion string, reason *string) error
+	SetFinished(ctx context.Context, templateID string, buildID uuid.UUID, rootfsSize int64, envdVersion string) error
 	GetStatus(ctx context.Context, buildId uuid.UUID, templateID string, clusterID *uuid.UUID, clusterNodeID *string) (*templatemanagergrpc.TemplateBuildStatusResponse, error)
 }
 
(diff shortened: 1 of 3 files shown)


jakubno (changes requested):
~~Currently the reason is returned only for failed builds? I would change the spec/name of the field and improve description~~

I missed part of the code. Is there a reason why completed build doesn't have reason? How is it different from e.g. statu waiting?

packages/api/internal/template-manager/template_status.go
```diff
@@ -66,8 +67,8 @@ func (tm *TemplateManager) BuildStatusSync(ctx context.Context, buildID uuid.UUI
 }
 
 type templateManagerClient interface {
-	SetStatus(ctx context.Context, templateID string, buildID uuid.UUID, status envbuild.Status, reason string) error
-	SetFinished(ctx context.Context, templateID string, buildID uuid.UUID, rootfsSize int64, envdVersion string) error
+	SetStatus(ctx context.Context, templateID string, buildID uuid.UUID, status envbuild.Status, reason *string) error
+	SetFinished(ctx context.Context, templateID string, buildID uuid.UUID, rootfsSize int64, envdVersion string, reason *string) error
```
jakubno: is there a reason why SetFinished has reason when it's expected that it will be `nil`?

dobrac: ~~the reason is not expected to be nil for SetFinished~~ Edit: I can omit it and set it to nil

dobrac: updated

packages/shared/pkg/db/envs.go
```diff
@@ -191,6 +191,7 @@ func (db *DB) FinishEnvBuild(
 		SetTotalDiskSizeMB(totalDiskSizeMB).
 		SetStatus(envbuild.StatusUploaded).
 		SetEnvdVersion(envdVersion).
+		SetNillableReason(nil).
```
jakubno, line 194: do you need this line at all?
```suggestion
```

jakubno (approved):

packages/shared/pkg/db/envs.go
dobrac: yes, if there was any reason previously, this will remove it

jakubno: Ok, let's merge

PR #831, opened 2025-06-25
BYOC: Support sandboxes workflow for team clusters 

sitole:
Support for BYOC orchestrators and sandbox spawning in API.

### DIFF
--- a/packages/api/internal/orchestrator/cache.go
+++ b/packages/api/internal/orchestrator/cache.go
@@ -227,8 +227,8 @@ func (o *Orchestrator) syncNodeState(ctx context.Context, node *Node, instanceCa
 	syncRetrySuccess := false
 
 	for range syncMaxRetries {
-		client, reqCtx := node.GetClient(ctx)
-		nodeInfo, err := client.Info.ServiceInfo(reqCtx, &emptypb.Empty{})
+		client, ctx := node.GetClient(ctx)
+		nodeInfo, err := client.Info.ServiceInfo(ctx, &emptypb.Empty{})
 		if err != nil {
 			zap.L().Error("Error getting node info", zap.Error(err), logger.WithNodeID(node.Info.ID))
 			continue
@@ -348,8 +348,8 @@ func (o *Orchestrator) getDeleteInstanceFunction(
 			info.PauseDone(nil)
 		} else {
 			req := &orchestrator.SandboxDeleteRequest{SandboxId: info.Instance.SandboxID}
-			client, reqCtx := node.GetClient(ctx)
-			_, err := client.Sandbox.Delete(reqCtx, req)
+			client, ctx := node.GetClient(ctx)
+			_, err := client.Sandbox.Delete(ctx, req)
 			if err != nil {
 				return fmt.Errorf("failed to delete sandbox '%s': %w", info.Instance.SandboxID, err)
 			}
--- a/packages/api/internal/orchestrator/client.go
+++ b/packages/api/internal/orchestrator/client.go
@@ -161,8 +161,8 @@ func (o *Orchestrator) GetClient(ctx context.Context, nodeID string) (*grpclient
 		return nil, nil, fmt.Errorf("node '%s' not found", nodeID)
 	}
 
-	client, reqCtx := n.GetClient(ctx)
-	return client, reqCtx, nil
+	client, ctx := n.GetClient(ctx)
+	return client, ctx, nil
 }
 
 func (o *Orchestrator) getNodeHealth(node *node.NodeInfo) (bool, error) {
--- a/packages/api/internal/orchestrator/create_instance.go
+++ b/packages/api/internal/orchestrator/create_instance.go
@@ -187,8 +187,8 @@ func (o *Orchestrator) CreateSandbox(
 			CPUs:      build.Vcpu,
 		})
 
-		client, reqCtx := node.GetClient(childCtx)
-		_, err = client.Sandbox.Create(reqCtx, sbxRequest)
+		client, childCtx := node.GetClient(childCtx)
+		_, err = client.Sandbox.Create(childCtx, sbxRequest)
 		// The request is done, we will either add it to the cache or remove it from the node
 		if err == nil {
 			// The sandbox was created successfully
--- a/packages/api/internal/orchestrator/node.go
+++ b/packages/api/internal/orchestrator/node.go
@@ -122,8 +122,8 @@ func (n *Node) SendStatusChange(ctx context.Context, s api.NodeStatus) error {
 		return fmt.Errorf("unknown service info status: %s", s)
 	}
 
-	client, reqCtx := n.GetClient(ctx)
-	_, err := client.Info.ServiceStatusOverride(reqCtx, &orchestratorinfo.ServiceStatusChangeRequest{ServiceStatus: nodeStatus})
+	client, ctx := n.GetClient(ctx)
+	_, err := client.Info.ServiceStatusOverride(ctx, &orchestratorinfo.ServiceStatusChangeRequest{ServiceStatus: nodeStatus})
 	if err != nil {
 		zap.L().Error("Failed to send status change", zap.Error(err))
 		return err
--- a/packages/api/internal/orchestrator/list_cached_builds.go
+++ b/packages/api/internal/orchestrator/list_cached_builds.go
@@ -15,12 +15,12 @@ func (o *Orchestrator) listCachedBuilds(ctx context.Context, nodeID string) ([]*
 	childCtx, childSpan := o.tracer.Start(ctx, "list-cached-builds")
 	defer childSpan.End()
 
-	client, reqCtx, err := o.GetClient(childCtx, nodeID)
+	client, childCtx, err := o.GetClient(childCtx, nodeID)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get GRPC client: %w", err)
 	}
 
-	res, err := client.Sandbox.ListCachedBuilds(reqCtx, &empty.Empty{})
+	res, err := client.Sandbox.ListCachedBuilds(childCtx, &empty.Empty{})
 
 	err = utils.UnwrapGRPCError(err)
 	if err != nil {
--- a/packages/api/internal/orchestrator/list_instances.go
+++ b/packages/api/internal/orchestrator/list_instances.go
@@ -19,12 +19,12 @@ func (o *Orchestrator) getSandboxes(ctx context.Context, node *nNode.NodeInfo) (
 	childCtx, childSpan := o.tracer.Start(ctx, "get-sandboxes-from-orchestrator")
 	defer childSpan.End()
 
-	client, reqCtx, err := o.GetClient(childCtx, node.ID)
+	client, childCtx, err := o.GetClient(childCtx, node.ID)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get GRPC client: %w", err)
 	}
 
-	res, err := client.Sandbox.List(reqCtx, &empty.Empty{})
+	res, err := client.Sandbox.List(childCtx, &empty.Empty{})
 
 	err = utils.UnwrapGRPCError(err)
 	if err != nil {
--- a/packages/api/internal/orchestrator/pause_instance.go
+++ b/packages/api/internal/orchestrator/pause_instance.go
@@ -85,13 +85,13 @@ func snapshotInstance(ctx context.Context, orch *Orchestrator, sbx *instance.Ins
 	childCtx, childSpan := orch.tracer.Start(ctx, "snapshot-instance")
 	defer childSpan.End()
 
-	client, reqCtx, err := orch.GetClient(childCtx, sbx.Node.ID)
+	client, childCtx, err := orch.GetClient(childCtx, sbx.Node.ID)
 	if err != nil {
 		return fmt.Errorf("failed to get client '%s': %w", sbx.Node.ID, err)
 	}
 
 	_, err = client.Sandbox.Pause(
-		reqCtx, &orchestrator.SandboxPauseRequest{
+		childCtx, &orchestrator.SandboxPauseRequest{
 			SandboxId:  sbx.Instance.SandboxID,
 			TemplateId: templateID,
 			BuildId:    buildID,
--- a/packages/api/internal/orchestrator/update_instance.go
+++ b/packages/api/internal/orchestrator/update_instance.go
@@ -28,13 +28,13 @@ func (o *Orchestrator) UpdateSandbox(
 	)
 	defer childSpan.End()
 
-	client, reqCtx, err := o.GetClient(childCtx, nodeID)
+	client, childCtx, err := o.GetClient(childCtx, nodeID)
 	if err != nil {
 		return fmt.Errorf("failed to get client '%s': %w", nodeID, err)
 	}
 
 	_, err = client.Sandbox.Update(
-		reqCtx, &orchestrator.SandboxUpdateRequest{
+		childCtx, &orchestrator.SandboxUpdateRequest{
 			SandboxId: sandboxID,
 			EndTime:   timestamppb.New(endTime),
 		},


packages/api/internal/orchestrator/node.go
```diff
@@ -139,12 +149,25 @@ func (o *Orchestrator) GetNode(nodeID string) *Node {
 	return n
 }
 
+// clusterNodeID - this way we don't need to worry about multiple clusters with the same node ID in shared pool
+func (o *Orchestrator) clusterNodeID(clusterID uuid.UUID, nodeID string) string {
+	clusterPrefix := clusterID.String()[0:7]
+	return fmt.Sprintf("%s-%s", clusterPrefix, nodeID)
+}
```
sitole: Thinking if there is a better way. We want to prefix cluster nodes so there is no collision where multiple clusters have the same node ID.

This takes the first 8 chars from the cluster UUID (generated by us) and puts it before the node ID.

packages/api/internal/orchestrator/create_instance.go
```diff
@@ -260,19 +268,22 @@ func (o *Orchestrator) CreateSandbox(
 		}
 	}
 
+	// we need to inform remote cluster proxy about newly spawned sandbox so it's registered in sandbox traffic proxy
+	go o.RegisterSandboxInsideClusterCatalog(node, startTime, sbxRequest.Sandbox)
```
sitole: Because the edge is now just a gRPC proxy, we need to somehow tell it that there is a new sandbox and where it's placed for traffic routing. 

Is there a better place for this?

packages/api/internal/orchestrator/delete_instance.go
```diff
@@ -17,5 +17,13 @@ func (o *Orchestrator) DeleteInstance(ctx context.Context, sandboxID string, pau
...
 	)
 
+	c, err := o.instanceCache.Get(sandboxID)
+	if err != nil && c != nil {
+		node := o.GetNode(c.Node.ID)
+		if node != nil {
+			go o.RemoveSandboxFromClusterCatalog(node, sandboxID, c.ExecutionID)
+		}
+	}
+
```
sitole: Edge API catalog removal, not sure if this is the proper place to do it

packages/api/internal/edge/cluster.go
```diff
@@ -90,36 +92,40 @@ func (c *Cluster) Close() error {
...
 }
 
-func (c *Cluster) GetAvailableTemplateBuilder(ctx context.Context) (*ClusterNode, error) {
-	_, span := c.tracer.Start(ctx, "template-builder-get-available-node")
+func (c *Cluster) GetInstanceByNodeID(nodeID string) (*ClusterOrchestratorInstance, bool) {
+	return c.instances.Get(nodeID)
+}
+
+func (c *Cluster) GetAvailableTemplateBuilder(ctx context.Context) (*ClusterOrchestratorInstance, error) {
+	_, span := c.tracer.Start(ctx, "template-builder-get-available-instance")
```
dobrac, line 123: use the context (just pending comment from earlier)

packages/api/internal/orchestrator/node.go
sitole: Resolved, NodeID is no longer returned to the user, because we stopped using the client part in sandbox, so it can be longer and contain characters that are not sandbox ID friendly, such as dashes.

packages/api/internal/orchestrator/create_instance.go
sitole: I make it sync so there is no race condition when the sandbox is created and returned to the user, but the catalog registry request has not finished yet.

For creation, not sure, but we should maybe return an error during sandbox creation when the catalog write will fail?

sitole: Still not sure if this logic is okay. Potentionally it will slow down sandbox spawn requests.

packages/api/internal/template-manager/template_manager.go
```diff
@@ -167,13 +167,13 @@ func (tm *TemplateManager) GetClusterBuildClient(clusterID uuid.UUID, nodeID str
 		return nil, errors.New("cluster not found")
 	}
 
-	node, err := cluster.GetTemplateBuilderByID(nodeID)
+	i, err := cluster.GetTemplateBuilderByNodeID(nodeID)
```
jakubno: why `i`?

packages/api/internal/edge/cluster.go
```diff
@@ -25,8 +27,8 @@ type Cluster struct {
 	httpClient *api.ClientWithResponses
 	grpcClient *grpclient.GRPCClient
 
-	nodes           *smap.Map[*ClusterNode]
-	synchronization *synchronization.Synchronize[api.ClusterOrchestratorNode, *ClusterNode]
+	instances       *smap.Map[*ClusterOrchestratorInstance]
```
jakubno: is there a reason behind the renaming?

packages/api/internal/edge/cluster.go
```diff
@@ -141,3 +146,50 @@ func (c *Cluster) GetGRPC(serviceInstanceID string) *ClusterGRPC {
...
+	body := api.V1SandboxCatalogCreateJSONRequestBody{
+		OrchestratorID: serviceInstanceID,
+
+		ExecutionID:      sandboxConfig.ExecutionId,
+		SandboxID:        sandboxConfig.SandboxId,
+		SandboxMaxLength: sandboxConfig.MaxSandboxLength,
+		SandboxStartTime: sandboxStartTime,
+	}
+
+	_, err := c.httpClient.V1SandboxCatalogCreate(context.Background(), body)
```
jakubno: why is background context here?

packages/api/internal/edge/cluster.go
```diff
@@ -141,3 +146,50 @@ func (c *Cluster) GetGRPC(serviceInstanceID string) *ClusterGRPC {
...
+	return nil
+}
+
+func (c *Cluster) RemoveSandboxFromCatalog(sandboxID string, executionID string) error {
+	body := api.V1SandboxCatalogDeleteJSONRequestBody{
+		SandboxID:   sandboxID,
+		ExecutionID: executionID,
+	}
+
+	_, err := c.httpClient.V1SandboxCatalogDelete(context.Background(), body)
```
jakubno: same here

packages/api/internal/edge/cluster_instances.go
```diff
@@ -34,64 +34,64 @@ type ClusterNode struct {
...
 	return slices.Contains(n.roles, r)
 }
 
-func (n *ClusterNode) IsBuilderNode() bool {
+func (n *ClusterOrchestratorInstance) IsBuilder() bool {
 	return n.hasRole(infogrpc.ServiceInfoRole_TemplateBuilder)
 }
 
-func (n *ClusterNode) IsOrchestratorNode() bool {
+func (n *ClusterOrchestratorInstance) IsOrchestrator() bool {
```
jakubno: the struct name is little bit confusing here. Is `ClusterOrchestratorInstance` an orchestrator?

packages/api/internal/orchestrator/client.go
```diff
@@ -114,13 +119,49 @@ func (o *Orchestrator) connectToNode(ctx context.Context, node *node.NodeInfo) e
...
+
+		buildCache:     buildCache,
+		sbxsInProgress: smap.New[*sbxInProgress](),
+		createFails:    atomic.Uint64{},
+	}
+
+	o.nodes.Insert(poolNodeID, orchestratorNode)
+}
+
+func (o *Orchestrator) GetClient(nodeID string) (*grpclient.GRPCClient, metadata.MD, error) {
```
jakubno: wouldn't it be better to return some struct with GRPCClient, metadata and some helper fuction for creating the outgoing context? (`NewOutgoingContext`)

packages/api/internal/orchestrator/node.go
```diff
@@ -140,12 +149,24 @@ func (o *Orchestrator) GetNode(nodeID string) *Node {
 	return n
 }
 
+// clusterNodeID - this way we don't need to worry about multiple clusters with the same node ID in shared pool
+func (o *Orchestrator) clusterNodeID(clusterID uuid.UUID, nodeID string) string {
```
jakubno, line 166:
```suggestion
func (o *Orchestrator) getClusterNodeID(clusterID uuid.UUID, nodeID string) string {
```

packages/api/internal/orchestrator/cache.go
```diff
@@ -78,40 +79,118 @@ func (o *Orchestrator) syncNodes(ctx context.Context, instanceCache *instance.In
...
+		for _, n := range cluster.GetOrchestrators() {
+			poolNodeID := o.clusterNodeID(cluster.ID, n.NodeID)
+
+			// If the node is not in the list, connect to it
+			if o.GetNode(poolNodeID) == nil {
+				wg.Add(1)
+				go func(n *edge.ClusterOrchestratorInstance) {
+					defer wg.Done()
+					o.connectToClusterNode(cluster, n)
+				}(n)
```
jakubno: FYI: you can already do only, from go 1.21 or something like that, go doesn't share the variable in range
no need to change

```suggestion
				go func() {
					defer wg.Done()
					o.connectToClusterNode(cluster, n)
				}()
```

packages/api/internal/orchestrator/client.go
```diff
@@ -114,13 +119,49 @@ func (o *Orchestrator) connectToNode(ctx context.Context, node *node.NodeInfo) e
...
 }
 
-func (o *Orchestrator) GetClient(nodeID string) (*grpclient.GRPCClient, error) {
+func (o *Orchestrator) connectToClusterNode(cluster *edge.Cluster, i *edge.ClusterOrchestratorInstance) {
+	// this way we don't need to worry about multiple clusters with the same node ID in shared pool
+	poolNodeID := o.clusterNodeID(cluster.ID, i.NodeID)
+	poolGrpc := cluster.GetGRPC(i.ServiceInstanceID)
+
+	buildCache := ttlcache.New[string, interface{}]()
+	go buildCache.Start()
```
jakubno, line 128: Is this ever stopped?

packages/api/internal/orchestrator/delete_instance.go
```diff
@@ -17,5 +17,16 @@ func (o *Orchestrator) DeleteInstance(ctx context.Context, sandboxID string, pau
 		zap.Bool("pause", pause),
 	)
 
+	c, err := o.instanceCache.Get(sandboxID)
+	if err != nil && c != nil {
```
jakubno: can this happen?

packages/api/internal/orchestrator/create_instance.go
```diff
@@ -311,6 +326,11 @@ func (o *Orchestrator) findLeastBusyNode(nodesExcluded map[string]*Node) (leastB
 			continue
 		}
 
+		// Node must be in the same cluster as requested
+		if node.ClusterID != clusterID {
+			continue
```
jakubno: I think this check should be first, because it will can happen very often?

packages/api/internal/orchestrator/cache.go
```diff
@@ -78,40 +79,118 @@ func (o *Orchestrator) syncNodes(ctx context.Context, instanceCache *instance.In
...
 	nodes, err := o.listNomadNodes(spanCtx)
 	if err != nil {
-		zap.L().Error("Error listing nodes", zap.Error(err))
-
+		zap.L().Error("Error listing orchestrator nodes", zap.Error(err))
 		return
 	}
 
+	o.syncLocalDiscoveredNodes(spanCtx, nodes)
+	o.syncClusterDiscoveredNodes(spanCtx)
```
jakubno: maybe these could be done in parallel?

packages/api/internal/edge/cluster.go
sitole: We want to name it properly. It's part of our discussion what is not and what is just an instance running on the node.

packages/api/internal/edge/cluster_instances.go
sitole: Yeah, naming is really strange here. Probably the name will be `ClusterInstance`? It can be both an orchestrator and/or a template builder.

packages/api/internal/orchestrator/node.go
sitole: Not sure what is current state is, but @dobrac mentioned we want to align new code with golang standard of not using `get` prefix (https://linear.app/e2b/project/bring-your-own-compute-8306ffb002a6/issues)

packages/api/internal/orchestrator/client.go
sitole: Hmm, looks like even for current orchestrators, we are not stopping it. Will do `Close()` method and stop it there.

packages/api/internal/template-manager/template_manager.go
sitole: Just an instance for short. Will do `instance` then.

packages/api/internal/orchestrator/client.go
sitole: Good idea! Adding

packages/api/internal/orchestrator/node.go
```diff
@@ -50,6 +60,18 @@ type Node struct {
 	createFails atomic.Uint64
 }
 
+func (n *Node) Close(closeClient bool) {
```
jakubno: wouldn't it be better to have two methods? `Close` and `CloseWithClient`? it's harder to read with only boolean

packages/api/internal/orchestrator/delete_instance.go
```diff
@@ -17,5 +17,16 @@ func (o *Orchestrator) DeleteInstance(ctx context.Context, sandboxID string, pau
 		zap.Bool("pause", pause),
 	)
 
+	c, err := o.instanceCache.Get(sandboxID)
+	if err == nil {
```
jakubno: wouldn't it be better like this? you don't really care about the error
```suggestion
	c, _ := o.instanceCache.Get(sandboxID)
	if c != nil {```

File: packages/api/internal/orchestrator/create_instance.go [hunk 19]
```diff
@@ -163,7 +164,12 @@ func (o *Orchestrator) CreateSandbox(
 		}
 
 		if node == nil {
-			node, err = o.getLeastBusyNode(childCtx, nodesExcluded)
+			nodeClusterID := uuid.Nil
+			if team.Team.ClusterID != nil {
+				nodeClusterID = *team.Team.ClusterID
+			}
```
jakubno, line 170: does `team.Team.ClusterID` need to be a nullable. Can we have `uuid.Nil` as a default, so we don't have to set it everywhere like this?

packages/api/internal/orchestrator/node.go
```diff
@@ -236,8 +282,10 @@ func (n *Node) InsertBuild(buildID string) {
 	n.buildCache.Set(buildID, struct{}{}, 2*time.Minute)
 }
 
-func (o *Orchestrator) NodeCount() int {
-	return o.nodes.Count()
+func (n *Node) GetClient() (*grpclient.GRPCClient, RequestContextBuilder) {
```
jakubno: can we set the metadata per client to be injected on every request automatically?

jakubno (changes requested):

packages/api/internal/orchestrator/create_instance.go
sitole: You mean like somehow adjust query logic and add a default empty UUID there?
Yeah, we can do this, but it can be an issue for things such as template creation, where you are storing the cluster ID, so you will send an empty UUID that does not exist in the cluster's database. This way we will just move IF to a different place, right?

packages/api/internal/orchestrator/node.go
sitole: Looks like it cannot be done. It cannot be set on the gRPC client or gRPC connection directly. Basically, to put it in, we need to adjust ctx provided during method invoke.

packages/api/internal/edge/cluster.go
sitole: Use context for what exactly?

dobrac: the context from c.tracer.Start is ignored, so the span is not correctly passed to the callstack

sitole: It's ignored because the only use of context in this method is the tracer start itself. If I re-assign ctx linter will fail, because iam trying to assign to a variable that is no longer used.

Not sure if you mean this...

packages/api/internal/orchestrator/node.go
sitole: This looks like most clean way how to do it https://github.com/e2b-dev/infra/pull/831/commits/52ff52cf9060bb04ed6c5e21f4dbf85fdabd83db

packages/api/internal/orchestrator/cache.go
```diff
@@ -348,8 +348,8 @@ func (o *Orchestrator) getDeleteInstanceFunction(
 			info.PauseDone(nil)
 		} else {
 			req := &orchestrator.SandboxDeleteRequest{SandboxId: info.Instance.SandboxID}
-			client, reqCtxBuilder := node.GetClient()
-			_, err := client.Sandbox.Delete(reqCtxBuilder(ctx), req)
+			client, reqCtx := node.GetClient(ctx)
```
jakubno: Can you do this, just to make sure it will be used correctly?
```suggestion
			client, ctx := node.GetClient(ctx)
```

sitole: Yes, we can do it, and probably it's just fine. Thinking if we cannot somehow break something, because auth and node-id are part of metadata, so if you use this context for a different client, you will append another token and another node-id, ending in wrong routing, because the first one will be used.

jakubno: It won't override the parent context, right?

sitole: No, but if you use a new one and send it to different methods, then the values are still there.

sitole: Will change it, iam maybe overthinking it

PR #833, opened 2025-06-27
Fixes in client proxy service discovery

sitole:
- Put edge/orchestrator node info private and accessible only via getter/setter to fix a possible race condition
- Store edge/orchestrator nodes in a pool with machine ID, not service ID, This was causing issues when the service restarted, service ID changes, but we updated it just inside node inf,o butthe  node pool key was still the same, so during request, it failed because "node was not found"

### DIFF
--- a/packages/client-proxy/internal/edge-pass-through/proxy.go
+++ b/packages/client-proxy/internal/edge-pass-through/proxy.go
@@ -79,7 +79,7 @@ func (s *NodePassThroughServer) director(ctx context.Context) (*grpc.ClientConn,
 		return nil, status.Error(codes.NotFound, "node not found")
 	}
 
-	return node.Client.Connection, nil
+	return node.GetClient().Connection, nil
 }
 
 // Handler - following code implement a gRPC pass-through proxy that forwards requests to the appropriate node
--- a/packages/client-proxy/internal/edge/handlers/service-discovery-node-drain.go
+++ b/packages/client-proxy/internal/edge/handlers/service-discovery-node-drain.go
@@ -62,7 +62,7 @@ func (a *APIStore) sendNodeRequest(ctx context.Context, serviceId string, status
 			return errors.New("failed to transform node status to orchestrator status")
 		}
 
-		_, err := o.Client.Info.ServiceStatusOverride(
+		_, err := o.GetClient().Info.ServiceStatusOverride(
 			findCtx, &orchestratorinfo.ServiceStatusChangeRequest{ServiceStatus: orchestratorStatus},
 		)
 		if err != nil {
--- a/packages/client-proxy/internal/edge/handlers/service-discovery-node-drain.go
+++ b/packages/client-proxy/internal/edge/handlers/service-discovery-node-drain.go
@@ -82,9 +82,9 @@ func (a *APIStore) sendNodeRequest(ctx context.Context, serviceId string, status
 
 	switch status {
 	case api.Draining:
-		_, err = e.Client.V1ServiceDiscoveryNodeDrain(ctx, serviceId)
+		_, err = e.GetClient().V1ServiceDiscoveryNodeDrain(ctx, serviceId)
 	case api.Unhealthy:
-		_, err = e.Client.V1ServiceDiscoveryNodeKill(ctx, serviceId)
+		_, err = e.GetClient().V1ServiceDiscoveryNodeKill(ctx, serviceId)
 	default:
 		return errors.New("failed to transform node status to api call")
 	}
--- a/packages/client-proxy/internal/edge/pool/edge-pool.go
+++ b/packages/client-proxy/internal/edge/pool/edge-pool.go
@@ -53,12 +53,13 @@ func (p *EdgePool) GetNodes() map[string]*EdgeNode {
 }
 
 func (p *EdgePool) GetNode(id string) (*EdgeNode, error) {
-	o, ok := p.nodes.Get(id)
-	if !ok {
-		return nil, ErrEdgeNodeNotFound
+	for _, node := range p.nodes.Items() {
+		if node.GetInfo().ServiceId == id {
+			return node, nil
+		}
 	}
 
-	return o, nil
+	return nil, ErrEdgeNodeNotFound
 }
 
 func (p *EdgePool) keepInSync(ctx context.Context) {
@@ -101,7 +102,6 @@ func (p *EdgePool) syncNodes(ctx context.Context) {
 			defer wg.Done()
 
 			var found *EdgeNode = nil
-
 			host := fmt.Sprintf("%s:%d", sdNode.NodeIp, sdNode.NodePort)
 
 			// skip self registration
@@ -172,7 +172,7 @@ func (p *EdgePool) connectNode(ctx context.Context, node *sd.ServiceDiscoveryIte
 		return err
 	}
 
-	p.nodes.Insert(o.GetInfo().ServiceId, o)
+	p.nodes.Insert(o.GetInfo().NodeId, o)
 	return nil
 }
 
@@ -189,7 +189,7 @@ func (p *EdgePool) removeNode(ctx context.Context, node *EdgeNode) error {
 		p.logger.Error("Error closing connection to node", zap.Error(err), l.WithClusterNodeID(info.ServiceId))
 	}
 
-	p.nodes.Remove(info.ServiceId)
+	p.nodes.Remove(info.NodeId)
 	p.logger.Info("Edge node node connection has been closed.", l.WithClusterNodeID(info.ServiceId))
 	return nil
 }
--- a/packages/client-proxy/internal/edge/pool/edge-pool.go
+++ b/packages/client-proxy/internal/edge/pool/edge-pool.go
@@ -20,7 +20,6 @@ type EdgePool struct {
 
 	nodeSelfHost string
 	nodes        *smap.Map[*EdgeNode]
-	mutex        sync.RWMutex
 
 	tracer trace.Tracer
 	logger *zap.Logger
@@ -38,7 +37,6 @@ func NewEdgePool(ctx context.Context, logger *zap.Logger, discovery sd.ServiceDi
 
 		nodeSelfHost: nodeSelfHost,
 		nodes:        smap.New[*EdgeNode](),
-		mutex:        sync.RWMutex{},
 
 		logger: logger,
 		tracer: tracer,
@@ -51,16 +49,10 @@ func NewEdgePool(ctx context.Context, logger *zap.Logger, discovery sd.ServiceDi
 }
 
 func (p *EdgePool) GetNodes() map[string]*EdgeNode {
-	p.mutex.RLock()
-	defer p.mutex.RUnlock()
-
 	return p.nodes.Items()
 }
 
 func (p *EdgePool) GetNode(id string) (*EdgeNode, error) {
-	p.mutex.RLock()
-	defer p.mutex.RUnlock()
-
 	o, ok := p.nodes.Get(id)
 	if !ok {
 		return nil, ErrEdgeNodeNotFound
@@ -118,7 +110,7 @@ func (p *EdgePool) syncNodes(ctx context.Context) {
 			}
 
 			for _, node := range p.nodes.Items() {
-				if host == node.Host {
+				if host == node.GetInfo().Host {
 					found = node
 					break
 				}
@@ -149,7 +141,7 @@ func (p *EdgePool) syncNodes(ctx context.Context) {
 
 			for _, sdNode := range sdNodes {
 				host := fmt.Sprintf("%s:%d", sdNode.NodeIp, sdNode.NodePort)
-				if host == node.Host {
+				if host == node.GetInfo().Host {
 					found = true
 					break
 				}
@@ -180,11 +172,7 @@ func (p *EdgePool) connectNode(ctx context.Context, node *sd.ServiceDiscoveryIte
 		return err
 	}
 
-	p.mutex.Lock()
-	defer p.mutex.Unlock()
-
-	p.nodes.Insert(o.ServiceId, o)
-
+	p.nodes.Insert(o.GetInfo().ServiceId, o)
 	return nil
 }
 
@@ -192,19 +180,16 @@ func (p *EdgePool) removeNode(ctx context.Context, node *EdgeNode) error {
 	_, childSpan := p.tracer.Start(ctx, "remove-edge-node")
 	defer childSpan.End()
 
-	p.logger.Info("Edge node connection is not active anymore, closing.", l.WithClusterNodeID(node.ServiceId))
+	info := node.GetInfo()
+	p.logger.Info("Edge node connection is not active anymore, closing.", l.WithClusterNodeID(info.ServiceId))
 
 	// stop background sync and close everything
 	err := node.Close()
 	if err != nil {
-		p.logger.Error("Error closing connection to node", zap.Error(err), l.WithClusterNodeID(node.ServiceId))
+		p.logger.Error("Error closing connection to node", zap.Error(err), l.WithClusterNodeID(info.ServiceId))
 	}
 
-	p.mutex.Lock()
-	defer p.mutex.Unlock()
-
-	p.nodes.Remove(node.ServiceId)
-
-	p.logger.Info("Edge node node connection has been closed.", l.WithClusterNodeID(node.ServiceId))
+	p.nodes.Remove(info.ServiceId)
+	p.logger.Info("Edge node node connection has been closed.", l.WithClusterNodeID(info.ServiceId))
 	return nil
 }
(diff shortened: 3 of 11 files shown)


packages/client-proxy/internal/edge/pool/edge-pool.go
```diff
@@ -51,22 +49,17 @@ func NewEdgePool(ctx context.Context, logger *zap.Logger, discovery sd.ServiceDi
...
 
 func (p *EdgePool) GetNode(id string) (*EdgeNode, error) {
-	p.mutex.RLock()
-	defer p.mutex.RUnlock()
-
-	o, ok := p.nodes.Get(id)
-	if !ok {
-		return nil, ErrEdgeNodeNotFound
+	for _, node := range p.nodes.Items() {
+		if node.GetInfo().ServiceId == id {
```
dobrac: should this be MachineID? Or is now serviceId MachineID? Can you clarify the naming please?

packages/client-proxy/internal/edge-pass-through/proxy.go
```diff
@@ -79,7 +79,7 @@ func (s *NodePassThroughServer) director(ctx context.Context) (*grpc.ClientConn,
 		return nil, status.Error(codes.NotFound, "node not found")
 	}
 
-	return node.Client.Connection, nil
+	return node.GetClient().Connection, nil
```
dobrac: Why the GetClient()? I don't see any lock there

packages/client-proxy/internal/edge/pool/edge-pool.go
```diff
@@ -180,31 +172,24 @@ func (p *EdgePool) connectNode(ctx context.Context, node *sd.ServiceDiscoveryIte
...
 	return nil
 }
 
 func (p *EdgePool) removeNode(ctx context.Context, node *EdgeNode) error {
 	_, childSpan := p.tracer.Start(ctx, "remove-edge-node")
 	defer childSpan.End()
 
-	p.logger.Info("Edge node connection is not active anymore, closing.", l.WithClusterNodeID(node.ServiceId))
+	info := node.GetInfo()
+	p.logger.Info("Edge node connection is not active anymore, closing.", l.WithClusterNodeID(info.ServiceId))
```
dobrac: please clarify the vars naming:
`l.WithClusterNodeID(info.ServiceId)`

packages/client-proxy/internal/edge/pool/edge-pool.go
```diff
@@ -51,22 +49,17 @@ func NewEdgePool(ctx context.Context, logger *zap.Logger, discovery sd.ServiceDi
...
 }
 
 func (p *EdgePool) GetNode(id string) (*EdgeNode, error) {
-	p.mutex.RLock()
-	defer p.mutex.RUnlock()
-
-	o, ok := p.nodes.Get(id)
-	if !ok {
-		return nil, ErrEdgeNodeNotFound
+	for _, node := range p.nodes.Items() {
```
dobrac: Also, why the loop and not nodes.Get()?

dobrac (changes requested):

dobrac:
> Store edge/orchestrator nodes in a pool with machine ID, not service ID, This was causing issues when the service restarted, service ID changes, but we updated it just inside node inf,o butthe node pool key was still the same, so during request, it failed because "node was not found"

This change hardens reliability to 1 node = 1 orch, right?

packages/client-proxy/internal/edge/pool/edge-pool.go
sitole: NodeID = ID that persists for whole machine lifetime (alias to MachineID or HostID)
ServiceID = ID of running process on machine. When process is restarted new ID is used

packages/client-proxy/internal/edge-pass-through/proxy.go
sitole: There is not, but like it this way more

packages/client-proxy/internal/edge/pool/edge-pool.go
sitole: Because map is using NodeID not ServiceID and we are providing ServiceID.

packages/client-proxy/internal/edge/pool/orchestrator.go
```diff
@@ -16,125 +16,90 @@ import (
...
+		freshInfo.ServiceInstanceID = status.ServiceId
+		freshInfo.ServiceStartup = status.ServiceStartup.AsTime()
+		freshInfo.ServiceStatus = getMappedStatus(status.ServiceStatus)
+		freshInfo.ServiceVersion = status.ServiceVersion
+		freshInfo.ServiceVersionCommit = status.ServiceCommit
+		freshInfo.Roles = status.ServiceRoles
 
-		o.Roles = status.ServiceRoles
+		freshInfo.ServiceVersion = status.ServiceVersion
+		freshInfo.ServiceVersionCommit = status.ServiceCommit
```
djeebus: this looks duplicated from lines 96-97

packages/client-proxy/internal/edge/pool/orchestrator-pool.go
```diff
@@ -13,214 +13,195 @@ import (
...
-		mutex: sync.RWMutex{},
+		instances: smap.New[*OrchestratorInstance](),
 
 		tracer: tracerProvider.Tracer("orchestrators-pool"),
 		logger: logger,
 
 		metricProvider: metricProvider,
 		tracerProvider: tracerProvider,
+
+		close: make(chan struct{}),
```
djeebus, line 59: what writes to this channel?

packages/client-proxy/internal/edge/pool/orchestrator.go
sitole: fixed, thanks!

packages/client-proxy/internal/edge/pool/orchestrator-pool.go
sitole: We are using it just for blocking wait and then closing it. Used in `statusLogSync` so it's closed when the orchestrator pool is closed.

sitole:
After internal discussion, we decided to remove node/service ids from the drain/kill endpoint path and move it to the request body, also with an additional service type. This way we make it more dynamic (body can add new features in the future) and also we make it a little bit safer because you also need to explicitly mention service type (edge/orchestrator), so you cannot by mistake nuke the wrong instance by mistake.

djeebus:
LGTM! 🚀

packages/client-proxy/internal/edge/handlers/service-discovery-node-drain.go
```diff
@@ -11,14 +12,25 @@ import (
...
+	l "github.com/e2b-dev/infra/packages/shared/pkg/logger"
+	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
 )
 
-func (a *APIStore) V1ServiceDiscoveryNodeDrain(c *gin.Context, serviceID string) {
-	_, templateSpan := a.tracer.Start(c, "service-discovery-node-drain-handler")
+func (a *APIStore) V1ServiceDiscoveryNodeDrain(c *gin.Context) {
+	ctx := c.Request.Context()
+
+	_, templateSpan := a.tracer.Start(ctx, "service-discovery-node-drain-handler")
```
dobrac: get the context, so the traces are properly passed down the stack

packages/client-proxy/internal/edge/handlers/service-discovery-node-kill.go
```diff
@@ -2,20 +2,31 @@ package handlers
...
 	"github.com/e2b-dev/infra/packages/shared/pkg/http/edge"
+	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
 )
 
-func (a *APIStore) V1ServiceDiscoveryNodeKill(c *gin.Context, serviceID string) {
-	_, templateSpan := a.tracer.Start(c, "service-discovery-node-kill-handler")
+func (a *APIStore) V1ServiceDiscoveryNodeKill(c *gin.Context) {
+	ctx := c.Request.Context()
+
+	_, templateSpan := a.tracer.Start(ctx, "service-discovery-node-kill-handler")
```
dobrac: get the context for traces too

packages/client-proxy/internal/edge/pool/edge-pool.go
```diff
@@ -4,205 +4,158 @@ import (
...
+	return false
+}
 
-			// skip self registration
-			if host == p.nodeSelfHost {
-				return
-			}
+func (e *edgeInstancesSyncStore) PoolList(ctx context.Context) []*EdgeInstance {
+	items := make([]*EdgeInstance, 0)
+	for _, item := range e.pool.instances.Items() {
```
dobrac, line 105: isn't items already a copy when using smap?

packages/client-proxy/internal/edge/pool/orchestrator-pool.go
```diff
@@ -13,214 +13,195 @@ import (
...
+}
 
-	// disconnect nodes that are not in the list anymore
-	for _, node := range p.GetOrchestrators() {
-		wg.Add(1)
-		go func(node *OrchestratorNode) {
-			defer wg.Done()
+func (e *orchestratorInstancesSyncStore) PoolList(ctx context.Context) []*OrchestratorInstance {
+	items := make([]*OrchestratorInstance, 0)
+	for _, item := range e.pool.instances.Items() {
```
dobrac, line 153: isn't .Items already a copy?

packages/client-proxy/internal/edge/service.go
```diff
@@ -10,19 +10,9 @@ import (
...
-	ctx context.Context,
-	logger *zap.Logger,
-	tracer trace.Tracer,
-	info *e2binfo.ServiceInfo,
-	edgeSD service_discovery.ServiceDiscoveryAdapter,
-	orchestrators *e2borchestrators.OrchestratorsPool,
-	catalog sandboxes.SandboxesCatalog,
-) (*handlers.APIStore, error) {
-	edges := e2borchestrators.NewEdgePool(ctx, logger, edgeSD, tracer, info.Host)
+func NewEdgeAPIStore(ctx context.Context, logger *zap.Logger, tracer trace.Tracer, info *e2binfo.ServiceInfo, edges *e2borchestrators.EdgePool, orchestrators *e2borchestrators.OrchestratorsPool, catalog sandboxes.SandboxesCatalog) (*handlers.APIStore, error) {
```
dobrac: can you revert the formatting of args to be on separate lines?

packages/client-proxy/internal/edge/pool/edge-pool.go
```diff
@@ -4,205 +4,158 @@ import (
...
+}
 
-	// Service discovery targets
-	sdNodes, err := p.discovery.ListNodes(spanCtx)
-	if err != nil {
-		return
+func (e *edgeInstancesSyncStore) SourceExists(ctx context.Context, s []sd.ServiceDiscoveryItem, p *EdgeInstance) bool {
+	for _, item := range s {
+		itemHost := e.getHost(item.NodeIP, item.NodePort)
+		if itemHost == p.GetInfo().Host {
```
dobrac, line 95: get the info only once and reuse

packages/client-proxy/internal/edge/handlers/service-discovery-node-drain.go
```diff
@@ -37,17 +49,20 @@ func (a *APIStore) V1ServiceDiscoveryNodeDrain(c *gin.Context, serviceID string)
...
 	defer findCtxCancel()
 
-	logger := a.logger.With(zap.String("service_id", serviceID))
+	logger := a.logger.With(l.WithServiceInstanceID(serviceInstanceID))
 
-	// try to find orchestrator node first
-	o, ok := a.orchestratorPool.GetOrchestrator(serviceID)
-	if ok {
-		logger.Info("found orchestrator node, calling status change request")
+	if serviceType == api.ClusterNodeTypeOrchestrator {
```
dobrac: can you split this function into two, one for the Orchestrator, one for the edge?

Basically create a mapping:
`serviceType -> handler`

dobrac (changes requested):
few small changes

packages/client-proxy/internal/edge/handlers/service-discovery-node-drain.go
sitole: fixed, thx

packages/client-proxy/internal/edge/pool/edge-pool.go
sitole: It's just because the interface expects to return an array, not map

packages/client-proxy/internal/edge/pool/edge-pool.go
sitole: You mean get it once for the whole sync struct?
It can change in time, so it is not a good idea to store it during sync struct creation.

packages/client-proxy/internal/edge/pool/edge-pool.go
sitole: Golang does not have a generic iterator type, so you need to always specify if it's map/array.

You can theoretically create this type, but it's really not flexible, and usage is not pretty.
```golang
type Iterable[T any] interface {
	[]T | map[any]T
}
```

In golang you need to do this
```golang
type Store[SourceItem any, SourceIterator Iterable[SourceItem], PoolItem any, PoolIterator Iterable[PoolItem]] interface {
SourceExists(ctx context.Context, s SourceIterator, p PoolItem) bool
}
```

Instead of more prretty version
```golang
type Store[SourceItem any, PoolItem any] interface {
SourceExists(ctx context.Context, s Iterable[SourceItem], p PoolItem) bool
}

```

packages/client-proxy/internal/edge/handlers/service-discovery-node-drain.go
```diff
@@ -50,44 +50,57 @@ func (a *APIStore) V1ServiceDiscoveryNodeDrain(c *gin.Context) {
...
+	// try to find orchestrator node first
+	o, ok := a.orchestratorPool.GetOrchestrator(serviceInstanceID)
+	if !ok {
+		return errors.New("orchestrator instance doesn't found")
+	}
+
+	logger.Info("orchestrator instance found, calling status change request")
+	var orchestratorStatus orchestratorinfo.ServiceInfoStatus
+
+	switch status {
```
dobrac: use the `ApiNodeToOrchestratorStateMapper`

dobrac (changes requested):
one small nit a gtg

packages/client-proxy/internal/edge/pool/edge-pool.go
dobrac: Oh I didn't realize the .Items() is a map, all good then

packages/client-proxy/internal/edge/pool/edge-pool.go
dobrac: Just move it above the range, because the p.GetInfo won't/shouldn't change while doing the iteration

packages/client-proxy/internal/edge/pool/edge.go
```diff
@@ -27,100 +29,97 @@ type EdgeNode struct {
...
+	defer o.mutex.Unlock()
+	o.info.ServiceStatus = s
 }
 
-func newEdgeApiClient(host string) (*api.ClientWithResponses, error) {
-	hostUrl := fmt.Sprintf("http://%s", host)
-	return api.NewClientWithResponses(hostUrl)
+func newEdgeApiClient(host string, auth authorization.AuthorizationService) (*api.ClientWithResponses, error) {
+	clientURL := fmt.Sprintf("http://%s", host)
+	clientSecret := auth.GetSecret()
```
dobrac: might make sense to request the secret for every request instead of static caching, not fully necessary right now as it's static afaik

packages/client-proxy/internal/edge/handlers/service-discovery-node-drain.go
sitole: Cannot use `ApiNodeToOrchestratorStateMapper`, because it's used for mapping different source types (api node status) and its located in API package, so even if we want to extract it to shared, we need to move the API openapi generated files also to share (we propably should at some point).

I changed it and now iam using the same pattern as in the API.

packages/client-proxy/internal/edge/pool/edge.go
sitole: Yes, it makes sense. This is partially why I abstracted authorization into a separate service, and it's not just a string (because it can be dynamic at some point).

Let's redo it with the implementation of OAuth for Edge API so we can do role-based authorization (for logs, control plane...)

packages/client-proxy/internal/edge/pool/edge-pool.go
sitole: Thx, missed that it's loop

PR #834, opened 2025-06-27
Update metadata during sync

jakubno:
# Description

If the API discovers the orchestrator before the orchestrator is healthy, it will add it with version, commit and orchestratorID equal to `unknown` and it won't be updated.

### DIFF
--- a/packages/api/internal/orchestrator/metrics.go
+++ b/packages/api/internal/orchestrator/metrics.go
@@ -22,7 +22,7 @@ func (o *Orchestrator) setupMetrics(meterProvider metric.MeterProvider) (metric.
 			for _, node := range o.nodes.Items() {
 				obs.ObserveInt64(gauge, 1, metric.WithAttributes(
 					attribute.String("status", string(node.status)),
-					attribute.String("node.id", node.orchestratorID),
+					attribute.String("node.id", node.metadata.orchestratorID),
 				))
 			}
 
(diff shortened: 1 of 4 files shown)


packages/api/internal/orchestrator/node.go
```diff
@@ -219,3 +223,19 @@ func (n *Node) InsertBuild(buildID string) {
 func (o *Orchestrator) NodeCount() int {
 	return o.nodes.Count()
 }
+
+func getNodeMetadata(n *orchestratorinfo.ServiceInfoResponse) *nodeMetadata {
```
sitole: Why pointer is used here?

sitole:
After discussion with @dobrac we agreend on adding removing metadata pointer and adding safe lock for concurrent R/W access to metadata and metadata structure fields.

packages/api/internal/orchestrator/metrics.go
```diff
@@ -22,7 +22,7 @@ func (o *Orchestrator) setupMetrics(meterProvider metric.MeterProvider) (metric.
 			for _, node := range o.nodes.Items() {
 				obs.ObserveInt64(gauge, 1, metric.WithAttributes(
 					attribute.String("status", string(node.status)),
```
dobrac: does the status not require read with lock?

packages/api/internal/orchestrator/node.go
```diff
@@ -220,3 +238,19 @@ func (n *Node) InsertBuild(buildID string) {
 func (o *Orchestrator) NodeCount() int {
 	return o.nodes.Count()
 }
+
+func getNodeMetadata(n *orchestratorinfo.ServiceInfoResponse) nodeMetadata {
+	if n == nil {
+		return nodeMetadata{
+			orchestratorID: "",
```
dobrac: before the orchestratorID was set as `node.ID` even without successful response

sitole: Yea, i know why Kuba removed it. Handling is not pretty with it

PR #835, opened 2025-06-27
Rework query metrics

jakubno:
# Description

Query the metrics from clickhouse.

- For list view, return only the latest metrics and fetch them all in one query
- For single sandbox we will return list of latest X ticks (we will probably aggregate on backend side - e.g. 5 minute average, if the metrics should be for last 24 hours). It will be done in separate PR
- Response times are pretty stable regardless of the number of concurrently running sandboxes (~50ms with multiple shards)


spec/openapi.yml
```diff
@@ -991,32 +991,6 @@ paths:
         "500":
           $ref: "#/components/responses/500"
 
-  /sandboxes/{sandboxID}/metrics:
```
jakubno, line 997: removed sandbox detail endpoint as we want to have different logic there and don't make this PR larger than it has to be

spec/openapi.yml
```diff
@@ -201,8 +201,8 @@ components:
         - timestamp
         - cpuCount
         - cpuUsedPct
-        - memUsedMiB
-        - memTotalMiB
+        - memUsed
```
jakubno, line 204: @ben-fornefeld This needs to be adjusted in dashboard

packages/api/internal/handlers/sandboxes_list_metrics.go
```diff
@@ -2,178 +2,105 @@ package handlers
...
-			results <- metricsResult{
-				sandbox: s,
-				metrics: metrics,
-				err:     err,
-			}
-		}(sandbox)
+	flagCtx := ldcontext.NewBuilder(featureflags.MetricsReadFlagName).Build()
+	metricsReadFlag, err := a.featureFlags.Ld.BoolVariation(featureflags.MetricsReadFlagName, flagCtx, featureflags.MetricsReadDefault)
+	if err != nil {
+		zap.L().Error("soft failing during metrics write feature flag receive", zap.Error(err))
```
0div:
```suggestion
		zap.L().Error("error getting metrics read feature flag, soft failing", zap.Error(err))
```

packages/api/internal/handlers/sandboxes_list_metrics.go
```diff
@@ -2,178 +2,105 @@ package handlers
...
+		// This is here just to have the possibility to turn off ClickHouse metrics reading
+		return nil, nil
+	}
 
-	// Collect results and build final response
-	sandboxesWithMetrics := make([]api.RunningSandboxWithMetrics, 0, len(sandboxes))
+	metrics, err := a.clickhouseStore.QueryLatestMetrics(ctx, sandboxIds, teamID.String())
+	if err != nil {
+		zap.L().Error("Error fetching sandbox metrics from ClickHouse",
+			logger.WithTeamID(teamID.String()),
```
0div, line 61: if this logger does not include sandbox id, i would add it here to help debug further

packages/api/internal/handlers/sandboxes_list_metrics.go
```diff
@@ -2,178 +2,105 @@ package handlers
...
+	// Process each result as it arrives
+	for _, sbx := range sandboxes {
+		var sbxMetrics *api.SandboxMetric
+		m, ok := apiMetrics[sbx.SandboxID]
+		if ok {
+			sbxMetrics = &m
 		}
 
-		err := errors.Join(metricsErrors...)
+		sbxWithMetrics := api.RunningSandboxWithMetrics{
```
0div, line 90: yeah it's annoying to have to do this bc of strict typing, prob some unwanted memory allocation at scale, i haven't found a way to use the api struct with the chdb methods neither :/

packages/api/internal/handlers/store.go
```diff
@@ -86,17 +85,13 @@ func NewAPIStore(ctx context.Context, tel *telemetry.Client) *APIStore {
...
 
-	if readMetricsFromClickHouse == "true" {
-		clickhouseStore, err = chdb.NewStore(chdb.ClickHouseConfig{
-			ConnectionString: os.Getenv("CLICKHOUSE_CONNECTION_STRING"),
-			Username:         os.Getenv("CLICKHOUSE_USERNAME"),
-			Password:         os.Getenv("CLICKHOUSE_PASSWORD"),
-			Database:         os.Getenv("CLICKHOUSE_DATABASE"),
-			Debug:            os.Getenv("CLICKHOUSE_DEBUG") == "true",
-		})
+	clickhouseConnectionString := strings.TrimSpace(os.Getenv("CLICKHOUSE_CONNECTION_STRING"))
```
0div, line 90: do we want this to be independent of feature flags?

packages/clickhouse/pkg/metrics.go
```diff
@@ -0,0 +1,66 @@
...
+	MemTotal       float64   `ch:"ram_total"`
+	MemUsed        float64   `ch:"ram_used"`
+}
+
+const metricsSelectQuery = `
+SELECT
+    Attributes['sandbox_id']                                            AS sandbox_id,
+    Attributes['team_id']                                               AS team_id,
+
+    argMaxIf(Value,  TimeUnix, MetricName = 'e2b.sandbox.cpu.total')    AS cpu_total,
```
0div, line 24: We're not inserting these values in the same row? Im which case i would make the query simpler by using only one most recent timestamp for the entire row for readability and maybe also performance.

0div (changes requested):

packages/api/internal/handlers/sandboxes_list_metrics.go
jakubno: There are multiple sandbox ids

packages/api/internal/handlers/sandboxes_list_metrics.go
jakubno: not optimal, but it shouldn't be an issue now

packages/api/internal/handlers/store.go
jakubno: It kinda has to be, because this is done only once during initialization. Also there's probably no reason why it should be feature flag dependant

jakubno: Worst case (not expecting this) it fails during initialization and the new API version won't be deployed

packages/clickhouse/pkg/metrics.go
jakubno: No, they are different metrics

0div: Different metrics tables? If so, I would put them all in the same table, i know we're not doing joins here per se but it kinda has the feel of it and CH is not meant for that.

jakubno: They are the same table, but different rows. Here's the table definiton: https://github.com/e2b-dev/infra/blob/main/packages/clickhouse/migrations/20250521131545_add_metrics_local.sql

0div: i see sorry i missed reviewing that part earlier. i'd say as a columnar storage, you'll likely get better compression, query performance and esp. aggregations speed if each metric type has their own column. if we need to add new metric type later it's relatively fast to add a new column in ClickHouse and doesn't require rewriting existing data, new columns get default values for existing rows.

0div: it this a mapping to prometheus or something?

jakubno: This can't be really done the way how we collect metrics (we are using OTEL). But it's completely fine  as we order by the column. Also we can add Index
```
    ORDER BY (ServiceName, MetricName, Attributes, toUnixTimestamp64Nano(TimeUnix))
```

spec/openapi.yml
dobrac: plus how about the CLI?

packages/clickhouse/pkg/metrics.go
0div: Typically `ORDER BY` clause can impact performance in SQL. 
You can index it that way if we're already ordering by timestamp (i hope) ?

jakubno: This is on the table and it defines how the data are stored and it improves the speed how the data are collected and also compressed

0div: Cool, looks good

0div (approved):

spec/openapi.yml
jakubno: Also CLI, but that isn't released

PR #748, opened 2025-06-09
Unite log fields names

dobrac:
Unite log fields names, e.g. `sandboxID` vs `sandbox_id`. All fields are now snake case.

PR #839, opened 2025-07-01
Fix sandbox logs ingestion from services

dobrac:
Fix sandbox logs ingestion from services (not envd).

The issue is that after the change https://github.com/e2b-dev/infra/pull/748, field names contain `.` instead of `_`. Dot symbol is converted in OTEL collector automatically to underscore symbol, but the same doesn't happen in Vector. This PR fixes that.

This PR also adds a commented sink to help debug potential future issues.

### DIFF
--- a/packages/nomad/logs-collector.hcl
+++ b/packages/nomad/logs-collector.hcl
@@ -153,7 +153,7 @@ source = '''
 del(.internal)
 '''
 
-# Enable debuging of logs to the console
+# Enable debugging of logs to the console
 # [sinks.console_loki]
 # type = "console"
 # inputs = ["remove_internal"]


packages/nomad/logs-collector.hcl
```diff
@@ -133,6 +153,22 @@ source = '''
 del(.internal)
 '''
 
+# Enable internal vector logs
+[sources.vector_logs]
+type = "internal_logs"
+
+# Print internal vector logs to console
+[sinks.console]
```
0div: not sure we should be doing this given the current OOM issues

dobrac: the OOM issues are on Loki, not Vector

ValentaTomas (approved):

PR #841, opened 2025-07-02
Handle Unexpected EOF for ReadAt in AWS S3 storage provider

sitole:
`io.ReadFull` will in that case return `ioErrUnexpectedEOF`, but service is expecting to receive `io.EOF` to know we that file download is complited.

### DIFF
--- a/packages/shared/pkg/storage/storage_aws.go
+++ b/packages/shared/pkg/storage/storage_aws.go
@@ -169,7 +169,7 @@ func (a *AWSBucketStorageObjectProvider) ReadAt(buff []byte, off int64) (n int, 
 
 	defer resp.Body.Close()
 
-	// When the object is smaller thant requested range there will be unexpected EOF,
+	// When the object is smaller than requested range there will be unexpected EOF,
 	// but backend expects to return EOF in this case.
 	n, err = io.ReadFull(resp.Body, buff)
 	if errors.Is(err, io.ErrUnexpectedEOF) {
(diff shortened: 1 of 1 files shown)


packages/shared/pkg/storage/storage_aws.go
```diff
@@ -169,7 +169,14 @@ func (a *AWSBucketStorageObjectProvider) ReadAt(buff []byte, off int64) (n int,
 
 	defer resp.Body.Close()
 
-	return io.ReadFull(resp.Body, buff)
+	// When the object is smaller thant requested range there will be unexpected EOF,
```
dobrac:
```suggestion
	// When the object is smaller than requested range there will be unexpected EOF,
```

dobrac (approved):
One nit. We should in the future return EOF correctly also for the GCP ObjectProvider.

sitole:
> > One nit. We should in the future return EOF correctly also for the GCP ObjectProvider.
> 
> In addition, I would check if we can use the `io.ReadFull` in the GCP too or if the `reader.Remain` was necessary. For both, we also need to check if the `io.EOF` handling in the `chunk.go` that is calling this `ReadAt` is correctly handling this error (disregarding if it happens in the last chunk, reporting if in another chunk).

Will create an issue so that we are not blocking this fix.
https://linear.app/e2b/issue/E2B-2617/slices-impl-should-properly-handle-ioeof-returned-from-object-storage

PR #842, opened 2025-07-03
Prevent nbd allocation log spam on error

dobrac:
Prevent nbd allocation log spam on erroring nbd allocation. Added delay of 50 ms between retries + logging only every 100th fail. In case of permanent failure, this will log at max 1 message per 5 seconds.

### DIFF
--- a/packages/orchestrator/internal/sandbox/nbd/pool.go
+++ b/packages/orchestrator/internal/sandbox/nbd/pool.go
@@ -130,11 +130,11 @@ func (d *DevicePool) Populate() error {
 		default:
 			device, err := d.getFreeDeviceSlot()
 			if err != nil {
-				failedCount++
 				if failedCount%100 == 0 {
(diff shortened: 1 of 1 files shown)


packages/orchestrator/internal/sandbox/nbd/pool.go
```diff
@@ -126,10 +130,15 @@ func (d *DevicePool) Populate() error {
 		default:
 			device, err := d.getFreeDeviceSlot()
 			if err != nil {
-				zap.L().Error("[nbd pool]: failed to create network", zap.Error(err))
+				failedCount++
```
ValentaTomas: I'm thinking if moving this after modulo might make more sense as it would log the error if it was "alone" error and not wait until there was 100 errors in a row.

dobrac: yes, totaly

dobrac: updated

PR #844, opened 2025-07-03
Open all ports on localhost interface to public by default

dobrac:
This PR reintroduces opening all ports listening on the loopback interface to public by default. 

It also fixes:
- socket closing when the port is closed
- routes also IPv6 listeners from the IPv4 interface

This is a regression from https://github.com/e2b-dev/infra/pull/681 and most of the implementation is copied from https://github.com/e2b-dev/infra/blob/703da3b2b8ef4af450f9874228e7406bdfc75d4a/packages/envd/internal/env/env.go (except replacing the GOnetstat library with already used gopsutil).

### DIFF
--- a/packages/envd/main.go
+++ b/packages/envd/main.go
@@ -38,7 +38,7 @@ const (
 )
 
 var (
-	Version = "0.2.1"
+	Version = "0.2.2"
 
 	commitSHA string
 
--- a/tests/integration/internal/tests/envd/localhost_bind_test.go
+++ b/tests/integration/internal/tests/envd/localhost_bind_test.go
@@ -1,0 +1,92 @@
+package envd
+
+import (
+	"context"
+	"errors"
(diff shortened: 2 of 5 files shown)


0div:
> We don't have currently any setup for IPv6 access to the sandbox from the outside, thats why the ipv6 <- ipv4. I'm not sure how the bindv6only plays role here, but if you do now `python -m http.server --bind localhost`, it won't be accessible through the IPv4 interface

To summarize convo about this: It's not expected behavior to have `localhost` accessible from the outside, wether inside a VPS or a container such as docker, it's expected that binding to `localhost` stays private whereas wanting to make servers accessible from outside we bind to `0.0.0.0`. There are also some security implications as well. However for the sake of reverting to what worked and making some bigger picture network changes later, it was decided to allow this for now.

PR #846, opened 2025-07-04
fix repositoryNameWithTemplate

warren830:
We should use both repositoryName and templateId to locate registry in aws ecr

### DIFF
--- a/packages/shared/pkg/artifacts-registry/registry_aws.go
+++ b/packages/shared/pkg/artifacts-registry/registry_aws.go
@@ -68,7 +68,8 @@ func (g *AWSArtifactsRegistry) Delete(ctx context.Context, templateId string, bu
 }
 
 func (g *AWSArtifactsRegistry) GetTag(ctx context.Context, templateId string, buildId string) (string, error) {
-	res, err := g.client.DescribeRepositories(ctx, &ecr.DescribeRepositoriesInput{RepositoryNames: []string{g.repositoryName}})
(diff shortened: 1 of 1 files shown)


warren830:
> Hi @warren830, do you have any exact reason to introduce the following change?
> 
> We are using this exact reason in Google Cloud, because you can dynamically create a repository by pushing to it, but this does not work the same in AWS, so for every template we would need to create ECR repository first.

Hi, sorry for the late reply. From my understanding,we need to define repository in env. That will be fixed. Everytime when we need a new template, it will generate a new image with buildId only. With this change, we can have templateId to identify different repos

sitole:
> > Hi @warren830, do you have any exact reason to introduce the following change?
> > We are using this exact reason in Google Cloud, because you can dynamically create a repository by pushing to it, but this does not work the same in AWS, so for every template we would need to create ECR repository first.
> 
> Hi, sorry for the late reply. From my understanding,we need to define repository in env. That will be fixed. Everytime when we need a new template, it will generate a new image with buildId only. With this change, we can have templateId to identify different repos

For single-tenant/single-template deployment, it's true, but we support multi-tenant environments. When your change is approved, it means that for every new template, we need to call the AWS ECR API and create a new repository, right? So it's not suitable for us.

In the Google Cloud implementation of the repository provider template ID is used because GCP Docker repository providers work differently. Images inside the repository are created dynamically when pushed. Feel free to prove me wrong, actually, we would love to implement it the same way as in GCP if that would work, because it offers bigger limits to us.

PR #847, opened 2025-07-04
Block changes from affecting the broken python sdk

djeebus:
TODO:

- [x] test the sdk against envd
- [x] convert the rest of the messages from new to old


djeebus:
This should be fully implemented now. One last complication: the current test requires adding a field and ensuring that it gets dropped when the right header is passed. This PR adds a boolean field called "testing" to the MoveResponse message, but I suspect we don't want to keep that. Do we have a field that we'd like to add that would also let us test this feature?

packages/envd/internal/services/legacy/conversion.go
```diff
@@ -0,0 +1,184 @@
...
+	return &EntryInfo{
+		Name: info.Name,
+		Type: FileType(info.Type),
+		Path: info.Path,
+	}
+}
+
+func Convert() ConversionInterceptor {
+	return ConversionInterceptor{
+		converters: map[reflect.Type]func(protoreflect.ProtoMessage) connect.AnyResponse{
```
jakubno: Can we return error instead of panic?
```suggestion
		converters: map[reflect.Type]func(protoreflect.ProtoMessage) (connect.AnyResponse, error){
```

packages/envd/internal/services/legacy/conversion.go
```diff
@@ -0,0 +1,184 @@
...
+
+		return response, nil
+	}
+}
+
+func (l ConversionInterceptor) WrapStreamingClient(clientFunc connect.StreamingClientFunc) connect.StreamingClientFunc {
+	return clientFunc
+}
+
+func (l ConversionInterceptor) WrapStreamingHandler(handlerFunc connect.StreamingHandlerFunc) connect.StreamingHandlerFunc {
```
jakubno: Will this work correctly for watch dir - it's a stream

djeebus: Nope, good call. Implemented streaming, but still need a bunch of tests

packages/envd/internal/services/legacy/stream.go
```diff
@@ -0,0 +1,42 @@
...
+func (s streamConverter) Receive(a any) error {
+	return s.conn.Receive(a)
+}
+
+func (s streamConverter) RequestHeader() http.Header {
+	return s.conn.RequestHeader()
+}
+
+func (s streamConverter) Send(a any) error {
+	a = maybeConvertValue(a)
```
jakubno, line 32: shouldn't this be done based on the header as in unary?

djeebus: I think it's different; at least the signature is different. We definitely need to test this though! I pushed up this implementation as a checkpoint, but I still have a lot of tests to write, and some manual work to verify.

packages/envd/internal/services/legacy/interceptor.go
```diff
@@ -0,0 +1,39 @@
...
+type ConversionInterceptor struct{}
+
+func (l ConversionInterceptor) WrapUnary(unaryFunc connect.UnaryFunc) connect.UnaryFunc {
+	return func(ctx context.Context, request connect.AnyRequest) (connect.AnyResponse, error) {
+		response, err := unaryFunc(ctx, request)
+		if err != nil {
+			return nil, err
+		}
+
+		if request.Header().Get("User-Agent") == "connect-python" {
```
jakubno: can we make it an constant? 🙏🏻

packages/envd/internal/services/legacy/stream.go
jakubno: I maybe missed it, but can you point me to where the condition for "if to convert or not to convert" is in streaming done?

djeebus:
I'm converting this back to a draft since it's become clear that streaming is potentially very different than the unary request/responses. Gonna need some real tests for this, both manual and unit. I'll bring it back when it's been verified to work.

packages/envd/internal/services/legacy/stream.go
djeebus: Just implemented it. it's in interceptor.go, in the `WrapStreamingHandler` function.

djeebus:
Fully implemented, tests pass, ready for review! Just waiting on end-to-end tests.

packages/envd/internal/services/legacy/interceptor.go
```diff
@@ -0,0 +1,58 @@
...
+			response = maybeConvertResponse(zerolog.Ctx(ctx), response)
+		}
+
+		return response, nil
+	}
+}
+
+func (l ConversionInterceptor) WrapStreamingClient(clientFunc connect.StreamingClientFunc) connect.StreamingClientFunc {
+	// only used for client, not server
+	panic("unused by server")
```
jakubno, line 47: Wouldn't it make sense to just return? e.g. If we decide to add a client streaming, we will have to change this line

packages/envd/internal/services/legacy/interceptor.go
```diff
@@ -0,0 +1,58 @@
...
+	"context"
+	"net/http"
+
+	"connectrpc.com/connect"
+	"github.com/rs/zerolog"
+)
+
+const brokenUserAgent = "connect-python"
+
+const notifyHeader = "X-E2B-Legacy-Sdk"
```
jakubno:
```suggestion
const notifyHeader = "X-E2B-Legacy-SDK"
```

packages/envd/internal/services/legacy/interceptor_test.go
```diff
@@ -0,0 +1,168 @@
...
+	return func(ctx context.Context, s connect.Spec) connect.StreamingClientConn {
+		conn := clientFunc(ctx, s)
+		conn.RequestHeader().Set("user-agent", u.userAgent)
+		return conn
+	}
+}
+
+func (u userAgentInterceptor) WrapStreamingHandler(handlerFunc connect.StreamingHandlerFunc) connect.StreamingHandlerFunc {
+	// TODO implement me
+	panic("implement me")
```
jakubno, line 161: same here, should be okay to just return  `clientFunc` right?

djeebus: This is just for testing, no reason to actually implement it. I can do that though, but I want to make it clear that it's literally unused. If the API changes in the future, I wanted to make it obvious that there's no value in spending effort in this function.

packages/envd/internal/services/legacy/interceptor.go
djeebus: This is a little weird. HTTP headers have a canonical formatting that it gets coerced to, at least in golang. Probably doesn't matter though. FWIW, I messed up the formatting. According to [golang](https://go.dev/play/p/LbTK2HCFRSm), it should be `E2b` =/

We can make this look good in code though, and let the canonical library do its job.

packages/envd/internal/services/legacy/interceptor.go
djeebus: We'd have to add this interceptor to a client, not the server, in order for this function to be called. Not sure why connectrpc built their interface this way, it's misleading. You can see the implementation of this function on the client side in interceptor_test.go, about line 151.

jakubno (approved):
lgtm

PR #854, opened 2025-07-08
WIP Run template build on the infrastructure II

dobrac:
WIP Run template build on the infrastructure II


packages/orchestrator/internal/template/build/layer_files.go
```diff
@@ -0,0 +1,7 @@
+package build
+
+import "fmt"
+
+func GetLayerFilesCachePath(templateID string, hash string) string {
+	return fmt.Sprintf("builder/%s/files/%s.tar", templateID, hash)
```
dobrac: I will create a new bucket for the layer files

packages/orchestrator/internal/template/build/hash_index.go
```diff
@@ -0,0 +1,91 @@
...
+
+	return nil
+}
+
+func uuidToHashPath(templateID, baseHash, hash string) string {
+	reSHA := sha256.New()
+	reSHA.Write([]byte(baseHash))
+	reSHA.Write([]byte(hash))
+	reHash := fmt.Sprintf("%x", reSHA.Sum(nil))
+	return fmt.Sprintf("builder/cache/%s/index/%s", templateID, reHash)
```
dobrac: the layers index is now saved to the templates bucket

The layers need to be without larger refactoring in the templates bucket. The reason is that the fetching for resuming layer needs to be from the same bucket as we resume sandboxes right now.

With next PR refactoring, we might be able to specify custom bucket for uploading and fetching the templates.

packages/orchestrator/internal/template/build/provision.go
```diff
@@ -0,0 +1,180 @@
...
+	rootfsFinalSize, err := ext4.Enlarge(ctx, b.tracer, rootfsPath, sizeDiff)
+	if err != nil {
+		// Debug filesystem stats on error
+		cmd := exec.Command("tune2fs", "-l", rootfsPath)
+		output, dErr := cmd.Output()
+		zap.L().Error(string(output), zap.Error(dErr))
+
+		return fmt.Errorf("error enlarging rootfs: %w", err)
+	}
+	template.RootfsSize = rootfsFinalSize
```
dobrac: this is not working now

packages/orchestrator/internal/template/build/rootfs/rootfs.go
```diff
@@ -105,7 +117,7 @@ func (r *Rootfs) createExt4Filesystem(ctx context.Context, tracer trace.Tracer,
 	if err != nil {
 		return containerregistry.Config{}, fmt.Errorf("error creating ext4 filesystem: %w", err)
 	}
-	r.template.rootfsSize = ext4Size
+	r.template.RootfsSize = ext4Size
```
dobrac: not working

packages/orchestrator/internal/template/build/rootfs/rootfs.go
```diff
@@ -134,7 +146,7 @@ func (r *Rootfs) createExt4Filesystem(ctx context.Context, tracer trace.Tracer,
 		if err != nil {
 			return containerregistry.Config{}, fmt.Errorf("error enlarging rootfs: %w", err)
 		}
-		r.template.rootfsSize = rootfsFinalSize
+		r.template.RootfsSize = rootfsFinalSize
```
dobrac: not working

packages/orchestrator/internal/template/build/ready_command.go
```diff
@@ -16,20 +18,21 @@ const (
...
 	sandboxID string,
-	envVars map[string]string,
+	cmdMetadata sandboxtools.CommandMetadata,
 ) error {
 	ctx, span := b.tracer.Start(ctx, "run-ready-command")
 	defer span.End()
 
 	postProcessor.WriteMsg("Waiting for template to be ready")
 
 	if template.ReadyCmd == "" {
```
dobrac: change to local var

packages/orchestrator/internal/template/build/builder.go
```diff
@@ -0,0 +1,590 @@
...
+	// Build Steps
+
+	// Run post-processing actions in the sandbox
+	err = b.buildLayer(
+		ctx,
+		postProcessor,
+		uploadErrGroup,
+		baseSandboxConfig,
+		metadata.TemplateID,
+		"config-run-cmd",
```
dobrac: these hashes are switched

packages/orchestrator/internal/template/build/builder.go
```diff
@@ -0,0 +1,590 @@
...
+			return nil
+		},
+	)
+	if err != nil {
+		return nil, fmt.Errorf("error running start and ready commands in sandbox: %w", err)
+	}
+
+	return &Result{
+		EnvdVersion:  envdVersion,
+		RootfsSizeMB: template.RootfsSizeMB(),
```
dobrac: this is now incorrect

packages/orchestrator/internal/template/build/oci/oci.go
```diff
@@ -37,7 +37,7 @@ func GetImage(ctx context.Context, tracer trace.Tracer, artifactRegistry artifac
 		Architecture: "amd64",
 	}
 
-	img, err := artifactRegistry.GetImage(childCtx, templateId, buildId, platform)
+	img, err := artifactRegistry.GetImage(childCtx, tag, platform)
```
sitole: How does this work with accessing public images? It's still called on GCP artifact registry implementation, and if so, can I theoretically provide `fromImage` that points to someone else's build images stored in our repository?

dobrac: Good point, I'll just revert the last commit

packages/orchestrator/internal/template/build/layer_files.go
dobrac: moved to a new bucket `fc-build-cache` containing the files as well as the index

packages/orchestrator/internal/template/build/hash_index.go
dobrac: index moved to a new bucket `fc-build-cache`

packages/orchestrator/internal/sandbox/template/cache.go
```diff
@@ -47,7 +49,11 @@ func NewCache(ctx context.Context) (*Cache, error) {
 		}
 	})
 
-	go cache.Start()
+	// Delete the old build cache directory content as it may contain stale data that are not managed by anyone.
+	err := cleanDir(build.DefaultCachePath)
+	if err != nil && !os.IsNotExist(err) {
+		return nil, fmt.Errorf("failed to remove old build cache directory: %w", err)
+	}
```
dobrac: Added deletion of the CachePath to remove any stale cached files after service restart

packages/api/internal/handlers/template_layer_files_upload.go
```diff
@@ -0,0 +1,53 @@
...
+
+		return
+	}
+
+	teamIDs := utils.Map(teams, func(t queries.GetTeamsWithUsersTeamsWithTierRow) uuid.UUID {
+		return t.Team.ID
+	})
+
+	// Check if the user has access to the template
+	_, err = a.db.Client.Env.Query().Where(env.ID(templateID), env.TeamIDIn(teamIDs...)).Only(ctx)
```
jakubno, line 35: can we use template cache here?

packages/orchestrator/internal/template/server/upload_layer_files_template.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	signedUrl, err := s.buildStorage.UploadSignedURL(ctx, path, signedUrlExpiration)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get signed url: %w", err)
+	}
+
+	_, err = obj.Size()
+	if err != nil {
+		zap.L().Warn("layer files not found", zap.Error(err))
```
jakubno: isn't this kinda expected?

packages/orchestrator/internal/template/cache/build_cache.go
```diff
@@ -84,12 +86,26 @@ func (b *BuildInfo) GetLogs() []string {
 }
 
 func (b *BuildInfo) Cancel() {
+	b.fail(cancelledBuildReason)
```
jakubno: log the error (if returned)

packages/orchestrator/internal/template/build/writer/postprocessor.go
```diff
@@ -26,23 +28,20 @@ func (p *PostProcessor) Start() {
...
 		select {
 		case postprocessingErr := <-p.errChan:
-			p.WriteMsg(msg)
-
 			if postprocessingErr != nil {
 				p.WriteMsg(fmt.Sprintf("Postprocessing failed: %s", postprocessingErr))
 				return
+			} else {
+				p.WriteMsg(fmt.Sprintf("Build finished, took %s", time.Since(startTime).Truncate(time.Second).String()))
+				return
```
jakubno: does this has to be in else branch?

packages/orchestrator/internal/template/build/writer/postprocessor.go
```diff
@@ -66,22 +65,29 @@ func (p *PostProcessor) Stop(ctx context.Context, err error) {
...
-	p.ticker.Reset(tickerInterval)
+	p.ticker.Reset(p.tickerInterval)
 	return p.writer.Write([]byte(prefixWithTimestamp(string(b))))
 }
 
-func NewPostProcessor(ctx context.Context, writer io.Writer) *PostProcessor {
+func NewPostProcessor(ctx context.Context, writer io.Writer, enableTicker bool) *PostProcessor {
+	// If ticker is not enabled, we use a ticker that ticks way past the build time
+	tickerInterval := 24 * time.Hour
+	if enableTicker {
```
jakubno, line 79: what's the purpose of this? (when it isn't enabled)

packages/orchestrator/internal/template/build/utils/utils.go
```diff
@@ -0,0 +1,32 @@
...
+
+func Sprintp(s *string) string {
+	if s == nil {
+		return "<nil>"
+	}
+
+	return *s
+}
+
+func GetFileHash(ctx context.Context, path string) (string, error) {
```
jakubno: what's the purpose of ctx?

jakubno (commented):
1st part for file hash endpoint

packages/orchestrator/internal/template/server/upload_layer_files_template.go
dobrac: it is, will remove

packages/orchestrator/internal/template/build/writer/postprocessor.go
dobrac: of the ticker? just to write "..." when something is happening in the CLI

but writing "..." is meaningless if you handle it correctly on the client side, so it's just backwards compatibility thing

packages/orchestrator/internal/template/build/utils/utils.go
dobrac: relict of the development, will remove

packages/orchestrator/internal/template/build/writer/postprocessor.go
jakubno: I was just wondering if it's turned off somewhere already

dobrac: yes, for the new build (isOldBuild variable)

spec/openapi.yml
```diff
@@ -1216,6 +1289,66 @@ paths:
...
+            schema:
+              $ref: "#/components/schemas/TemplateBuildRequestV2"
+
+      responses:
+        "202":
+          description: The build was requested successfully
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Template"
```
jakubno, line 1293: Can we get rid of multiple aliases? The whole type can be probably simplified to prevent any confussion

packages/orchestrator/internal/template/build/writer/postprocessor.go
dobrac: no, but I wanted to make it clear it's one or the other (I'll move the returns though)

packages/api/internal/handlers/template_layer_files_upload.go
dobrac: do we have any cache that I could use? I think the templateCache works only for already built templates because of `AND eb.status = 'uploaded'`

packages/api/internal/handlers/template_request_build_v2.go
```diff
@@ -0,0 +1,110 @@
...
+		return
+	}
+	span.End()
+
+	// Create the build, find the template ID by alias or generate a new one
+	_, span = a.Tracer.Start(ctx, "find-template-alias")
+	defer span.End()
+	templateID := id.Generate()
+	isNew := true
+	templateAlias, err := a.db.Client.EnvAlias.Query().Where(envalias.ID(body.Alias)).Only(ctx)
```
jakubno: can you use sqlc?

packages/api/internal/handlers/template_request_build.go
```diff
@@ -53,233 +82,157 @@ func (a *APIStore) PostTemplatesTemplateID(c *gin.Context, templateID api.Templa
...
 		Create().
-		SetID(templateID).
-		SetTeamID(team.ID).
-		SetCreatedBy(*userID).
+		SetID(req.TemplateID).
+		SetTeamID(req.Team.ID).
+		SetCreatedBy(req.UserID).
 		SetPublic(false).
-		SetNillableClusterID(team.ClusterID).
 		OnConflictColumns(env.FieldID).
```
jakubno: Why is this deleted?

packages/api/internal/handlers/template_request_build.go
```diff
@@ -53,233 +82,157 @@ func (a *APIStore) PostTemplatesTemplateID(c *gin.Context, templateID api.Templa
...
-
-		clusterNode, err := cluster.GetAvailableTemplateBuilder(ctx)
-		if err != nil {
-			a.sendAPIStoreError(c, http.StatusInternalServerError, fmt.Sprintf("Error when getting available template builder: %s", err))
-			telemetry.ReportCriticalError(ctx, "error when getting available template builder", err, telemetry.WithTemplateID(templateID))
-			return nil
-		}
-
-		builderNodeID = &clusterNode.NodeID
 	}
```
jakubno, line 215: why is this deleted?

packages/api/internal/handlers/template_request_build.go
```diff
@@ -53,233 +82,157 @@ func (a *APIStore) PostTemplatesTemplateID(c *gin.Context, templateID api.Templa
...
+		SetEnvID(req.TemplateID).
 		SetStatus(envbuild.StatusWaiting).
 		SetRAMMB(ramMB).
 		SetVcpu(cpuCount).
 		SetKernelVersion(schema.DefaultKernelVersion).
 		SetFirecrackerVersion(schema.DefaultFirecrackerVersion).
-		SetFreeDiskSizeMB(tier.DiskMb).
-		SetNillableStartCmd(body.StartCmd).
-		SetNillableReadyCmd(body.ReadyCmd).
-		SetNillableClusterNodeID(builderNodeID).
```
jakubno, line 273: why is this deleted?

packages/api/internal/handlers/template_request_build.go
```diff
@@ -378,18 +333,125 @@ func (a *APIStore) TemplateRequestBuild(c *gin.Context, templateID api.TemplateI
...
 		aliases = append(aliases, alias)
 	}
 
-	zap.L().Info("Built template", logger.WithTemplateID(templateID), logger.WithBuildID(buildID.String()))
+	zap.L().Info("template build requested", logger.WithTemplateID(req.TemplateID), logger.WithBuildID(buildID.String()))
+
+	return &TemplateBuildResponse{
+		TemplateID:         *build.EnvID,
+		BuildID:            build.ID.String(),
+		Public:             false,
```
jakubno: This is possibly incorrect for already existing templates

packages/api/internal/handlers/template_start_build_v2.go
```diff
@@ -0,0 +1,222 @@
...
+		a.templateCache.Invalidate(templateID)
+	}()
+
+	a.posthog.CreateAnalyticsUserEvent(userID.String(), team.ID.String(), "built environment", posthog.NewProperties().
+		Set("user_id", userID).
+		Set("environment", templateID).
+		Set("build_id", buildID).
+		Set("duration", time.Since(startTime).String()).
+		Set("success", err != nil),
+	)
```
jakubno: this part could be also probably shared

packages/api/internal/handlers/template_start_build_v2.go
```diff
@@ -0,0 +1,222 @@
...
+	startTime := time.Now()
+	build := envDB.Edges.Builds[0]
+
+	// only waiting builds can be triggered
+	if build.Status != envbuild.StatusWaiting {
+		a.sendAPIStoreError(c, http.StatusBadRequest, "build is not in waiting state")
+		telemetry.ReportCriticalError(ctx, "build is not in waiting state", fmt.Errorf("build is not in waiting state: %s", build.Status), telemetry.WithTemplateID(templateID))
+		return
+	}
+
```
jakubno, line 114: you are missing new cluster logic here

packages/api/internal/handlers/template_start_build_v2.go
```diff
@@ -0,0 +1,222 @@
...
+	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
+)
+
+type dockerfileStore struct {
+	FromImage string              `json:"from_image"`
+	Steps     *[]api.TemplateStep `json:"steps"`
+}
+
+// PostV2TemplatesTemplateIDBuildsBuildID triggers a new build
+func (a *APIStore) PostV2TemplatesTemplateIDBuildsBuildID(c *gin.Context, templateID api.TemplateID, buildID api.BuildID) {
```
jakubno, line 29: almost all here is reused, can you extract it?

spec/openapi.yml
```diff
@@ -1190,6 +1263,66 @@ paths:
...
+            description: Hash of the files
+
+      responses:
+        "201":
+          description: The upload link where to upload the tar file
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/TemplateBuildFileUpload"
+        "204":
```
jakubno: The server behaves differently

packages/orchestrator/template-manager.proto
```diff
@@ -78,4 +100,7 @@ service TemplateService {
 
   // todo (2025-05): this is deprecated, please use InfoService that is used for both orchestrator and template manager
   rpc HealthStatus (google.protobuf.Empty) returns (HealthStatusResponse);
+
+  // TemplateLayerFilesUpload requests an upload URL for a file to be cached for the template build.
+  rpc TemplateLayerFilesUpload (TemplateLayerFilesUploadRequest) returns (TemplateLayerFilesUploadResponse);
```
jakubno: not sure about the name, sounds like actual upload, I would call it
```suggestion
  rpc TemplateLayerFilesUploadUrl (TemplateLayerFilesUploadRequest) returns (TemplateLayerFilesUploadResponse);
```

packages/shared/pkg/storage/template.go
```diff
@@ -28,79 +28,80 @@ const (
...
-	BuildId            string
-	KernelVersion      string
-	FirecrackerVersion string
+	TemplateID         string `json:"template_id"`
+	BuildID            string `json:"build_id"`
+	KernelVersion      string `json:"kernel_version"`
+	FirecrackerVersion string `json:"firecracker_version"`
 }
 
+// Deprecated: Use TemplateFiles directly instead.
```
jakubno, line 37: Just curious, what made you to change it? Is there a reason why you didn't refactor the rest? Do you plan do it in some subsequent PR?

packages/orchestrator/internal/template/cache/build_cache.go
```diff
@@ -83,11 +85,30 @@ func (b *BuildInfo) GetLogs() []string {
...
+func (b *BuildInfo) Cancel() error {
+	err := b.fail(cancelledBuildReason)
+	if err != nil {
+		return fmt.Errorf("failed to cancel build: %w", err)
+	}
+
 	b.mu.Lock()
 	defer b.mu.Unlock()
-
 	b.ctxCancel()
```
jakubno, line 96: looking into it, the context is used only for tracing, maybe we don't need any context on the build info?

I want to prevent situation where the ctx will be used and will be bind to some actual logic and we will introduce race condition because of the locking inside `fail` the function and then again after `fail` execution.

packages/api/internal/handlers/template_start_build_v2.go
dobrac: I think I've done incorrect rebase here, there were a lot of conflicts

packages/api/internal/handlers/template_request_build.go
dobrac: reverted

packages/api/internal/handlers/template_request_build.go
dobrac: reverted

packages/api/internal/handlers/template_request_build.go
dobrac: reverted

packages/api/internal/handlers/template_request_build.go
dobrac: fixed

packages/api/internal/handlers/template_start_build_v2.go
dobrac: maybe, but I didn't want to share parts unless necessary as I expect them to diverge in the future

packages/api/internal/handlers/template_start_build_v2.go
dobrac: reverted

spec/openapi.yml
dobrac: fixed

packages/shared/pkg/storage/template.go
dobrac: I've changed it because imo it was hiding the fact, that it's not a stateful class, but it's just few variables together. I would do the rest of refactoring in a subsequent PR

spec/openapi.yml
dobrac: I don't follow, could you clarify please?

packages/orchestrator/internal/template/cache/build_cache.go
dobrac: the context should be used for the whole build
```
buildContext, buildSpan := s.tracer.Start(
  trace.ContextWithSpanContext(buildInfo.GetContext(), childSpan.SpanContext()),
  "template-background-build",
)
```

This is a good point, might need fix in a future PR

packages/orchestrator/template-manager.proto
dobrac: I'm thinking that it might change in the future to request multiple file uploads (also TemplateLayerFilesUploadUrl is really long), maybe different name then?

packages/api/internal/handlers/template_start_build_v2.go
dobrac: same answer as the previous comment:

maybe, but I didn't want to share parts unless necessary as I expect them to diverge in the future

packages/orchestrator/template-manager.proto
dobrac: maybe `InitLayerFileUpload` ?

packages/api/internal/handlers/template_start_build_v2.go
dobrac: I'll refactor the bottom part with BuildStatusSync, the top part is already refactored a bit in this PR https://github.com/e2b-dev/infra/pull/881

packages/api/internal/handlers/template_start_build_v2.go
dobrac: same answers as the other commit, I'll resolve this one so it's not in two places

packages/api/internal/handlers/template_start_build_v2.go
dobrac: refactored

spec/openapi.yml
dobrac: the aliases are not decided yet, let's change it in a future PR if necessary

packages/api/internal/handlers/template_request_build_v2.go
dobrac: done

packages/orchestrator/internal/sandbox/block/local.go
```diff
@@ -92,13 +92,18 @@ func (d *Local) Header() *header.Header {
...
 
-func (d *Local) UpdateSize() error {
+func (d *Local) UpdateSize(size int64) error {
 	info, err := d.f.Stat()
 	if err != nil {
 		return fmt.Errorf("failed to get file info: %w", err)
 	}
 
-	d.header.Metadata.Size = uint64(info.Size())
+	// Safety check to ensure the size matches the file size
```
jakubno: Does it make sense to do such check here? looks kinda random to pass an expected size

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -162,8 +172,11 @@ func CreateSandbox(
 		ips.slot,
 		sandboxFiles,
 		rootfsPath,
-		config.BaseTemplateId,
-		config.BuildId,
+		// The BaseTemplateID is always the same as config value for new sandboxes
```
```suggestion
		//The BaseTemplateID is always the same as TemplateID when creating a new sandbox.
```

packages/orchestrator/internal/template/build/builder.go
```diff
@@ -0,0 +1,685 @@
...
+		}
+	}()
+
+	cmdMeta := sandboxtools.CommandMetadata{
+		User:    defaultUser,
+		WorkDir: nil,
+		EnvVars: make(map[string]string),
+	}
+
+	// This is a compability for old template builds
```
jakubno:
```suggestion
	// This is a compatibility for old template builds
```

packages/orchestrator/internal/template/build/builder.go
```diff
@@ -0,0 +1,685 @@
...
+	}
+
+	return template
+}
+
+func (b *Builder) setupBase(
+	ctx context.Context,
+	finalMetadata storage.TemplateFiles,
+	template config.TemplateConfig,
+	envdVersion string,
```
jakubno: not used

packages/orchestrator/internal/template/build/hash_index.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	if templateMetadata.TemplateID == "" ||
+		templateMetadata.BuildID == "" ||
+		templateMetadata.KernelVersion == "" ||
+		templateMetadata.FirecrackerVersion == "" {
+		return newTemplate
+	}
+
+	return templateMetadata
```
jakubno: what if the firecracker version differs should we still use the cache?

packages/orchestrator/internal/template/build/builder.go
```diff
@@ -0,0 +1,685 @@
...
+		EnvdVersion:  envdVersion,
+		RootfsSizeMB: int64(rootfsSize >> constants.ToMBShift),
+	}, nil
+}
+
+func getRootfsSize(
+	ctx context.Context,
+	s storage.StorageProvider,
+	metadata storage.TemplateFiles,
+	config config.TemplateConfig,
```
jakubno: not used

packages/orchestrator/internal/template/build/hash_index.go
```diff
@@ -0,0 +1,101 @@
...
+	buf := bytes.NewBuffer(marshalled)
+	_, err = obj.ReadFrom(buf)
+	if err != nil {
+		return fmt.Errorf("error writing UUID to object: %w", err)
+	}
+
+	return nil
+}
+
+func hashKeys(previousHash string, keys ...string) string {
```
jakubno: in `hashBase` it isn't used with `previousHash`, I think you can make it more general by having only `keys`

packages/orchestrator/internal/template/build/hash_index.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/e2b-dev/infra/packages/orchestrator/internal/template/build/envd"
+	"github.com/e2b-dev/infra/packages/orchestrator/internal/template/build/utils"
+	templatemanager "github.com/e2b-dev/infra/packages/shared/pkg/grpc/template-manager"
+	"github.com/e2b-dev/infra/packages/shared/pkg/id"
+	"github.com/e2b-dev/infra/packages/shared/pkg/storage"
+)
+
+const hashingVersion = "v1"
+
+func getTemplateFromHash(ctx context.Context, s storage.StorageProvider, m storage.TemplateFiles, finalTemplateID string, hash string) storage.TemplateFiles {
```
jakubno: wouldn't `getMetadataFromHash` be a better name? Maybe even `getOrCreate`

packages/orchestrator/internal/template/build/base_files.go
```diff
@@ -4,29 +4,26 @@ import (
...
-//go:embed provision.sh
-var provisionScriptFile string
-var ProvisionScriptTemplate = template.Must(template.New("provisioning-script").Parse(provisionScriptFile))
-
-//go:embed configure.sh
-var configureScriptFile string
-var ConfigureScriptTemplate = template.Must(template.New("provisioning-finish-script").Parse(configureScriptFile))
-
-func Build(
+func constructBaseLayerFiles(
```
jakubno, line 21: why are there two buildIIDs? the first in `metadata` and the second explicitly passed

packages/orchestrator/internal/template/build/builder.go
```diff
@@ -0,0 +1,685 @@
...
+		}
+
+		err = rootfs.UpdateSize(rootfsSize)
+		if err != nil {
+			return nil, fmt.Errorf("error updating rootfs size: %w", err)
+		}
+
+		// Create sandbox for building template
+		postProcessor.WriteMsg("Creating sandbox template")
+		baseSandboxConfig = proto.Clone(baseSandboxConfig).(*orchestrator.SandboxConfig)
```
jakubno, line 295: why do you need to clone here if you then overwrite the variable? Also is this really needed? I would add a comment this is only to make sure there isn't any contamination and sandbox id nor build id is used in any meaningful way in create

packages/orchestrator/internal/template/build/hash_index.go
```diff
@@ -0,0 +1,101 @@
...
+		return fmt.Errorf("error writing UUID to object: %w", err)
+	}
+
+	return nil
+}
+
+func hashKeys(previousHash string, keys ...string) string {
+	sha := sha256.New()
+	sha.Write([]byte(previousHash))
+	for _, key := range keys {
```
jakubno, line 72: wouldn't it make sense to add some separator between each key?

packages/orchestrator/internal/template/build/build_layer.go
```diff
@@ -0,0 +1,265 @@
...
+	postProcessor *writer.PostProcessor,
+	uploadErrGroup *errgroup.Group,
+	sourceSbxConfig *orchestrator.SandboxConfig,
+	finalTemplateID string,
+	hash string,
+	sourceMeta storage.TemplateFiles,
+	exportMeta storage.TemplateFiles,
+	resumeSandbox bool,
+	allowInternet bool,
+	fun func(ctx context.Context, sbx *sandbox.Sandbox) error,
```
jakubno: no fun :D 
```suggestion
	fn func(ctx context.Context, sbx *sandbox.Sandbox) error,
```

packages/orchestrator/internal/template/build/build_layer.go
```diff
@@ -0,0 +1,265 @@
...
+		sourceMeta.BuildID,
+		sourceMeta.KernelVersion,
+		sourceMeta.FirecrackerVersion,
+	)
+	if err != nil {
+		return fmt.Errorf("failed to get template snapshot data: %w", err)
+	}
+
+	if resumeSandbox {
+		sbxConfig := &orchestrator.SandboxConfig{
```
jakubno, line 59: are the configs for resume and create here different?

packages/orchestrator/internal/template/build/builder.go
```diff
@@ -0,0 +1,685 @@
...
+		// Run commands in the sandbox only if not cached
+		if !isCached {
+			err = b.buildLayer(
+				ctx,
+				postProcessor,
+				uploadErrGroup,
+				&orchestrator.SandboxConfig{
+					BaseTemplateId: baseMetadata.TemplateID,
+
+					// TODO: Here might be invalid data for the template resume, but they might now be used
```
jakubno: in which sense can they be invalid?

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -162,8 +172,11 @@ func CreateSandbox(
...
+		config.TemplateId,
+		// The rootfs build ID is from the header, because it needs to be the same from
+		// the first FS creation.
```
jakubno, line 178: Imho this comment doesn't help much. What's rootfs build ID? Why does this make it the same.

packages/orchestrator/internal/sandbox/sandbox.go
jakubno: it would be nice to get rid of the variables from the paths...

packages/orchestrator/internal/template/build/build_layer.go
```diff
@@ -0,0 +1,265 @@
...
+func (b *Builder) buildLayer(
+	ctx context.Context,
+	postProcessor *writer.PostProcessor,
+	uploadErrGroup *errgroup.Group,
+	sourceSbxConfig *orchestrator.SandboxConfig,
+	finalTemplateID string,
+	hash string,
+	sourceMeta storage.TemplateFiles,
+	exportMeta storage.TemplateFiles,
+	resumeSandbox bool,
```
jakubno, line 37: resume is only for performance? would be nice to add comment explaining it

packages/orchestrator/internal/template/build/apply_command.go
```diff
@@ -0,0 +1,284 @@
...
+		attribute.String("metadata.user", buildMetadata.User),
+		attribute.String("metadata.workdir", utils.Sprintp(buildMetadata.WorkDir)),
+		attribute.String("metadata.env_vars", fmt.Sprintf("%v", buildMetadata.EnvVars)),
+	))
+	defer span.End()
+
+	cmdType := strings.ToUpper(step.Type)
+	args := step.Args
+
+	switch cmdType {
```
jakubno, line 38: it would be nice to make cmdType an enum

packages/orchestrator/internal/template/build/sandboxtools/file.go
```diff
@@ -0,0 +1,106 @@
...
+
+	// Pipe to stream data
+	pr, pw := io.Pipe()
+	writer := multipart.NewWriter(pw)
+
+	go func() {
+		defer pw.Close()
+		part, err := writer.CreateFormFile("file", filepath.Base(sourcePath))
+		if err != nil {
+			log.Fatalf("Failed to create form file: %v", err)
```
jakubno: This is probably copied from tests, we don't want to fatal here

packages/orchestrator/internal/template/build/sandboxtools/file.go
```diff
@@ -0,0 +1,106 @@
...
+		}
+		defer writer.Close()
+		if _, err := io.Copy(part, file); err != nil {
+			log.Fatalf("Failed to copy file: %v", err)
```
jakubno: This is probably copied from tests, we don't want to fatal here

packages/orchestrator/internal/template/build/builder.go
```diff
@@ -0,0 +1,685 @@
...
+					EnvdVersion: envdVersion,
+				},
+				finalMetadata.TemplateID,
+				lastHash,
+				sourceMetadata,
+				stepMetadata,
+				true,
+				globalconfig.AllowSandboxInternet,
+				func(ctx context.Context, sbx *sandbox.Sandbox) error {
+					postProcessor.WriteMsg(fmt.Sprintf("Running action in: %s/%s", sourceMetadata.TemplateID, sourceMetadata.BuildID))
```
jakubno, line 423: it would be nice to write what action

packages/orchestrator/internal/template/build/sandboxtools/file.go
```diff
@@ -0,0 +1,106 @@
...
+	pr, pw := io.Pipe()
+	writer := multipart.NewWriter(pw)
+
+	go func() {
+		defer pw.Close()
+		part, err := writer.CreateFormFile("file", filepath.Base(sourcePath))
+		if err != nil {
+			log.Fatalf("Failed to create form file: %v", err)
+		}
+		defer writer.Close()
```
jakubno: you should call CloseWithError if there's an error

packages/orchestrator/internal/template/build/apply_command.go
```diff
@@ -0,0 +1,284 @@
...
+
+		err = sandboxtools.RunCommand(
+			ctx,
+			b.tracer,
+			b.proxy,
+			b.buildLogger,
+			nil,
+			prefix,
+			sbx.Metadata.Config.SandboxId,
+			fmt.Sprintf(`tar -xzvf "%s" -C "%s"`, sbxTargetPath, sbxUnpackPath),
```
jakubno: can't you just copy it to `sbxTargetPath` and run somethign like or at least do mkdir in the same command?
```
	fmt.Sprintf(`tar -xzvf "%s" -C "%s" && rm "%s"`, sbxTargetPath, args[1], sbxTargetPath),
```

packages/orchestrator/internal/template/build/apply_command.go
```diff
@@ -0,0 +1,284 @@
...
+		return sandboxtools.RunCommand(
+			ctx,
+			b.tracer,
+			b.proxy,
+			b.buildLogger,
+			postProcessor,
+			prefix,
+			sbx.Metadata.Config.SandboxId,
+			fmt.Sprintf(`mkdir -p "%s"`, utils.Sprintp(cmdMetadata.WorkDir)),
+			sandboxtools.CommandMetadata{
```
jakubno, line 258: why can't you use `cmdMetadata` ?

packages/orchestrator/internal/template/build/apply_command.go
```diff
@@ -0,0 +1,284 @@
...
+	defer span.End()
+
+	cmdType := strings.ToUpper(step.Type)
+	args := step.Args
+
+	switch cmdType {
+	case "ADD":
+		// args: [localPath containerPath]
+		fallthrough
+	case "COPY":
```
jakubno:
```suggestion
	case "ADD", "COPY":
```

packages/orchestrator/internal/template/build/apply_command.go
```diff
@@ -0,0 +1,284 @@
...
+			postProcessor,
+			prefix,
+			sbx.Metadata.Config.SandboxId,
+			"adduser "+args[0],
+			sandboxtools.CommandMetadata{
+				User:    "root",
+				EnvVars: cmdMetadata.EnvVars,
+			},
+		)
+	case "WORKDIR":
```
jakubno, line 243: here you could probably use `fallthrough`

packages/orchestrator/internal/template/build/configure.sh
```diff
@@ -4,6 +4,12 @@ set -euo pipefail
 
 echo "Starting configuration script"
 
+cat <<EOF > /.e2b
+ENV_ID={{ .TemplateID }}
```
jakubno, line 8: Do we need this?

packages/orchestrator/internal/template/build/provision.sh
```diff
@@ -1,5 +1,5 @@
-#!/bin/bash
-set -euo pipefail
+#!/bin/sh
```
jakubno, line 1: this is because some images doesn't have to have bash right?

packages/orchestrator/internal/template/build/builder.go
```diff
@@ -0,0 +1,685 @@
...
+	baseCached, err := isCached(ctx, b.templateStorage, baseMetadata, template)
+	if err != nil {
+		return false, storage.TemplateFiles{}, fmt.Errorf("error checking if base layer is cached: %w", err)
+	}
+
+	return baseCached, baseMetadata, nil
+}
+
+func (b *Builder) runPostprocessing(
+	ctx context.Context,
```
jakubno:
```suggestion
	_ context.Context,
```

packages/orchestrator/internal/template/build/builder.go
```diff
@@ -0,0 +1,685 @@
...
+
+	baseCached, err := isCached(ctx, b.templateStorage, baseMetadata, template)
+	if err != nil {
+		return false, storage.TemplateFiles{}, fmt.Errorf("error checking if base layer is cached: %w", err)
+	}
+
+	return baseCached, baseMetadata, nil
+}
+
+func (b *Builder) runPostprocessing(
```
jakubno: maybe `getPostprocessingFn`, it isn't running anyything

packages/orchestrator/internal/template/build/builder.go
```diff
@@ -0,0 +1,685 @@
...
+	cmdMeta sandboxtools.CommandMetadata,
+) func(ctx context.Context, sbx *sandbox.Sandbox) error {
+	return func(ctx context.Context, sbx *sandbox.Sandbox) error {
+		// Run configuration script
+		err := runConfiguration(
+			ctx,
+			b.tracer,
+			b.proxy,
+			b.buildLogger,
+			// TODO: Make this debug level
```
jakubno, line 608: what about this?

packages/orchestrator/internal/template/build/build_layer.go
```diff
@@ -0,0 +1,265 @@
...
+			sbxConfig,
+			uuid.New().String(),
+			time.Now(),
+			time.Now().Add(layerTimeout),
+			b.devicePool,
+			allowInternet,
+			false,
+		)
+	} else {
+		var oldMemfile block.ReadonlyDevice
```
jakubno, line 89: why old?

packages/orchestrator/internal/template/build/hash_index.go
```diff
@@ -0,0 +1,101 @@
...
+		templateMetadata.BuildID == "" ||
+		templateMetadata.KernelVersion == "" ||
+		templateMetadata.FirecrackerVersion == "" {
+		return newTemplate
+	}
+
+	return templateMetadata
+}
+
+func saveTemplateToHash(ctx context.Context, s storage.StorageProvider, finalTemplateID string, hash string, template storage.TemplateFiles) error {
```
jakubno: Are you saving it to "hash" or to a storage? (`saveTemplateToHash`)

packages/orchestrator/internal/sandbox/template/cache.go
```diff
@@ -47,7 +49,11 @@ func NewCache(ctx context.Context) (*Cache, error) {
 		}
 	})
 
-	go cache.Start()
+	// Delete the old build cache directory content as it may contain stale data that are not managed by anyone.
```
jakubno: maybe this could be in function description

jakubno (changes requested):

packages/api/internal/handlers/template_layer_files_upload.go
jakubno: new cache than? It may be worth it to generalized the cache

packages/orchestrator/internal/sandbox/sandbox.go
dobrac: it would, not in this PR

packages/orchestrator/internal/template/build/apply_command.go
dobrac: I plan to refactor this using map (cmd type -> function handler), but I want to do it in a new PR

packages/orchestrator/internal/template/build/builder.go
dobrac: we already write it, it's in the layer name (also, this is a debug message, but thats updated in this PR https://github.com/e2b-dev/infra/pull/880)

packages/orchestrator/internal/template/build/apply_command.go
dobrac: I can do mkdir, not rm (also rm is not necessary to do, because /tmp is cleared on restart)

packages/orchestrator/internal/sandbox/block/local.go
dobrac: I'm also setting the size to the passed one. The idea is that it's transparent what the function does and there is no hidden logic from where it gets the size

packages/orchestrator/internal/template/build/hash_index.go
dobrac: we call everywhere `TemplateFiles` template, so I was keeping the same logic

Not a fan of this length `getOrCreateTemplateFromHash` (also, it doesn't really add any new meaning)

packages/orchestrator/internal/template/build/base_files.go
dobrac: the buildID is the base layer buildID, the finalMetadata.buildID is the final buildID

The base and the final can differ

packages/orchestrator/internal/template/build/builder.go
dobrac: I do clone, because proto struct contains mutex

> I would add a comment this is only to make sure there isn't any contamination and sandbox id nor build id is used in any meaningful way in create

I don't understand this, it's set because you start a new sandbox

packages/orchestrator/internal/template/build/hash_index.go
dobrac: sure, what would you add? Maybe `|` or `;`?

packages/orchestrator/internal/template/build/build_layer.go
dobrac: yes they are in half of the fields, unfortunately it's grpc struct, so it's not that easy to work with it + you probably don't even need half of the fields (but thats for another PR refactor)

packages/orchestrator/internal/template/build/builder.go
dobrac: because the data for resume are taken from the new build request, but you're resuming snapshot of older build cache (e.g. the layer was cached with 1024 memory, but you want now to build it with 2048 memory).

But I think these params are not used at all and they're just unnecessary to pass. But this requires large refactor

packages/orchestrator/internal/sandbox/sandbox.go
dobrac: This will be longer, TLDR:

When we build rootfs right now, we encode the buildID and templateID to the mount path of the firecracker start. This mount path can't be changed later when you do pause and resume on the same sandbox.
Because of the encoding, we save buildID in the rootfs header (because we assume the templateID we know always - tricky now though, that's why it's not saved in the header though). Each resume, we need to know the base build ID of the rootfs.

Now, the same applies to creating a sandbox from already "paused" rootfs filesystem (the rootfs filesystem has many layers that it's composed from). This is because of internals how we reference the build layers in the diff system.

(kinda +- there are still lot of details here)

packages/orchestrator/internal/template/build/build_layer.go
dobrac: I guess, actually we could do probably create, but I was assuming resume should be faster in the future and create doesn't have to be that fast. Haven't tested it though

packages/orchestrator/internal/template/build/sandboxtools/file.go
dobrac: refactored

packages/orchestrator/internal/template/build/apply_command.go
dobrac: the workdir shouldn't be set

packages/orchestrator/internal/template/build/apply_command.go
dobrac: how?

packages/orchestrator/internal/template/build/configure.sh
dobrac: do you mean the .e2b file or ENV_ID? If file, I think we do, people use it, if ENV_ID, I think we do, people rely on it now

packages/orchestrator/internal/template/build/provision.sh
dobrac: yeah, I was thinking that we might support images without bash, but it would be long way anyway right now so it's more like I've had that already done in /sh

packages/orchestrator/internal/template/build/builder.go
dobrac: not gets in go though

packages/orchestrator/internal/template/build/build_layer.go
dobrac: because I'm creating a new one that is up to date to the requested memory size, but I need the previous one for some metadata

packages/orchestrator/internal/template/build/hash_index.go
dobrac: renamed

packages/orchestrator/internal/sandbox/template/cache.go
dobrac: moved

packages/orchestrator/internal/template/build/hash_index.go
dobrac: I wanted to stress that you need to always provide at least one key

dobrac: renamed to "baseKey"

packages/orchestrator/internal/template/build/builder.go
dobrac: next PR: https://github.com/e2b-dev/infra/pull/880

packages/orchestrator/internal/template/build/hash_index.go
dobrac: it's the other way around, we will still use the old FC version

packages/api/internal/handlers/template_layer_files_upload.go
dobrac: yes, let's do it in another PR though

packages/orchestrator/template-manager.proto
jakubno: How can you request multiple file if the request can send only 1? I think in that case it will be cleaner to add a new rpc.

`InitLayerFileUpload` could work

packages/orchestrator/internal/template/build/hash_index.go
jakubno: `;` seems as a standard way to do it

dobrac: just to be sure, it's just a hash computation, the character won't show anywhere

packages/orchestrator/internal/template/build/hash_index.go
jakubno: image

even inside 
[the body of the function](https://github.com/e2b-dev/infra/blob/ceaee7835b9e546cbb68b59c77f3526c73556f5d/packages/orchestrator/internal/template/build/hash_index.go#L44
)
I think it would be much cleaner to return error if the template isn't here, you are almost doing this in `stepMetadata` anyway

dobrac: I've already renamed it. The error is a good point, I'll redo it

packages/orchestrator/internal/template/build/base_files.go
jakubno: let's call it finalBuildID then to make it consistent with other places

packages/orchestrator/internal/template/build/builder.go
jakubno: but why do you need a copy in the first place

packages/orchestrator/internal/template/build/base_files.go
dobrac: its the other way around, it's not the final one, it's the current one

packages/orchestrator/internal/template/build/builder.go
dobrac: to share these settings and not have to define them again:

```
TemplateId:         baseMetadata.TemplateID,
BuildId:            baseMetadata.BuildID,
KernelVersion:      baseMetadata.KernelVersion,
FirecrackerVersion: baseMetadata.FirecrackerVersion,

BaseTemplateId: baseMetadata.TemplateID,

Vcpu:        template.VCpuCount,
RamMb:       template.MemoryMB,
HugePages:   template.HugePages,
EnvdVersion: envdVersion,
```

packages/orchestrator/internal/template/build/build_layer.go
jakubno: please write it down

packages/orchestrator/internal/template/build/builder.go
jakubno: shouldn't there be `but they might noT be used`? can you also remove the uncertainty from the sentence?

packages/orchestrator/internal/sandbox/sandbox.go
jakubno: Add this information to the comment, that we are referencing the build id in the path of filesystem when resuming the sandbox

packages/orchestrator/internal/template/build/builder.go
dobrac: yes, it's already fixed

I'm not 100% sure though 👀

packages/orchestrator/internal/template/build/apply_command.go
jakubno: What's the reason you have to copy it to tmp, extract it there and then copy it to the destination?

packages/orchestrator/internal/template/build/apply_command.go
jakubno: set cmd + metadata and execute the command, but it's fine

packages/api/internal/handlers/template_layer_files_upload.go
jakubno, line 35: can you make it sqlc 🙏🏻 I would really like to end up with only one package for DB calls

packages/orchestrator/internal/template/build/hash_index.go
jakubno: I know, this is just to prevent any unexpected collisions, even though it's probably impossible right now.

packages/orchestrator/internal/template/build/sandboxtools/file.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	req.Header.Set("Content-Type", writer.FormDataContentType())
+	err = grpc.SetSandboxHeader(req.Header, proxyHost, sandboxID)
+	if err != nil {
+		return fmt.Errorf("failed to set request header: %w", err)
+	}
+	req.Host = req.Header.Get("Host")
+
+	// Send request
+	client := http.Client{
```
jakubno: can you reuse the client?

packages/orchestrator/internal/template/server/upload_layer_files_template.go
```diff
@@ -0,0 +1,41 @@
...
+	"fmt"
+	"time"
+
+	"github.com/e2b-dev/infra/packages/orchestrator/internal/template/build"
+	templatemanager "github.com/e2b-dev/infra/packages/shared/pkg/grpc/template-manager"
+)
+
+const signedUrlExpiration = time.Minute * 30
+
+func (s *ServerStore) TemplateLayerFilesUpload(ctx context.Context, in *templatemanager.InitLayerFileUploadRequest) (*templatemanager.InitLayerFileUploadResponse, error) {
```
jakubno: this needs to be renamed as well

packages/orchestrator/internal/template/build/builder.go
```diff
@@ -559,10 +562,21 @@ func (b *Builder) setupBase(
...
+		KernelVersion:      finalMetadata.KernelVersion,
+		FirecrackerVersion: finalMetadata.FirecrackerVersion,
+	}
+	bm, err := templateMetaFromHash(ctx, b.buildStorage, finalMetadata.TemplateID, hash)
+	if err != nil {
+		b.logger.Info("base layer not found in cache, building new base layer", zap.Error(err), zap.String("hash", hash))
+	} else {
+		baseMetadata = bm
+	}
+
```
jakubno: isn't this more readable? 
```
	var baseMetadata storage.TemplateFiles
	bm, err := templateMetaFromHash(ctx, b.buildStorage, finalMetadata.TemplateID, hash)
	if err != nil {
		b.logger.Info("base layer not found in cache, building new base layer", zap.Error(err), zap.String("hash", hash))
		baseMetadata = storage.TemplateFiles{
			TemplateID:         id.Generate(),
			BuildID:            uuid.New().String(),
			KernelVersion:      finalMetadata.KernelVersion,
			FirecrackerVersion: finalMetadata.FirecrackerVersion,
		}
	} else {
		baseMetadata = bm
	}
```

packages/orchestrator/internal/template/cache/build_cache.go
jakubno: What's the resolution then? You will create a linear issue and handle it later?

packages/orchestrator/internal/sandbox/block/local.go
jakubno: The check seems kinda random. You know the size, but you still expect the user to pass it.

packages/orchestrator/internal/template/build/base_files.go
jakubno: *baseBuildID + add comment, from the signature of the function you can't tell what's expected

packages/orchestrator/internal/template/build/builder.go
dobrac: not a fan of big if statements, but doesn't matter much here, will change

packages/orchestrator/internal/template/build/builder.go
jakubno: but why you can't use the `baseSandboxConfig` variable (without copying it)? I am maybe missing something

packages/orchestrator/internal/template/build/builder.go
jakubno: and who should be?

packages/orchestrator/internal/template/build/apply_command.go
dobrac: it's different in following PR (https://github.com/e2b-dev/infra/pull/885) already, so let's keep it

packages/orchestrator/internal/template/build/hash_index.go
dobrac: added `;`

packages/orchestrator/internal/template/build/base_files.go
dobrac: added

packages/orchestrator/internal/template/build/builder.go
dobrac: I can't safely use it because it's protobuf defined type (and a pointer) and it contains mutex, so any copy operation causes warning

```
type SandboxConfig struct {
	state         protoimpl.MessageState
	sizeCache     protoimpl.SizeCache
	unknownFields protoimpl.UnknownFields

	// Data required for creating a new sandbox.
	TemplateId         string            `protobuf:"bytes,1,opt,name=template_id,json=templateId,proto3" json:"template_id,omitempty"`
	BuildId            string            `protobuf:"bytes,2,opt,name=build_id,json=buildId,proto3" json:"build_id,omitempty"`
```

dobrac: hmm, it might probably work to just update the two fields in the old one, but because it's a pointer, it might have undesired side effects in the future (now it should be ok)

packages/orchestrator/internal/template/build/builder.go
dobrac: updated

I would be 100% sure when this whole codepath is refactored, and no more proto struct with unnecessary fields is used, but it's fine now

packages/orchestrator/internal/sandbox/sandbox.go
dobrac: it's connected with pause, but that's not the cause

The important part is that it's dependent on the first filesystem creation buildID (which is saved i the metadata), because we use it that way in the approach how we are mounting the filesystem. So imo the important information at that point in the code is, that this buildID needs to be the first base build ID when you created the filesystem block device. (It doesn't matter how you created the FS)

packages/orchestrator/internal/template/cache/build_cache.go
dobrac: yes, done

packages/orchestrator/internal/sandbox/block/local.go
dobrac: refactored

packages/orchestrator/internal/template/build/build_layer.go
dobrac: will be refactored in a future PR where the create will be used for first uncached layer and resume afterwards

packages/api/internal/handlers/template_layer_files_upload.go
dobrac: can I refactor it in this PR https://github.com/e2b-dev/infra/pull/881/files#diff-a82be3f1a0990adbbfa2845f263be11b77e6016f8610bf8555ce77bd991687a3?

It's gonna be completely different query

PR #855, opened 2025-07-08
Remove use of sandbox id client part

sitole:
Removes the use of the sandbox ID client part and always falls back to the cache/database to get the node ID where the sandbox is running.

This is the first part that is needed for the client part removal.


packages/api/internal/handlers/sandbox_resume.go
```diff
@@ -121,6 +102,30 @@ func (a *APIStore) PostSandboxesSandboxIDResume(c *gin.Context, sandboxID api.Sa
 	snap := lastSnapshot.Snapshot
 	build := lastSnapshot.EnvBuild
 
+	var nodeID *string
+	if snap.OriginNodeID != nil {
+		nodeID = snap.OriginNodeID
+	} else {
+		// After migration period, we can remove this part, because all actively used snapshots will be stored in the database with the node ID.
```
jakubno: Can you create a linear task and reference it here, ideally with TODO

jakubno (commented):
There are still some places where we use `instance.ClientID`, could you update these?

sitole:
> There are still some places where we use `instance.ClientID`, could you update these?

Do you have some exact places? Currently, it's used only for sbx start, detail, and list endpoints, and it's communicated just to the API response, not used for actual logic. We should probably leave it here.

jakubno (approved):

PR #856, opened 2025-07-08
Resolve symlinks in envd filesystem links

jakubno:
# Description

- Resolves the symlink if provided as dir path. The resolution of symlinks is on purpose turned off to prevent issues with circular dependancies. With #745 you should be able to see symlinks and you can resolve the list content yourself


packages/envd/internal/services/filesystem/dir.go
```diff
@@ -30,7 +32,22 @@ func (Service) ListDir(ctx context.Context, req *connect.Request[rpc.ListDirRequ
 		return nil, connect.NewError(connect.CodeInvalidArgument, err)
 	}
 
-	stat, err := os.Stat(dirPath)
+	// Resolve symlinks
+	resolved, err := filepath.EvalSymlinks(dirPath)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil, connect.NewError(connect.CodeNotFound, fmt.Errorf("directory not found: %w", err))
```
dobrac: symlink doesn't have to be directory only

packages/envd/internal/services/filesystem/dir.go
```diff
@@ -68,6 +85,7 @@ func (Service) ListDir(ctx context.Context, req *connect.Request[rpc.ListDirRequ
 		entries = append(entries, &rpc.EntryInfo{
 			Name: entry.Name(),
 			Type: getEntryType(entry),
+			// Returns the "real" path - resolved symlinks
```
dobrac: Where do you get the `path` as resolved symlink?

packages/envd/internal/services/filesystem/dir.go
```diff
@@ -30,7 +32,22 @@ func (Service) ListDir(ctx context.Context, req *connect.Request[rpc.ListDirRequ
...
+
+		if strings.Contains(err.Error(), "too many links") {
+			return nil, connect.NewError(connect.CodeInvalidArgument, fmt.Errorf("cyclic symlink or chain >255 links at %q", dirPath))
+		}
+
+		return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("error resolving symlink: %w", err))
+	}
+
+	// Check if the path is a directory
+	stat, err := os.Stat(resolved)
```
dobrac: still calling `filepath.WalkDir(dirPath,...)`, shouldn't you use the resolved path or it's ok as symlink?

tests/integration/internal/tests/envd/filesystem/list_test.go
```diff
@@ -0,0 +1,195 @@
...
+	linkToFile := fmt.Sprintf("%s/link-file", symlinkRoot) // → file
+	cyclicLink := fmt.Sprintf("%s/cyclic", symlinkRoot)    // → itself
+
+	createSymlink(t, ctx, envdClient, sbx, realDir, linkToDir)
+	createSymlink(t, ctx, envdClient, sbx, realFile, linkToFile)
+	createSymlink(t, ctx, envdClient, sbx, cyclicLink, cyclicLink)
+
+	t.Run("symlink to directory behaves like directory", func(t *testing.T) {
+		req := connect.NewRequest(&filesystem.ListDirRequest{
+			Path:  realDir,
```
dobrac: can you also test:

`eq := connect.NewRequest(&filesystem.ListDirRequest{
			Path:  linkToDir,...})`

tests/integration/internal/utils/filesystem.go
```diff
@@ -0,0 +1,74 @@
...
+	"connectrpc.com/connect"
+
+	"github.com/e2b-dev/infra/packages/shared/pkg/grpc/envd/filesystem"
+	"github.com/e2b-dev/infra/tests/integration/internal/api"
+	envdapi "github.com/e2b-dev/infra/tests/integration/internal/envd/api"
+	"github.com/e2b-dev/infra/tests/integration/internal/setup"
+)
+
+func UploadFile(tb testing.TB, ctx context.Context, sbx *api.Sandbox, envdClient *setup.EnvdClient, path string, content []byte) error {
+	buffer, contentType := CreateTextFile(tb, path, string(content))
```
dobrac, line 22: add `tb.Helper()`

tests/integration/internal/utils/filesystem.go
```diff
@@ -0,0 +1,74 @@
...
+
+	if writeRes.StatusCode() != http.StatusOK {
+		return fmt.Errorf("failed to upload file %s, status code: %d", path, writeRes.StatusCode())
+	}
+
+	return nil
+}
+
+func CreateTextFile(tb testing.TB, path string, content string) (*bytes.Buffer, string) {
+	body := &bytes.Buffer{}
```
dobrac, line 48: add `tb.Helper()`

tests/integration/internal/utils/process.go
```diff
@@ -0,0 +1,56 @@
...
+
+	"connectrpc.com/connect"
+
+	"github.com/e2b-dev/infra/packages/shared/pkg/grpc/envd/process"
+	"github.com/e2b-dev/infra/tests/integration/internal/api"
+	"github.com/e2b-dev/infra/tests/integration/internal/setup"
+)
+
+func ExecCommand(t *testing.T, ctx context.Context, sbx *api.Sandbox, envdClient *setup.EnvdClient, command string, args ...string) error {
+	req := connect.NewRequest(&process.StartRequest{
```
dobrac, line 18: `t.Helper()`

packages/envd/internal/services/filesystem/dir.go
jakubno: Let's say you want to list files from `/a` and `/a` is symlink to `/b` then the result will be 
```
b/x.txt
b/folder/x.txt
```

jakubno: Should we maybe replace the "base"?

packages/envd/internal/services/filesystem/dir.go
```diff
@@ -154,3 +95,84 @@ func (Service) MakeDir(ctx context.Context, req *connect.Request[rpc.MakeDirRequ
...
+
+	// Resolve symlinks
+	resolvedPath, err := filepath.EvalSymlinks(expandedPath)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return "", connect.NewError(connect.CodeNotFound, fmt.Errorf("path not found: %w", err))
+		}
+
+		if strings.Contains(err.Error(), "too many links") {
+			return "", connect.NewError(connect.CodeInvalidArgument, fmt.Errorf("cyclic symlink or chain >255 links at %q", expandedPath))
```
dobrac:
```suggestion
			return "", connect.NewError(connect.CodeFailedPrecondition, fmt.Errorf("cyclic symlink or chain >255 links at %q", expandedPath))
```

dobrac (changes requested):
one nit

packages/envd/internal/services/filesystem/dir.go
dobrac: I think it's fine now, or do you think it could cause any issues?

jakubno: Now returning the requested path (not the resolved)

tests/integration/internal/tests/envd/filesystem_test.go
```diff
@@ -1,145 +1,18 @@
...
 	"github.com/e2b-dev/infra/packages/shared/pkg/grpc/envd/process"
-	"github.com/e2b-dev/infra/tests/integration/internal/api"
-	envdapi "github.com/e2b-dev/infra/tests/integration/internal/envd/api"
 	"github.com/e2b-dev/infra/tests/integration/internal/setup"
 	"github.com/e2b-dev/infra/tests/integration/internal/utils"
 )
 
-const testFolder = "/test"
-
-func TestListDir(t *testing.T) {
```
dobrac: I don't think that the unit tests replace these tests fully. You can have issue in the constructed envd filesystem that won't show in the unit tests, but might show here. (you might still reduce the test amount though)

dobrac (changes requested):
one more thing

PR #861, opened 2025-07-10
Add sandbox metrics endpoint

jakubno:
# Description

Adds an endpoint for getting sandbox metrics
```
GET /sandboxes/{id}/metrics
```

Possible to specify `start` and `end`.
If no `start` and `end` is provided (default), then it will query clickhouse to get the max range.

The `step` is calculated automatically based on following table

```
duration             aggregation step
-----------------------------------------
≤   1 h                  5 s (all points)
≤   6 h                 30 s
≤  12 h                  1 m
≤  24 h                  2 m
```


packages/api/internal/handlers/sandbox_metrics.go
```diff
@@ -0,0 +1,127 @@
...
+			MemUsed:    int64(m.MemUsed),
+		}
+	}
+
+	c.JSON(http.StatusOK, apiMetrics)
+}
+
+// calculateStep determines the step size for metrics based on the time range.
+// The result should always contain less than 1000 points.
+func calculateStep(start, end time.Time) time.Duration {
```
jakubno, line 79: Not sure here if it should be just based on start and end - for paused and resumed sandbox it can result in only few data points, but on the other side, you probably won't be able to use more point in the plot it in the dashboard

dobrac: I hope we can tweak this based on how it works on some data in the dashboard

packages/clickhouse/pkg/metrics.go
```diff
@@ -64,3 +67,84 @@ func (c *Client) QueryLatestMetrics(ctx context.Context, sandboxIDs []string, te
...
+        'e2b.sandbox.ram.used'
+    )
+  AND TimeUnix >= {start_time:DateTime64}
+  AND TimeUnix <= {end_time:DateTime64}
+GROUP BY timestamp
+ORDER BY timestamp;
+`
+
+func (c *Client) QuerySandboxTimeRange(ctx context.Context, sandboxID string, teamID string) (time.Time, time.Time, error) {
+	// we can't use time.Time, because the time is Datetime64
```
dobrac: what does this comment mean? (because right underneath is used the time.Time)

packages/clickhouse/pkg/metrics.go
```diff
@@ -64,3 +67,84 @@ func (c *Client) QueryLatestMetrics(ctx context.Context, sandboxIDs []string, te
...
+
+	err := c.conn.QueryRow(ctx, sandboxMetricsTimeRangeSelectQuery,
+		clickhouse.Named("sandbox_id", sandboxID),
+		clickhouse.Named("team_id", teamID),
+	).Scan(&start, &end)
+	if err != nil {
+		return time.Time{}, time.Time{}, fmt.Errorf("query time range: %w", err)
+	}
+
+	fmt.Printf("start: %v, end: %v\n", start, end)
```
dobrac:
```suggestion
```

packages/clickhouse/pkg/metrics.go
```diff
@@ -64,3 +67,84 @@ func (c *Client) QueryLatestMetrics(ctx context.Context, sandboxIDs []string, te
...
 	return out, rows.Err()
 }
+
+const sandboxMetricsTimeRangeSelectQuery = `
+    SELECT
+        min(TimeUnix) AS start_time,
+        max(TimeUnix) AS end_time
+    FROM metrics_gauge m
+    WHERE Attributes['sandbox_id'] = 'iawmddux18qj3m0ifwfop'
+      AND Attributes['team_id'] = '460355b3-4f64-48f9-9a16-4442817f79f5'
```
dobrac:
```suggestion
    WHERE Attributes['sandbox_id'] = {sandbox_id:String}
      AND Attributes['team_id'] = {team_id:String}
```

packages/clickhouse/pkg/metrics.go
jakubno: Old comment

packages/api/internal/handlers/sandbox_metrics.go
jakubno: We may change it once we know how exactly will we use it and what's exactly needed

dobrac (approved):

PR #868, opened 2025-07-12
Fix race condition for sandbox resume

jakubno:
# Description

Fix race condition when 2 requests are send such for the later requests `a.orchestrator.GetSandbox(sandboxID)`(`api/internal/handlers/sandbox_resume.go:77`)  is called before the first sandbox called is satisfied, but `orchestrator.CreateSandbox` (`api/internal/handlers/sandbox.go:41`) is called after that.

There's a DB call in between those so it isn't that hard to reproduce


packages/api/internal/cache/instance/reservation.go
```diff
@@ -93,13 +93,18 @@ func (c *InstanceCache) Reserve(instanceID string, team uuid.UUID, limit int64)
 		return nil, &ErrSandboxLimitExceeded{teamID: team.String()}
 	}
 
-	inserted := c.reservations.insertIfAbsent(instanceID, team)
-	if !inserted {
+	if _, ok := ids[instanceID]; ok {
```
dobrac, line 97: if I understand it correctly, this works only because it also includes the `c.list(team)` sandboxes?

If so, could you please add comment to the code here?

packages/api/internal/cache/instance/reservation.go
```diff
@@ -93,13 +93,18 @@ func (c *InstanceCache) Reserve(instanceID string, team uuid.UUID, limit int64)
...
 		return nil, &ErrAlreadyBeingStarted{
 			sandboxID: instanceID,
 		}
 	}
 
+	inserted := c.reservations.insertIfAbsent(instanceID, team)
+	if !inserted {
+		// This shouldn't happen
+		return nil, &ErrAlreadyBeingStarted{}
```
dobrac: should you also return the sandboxID like in the previous if?

packages/api/internal/cache/instance/reservation.go
dobrac, line 97: would any integration test help to simulate this? Would be nice to catch similar racy errors in the future

dobrac (changes requested):
note: I think ideally we should make all sandbox lifecycle operations sturdy by a global locking mechanism

jakubno:
> note: I think ideally we should make all sandbox lifecycle operations sturdy by a global locking mechanism

We plan this with introducing some distributed storage for saving this info

packages/api/internal/cache/instance/reservation.go
jakubno: It will be necessarily flaky, as you have to basically hit the spot database call (the rest is super short in comparison). And in integration test we run the database locally with almost no data, so I expect the response to be super fast as well. I will give it a try though

jakubno:
The ideal case here would be to "merge" those requests and return success on both, but resume the sandbox only once

packages/api/internal/cache/instance/reservation.go
jakubno: I tried, but it's effectiveness would be close to 0, so I removed it. There's unit test for the reservation at least. As the reservation is part of the actual create call, it should suffice

dobrac: ok, good then

PR #870, opened 2025-07-14
Return traffic domain for sandboxes running in cluster

sitole:
This change allows future use in the SDK/CLI that can generate proper URLs pointing to a target that is listening for sandbox traffic.

The domain return field is optional and will be back to our current production domain then.


spec/openapi.yml
```diff
@@ -250,6 +250,10 @@ components:
         envdAccessToken:
           type: string
           description: Access token used for envd communication
+        domain:
+          type: string
+          nullable: true
```
jakubno, line 256: is there a reason why would we not want to send it for our cluster?

packages/shared/pkg/schema/cluster.go
```diff
@@ -30,6 +30,8 @@ func (Cluster) Fields() []ent.Field {
 			NotEmpty().
 			Sensitive().
 			SchemaType(map[string]string{dialect.Postgres: "text"}),
+		field.String("sandbox_proxy_domain").
```
jakubno, line 33: isn't it just cluster domain? Maybe builds/logs may need to use this domain

packages/api/internal/orchestrator/create_instance.go
```diff
@@ -105,6 +105,20 @@ func (o *Orchestrator) CreateSandbox(
...
+			return nil, &api.APIError{
+				Code:      http.StatusInternalServerError,
+				ClientMsg: "Error while looking for sandbox cluster information",
+				Err:       fmt.Errorf("cannot access cluster %s associated with team id %s that spawned sandbox %s", *team.Team.ClusterID, team.Team.ID, sandboxID),
+			}
+		}
+
+		sbxDomain = cluster.SandboxDomain
+	}
+
```
sitole, line 121: Moved here from sandbox create handler, because we are now just re-using the sbx api response in this handler and not creating it again from received values.

https://github.com/e2b-dev/infra/pull/872

packages/api/internal/handlers/sandbox_get.go
```diff
@@ -95,6 +108,7 @@ func (a *APIStore) GetSandboxesSandboxID(c *gin.Context, id string) {
 		State:           api.Paused,
 		EnvdVersion:     lastSnapshot.EnvBuild.EnvdVersion,
 		EnvdAccessToken: sbxAccessToken,
+		Domain:          nil,
```
jakubno, line 111: why is it nil in here?

spec/openapi.yml
sitole: There is not. We just need to add a new environment with that.

packages/shared/pkg/schema/cluster.go
sitole: It may be a different domain, for example, if sandboxes should be available only via the internal network, you will return something different than for "cluster domain".

For logs and all stuff, we already have an endpoint column that should point to the edge API that will return this type of stuff.

packages/api/internal/handlers/sandbox_get.go
sitole: Is it correct to return the domain for a sandbox that does not run, so you technically don't know what URL you should return? Looks to me the same as previously, did not return client ID here, because in this state, you don't know on what node the sandbox was running. 

If we decide that we want to always return the domain, then yes, we don't have any other way than to return the cluster domain/default domain here.

jakubno (approved):

PR #872, opened 2025-07-14
Migration step to client id const

jakubno:
# Description

Return const node id when creating sandbox
Return the "used" client id when getting/listing


sitole (approved):
LGTM, will followup with client id issue in orchestrators sync (https://linear.app/e2b/issue/E2B-2702/remove-nodeid-as-clientid-during-after-api-restart-sandboxes)

PR #873, opened 2025-07-14
Make client ID deprecated in API openapi schema

sitole:


ValentaTomas:
lgtm

Also we still have it exposed in the CLI and before we start removing this (especially from the sdk.getHost) we should modify proxies to also accept an URL without the client ID.

Screenshot 2025-07-14 at 16 06 04

jakubno:
We know about this, but we just want to make sure, this won't be used in code as there are two similar variables and it was easy to get it wrong

jakubno (approved):

PR #874, opened 2025-07-14
Kill and wait for FC process exit before killing the parent unshare process

ValentaTomas:
There seems to be a few bugs related to the sandbox cleanup that results in unexpected errors.
Both (io uring, epoll) might be caused by a bad cleanup order in the orchestrator, resulting in hanging syscalls.

This PR adds killing of the FC (process) in addition to killing the FC's parent unshare process. The unshare process propagates the SIGKILL, but there might be some problem with synchronizing the kill of unshare (and the followup cleanup of network ns, etc.) and the FC exit.


packages/orchestrator/internal/sandbox/fc/process.go
```diff
@@ -422,14 +429,90 @@ func (p *Process) Pid() (int, error) {
...
+
 func (p *Process) Stop() error {
 	if p.cmd.Process == nil {
 		return fmt.Errorf("fc process not started")
 	}
 
-	err := p.cmd.Process.Kill()
+	err := killFirecrackerProcess(p.cmd.Process.Pid)
+	if err != nil {
+		zap.L().Warn("failed to kill firecracker process before killing unshare process", zap.Error(err), zap.String("sandbox.id", p.files.SandboxID))
```
jakubno:
```suggestion
		zap.L().Warn("failed to kill firecracker process before killing unshare process", zap.Error(err), logger.WithSandboxID(p.files.SandboxID))
```

packages/orchestrator/internal/sandbox/fc/process.go
```diff
@@ -422,14 +429,90 @@ func (p *Process) Pid() (int, error) {
...
+	}
+
+	// Send SIGKILL to the unshare process to make sure the parent of the FC process is killed.
+	err = p.cmd.Process.Kill()
 	if err != nil {
-		zap.L().Warn("failed to send KILL to FC process", zap.Error(err))
+		zap.L().Warn("failed to send KILL to unshare process (parent of the FC process)", zap.Error(err))
```
jakubno:
```suggestion
		zap.L().Warn("failed to send KILL to unshare process (parent of the FC process)", zap.Error(err), logger.WithSandboxID(p.files.SandboxID))
```

packages/orchestrator/internal/sandbox/fc/process.go
```diff
@@ -422,14 +429,90 @@ func (p *Process) Pid() (int, error) {
...
+// killFirecrackerProcess tries to find a firecracker process in the process tree to stop it before we start killing the parent.
+func killFirecrackerProcess(unsharePid int) error {
+	fc, err := findFirecrackerProcess(int32(unsharePid))
+	if err != nil {
+		return fmt.Errorf("cannot find firecracker process: %w", err)
+	}
+
+	// Send SIGKILL to the firecracker process to make sure it is killed before the unshare process, nbd, and uffd cleanup.
+	// Because of our unshare (-p, --kill-child) setup, other signals don't propagate properly.
+	err = fc.Kill()
```
jakubno: Don't we want to try to send sigterm only in here? wait a little and then sigkill as a last resort? especially if there's a some issue with resource release

packages/orchestrator/internal/sandbox/fc/process.go
```diff
@@ -422,14 +429,90 @@ func (p *Process) Pid() (int, error) {
...
+	return nil
+}
+
 func (p *Process) Stop() error {
 	if p.cmd.Process == nil {
 		return fmt.Errorf("fc process not started")
 	}
 
-	err := p.cmd.Process.Kill()
+	err := killFirecrackerProcess(p.cmd.Process.Pid)
```
dobrac: why not use pgroup kill on unshare instead? (https://github.com/e2b-dev/infra/pull/326 - I don't remember why this PR was reverted)

The pgid kill should kill all child processed (fork), and shouldn't rely on the unshare -kill-child parameter.

Also, why not use `--mount-proc` in the unshare command, or why are we even using `--fork`?

packages/orchestrator/internal/sandbox/fc/process.go
```diff
@@ -422,14 +429,90 @@ func (p *Process) Pid() (int, error) {
...
+	}
+
+	queue := []*gopsutil.Process{proc}
+
+	for len(queue) > 0 {
+		current := queue[0]
+		queue = queue[1:]
+
+		name, _ := current.Name()
+		if name == "firecracker" {
```
dobrac: this is dependent on the binary name, might break in the future

packages/orchestrator/internal/sandbox/fc/process.go
```diff
@@ -422,14 +429,90 @@ func (p *Process) Pid() (int, error) {
...
+	}
+
+	return nil, fmt.Errorf("firecracker not found")
+}
+
+// waitForFirecrackerExit waits for the firecracker process to exit.
+func waitForFirecrackerExit(fc *gopsutil.Process) error {
+	timeout := time.After(waitForFirecrackerExitTimeout)
+	// It usually took 50-75ms for the FC to stop being in "running" state after SIGKILL.
+	tick := time.Tick(waitForFirecrackerExitTick)
```
dobrac: this adds delay of 25ms (waitForFirecrackerExitTick) that will be there always, regardless of the FC process status

dobrac (changes requested):
I would be for implementing this more properly, focusing on the unshare command and kill, rather than doing fix that might be flaky in the future (different process name, more processes). Or, that might just hide the problem so it would happen "less" often, but not fixing it - let's say the kill on unshare FC child times out (250ms), then we're in the same situation as we're right now

packages/orchestrator/internal/sandbox/fc/process.go
ValentaTomas: SIGTERM has no effect, probably because of the propagation via the unshare. I saw SIGKILL used for FC cleanup in other repos, and in general it should not get stuck on syscalls if we cleanup the FC process first, regardless of the way.

Would try the SIGTERM and reworking unshare if this does not help, but I'm starting with a minimal change that should not affect anything else.

packages/orchestrator/internal/sandbox/fc/process.go
ValentaTomas: Yes, but for now, for the actual kill, I would be ok with that until we rework it. We can tighten the loop alternatively.

packages/orchestrator/internal/sandbox/fc/process.go
ValentaTomas: Valid, might want to link it with the fc name passed to the orchestrator.

packages/orchestrator/internal/sandbox/fc/process.go
ValentaTomas: > why not use pgroup kill on unshare instead? (https://github.com/e2b-dev/infra/pull/326 - I don't remember why this PR was reverted)
I think it changed nothing with the problem it was trying to fix.

We use fork, because it is necessary for `--kill-child`, and kill child because it seems to be the recommended way for killing the whole process tree which we want to do.

dobrac: If you don’t use fork, you don’t need kill child because no child process is spawned (assuming you also don’t use -p)

ValentaTomas: Sounds actually good, I'll explore configuration without fork.

packages/orchestrator/internal/sandbox/fc/process.go
ValentaTomas: Reworked the unshare to allow this.

packages/orchestrator/internal/sandbox/fc/process.go
ValentaTomas: Removed this search.

packages/orchestrator/internal/sandbox/fc/process.go
ValentaTomas: Removed this kind of wait for FC exit.

packages/orchestrator/internal/sandbox/fc/process.go
ValentaTomas: Pushed latest changes without the fork and PID for now.

ValentaTomas:
Reworked the hotfix.

ValentaTomas:
I'm also thinking of cleaning the `.Exit` as right now it can return error, but it seems there several places it is called.

packages/orchestrator/internal/sandbox/fc/process.go
```diff
@@ -427,11 +428,27 @@ func (p *Process) Stop() error {
...
+		// Wait 10 sec for the FC process to exit, if it doesn't send SIGKILL.
+		case <-time.After(10 * time.Second):
+			err := p.cmd.Process.Kill()
+			if err != nil {
+				zap.L().Warn("failed to send SIGKILL to fc process", zap.Error(err), logger.WithSandboxID(p.files.SandboxID))
+			} else {
+				zap.L().Info("sent SIGKILL to fc process because it was not responding to SIGTERM for 10 seconds", logger.WithSandboxID(p.files.SandboxID))
+			}
+		// If the FC process exited, we can return.
+		case <-p.Exit:
```
dobrac: as the p.Exit is a channel that is read somewhere else and contains an error, I don't think you can rely on it here

(there actually might be other places where it's incorrect - as per you message)

ValentaTomas: Yeah, I'm probably multiplex it or check using some other signalization.

packages/orchestrator/internal/sandbox/fc/process.go
```diff
@@ -427,11 +428,27 @@ func (p *Process) Stop() error {
...
 	}
 
-	err := p.cmd.Process.Kill()
+	err := p.cmd.Process.Signal(syscall.SIGTERM)
 	if err != nil {
-		zap.L().Warn("failed to send KILL to FC process", zap.Error(err))
+		zap.L().Warn("failed to send SIGTERM to fc process", zap.Error(err), logger.WithSandboxID(p.files.SandboxID))
 	}
 
+	go func() {
```
jakubno, line 436: what's the reason for running this in goroutine? also 10seconds is probably unnecessarily too long

ValentaTomas: I looked around and for example systemd uses 10 sec delay of SIGKILL after SIGTERM.
There is no overall guide for how long this should be.

If we have this outside of the goroutine it would be in the direct path of kill, so it would slow the kill, especially if we want to go with reasonable delay.

ValentaTomas: And although we can theoretically send SIGKILL right after SIGTERM that seems pretty rude.

packages/orchestrator/internal/sandbox/fc/process.go
ValentaTomas: Added

ValentaTomas:
Reworked the Exit signalization as it was used in several places even before, and was not handling that.

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -874,7 +880,12 @@ func (s *Sandbox) WaitForEnvd(
 			syncCancel(fmt.Errorf("syncing took too long"))
 		case <-syncCtx.Done():
 			return
-		case err := <-s.process.Exit:
+		case <-s.process.Exit.Done:
+			_, err := s.process.Exit.Result()
+			if err == nil {
+				return
+			}
```
dobrac: this should not be here, any exit during waiting for envd is an error

dobrac (changes requested):

packages/orchestrator/internal/sandbox/sandbox.go
ValentaTomas: Got it, removed.

dobrac (approved):

PR #880, opened 2025-07-15
Improve build logs with structured messages

dobrac:
Improve build logs with structured messages. Adds `logEntries` field to the template status endpoint which contains the following entry format:
```
{
    "level": "info",
    "message": "Waiting for template to be ready",
    "timestamp": "2025-07-16T10:04:58.564568281Z"
}
```

It also adds a query parameter `level` to filter messages only for the certain level.

- [x] Depends on https://github.com/e2b-dev/infra/pull/854


packages/orchestrator/template-manager.proto
```diff
@@ -47,10 +48,18 @@ message TemplateCreateRequest {
...
+    Info = 1;
+    Warn = 2;
+    Error = 3;
+}
+
 message TemplateStatusRequest {
   string templateID = 1;
   string buildID = 2;
   optional int32 offset = 3;
+  optional LogLevel level = 4;
```
sitole, line 62: Maybe name it something like `minLevel` to make obvious it's used for filtration here?

spec/openapi-edge.yml
```diff
@@ -199,16 +199,49 @@ components:
...
         - logs
+        - logEntries
       properties:
         logs:
           default: []
           description: Build logs
           type: array
           items:
             type: string
+        logEntries:
```
sitole, line 239: It's not implemented yet?

packages/api/internal/template-manager/logs/edge_provider.go
```diff
@@ -28,5 +29,48 @@ func (c *ClusterPlacementProvider) GetLogs(ctx context.Context, templateID strin
 		return nil, errors.New("failed to get build logs in template manager")
 	}
 
-	return res.JSON200.Logs, nil
+	if res.JSON200 == nil {
```
sitole: Its unexpected response from edge api

packages/api/internal/template-manager/logs/edge_provider.go
```diff
@@ -28,5 +29,48 @@ func (c *ClusterPlacementProvider) GetLogs(ctx context.Context, templateID strin
...
+}
+
+func edgeLevelToAPILevel(level edgeapi.LogLevel) api.LogLevel {
+	if apiLevel, ok := edgeToAPILevelMap[level]; ok {
+		return apiLevel
+	}
+	return api.LogLevelInfo
+}
+
+func apiLevelToEdgeLevel(level *api.LogLevel) *edgeapi.LogLevel {
```
sitole, line 68: Cannot we do also map instead of this?

packages/api/internal/template-manager/logs/edge_provider.go
```diff
@@ -28,5 +29,48 @@ func (c *ClusterPlacementProvider) GetLogs(ctx context.Context, templateID strin
...
 	}
 
-	return res.JSON200.Logs, nil
+	if res.JSON200 == nil {
+		zap.L().Error("unexpected response from template manager", zap.String("body", string(res.Body)))
+		return nil, errors.New("unexpected response from template manager")
+	}
+
+	logs := make([]api.BuildLogEntry, 0)
+	for _, entry := range res.JSON200.LogEntries {
```
sitole, line 43: This will not work, because log entries is not filled in edge api

packages/orchestrator/internal/template/server/create_template.go
```diff
@@ -59,16 +57,23 @@ func (s *ServerStore) TemplateCreate(ctx context.Context, templateRequest *templ
...
 	if err != nil {
 		return nil, fmt.Errorf("error while creating build cache: %w", err)
 	}
 
+	// Add new core that will log all messages using logger (zap.Logger) to the logs buffer too
+	encoder := zapcore.NewJSONEncoder(zap.NewProductionEncoderConfig())
+	bufferCore := zapcore.NewCore(encoder, logs, zapcore.DebugLevel)
+	core := zapcore.NewTee(bufferCore, s.buildLogger.Core().
+		With([]zap.Field{
+			{Type: zapcore.StringType, Key: "envID", String: metadata.TemplateID},
```
sitole, line 71: Can we use logger helpers for env id and build id here?

dobrac: this logger is external and the field names don't adhere to our internal names

packages/orchestrator/template-manager.proto
dobrac: it's standard to call it just level and treat it as the minimum level

spec/openapi-edge.yml
dobrac: good call, implemented

packages/api/internal/template-manager/logs/edge_provider.go
dobrac: removed

packages/api/internal/template-manager/logs/edge_provider.go
dobrac: I didn't want second map, but otherwise yes

packages/api/internal/template-manager/logs/edge_provider.go
dobrac: should be fixed now with them being implemented

sitole:
> I'm thinking about replacing the enums with strings, so there would be no need for so many converters in the communication

For the rest i like string enums more

packages/api/internal/template-manager/logs/edge_provider.go
sitole: I'm not sure what the correct way is with the HTTP client we are using, but I probably prefer checking if `res.JSON200` is not nil, rather than just checking if the response status code is 200.

dobrac: the check was there before and you've said that it's unexpected 👀 
https://github.com/e2b-dev/infra/pull/880#discussion_r2216409767

sitole:
> > Can we share Loki query logic from the edge api and api?
> 
> totally, do you want me to refactor the whole logic in this PR?

Yep, we can do it here if you want.

packages/api/internal/template-manager/logs/edge_provider.go
sitole: I was just pointing out that it will always unexpectedly fail (because the response api from Edge was not returning log entries at all, because not implemented).

dobrac: added

dobrac:
Refactored the Loki mapping to shared + REST/GRPC enum types handling

PR #882, opened 2025-07-16
Add OTEL collector metrics

jakubno:
# Description

- Add otel collector metrics 
- Set resource names on all metrics
- Fix incorrect name (nomad_client_allocs_cpu_usage -> nomad_client_allocs_cpu_total_percent)

### DIFF
--- a/packages/nomad/otel-collector.hcl
+++ b/packages/nomad/otel-collector.hcl
@@ -44,7 +44,7 @@ job "otel-collector" {
 
       config {
         network_mode = "host"
-        image        = "otel/opentelemetry-collector-contrib:0.123.0"
+        image        = "otel/opentelemetry-collector-contrib:0.130.0"
 
         volumes = [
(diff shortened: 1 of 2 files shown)


packages/nomad/otel-collector.hcl
```diff
@@ -44,7 +44,7 @@ job "otel-collector" {
 
       config {
         network_mode = "host"
-        image        = "otel/opentelemetry-collector-contrib:0.123.0"
+        image        = "otel/opentelemetry-collector-contrib:0.130.0"
```
jakubno, line 47: I had to update otel collector because of https://github.com/open-telemetry/opentelemetry-collector/issues/12701
Insecure flag was ignored and then the connection to itself was failing due to `authentication handshake failed: tls: first record does not look like a TLS handshake`

dobrac (approved):

PR #883, opened 2025-07-18
Fix envd resolving path

jakubno:
# Description

Return whole path not only relative one

### DIFF
--- a/packages/envd/internal/services/filesystem/dir.go
+++ b/packages/envd/internal/services/filesystem/dir.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"os"
-	"os/user"
 	"path"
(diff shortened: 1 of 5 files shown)


packages/envd/internal/services/filesystem/dir.go
```diff
@@ -27,7 +26,14 @@ func (Service) ListDir(ctx context.Context, req *connect.Request[rpc.ListDirRequ
 	}
 
 	requestedPath := req.Msg.GetPath()
-	resolvedPath, err := resolvePath(requestedPath, u)
+
+	// Expand the path
```
dobrac: this comments is not helpful, if you want to describe the issue, you could mention that you convert all symlinks to absolute paths

PR #885, opened 2025-07-18
Build layers caching improvements and fixes

dobrac:
This PR changes

- Make ENV, ARG, WORKDIR, USER resolved using bash and persisted as a separate layer
This fixes bug where you want to define e.g. PATH using `PATH=$PATH:/my/path`
- **Fixes CPU/Memory change when the cached layer is cached with a different definitions**
- Refactors the commands definitions to separate files

- [x] Depends on https://github.com/e2b-dev/infra/pull/880


packages/orchestrator/internal/template/build/apply_command.go
```diff
@@ -78,20 +35,28 @@ func (b *Builder) applyCommand(
...
 
+	cmdMetadata, err := b.readCommandMetadata(ctx, sbx.Metadata.Config.SandboxId, baseCmdMetadata)
+	if err != nil {
+		return fmt.Errorf("failed to read command metadata: %w", err)
+	}
+
+	telemetry.ReportEvent(ctx, "command metadata",
+		attribute.String("user", cmdMetadata.User),
+		attribute.String("workdir", utils.Sprintp(cmdMetadata.WorkDir)),
+		attribute.String("env_vars", fmt.Sprintf("%v", cmdMetadata.EnvVars)),
```
0div: not sure we want to log env vars if they can potentially contain privileged information

packages/orchestrator/internal/template/build/apply_command.go
```diff
@@ -227,41 +192,201 @@ fi
...
 				User:    cmdMetadata.User,
 				EnvVars: cmdMetadata.EnvVars,
 			},
 		)
+		if err != nil {
+			return fmt.Errorf("failed to create workdir in sandbox: %w", err)
+		}
+
+		return b.saveWorkdir(ctx, sbx.Metadata.Config.SandboxId, cmdMetadata, workdirArg)
+	case "ENV", "ARG":
```
0div: Should we not stick to the OCI spec stating that `ARG` defines variables that are only available during the image build process and not in the final environment?

packages/orchestrator/internal/template/build/apply_command.go
```diff
@@ -227,41 +192,201 @@ fi
...
+	}
+
+	return sandboxtools.CommandMetadata{
+		User:    user,
+		WorkDir: workdir,
+		EnvVars: envVars,
+	}, nil
+}
+
+func (b *Builder) saveEnv(
```
0div: high-level for my understanding: besides workdir, why are we saving env and user as files in a metadata folder when we could set them as the entity they represent? is this an intermediary step?

0div (dismissed):

packages/orchestrator/internal/template/build/apply_command.go
dobrac: good point, will remove

packages/orchestrator/internal/template/build/apply_command.go
dobrac: none of the variables are now in the final environment, but good point, maybe we should rethink that

packages/orchestrator/internal/template/build/apply_command.go
dobrac: the reason for this PR is to allow `PATH=$PATH:/my/path` to be resolved correctly. For that, I need to execute the $PATH in the bash environment (sandbox)

Now, because the layers are cached, I don't know from which point the build will resume. And because it might resume from, let's say, the middle, after the $PATH is executed, I need to persist somewhere the information about expanded $PATH at that particular moment. I can't do it retrospectively just in memory (unfortunately), because i would basically need to re-run the $PATH in all relevant layers

does this explanation help?

packages/orchestrator/internal/template/build/apply_command.go
dobrac: (let's do it in another PR though if necessary)

sitole (commented):
After checking in, the logic for writing and loading temporary files is obvious, but it would be helpful to add comments that clarify the reasoning behind these actions so next time you don't need to read all the code just to notify the issue with inconsistency during cached layers.

dobrac:
> After checking in, the logic for writing and loading temporary files is obvious, but it would be helpful to add comments that clarify the reasoning behind these actions so next time you don't need to read all the code just to notify the issue with inconsistency during cached layers.

Makes sense, I'll add comments explaining it more

jakubno (changes requested):
Can we save cmd metadata (user, cwd, envs) to LayerMetadata instead of saving it to disk?

dobrac:
> Can we save cmd metadata (user, cwd, envs) to LayerMetadata instead of saving it to disk?

Yes, I can do that, will redo

PR #886, opened 2025-07-18
Create ClickHouse table and queries for sandbox events complete with batcher

0div:
This PR is part of a broader effort to store events around sandboxes, the first steps being:
- [x] create makefile rule to build the clickhouse-migrator locally for integration testing
- [x] integrate the clickhouse pkg in orchestrator 
- [x] define the general sandbox events table in ClickHouse via migration files 
- [x] add new methods to the `Clickhouse` client interface to insert and select these sandbox events 
- [x] start using these methods in orchestrator to track sandbox lifecycle events. 
- [x] create a client-side query batcher


packages/clickhouse/migrations/20250520223340_add_sandbox_events.sql
```diff
@@ -0,0 +1,18 @@
...
+CREATE TABLE sandbox_events (
+    timestamp DateTime64(9),
+    sandbox_id UUID,
+    sandbox_execution_id UUID,
+    sandbox_template_id UUID,
+    sandbox_team_id UUID,
+    event_category String,
+    event_label String,
+    event_data Nullable(JSON)
+) ENGINE = MergeTree ORDER BY (timestamp, sandbox_id, event_category, event_label);
```
ValentaTomas: Will we need some partitioning like in (https://github.com/e2b-dev/infra/blob/main/packages/clickhouse/migrations/20250521131545_add_metrics_local.sql#L34)?

Also, the metrics gauge is a distributed table here (https://github.com/e2b-dev/infra/blob/main/packages/clickhouse/migrations/20250521131546_add_metrics.sql#L4), do we need to make the events the same?

packages/clickhouse/migrations/20250520223340_add_sandbox_events.sql
```diff
@@ -0,0 +1,18 @@
+-- +goose Up
+-- +goose StatementBegin
+CREATE TABLE sandbox_events (
+    timestamp DateTime64(9),
+    sandbox_id UUID,
```
ValentaTomas: I don't think sandbox ID is UUID.

ValentaTomas: The same for template ID.

packages/clickhouse/migrations/20250520223340_add_sandbox_events.sql
```diff
@@ -0,0 +1,18 @@
...
+    sandbox_execution_id UUID,
+    sandbox_template_id UUID,
```
ValentaTomas: We might be able to add buildID here as it identifies the exact build.

packages/clickhouse/pkg/clickhouse.go
```diff
@@ -13,16 +13,20 @@ type Clickhouse interface {
...
 
 	// Metrics queries
-	QuerySandboxTimeRange(ctx context.Context, sandboxID string, teamID string) (start time.Time, end time.Time, err error)
-	QuerySandboxMetrics(ctx context.Context, sandboxID string, teamID string, start time.Time, end time.Time, step time.Duration) ([]Metrics, error)
+	QuerySandboxTimeRange(ctx context.Context, sandboxID, teamID string) (start time.Time, end time.Time, err error)
+	QuerySandboxMetrics(ctx context.Context, sandboxID, teamID string, start time.Time, end time.Time, step time.Duration) ([]Metrics, error)
 	QueryLatestMetrics(ctx context.Context, sandboxIDs []string, teamID string) ([]Metrics, error)
+
+	// Events queries
+	QueryLatestSandboxEvent(ctx context.Context, sandboxID string, limit, offset int) ([]SandboxEvent, error)
```
ValentaTomas: Would name this just `QuerySandboxEvents`—the offset and limit (would first put the offset) should define if it is last, etc.

packages/clickhouse/pkg/clickhouse.go
```diff
@@ -13,16 +13,20 @@ type Clickhouse interface {
...
+	InsertSandboxEvent(ctx context.Context, event SandboxEvent) error
```
ValentaTomas: Curious—where is the insert for metrics and why it is also not here?

packages/clickhouse/pkg/mock.go
```diff
@@ -7,7 +7,7 @@ import (
 
 type NoopClient struct{}
 
-func NewNoopClient() *NoopClient {
+func NewNoopClient() Clickhouse {
```
ValentaTomas: Would keep the NoopClient return to be as specific as possible. It was a little weird when I initially saw that, but as a Go recommendation it makes sense.

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -147,6 +169,21 @@ func (s *server) Update(ctx context.Context, req *orchestrator.SandboxUpdateRequ
...
 	item.EndAt = req.EndTime.AsTime()
 
+	go func() {
+		err := s.clickhouseClient.InsertSandboxEvent(ctx, clickhouse.SandboxEvent{
+			Timestamp:          time.Now().UTC(),
+			SandboxID:          item.Config.SandboxId,
+			SandboxTemplateID:  item.Config.TemplateId,
+			SandboxTeamID:      item.Config.TeamId,
+			SandboxExecutionID: item.Config.ExecutionId,
+			EventCategory:      "lifecycle",
```
ValentaTomas: Please constantize/make enum. Same for event label.

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -124,6 +125,27 @@ func (s *server) Create(ctxConn context.Context, req *orchestrator.SandboxCreate
...
 
+	// If there is no snapshot flag in the request, this should mean the sandbox is being created for the first time.
+	label := "resume"
+	if !req.Sandbox.Snapshot {
+		label = "create"
+	}
+
+	go func() {
+		err := s.clickhouseClient.InsertSandboxEvent(ctx, clickhouse.SandboxEvent{
+			Timestamp:          time.Now().UTC(),
```
ValentaTomas: Curious—is using `time.Now().UTC()` here vs using just `time.Now()` in `packages/clickhouse/pkg/sandbox_events.go` intended?

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -124,6 +125,27 @@ func (s *server) Create(ctxConn context.Context, req *orchestrator.SandboxCreate
 		sbxlogger.E(sbx).Info("Sandbox killed")
 	}()
 
+	// If there is no snapshot flag in the request, this should mean the sandbox is being created for the first time.
+	label := "resume"
+	if !req.Sandbox.Snapshot {
+		label = "create"
+	}
+
+	go func() {
```
ValentaTomas: Are we expecting some problems with congestion/buffering?

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -147,6 +169,21 @@ func (s *server) Update(ctx context.Context, req *orchestrator.SandboxUpdateRequ
 
 	item.EndAt = req.EndTime.AsTime()
```
ValentaTomas, line 199: For the sandbox start/end event, I think we might want to capture the data from the sandbox `StartTime`, `EndTime`.
Though these can be wrong right at the start by a few milliseconds, so maybe just having them in the `eventData` makes more sense.

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -147,6 +169,21 @@ func (s *server) Update(ctx context.Context, req *orchestrator.SandboxUpdateRequ
...
 
+	go func() {
```
ValentaTomas: It is possible that update will update a different values on sandbox than timeout in the future, so let's guard agains that.

ValentaTomas: We can also add the info about the timeout here to the event.

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -214,6 +251,21 @@ func (s *server) Delete(ctxConn context.Context, in *orchestrator.SandboxDeleteR
 		sbxlogger.I(sbx).Error("error stopping sandbox", logger.WithSandboxID(in.SandboxId), zap.Error(err))
 	}
 
+	go func() {
```
ValentaTomas: Although the `Delete` is called both if a customer does request to kill the sandbox and if it times out, the sandbox itself can (for some reasons, generally it should not) die and to get that we can "wait" until this (https://github.com/e2b-dev/infra/blob/create-clickhouse-table-for-sandbox-events-e2b-2699/packages/orchestrator/internal/server/sandboxes.go#L97) method passes.

packages/clickhouse/pkg/clickhouse.go
jakubno: We use otel for recording and then otel collector for inserting metrics

packages/clickhouse/pkg/sandbox_events.go
```diff
@@ -0,0 +1,92 @@
...
+    ?,
+    ?,
+    ?,
+    ?,
+    ?,
+    ?
+)`
+
+func (c *Client) InsertSandboxEvent(ctx context.Context, event SandboxEvent) error {
+	return c.conn.Exec(ctx, insertSandboxEventQuery,
```
jakubno: We don't want to send per event, this has to be batched. Clickhouse isn't made to ingest single entries at scale

packages/clickhouse/pkg/sandbox_events.go
```diff
@@ -0,0 +1,92 @@
...
+	SandboxID          string    `ch:"sandbox_id"`
+	SandboxExecutionID string    `ch:"sandbox_execution_id"`
+	SandboxTemplateID  string    `ch:"sandbox_template_id"`
+	SandboxTeamID      string    `ch:"sandbox_team_id"`
+	EventCategory      string    `ch:"event_category"`
+	EventLabel         string    `ch:"event_label"`
+	EventData          *string   `ch:"event_data"`
+}
+
+const latestSandboxEventSelectQuery = `
```
jakubno: can you use the named params?

packages/clickhouse/pkg/sandbox_events.go
```diff
@@ -0,0 +1,92 @@
...
+		if err := rows.ScanStruct(&m); err != nil {
+			return nil, fmt.Errorf("error scaning SandboxEvent: %w", err)
+		}
+		out = append(out, m)
+	}
+
+	return out, rows.Err()
+}
+
+const insertSandboxEventQuery = `
```
jakubno: also here, can you use named params?

packages/nomad/main.tf
```diff
@@ -403,6 +403,7 @@ locals {
     otel_collector_grpc_endpoint = "localhost:${var.otel_collector_grpc_port}"
     allow_sandbox_internet       = var.allow_sandbox_internet
     launch_darkly_api_key        = trimspace(data.google_secret_manager_secret_version.launch_darkly_api_key.secret_data)
+    clickhouse_connection_string = var.clickhouse_server_count > 0 ? "clickhouse://${var.clickhouse_username}:${random_password.clickhouse_password.result}@clickhouse.service.consul:${var.clickhouse_server_port.port}/${var.clickhouse_database}" : 0
```
jakubno: do you really want to return `0` as `clickhouse_connection_string` if there's no clickhouse?

packages/nomad/main.tf
```diff
@@ -486,6 +487,7 @@ resource "nomad_job" "template_manager" {
     logs_collector_public_ip     = var.logs_proxy_address
     orchestrator_services        = "template-manager"
     allow_sandbox_internet       = var.allow_sandbox_internet
+    clickhouse_connection_string = var.clickhouse_server_count > 0 ? "clickhouse://${var.clickhouse_username}:${random_password.clickhouse_password.result}@clickhouse.service.consul:${var.clickhouse_server_port.port}/${var.clickhouse_database}" : 0
```
jakubno: do you really want to return `0` as `clickhouse_connection_string` if there's no clickhouse?

packages/clickhouse/migrations/20250520223340_add_sandbox_events.sql
jakubno: This ordering won't help you with querying at all. This is primary index (data are stored based on this ordering). You probably want to have sandbox_id first. Can't tell more, what do you expect the event_category and labels to be?

packages/clickhouse/migrations/20250520223340_add_sandbox_events.sql
```diff
@@ -0,0 +1,18 @@
+-- +goose Up
+-- +goose StatementBegin
+CREATE TABLE sandbox_events (
+    timestamp DateTime64(9),
+    sandbox_id UUID,
+    sandbox_execution_id UUID,
```
jakubno: does everything has to be prefixed with `sandbox_`?

packages/orchestrator/main.go
```diff
@@ -246,12 +247,24 @@ func run(port, proxyPort uint) (success bool) {
 		zap.L().Fatal("failed to create feature flags client", zap.Error(err))
 	}
 
+	var clickhouseClient clickhouse.Clickhouse
+	clickhouseConnectionString := os.Getenv("CLICKHOUSE_CONNECTION_STRING")
```
jakubno, line 272: can we use feature flags for turning collecting events on and off if something goes wrong?

jakubno (changes requested):

packages/clickhouse/migrations/20250520223340_add_sandbox_events.sql
0div: it helps disambiguate during joins and most importantly subqueries.

packages/orchestrator/internal/server/sandboxes.go
0div: If you mean integer enums, I'd rather not use those when interacting with a persistence layer bc it limits flexibility and schema evolution right off the bat for limited benefits. If you meant to store them as string in the end, sure.

packages/orchestrator/main.go
0div: i tried using launch darkly but it was missing SDK key, is it documented somewhere how to use it in our infra?

packages/nomad/main.tf
0div: i mistakenly thought this was a wanted pattern across our IaC (e.g. [here](https://github.com/e2b-dev/infra/blob/main/packages/nomad/main.tf#L94))

packages/nomad/main.tf
0div: i mistakenly thought this was a wanted pattern across our IaC (e.g. [here](https://github.com/e2b-dev/infra/blob/main/packages/nomad/main.tf#L94))

packages/clickhouse/pkg/sandbox_events.go
0div: Yep, was planning on adding batching.

packages/clickhouse/migrations/20250520223340_add_sandbox_events.sql
0div: agreed on flipping `timestamp` with `sandbox_id`. We'll likely want to drill down on event category and label when serving information about sandboxes tho.

packages/orchestrator/internal/server/sandboxes.go
0div: made it consistently UTC 👍

packages/orchestrator/internal/server/sandboxes.go
0div: Yeah i'm gonna add batching logic

packages/clickhouse/pkg/mock.go
0div: "accept interfaces, return structs", so goes the idiom. That being said, concretely I want existing code to break when i add new methods and compile, since i want it to be consistent/sealed, in other words i don't want it to be extensible, without it being consistent everywhere. Also with that layer abstraction i want to allow polymorphism and facilitate testing.

packages/nomad/main.tf
jakubno: Can you fix it there as well, I think it should be an empty string

packages/clickhouse/migrations/20250520223340_add_sandbox_events.sql
jakubno: you can always reference the table

0div: that's one school of thought, but the other is not having to do `<table>.*`  and `AS` often and the peace of mind of not having column name collision.

packages/clickhouse/go.mod
```diff
@@ -1,8 +1,6 @@
 module github.com/e2b-dev/infra/packages/clickhouse
 
-go 1.24.0
-
-toolchain go1.24.3
```
jakubno, line 5: why?

packages/clickhouse/Makefile
```diff
@@ -2,6 +2,7 @@ ENV := $(shell cat ../../.last_used_env || echo "not-set")
 -include ../../.env.${ENV}
 
 CLICKHOUSE_MIGRATOR_IMAGE ?= $(GCP_REGION)-docker.pkg.dev/$(GCP_PROJECT_ID)/e2b-orchestration/clickhouse-migrator
+CLICKHOUSE_MIGRATOR_LOCAL_IMAGE ?= e2b-orchestration/clickhouse-migrator
```
jakubno: if you put `CLICKHOUSE_MIGRATOR_IMAGE` into your `.env` it will be used instead of the default one, so there's no need for `build-locally`

packages/nomad/main.tf
```diff
@@ -407,6 +407,7 @@ locals {
     otel_collector_grpc_endpoint = "localhost:${var.otel_collector_grpc_port}"
     allow_sandbox_internet       = var.allow_sandbox_internet
     launch_darkly_api_key        = trimspace(data.google_secret_manager_secret_version.launch_darkly_api_key.secret_data)
+    clickhouse_connection_string = var.clickhouse_server_count > 0 ? "clickhouse://${var.clickhouse_username}:${random_password.clickhouse_password.result}@clickhouse.service.consul:${var.clickhouse_server_port.port}/${var.clickhouse_database}" : ""
```
jakubno, line 411:
```suggestion
    clickhouse_connection_string = local.clickhouse_connection_string
```

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -31,6 +32,7 @@ const (
 func (s *server) Create(ctxConn context.Context, req *orchestrator.SandboxCreateRequest) (*orchestrator.SandboxCreateResponse, error) {
 	ctx, cancel := context.WithTimeoutCause(ctxConn, requestTimeout, fmt.Errorf("request timed out"))
 	defer cancel()
+	fmt.Println("~~~~~cREATE")
```
jakubno:
```suggestion
```

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -122,6 +124,27 @@ func (s *server) Create(ctxConn context.Context, req *orchestrator.SandboxCreate
 		sbxlogger.E(sbx).Info("Sandbox killed")
 	}()
 
+	label := clickhouse.SandboxEventLabelResume
+	if !req.Sandbox.Snapshot {
+		label = clickhouse.SandboxEventLabelCreate
+	}
```
jakubno: I would switch the logic as Create is expected to happen more often

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -122,6 +124,27 @@ func (s *server) Create(ctxConn context.Context, req *orchestrator.SandboxCreate
...
+
+	go func(label clickhouse.SandboxEventLabel) {
```
jakubno: As I mentioned earlier, it has to be batched

packages/orchestrator/internal/server/sandboxes.go
0div: do you mean using the sandbox `StartTime` & `EndTime` respecitvely for `Create` and `Kill` lifecyle event timestamps?

packages/clickhouse/migrations/20250725223330_settings.sql
jakubno, line 1: are you even using string?

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -214,6 +274,31 @@ func (s *server) Delete(ctxConn context.Context, in *orchestrator.SandboxDeleteR
...
+				SandboxTeamID:      sbx.Config.TeamId,
+				SandboxExecutionID: sbx.Config.ExecutionId,
+				EventCategory:      string(clickhouse.SandboxEventCategoryLifecycle),
+				EventLabel:         string(clickhouse.SandboxEventLabelKill),
+			})
+			if err != nil {
+				sbxlogger.I(sbx).Error("error inserting sandbox event during kill", zap.Error(err))
+			}
+		}
+	}()
```
jakubno: I would do the check outside of go routine, if the check is false you don't need to create go routine which is more intensive then bool check (which you will do anyway)

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -143,6 +174,35 @@ func (s *server) Update(ctx context.Context, req *orchestrator.SandboxUpdateRequ
 
 	item.EndAt = req.EndTime.AsTime()
 
+	// TODO: adapt to new types of update events
```
jakubno, line 201: There's TODO:

packages/clickhouse/pkg/sandbox_events.go
```diff
@@ -0,0 +1,188 @@
...
+    ?,
+    ?,
+    ?,
+    ?,
+    ?,
+    ?,
+	?,
+)`
+
+func (c *Client) InsertSandboxEvent(ctx context.Context, event SandboxEvent) error {
```
jakubno: this is still sync

packages/clickhouse/pkg/sandbox_events.go
```diff
@@ -0,0 +1,188 @@
...
+    sandbox_id, 
+    sandbox_execution_id,
+    sandbox_template_id,
+    sandbox_build_id,
+    sandbox_team_id,
+    event_category,
+    event_label,
+    event_data
+)
+SETTINGS async_insert=1, wait_for_async_insert=1
```
jakubno: it's still recommended to do batching on client 
https://kb.altinity.com/altinity-kb-queries-and-syntax/async-inserts/

also `SETTINGS async_insert=1, wait_for_async_insert=1` should be default (it's in clickhouse config)

packages/clickhouse/pkg/sandbox_events.go
```diff
@@ -0,0 +1,188 @@
...
+VALUES (
```
jakubno: please use the named arguments (also in other queries)

jakubno (changes requested):

packages/clickhouse/go.mod
0div: it's remove when i run `go mod tidy` in `packages/clickhouse/`

packages/orchestrator/internal/server/sandboxes.go
0div: AFAIK we only have `set_timeout`, once a new update event is introduced we should adapt event data.

packages/clickhouse/migrations/20250725223330_settings.sql
0div: yeah for now, JSON was causing me serialization problems, idk why bc i set 
```
SET enable_json_type = 1, allow_experimental_json_type = 1, output_format_native_write_json_as_string = 1;
```
i'll try again to make it work

packages/clickhouse/pkg/sandbox_events.go
0div: > We recommend to batch at app/ingestor level because you will have more control and you decouple this responsibility from ClickHouse [...] Async inserts come in handy if you have hundreds or thousands of clients doing small inserts.

How about we monitor to see how this is scaling first? Did you get any sense of it with telemetry, is that currently batching? What are the tradeoffs in terms of our app memory vs deferring to CH buffer and what improvements do we exactly gain?
There is nice stmt batching out of the box with `clickhouse-go` which would come int handy with multi transaction queries within a controllable synchronous scope. 
We could always implement an accumulator and ticker that flushes from our pkg client but do we need that kind of control right now?

packages/clickhouse/pkg/sandbox_events.go
0div: What is the advantage of this? it adds more code complexity for limited improvements in readability for these simple type of queries; also the query builder already knows the types expected.

packages/clickhouse/pkg/sandbox_events.go
0div: >  should be default (it's in clickhouse config)

Where is it configured exactly? I don't see it in config.xml ?

packages/clickhouse/pkg/sandbox_events.go
jakubno: To prevent some unintended mistakes - let's say you will add a new column, then you can easily insert the argument on incorrect position, if the type matches, it can go unnoticed

packages/clickhouse/pkg/sandbox_events.go
jakubno: sorry, it's default, so I didn't add it, my bad

jakubno: If you can do some benchmark which will indicate that it's okay to insert separately I am fine. I really doubt it will scale, because if you do `async_insert=1, wait_for_async_insert=1` you will wait for the response, let's say it will be ~200ms. If I am not mistaken, during this call you are still using the connection -> each orchestrator will use multiple (at least tens) of connections and it will grow with the granularity of the events.

packages/clickhouse/migrations/20250725223330_settings.sql
jakubno: Did you try to use `Map`?

0div: `Map` is nice but i have to specify the type of the key which can be `any`

packages/clickhouse/pkg/sandbox_events.go
0div: it seems like u have to manually convert all the non-string values to string in `clickhouse-go`, eg:
```
"SELECT {num:UInt64}, {str:String}",
  clickhouse.Named("num", "42"),
  clickhouse.Named("str", "hello"),
```
i'm not a fan of having to do this for the sake of not having to spot check argument position.

packages/orchestrator/internal/server/sandboxes.go
ValentaTomas: In the future (with the updatable metadata) we will need to handle this, but for now this should be resolved.

packages/clickhouse/migrations/20250725223330_settings.sql
ValentaTomas: @0div The JSON is ok now?

0div: I'm testing it as we speak, it was tricky to make it work as a `Nullable` type, but `sql.NullString` seems to do the trick.

packages/clickhouse/pkg/sandbox_events.go
ValentaTomas: @0div Please, fix the rest of the issues first. If this is too troublesome now, let's get the core fixed PR to staging now, but then you open a new one with this change.

packages/clickhouse/migrations/20250725223330_settings.sql
ValentaTomas: Great!

packages/clickhouse/pkg/sandbox_events.go
ValentaTomas: If the async insert wait really hogs connections/resources too much, we can for now do without it (I wasn't sure about that).

@jakubno We will see failed inserts in Clickhouse logs with async inserts, without wait, right?

I'm also okay with the client insert—the things I want:
1. For it not to block anything in api/orchestrator
2. To be able to handle the scale *and* if not for us to get error somewhere without crashing the system

packages/clickhouse/migrations/20250725223330_settings.sql
0div: Inserts work but i keep getting 
```
Error returning events for sandboxes by sandbox id: error querying events: clickhouse [Decode]: event_data unsupported JSON serialization version for decode: 18446744073709551615"
```
When selecting, i've spent too much time trying to make it work, i'm going to revert to `Nullable(String)` type for now, until we need to inspect complex JSON objects.

packages/orchestrator/internal/server/sandboxes.go
0div: @ValentaTomas are you suggesting I should also insert the `kill` after that wait ?

packages/orchestrator/internal/server/sandboxes.go
0div: @ValentaTomas reviving this

packages/clickhouse/pkg/sandbox_events.go
jakubno: Async insert is fine, but it isn't solution for spamming clickhouse with single row inserts. Ideally you combine batched inserts with async inserts, that way clickhouse can optimize the inserts the most

packages/clickhouse/pkg/sandbox_events.go
jakubno: Imho it isn't that hard to convert to string, at least you have the control how you convert it, because the conversion has to happen anyway and with that syntax it only calls `String()` method I believe

packages/clickhouse/migrations/20250725223340_add_sandbox_events_local.sql
```diff
@@ -0,0 +1,22 @@
+-- +goose Up
+-- +goose StatementBegin
+CREATE TABLE sandbox_events_local (
```
jakubno, line 3: use codecs to optimize storage + low cardinality for fields as event category

packages/clickhouse/pkg/sandbox_events.go
jakubno: there's 10 `MaxOpenConns` now, so there's no way this will work with production load, you will need to implement batching

packages/clickhouse/pkg/sandbox_events.go
```diff
@@ -0,0 +1,187 @@
...
+    ?,
+    ?,	
+    ?,
+    ?,
+    ?
+)`
+
+func (c *Client) InsertSandboxEvent(ctx context.Context, event SandboxEvent) error {
+	return c.conn.Exec(ctx, insertSandboxEventQueryAsync,
+		time.Now().UTC(),
```
jakubno: the event has its own timestamp, you should use that one

packages/clickhouse/migrations/20250725223330_settings.sql
jakubno, line 1: is this migration needed for anything?

jakubno (dismissed):

packages/clickhouse/pkg/sandbox_events.go
jakubno: @0div, sorry my bad, async insert isnt enabled by default

packages/clickhouse/migrations/20250725223330_settings.sql
0div: not anymore 😢

0div:
@jakubno @tomassrnka can we implement a client0side batcher in another PR? this is exceeding the scope of the initial object of the PR.

0div:
> To sync everybody here—what is the status @0div? I know we talked about the client side, but what was the conclusion we are going with?

@ValentaTomas i'm trying to run some load tests to validate some assumptions about the behavior of `clickhouse-go`'s connection pooling and orchestrator at scale.

packages/clickhouse/pkg/sandbox_events.go
0div: client-side batcher has been added

packages/shared/pkg/feature-flags/flags.go
```diff
@@ -20,11 +21,19 @@ const (
...
+	// ClickhouseQueueSize - size of the channel buffer used to queue incoming sandbox events
+	ClickhouseBatcherQueueSize IntFlag = "clickhouse-batcher-queue-size"
 )
 
 var flagsBool = map[BoolFlag]bool{
-	MetricsWriteFlagName: env.IsDevelopment(),
-	MetricsReadFlagName:  env.IsDevelopment(),
+	MetricsWriteFlagName:                env.IsDevelopment(),
+	MetricsReadFlagName:                 env.IsDevelopment(),
+	SandboxLifeCycleEventsWriteFlagName: env.IsDevelopment(),
```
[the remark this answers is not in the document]
0div: it's fine, we have default values already, you said so yourself.

packages/nomad/template-manager.hcl
```diff
@@ -65,6 +65,7 @@ job "template-manager" {
         ORCHESTRATOR_SERVICES         = "${orchestrator_services}"
         LOGS_COLLECTOR_PUBLIC_IP      = "${logs_collector_public_ip}"
         ALLOW_SANDBOX_INTERNET        = "${allow_sandbox_internet}"
+        CLICKHOUSE_CONNECTION_STRING  = "${clickhouse_connection_string}"
```
ValentaTomas, line 68: Will this cause the template manager to also produce lifecycle events?

packages/nomad/main.tf
ValentaTomas: Out of curiosity, what was the original blocker for using the otel also for the events?

ValentaTomas:
I think this batching system should work. I think it could be simplified, but let's focus on the correctness/non-blocking and merge if that is satisfied.

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -154,6 +198,45 @@ func (s *server) Update(ctx context.Context, req *orchestrator.SandboxUpdateRequ
...
+			err = s.sandboxEventBatcher.Push(clickhouse.SandboxEvent{
+				Timestamp:          time.Now().UTC(),
+				SandboxID:          item.Runtime.SandboxID,
+				SandboxTemplateID:  item.Config.BaseTemplateID,
+				SandboxBuildID:     buildId,
+				SandboxTeamID:      teamID,
+				SandboxExecutionID: item.Runtime.ExecutionID,
+				EventCategory:      string(clickhouse.SandboxEventCategoryLifecycle),
+				EventLabel:         string(clickhouse.SandboxEventLabelUpdate),
+				EventData:          sql.NullString{String: eventData, Valid: true},
```
ValentaTomas, line 231: Was the JSON still problematic?

ValentaTomas:
Also, the cursor bot makes the convo super confusing if most of these are false positives.

packages/orchestrator/internal/server/sandboxes.go
0div: I made it work for inserts but for reads i kept getting 
`unsupported JSON serialization version for decode: 18446744073709551615 clickhouse` 
despite having set all the settings needed for JSON so i reverted back to strings given how small the payload is right now.

packages/nomad/template-manager.hcl
0div: when it creates and pauses the sandbox.

jakubno: It doesn't make much sense right now, does it? As during build it doesn't have any meaningful sandbox id (it will start sandboxes with random id), but that's more of an internal thing. I don't think we want to expose anyway, it would be just confusing

packages/orchestrator/internal/server/sandboxes.go
jakubno: Have you tried ma instead of json? I think it's more native way for clickhouse

packages/nomad/template-manager.hcl
0div: To be clear, i ran tests with this binary deployed for `template-manager` as well and the life-cyle events were registered correctly as u would expect `create, update, update, etc., kill` there no redundancy in them

packages/orchestrator/internal/server/sandboxes.go
0div: yeah map is great but unfortunately it is strictly typed values so can't pass arbitrary `[string]any` maps

PR #887, opened 2025-07-18
Clickhouse - Add dedicated sandbox metrics table

jakubno:
# Description

Adds a dedicated table for sandboxes. There will be following PR (#892) that will remove the old table, it has to be in separate PR to avoid breaking change. This should significantly improve retrieval speed as the column we query by is a primary index.


packages/clickhouse/migrations/20250717135224_sandbox_metrics.sql
```diff
@@ -0,0 +1,42 @@
...
+PARTITION BY toDate(timestamp)
+ORDER BY (sandbox_id, metric_name, toUnixTimestamp64Nano(timestamp))
+TTL toDateTime(timestamp) + INTERVAL 7 DAY;
+
+-- Create routing table that routes sandbox metrics
+CREATE TABLE sandbox_metrics_gauge AS sandbox_metrics_gauge_local
+    ENGINE = Distributed('cluster', currentDatabase(), 'sandbox_metrics_gauge_local', xxHash64(sandbox_id));
+
+-- Create materialized view that routes sandbox metrics
+CREATE MATERIALIZED VIEW sandbox_metrics_gauge_mv
```
jakubno, line 20: All data inserted in `metrics_gauge` will be automatically also added to `sandbox_metrics_gauge`. This is only temporary, `metrics_gauge` will be made "routing" only table - it won't store anything

packages/clickhouse/migrations/20250717135224_sandbox_metrics.sql
```diff
@@ -0,0 +1,42 @@
...
+    toDateTime64(TimeUnix, 9) AS timestamp,
+    Attributes['sandbox_id'] AS sandbox_id,
+    Attributes['team_id'] AS team_id,
+    MetricName AS metric_name,
+    Value AS value
+FROM metrics_gauge
+WHERE Attributes['sandbox_id'] IS NOT NULL;
+
+-- Insert existing sandbox metrics into the local table
+INSERT INTO sandbox_metrics_gauge SELECT
```
jakubno, line 31: Insert old data, so we can start serving from this table and then remove the data from `metrics_gauge` in following PR

packages/clickhouse/migrations/20250717135224_sandbox_metrics.sql
jakubno: https://clickhouse.com/docs/materialized-view/incremental-materialized-view

packages/clickhouse/pkg/metrics.go
```diff
@@ -70,35 +70,35 @@ func (c *Client) QueryLatestMetrics(ctx context.Context, sandboxIDs []string, te
...
         'e2b.sandbox.cpu.used',
         'e2b.sandbox.ram.total',
         'e2b.sandbox.ram.used'
     )
-  AND TimeUnix >= {start_time:DateTime64}
-  AND TimeUnix <= {end_time:DateTime64}
+  AND timestamp >= {start_time:DateTime64}
+  AND timestamp <= {end_time:DateTime64}
 GROUP BY timestamp
 ORDER BY timestamp;
```
dobrac: isn't in this query some clash of "timestamp" naming?

packages/clickhouse/migrations/20250717135224_sandbox_metrics.sql
dobrac: (note, all data that have sandbox_id not null)

dobrac (changes requested):

packages/clickhouse/pkg/metrics.go
```diff
@@ -21,18 +21,18 @@ type Metrics struct {
...
-    Attributes['sandbox_id']                                            AS sandbox_id,
-    Attributes['team_id']                                               AS team_id,
-
-    argMaxIf(Value,  TimeUnix, MetricName = 'e2b.sandbox.cpu.total')    AS cpu_total,
-    argMaxIf(Value,  TimeUnix, MetricName = 'e2b.sandbox.cpu.used')     AS cpu_used,
-    argMaxIf(Value,  TimeUnix, MetricName = 'e2b.sandbox.ram.total')    AS ram_total,
-    argMaxIf(Value,  TimeUnix, MetricName = 'e2b.sandbox.ram.used')     AS ram_used
-FROM metrics_gauge
+    sandbox_id                                            				  AS sandbox_id,
+    team_id                                               				  AS team_id,
```
dobrac: there is a combination of spaces and tabs, not sure if desired

(also, the AS here is unnecessary)

packages/clickhouse/pkg/metrics.go
jakubno: it works fine, but I added an alias for the table to make it more obvious

packages/clickhouse/pkg/metrics.go
```diff
@@ -69,36 +63,28 @@ func (c *Client) QueryLatestMetrics(ctx context.Context, sandboxIDs []string, te
...
-  AND Attributes['team_id'] = {team_id:String}
-  AND MetricName IN (
-        'e2b.sandbox.cpu.total',
-        'e2b.sandbox.cpu.used',
-        'e2b.sandbox.ram.total',
-        'e2b.sandbox.ram.used'
-    )
-  AND TimeUnix >= {start_time:DateTime64}
-  AND TimeUnix <= {end_time:DateTime64}
+SELECT   toStartOfInterval(timestamp, interval {step:UInt32} second) AS timestamp,
```
dobrac:
```suggestion
SELECT   toStartOfInterval(s.timestamp, interval {step:UInt32} second) AS timestamp,
```

dobrac (approved):

PR #889, opened 2025-07-18
Improve build context + cancelation

dobrac:
Improve build context + cancelation. This is a continuation from https://github.com/e2b-dev/infra/pull/854

### DIFF
--- a/tests/integration/internal/utils/sandbox.go
+++ b/tests/integration/internal/utils/sandbox.go
@@ -58,7 +58,7 @@ func SetupSandboxWithCleanup(t *testing.T, c *api.ClientWithResponses, options .
 		Metadata:   &config.metadata,
 	}, setup.WithAPIKey())
 
-	assert.NoError(t, err)
+	require.NoError(t, err)
 	assert.Equal(t, http.StatusCreated, createSandboxResponse.StatusCode())
 	require.NotNil(t, createSandboxResponse.JSON201)
 


dobrac:
> Would it make sense to move the `status` + `metadata` to value in the `utils.SetOnce[V]` and use the `SetValue` and `SetError` to signalize the finished state in both cases, instead of just using it for cancel?

makes sense, I'll try it

packages/orchestrator/internal/template/cache/build_cache.go
```diff
@@ -17,96 +16,78 @@ import (
...
-	return b.metadata
+	return res.Metadata
 }
 
 func (b *BuildInfo) GetStatus() template_manager.TemplateBuildState {
-	b.mu.RLock()
-	defer b.mu.RUnlock()
+	res, err := b.Result.Result()
+	if err != nil {
+		return template_manager.TemplateBuildState_Building
```
jakubno: I would add error explaining this behavior

packages/orchestrator/internal/template/cache/build_cache.go
```diff
@@ -17,96 +16,78 @@ import (
...
+	}
 
-	return b.status
+	return res.Status
 }
 
 func (b *BuildInfo) GetReason() *string {
```
jakubno: wouldn't it make sense to use GetResult

packages/orchestrator/internal/template/cache/build_cache.go
```diff
@@ -17,96 +16,78 @@ import (
...
+func (b *BuildInfo) SetSuccess(metadata *template_manager.TemplateBuildMetadata) {
+	_ = b.Result.SetValue(BuildInfoResult{
+		Status:   template_manager.TemplateBuildState_Completed,
+		Metadata: metadata,
+		Reason:   nil,
+	})
+}
 
-	return b.ctx
+func (b *BuildInfo) SetFail(reason *string) {
```
jakubno: i would require string here, we want to always set the reason

packages/orchestrator/internal/template/cache/build_cache.go
dobrac: there is no GetResult method

packages/orchestrator/internal/template/cache/build_cache.go
dobrac: added comment

dobrac:
> * Why don't you use `context.WithoutCancel(ctx)`

I didn't know, refactored, thx

> * Wouldn't it be more correct to return `Result` in `template_status` to prevent some inconsistent state? e.g. you get status building and then reason for failure

yes, refactored

packages/orchestrator/internal/template/cache/build_cache.go
dobrac: added the method

PR #895, opened 2025-07-21
Add disk usage metrics

jakubno:
# Description

- Adds disk usage metrics
- Returns more precise metrics for memory


packages/clickhouse/pkg/metrics.go
```diff
@@ -17,20 +17,22 @@ type Metrics struct {
...
        team_id,
-       Argmaxif(value, timestamp, metric_name = 'e2b.sandbox.cpu.total') AS cpu_total,
-       Argmaxif(value, timestamp, metric_name = 'e2b.sandbox.cpu.used')  AS cpu_used,
-       Argmaxif(value, timestamp, metric_name = 'e2b.sandbox.ram.total') AS ram_total,
-       Argmaxif(value, timestamp, metric_name = 'e2b.sandbox.ram.used')  AS ram_used
+       argMaxIf(value, timestamp, metric_name = 'e2b.sandbox.cpu.total')  AS cpu_total,
+       argMaxIf(value, timestamp, metric_name = 'e2b.sandbox.cpu.used')   AS cpu_used,
+       argMaxIf(value, timestamp, metric_name = 'e2b.sandbox.ram.total')  AS ram_total,
+       argMaxIf(value, timestamp, metric_name = 'e2b.sandbox.ram.used')   AS ram_used,
+       argMaxIf(value, timestamp, metric_name = 'e2b.sandbox.disk.total') AS disk_total,
```
dobrac, line 31: what's returned when the sandbox doesn't export the disk metrics?

packages/envd/internal/host/metrics.go
```diff
@@ -41,11 +51,24 @@ func GetMetrics() (*Metrics, error) {
...
 	}
 
+	var stat unix.Statfs_t
+	err = unix.Statfs("/", &stat)
+	if err != nil {
+		return nil, err
+	}
+
+	diskTotal := stat.Blocks * uint64(stat.Bsize)
+	diskUsed := diskTotal - (stat.Bfree * uint64(stat.Bsize))
```
dobrac: did you consider Bavail instead of Bfree to show available space?

jakubno: Good catch, that's probably closer what you would expect with "used". I am thinking if we shouldn't report `disk.available` to make it clearer

dobrac: let's keep it at used so it's on par with memory and cpu imo

packages/clickhouse/pkg/metrics.go
jakubno: It will return 0, the same way as for old sandboxes with old envd

packages/orchestrator/internal/metrics/sandboxes.go
```diff
@@ -141,7 +141,13 @@ func (so *SandboxObserver) startObserving() (metric.Registration, error) {
...
 
 			for _, sbx := range so.sandboxes.Items() {
-				if !utils.IsGTEVersion(sbx.Config.EnvdVersion, minEnvdVersionForMetrics) {
+				ok, err := utils.IsGTEVersion(sbx.Config.EnvdVersion, minEnvdVersionForMetrics)
+				if err != nil {
+					zap.L().Error("Failed to check envd version", zap.Error(err), zap.String("sandbox_id", sbx.Config.SandboxId))
+					continue
+				}
+				if !ok {
+					zap.L().Warn("Sandbox metrics are not supported for this envd version", zap.String("envd_version", sbx.Config.EnvdVersion), zap.String("sandbox_id", sbx.Config.SandboxId))
```
dobrac: this will be writing the log entry for all running in perpetuity repeatedly, same applies to the rest of logs

jakubno: forgotten debug logs, good catch!

dobrac (approved):

PR #896, opened 2025-07-21
Fix sandbox cleanup on timeout error

dobrac:
If the context is canceled (e.g. timeout), the cleanup is not run properly. This PR tries to fix it by running the cleanup using the background context


jakubno:
> note: If the cleanup would hang indefinitely, maybe we should return the request and continue hanging in the background goroutine?

In `Delete`?

ValentaTomas (approved):

ValentaTomas:
I moved the cleanup into a separate goroutine as discussed in a coming PR, but this PR fixing the background context still makes sense.

ValentaTomas:
For waiting in create, yes, that is an optimization we can do later.

PR #898, opened 2025-07-21
Add clickhouse tests

jakubno:
# Description

Add clickhouse to integration tests


packages/shared/pkg/telemetry/config.go
```diff
@@ -10,7 +10,7 @@ import (
 	"go.opentelemetry.io/otel/semconv/v1.21.0"
 )
 
-var otelCollectorGRPCEndpoint = os.Getenv("OTEL_COLLECTOR_GRPC_ENDPOINT")
+var OtelCollectorGRPCEndpoint = os.Getenv("OTEL_COLLECTOR_GRPC_ENDPOINT")
```
dobrac, line 13: why making it exported?

tests/integration/internal/tests/api/sandbox_list_metrics_test.go
```diff
@@ -0,0 +1,44 @@
...
+	// Ensure there are some metrics
+	time.Sleep(7 * time.Second)
+
+	response, err := c.GetSandboxesMetricsWithResponse(t.Context(), &api.GetSandboxesMetricsParams{SandboxIds: []string{sbx1.SandboxID, sbx2.SandboxID}}, setup.WithAPIKey())
+	require.NoError(t, err)
+	require.Equal(t, http.StatusOK, response.StatusCode())
+
+	require.NotNil(t, response.JSON200)
+	require.NotNil(t, response.JSON200.Sandboxes)
+	require.Equal(t, 2, len(response.JSON200.Sandboxes), "Expected at least one metric in the response")
```
dobrac:
```suggestion
	require.Equal(t, 2, len(response.JSON200.Sandboxes), "Expected two metrics in the response")
```

tests/integration/configs/otel-collector.tpl.yaml
dobrac, line 1: it would be nice to merge the config with the other otel-collector, so when making changes, you don't need to think that it's duplicated in the tests and it's no the same configuration

tests/integration/Makefile
```diff
@@ -5,6 +5,9 @@ openapi_api := ../../spec/openapi.yml
 openapi_envd := ../../packages/envd/spec/envd.yaml
 codegen := go tool github.com/oapi-codegen/oapi-codegen/v2/cmd/oapi-codegen
 
+# Docker bridge network
+TESTS_CLICKHOUSE_HOST ?= clickhouse
```
dobrac: is this relying that in the integration tests we start the Docker Container with the name "clickhouse"? If so, I would redo a bit how this configuration is constructed so the connection is obvious

Also, can we have the makefile 
```
.PHONY: run-otel-collector
run-otel-collector:
```
in otel collector (and not in the tests)? Maybe creating a new package is fine

dobrac (changes requested):

packages/shared/pkg/telemetry/config.go
jakubno: I use it in `main.go` to determine if we want to use `noop` or not

PR #900, opened 2025-07-21
Limit resource usage and parallelism of `gcloud` when uploading snapshots to reduce `EOF` errors

ValentaTomas:
- Limit CPU and memory usage of `gcloud` upload command to reduce CPU stall that was causing the EOF to manifest under big parallism
- Limit max concurrent upload via `gcloud` upload command to further reduce the CPU stall
  - This might have an effect on the performance, so we will be checking this out
- Also add a retry mechanism for `gcloud` as some errors in resume can be caused by failed upload of files
- Move the `ReadFrom` close to defer

### TODO
- [ ] Try replicating the EOF bug with this setup
- [ ] Test if upload works
- [ ] Check possible performance impact (can be tested separately)


packages/shared/pkg/storage/storage_google.go
```diff
@@ -26,8 +28,22 @@ const (
...
 	googleBackoffMultiplier = 2
 	googleMaxAttempts       = 10
+
+	// GCloud imposed limits:
+
+	// 10 concurrent uploads
+	gcloudMaxConcurrentUploads = 8
+	gcloudMaxRetries           = 3
+	// 200% CPU quota (200% / 9600% CPU total -> 2.083% total CPU per gcloud), total max CPU used for uploads is 2.083% * 10 = 20.83%
+	gcloudMaxCPUQuota = 200
```
dobrac: this would be probably better to compute based on the actual CPU count, same applies to memory

packages/shared/pkg/storage/storage_google.go
```diff
@@ -171,17 +187,13 @@ func (g *GCPBucketStorageObjectProvider) ReadAt(buff []byte, off int64) (n int,
 
 func (g *GCPBucketStorageObjectProvider) ReadFrom(src io.Reader) (int64, error) {
 	w := g.handle.NewWriter(g.ctx)
+	defer w.Close()
```
dobrac, line 185: nice

packages/shared/pkg/storage/storage_google.go
```diff
@@ -26,8 +28,22 @@ const (
...
+
+	// 10 concurrent uploads
+	gcloudMaxConcurrentUploads = 8
+	gcloudMaxRetries           = 3
+	// 200% CPU quota (200% / 9600% CPU total -> 2.083% total CPU per gcloud), total max CPU used for uploads is 2.083% * 10 = 20.83%
+	gcloudMaxCPUQuota = 200
+	// 2048M memory limit (2048M / 3865471M -> 0.053% of total memory), total max memory used for uploads is 0.053% * 10 = 0.53%
+	gcloudMaxMemoryLimit = 2048
+	// 16 tasks max, total max tasks used for uploads is 16 * 10 = 160
+	gcloudMaxTasksMax = 16
```
dobrac: should we maybe use LaunchDarkly for these so we can update them in runtime? e.g. if there would be any issue, we would need to roll all orchestrators now

packages/shared/pkg/storage/storage_google.go
```diff
@@ -210,20 +222,49 @@ func (g *GCPBucketStorageObjectProvider) WriteTo(dst io.Writer) (int64, error) {
...
+	for range gcloudMaxRetries {
+		err := upload()
+		if err != nil {
+			// Failed to upload file, retrying.
+			zap.L().Warn("Failed to upload file to GCS, retrying", zap.Error(err), zap.String("path", g.path))
+
+			continue
+		}
+
+		// Files was successfully uploaded
```
dobrac, line 270: when all retries fail, we should probably return an error

ValentaTomas: Oh, nice catch.

packages/shared/pkg/storage/storage_google.go
ValentaTomas: Good idea, I'll check that and we can reduce the amount of time on perf monitoring.

packages/shared/pkg/storage/storage_google.go
ValentaTomas: I'm thinking if we are using launch darkly, do we want to still calculate them? The default value will then be unclear.

dobrac: we need to provide some default when launchdarkly is not available, but the default could be also that the limits are basically turned off

packages/orchestrator/internal/template/server/main.go
```diff
@@ -52,14 +52,10 @@ func New(
 	proxy *proxy.SandboxProxy,
 	sandboxes *smap.Map[*sandbox.Sandbox],
 	templateCache *sbxtemplate.Cache,
+	templatePersistence storage.StorageProvider,
```
dobrac, line 56: just fyi, there is also GetBuildCacheStorageProvider which is also type of StorageProvider and is used for files and index

packages/shared/pkg/feature-flags/flags.go
```diff
@@ -17,3 +23,49 @@ const (
...
+const GcloudConcurrentUploadLimitDefault = 8
+
+// Flag for setting the maximum number of CPUs for GCloud uploads
+// https://app.launchdarkly.com/projects/default/flags/gcloud-max-cpu-quota
+const (
+	GcloudMaxCPUQuota = "gcloud-max-cpu-quota"
+)
+
+// GcloudMaxCPUQuotaDefault default is 2% of total CPU
+var GcloudMaxCPUQuotaDefault = runtime.NumCPU() * 2 / 100
```
dobrac: this is always 0 for NumCPU < 50 as it's integer division

packages/shared/pkg/feature-flags/flags.go
```diff
@@ -17,3 +23,49 @@ const (
...
+
+// Flag for setting the maximum memory limit for GCloud uploads
+// https://app.launchdarkly.com/projects/default/flags/gcloud-max-memory-limit
+const (
+	GcloudMaxMemoryLimit = "gcloud-max-memory-limit"
+)
+
+// GcloudMaxMemoryLimitDefault default is 0.5% of total memory
+var GcloudMaxMemoryLimitDefault = getDefaultMemoryLimit()
```
dobrac: Specify that it's in MB

packages/shared/pkg/feature-flags/flags.go
```diff
@@ -17,3 +23,49 @@ const (
...
+// getDefaultMemoryLimit returns the default memory limit for GCloud uploads in MiB
+func getDefaultMemoryLimit() int {
+	vmStat, err := mem.VirtualMemory()
+	if err != nil {
+		panic(err)
+	}
+
+	totalMemory := vmStat.Total
+	// Calculate the memory limit based on the percentage
+	return int(0.005 * float64(totalMemory/1024/1024)) // Convert to MiB
```
dobrac: wouldn't it be better to do (0.005 * totalMemory)/1024/1024

packages/shared/pkg/storage/storage_google.go
```diff
@@ -26,18 +30,29 @@ const (
 	googleMaxBackoff        = 10 * time.Second
 	googleBackoffMultiplier = 2
 	googleMaxAttempts       = 10
+
+	// GCloud imposed limits:
```
dobrac: what does the comment mean?

packages/shared/pkg/feature-flags/upload-limit.go
```diff
@@ -0,0 +1,34 @@
...
+func (c *Client) UpdateUploadLimitSemaphore(ctx context.Context, uploadSemaphore *utils.AdjustableSemaphore) {
+	flagCtx := ldcontext.NewBuilder(GcloudConcurrentUploadLimit).Build()
+
+	ticker := time.NewTicker(5 * time.Second)
+	for {
+		select {
+		case <-ticker.C:
+			uploadLimitFlag, flagErr := c.Ld.IntVariation(GcloudConcurrentUploadLimit, flagCtx, GcloudConcurrentUploadLimitDefault)
+			if flagErr != nil {
+				zap.L().Error("soft failing during metrics write feature flag receive", zap.Error(flagErr))
```
dobrac: is this gonna log every 5 seconds?

packages/shared/pkg/storage/storage_google.go
```diff
@@ -112,11 +151,34 @@ func (g *GCPBucketStorageProvider) OpenObject(ctx context.Context, path string)
 		),
 	)
 
+	flagCtx := ldcontext.NewBuilder(featureflags.GcloudMaxCPUQuota).SetString("path", path).Build()
+	maxCPU, flagErr := g.featureFlags.Ld.IntVariation(featureflags.GcloudMaxCPUQuota, flagCtx, featureflags.GcloudMaxCPUQuotaDefault)
+	if flagErr != nil {
+		zap.L().Error("soft failing during metrics write feature flag receive", zap.Error(flagErr))
```
dobrac:
```suggestion
		zap.L().Warn("soft failing during metrics write feature flag receive", zap.Error(flagErr))
```

packages/shared/pkg/storage/storage_google.go
```diff
@@ -112,11 +151,34 @@ func (g *GCPBucketStorageProvider) OpenObject(ctx context.Context, path string)
...
+	}
+
+	flagCtx = ldcontext.NewBuilder(featureflags.GcloudMaxMemoryLimit).SetString("path", path).Build()
+	maxMemory, flagErr := g.featureFlags.Ld.IntVariation(featureflags.GcloudMaxMemoryLimit, flagCtx, featureflags.GcloudMaxMemoryLimitDefault)
+	if flagErr != nil {
+		zap.L().Error("soft failing during metrics write feature flag receive", zap.Error(flagErr))
```
dobrac:
```suggestion
		zap.L().Warn("soft failing during metrics write feature flag receive", zap.Error(flagErr))
```

packages/shared/pkg/storage/storage_google.go
```diff
@@ -112,11 +151,34 @@ func (g *GCPBucketStorageProvider) OpenObject(ctx context.Context, path string)
...
+	}
+
+	flagCtx = ldcontext.NewBuilder(featureflags.GcloudMaxTasks).SetString("path", path).Build()
+	maxTasks, flagErr := g.featureFlags.Ld.IntVariation(featureflags.GcloudMaxTasks, flagCtx, featureflags.GcloudMaxTasksDefault)
+	if flagErr != nil {
+		zap.L().Error("soft failing during metrics write feature flag receive", zap.Error(flagErr))
```
dobrac:
```suggestion
		zap.L().Warn("soft failing during metrics write feature flag receive", zap.Error(flagErr))
```

packages/shared/pkg/feature-flags/upload-limit.go
dobrac:
```suggestion
				zap.L().Warn("soft failing during metrics write feature flag receive", zap.Error(flagErr))
```

dobrac (changes requested):

packages/shared/pkg/storage/storage_google.go
jakubno: I don't know, removed

packages/shared/pkg/feature-flags/upload-limit.go
jakubno: Yes, decrease the level to warning. This should only happen if the flag isn't present / there's incorrect api key / request failed

packages/orchestrator/internal/template/server/main.go
jakubno: do you think we should add it there as well?

jakubno: added

packages/shared/pkg/limit/upload.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/launchdarkly/go-sdk-common/v3/ldcontext"
+	"go.uber.org/zap"
+
+	featureflags "github.com/e2b-dev/infra/packages/shared/pkg/feature-flags"
+)
+
+func (l *Limiter) UpdateUploadLimitSemaphore() {
+	flagCtx := ldcontext.NewBuilder(featureflags.GcloudConcurrentUploadLimit).Build()
+
+	ticker := time.NewTicker(5 * time.Second)
```
djeebus, line 12: If we scale wide enough, this might become an issue. I'm not sure what launch darkly's rate limits are for this API, but we may want to increase this number (1 minute?) to give us some more head room. At the least, it should be a constant.

Example: if the rate limit is 100 requests per 10 second window, then this 5 second interval means we can only spin up 50 orchestrators before we hit that rate limit.

packages/shared/pkg/storage/storage_google.go
```diff
@@ -210,23 +217,61 @@ func (g *GCPBucketStorageObjectProvider) WriteTo(dst io.Writer) (int64, error) {
...
+		}
 
-	output, err := cmd.CombinedOutput()
-	if err != nil {
-		return fmt.Errorf("failed to upload file to GCS: %w\n%s", err, string(output))
+		args := append(systemRunArgs,
+			"gcloud",
+			"storage",
+			"cp",
+			"--verbosity", "error",
```
djeebus: To keep these together, we could merge them into a single argument:

```suggestion
			"--verbosity=error",
```

packages/shared/pkg/limit/upload.go
jakubno: They bill you per connections and there shouldn't be any rate limits. We used this to roll metrics or proxying which was calculated per sandbox.
Source: https://launchdarkly.com/docs/api#rate-limiting-and-sdks

packages/shared/pkg/storage/storage_google.go
```diff
@@ -210,23 +217,61 @@ func (g *GCPBucketStorageObjectProvider) WriteTo(dst io.Writer) (int64, error) {
...
+			zap.L().Warn("Failed to upload file to GCS, retrying", zap.Error(err), zap.String("path", g.path))
+
+			continue
+		}
+
+		// Files was successfully uploaded
+		return nil
+	}
+
+	return fmt.Errorf("failed to upload file to GCS after %d retries", gcloudMaxRetries)
```
djeebus: Should we include the last error as well? It would probably make debugging issues easier; you wouldn't have to go hunting for the last error.

packages/shared/pkg/utils/resizable_semaphore_test.go
```diff
@@ -0,0 +1,311 @@
...
+// basic correctness
+// -----------------------------------------------------------------------------
+
+func TestBasicAcquireTryRelease(t *testing.T) {
+	s, err := NewAdjustableSemaphore(2)
+	require.NoError(t, err)
+
+	if err := s.Acquire(context.Background(), 1); err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
```
djeebus: You can remove a lot of the conditionals by using testify libraries:

```golang
import "github.com/stretchr/testify/require"
```

```suggestion
	err := s.Acquire(context.Background(), 1)
	require.NoError(t, err)
```

packages/shared/pkg/utils/resizable_semaphore_test.go
```diff
@@ -0,0 +1,311 @@
...
+	if got := s.TryAcquire(1); !got {
+		t.Fatalf("TryAcquire should have succeeded with remaining capacity")
+	}
```
djeebus: Same for these:

```golang
import "github.com/stretchr/testify/require"
```

```suggestion
	got := s.TryAcquire(1)
	require.True(t, got, "TryAcquire should have succeeded with remaining capacity")
```

packages/shared/pkg/utils/resizable_semaphore_test.go
```diff
@@ -0,0 +1,311 @@
...
+	// should panic on negative release
+	mustPanic(t, func() { s.Release(2) })
+}
+
+func TestReleaseErrorsOnZero(t *testing.T) {
+	s, err := NewAdjustableSemaphore(2)
+	require.NoError(t, err)
+
+	// should panic on negative release
+	mustPanic(t, func() { s.Release(0) })
```
djeebus: slightly different DX, but you could probably remove the `func() { ... }` with

```
defer mustPanic()
s.Release(0)
```

not necessary though, just offering an alternative!

packages/shared/pkg/utils/resizable_semaphore_test.go
```diff
@@ -0,0 +1,311 @@
...
+	// should panic on negative limit
+	_, err := NewAdjustableSemaphore(-1)
+	assert.Error(t, err, "NewAdjustableSemaphore should return an error for negative limit")
+}
+
+func TestNewAdjustableSemaphoreErrorsOnZeroLimit(t *testing.T) {
+	// should panic on zero limit
+	_, err := NewAdjustableSemaphore(0)
+	assert.Error(t, err, "NewAdjustableSemaphore should return an error for zero limit")
+}
```
djeebus: Seems like these do _not_ panic, we should probably change the comment on each.

packages/shared/pkg/utils/resizable_semaphore_test.go
```diff
@@ -0,0 +1,311 @@
...
+	case <-time.After(100 * time.Millisecond):
+		t.Fatalf("Acquire didn’t return after context cancellation")
+	}
+}
+
+// -----------------------------------------------------------------------------
+// race-detector stress
+// -----------------------------------------------------------------------------
+
+func TestConcurrentStressNoDeadlockOrRace(t *testing.T) {
```
djeebus, line 269: nice!

packages/shared/pkg/storage/storage_google.go
```diff
@@ -210,23 +217,61 @@ func (g *GCPBucketStorageObjectProvider) WriteTo(dst io.Writer) (int64, error) {
...
-		"gcloud",
-		"storage",
-		"cp",
-		"--verbosity",
-		"error",
-		path,
-		fmt.Sprintf("gs://%s/%s", g.storage.bucket.BucketName(), g.path),
-	)
+	upload := func() error {
+		systemRunArgs := []string{
```
djeebus: should probably be named `systemDRunArgs`

packages/shared/pkg/storage/storage_google.go
```diff
@@ -210,23 +217,61 @@ func (g *GCPBucketStorageObjectProvider) WriteTo(dst io.Writer) (int64, error) {
...
+			"storage",
+			"cp",
+			"--verbosity", "error",
+			path,
+			fmt.Sprintf("gs://%s/%s", g.storage.bucket.BucketName(), g.path),
+		)
+
+		cmd := exec.CommandContext(
+			g.ctx,
+			"systemd-run",
```
djeebus, line 248: I'm not sure, but we might want to add `--scope` here if `cmd.CombinedOutput()` no longer contains stdout/stderr. It's not super clear to me, but if we're still receiving output, feel free to ignore this!

packages/shared/pkg/limit/limiter.go
```diff
@@ -0,0 +1,41 @@
...
+		done:                make(chan struct{}),
+	}
+
+	go l.UpdateUploadLimitSemaphore()
+
+	return l, nil
+}
+
+func (l *Limiter) Close(ctx context.Context) error {
+	if l.done != nil {
```
dobrac: can l.done be nil?

dobrac: also, you may consider adding sync.Once to prevent double close errors

jakubno: It shouldn't, updated

packages/shared/pkg/storage/storage_google.go
jakubno: we already use `--scope` flag, refactored the args, to make it clearer

packages/shared/pkg/limit/gcloud.go
```diff
@@ -0,0 +1,42 @@
...
+	flagCtx := ldcontext.NewBuilder(featureflags.GcloudMaxCPUQuota).SetString("path", path).Build()
+	maxCPU, flagErr := l.featureFlags.Ld.IntVariation(featureflags.GcloudMaxCPUQuota, flagCtx, featureflags.GcloudMaxCPUQuotaDefault)
+	if flagErr != nil {
+		zap.L().Warn("soft failing during gcloud cmd limits feature flag receive", zap.Error(flagErr), zap.Int("maxCPU", featureflags.GcloudMaxCPUQuotaDefault))
+	}
+
+	flagCtx = ldcontext.NewBuilder(featureflags.GcloudMaxMemoryLimitMiB).SetString("path", path).Build()
+	maxMemory, flagErr := l.featureFlags.Ld.IntVariation(featureflags.GcloudMaxMemoryLimitMiB, flagCtx, featureflags.GcloudMaxMemoryLimitMiBDefault)
+	if flagErr != nil {
+		zap.L().Warn("soft failing during gcloud cmd limits feature flag receive", zap.Error(flagErr), zap.Int("maxMemory", featureflags.GcloudMaxMemoryLimitMiBDefault))
```
dobrac:
```suggestion
		zap.L().Warn("soft failing during gcloud cmd limits feature flag receive", zap.Error(flagErr), zap.Int("maxMemory", maxMemory))
```

packages/shared/pkg/limit/gcloud.go
```diff
@@ -0,0 +1,42 @@
...
+
+func (l *Limiter) GCloudUploadLimiter() *utils.AdjustableSemaphore {
+	return l.gCloudUploadLimiter
+}
+
+func (l *Limiter) GCloudCmdLimits(path string) []string {
+	flagCtx := ldcontext.NewBuilder(featureflags.GcloudMaxCPUQuota).SetString("path", path).Build()
+	maxCPU, flagErr := l.featureFlags.Ld.IntVariation(featureflags.GcloudMaxCPUQuota, flagCtx, featureflags.GcloudMaxCPUQuotaDefault)
+	if flagErr != nil {
+		zap.L().Warn("soft failing during gcloud cmd limits feature flag receive", zap.Error(flagErr), zap.Int("maxCPU", featureflags.GcloudMaxCPUQuotaDefault))
```
dobrac:
```suggestion
		zap.L().Warn("soft failing during gcloud cmd limits feature flag receive", zap.Error(flagErr), zap.Int("maxCPU", maxCPU))
```

packages/shared/pkg/limit/gcloud.go
```diff
@@ -0,0 +1,42 @@
...
+	flagCtx = ldcontext.NewBuilder(featureflags.GcloudMaxMemoryLimitMiB).SetString("path", path).Build()
+	maxMemory, flagErr := l.featureFlags.Ld.IntVariation(featureflags.GcloudMaxMemoryLimitMiB, flagCtx, featureflags.GcloudMaxMemoryLimitMiBDefault)
+	if flagErr != nil {
+		zap.L().Warn("soft failing during gcloud cmd limits feature flag receive", zap.Error(flagErr), zap.Int("maxMemory", featureflags.GcloudMaxMemoryLimitMiBDefault))
+	}
+
+	flagCtx = ldcontext.NewBuilder(featureflags.GcloudMaxTasks).SetString("path", path).Build()
+	maxTasks, flagErr := l.featureFlags.Ld.IntVariation(featureflags.GcloudMaxTasks, flagCtx, featureflags.GcloudMaxTasksDefault)
+	if flagErr != nil {
+		zap.L().Warn("soft failing during gcloud cmd limits feature flag receive", zap.Error(flagErr), zap.Int("maxTasks", featureflags.GcloudMaxTasksDefault))
```
dobrac:
```suggestion
		zap.L().Warn("soft failing during gcloud cmd limits feature flag receive", zap.Error(flagErr), zap.Int("maxTasks", maxTasks))
```

packages/shared/pkg/feature-flags/client.go
```diff
@@ -18,37 +21,78 @@ var launchDarklyApiKey = os.Getenv("LAUNCH_DARKLY_API_KEY")
...
 	ldClient, err = ldclient.MakeClient(launchDarklyApiKey, waitForInit)
 	if err != nil {
 		return nil, err
 	}
 
-	return &Client{Ld: ldClient}, nil
+	return &Client{ld: ldClient}, nil
+}
+
+func (c *Client) BoolFlag(flagName BoolFlag, evalKey string) (bool, error) {
```
dobrac: can you rename "eval" to something else? eval is kinda "evil" to me and signifies some potential security issue (if that's not the case though)

packages/shared/pkg/feature-flags/flags.go
```diff
@@ -1,19 +1,69 @@
...
-import "github.com/e2b-dev/infra/packages/shared/pkg/env"
+import (
+	"runtime"
+
+	"github.com/shirou/gopsutil/v4/mem"
+
+	"github.com/e2b-dev/infra/packages/shared/pkg/env"
+)
+
+// https://app.launchdarkly.com/projects/default/flags/
```
dobrac: unattached comment

dobrac (changes requested):

dobrac (approved):

PR #903, opened 2025-07-22
Send edge sandbox catalog as gRPC metadata

sitole:
Remove the need to call the Edge REST API for sandbox registry/remove from cluster shared cache.

This PR introduces changes that will move sandbox catalog create/delete requests directly into sandbox start/delete/pause metadata. It should simplify the process of calling and speed up startup, because it's no longer another request.

### DIFF
--- a/packages/api/internal/orchestrator/client.go
+++ b/packages/api/internal/orchestrator/client.go
@@ -107,8 +107,7 @@ func (o *Orchestrator) connectToNode(ctx context.Context, node *node.NodeInfo) e
 			client:   client,
 			clientMd: make(metadata.MD),
 
-			Info:              node,
-			serviceInstanceID: nodeInfo.ServiceId,
+			Info: node,
 
 			meta:           getNodeMetadata(nodeInfo, node.ID, nodeInfo.ServiceId),
 			buildCache:     buildCache,
@@ -141,7 +140,6 @@ func (o *Orchestrator) connectToClusterNode(cluster *edge.Cluster, i *edge.Clust
 		Info: &node.NodeInfo{
 			ID: poolNodeID,
 		},
-		serviceInstanceID: i.ServiceInstanceID,
 
 		status: OrchestratorToApiNodeStateMapper[i.GetStatus()],
 		meta: nodeMetadata{
--- a/packages/api/internal/orchestrator/client.go
+++ b/packages/api/internal/orchestrator/client.go
@@ -107,7 +107,9 @@ func (o *Orchestrator) connectToNode(ctx context.Context, node *node.NodeInfo) e
 			client:   client,
 			clientMd: make(metadata.MD),
 
-			Info:           node,
+			Info:              node,
+			serviceInstanceID: nodeInfo.ServiceId,
+
 			meta:           getNodeMetadata(nodeInfo, node.ID),
 			buildCache:     buildCache,
 			status:         nodeStatus,
@@ -139,6 +141,7 @@ func (o *Orchestrator) connectToClusterNode(cluster *edge.Cluster, i *edge.Clust
 		Info: &node.NodeInfo{
 			ID: poolNodeID,
 		},
+		serviceInstanceID: i.ServiceInstanceID,
 
 		status: OrchestratorToApiNodeStateMapper[i.GetStatus()],
 		meta: nodeMetadata{
--- a/packages/api/internal/orchestrator/client.go
+++ b/packages/api/internal/orchestrator/client.go
@@ -110,7 +110,7 @@ func (o *Orchestrator) connectToNode(ctx context.Context, node *node.NodeInfo) e
 			Info:              node,
 			serviceInstanceID: nodeInfo.ServiceId,
 
-			meta:           getNodeMetadata(nodeInfo, node.ID),
+			meta:           getNodeMetadata(nodeInfo, node.ID, nodeInfo.ServiceId),
 			buildCache:     buildCache,
 			status:         nodeStatus,
 			sbxsInProgress: smap.New[*sbxInProgress](),
--- a/packages/api/internal/orchestrator/client.go
+++ b/packages/api/internal/orchestrator/client.go
@@ -109,7 +109,7 @@ func (o *Orchestrator) connectToNode(ctx context.Context, node *node.NodeInfo) e
 
 			Info: node,
 
-			meta:           getNodeMetadata(nodeInfo, node.ID, nodeInfo.ServiceId),
+			meta:           getNodeMetadata(nodeInfo, node.ID),
 			buildCache:     buildCache,
 			status:         nodeStatus,
 			sbxsInProgress: smap.New[*sbxInProgress](),
--- a/packages/api/internal/orchestrator/node.go
+++ b/packages/api/internal/orchestrator/node.go
@@ -107,7 +107,7 @@ func (n *Node) setStatus(status api.NodeStatus) {
 	}
 }
 
-func (n *Node) setMetadata(i *orchestratorinfo.ServiceInfoResponse, nodeID string, serviceInstanceID string) {
+func (n *Node) setMetadata(i *orchestratorinfo.ServiceInfoResponse, nodeID string) {
 	n.mutex.Lock()
 	defer n.mutex.Unlock()
 	n.meta = getNodeMetadata(i, nodeID)
--- a/packages/api/internal/orchestrator/node.go
+++ b/packages/api/internal/orchestrator/node.go
@@ -31,9 +31,11 @@ type sbxInProgress struct {
 }
 
 type nodeMetadata struct {
-	orchestratorID string
-	commit         string
-	version        string
+	orchestratorID    string
+	serviceInstanceID string
+
+	commit  string
+	version string
 }
 
 type Node struct {
@@ -46,8 +48,7 @@ type Node struct {
 	ClusterID     uuid.UUID
 	ClusterNodeID string
 
-	Info              *node.NodeInfo
-	serviceInstanceID string
+	Info *node.NodeInfo
 
 	meta   nodeMetadata
 	status api.NodeStatus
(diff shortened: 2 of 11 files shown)


packages/api/internal/orchestrator/node.go
```diff
@@ -45,7 +46,8 @@ type Node struct {
 	ClusterID     uuid.UUID
 	ClusterNodeID string
 
-	Info *node.NodeInfo
+	Info              *node.NodeInfo
+	serviceInstanceID string
```
sitole: Service instance ID is known for both local and cluster orchestrators, but for back compatibility API is still using node ID instead of service instance ID (already used for orchestrators' communication in edge). Before now, it was used only when creating gRPC metadata for the cluster orchestrator.

jakubno: What's missing to use serviceInstanceID everywhere?

sitole: 1. Need to know it when doing nodes sync via Nomad (we need to fetch instance status before adding to the map)
2. Naming re-write (replace use of nodeID to instanceID) to make it more obvious and less confusing

We can probably do it as part of the "internal cluster" migration.

packages/api/internal/orchestrator/client.go
```diff
@@ -107,7 +107,9 @@ func (o *Orchestrator) connectToNode(ctx context.Context, node *node.NodeInfo) e
 			client:   client,
 			clientMd: make(metadata.MD),
 
-			Info:           node,
+			Info:              node,
+			serviceInstanceID: nodeInfo.ServiceId,
```
jakubno: can't it be part of meta?

packages/api/internal/orchestrator/node.go
```diff
@@ -288,6 +290,36 @@ func (n *Node) GetClient(ctx context.Context) (*grpclient.GRPCClient, context.Co
 	return n.client, metadata.NewOutgoingContext(ctx, n.clientMd)
 }
 
+func (n *Node) GetSandboxCreateCtx(ctx context.Context, req *orchestrator.SandboxCreateRequest) context.Context {
+	if n.ClusterID == uuid.Nil {
+		return metadata.NewOutgoingContext(ctx, n.clientMd)
```
jakubno, line 301: do you need to set this even for local orchestrator?

packages/client-proxy/internal/edge-pass-through/proxy.go
```diff
@@ -204,3 +230,53 @@ func (s *NodePassThroughServer) forwardServerToClient(src grpc.ServerStream, dst
 
 	return ret
 }
+
+func (s *NodePassThroughServer) eventsHandler(md metadata.MD) (func(error), error) {
```
jakubno: can't you do some switch and then handle the logic in one place? you call `HandleSandboxCatalogCreateEvent` and then the rest of logic is in this function

packages/client-proxy/internal/edge-pass-through/proxy.go
```diff
@@ -103,14 +117,24 @@ func (s *NodePassThroughServer) Handler(srv interface{}, serverStream grpc.Serve
...
-	clientConnection, err := s.director(serverStream.Context())
+	clientConnection, md, err := s.director(serverStream.Context())
 	if err != nil {
 		return err
 	}
 
 	clientCtx, clientCancel := context.WithCancel(s.ctx)
 	defer clientCancel()
 
+	callback, err := s.eventsHandler(md)
```
jakubno, line 129: wouldn't it make sense to run this as the last thing and only if there's no error?

packages/shared/pkg/edge/events.go
```diff
@@ -0,0 +1,147 @@
...
+
+func HandleSandboxCatalogCreateEvent(md metadata.MD) (e *SandboxCatalogCreateEvent, err error) {
+	v, f := getMetadataValue(md, eventTypeHeader)
+	if !f || v != catalogCreateEventType {
+		return nil, nil
+	}
+
+	sandboxID, found := getMetadataValue(md, sbxIdHeader)
+	if !found {
+		return nil, ErrSandboxCreateEventRequiredFieldsMissing
```
jakubno: can't you create a new struct and also pass what header is missing? 🙏🏻

packages/api/internal/orchestrator/node.go
sitole: Nope, added comment 6f63cdcb32407d19eb777b2754b42410b02c5e16 that IF is not needed, but it's more explicit to explain why its not needed.

packages/api/internal/orchestrator/client.go
sitole: Yes, actually it can be!

packages/client-proxy/internal/edge-pass-through/proxy.go
sitole: Calling sandbox catalog add and then processing sandbox request ensures that when sandbox catalog add fails, we can safely return err and no sandbox is spawned. When it is moved to the last position, the request is processed first, and if there is an error in the catalog add, killing the sandbox to ensure consistency is more complicated.

packages/client-proxy/internal/edge-pass-through/proxy.go
sitole: I cleaned it a little bit. Create/delete is called from the switch and runs in separate method. Edge helpers renamed from "HandleEvent" to "ParseEvent".

packages/client-proxy/internal/edge-pass-through/proxy.go
```diff
@@ -135,6 +159,7 @@ func (s *NodePassThroughServer) Handler(srv interface{}, serverStream grpc.Serve
 				// to cancel the clientStream to the backend, let all of its goroutines be freed up by the CancelFunc and
 				// exit with an error to the stack
 				clientCancel()
+				returnErr = s2cErr
```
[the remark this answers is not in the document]
sitole: Not sure if this is cleaner

packages/api/internal/orchestrator/node.go
jakubno: so why do you return `metadata.NewOutgoingContext(ctx, n.clientMd)` instead of `ctx`?

packages/client-proxy/internal/edge-pass-through/proxy.go
```diff
@@ -232,50 +232,63 @@ func (s *NodePassThroughServer) forwardServerToClient(src grpc.ServerStream, dst
 }
 
 func (s *NodePassThroughServer) eventsHandler(md metadata.MD) (func(error), error) {
-	c, err := edge.HandleSandboxCatalogCreateEvent(md)
+	eventTypeHeaders := md.Get(edge.EventTypeHeader)
+	if len(eventTypeHeaders) == 0 {
```
jakubno: shouldn't you return error if the header has more than 1 element?

packages/api/internal/orchestrator/node.go
sitole: True, will return that

sitole: It may make sense to do it this way. Metadata for the local cluster is empty, but they are still filled on the node struct, so it's probably good to use them as expected.

packages/client-proxy/internal/edge-pass-through/proxy.go
sitole: Added err, that is clear for now.

packages/client-proxy/internal/edge-pass-through/proxy.go
```diff
@@ -204,3 +230,70 @@ func (s *NodePassThroughServer) forwardServerToClient(src grpc.ServerStream, dst
...
+	case edge.CatalogCreateEventType:
+		return s.catalogCreateEventHandler(md)
+	case edge.CatalogDeleteEventType:
+		return s.catalogDeleteEventHandler(md)
+	default:
+		return nil, status.Errorf(codes.InvalidArgument, "event type %s is not supported", eventType)
+	}
+}
+
+func (s *NodePassThroughServer) catalogCreateEventHandler(md metadata.MD) (func(error), error) {
```
jakubno: can we move this to separate file? 🙏🏻

packages/client-proxy/internal/edge-pass-through/proxy.go
```diff
@@ -103,14 +117,24 @@ func (s *NodePassThroughServer) Handler(srv interface{}, serverStream grpc.Serve
...
 
+	callback, err := s.eventsHandler(md)
+	if err != nil {
+		return status.Errorf(codes.Internal, "failed to handle events: %v", err)
+	}
+
+	var returnErr error
+	if callback != nil {
+		defer callback(returnErr)
+	}
```
jakubno: please handle this better, you are missing one error now

packages/api/internal/orchestrator/node.go
```diff
@@ -104,10 +107,10 @@ func (n *Node) setStatus(status api.NodeStatus) {
 	}
 }
 
-func (n *Node) setMetadata(i *orchestratorinfo.ServiceInfoResponse, nodeID string) {
+func (n *Node) setMetadata(i *orchestratorinfo.ServiceInfoResponse, nodeID string, serviceInstanceID string) {
```
jakubno: do you need to pass serviceInstanceID as extra argument instead of taking it from `ServiceInfoResponse`?

jakubno (changes requested):

packages/api/internal/orchestrator/node.go
```diff
@@ -30,9 +31,11 @@ type sbxInProgress struct {
 }
 
 type nodeMetadata struct {
-	orchestratorID string
-	commit         string
-	version        string
+	orchestratorID    string
+	serviceInstanceID string
```
jakubno: can you add comments what each represents 🙏🏻

jakubno (approved):
Lgtm

PR #905, opened 2025-07-22
Remove template manager health status

sitole:
- Removed now unused template manager method for health status.
- Renamed `orchestrator` prefixed info service status


packages/shared/pkg/grpc/orchestrator-info/info.pb.go
```diff
@@ -26,22 +26,22 @@ const (
...
+		1: "Draining",
+		2: "Unhealthy",
 	}
 	ServiceInfoStatus_value = map[string]int32{
-		"OrchestratorHealthy":   0,
-		"OrchestratorDraining":  1,
-		"OrchestratorUnhealthy": 2,
+		"Healthy":   0,
+		"Draining":  1,
+		"Unhealthy": 2,
```
sitole, line 44: These maps are not used because we are using a binary backend for the gRPC protocol, so just enum value numbers are sent, and they are still the same.

packages/orchestrator/internal/template/server/main.go
```diff
@@ -107,23 +110,16 @@ func (s *ServerStore) Close(ctx context.Context) error {
...
 
 		if !env.IsLocal() {
-			// give some time so all connected services can check build status
-			s.logger.Info("waiting before shutting down server")
+			// Give some time so all connected services can check build status
+			s.logger.Info("Waiting before shutting template builder down server")
 			time.Sleep(15 * time.Second)
 		}
+
+		s.logger.Info("Template builder shutdown complete")
```
sitole: The health status set is removed from here, so we are not overriding it between close methods. Now it's controlled by the main close handler.

sitole: After building jobs are done, we are still waiting 15s for the API to fetch the build status.

packages/orchestrator/internal/template/server/main.go
```diff
@@ -107,23 +110,16 @@ func (s *ServerStore) Close(ctx context.Context) error {
 	case <-ctx.Done():
 		return errors.New("context canceled during server graceful shutdown")
```
jakubno: this comment isn't true

packages/orchestrator/internal/template/server/main.go
jakubno: Depending on being 1st element in closers array seems to me to be little bit fragile. I would remove it from closers and call it explicitly (in main.go)

sitole: Changed, now if initialized it's closed before iterating over closers.

packages/orchestrator/internal/template/server/main.go
sitole: Main ctx is never calling done, so it should never happen. We can remove the whole select here.

packages/orchestrator/internal/template/server/main.go
sitole: Moved back after in-person discussions that order matters for all list items already.

packages/orchestrator/internal/template/server/main.go
```diff
@@ -103,27 +105,15 @@ func New(
...
-		s.logger.Info("waiting for all jobs to finish")
-		s.wg.Wait()
-
-		if !env.IsLocal() {
-			// give some time so all connected services can check build status
-			s.logger.Info("waiting before shutting down server")
-			time.Sleep(15 * time.Second)
-		}
-		return nil
+	s.logger.Info("Waiting for all build jobs to finish")
```
jakubno: you are ignoring `forceStop` was correctly there, but it should say, force exit, not waiting for builds` 
```go	
case <-ctx.Done():
		return errors.New("context canceled during server graceful shutdown")`
```

packages/orchestrator/internal/template/server/main.go
```diff
@@ -103,27 +105,15 @@ func New(
...
-	case <-ctx.Done():
-		return errors.New("context canceled during server graceful shutdown")
-	default:
-		// no new jobs should be started
-		s.logger.Info("marking service as draining")
-		s.healthStatus = templatemanager.HealthState_Draining
-		// wait for registering the node as draining
-		if !env.IsLocal() {
-			time.Sleep(5 * time.Second)
-		}
```
jakubno: you removed this part, i think you are risking a race condition here. There's no ongoing build, API thinks the builder is healthy and send a new request, you will accept it, but you won't await it anymore

packages/orchestrator/main.go
```diff
@@ -371,6 +378,9 @@ func run(port, proxyPort uint) (success bool) {
 		}
 	}
 
+	// Make service unhealthy after closing all services
+	serviceInfo.SetStatus(orchestrator.ServiceInfoStatus_Unhealthy)
```
jakubno: I think the builder becomes unhealthy right after you close store, once you will stop grpc, making it basically unhealthy (there's no way how to connect to server anymore)

packages/orchestrator/internal/template/server/main.go
```diff
@@ -105,25 +108,23 @@ func New(
...
 	default:
-		// no new jobs should be started
-		s.logger.Info("marking service as draining")
-		s.healthStatus = templatemanager.HealthState_Draining
-		// wait for registering the node as draining
+		// Wait for draining state to propagate to all consumers
 		if !env.IsLocal() {
-			time.Sleep(5 * time.Second)
+			time.Sleep(15 * time.Second)
 		}
```
jakubno, line 114: shouldn't this wait then be outside of the function? I think the store shouldn't care about it

packages/orchestrator/internal/template/server/main.go
```diff
@@ -105,25 +108,23 @@ func New(
...
+		s.logger.Info("Waiting for all build jobs to finish")
 		s.wg.Wait()
 
 		if !env.IsLocal() {
-			// give some time so all connected services can check build status
-			s.logger.Info("waiting before shutting down server")
+			// Give some time so all connected services can check build status
+			s.logger.Info("Waiting before shutting template builder down server")
 			time.Sleep(15 * time.Second)
 		}
```
jakubno: this isn't true

packages/orchestrator/internal/template/server/main.go
sitole: It's related only to the template manager and it will be used only when running the orchestrator as the template manager. The store should wait on it, because it's still time when the gRPC API should be working and resolving requests.

packages/orchestrator/internal/template/server/main.go
sitole: changed

PR #908, opened 2025-07-23
Allow specifying custom user for the RUN

dobrac:
Allow specifying custom user for the RUN `[command optional_user]`


dobrac:
> isn't this breaking change? shouldn't this be handled in API somehow, wouldn't it make the spec confusing otherwise?

it's not (we've never used more than the first arg) + the API spec defines only type and args in general

jakubno (approved):

PR #909, opened 2025-07-23
Allow specifying user and permissions for file COPY

dobrac:
Allow specifying user and permissions for file COPY. Format
`[localPath containerPath optional_owner optional_permissions]`

Owner can be empty to skip it.

The owner is in format of `chown` command, the permissions are in format of `chmod` command.


packages/orchestrator/internal/template/build/command/copy.go
```diff
@@ -172,5 +173,46 @@ func (c *Copy) Execute(
...
 
+	// If optional owner is provided, set them
+	if len(args) >= 3 {
+		// Assumes the format is "owner:group"
+		owner := args[2]
+		if owner != "" {
+			err = sandboxtools.RunCommand(
+				ctx,
+				tracer,
+
```
jakubno:
```suggestion
```

packages/orchestrator/internal/template/build/command/copy.go
```diff
@@ -172,5 +173,46 @@ func (c *Copy) Execute(
 		return sandboxtools.CommandMetadata{}, fmt.Errorf("failed to move files in sandbox: %w", err)
 	}
 
+	// If optional owner is provided, set them
+	if len(args) >= 3 {
+		// Assumes the format is "owner:group"
```
jakubno: shouldn't you validate then? Is the error verbose enough?

packages/orchestrator/internal/template/build/command/copy.go
```diff
@@ -172,5 +173,46 @@ func (c *Copy) Execute(
...
+			)
+			if err != nil {
+				return sandboxtools.CommandMetadata{}, fmt.Errorf("failed to set ownership and permissions in sandbox: %w", err)
+			}
+		}
+	}
+
+	// If optional permissions are provided, set them
+	if len(args) >= 4 {
+		// This assumes the format is "755"
```
jakubno: same here

packages/orchestrator/internal/template/build/command/copy.go
dobrac: update comment, I think the `chown` and `chmod` validation is enough

packages/orchestrator/internal/template/build/command/copy.go
```diff
@@ -172,5 +173,45 @@ func (c *Copy) Execute(
...
+	// If optional permissions are provided, set them
+	if len(args) >= 4 {
+		// This assumes the format of chmod
+		permissions := args[3]
+
+		if permissions != "" {
+			err = sandboxtools.RunCommand(
+				ctx,
+				tracer,
+
```
jakubno:
```suggestion
```

jakubno (approved):

packages/orchestrator/internal/template/build/command/copy.go
dobrac: 🤦

PR #913, opened 2025-07-24
Added fields to envd stat/list method response (#745)

jakubno:
# Description

Adds follwing fields for filesystem operation to envd
   - size
   - mode
   - permissions
   - owner
   - group
   - modified_time
   - symlink_target


packages/envd/internal/services/filesystem/utils.go
```diff
@@ -1,28 +1,88 @@
...
+// getFileOwnership returns the owner and group names for a file.
+// If the lookup fails, it returns the numeric UID and GID as strings.
+func getFileOwnership(fileInfo os.FileInfo) (owner, group string) {
+	sys, ok := fileInfo.Sys().(*syscall.Stat_t)
+	if !ok {
+		return "", ""
+	}
+
+	// Look up username
+	if u, err := user.LookupId(fmt.Sprintf("%d", sys.Uid)); err == nil {
```
dobrac: I think the `fmt.Sprintf("%d", sys.Uid)` can be set as default and rewritten if user.LookupId succeeds. The same applies for the group too

packages/envd/internal/services/filesystem/utils.go
```diff
@@ -1,28 +1,88 @@
...
+}
+
+func entryInfoFromFileInfo(fileInfo os.FileInfo, path string) *rpc.EntryInfo {
+	owner, group := getFileOwnership(fileInfo)
+	fileMode := fileInfo.Mode()
+
+	var symlinkTarget string
+	if fileMode&os.ModeSymlink != 0 {
+		// If we can't resolve the symlink target, we won't set the target
+		symlinkTarget, _ = followSymlink(path)
```
dobrac: can we remove resolving the symlink twice? (once here, once in the `getEntryType`)

packages/envd/spec/filesystem/filesystem.proto
```diff
@@ -52,6 +54,14 @@ message EntryInfo {
...
     FileType type = 2;
     string path = 3;
+    int64 size = 4;
+    uint32 mode = 5;
+    string permissions = 6;
+    string owner = 7;
+    string group = 8;
+    google.protobuf.Timestamp modified_time = 9;
+    // If the entry is a symlink, this field contains the target of the symlink.
+    string symlink_target = 10;
```
dobrac: shouldn't this be optional?

dobrac (changes requested):
few nits + can you please increase the envd version?

packages/envd/internal/services/filesystem/utils_test.go
```diff
@@ -25,19 +25,8 @@ func TestGetEntryType(t *testing.T) {
...
-	symlinkToDir := filepath.Join(tempDir, "symlink_to_dir")
-	require.NoError(t, os.Symlink(testDir, symlinkToDir))
-
-	// Create broken symlink
-	brokenSymlink := filepath.Join(tempDir, "broken_symlink")
-	require.NoError(t, os.Symlink("/nonexistent/path", brokenSymlink))
-
-	// Create cyclic symlink
-	cyclicSymlink := filepath.Join(tempDir, "cyclic_symlink")
-	require.NoError(t, os.Symlink(cyclicSymlink, cyclicSymlink))
```
dobrac: why removing all these tests?

dobrac (changes requested):
one thing and gtg

packages/envd/internal/services/filesystem/utils_test.go
jakubno: they were kinda duplicated, the logic was removed from the helper funtiion

dobrac: ok

dobrac (approved):

PR #917, opened 2025-07-25
Remove old unused terraform modules

jakubno:
# Description

Removing old terraform modules as they aren't maintained anymore and they are obsolete


terraform/github/main.tf
```diff
@@ -1,177 +0,0 @@
...
-  }
-}
-
-data "google_project" "gcp_project" {}
-
-resource "google_service_account_iam_member" "gha_service_account_wif_tokencreator_iam_member" {
-  service_account_id = google_service_account.github_action_service_account.name
-  role               = "roles/iam.workloadIdentityUser"
-  member             = "principalSet://iam.googleapis.com/projects/${data.google_project.gcp_project.number}/locations/global/workloadIdentityPools/${google_iam_workload_identity_pool.github_actions_wip.workload_identity_pool_id}/attribute.repository/${var.github_organization}/${var.github_repository}"
-}
```
dobrac, line 90: this identity might still be used in GH actions

Makefile
```diff
@@ -34,7 +34,6 @@ tf_vars := 	TF_VAR_environment=$(TERRAFORM_ENVIRONMENT) \
 	$(call tfvar, DOMAIN_NAME) \
 	$(call tfvar, ADDITIONAL_DOMAINS) \
 	$(call tfvar, PREFIX) \
-	$(call tfvar, TERRAFORM_STATE_BUCKET) \
```
dobrac, line 37: is this not used anymore?

terraform/github/main.tf
jakubno: it isn't

Makefile
jakubno: not in `tfvars`, it's only used for terraform init

PR #918, opened 2025-07-25
Allow extending existing sandbox templates

dobrac:
Allow extending existing sandbox templates when creating a new template. This includes also the start and ready commands.


packages/api/internal/template-manager/create_template.go
```diff
@@ -80,30 +81,67 @@ func (tm *TemplateManager) CreateTemplate(
...
+		FirecrackerVersion: firecrackerVersion,
+		HugePages:          features.HasHugePages(),
+		StartCommand:       startCmd,
+		ReadyCommand:       readyCmd,
+		Force:              force,
+		Steps:              convertTemplateSteps(steps),
+	}
+
+	// Set the source (either fromImage or fromTemplate)
+	if fromTemplate != nil && *fromTemplate != "" {
```
jakubno: can you check if both are provided?

packages/orchestrator/internal/template/build/builder.go
```diff
@@ -123,7 +123,7 @@ func (b *Builder) Build(ctx context.Context, finalMetadata storage.TemplateFiles
 	// Validate template, update force layers if needed
 	template = forceSteps(template)
 
-	isOldBuild := template.FromImage == ""
+	isOldBuild := template.FromImage == "" && template.FromTemplate == nil
```
jakubno: How about renaming it to `isV1Build`? I think this would make the variable's purpose more obvious.

packages/orchestrator/internal/template/build/builder.go
```diff
@@ -588,6 +590,19 @@ func (b *Builder) setupBase(
...
+	// If the template is built from another template, use its metadata always
+	if template.FromTemplate != nil {
+		baseMetadata = LayerMetadata{
+			Template: storage.TemplateFiles{
+				TemplateID:         template.FromTemplate.GetTemplateID(),
+				BuildID:            template.FromTemplate.GetBuildID(),
+				KernelVersion:      template.FromTemplate.GetKernelVersion(),
+				FirecrackerVersion: template.FromTemplate.GetFirecrackerVersion(),
+			},
+			Metadata: cmdMeta,
```
jakubno: do you need to do whole "base setup", e.g. run provisioning script if you do from template?

jakubno (changes requested):

packages/orchestrator/internal/template/build/builder.go
dobrac: yes, thats good

packages/orchestrator/internal/template/build/builder.go
dobrac: no I don't, and actually it needs to always return that it's cached, otherwise the base setup doesn't even make any sense

I've refactored the setup a bit

jakubno: What about `cmdMeta` should they be based on the template?

dobrac: good catch, I need to rethink how to get these

dobrac: refactored by saving the template metadata with the template files in `metadata.json`. This file includes the start/ready commands, as well as the context (workdir, user, etc) for further execution

packages/orchestrator/internal/template/build/builder.go
```diff
@@ -448,8 +449,27 @@ func (b *Builder) Build(ctx context.Context, finalMetadata storage.TemplateFiles
...
+	var startMetadata *StartMetadata
+	if template.StartCmd != "" || template.ReadyCmd != "" {
+		startMetadata = &StartMetadata{
+			StartCmd: template.StartCmd,
+			ReadyCmd: template.ReadyCmd,
+			Metadata: sourceMetadata.Metadata,
+		}
+	}
+
+	// If the template is built from another template, and the start metadata are not set,
```
jakubno, line 461: can you "clear" the start command?

packages/orchestrator/internal/template/build/builder.go
```diff
@@ -661,14 +736,20 @@ func (b *Builder) postProcessingFn(
 		}
 
 		// Ready command
+		readyCmd := start.ReadyCmd
+		if readyCmd == "" {
+			if start.StartCmd == "" {
```
jakubno, line 741: just curious here, can it happen that `start` is not `nil`, but `StartCmd` and `ReadyCmd` are both `""`?

packages/orchestrator/internal/template/build/hash_index.go
```diff
@@ -89,7 +89,16 @@ func hashBase(template config.TemplateConfig) (string, error) {
 		return "", fmt.Errorf("error getting envd binary hash: %w", err)
 	}
 
-	return hashKeys(hashingVersion, envdHash, provisionScriptFile, strconv.FormatInt(template.DiskSizeMB, 10), template.FromImage), nil
+	var baseSource string
+	if template.FromTemplate != nil {
+		// When building from template, use the base template's ID and build ID as the source
+		baseSource = fmt.Sprintf("template:%s", template.FromTemplate.GetBuildID())
```
jakubno: you are using only buildID here

packages/orchestrator/internal/template/build/hash_index.go
```diff
@@ -89,7 +89,16 @@ func hashBase(template config.TemplateConfig) (string, error) {
...
+	} else {
+		// When building from image, use the image name
+		baseSource = template.FromImage
```
jakubno, line 103: this one is little bit annoying, but couldn't this cause some issues when the cache is shared and even though you don't force rebuild it can use different base, because you use `latest`? My problem here is, in logs we say it's cached, but it isn't in a way you would probably expect. But it's okay to ignore

packages/orchestrator/internal/template/build/builder.go
dobrac: only by setting something dummy like `exit 0`

packages/orchestrator/internal/template/build/builder.go
dobrac: no, it shouldn't happen right now with the flow

but, I don't think you know it when you are in this function at that point

packages/orchestrator/internal/template/build/hash_index.go
dobrac: right now the sharing is only per team (this would need to be changed to have it globally), but I understand the concern. 

If we need global cache or this becomes an issue, the solution would be to include the image hash

packages/orchestrator/internal/template/build/builder.go
jakubno: I have nothing against being defensive, I just wnated to make sure I am not missing something

packages/orchestrator/internal/template/build/hash_index.go
jakubno: Could you please add a comment to warn about this behavior? It might have happened before, but since we moved this to our side, it's not as clear.

.github/workflows/integration_tests.yml
```diff
@@ -32,17 +32,19 @@ jobs:
           TESTS_ORCHESTRATOR_HOST: "localhost:5008"
           TESTS_ENVD_PROXY: "http://localhost:5007"
         run: |
-            # Monitor logs of the services
```
jakubno, line 35: did you remove the logs on purpose?

packages/orchestrator/internal/template/build/hash_index.go
dobrac: added

tests/integration/internal/tests/api/templates/build_template_test.go
```diff
@@ -289,3 +303,530 @@ func TestTemplateBuildCache(t *testing.T) {
...
+			// First build the base template
+			assert.True(t, buildTemplate(t, tc.baseTemplate, tc.baseConfig, defaultBuildLogHandler(t)))
+
+			// Then build the derived template from the base template
+			assert.True(t, buildTemplate(t, tc.derivedTemplate, tc.derivedConfig, defaultBuildLogHandler(t)))
+		})
+	}
+}
+
+func TestTemplateBuildFromTemplateInheritance(t *testing.T) {
```
jakubno, line 437: taking into account how long it takes, can you check all 3 in one tests?

tests/integration/internal/tests/api/templates/build_template_test.go
```diff
@@ -289,3 +303,530 @@ func TestTemplateBuildCache(t *testing.T) {
...
+			// First build the base template
+			assert.True(t, buildTemplate(t, tc.baseTemplate, tc.baseConfig, defaultBuildLogHandler(t)))
+
+			// Then build the derived template from the base template
+			assert.True(t, buildTemplate(t, tc.derivedTemplate, tc.derivedConfig, defaultBuildLogHandler(t)))
+		})
+	}
+}
+
+func TestTemplateBuildFromTemplateBaseCommandsInheritance(t *testing.T) {
```
jakubno, line 544: how are the 2 tests different?

tests/integration/internal/tests/api/templates/build_template_test.go
```diff
@@ -289,3 +303,530 @@ func TestTemplateBuildCache(t *testing.T) {
...
+			// First build the base template
+			assert.True(t, buildTemplate(t, tc.baseTemplate, tc.baseConfig, defaultBuildLogHandler(t)))
+
+			// Then build the derived template from the base template
+			assert.True(t, buildTemplate(t, tc.derivedTemplate, tc.derivedConfig, defaultBuildLogHandler(t)))
+		})
+	}
+}
+
+func TestTemplateBuildFromTemplateStartCommand(t *testing.T) {
```
jakubno, line 498: these could be probably tested together as well

jakubno (changes requested):
lgtm, small nits

.github/workflows/integration_tests.yml
dobrac: yes, I've removed them as the tests logs run was not loading at all because of the amount. They're now included in the step as an artifact

do you have any better idea how to handle them?

tests/integration/internal/tests/api/templates/build_template_test.go
dobrac: removed the second one

tests/integration/internal/tests/api/templates/build_template_test.go
dobrac: it's not really that slow now, the base is always cached

tests/integration/internal/tests/api/templates/build_template_test.go
dobrac: the tests are not really that slow now, the base is always cached

.github/workflows/integration_tests.yml
jakubno: it's fine, as long you can download them, they aren't that important and it should make finding failed test easier

PR #921, opened 2025-07-28
Allow to block network from SDK (API part)

jakubno:
# Description
This pull request introduces a new feature to allow sandboxes to access the internet.

For resumed sandbox this option is inherited


packages/api/internal/handlers/sandbox_create.go
```diff
@@ -154,6 +154,8 @@ func (a *APIStore) PostSandboxes(c *gin.Context) {
 		envdAccessToken = &accessToken
 	}
 
+	allowInternetAccess := body.AllowInternetAccess
```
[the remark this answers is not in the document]
dobrac: doesn't make sense, it's nillable

packages/db/migrations/20250728085406_add_create_network_flag.sql
```diff
@@ -0,0 +1,19 @@
+-- +goose Up
+-- +goose StatementBegin
+BEGIN;
```
dobrac: this might be unnecessary

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -43,9 +42,9 @@ func (s *server) Create(ctxConn context.Context, req *orchestrator.SandboxCreate
 		attribute.String("envd.version", req.Sandbox.EnvdVersion),
 	)
 
-	// TODO: Temporary workaround, remove API changes deployed
-	if req.Sandbox.GetExecutionId() == "" {
-		req.Sandbox.ExecutionId = uuid.New().String()
+	// TODO: Temporary workaround, remove when API changes deployed
```
dobrac:
```suggestion
	// TODO: Temporary workaround, remove when API changes are deployed
```

packages/orchestrator/internal/template/build/build_layer.go
```diff
@@ -139,7 +139,6 @@ func (b *Builder) buildLayer(
 				KernelLogs:          env.IsDevelopment(),
 				SystemdToKernelLogs: false,
 			},
-			allowInternet,
```
dobrac, line 142: missing in sbxConfig definition

packages/shared/pkg/schema/snapshots.go
```diff
@@ -29,6 +29,7 @@ func (Snapshot) Fields() []ent.Field {
 		field.Time("sandbox_started_at"),
 		field.Bool("env_secure").Default(false),
 		field.String("origin_node_id").SchemaType(map[string]string{dialect.Postgres: "text"}),
+		field.Bool("allow_internet_access").Nillable().Optional(),
```
dobrac, line 32: sqlc is better, but can be done in different PR

dobrac (changes requested):

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -86,14 +86,13 @@ func CreateSandbox(
...
 	processOptions fc.ProcessOptions,
-	allowInternet bool,
 ) (*Sandbox, *Cleanup, error) {
 	childCtx, childSpan := tracer.Start(ctx, "new-sandbox")
 	defer childSpan.End()
 
 	cleanup := NewCleanup()
 
-	ipsCh := getNetworkSlotAsync(childCtx, tracer, networkPool, cleanup, allowInternet)
+	ipsCh := getNetworkSlotAsync(childCtx, tracer, networkPool, cleanup, config.GetAllowInternetAccess())
```
dobrac, line 96: just a note, this will default to false, but it should probably default to the config value or true

dobrac (changes requested):

packages/orchestrator/internal/sandbox/sandbox.go
jakubno: It's set in all places to the global config or `true`

dobrac: yes, but you never know in new usages 👀

PR #923, opened 2025-07-28
add metrics for sandbox creation

djeebus:
this allows us to track:
* count of sandboxes scheduled
* distribution of attempts to schedule
* resume vs create counts
* success vs failed attempts to use node affinity


packages/api/internal/orchestrator/create_instance.go
```diff
@@ -209,6 +210,14 @@ func (o *Orchestrator) CreateSandbox(
...
 		if err == nil {
 			// The sandbox was created successfully
+			attributes := []attribute.KeyValue{
+				attribute.Int("attempts", attempt),
+				attribute.String("node_id", node.Info.ID),
+				attribute.Bool("is_resume", isResume),
+				attribute.Bool("node_affinity_requested", nodeID != nil),
+				attribute.Bool("node_affinity_success", nodeID != nil && node.Info.ID == *nodeID),
+			}
+			o.createdSandboxesCounter.Add(ctx, 1, metric.WithAttributes(attributes...))
```
jakubno, line 219: won't this have huge cardinality?

djeebus: it'll be ~ `len(nodes) * 2 * 2 * 2` plus a little bit more for sandboxes that take more than one attempt to get assigned. at present size, that's something like ~400 series, give or take. probably fine if the data's useful, but I don't have a problem removing the node_id tag, which brings us down to ~8 (plus a bit). keeping the data might be useful for identifying hot spots, but not likely. we're currently at ~12000 series in foxtrot, so this would represent a ~3% increase. I'll remove it 👍

djeebus: oh, actually it's greater than that. I forgot about the `instance` that's attachd, so it'll be `<above> * len(api_nodes)`. not sure how many that is, but it probably doesn't matter. Removing!

dobrac: you can actually test this and debug in Grafana on your cluster. There should be a dashboard where you can see the cardinality of the metric

djeebus: I don't have enough data in my sandbox for that to be useful. It does claims that `otelcol_processor_batch_batch_send_size_bucket` has the most series, at 184 😁

jakubno: That's fine, i just quickly go over it and I thought there the requested node and node id making it n^2. My bad. Also grafana should now allows to drop attributes during ingestion via UI, so it shouldn't be such a problem

packages/api/internal/orchestrator/metrics.go
```diff
@@ -53,8 +53,12 @@ func (o *Orchestrator) setupMetrics(meterProvider metric.MeterProvider) (metric.
 			return nil
 		}, gauge)
 	if err != nil {
-		return nil, fmt.Errorf("failed to register orchestrators gauge: %w", err)
+		return fmt.Errorf("failed to register orchestrators gauge: %w", err)
 	}
 
-	return registration, nil
+	if o.createdSandboxesCounter, err = meter.Int64Counter("api.orchestrator.created_sandboxes"); err != nil {
```
jakubno: can you move the meter to `packages/shared/pkg/telemetry/meters.go`, we have other meters defined there?

jakubno (commented):
Otherwise look good

djeebus:
> I am thinking how is it different from `api.orchestrator.sandbox.create.success` and `api.orchestrator.sandbox.create.failure`. Wouldn't it make sense to just "enrich" that metric?

This meter looks like it's calculated on a timer from in-memory data. We would need to add a multi-dimensional matrix to the `Node` struct in order to track counts based on attempts/is_resume/node_affinity_requested/node_affinity_success, which is effectively what otel is doing (collecting the series in memory, and allowing them to be scraped every 30 or 60 seconds). I don't know why we're observing metrics this way, so I can't say if it's worth the effort to build up the memory structures to avoid using otel's code.

jakubno (approved):

PR #924, opened 2025-07-28
Fix envd symlink resolution

jakubno:
# Description

os.Stat gives information about resolved symlinks, that caused issues with `symlinkTarget` for `Stat` method.
Refactor the entryInfo method to prevent such bugs

### DIFF
--- a/packages/envd/internal/services/filesystem/dir.go
+++ b/packages/envd/internal/services/filesystem/dir.go
@@ -156,14 +156,16 @@ func walkDir(requestedPath string, dirPath string, depth int) (entries []*rpc.En
 			return filepath.SkipDir
 		}
 
-		fileInfo, err := entry.Info()
-		if err != nil {
+		entryInfo, err := entryInfo(path)
+		if entry == nil {
 			return err
 		}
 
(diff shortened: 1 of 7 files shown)


packages/envd/internal/services/filesystem/dir.go
```diff
@@ -156,14 +156,16 @@ func walkDir(requestedPath string, dirPath string, depth int) (entries []*rpc.En
...
-		if err != nil {
+		entryInfo, err := entryInfo(path)
+		if entry == nil {
 			return err
 		}
 
 		// Return the requested path as the base path instead of the symlink-resolved path
 		path = filepath.Join(requestedPath, relPath)
-		entries = append(entries, entryInfoFromFileInfo(fileInfo, path))
+		entryInfo.Path = path
```
dobrac, line 166: can we avoid this somehow?

dobrac (changes requested):

packages/envd/internal/services/filesystem/dir.go
jakubno: Not really, if we want to return "requested" path

dobrac (approved):

PR #929, opened 2025-07-29
Fix rename file if doesn't exist to return correct error

jakubno:
# Description

If file doesn't exist it should return `CodeNotFound` error


ben-fornefeld (approved):
lgtm

PR #941, opened 2025-08-01
Update envd on first layer build

dobrac:
Update envd on first layer build (sandbox create). This removes the envd dependency in provisioning which improves caches invalidation on envd update.

Also, this change significantly speeds up envd updates on template rebuild.

The envd replacement takes around 300ms.

### DIFF
--- a/packages/orchestrator/internal/template/build/builder.go
+++ b/packages/orchestrator/internal/template/build/builder.go
@@ -163,7 +163,7 @@ func (b *Builder) Build(ctx context.Context, finalMetadata storage.TemplateFiles
 		return nil, fmt.Errorf("error getting base hash: %w", err)
 	}
 
-	lastCached, baseMetadata, err := b.setupBase(ctx, cacheScope, finalMetadata, template, lastHash, isV1Build)
+	isLastLayerCached, baseMetadata, err := b.setupBase(ctx, cacheScope, finalMetadata, template, lastHash, isV1Build)
 	if err != nil {
 		return nil, fmt.Errorf("error setting up build: %w", err)
 	}
@@ -183,10 +183,10 @@ func (b *Builder) Build(ctx context.Context, finalMetadata storage.TemplateFiles
 		}
 		baseSource = "FROM " + fromImage
 	}
-	postProcessor.Info(layerInfo(lastCached, "base", baseSource, lastHash))
+	postProcessor.Info(layerInfo(isLastLayerCached, "base", baseSource, lastHash))
 
 	// Build the base layer if not cached
-	if !lastCached {
+	if !isLastLayerCached {
 		templateBuildDir := filepath.Join(templatesDirectory, finalMetadata.BuildID)
 		err = os.MkdirAll(templateBuildDir, 0o777)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


packages/orchestrator/internal/template/build/builder.go
```diff
@@ -412,6 +412,7 @@ func (b *Builder) Build(ctx context.Context, finalMetadata storage.TemplateFiles
 				sourceMetadata,
 				stepMetadata.Template,
 				shouldResumeSandbox,
+				!shouldResumeSandbox,
```
jakubno: can you define var `updateEnvd` so it's easier to read? 🙏🏻

packages/orchestrator/internal/template/build/build_layer.go
```diff
@@ -36,6 +37,7 @@ func (b *Builder) buildLayer(
 	sourceMeta LayerMetadata,
 	exportTemplate storage.TemplateFiles,
 	resumeSandbox bool,
+	updateEnvd bool,
```
jakubno, line 40: is it ever True?

packages/orchestrator/internal/template/build/builder.go
```diff
@@ -488,6 +489,7 @@ func (b *Builder) Build(ctx context.Context, finalMetadata storage.TemplateFiles
 		sourceMetadata,
 		finalMetadata,
 		false,
+		!shouldResumeSandbox,
```
jakubno: this is little bit confusing as `shouldResumeSandbox` here doesn't really do what it says

packages/orchestrator/internal/template/build/build_layer.go
dobrac: yes, for the first uncached layer

packages/orchestrator/internal/template/build/builder.go
dobrac: I've reverted it to the isLastLayerCached, I think it's much clearer

packages/orchestrator/internal/template/build/builder.go
dobrac: Reverted it to the isLastLayerCached

packages/orchestrator/internal/template/build/builder.go
```diff
@@ -474,8 +468,6 @@ func (b *Builder) Build(ctx context.Context, finalMetadata storage.TemplateFiles
 		postProcessor,
 		uploadErrGroup,
 		&orchestrator.SandboxConfig{
-			BaseTemplateId: baseMetadata.Template.TemplateID,
```
dobrac: this was unused

packages/orchestrator/internal/template/build/builder.go
```diff
@@ -437,14 +435,10 @@ func (b *Builder) Build(ctx context.Context, finalMetadata storage.TemplateFiles
 				return nil, fmt.Errorf("error running build layer: %w", err)
 			}
 			stepMetadata = meta
-
-			if !shouldResumeSandbox {
-				baseMetadata = stepMetadata
```
dobrac, line 442: it doesn't make much sense, also makes the code quite complicated

I'll revisit this in a future PR as I've found a bug in some of the metadata passing

PR #942, opened 2025-08-01
Deduplicate requests to the UFFD

ValentaTomas:
Especially for larger memfiles, numerous `EEXIST` errors occur during the UFFD COPY, as the page has already been served.

In some extreme cases (looks like to be caused by either slower download or concurrency), we get to 95% of UFFD events being marked as `EEXIST`.

This deduplicated the calls, so we don't spawn a new goroutine to serve these pages.

Some very minor gains (<1ms) in the worst cases could be gained by switching from map to bitset.
This does not need a lock, because the FD reading happens in a single thread.


jakubno:
Have you run any bencharks before / after?

PR #951, opened 2025-08-04
Sandbox start rate and currently running sandboxes

jakubno:


packages/clickhouse/migrations/20250801113225_sandbox_filter.sql
jakubno, line 1: changed the routing so we can distinguish based on the metric name and not on the attributes (the condition before is always true

```
SELECT (CAST(map(), 'Map(String, String)')['sandbox_id']) IS NOT NULL AS empty_map

Query id: d67ff78d-509e-4c76-86e3-a2fba86c00fc

   ┌─empty_map─┐
1. │         1 │
   └───────────┘
   ```

File: packages/clickhouse/migrations/20250801113226_team_counters.sql
Comment (jakubno) [inline packages/clickhouse/migrations/20250801113226_team_counters.sql:1]: Table for counters

File: spec/openapi.yml [hunk 1]
```diff
@@ -1347,6 +1367,43 @@ paths:
         "404":
           $ref: "#/components/responses/404"
 
+  /teams/metrics:
```
dobrac: should this endpoint be accessible with the API Key?

Also, do we want to keep the API more REST-like? (then having `/teams/{teamID}/metrics`

packages/shared/pkg/telemetry/meters.go
```diff
@@ -53,20 +56,22 @@ const (
...
 	ApiOrchestratorCreatedSandboxes: "Number of successfully created sandboxes",
+
+	TeamSandboxCreated: "Counter of started sandboxes for the team in the interval",
 }
 
 var counterUnits = map[CounterType]string{
 	SandboxCreateMeterName:          "{sandbox}",
 	ApiOrchestratorCreatedSandboxes: "{sandbox}",
+
+	TeamSandboxCreated: "1",
```
dobrac: why is it "1"?

packages/shared/pkg/telemetry/meters.go
```diff
@@ -128,13 +133,15 @@ var gaugeIntDesc = map[GaugeIntType]string{
...
 	SandboxCpuTotalGaugeName:      "Amount of CPU available to the sandbox.",
+	TeamSandboxMaxGaugeName:       "Maximum number of sandboxes running for the team in the interval.",
 }
 
 var gaugeIntUnits = map[GaugeIntType]string{
 	ApiOrchestratorCountMeterName: "{orchestrator}",
 	SandboxRamUsedGaugeName:       "{By}",
 	SandboxRamTotalGaugeName:      "{By}",
 	SandboxCpuTotalGaugeName:      "{count}",
+	TeamSandboxMaxGaugeName:       "1",
```
dobrac: same here, why "1"?

packages/api/internal/orchestrator/cache.go
```diff
@@ -326,6 +326,7 @@ func (o *Orchestrator) getDeleteInstanceFunction(
 		node.RamUsage.Add(-info.RamMB)
 
 		o.dns.Remove(ctx, info.Instance.SandboxID, node.Info.IPAddress)
+		o.teamMetricsObserver.Remove(info.TeamID)
```
dobrac: how will this work with multiple API instances?

packages/shared/pkg/telemetry/meters.go
jakubno: https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-units

packages/shared/pkg/telemetry/meters.go
jakubno: https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-units

packages/api/internal/orchestrator/cache.go
jakubno: Started will work because we will just sum it, max will use the same logic as the check when creating sandbox

spec/openapi.yml
jakubno: We can do that, it's just annoying that you have two team IDs which you have to compare

packages/api/internal/handlers/team_metrics.go
```diff
@@ -0,0 +1,100 @@
...
+	if teamID != team.ID.String() {
+		zap.L().Warn("user tried to access metrics for a team they are not authorized to access", logger.WithTeamID(team.ID.String()), zap.String("requested_team_id", teamID))
+		a.sendAPIStoreError(c, http.StatusForbidden, fmt.Sprintf("You (%s) are not authorized to access this team's (%s) metrics", team.ID, teamID))
+
+		return
+	}
+
+	metricsReadFlag, err := a.featureFlags.BoolFlag(featureflags.MetricsReadFlagName, team.ID.String())
+	if err != nil {
+		zap.L().Error("error getting metrics read feature flag, soft failing", zap.Error(err))
```
dobrac:
```suggestion
		zap.L().Warn("error getting metrics read feature flag, soft failing", zap.Error(err))
```

packages/api/internal/handlers/team_metrics.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	if !metricsReadFlag {
+		zap.L().Debug("sandbox metrics read feature flag is disabled")
+		// If we are not reading from ClickHouse, we can return an empty map
+		// This is here just to have the possibility to turn off ClickHouse metrics reading
+
+		c.JSON(http.StatusOK, []api.TeamMetric{})
```
dobrac, line 44: is this status OK?

packages/api/internal/handlers/team_metrics.go
```diff
@@ -0,0 +1,100 @@
...
+		return
+	}
+
+	// Default time range is the last 7 days
+	start, end := time.Now().Add(-7*24*time.Hour), time.Now()
```
dobrac: move the default to const

packages/api/internal/handlers/team_metrics.go
```diff
@@ -0,0 +1,100 @@
...
+		step = 30 * time.Second
+	case duration < 12*time.Hour:
+		step = time.Minute
+	case duration < 24*time.Hour:
+		step = 2 * time.Minute
+	case duration < 7*24*time.Hour:
+		step = 5 * time.Minute
+	default:
+		step = 15 * time.Minute
+	}
```
dobrac: can you make a helper function for this conversion?

packages/api/internal/handlers/team_metrics.go
```diff
@@ -0,0 +1,100 @@
...
+
+	metrics, err := a.clickhouseStore.QueryTeamMetrics(ctx, team.ID.String(), start, end, step)
```
dobrac:
```suggestion
	metrics, err := a.clickhouseStore.QueryTeamMetrics(ctx, teamID, start, end, step)
```

packages/clickhouse/pkg/team.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+const teamMetricsSelectQuery = `
+WITH
+  created AS (
+    SELECT
+      toStartOfInterval(timestamp, interval {step:UInt32} second) AS ts,
+      sum(value) as created_sandboxes
+    FROM team_metrics_sum
+    WHERE metric_name = 'e2b.team.sandbox.created'
```
dobrac: can we make the `e2b.team` constant somewhere?

Also, it's kinda unfortunate there is no DSL for the SQL here, as the inline queries will be difficult to manage in the future

packages/api/internal/orchestrator/cache.go
dobrac: the issue is that the counts can be inaccurate now when you're deploying a new API - because all APIs keep all sandboxes in the cache

also, I'm not sure how would this play out with the API cache synchronizations. Basically same issues as we've had with the currently running sbx counter

dobrac (changes requested):

packages/api/internal/orchestrator/cache.go
jakubno: It isn't the same. The problem with old metrics was it was cumulative and some sbxs were loaded before the otel was connected.

packages/api/internal/handlers/team_metrics.go
jakubno: yes, it's expected when the reading is turned off. I think we may remove it now

packages/api/internal/orchestrator/cache.go
dobrac: this is synchronous for pause/kill

packages/api/internal/orchestrator/cache.go
```diff
@@ -429,6 +430,8 @@ func (o *Orchestrator) getInsertInstanceFunction(parentCtx context.Context, time
 			o.dns.Add(ctx, info.Instance.SandboxID, node.Info.IPAddress)
 		}
 
+		o.teamMetricsObserver.Add(ctx, info.TeamID, created)
```
dobrac, line 435: this should be now async, but If we want to have it always async, but I'm not sure if it shouldn't be at least after the markAsPausing

packages/clickhouse/pkg/team.go
jakubno: sqlc has to catch up, but the issue looks little bit stale
https://github.com/sqlc-dev/sqlc/issues/1628

jakubno:
@cursor review

jakubno:
@cursor review

packages/api/internal/orchestrator/cache.go
jakubno: it isn't blocking anymore

packages/api/internal/handlers/team_metrics.go
```diff
@@ -0,0 +1,87 @@
...
+func (a *APIStore) GetTeamsTeamIDMetrics(c *gin.Context, teamID string, params api.GetTeamsTeamIDMetricsParams) {
+	ctx := c.Request.Context()
+	ctx, span := a.Tracer.Start(ctx, "sandbox-metrics")
+	defer span.End()
+
+	team := c.Value(auth.TeamContextKey).(authcache.AuthTeamInfo).Team
+
+	if teamID != team.ID.String() {
+		zap.L().Warn("user tried to access metrics for a team they are not authorized to access", logger.WithTeamID(team.ID.String()), zap.String("requested_team_id", teamID))
+		a.sendAPIStoreError(c, http.StatusForbidden, fmt.Sprintf("You (%s) are not authorized to access this team's (%s) metrics", team.ID, teamID))
```
dobrac, line 29: should we use also these for traces? You can then remove the zap.L().Warn

telemetry.ReportCriticalError(ctx, msg, err) / ReportError

tests/integration/internal/tests/api/team_metrics_test.go
```diff
@@ -0,0 +1,124 @@
...
+	c := setup.GetAPIClient()
+
+	// Create multiple sandboxes to generate team metrics
+	utils.SetupSandboxWithCleanup(t, c)
+	utils.SetupSandboxWithCleanup(t, c)
+
+	// Wait a bit to ensure metrics are generated
+	maxRetries := 15
+	var metrics []api.TeamMetric
+	for i := 0; i < maxRetries; i++ {
```
dobrac: can we make a helper for this? It's in many places now

tests/integration/internal/tests/api/team_metrics_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+func TestTeamMetricsEmpty(t *testing.T) {
+	c := setup.GetAPIClient()
+
+	// Test getting metrics for a time range where no sandboxes existed
+	now := time.Now()
+	start := now.Add(-240 * time.Hour).Unix()
+	end := now.Add(-216 * time.Hour).Unix()
+
+	// Wait a bit to ensure metrics are generated
```
dobrac: where do we wait here?

dobrac (changes requested):
small nits

packages/api/internal/handlers/team_metrics.go
jakubno: these what?

tests/integration/internal/tests/api/team_metrics_test.go
jakubno: in the loop, i will improve the comment

packages/api/internal/handlers/team_metrics.go
dobrac: `telemetry.ReportCriticalError(ctx, msg, err)` or `telemetry.ReportError`

tests/integration/internal/tests/api/team_metrics_test.go
dobrac: but there is no loop 👀

jakubno: copy pasta 🙈

tests/integration/internal/tests/api/sandbox_list_metrics_test.go
```diff
@@ -20,26 +20,24 @@ func TestSandboxListMetrics(t *testing.T) {
 	sbx1 := utils.SetupSandboxWithCleanup(t, c)
 	sbx2 := utils.SetupSandboxWithCleanup(t, c)
 
-	maxRetries := 15
+	maxDuration := 15 * time.Second
+	tick := 500 * time.Millisecond
 	var metrics map[string]api.SandboxMetric
 
-	for i := 0; i < maxRetries; i++ {
+	require.Eventually(t, func() bool {
```
dobrac, line 27: cool

packages/api/internal/handlers/team_metrics.go
```diff
@@ -0,0 +1,88 @@
...
+		start = time.Unix(*params.Start, 0)
+	}
+
+	if params.End != nil {
+		end = time.Unix(*params.End, 0)
+	}
+
+	// Validate time range parameters
+	if start.After(end) {
+		a.sendAPIStoreError(c, http.StatusBadRequest, "start time cannot be after end time")
```
dobrac, line 61: add telemetry also here

packages/api/internal/handlers/team_metrics.go
```diff
@@ -0,0 +1,88 @@
...
+	step := calculateStep(start, end)
+
+	metrics, err := a.clickhouseStore.QueryTeamMetrics(ctx, teamID, start, end, step)
+	if err != nil {
+		zap.L().Error("Error fetching sandbox metrics from ClickHouse",
+			logger.WithTeamID(team.ID.String()),
+			zap.Error(err),
+		)
+
+		a.sendAPIStoreError(c, http.StatusInternalServerError, fmt.Sprintf("error querying sandbox metrics: %s", err))
```
dobrac: add telemetry also here

packages/api/internal/handlers/team_metrics.go
```diff
@@ -0,0 +1,90 @@
...
+
+	step := calculateStep(start, end)
+
+	metrics, err := a.clickhouseStore.QueryTeamMetrics(ctx, teamID, start, end, step)
+	if err != nil {
+		zap.L().Error("Error fetching team metrics from ClickHouse",
+			logger.WithTeamID(team.ID.String()),
+			zap.Error(err),
+		)
+
```
dobrac:
```suggestion
```

PR #952, opened 2025-08-04
Stop using Nomad Short NodeID in API internals

sitole:
- Handle build cache cleanup when zero keys are not found (fix for AWS storage provider).
- Use the cluster template manager when requesting a file upload.
- Stores nodeID for local template builder, so we are able to support multiple local builders.
- Stop using Nomad short ID as the orchestrator primary identifier in the API.


packages/api/internal/handlers/template_request_build.go
```diff
@@ -422,23 +422,11 @@ func (a *APIStore) TemplateRequestBuild(c *gin.Context, templateID api.TemplateI
...
-			a.sendAPIStoreError(c, http.StatusInternalServerError, fmt.Sprintf("Error when getting available template builder: %s", err))
-			telemetry.ReportCriticalError(ctx, "error when getting available template builder", err, telemetry.WithTemplateID(templateID))
-			return nil
-		}
-
-		builderNodeID = &clusterNode.NodeID
+	builderNodeID, err := a.templateManager.GetAvailableBuildClient(ctx, team.ClusterID)
+	if err != nil {
+		telemetry.ReportCriticalError(ctx, "error when getting available build client", err, telemetry.WithTemplateID(templateID))
+		a.sendAPIStoreError(c, http.StatusBadRequest, "Error when getting available build client")
```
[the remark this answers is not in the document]
sitole: It's not an internal server error, because it can be caused by the user not scaling the cluster to meet at least one healthy template manager.

packages/api/internal/handlers/template_request_build.go
```diff
@@ -226,7 +226,7 @@ func (a *APIStore) BuildTemplate(ctx context.Context, req BuildTemplateRequest)
 		SetFreeDiskSizeMB(req.Tier.DiskMb).
 		SetNillableStartCmd(req.StartCmd).
 		SetNillableReadyCmd(req.ReadyCmd).
-		SetNillableClusterNodeID(req.BuilderNodeID).
+		SetNillableClusterNodeID(&req.BuilderNodeID).
```
dobrac: can it be nil now?

packages/api/internal/template-manager/template_manager.go
```diff
@@ -229,7 +262,13 @@ func (tm *TemplateManager) DeleteBuild(ctx context.Context, t trace.Tracer, buil
 
 func (tm *TemplateManager) DeleteBuilds(ctx context.Context, builds []DeleteBuild) error {
 	for _, build := range builds {
-		err := tm.DeleteBuild(ctx, tm.tracer, build.BuildID, build.TemplateID, build.ClusterID, build.ClusterNodeID)
+		// Temporarily skip builds without cluster node ID
```
dobrac: why skip and not use local template manager?

packages/api/internal/template-manager/template_manager_local.go
```diff
@@ -41,21 +41,25 @@ func (tm *TemplateManager) localClientHealthSync(ctx context.Context) {
 	err = utils.UnwrapGRPCError(err)
 	if err != nil {
 		zap.L().Error("Failed to get health status of template manager", zap.Error(err))
-		tm.setLocalClientStatus(orchestratorinfo.ServiceInfoStatus_Unhealthy)
+		tm.setLocalClientInfo(orchestratorinfo.ServiceInfoStatus_Unhealthy, "unknown")
```
dobrac: can you make "unknown" to a const? I saw it in 2 places

packages/api/internal/template-manager/upload_template_layer_files.go
```diff
@@ -5,15 +5,22 @@ import (
...
 
 	"github.com/e2b-dev/infra/packages/api/internal/utils"
 	template_manager "github.com/e2b-dev/infra/packages/shared/pkg/grpc/template-manager"
 	ut "github.com/e2b-dev/infra/packages/shared/pkg/utils"
 )
 
-func (tm *TemplateManager) InitLayerFileUpload(ctx context.Context, teamID uuid.UUID, templateID string, hash string) (*template_manager.InitLayerFileUploadResponse, error) {
-	resp, err := tm.grpc.Template.InitLayerFileUpload(
-		ctx, &template_manager.InitLayerFileUploadRequest{
+func (tm *TemplateManager) InitLayerFileUpload(ctx context.Context, clusterID *uuid.UUID, nodeID *string, teamID uuid.UUID, templateID string, hash string) (*template_manager.InitLayerFileUploadResponse, error) {
```
dobrac: should the nodeID be still nillable?

dobrac (changes requested):
can you also check other places if the nodeID needs to be still nillable in the functions declarations?

packages/api/internal/template-manager/upload_template_layer_files.go
sitole: Calls of `InitLayerFileUpload` always know NodeID, so the pointer has been removed here. Now, it is propagated as a pointer to `tm.GetBuildClient`, where it is still needed to work with it as nullable.

packages/api/internal/handlers/template_request_build.go
sitole: Replaced with `SetClusterNodeID`. After this change is deployed, we can remove null on that field.

dobrac (approved):

packages/shared/pkg/db/snapshot.go
```diff
@@ -35,7 +35,8 @@ func (db *DB) NewSnapshotBuild(
 	ctx context.Context,
 	snapshotConfig *SnapshotInfo,
 	teamID uuid.UUID,
-	originNodeID string,
+	originNodeID string, // Node ID corresponds to Nomad short ID
+	originClusterNodeID string, // Cluster Node ID corresponds Node ID provided by the orchestrator itself
```
dobrac: can you rename these so it clear which is which and also, why are we using nomadID for the origin node, but cluster node ID for the build target?

packages/api/internal/orchestrator/pause_instance.go
```diff
@@ -47,11 +47,17 @@ func (o *Orchestrator) PauseInstance(
...
+	if !ok {
+		return fmt.Errorf("orchestrator node '%s' not found", sbx.Node.ID)
+	}
+
 	envBuild, err := o.dbClient.NewSnapshotBuild(
 		ctx,
 		snapshotConfig,
 		teamID,
 		sbx.Node.ID,
+		orch.ClusterNodeID,
```
dobrac: why is the ClusterNodeID needed for snapshots?

packages/api/internal/orchestrator/client.go
```diff
@@ -93,23 +94,30 @@ func (o *Orchestrator) connectToNode(ctx context.Context, node *node.NodeInfo) e
...
 	nodeInfo, err := client.Info.ServiceInfo(ctx, &emptypb.Empty{})
 	if err != nil {
-		zap.L().Error("Failed to get node info", zap.Error(err))
-	} else {
-		nodeStatus, ok = OrchestratorToApiNodeStateMapper[nodeInfo.ServiceStatus]
-		if !ok {
-			zap.L().Error("Unknown service info status", zap.Any("status", nodeInfo.ServiceStatus), logger.WithNodeID(node.ID))
-			nodeStatus = api.NodeStatusUnhealthy
-		}
+		return fmt.Errorf("failed to get node service info: %w", err)
```
dobrac, line 81: if the connectToNode will fail, there is no way to know that this node should be here, but is for some reason unresponsive

based on a past experience, this took some time to found out when for example the node was running, but the orchestrator on the node was dead

packages/api/internal/node/node.go
```diff
@@ -1,7 +1,9 @@
 package node
 
 type NodeInfo struct {
-	ID                  string
+	// Corresponds to Nomad short ID that is different from NodeID provided by the orchestrator itself.
+	ID string
```
dobrac: why don't we name it NomadShortID then?

packages/api/internal/orchestrator/node.go
```diff
@@ -111,10 +108,15 @@ func (n *Node) setStatus(status api.NodeStatus) {
 	}
 }
 
-func (n *Node) setMetadata(i *orchestratorinfo.ServiceInfoResponse, nodeID string) {
+func (n *Node) setMetadata(serviceInstanceID string, commit string, version string) {
```
dobrac: pass the struct `nodeMetadata ` directly in the function params

packages/api/internal/orchestrator/client.go
sitole: Yep, agree.
The proper approach would be to introduce an additional state indicating the nodeis presence, but since we don't know the state yet, it won't be used.

The current issue was that we are getting some IDs from this service info calls, and it can happen that they will be created with `unknown` value that can be used somewhere and that is definitely not a valid state.

packages/api/internal/node/node.go
sitole: At the end, this way we will choose, just mentioning that "nomad" in the name is relative just to our cluster and for BYOC it is not related to nomad at all and is there just for back compatibility (mainly internal cli usage).

So, lets rename it and make it as deprecated so no new code will use it. Sounds it okay?

packages/api/internal/orchestrator/pause_instance.go
sitole: Because cluster node id (node id coming from template manager/orchestrator) is different than `sbx.Node.ID` (Nomad short ID) that is used for orchestrators map primary identifier

sitole:
After the first code review batch, I decided that when we are now taking the node ID reported by the orchestrator, we no longer need to use the Nomad short ID as an identifier for the orchestrator in the API.

Now, the API uses the node ID reported by the orchestrator info service and the cluster prefix (if a remote cluster is used). There is a back compatibility layer for admin endpoints that still need to see Nomad IDs.

These changes also address the [inconsistency between the nodeID stored in the template build and the nodeID](https://github.com/e2b-dev/infra/pull/952#discussion_r2260792591) stored in a build created with the snapshot mechanism.

Currently, a migration path exists that checks the length of the ID stored in the template build. If the ID is exactly eight characters long, it attempts to find the related orchestrator; otherwise, it uses a new node ID for orchestrator search.

cc @dobrac @jakubno

packages/api/internal/template-manager/template_manager_local.go
```diff
@@ -41,21 +45,25 @@ func (tm *TemplateManager) localClientHealthSync(ctx context.Context) {
...
+		nodeID: nodeID,
+	}
 }
 
-func (tm *TemplateManager) GetLocalClientStatus() orchestratorinfo.ServiceInfoStatus {
+func (tm *TemplateManager) GetLocalClientInfo() LocalTemplateManagerInfo {
 	tm.localClientMutex.Lock()
 	defer tm.localClientMutex.Unlock()
-	return tm.localClientStatus
+	return tm.localClientInfo
```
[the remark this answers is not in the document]
sitole: Good catch!

packages/api/internal/template-manager/template_manager_local.go
```diff
@@ -41,21 +45,25 @@ func (tm *TemplateManager) localClientHealthSync(ctx context.Context) {
...
+	tm.localClientInfo = LocalTemplateManagerInfo{
+		status: status,
+		nodeID: nodeID,
+	}
 }
 
-func (tm *TemplateManager) GetLocalClientStatus() orchestratorinfo.ServiceInfoStatus {
+func (tm *TemplateManager) GetLocalClientInfo() LocalTemplateManagerInfo {
 	tm.localClientMutex.Lock()
 	defer tm.localClientMutex.Unlock()
```
dobrac: is read lock here enough?

dobrac (commented):
the nodeID management is soooo much nicer now!

packages/api/internal/orchestrator/cache.go
```diff
@@ -169,17 +166,17 @@ func (o *Orchestrator) syncClusterDiscoveredNodes(ctx context.Context) {
 
 func (o *Orchestrator) syncClusterNode(ctx context.Context, node *Node, instanceCache *instance.InstanceCache) {
 	ctx, childSpan := o.tracer.Start(ctx, "sync-cluster-node")
-	telemetry.SetAttributes(ctx, telemetry.WithNodeID(node.Info.ID), telemetry.WithClusterID(node.ClusterID), telemetry.WithClusterNodeID(node.ClusterNodeID))
+	telemetry.SetAttributes(ctx, telemetry.WithNodeID(node.Info.NodeID), telemetry.WithClusterID(node.Info.ClusterID), telemetry.WithClusterNodeID(node.Info.NodeID))
```
dobrac: is the NodeID here twice?

packages/api/internal/orchestrator/cache.go
```diff
@@ -191,31 +188,31 @@ func (o *Orchestrator) syncClusterNode(ctx context.Context, node *Node, instance
 	if !nodeFound {
 		// we are not closing grpc connection, because it is shared between all cluster nodes, and it's handled by the cluster
 		node.Close()
-		o.nodes.Remove(node.Info.ID)
+		o.nodes.Remove(o.scopedNodeID(node.Info.ClusterID, node.Info.NodeID))
```
dobrac: this change is not in other places where ID is replaced with NodeID. Is this only for the nodes key, or could there be other places where it's important?

packages/api/internal/orchestrator/client.go
```diff
@@ -122,68 +120,50 @@ func (o *Orchestrator) connectToNode(ctx context.Context, node *node.NodeInfo) e
...
-		ClusterNodeID: i.NodeID,
-
-		// some places are using this id to get node from orchestrator pool
-		// probably we can get rid of this and just create ID directly on Node struct
 		Info: &node.NodeInfo{
-			ID: poolNodeID,
+			NomadNodeShortID: node.UnknownNomadNodeShortID,
+
+			ClusterID: cluster.ID,
+			NodeID:    i.NodeID,
```
dobrac, line 140: this is now different ID, right?

packages/api/internal/orchestrator/client.go
```diff
@@ -122,68 +120,50 @@ func (o *Orchestrator) connectToNode(ctx context.Context, node *node.NodeInfo) e
...
 		},
 
 		buildCache:     buildCache,
 		sbxsInProgress: smap.New[*sbxInProgress](),
 		createFails:    atomic.Uint64{},
 	}
 
-	o.nodes.Insert(poolNodeID, orchestratorNode)
+	scopedNodeID := o.scopedNodeID(cluster.ID, i.NodeID)
+	o.nodes.Insert(scopedNodeID, orchestratorNode)
```
dobrac: can we maybe make a helper for the insert/remove maybe so we won't forget that the key has to be clusterID and nodeID composed together?

packages/api/internal/orchestrator/orchestrator.go
```diff
@@ -198,3 +203,95 @@ func (o *Orchestrator) Close(ctx context.Context) error {
 func (o *Orchestrator) WaitForPause(ctx context.Context, sandboxID string) (*node.NodeInfo, error) {
 	return o.instanceCache.WaitForPause(ctx, sandboxID)
 }
+
+func (o *Orchestrator) AdminNodes() []*api.Node {
+	nodes := make(map[string]*api.Node)
+
+	for _, n := range o.nodes.Items() {
+		// Skip all nodes that are not running in local (Nomad) cluster
+		if n.Info.ClusterID != uuid.Nil {
```
dobrac: why not filter by empty NomadNodeShortID?

packages/api/internal/orchestrator/orchestrator.go
```diff
@@ -198,3 +203,95 @@ func (o *Orchestrator) Close(ctx context.Context) error {
...
+		return cmp.Compare(i.NodeID, j.NodeID)
+	})
+
+	return result
+}
+
+func (o *Orchestrator) AdminNodeDetail(nomadNodeShortID string) *api.NodeDetail {
+	n := o.GetNodeByNomadShortID(nomadNodeShortID)
+	if n == nil {
+		return nil
```
dobrac: probably better to return an error here too

packages/api/internal/orchestrator/cache.go
sitole: It's needed to be just for the map. We are using cluster ID as a prefix for node ID, so there is no problem if multiple clusters connect with the same node IDs, causing issues with the nodes map using the same key for different cluster nodes.

packages/api/internal/orchestrator/cache.go
sitole: Yep, i will remove ClusterNodeID as its not basically NodeID. There is multiple places we are using it.

packages/api/internal/orchestrator/client.go
sitole: Yep, it is. ID reported by orchestrator (gcloud machine name).

dobrac (approved):

PR #955, opened 2025-08-04
Use Filestore over NFS to speed up and share GCS Object Storage cached slabs

djeebus:
Early tests show that this can reduce the startup time for sandboxes by ~60% if the sandbox is starting on a fresh node. 

Notes:
- this only stores template slabs in NFS, in order to reduce disk usage and blast radius
- the launch darkly flags `use-nfs-for-snapshots` and `use-nfs-for-templates` control whether we write slabs to NFS for snapshots and templates.
- 

Remaining tasks:
- [x] Bring back terraform, make sure it is production ready
- [x] Remove unused files from the NFS mount when appropriate (in progress)
- [x] Monitor usage
- [x] Protect against short writes to GCP Object Storage during batched writes

Other items worth noting:
1. the "build-packages" github action now calculates its cache key based on go.work/go.work.sum, since that collects all dependencies and is simpler than recursively searching the whole repository
2. there is a new tfvar called "use_filestore_cache" that enables the creation of the filestore (defaults to `false`)
3. `make plan-only-jobs` no longer errors if a change needs to be made, allowing `make plan-only-jobs && make apply`
4. `make test` now tests every package in the go.work file, rather than requiring them to be listed out one at a time
5. `make lint` can now be run to lint all the packages
6. I fixed a race condition in the `lifecycle_cache_test.go`, revealed when running the tests with `-race`
7. I refactored the boolean flags to combine key and default into a single struct. If we like it, we can refactor the int flags similarly.


packages/shared/pkg/telemetry/main.go
```diff
@@ -75,6 +75,7 @@ func New(ctx context.Context, serviceName, commitSHA, clientID string) (*Client,
 	if err != nil {
 		return nil, fmt.Errorf("failed to create tracer provider: %w", err)
 	}
+	otel.SetTracerProvider(tracerProvider)
```
djeebus, line 78: We can stop passing an `tracer trace.Tracer` everywhere, and simply create globals like ` var tracer = ote.Tracer("<package name>"), which should make it easier to trace.

packages/cluster/filestore/main.tf
```diff
@@ -0,0 +1,21 @@
+resource "google_filestore_instance" "slab-cache" {
+  name        = var.name
+  description = "High performance slab cache"
+  tier        = "ZONAL"
```
ValentaTomas: Is this the one we want to go long term? In terms of zone/regional and also the enterprise/ssd variant?

packages/shared/pkg/storage/storage_aws.go
```diff
@@ -155,22 +156,25 @@ func (a *AWSBucketStorageObjectProvider) WriteFromFileSystem(path string) error
 	return err
 }
 
-func (a *AWSBucketStorageObjectProvider) ReadFrom(src io.Reader) (int64, error) {
+func (a *AWSBucketStorageObjectProvider) ReadFrom(data []byte) (int64, error) {
```
ValentaTomas: This method now stops conforming to the `io.ReaderFrom` interface—which might be ok, but then we should probably rename it?

djeebus: I made the changes in a separate PR, mind taking a look? https://github.com/e2b-dev/infra/pull/1015

packages/cluster/filestore/main.tf
djeebus: The current available service tiers are Basic HDD, Basic SSD, Zonal, and Regional. 

At the low end, Basic SSD givers us the best speed (1,200 MiB/s and 60k IOPS no matter how much we provision), but as we scale up Zonal gives us better speed.

| Tier | Size | Cost | Read Bandwidth | Read IOPS |
| --- | --- | ---- | ---- | ---- |
| SSD | 1 TiB | $307 | 1,200 MiB/s | 60,000 IOPS |
| Zonal | 1 TiB | $256 | 260 MiB/s | 9,200 IOPS |
| Regional | 1 TiB | $461 | 120 MiB/s | 12,000 IOPS |
| SSD | 10 TiB | $3,072 | 1,200 MiB/s | 60,000 IOPS | 
| Zonal | 10 TiB | $2,560 | 2,600 MiB/s | 92,000 IOPS |
| Regional | 10 TiB | $4,610 | 2,600 MiB/s | 92,000 IOPS |

As we scale up, Regional is the same as Zonal, but nearly twice as expensive.

djeebus:
Waiting on #1016 and #1015 to simplify this PR

packages/cluster/filestore/main.tf
ValentaTomas: Got it.
I think there is also permanent choice around capacity range—let's go for the 10TiB for now, so we are not blocked.

Screenshot 2025-08-20 at 14 39 55

djeebus: I've added some config parameters to allow us to create appropriate filestores in each env

ValentaTomas:
@djeebus Maybe we talked about this before, but could there be a race between:

1. Orchestrator A creates an empty file X to write the chunk to NFS
2. Orchestrator B checks NFS if there is chunk and finds X
3. Orchestrator B reads the X and tries to serve the content (nothing)
4. Orchestrator A writes the content of the chunk to the file X

djeebus:
> @djeebus Maybe we talked about this before, but could there be a race between:
> 
> 1. Orchestrator A creates an empty file X to write the chunk to NFS
> 2. Orchestrator B checks NFS if there is chunk and finds X
> 3. Orchestrator B reads the X and tries to serve the content (nothing)
> 4. Orchestrator A writes the content of the chunk to the file X

It may or may not be a problem, depending on Filestore's locking implementation. We can write to a temp file first though, then move it to the right spot when it's done. Pretty easy to implement, I'll do that. That way the file is atomic (it is either there or not there).

packages/orchestrator/Dockerfile
```diff
@@ -20,12 +20,12 @@ WORKDIR /build/orchestrator
 COPY go.mod go.sum ./
 RUN go mod download
 
-COPY main.go Makefile ./
-COPY internal internal
+COPY . ./
```
ValentaTomas, line 23: Why copy everything here?

main.tf
```diff
@@ -269,3 +275,9 @@ module "redis" {
 
   depends_on = [module.api]
 }
+
+
+import {
+  id = "e2b-staging-joe/memory-cache-disk-usage-low"
```
ValentaTomas: This is the alert, right? Should not be here.

packages/orchestrator/Dockerfile
djeebus: This command copies "Everything that is necessary for compilation", which changes over time. As an example of that, I added a new compilation target in ./cmd/clean-nfs-cache and found out that the Dockerfile didn't copy any of my code files over. Having to add a new COPY step every time we add a new folder under /packages/orchestrator doesn't seem useful, as the only thing it buys us is saving a few dozen kilobytes of data, and changes to a few rarely changed files won't trigger a new build. "everything" (beyond what we already copy) are some .sh, .md, .proto, and a Makefile, and some other text files. if there was anything big here, we should probably add a .dockerignore file, so it doesn't make it into the context (which potentially saves some time, if the files are large).

Edit: the /bin folder in this directory can get to be 90+ MB, I'll add that .dockerignore file to hide that one from the context

packages/cluster/filestore/main.tf
```diff
@@ -0,0 +1,132 @@
...
+    network = var.network_name
+  }
+}
+
+data "google_monitoring_notification_channel" "notification" {
+  count        = var.notification_display_name == null ? 0 : 1
+  display_name = var.notification_display_name
+}
+
+resource "google_monitoring_alert_policy" "warning" {
```
ValentaTomas: Let's switch to Grafana as we talked. We can approve and merge before, but would add this right after/alongside.

packages/orchestrator/Dockerfile
djeebus: Edit 2: turns out docker falls back to the .gitignore file if .dockerignore doesn't exist, so it's already avoiding the large files. ... or possibly it's a superset? or in addition to? Regardless, it's ignoring the bin folder on my local and stops when I remove `bin` from the `.gitignore` file in that folder.

packages/shared/pkg/storage/storage_cache.go
```diff
@@ -0,0 +1,461 @@
...
+		trace.WithAttributes(attribute.String("path", path)))
+	defer endSpan(span, err)
+
+	// write the file to the disk and the remote system at the same time.
+	// this opens the file twice, but the API makes it difficult to use a MultiWriter
+
+	var eg errgroup.Group
+
+	eg.Go(func() error {
+		return c.createCacheBlocksFromFile(ctx, path)
```
ValentaTomas: We have this implemented, but right now it is not used in places where the writing can happen (template-builder, paused snapshot upload), right?

packages/shared/pkg/storage/storage_cache.go
```diff
@@ -0,0 +1,461 @@
...
+	cacheDoesNotExist := os.IsNotExist(err)
+
+	// read remote file
+	readCount, err := c.inner.ReadAt(buff, offset)
+	if err != nil {
+		return 0, fmt.Errorf("failed to perform uncached read: %w", err)
+	}
+
+	if cacheDoesNotExist {
+		c.writeBytesToLocal(ctx, offset, chunkPath, buff[:readCount])
```
ValentaTomas: Should this be synchronous with the read?

packages/shared/pkg/storage/storage_cache.go
```diff
@@ -0,0 +1,461 @@
...
+	}
+
+	return readCount, nil
+}
+
+var (
+	ErrOffsetUnaligned = errors.New("offset must be a multiple of chunk size")
+	ErrBufferTooSmall  = errors.New("buffer is too small")
```
ValentaTomas, line 167: Not related to the specific errors here, but potentially the last chunk from the remote storage can be "short" as it might be less than 4MB because of the alignment. We are never reading the "missing" part of the chunk, so we should be able to handle it the same in the cache chunks, just wanted to mention, so we handle that 100%.

packages/shared/pkg/storage/storage_cache.go
```diff
@@ -0,0 +1,461 @@
...
+	} else if num != len(src) {
+		return 0, fmt.Errorf("failed to copy %d bytes from cache: %w", num, err)
+	}
+
+	return num, nil
+}
+
+func (c *CachedFileObjectProvider) ReadAt(buff []byte, offset int64) (int, error) {
+	var err error
+	ctx, span := tracer.Start(c.ctx, "CachedFileObjectProvider.WriteTo", trace.WithAttributes(
```
ValentaTomas: Do we want to do spans/traces on these? It will probably be tens of thousands of these and even the traces as we have now in grafana might not handle that well.

packages/shared/pkg/storage/storage_cache.go
```diff
@@ -0,0 +1,461 @@
...
+			zap.String("path", tempPath),
+			zap.Int64("offset", offset),
+			zap.Int("length", len(bytes)),
+			zap.Error(err),
+		)
+		return
+	}
+}
+
+func (c *CachedFileObjectProvider) createCacheBlocksFromFile(ctx context.Context, inputPath string) error {
```
ValentaTomas: @djeebus Most of the methods that follow are for handling the writes during template build/sandbox pause, right? Not necessarily needed for pure read cache and disabled for these cases as the `CachedProvider` is not used.

packages/orchestrator/internal/template/build/layer/layer_executor.go
```diff
@@ -73,9 +73,11 @@ func (lb *LayerExecutor) BuildLayer(
 	defer childSpan.End()
 
 	localTemplate, err := lb.templateCache.GetTemplate(
```
ValentaTomas: @djeebus Won't the NFS cache be used here (which will use the write)?

djeebus:
> Also not sure if https://github.com/e2b-dev/infra/blob/nfs-file-cache/packages/orchestrator/internal/server/sandboxes.go#L423 will the snapshot upload use the cached persistence and therefore the mentioned writes.

No, it's only used in the GetTemplate function, which is called on Create (line 52 in the same file). The wrapped persistence value only gets used inside GetTemplate and in the Template instance, which isn't used to upload the snapshot.

packages/shared/pkg/storage/storage_cache.go
djeebus: Correct, I believe the only functions that are called are:
- ReadAt (called by Chunker)
- Size (called by StorageDiff and NewStorage)
- WriteTo (called by NewStorage, newStorageFile)

WriteFromFileSystem is only called by TemplateBuild.Upload, which won't have a wrapped persistence instance.

packages/shared/pkg/storage/storage_cache.go
djeebus: Good point! The only thing preventing us from making this async is that we need to ensure that `buff` won't be changed while we're uploading. I think the system is built that way, but I'm not confident that's true.

packages/shared/pkg/storage/storage_cache.go
djeebus: Thanks for the call out! Just to be clear, as an example:
- ReadFrom is called with a 2 MB buffer
- we only have 1 MB of data
- we should set 1 MB of the buffer and return <1 mb> as an int, with no error
Does that sound correct? I do suspect that's a bug that I need to fix. Interesting that the sandbox still managed to work, but maybe it didn't grab the last block.

packages/shared/pkg/storage/storage_cache.go
djeebus: I can remove them, they're not providing much value right now. Since we have `ctx` on the struct and not passed in, they don't get attached to the correct parents, so they're hard to find in tempo.

packages/orchestrator/internal/template/build/layer/layer_executor.go
djeebus: yes, potentially the template could get cached while we build from it. It wouldn't be hard to disable it, but I could see it being useful. Do you have an opinion one way or the other?

packages/shared/pkg/storage/storage_cache.go
ValentaTomas: Slightly differently:

1. We `ReadAt` for a 2MB buffer (the page), but this is the last page in the file.
2. We try to download a 4MB chunk, the first 2MB is ok, but the last 2MB are not there, we handle the expected EOF (as opposed to UnexpectedEOF) by just returning the first 2 MB
3. We return the 2MB chunk the ReadAt asked for.

packages/orchestrator/internal/template/build/layer/layer_executor.go
ValentaTomas: Let's not involve it now—we only have one template build node too, and I don't want to combine it with the rest of the logic.

packages/shared/pkg/storage/storage_cache.go
ValentaTomas: Let's recheck if it is so and move it to async.

ValentaTomas (approved):

PR #956, opened 2025-08-05
Update sandbox metadata

jakubno:
# Description

This pull request introduces support for setting sandbox metadata via a new PUT endpoint in the API. These changes allow clients to set the metadata of a running sandbox.

### DIFF
--- a/tests/integration/internal/tests/api/kill_test.go
+++ b/tests/integration/internal/tests/api/kill_test.go
@@ -66,20 +66,20 @@ func TestSandboxKill(t *testing.T) {
 		// pause the sandbox
 		pauseSandboxResponse, err := c.PostSandboxesSandboxIDPauseWithResponse(t.Context(), sandboxID, setup.WithAPIKey())
 
-		assert.NoError(t, err)
-		assert.Equal(t, http.StatusNoContent, pauseSandboxResponse.StatusCode())
+		require.NoError(t, err)
+		require.Equal(t, http.StatusNoContent, pauseSandboxResponse.StatusCode())
 
 		// kill the sandbox
 		killSandboxResponse, err := c.DeleteSandboxesSandboxIDWithResponse(t.Context(), sandboxID, setup.WithAPIKey())
 
-		assert.NoError(t, err)
-		assert.Equal(t, http.StatusNoContent, killSandboxResponse.StatusCode())
+		require.NoError(t, err)
+		require.Equal(t, http.StatusNoContent, killSandboxResponse.StatusCode())
 
 		// list all sandboxes and check that the sandbox is not in the list
 		listSandboxesResponse, err := c.GetSandboxesWithResponse(t.Context(), &api.GetSandboxesParams{}, setup.WithAPIKey())
 
-		assert.NoError(t, err)
-		assert.Equal(t, http.StatusOK, listSandboxesResponse.StatusCode())
+		require.NoError(t, err)
+		require.Equal(t, http.StatusOK, listSandboxesResponse.StatusCode())
 
 		runningSandboxes := listSandboxesResponse.JSON200
 		if runningSandboxes == nil {


packages/api/internal/orchestrator/update_sandbox_metadata.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		return &api.APIError{
+			Err:       err,
+			ClientMsg: "Failed to update sandbox metadata",
+			Code:      500,
+		}
+	}
+
+	// Only update local cache after orchestrator call succeeds
+	instanceInfo.Metadata = newMetadata
```
dobrac: there should be lock around the setter to prevent data race

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -135,7 +135,25 @@ func (s *server) Update(ctx context.Context, req *orchestrator.SandboxUpdateRequ
...
+	// Update metadata if provided
+	if req.Metadata != nil {
+		item.Config.Metadata = req.Metadata
+		updated = true
+		telemetry.ReportEvent(ctx, "Updated sandbox metadata")
+	}
+
+	if !updated {
+		telemetry.ReportEvent(ctx, "Sandbox update request received but no changes applied")
+	}
```
dobrac: this looks a bit unnecessary when you have the report event for each change

dobrac (changes requested):

dobrac:
> # Description
> This PR introduces a comprehensive team metrics system for tracking sandbox usage, including:
> 
> * New API endpoint `/teams/metrics` for retrieving team-level metrics
> * Team metrics collection using OpenTelemetry for real-time concurrent sandbox tracking
> * ClickHouse integration for historical team metrics storage
> * Integration tests for the new functionality

the description looks to be incorrect

packages/api/internal/handlers/sandbox_update.go
```diff
@@ -0,0 +1,89 @@
...
+		telemetry.WithSandboxID(sandboxID),
+		telemetry.WithTeamID(teamID.String()),
+	)
+
+	sbx, err := a.orchestrator.GetSandbox(sandboxID)
+	if err == nil {
+		// Verify the sandbox belongs to the team
+		if sbx.TeamID != teamID {
+			telemetry.ReportCriticalError(ctx, fmt.Sprintf("sandbox '%s' doesn't belong to team '%s'", sandboxID, teamID.String()), nil)
+			a.sendAPIStoreError(c, http.StatusUnauthorized, fmt.Sprintf("Error updating sandbox - sandbox '%s' does not belong to your team '%s'", sandboxID, teamID.String()))
```
jakubno: this is currently inconsistent with paused sandbox, we return 401 for running and 404 for paused sandboxes if the team doesn't match with the sandbox's team. Not sure if it's worth doing two queries as this shouldn't happen

packages/api/internal/handlers/sandbox_update.go
```diff
@@ -0,0 +1,92 @@
...
+			a.sendAPIStoreError(c, apiErr.Code, apiErr.ClientMsg)
+
+			return
+		}
+	} else {
+		// Sandbox not found in cache, might be paused
+		// Try to update the snapshot metadata in the database
+		zap.L().Debug("Sandbox not found in cache, checking if it's paused",
+			logger.WithSandboxID(sandboxID),
+			zap.String("team.id", teamID.String()))
```
dobrac: use WithTeamID helper

packages/api/internal/handlers/sandbox_update.go
dobrac: can we do 404 instead? This way you can get which sandbox ID are run by different teams

(also obfuscate the error message sent to the user that the sandbox was simply not found)

packages/api/internal/handlers/sandbox_update.go
```diff
@@ -0,0 +1,92 @@
...
+		// Verify the sandbox belongs to the team
+		if sbx.TeamID != teamID {
+			telemetry.ReportCriticalError(ctx, fmt.Sprintf("sandbox '%s' doesn't belong to team '%s'", sandboxID, teamID.String()), nil)
+			a.sendAPIStoreError(c, http.StatusUnauthorized, fmt.Sprintf("Error updating sandbox - sandbox '%s' does not belong to your team '%s'", sandboxID, teamID.String()))
+
+			return
+		}
+
+		sbx.Lock()
+		defer sbx.Unlock()
```
dobrac: looks like a path to the global sandbox lock 👀 (this lock might be more interesting with having multiple APIs)

packages/api/internal/orchestrator/update_sandbox_metadata.go
```diff
@@ -0,0 +1,55 @@
...
+	if err != nil {
+		return &api.APIError{
+			Err:       err,
+			ClientMsg: "Failed to update sandbox metadata",
+			Code:      http.StatusInternalServerError,
+		}
+	}
+
+	// Only update local cache after orchestrator call succeeds
+	sbx.Metadata = metadata
```
dobrac, line 50: there is a lock around metadata for writing, but is there also for reading?

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -197,18 +203,37 @@ func (s *server) Update(ctx context.Context, req *orchestrator.SandboxUpdateRequ
 		return nil, status.Error(codes.NotFound, "sandbox not found")
 	}
 
-	item.EndAt = req.EndTime.AsTime()
+	item.Mu.Lock()
+	defer item.Mu.Unlock()
```
dobrac: this lock is not implemented for getters

also, might be worth to move the updates to the sandbox struct and not expose the lock

packages/orchestrator/internal/server/sandboxes_update_test.go
dobrac, line 1: might be worth adding also tests for the concurrent changes

tests/integration/internal/tests/api/sandbox_metadata_update_test.go
```diff
@@ -0,0 +1,367 @@
...
+	require.NotNil(t, getSandboxResponse.JSON200)
+
+	// Metadata should be empty or nil after PUT with empty map
+	if getSandboxResponse.JSON200.Metadata != nil {
+		metadata := *getSandboxResponse.JSON200.Metadata
+		assert.Empty(t, metadata, "Metadata should be empty after PUT with empty map")
+	}
+}
+
+// TestSandboxMetadataUpdateNil is removed as PUT requires a body
```
dobrac: ?

tests/integration/internal/tests/api/sandbox_metadata_update_test.go
```diff
@@ -0,0 +1,367 @@
...
+	assert.Equal(t, http.StatusOK, getUpdatedSandboxResponse.StatusCode())
+	require.NotNil(t, getUpdatedSandboxResponse.JSON200)
+	require.NotNil(t, getUpdatedSandboxResponse.JSON200.Metadata)
+
+	updatedMeta := *getUpdatedSandboxResponse.JSON200.Metadata
+
+	// Verify updated values
+	assert.Equal(t, "e2b-is-place-to-be", updatedMeta["environment"])
+	assert.Equal(t, "1.1.0", updatedMeta["version"])
+	assert.Equal(t, "feature-test", updatedMeta["branch"])
```
dobrac: this tests might look better when defined using the Golang:

test cases (test, expected result)

runtime

structure

tests/integration/internal/tests/api/sandbox_metadata_update_test.go
```diff
@@ -0,0 +1,367 @@
...
+		"test": "value",
+	}
+
+	updateResponse, err := c.PutSandboxesSandboxIDMetadataWithResponse(t.Context(), "nonexistent-sandbox-id", updateMetadata, setup.WithAPIKey())
+
+	require.NoError(t, err)
+	assert.Equal(t, http.StatusNotFound, updateResponse.StatusCode())
+}
+
+func TestSandboxMetadataUpdateWithSpecialCharacters(t *testing.T) {
```
dobrac: not sure if necessary

tests/integration/internal/tests/api/sandbox_metadata_update_test.go
```diff
@@ -0,0 +1,367 @@
...
+	assert.Equal(t, "/opt/app/config", metadata["config/path"])
+	assert.Equal(t, "2024.01.15", metadata["build-number"])
+	assert.Equal(t, "true", metadata["feature_flag"])
+	assert.Equal(t, "🚀", metadata["emoji"])
+	assert.Equal(t, "αβγδε", metadata["unicode"])
+	assert.Equal(t, `{"key": "value"}`, metadata["json_like"])
+	assert.Equal(t, "value with spaces", metadata["spaces"])
+}
+
+func TestSandboxMetadataUpdateInvalidAuth(t *testing.T) {
```
dobrac: probably not necessary, better to do test that verifies the paused/vs running foreign team sandbox scenarios

tests/integration/internal/tests/api/sandbox_metadata_update_test.go
```diff
@@ -0,0 +1,367 @@
...
+	assert.Equal(t, http.StatusUnauthorized, updateResponse.StatusCode())
+
+	// Try to update with invalid API key
+	updateResponse, err = c.PutSandboxesSandboxIDMetadataWithResponse(t.Context(), sandbox.SandboxID, updateMetadata, setup.WithAPIKey("invalid-api-key"))
+
+	require.NoError(t, err)
+	assert.Equal(t, http.StatusUnauthorized, updateResponse.StatusCode())
+}
+
+func TestSandboxMetadataUpdateMultipleUpdates(t *testing.T) {
```
dobrac: might be better to test the async/parallel locking

tests/integration/internal/tests/api/sandbox_metadata_update_test.go
jakubno: I think it still make sense to keep no api key, removed invalid api key

jakubno:
@cursoragent review

packages/api/internal/cache/instance/instance.go
```diff
@@ -111,7 +110,7 @@ type InstanceInfo struct {
 	ClusterID           uuid.UUID
 	AutoPause           atomic.Bool
 	Pausing             *utils.SetOnce[string]
-	mu                  sync.RWMutex
+	sync.RWMutex
```
dobrac: why exposing the mutex and not having all required methods on the struct?

packages/api/internal/handlers/sandbox_get.go
```diff
@@ -41,35 +41,39 @@ func (a *APIStore) GetSandboxesSandboxID(c *gin.Context, id string) {
...
-			EnvdVersion:     info.EnvdVersion,
-			EnvdAccessToken: info.EnvdAccessToken,
+			EnvdVersion:     sbx.EnvdVersion,
+			EnvdAccessToken: sbx.EnvdAccessToken,
 			Domain:          sbxDomain,
 		}
 
-		if info.Metadata != nil {
-			meta := api.SandboxMetadata(info.Metadata)
+		metadata := sbx.Metadata()
```
dobrac: here is a lock inside of lock (it's read only, so it should be okay-ish now)

packages/api/internal/handlers/sandboxes_list.go
```diff
@@ -286,8 +286,9 @@ func instanceInfoToPaginatedSandboxes(runningSandboxes []*instance.InstanceInfo)
 			PaginationTimestamp: info.StartTime,
```
dobrac, line 288: missing lock here

packages/api/internal/orchestrator/pause_instance.go
```diff
@@ -37,7 +37,7 @@ func (o *Orchestrator) PauseInstance(
 		VCPU:                sbx.VCpu,
 		RAMMB:               sbx.RamMB,
 		TotalDiskSizeMB:     sbx.TotalDiskSizeMB,
```
dobrac: missing rlock?

packages/api/internal/orchestrator/update_sandbox_metadata.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/e2b-dev/infra/packages/api/internal/utils"
+	"github.com/e2b-dev/infra/packages/shared/pkg/grpc/orchestrator"
+	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
+)
+
+func (o *Orchestrator) UpdateSandboxMetadata(
+	ctx context.Context,
+	sbx *instance.InstanceInfo,
+	metadata map[string]string,
+) *api.APIError {
```
dobrac, line 18: is there a missing lock for concurrent metadata updates? e.g. first request succeeds second on the orchestrator, but second request sets the metadata on the API -> inconsistency

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -89,6 +90,8 @@ type Sandbox struct {
 	*Resources
 	*Metadata
 
+	sync.RWMutex
```
dobrac: why having the lock on the sandbox and not in the metadata handler only

also, I don't like the embedding

tests/integration/internal/tests/api/sandbox_metadata_update_test.go
```diff
@@ -0,0 +1,523 @@
...
+	}
+
+	// Wait for all updates to complete
+	wg.Wait()
+
+	// All updates should succeed (no errors due to concurrent access)
+	assert.Equal(t, count, successCount, "All concurrent updates should succeed")
+
+	// Verify final state - should be from one of the updates
+	getSandboxResponse, err := c.GetSandboxesSandboxIDWithResponse(t.Context(), sandbox.SandboxID, setup.WithAPIKey())
```
dobrac, line 495: this doesn't check that the metadata reported by the API is the same as the one reported by the orchestrator

packages/api/internal/cache/instance/instance.go
jakubno: What's the advantage of reimplementing it on the struct?

dobrac: to not expose internal mechanisms/implementation details to the usage

packages/api/internal/orchestrator/update_sandbox_metadata.go
jakubno: I refactor how we lock, now if there's some handler or background process working with sandbox it should always acquire lock

PR #957, opened 2025-08-05
Build and sandbox refactoring cleanup

dobrac:
Build and sandbox refactoring cleanup


packages/orchestrator/internal/template/build/phases/base/provision.sh
```diff
@@ -75,7 +75,7 @@ chmod 600 /swap/swapfile
 mkswap /swap/swapfile
 
 echo "Increasing inotify watch limit"
-echo 'fs.inotify.max_user_watches=65536' | sudo tee -a /etc/sysctl.conf
+echo 'fs.inotify.max_user_watches=65536' | tee -a /etc/sysctl.conf
```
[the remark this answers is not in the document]
dobrac: this was a bug previously, there is no sudo in this context and everything is run as root

packages/orchestrator/internal/template/build/phases/steps/builder.go
```diff
@@ -0,0 +1,233 @@
...
+		},
+	}
+
+	// First not cached layer is create (to change CPU, Memory, etc), subsequent are layers are resumes.
+	var sandboxCreator layer.SandboxCreator
+	if sourceLayer.Cached {
+		sandboxCreator = layer.NewCreateSandbox(sbxConfig, fc.FirecrackerVersions{
+			KernelVersion:      sb.Template.KernelVersion,
+			FirecrackerVersion: sb.Template.FirecrackerVersion,
+		}, currentLayer.Metadata.Template.TemplateID)
```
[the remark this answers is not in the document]
dobrac: This is correct, when creating a new sandbox, there is no need to keep the old rootfs snapshot kernel version, as well as the firecracker version. At that point, the base template ID is the current layer template ID for future sandboxes

packages/orchestrator/internal/template/build/phases/base/builder.go
```diff
@@ -0,0 +1,373 @@
...
+		layerExecutor: layerExecutor,
+		index:         index,
+	}
+}
+
+func (bb *BaseBuilder) Build(
+	ctx context.Context,
+	_ phases.LayerResult,
+) (phases.LayerResult, error) {
+	hash := HashBase(bb.index, bb.Config, provisionScriptFile)
```
jakubno: do you need to compute the hash here? I could be already part of the struct or a method on the struct? More generally I would add Hash method on all builders, so you can easily tell from what the hash should be build

dobrac: 👍  I'll move it

dobrac: I couldn't add it more generally, because the step builder needs to manage multiple layers as opposed the base/finalized builders manage only one layer

packages/orchestrator/internal/template/build/phases/phase.go
```diff
@@ -20,6 +20,16 @@ type LayerResult struct {
 	StartMetadata *metadata.StartMetadata
 }
 
+func (l LayerResult) Next(next LayerResult) LayerResult {
```
jakubno: this isn't used

packages/orchestrator/internal/template/build/phases/base/hash.go
```diff
@@ -8,7 +8,7 @@ import (
 	"github.com/e2b-dev/infra/packages/orchestrator/internal/template/build/storage/cache"
 )
 
-func HashBase(index cache.Index, template config.TemplateConfig, provisionScriptFile string) string {
+func (bb *BaseBuilder) Hash(index cache.Index, template config.TemplateConfig, provisionScriptFile string) string {
```
jakubno: is there a reason to pass index and config? why not use it directly from the struct?

packages/orchestrator/internal/template/build/phases/phase.go
dobrac: removed

packages/orchestrator/internal/template/build/phases/base/hash.go
dobrac: moved to the struct properties

PR #965, opened 2025-08-06
Refactor UFFD mapping handling

ValentaTomas:
Extract UFFD mapping handling to prepare for adding UFFD write protection.

### DIFF
--- a/packages/orchestrator/internal/sandbox/uffd/mapping/firecracker.go
+++ b/packages/orchestrator/internal/sandbox/uffd/mapping/firecracker.go
@@ -16,13 +16,9 @@ func (m *GuestRegionUffdMapping) relativeOffset(addr uintptr) int64 {
 
 type FcMappings []GuestRegionUffdMapping
 
-func (m FcMappings) GetRange(addr uintptr) (int64, int64, error) {
-	return getMappedRange(addr, m)
-}
-
 // Returns the relative offset and the page size of the mapped range for a given address
-func getMappedRange(addr uintptr, mappings []GuestRegionUffdMapping) (int64, int64, error) {
-	for _, m := range mappings {
+func (m FcMappings) GetRange(addr uintptr) (int64, int64, error) {
+	for _, m := range m {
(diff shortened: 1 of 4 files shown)


packages/orchestrator/internal/sandbox/uffd/serve.go
```diff
@@ -142,16 +143,13 @@ outerLoop:
 
 		addr := constants.GetPagefaultAddress(&pagefault)
 
-		mapping, err := getMapping(uintptr(addr), mappings)
```
jakubno, line 147: this function isn't used anymore, remove it

packages/orchestrator/internal/sandbox/uffd/mapping/firecracker.go
```diff
@@ -0,0 +1,31 @@
+package mapping
+
+import "fmt"
+
+type GuestRegionUffdMapping struct {
+	BaseHostVirtAddr uintptr `json:"base_host_virt_addr"`
+	Size             uintptr `json:"size"`
+	Offset           uintptr `json:"offset"`
+	// This is actually in bytes—it is deprecated and they introduced "page_size"
```
jakubno: can we use page_size then?

ValentaTomas: Not sure if we are using FC when the new field is already added, will clarify in comment.

packages/orchestrator/internal/sandbox/uffd/serve.go
ValentaTomas: I think it was removed (https://github.com/e2b-dev/infra/pull/965/files/59c17c566fd31a4471272814e4cd6c509c386767#diff-b3919c77ca6877c540824059718fc4b6c8552369a2c2c9884e3772a5785ff46d)—not sure why it was shown.

jakubno: removed here https://github.com/e2b-dev/infra/pull/965/commits/e2eb48478621bc832731a1615f6034399a95c84d

jakubno (approved):
Removed the unused function to speed it up, I hope you don't mind

PR #966, opened 2025-08-06
Refactor UFFD logs sandbox field

ValentaTomas:
Refactor logging sandbox ID in UFFD serve. When I later add tests this makes it easier to do that.

### DIFF
--- a/packages/orchestrator/internal/sandbox/uffd/serve.go
+++ b/packages/orchestrator/internal/sandbox/uffd/serve.go
@@ -6,12 +6,12 @@ import (
 	"syscall"
 	"unsafe"
 
-	"github.com/loopholelabs/userfaultfd-go/pkg/constants"
 	"go.uber.org/zap"
 	"golang.org/x/sync/errgroup"
 	"golang.org/x/sys/unix"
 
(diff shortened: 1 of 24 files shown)


packages/orchestrator/internal/sandbox/uffd/serve.go
```diff
@@ -155,8 +154,7 @@ outerLoop:
 		eg.Go(func() error {
 			defer func() {
 				if r := recover(); r != nil {
-					zap.L().Error("UFFD serve panic", logger.WithSandboxID(sandboxId), zap.Any("offset", offset), zap.Any("pagesize", pagesize), zap.Any("panic", r))
-					fmt.Printf("[sandbox %s]: recovered from panic in uffd serve (offset: %d, pagesize: %d): %v\n", sandboxId, offset, pagesize, r)
```
ValentaTomas, line 167: I think removing the duplicate stdout log makes sense if we trust the zap to log it.

jakubno (changes requested):
Can you pass a logger (with given fields set) instead of fields? It's more general

jakubno (approved):

PR #967, opened 2025-08-06
Add extended userfaultfd constants

ValentaTomas:
Add extended userfaultfd constants that are needed for adding the write protection.

They are not available in the https://github.com/loopholelabs/userfaultfd-go/tree/main/pkg/constants, and this is essentially the only file we need from there.


packages/orchestrator/internal/sandbox/uffd/userfaultfd/constants.go
```diff
@@ -0,0 +1,116 @@
+package userfaultfd
+
```
jakubno, line 2: i would also add a link to loopholelabs as a reference

ValentaTomas: Done

PR #969, opened 2025-08-07
Make rootfs path constant

dobrac:
Make rootfs path constant for new templates


packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -218,12 +220,9 @@ func CreateSandbox(
...
 		rootfsPath,
-		storage.RootfsPaths{
-			// This is the potential ID which might become the TemplateID for future resumes.
-			TemplateID: config.BaseTemplateID,
-			// The rootfs build ID is from the header, because it needs to be the same from
-			// the first FS creation.
-			BuildID: rootFS.Header().Metadata.BaseBuildId.String(),
+		fc.RootfsPaths{
+			// The version is always 2 for the rootfs paths format change.
+			Version: 2,
```
jakubno: use const

dobrac: moved to 
```
var ConstantRootfsPaths = RootfsPaths{
	// The version is always 2 for the constant rootfs paths format change.
	Version: 2,
}
```

PR #973, opened 2025-08-07
Extract low level uffd handler to a separate package

ValentaTomas:
- [x] Extract low level uffd handler to a separate package
- [x] Remove unused struct
- [ ] Add uffd tests

### DIFF
--- a/packages/orchestrator/internal/sandbox/uffd/userfaultfd/serve.go
+++ b/packages/orchestrator/internal/sandbox/uffd/userfaultfd/serve.go
@@ -1,4 +1,4 @@
-package uffd
+package userfaultfd
 
 import (
 	"errors"
@@ -13,27 +13,16 @@ import (
 	"github.com/e2b-dev/infra/packages/orchestrator/internal/sandbox/block"
 	"github.com/e2b-dev/infra/packages/orchestrator/internal/sandbox/uffd/fdexit"
 	"github.com/e2b-dev/infra/packages/orchestrator/internal/sandbox/uffd/mapping"
-	"github.com/e2b-dev/infra/packages/orchestrator/internal/sandbox/uffd/userfaultfd"
 )
 
-var ErrUnexpectedEventType = errors.New("unexpected event type")
-
-type GuestRegionUffdMapping struct {
-	BaseHostVirtAddr uintptr `json:"base_host_virt_addr"`
-	Size             uintptr `json:"size"`
-	Offset           uintptr `json:"offset"`
-	PageSize         uintptr `json:"page_size_kib"`
-}
-
-func Serve(
-	uffd int,
+func (u *userfaultfd) Serve(
 	mappings mapping.Mappings,
 	src block.Slicer,
 	fdExit *fdexit.FdExit,
 	logger *zap.Logger,
 ) error {
 	pollFds := []unix.PollFd{
-		{Fd: int32(uffd), Events: unix.POLLIN},
+		{Fd: int32(u.fd), Events: unix.POLLIN},
 		{Fd: int32(fdExit.Reader()), Events: unix.POLLIN},
 	}
 
@@ -92,10 +81,10 @@ outerLoop:
 			continue
 		}
 
-		buf := make([]byte, unsafe.Sizeof(userfaultfd.UffdMsg{}))
+		buf := make([]byte, unsafe.Sizeof(UffdMsg{}))
 
 		for {
-			n, err := syscall.Read(uffd, buf)
+			n, err := syscall.Read(int(u.fd), buf)
 			if err == syscall.EINTR {
 				logger.Debug("uffd: interrupted read, reading again")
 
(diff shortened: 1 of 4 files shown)


ValentaTomas:
I'll improve the tests a little, but the rest should be ready.

ValentaTomas:
✅ Extracted all non-relevant (mostly write-protection related) code to a separated PR so it can be merged in steps.

packages/orchestrator/internal/sandbox/uffd/userfaultfd/userfaultfd_test.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+func TestUffdMissing(t *testing.T) {
+	tests := []pageTest{
+		// Standard 4K page, operation at start
+		{
+			pagesize:        header.PageSize,
+			numberOfPages:   32,
+			operationOffset: 0,
+		},
```
djeebus: Instead of tracking the purpose of a test via comment, you can use a map to give the tests meaningful names:


```suggestion
	tests := map[string]pageTest{
		"Standard 4K page, operation at start": {
			pagesize:        header.PageSize,
			numberOfPages:   32,
			operationOffset: 0,
		},
```

Then when you iterate over the tests, you can use the name:

```golang
for name, tt := range tests {
	t.Run(name, func(t *testing.T) {
```

packages/orchestrator/internal/sandbox/uffd/memory/map.go
```diff
@@ -0,0 +1,5 @@
+package memory
+
+type MemoryMap interface {
+	GetOffset(hostVirtAddr uintptr) (offset int64, pagesize uint64, err error)
+}
```
djeebus: In general, golang would rather you put the interface as close to its consumption as possible. In this case, it probably makes the most sense to put in `uffd/userfaultfd`, since the only function that accepts it as a parameter is located there.

packages/orchestrator/internal/sandbox/uffd/userfaultfd/userfaultfd_test.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	for _, tt := range tests {
+		t.Run(fmt.Sprintf("pagesize-%d-offset-%d", tt.pagesize, tt.operationOffset), func(t *testing.T) {
+			data, size := testutils.RandomPages(tt.pagesize, tt.numberOfPages)
+
+			uffd, err := newUserfaultfd(syscall.O_CLOEXEC | syscall.O_NONBLOCK)
+			if err != nil {
+				t.Fatal("failed to create userfaultfd", err)
+			}
```
djeebus: Instead of the conditionals, you can use `testify` to simplify the tests.


```suggestion
			require.NoError(t, err)
```

packages/orchestrator/internal/sandbox/uffd/memory/map.go
ValentaTomas: Make sense, one hidden thing here—I'll be expanding the interface with a method (direct opposite of the `GetOffset`) that is not used in the `uffd/userfaultfd`, but elsewhere.

packages/orchestrator/internal/sandbox/uffd/userfaultfd/userfaultfd_test.go
```diff
@@ -0,0 +1,141 @@
...
+				fdExit.Close()
+			})
+
+			exitUffd := make(chan struct{}, 1)
+
+			go func() {
+				err := uffd.Serve(t.Context(), m, data, fdExit, zap.L())
+				if err != nil {
+					fmt.Println("[TestUffdMissing] failed to serve uffd", err)
+				}
```
djeebus: Failures to serve should probably fail the test explicitly:


```suggestion
				err := uffd.Serve(t.Context(), m, data, fdExit, zap.L())
				assert.NoError(t, err)
```

Go would prefer you not panic in a test, which prevents us from using `require`, but `assert` works just fine here as well.

packages/orchestrator/internal/sandbox/uffd/userfaultfd/userfaultfd_test.go
```diff
@@ -0,0 +1,141 @@
...
+
+			d, err := data.Slice(t.Context(), int64(tt.operationOffset), int64(tt.pagesize))
+			if err != nil {
+				t.Fatal("cannot read content", err)
+			}
+
+			if !bytes.Equal(memoryArea[tt.operationOffset:tt.operationOffset+tt.pagesize], d) {
+				idx, want, got := testutils.DiffByte(memoryArea[tt.operationOffset:tt.operationOffset+tt.pagesize], d)
+				t.Fatalf("content mismatch: want %q, got %q at index %d", want, got, idx)
+			}
```
djeebus: `testify` helps us here too:


```suggestion
			assert.Equal(t, memoryArea[tt.operationOffset:tt.operationOffset+tt.pagesize], d)
```

We can use `assert` instead of `require` here so that we get any further test failures, which can help in diagnosing test failures.

packages/orchestrator/internal/sandbox/uffd/userfaultfd/userfaultfd_test.go
```diff
@@ -0,0 +1,141 @@
...
+
+			if !reflect.DeepEqual(m.Map(), map[uint64]struct{}{tt.operationOffset: {}}) {
+				t.Fatalf("accessed mismatch: should be accessed %v, actually accessed %v", []uint64{tt.operationOffset}, m.Keys())
+			}
```
djeebus: `assert` can help with comparing maps as well:


```suggestion
			assert.Equal(t, map[uint64]struct{}{tt.operationOffset: {}}, m.Map())
```

packages/orchestrator/internal/sandbox/uffd/userfaultfd/userfaultfd_test.go
```diff
@@ -0,0 +1,141 @@
...
+			signalExitErr := fdExit.SignalExit()
+			if signalExitErr != nil {
+				t.Fatal("failed to signal exit", err)
+			}
+
+			select {
+			case <-exitUffd:
+			case <-t.Context().Done():
+				t.Fatal("timeout waiting for uffd to exit")
+			}
```
djeebus: Do tests timeout by default? If not, we probably want to set an explicit timeout here.

djeebus (approved):
Looks good! Just added some opinions, none of which are critical.

packages/orchestrator/internal/sandbox/uffd/userfaultfd/userfaultfd_test.go
ValentaTomas: Kept the old version—we want the diff not to explode all over terminal, reporting just the first difference.

packages/orchestrator/internal/sandbox/uffd/userfaultfd/serve.go
```diff
@@ -153,41 +145,32 @@ outerLoop:
...
-			); errno != 0 {
-				if errno == unix.EEXIST {
-					logger.Debug("UFFD serve page already mapped", zap.Any("offset", offset), zap.Any("pagesize", pagesize))
-
-					// Page is already mapped
-					return nil
-				}
+			copyErr := u.copy(addr, b, pagesize, copyMode)
+			if errors.Is(copyErr, unix.EEXIST) {
+				logger.Debug("UFFD serve page already mapped", zap.Any("offset", addr), zap.Any("pagesize", pagesize))
```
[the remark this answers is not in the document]
ValentaTomas: This seems to be outdated commit.

ValentaTomas:
Included in https://github.com/e2b-dev/infra/pull/1385, very hard to merge separately.

PR #976, opened 2025-08-08
Add the bodyclose linter

djeebus:
implementation: https://github.com/timakin/bodyclose


.golangci.yml
```diff
@@ -28,4 +29,5 @@ linters:
         - -QF1008  # Omit embedded fields from selector expression
 
 run:
-  go: 1.24.3
\ No newline at end of file
+  go: 1.24.3
+  tests: false
```
jakubno: Is there a reason why not to? I can see it's little bit annoying, but on the other side, it can help catch some unintended behavior

djeebus: valid point, I'll clean up the closes in the tests instead.

djeebus:
the last few tests need some refactoring still

PR #980, opened 2025-08-08
Add the contextcheck linter

djeebus:
implementation: https://github.com/kkHAIKE/contextcheck


djeebus:
Too difficult to figure out what's going on here; I'm going to recreate this one file at a time to see which change caused the issue here.

packages/api/internal/dns/server.go
```diff
@@ -228,12 +226,13 @@ func (d *DNS) Start(ctx context.Context, address string, port string) {
 	go func() {
 		<-ctx.Done()
 
-		shutdownCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+		ctx = context.WithoutCancel(ctx)
+		ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
```
0div, line 230: i see a deferred cancellation below and also does WithTimeout re-enable cancellation here?

packages/client-proxy/main.go
```diff
@@ -357,13 +357,14 @@ func run() int {
 		// close the mux server
 		muxServer.Close()
 
-		closeCtx, cancelCloseCtx := context.WithCancel(context.Background())
-		defer cancelCloseCtx()
+		ctx = context.WithoutCancel(ctx)
+		ctx, cancel := context.WithCancel(ctx)
```
0div, line 361: this will re-enable cancellation

packages/shared/pkg/artifacts-registry/registry.go
```diff
@@ -31,17 +31,18 @@ type ArtifactsRegistry interface {
...
 }
 
-func GetArtifactsRegistryProvider() (ArtifactsRegistry, error) {
+func GetArtifactsRegistryProvider(ctx context.Context) (ArtifactsRegistry, error) {
 	provider := RegistryProvider(env.GetEnv(storageProviderEnv, string(DefaultRegistryProvider)))
 
-	setupCtx, setupCtxCancel := context.WithTimeout(context.Background(), 10*time.Second)
-	defer setupCtxCancel()
+	ctx = context.WithoutCancel(ctx)
+	ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
```
0div, line 38: same comment as the other `WithoutCancel` followed by `WithTimeout`

packages/api/internal/dns/server.go
djeebus: the `WithoutCancel` prevents cancellations further up from propagating past this call. The `WithTimeout` call creates a new cancellation opportunity that propagates down, but not up.

packages/client-proxy/main.go
djeebus: Same as above, it prevents cancellations from propagating through this call, then creates its own timeout

Looking at this function though, I can't see a way that the parent context _could_ be cancelled; I added the `WithoutCancel` assuming that there was a reason for the `context.Background()` call in the first place, but I'm not sure there was.

packages/shared/pkg/artifacts-registry/registry.go
djeebus: Same as above

jakubno:
Can we rather not do any changes in this PR, just put `nolint` everywhere and then solve these problems one by one. There have been a few issues, especially related to context and adding new linters, I would like to prevent that from happening as much as possible.

djeebus:
> Can we rather not do any changes in this PR, just put `nolint` everywhere and then solve these problems one by one. There have been a few issues, especially related to context and adding new linters, I would like to prevent that from happening as much as possible.

Yeah, I can do that. This PR got a bit large, eh? 😅

PR #982, opened 2025-08-08
Add versioning and lifecycle rules to the terraform state bucket

jakubno:
Enhancements to Terraform state bucket management:

* Enabled object versioning and soft delete for the Terraform state bucket to retain deleted or replaced objects for 30 days.
* Added lifecycle rules to automatically delete non-live objects older than 30 days and to remove objects if there are more than 50 newer versions, using a temporary lifecycle configuration file during initialization.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,6 +66,21 @@ init:
 	@ printf "Initializing Terraform for env: `tput setaf 2``tput bold`$(ENV)`tput sgr0`\n\n"
 	./scripts/confirm.sh $(TERRAFORM_ENVIRONMENT)
 	gcloud storage buckets create gs://$(TERRAFORM_STATE_BUCKET) --location $(GCP_REGION) --project $(GCP_PROJECT_ID) --default-storage-class STANDARD  --uniform-bucket-level-access > /dev/null 2>&1 || true
+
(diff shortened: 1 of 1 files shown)


jakubno:
> Does soft delete/versioning make sense for terraform state as we cannot effectively go back?
> The only reason I can think of is if we delete the whole bucket.

There have been such cases, right @0div

Makefile
```diff
@@ -66,6 +66,21 @@ init:
 	@ printf "Initializing Terraform for env: `tput setaf 2``tput bold`$(ENV)`tput sgr0`\n\n"
 	./scripts/confirm.sh $(TERRAFORM_ENVIRONMENT)
 	gcloud storage buckets create gs://$(TERRAFORM_STATE_BUCKET) --location $(GCP_REGION) --project $(GCP_PROJECT_ID) --default-storage-class STANDARD  --uniform-bucket-level-access > /dev/null 2>&1 || true
+
+	# Enable object versioning (keeps deleted/replaced objects as older versions)
+	gcloud storage buckets update gs://$(TERRAFORM_STATE_BUCKET) --versioning --soft-delete-duration=30d
```
dobrac, line 71: the 30 days might be maybe a little bit excessive, but it's fine

PR #983, opened 2025-08-08
Add build metrics

dobrac:
Add build metrics


packages/shared/pkg/telemetry/meters.go
```diff
@@ -62,11 +79,15 @@ const (
...
 	ApiOrchestratorCreatedSandboxes: "Number of successfully created sandboxes",
+	BuildResultCounterName:          "Number of template build results by success/failure status",
+	BuildCacheResultCounterName:     "Number of build cache results by hit/miss status",
 }
 
 var counterUnits = map[CounterType]string{
 	SandboxCreateMeterName:          "{sandbox}",
 	ApiOrchestratorCreatedSandboxes: "{sandbox}",
+	BuildResultCounterName:          "{build}",
+	BuildCacheResultCounterName:     "{layer}",
```
jakubno, line 90: use `1` for units

packages/orchestrator/internal/template/build/phases/steps/builder.go
```diff
@@ -89,8 +95,12 @@ func (sb *StepsBuilder) Build(
 		cmd := fmt.Sprintf("%s %s", strings.ToUpper(step.Type), strings.Join(step.Args, " "))
 		sb.UserLogger.Info(phases.LayerInfo(currentLayer.Cached, prefix, cmd, currentLayer.Hash))
 
+		sb.metrics.RecordCacheResult(ctx, "step", currentLayer.Cached)
```
jakubno: can you have constants for `base`, `step`, ..

packages/shared/pkg/telemetry/meters.go
dobrac: the units are not a fraction of total, rather an integer count (https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-units)

packages/orchestrator/internal/template/build/phases/steps/builder.go
dobrac: changed

PR #985, opened 2025-08-09
PoC: Remove unnecessary fail and error text from the errors

dobrac:
Remove unnecessary fail and error test from the errors

### DIFF
--- a/packages/orchestrator/cmd/build-template/main.go
+++ b/packages/orchestrator/cmd/build-template/main.go
@@ -71,7 +71,7 @@ func buildTemplate(
 		EnableConsole: true,
 	})
 	if err != nil {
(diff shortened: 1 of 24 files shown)


dobrac:
> Not sure about this change—where I went through other Go repos, the "problem" seems to be always indicated in the message instead of using the text for just specifying where it originated.
> 
> I get the intention, because it is an _error_ already, so the fact it is not working is almost given, but in plain text and overall it might not be clear.

It’s based on Uber guidelines: https://github.com/uber-go/guide/blob/master/style.md#errors

ValentaTomas:
Also, if we are doing this, enabling some linters for errors format/this might be useful.

PR #986, opened 2025-08-10
Make envd required (it should be already always present)

jakubno:
**API Type and OpenAPI Spec Updates**
- Made `envdVersion` as required in the OpenAPI spec (`openapi.yml`).


packages/api/internal/handlers/sandbox_get.go
```diff
@@ -93,6 +93,11 @@ func (a *APIStore) GetSandboxesSandboxID(c *gin.Context, id string) {
 		zap.L().Error("disk size is not set for the sandbox", logger.WithSandboxID(id))
 	}
 
+	envdVersion := ""
```
dobrac: if envd should be always present, is this invalid state? -> error?

dobrac (changes requested):

packages/api/internal/handlers/sandbox_get.go
jakubno: It should be always present, otherwise the db in some corrupted state (envd column is nullable as we don't know the version when we "wait for the build". I may be worth refactoring and move the build requests to separate table. I think it doesn't make sense to return error here though, I will add log though

dobrac: if not returning an error, what would be the correct behavior when proceeding?

If it should be always present otherwise it's corrupted state, then it should be an error. Imagine relying on the envd version somewhere down the chain, discovering that it's "" instead of a version

jakubno: It's for paused sandbox, so it doesn't really matter that much IMHO. I can return v1.0.0 which will basically say, so if there some check (there could be something in dashboard in the future), it wouldn't expect there's any feature (because we can't really tell)

dobrac: why do we even need the envd version here?

jakubno: Currently we don't, but it's the same type for both paused and running. For running I think we use it to distinguish if the sandbox has metrics or not, if the sandbox has version high enough for inspect etc

dobrac: okay, makes sense. I've missed a bit the usecase, let's either keep it empty or use something like 0.0.0-invalid

dobrac (approved):

PR #987, opened 2025-08-11
Update self-host.md - switch steps 3 & 4

tomassrnka:
First, you need to log in to google cloud, then you can switch env.


jakubno:
This will need little bit more work and will be probably rewritten in near future, closing for now

PR #638, opened 2025-05-12
Specify the clock source as kvm-clock

mengqiqi-github:
image
When the virtual machine is resumed, the time is the time before the snapshot is created. The reason is the time drift of the TSC clock source in the virtualization scenario. It takes about 3 seconds for the chronyd service to synchronize the system time to the accurate time. Using the kvm-clock clock source can solve this problem.
image

PR #988, opened 2025-08-11
Fix not actual sandbox time when clock sync takes too long

sitole:
Currently, there is an issue where sandbox clock sync takes too long. Additionally, requests with a signature are handled before the successful clock sync, and signature expiration is handled incorrectly.


jakubno (changes requested):
Can we move this to some middleware and wait on all requests to prevent this problem in future?

jakubno:
So you will check #638 ?

sitole:
Sandbox kernel is properly registering the KVM clock as the source of time. Synchronization is instant now.
Time sync after resume works properly, and there are no Kernel logs complaining about drifts.

packages/envd/internal/api/init.go
```diff
@@ -52,15 +52,6 @@ func (a *API) PostInit(w http.ResponseWriter, r *http.Request) {
 
 	logger.Debug().Msg("Syncing host")
```
jakubno:
```suggestion
```

jakubno (changes requested):

packages/envd/internal/api/init.go
sitole: There is sync of metadata coming from Firecracker, so it's still relevant IMHO.

packages/envd/internal/services/process/start.go
```diff
@@ -62,7 +61,6 @@ func (s *Service) handleStart(ctx context.Context, req *connect.Request[rpc.Star
 	defer cancel(nil)
 
 	s.logger.Trace().Str(string(logs.OperationIDKey), ctx.Value(logs.OperationIDKey).(string)).Msg("Process start: Waiting for clock to sync")
```
jakubno:
```suggestion
```

packages/envd/internal/services/process/start.go
```diff
@@ -62,7 +61,6 @@ func (s *Service) handleStart(ctx context.Context, req *connect.Request[rpc.Star
...
-	host.WaitForSync()
 	s.logger.Trace().Str(string(logs.OperationIDKey), ctx.Value(logs.OperationIDKey).(string)).Msg("Process start: Clock synced")
```
jakubno:
```suggestion
```

packages/envd/internal/api/init.go
jakubno: but that isn't syncing host, is it?

jakubno: maybe setting mmds

sitole: true, removing.

jakubno (approved):

PR #990, opened 2025-08-11
Fix pagination query

jakubno:
# Description

The problem is that we created the cursor from `StartedAt`, but the query was using `CreatedAt` at DB, causing a potential race condition.


tests/integration/internal/tests/api/sandbox_list_test.go
```diff
@@ -240,7 +240,7 @@ func TestSandboxListPaginationRunningAndPaused(t *testing.T) {
 	sandbox2ID := sbx2.SandboxID
 
 	// Pause the second sandbox
```
dobrac:
```suggestion
	// Pause the first sandbox
```

dobrac (approved):

PR #991, opened 2025-08-11
Report orch metrics

jakubno:
This pull request introduces enhanced host-level resource metrics reporting for orchestrator nodes, including detailed disk metrics, CPU, and memory usage. It adds new data structures and logic to collect, store, and expose these metrics, and updates the API to make them available to consumers.


packages/api/internal/orchestrator/node.go
```diff
@@ -46,6 +46,16 @@ type Node struct {
 	CPUUsage atomic.Int64
 	RamUsage atomic.Int64
 
+	// Host metrics
+	hostCPUPercent     atomic.Int64
+	hostMemoryUsedMiB  atomic.Int64
+	hostCPUCount       atomic.Int64
+	hostMemoryTotalMiB atomic.Int64
```
dobrac: can we report it in bytes?

packages/api/internal/orchestrator/node.go
```diff
@@ -117,12 +127,57 @@ func (n *Node) setMetadata(i *orchestratorinfo.ServiceInfoResponse, nodeID strin
...
+	for i, disk := range info.MetricHostDisks {
+		n.hostDisks[i] = orchestratorinfo.DiskMetrics{
+			MountPoint:     disk.MountPoint,
+			Device:         disk.Device,
+			FilesystemType: disk.FilesystemType,
+			UsedMb:         disk.UsedMb,
+			TotalMb:        disk.TotalMb,
+		}
+	}
+	n.hostDisksMutex.Unlock()
```
dobrac: can you do a separate function where the mutes would be used with defer?

packages/orchestrator/internal/metrics/host.go
```diff
@@ -0,0 +1,108 @@
...
+	MountPoint     string
+	Device         string
+	FilesystemType string
+	UsedMB         int64
+	TotalMB        int64
+	UsedPercent    float64
+}
+
+func GetHostMetrics() (*HostMetrics, error) {
+	// Get CPU count
```
dobrac, line 32: can you split this function to 3?
- CPU
- Memory
- Devices

packages/client-proxy/internal/edge/handlers/service-discovery-orchestrators.go
```diff
@@ -33,11 +33,6 @@ func (a *APIStore) V1ServiceDiscoveryGetOrchestrators(c *gin.Context) {
 				ServiceStatus:        getOrchestratorStatusResolved(info.ServiceStatus),
 
 				Roles: getOrchestratorRolesResolved(info.Roles),
-
-				MetricRamMBUsed:        node.MetricMemoryUsedInMB.Load(),
```
jakubno, line 37: We can delete these as we will have them directly from orchestrator (confirmed by @sitole)

PR #994, opened 2025-08-12
Always use actual service instance ID during gRPC call 

sitole:
Context: For clusters, we are adding the service instance ID to requests used in the edge proxy to determine which orchestrator/template builder should route the request. This ID changes when the orchestrator is restarted.

In production environments, we are not supporting orchestrator restarts. It should never happen to have the same node ID and changed service instance ID, but when developing, restarts are allowed. This was causing the API to be unable to communicate with the orchestrator, because the service instance ID was outdated, as it was issued only during cluster node connection and was never updated.

### DIFF
--- a/packages/api/internal/orchestrator/node.go
+++ b/packages/api/internal/orchestrator/node.go
@@ -66,8 +66,6 @@ type Node struct {
 	createFails   atomic.Uint64
 }
 
-var emptyMetadata = metadata.New(map[string]string{})
-
 func (n *Node) Close() {
 	n.buildCache.Stop()
 }
--- a/packages/api/internal/orchestrator/node.go
+++ b/packages/api/internal/orchestrator/node.go
@@ -202,10 +202,13 @@ func (n *Node) InsertBuild(buildID string) {
 	n.buildCache.Set(buildID, struct{}{}, 2*time.Minute)
 }
 
+// Ensures that GRPC client request context always has the latest service instance ID
 func (n *Node) getClient(ctx context.Context) (*grpclient.GRPCClient, context.Context) {
 	return n.client, metadata.NewOutgoingContext(ctx, n.getClientMetadata())
 }
 
+// Generates metadata with the current service instance ID
+// to ensure we always use the latest ID (e.g. after orchestrator restarts)
 func (n *Node) getClientMetadata() metadata.MD {
 	return metadata.New(map[string]string{consts.EdgeRpcServiceInstanceIDHeader: n.metadata().serviceInstanceID})
 }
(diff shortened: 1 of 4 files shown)


packages/api/internal/orchestrator/node.go
```diff
@@ -203,14 +204,18 @@ func (n *Node) InsertBuild(buildID string) {
...
 }
 
-func (n *Node) GetClient(ctx context.Context) (*grpclient.GRPCClient, context.Context) {
-	return n.client, metadata.NewOutgoingContext(ctx, n.clientMd)
+func (n *Node) getClient(ctx context.Context) (*grpclient.GRPCClient, context.Context) {
+	return n.client, metadata.NewOutgoingContext(ctx, n.getClientMetadata())
+}
+
+func (n *Node) getClientMetadata() metadata.MD {
+	return metadata.New(map[string]string{consts.EdgeRpcServiceInstanceIDHeader: n.metadata().serviceInstanceID})
```
jakubno, line 213: I think you can do directly `n.meta.serviceInstanceID`

packages/api/internal/orchestrator/node.go
```diff
@@ -67,6 +66,8 @@ type Node struct {
 	createFails   atomic.Uint64
 }
 
+var emptyMetadata = metadata.New(map[string]string{})
```
jakubno: what's this for?

dobrac: is this used?

packages/api/internal/orchestrator/node.go
```diff
@@ -203,14 +204,18 @@ func (n *Node) InsertBuild(buildID string) {
 	n.buildCache.Set(buildID, struct{}{}, 2*time.Minute)
 }
 
-func (n *Node) GetClient(ctx context.Context) (*grpclient.GRPCClient, context.Context) {
-	return n.client, metadata.NewOutgoingContext(ctx, n.clientMd)
+func (n *Node) getClient(ctx context.Context) (*grpclient.GRPCClient, context.Context) {
```
jakubno, line 206: Can you add a comment here?
```go
  // generates metadata with the current service instance ID
  // to ensure we always use the latest ID (e.g. after orchestrator restarts)
```

packages/api/internal/orchestrator/node.go
sitole: You cannot, because there can be a race condition as it's updated with every service info run.

jakubno (approved):

dobrac (approved):

PR #1002, opened 2025-08-14
Fix chattr missing in busybox

dobrac:
Fix chattr missing in BusyBox provided by Debian Bookworm.

Fixes
```
provision.sh: 10: chattr: not found
```


dobrac:
@cursor review

packages/orchestrator/internal/template/build/core/systeminit/busybox_1.35.0-x86_64-linux-musl
jakubno, line 1: Why did you decided to go with 1.35.0?
Based on: https://busybox.net/
image

jakubno: Does it make sense to add the binary to the repo directly?

dobrac: > Why did you decided to go with 1.35.0?

I've used the latest built binary from here: https://busybox.net/downloads/binaries/ Completely missed that one is unstable. I'll try to get it from elsewhere.

> Does it make sense to add the binary to the repo directly?

Here are other options:
- adding it to the public bucket and then use similar approaches as with kernel (kinda difficult to update/manage though, it's only important for the provisioning script)
- adding it to the public bucket and download it from there using an url on template manager start
- downloading the latest version on node init script (kinda difficult to update though, especially with template manager)
- we could use Git LFS (we probably should if adding to a Git)

(note: it's size is around 1MB)

dobrac: I'll use this version instead: https://archlinux.org/packages/extra/x86_64/busybox/

jakubno (approved):

packages/orchestrator/internal/template/build/core/systeminit/busybox_1.35.0-x86_64-linux-musl
dobrac: let's keep it in git for now

PR #1003, opened 2025-08-14
Fix error types returned from API

jakubno:
This pull request refactors error handling in API handlers to use a centralized method for sending error responses, making sure the correct error type is returned (json, not a string)

### DIFF
--- a/packages/api/internal/handlers/sandbox_get.go
+++ b/packages/api/internal/handlers/sandbox_get.go
@@ -31,8 +31,8 @@ func (a *APIStore) GetSandboxesSandboxID(c *gin.Context, id string) {
 	if team.ClusterID != nil {
 		cluster, ok := a.clustersPool.GetClusterById(*team.ClusterID)
 		if !ok {
-			zap.L().Error("Sandbox attached cluster not found", logger.WithClusterID(*team.ClusterID))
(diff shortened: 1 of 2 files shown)


dobrac (changes requested):
is this a breaking change? Also, can you please use the telemetry here too? (might be worth refactoring to the sendAPIStoreError)

packages/api/internal/handlers/sandbox_get.go
```diff
@@ -31,8 +31,9 @@ func (a *APIStore) GetSandboxesSandboxID(c *gin.Context, id string) {
 	if team.ClusterID != nil {
 		cluster, ok := a.clustersPool.GetClusterById(*team.ClusterID)
 		if !ok {
-			zap.L().Error("Sandbox attached cluster not found", logger.WithClusterID(*team.ClusterID))
-			c.JSON(http.StatusInternalServerError, fmt.Sprintf("cluster with id %s not found", *team.ClusterID))
+			telemetry.ReportCriticalError(ctx, "error getting cluster by ID", fmt.Errorf("cluster with ID '%s' not found", *team.ClusterID))
```
dobrac: You can do just ReportCriticalError(ctx, message, nil, attributes...)

same applies to other places

jakubno: On the other places, there's actual error, so I kept it

PR #1011, opened 2025-08-18
Fix build caches with constant rootfs path

dobrac:
Fix build caches with constant rootfs path (correctly updating the template version). **This change requires deploying orchestrator first.**

The change in this PR adds download of the `metadata.json` with all templates for the sandbox start. The only current use is the `version` field (used for this fix, the constant rootfs is distinguished based on this).

This version and additional metadata could be also saved in the API database and there would be no need for this metadata file. Unfortunately, this approach is not compatible with the BYOC, where the customer data need to reside in their deployment, and not our API database. 

The usage for the metadata.json file altogether is now:
- version (constant vs dynamic rootfs path), used in orchestrator
- fromTemplate (when building a template based on another template), used in template manager

Potential additional usages:
- replace DB FC version and Kernel version with the metadata, used in orchestrator


dobrac:
> The script builder's template version selection changed from an exact match (== 1) to a range check (<= 1). This means templates with an implicit version 0 (from LocalTemplate.Metadata()) now incorrectly use templateV1 instead of templateV2, which can lead to incorrect rootfs path handling.

this is wanted change

ValentaTomas:
@dobrac Is the metadata.json the previous header metadata, but extracted, or something separate?

dobrac:
> @dobrac Is the metadata.json the previous header metadata, but extracted, or something separate?

It's this `Template` struct: https://github.com/e2b-dev/infra/blob/d8ab543d3f30dafcf454fbfa54f8dd2f14ef5ebd/packages/orchestrator/internal/template/metadata/template_metadata.go#L36

packages/orchestrator/internal/sandbox/snapshot.go
```diff
@@ -53,21 +55,28 @@ func (s *Snapshot) Upload(
 	templateBuild := storage.NewTemplateBuild(
 		s.MemfileDiffHeader,
 		s.RootfsDiffHeader,
-		persistance,
+		persistence,
 		templateFiles,
 	)
 
+	metadataReader, err := os.Open(s.Metafile.Path())
```
jakubno: why do you open here and the rest in open in upload? Is there reason why not to keep it consistent?

packages/orchestrator/internal/template/metadata/template_metadata.go
```diff
@@ -5,83 +5,139 @@ import (
...
-	Version      string                       `json:"version"`
-	Template     storage.TemplateFiles        `json:"template"`
-	Metadata     sandboxtools.CommandMetadata `json:"metadata"`
-	Start        *StartMetadata               `json:"start,omitempty"`
-	FromImage    *string                      `json:"from_image,omitempty"`
-	FromTemplate *FromTemplateMetadata        `json:"from_template,omitempty"`
+type Template struct {
+	Version      uint64                `json:"version"`
+	Template     storage.TemplateFiles `json:"template"`
+	Metadata     Command               `json:"metadata"`
```
jakubno: this is little bit weird, as you have in `LayerResult` field 
```
	Metadata metadata.Template
```

and then `Template` also has `Metadata`

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -563,6 +568,20 @@ func (s *Sandbox) Pause(
 	ctx context.Context,
 	tracer trace.Tracer,
 	snapshotTemplateFiles storage.TemplateCacheFiles,
+) (*Snapshot, error) {
+	meta, err := s.template.Metadata()
```
jakubno: do you think this is cleaner than having only 1 pause and handling this outside?
Also add a comment explaining why we are calling `PauseWithMetadata` during build

jakubno (changes requested):

packages/orchestrator/internal/sandbox/snapshot.go
dobrac: good point, will redo

packages/orchestrator/internal/sandbox/sandbox.go
dobrac: ~~you're right, it's better to have just one method now as the pause for the sandbox is very specific use case~~

EDIT: I'll redo it a bit

dobrac: refactored to alway pass the metadata

packages/orchestrator/internal/template/metadata/template_metadata.go
dobrac: renamed to `Context` (or if you have any better idea?)

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -673,6 +664,7 @@ func (s *Sandbox) PauseWithMetadata(
 		return nil, fmt.Errorf("error while post processing: %w", err)
 	}
 
+	m.Template = snapshotTemplateFiles.TemplateFiles
```
jakubno: what's the reason for this? It should be the same, isn't?

dobrac: thx, forgotten line after the refactor

```suggestion
```

PR #1012, opened 2025-08-19
Allow multiple ENV variables as a single layer

dobrac:
Allow multiple ENV variables as a single layer

Extends the syntax from:
```
[key value]
```

To:
```
[key1 value1 key2 value2 ...]
```

### DIFF
--- a/packages/orchestrator/internal/template/build/commands/env.go
+++ b/packages/orchestrator/internal/template/build/commands/env.go
@@ -29,8 +29,12 @@ func (e *Env) Execute(
 	cmdType := strings.ToUpper(step.Type)
 	args := step.Args
 	// args: [key1 value1 key2 value2 ...]
-	if len(args) == 0 || len(args)%2 != 0 {
-		return sandboxtools.CommandMetadata{}, fmt.Errorf("%s requires a key and value arguments", cmdType)
+	if len(args) == 0 {
+		return sandboxtools.CommandMetadata{}, fmt.Errorf("%s does not support passing no arguments", cmdType)
+	}
+
+	if len(args)%2 != 0 {
+		return sandboxtools.CommandMetadata{}, fmt.Errorf("%s requires both a key and value arguments", cmdType)
 	}
 
 	envVars := maps.Clone(cmdMetadata.EnvVars)


packages/orchestrator/internal/template/build/commands/env.go
```diff
@@ -28,12 +28,20 @@ func (e *Env) Execute(
 ) (sandboxtools.CommandMetadata, error) {
 	cmdType := strings.ToUpper(step.Type)
 	args := step.Args
-	// args: [key value]
-	if len(args) < 2 {
-		return sandboxtools.CommandMetadata{}, fmt.Errorf("%s requires a key and value argument", cmdType)
+	// args: [key1 value1 key2 value2 ...]
+	if len(args) < 2 || len(args)%2 != 0 {
```
ValentaTomas: `len(args) != 0` instead of `len(args) < 2` might be cleaner here, as mod 2 already expresses the `!= 1`.

packages/orchestrator/internal/template/build/commands/env.go
```diff
@@ -28,12 +28,20 @@ func (e *Env) Execute(
...
+		return sandboxtools.CommandMetadata{}, fmt.Errorf("%s requires a key and value arguments", cmdType)
 	}
 
-	return saveEnvMeta(ctx, tracer, proxy, sandboxID, cmdMetadata, args[0], args[1])
+	for i := 0; i < len(args)-1; i += 2 {
+		m, err := saveEnvMeta(ctx, tracer, proxy, sandboxID, cmdMetadata, args[i], args[i+1])
+		if err != nil {
+			return sandboxtools.CommandMetadata{}, err
+		}
+		cmdMetadata = m
```
ValentaTomas: Do we need to pass the cmdMetadata to the `saveEnvMeta` whole, if we are just modifying one field?

The same thing for the `Execute`—I was initially confused about modifying *and* returning argument that was passed.

ValentaTomas:
Nothing blocking, added just style notes.

packages/orchestrator/internal/template/build/commands/env.go
dobrac: I know, I wanted to be more explicit that you need at least one full pair. I'll change it

packages/orchestrator/internal/template/build/commands/env.go
dobrac: > Do we need to pass the cmdMetadata to the saveEnvMeta whole, if we are just modifying one field?

no, will redo

> The same thing for the Execute—I was initially confused about modifying and returning argument that was passed.

what do you mean here? The idea behind passing and returning the argument is for the purpose of immutability

packages/orchestrator/internal/template/build/commands/env.go
```diff
@@ -28,12 +28,20 @@ func (e *Env) Execute(
...
-	// args: [key value]
-	if len(args) < 2 {
-		return sandboxtools.CommandMetadata{}, fmt.Errorf("%s requires a key and value argument", cmdType)
+	// args: [key1 value1 key2 value2 ...]
+	if len(args) < 2 || len(args)%2 != 0 {
+		return sandboxtools.CommandMetadata{}, fmt.Errorf("%s requires a key and value arguments", cmdType)
 	}
 
-	return saveEnvMeta(ctx, tracer, proxy, sandboxID, cmdMetadata, args[0], args[1])
+	for i := 0; i < len(args)-1; i += 2 {
```
ValentaTomas, line 41: Won't thing behave the same with or without the `-1` because of the `+=2` and `%2`?

dobrac: yes, if you don't forget the %2 before, then it will behave the same. This is just to be defensive in case of future refactorings

packages/orchestrator/internal/template/build/commands/env.go
ValentaTomas: Got it, I also felt that the < 2 might be more explicit, but not sure if much.

packages/orchestrator/internal/template/build/commands/env.go
ValentaTomas: > what do you mean here? The idea behind passing and returning the argument is for the purpose of immutability

Got it on the level of the argument value, but I was slightly confused by the `cmdMetadata = m` on the variable level here.

ValentaTomas: I know you are using it for the loop this way, just was unexpected, so wanted to mention it. If you are refactoring the envd from cmdMetadata it will probably work slightly differently, so no need to tackle this separately.

dobrac: > Got it on the level of the argument value, but I was slightly confused by the cmdMetadata = m on the variable level here.

not sure if there is any more idiomatic way to do this

packages/orchestrator/internal/template/build/commands/env.go
ValentaTomas: Oh, it is error returning—`len(args) == 0` then? I think explicitly splitting these error conditions might be actually most clear.

dobrac: I didn't like this much:

```
	if len(args) < 2 {
		return sandboxtools.CommandMetadata{}, fmt.Errorf("%s requires a key and value argument", cmdType)
	}
	if len(args)%2 != 0 {
		return sandboxtools.CommandMetadata{}, fmt.Errorf("%s requires a key and value arguments", cmdType)
	}
```

ValentaTomas: Is 0 passed env vars to the layer `ENV` valid?

dobrac: no, it shouldn't be

ValentaTomas: Damn, just the loop and one check would be nice :D

ValentaTomas: I would just with the split error~

```go
	if len(args) == 0 {
		return sandboxtools.CommandMetadata{}, fmt.Errorf("%s does not support passing no arguments", cmdType)
	}

	if len(args)%2 != 0 {
		return sandboxtools.CommandMetadata{}, fmt.Errorf("%s requires both a key and value arguments", cmdType)
	}
```

Is this error directly passed to users as build logs?

dobrac: > I would just with the split error~

ok

> Is this error directly passed to users as build logs?

yes, but it should be handled in the SDK so these errors never show to the user under normal circumstances

PR #1013, opened 2025-08-19
Refactor node / orchestrator

jakubno:
Refactors node out from orchestrator


packages/api/internal/orchestrator/node.go
jakubno, line 1: split to different files in nodes package

packages/api/internal/orchestrator/nodes/node.go
```diff
@@ -0,0 +1,187 @@
...
+		Commit:            i.ServiceVersionCommit,
+		Version:           i.ServiceVersion,
+	}
+
+	clusterID := uuid.Nil
+	n := &Node{
+		NomadNodeShortID: UnknownNomadNodeShortID,
+		ClusterID:        clusterID,
+		ID:               i.NodeID,
+		IPAddress:        "",
```
jakubno: cluster node doesn't have IP address?

packages/api/internal/orchestrator/cache.go
```diff
@@ -157,123 +154,73 @@ func (o *Orchestrator) syncClusterDiscoveredNodes(ctx context.Context) {
...
 	defer childSpan.End()
 
-	nodeFound := false
-	for range syncMaxRetries {
-		cluster, clusterFound := o.clusters.GetClusterById(node.Info.ClusterID)
-		if !clusterFound {
-			continue
-		}
+	nodeFound := true
+	defer func() {
```
jakubno: removed the retries as it didn't make sense, it's read from in memory map

packages/api/internal/orchestrator/nodes/node.go
sitole: It does, but it's useless for direct communication because it cannot be accessed.

packages/api/internal/orchestrator/nodes/node.go
```diff
@@ -0,0 +1,185 @@
...
+
+type NomadServiceDiscovery struct {
+	NomadNodeShortID string
+
+	OrchestratorAddress string
+	IPAddress           string
+}
+
+type Node struct {
+	NomadNodeShortID string
```
sitole, line 40: Can we mark it as deprecated so no one will use it in new code?

packages/api/internal/orchestrator/nodes/node.go
```diff
@@ -0,0 +1,185 @@
...
+}
+
+func (n *Node) CloseWithClient() error {
+	err := n.client.Close()
+	if err != nil {
+		zap.L().Error("Error closing connection to node", zap.Error(err), logger.WithNodeID(n.ID))
+	}
+
+	return n.Close()
+}
```
sitole: What `Close()` also canceling the client connection? It should not be done for clusters as the connection is shared for the whole cluster.

A close call is missed during cluster node removal.

packages/api/internal/orchestrator/placement/placement.go
```diff
@@ -0,0 +1,19 @@
...
+type SandboxResources struct {
+	CpuCount int64
+	RamMib   int64
+}
+
+// Algorithm defines the interface for sandbox placement strategies.
+// Implementations should choose an optimal node based on available resources
+// and current load distribution.
+type Algorithm interface {
+	ChooseNode(ctx context.Context, nodes []*nodes.Node, nodesExcluded map[string]struct{}, resources SandboxResources) (*nodes.Node, error)
```
sitole: Maybe renamed "resources" to something like "allocation", or "requested"?

packages/api/internal/orchestrator/client.go
```diff
@@ -177,6 +69,62 @@ func (o *Orchestrator) GetClient(ctx context.Context, clusterID uuid.UUID, nodeI
...
+func (o *Orchestrator) GetClusterNodes(clusterID uuid.UUID) []*nodes.Node {
+	clusterNodes := make([]*nodes.Node, 0)
+	for _, n := range o.nodes.Items() {
+		if n.ClusterID == clusterID {
+			clusterNodes = append(clusterNodes, n)
+		}
+	}
+
+	return clusterNodes
+}
```
sitole, line 116: If we dont want to do this for every node allocation call, we can store nodes in [clusterID][nodeID] map to speed up?

sitole (changes requested):

packages/api/internal/orchestrator/client.go
jakubno: We could, but I would leave it up to next PR, this is just moved

sitole: Agree

PR #1015, opened 2025-08-19
Make the interface support composability

djeebus:
The previous `ReadFrom(src io.Reader) (int64, error)` method prevented us from streaming the source to multiple destinations efficiently. Turns out we always have a `[]byte` array when calling this method, so changing the method to accept the byte array actually simplifies our code.

It also simplifies the TemplateBuild.Upload function by allowing the gcp object provider to decide whether multipart or one-shot uploading is more efficient.


packages/shared/pkg/storage/storage_google.go
```diff
@@ -21,16 +21,15 @@ import (
...
-	gcloudMaxRetries               = 3
+	googleReadTimeout              = 10 * time.Second
+	googleOperationTimeout         = 5 * time.Second
+	googleBufferSize               = 2 << 21
+	googleInitialBackoff           = 10 * time.Millisecond
+	googleMaxBackoff               = 10 * time.Second
+	googleBackoffMultiplier        = 2
+	googleMaxAttempts              = 10
 	gcloudDefaultUploadConcurrency = 16
+	gcloudParallelMinFileSize      = 10 * 1024 * 1024
```
jakubno: why are you not using `storage.ChunkSize`?

packages/shared/pkg/storage/storage_google.go
```diff
@@ -222,6 +221,26 @@ func (g *GCPBucketStorageObjectProvider) WriteFromFileSystem(path string) error
 	objectName := g.path
 	filePath := path
 
+	fileInfo, err := os.Stat(filePath)
```
jakubno, line 223: wouldn't it make sense to have this as a part of the multipart uploader? If we ever decide to use it elsewhere, it make sense to do it the same way, right?

ValentaTomas (approved):

packages/shared/pkg/storage/storage_google.go
ValentaTomas: I think it is effectively unrelated to the chunk size as that was optimized for download+we don't want to cache everything, so also for granularity.

jakubno: But then the comment doesn't make sense as the direct upload is limit is smaller than chunk size, so you multipart with 1 chunk for sizes 10-50MiB. I would be okay if the limit would be higher than chunk size, this is rather weird

ValentaTomas: Oh, I though you meant https://github.com/e2b-dev/infra/blob/tweak-interface/packages/orchestrator/internal/sandbox/block/chunk.go#L20 instead of https://github.com/e2b-dev/infra/blob/tweak-interface/packages/shared/pkg/storage/gcp_multipart.go#L26

ValentaTomas: Yes, this should be https://github.com/e2b-dev/infra/blob/tweak-interface/packages/shared/pkg/storage/gcp_multipart.go#L26 so it is not confusing.

packages/shared/pkg/storage/storage_google.go
ValentaTomas: The multipart effectively allows you to not know how big the final file will be, which we plan to exploit with compression, so introducing this check on the level of multipart would be complicated.

packages/shared/pkg/storage/storage_google.go
djeebus: Ah, I was looking for that, but missed it. Thanks!

djeebus: Pointed at the right place 👍

jakubno (approved):

PR #1018, opened 2025-08-20
Support Kubernetes pod service discovery

sitole:
Support for Edge API to take orchestrator/edge target from Kubernetes Pods API.

### DIFF
--- a/packages/client-proxy/internal/service-discovery/k8s-service-discovery.go
+++ b/packages/client-proxy/internal/service-discovery/k8s-service-discovery.go
@@ -89,6 +89,8 @@ func (sd *K8sServiceDiscovery) sync(ctx context.Context) {
 	foundPods := make(map[string]string)
 	for _, pod := range list.Items {
 		ip := pod.Status.PodIP
+
+		// Allow to optionally switch and use HostIP as service discovery entry
 		if sd.hostIP {
 			ip = pod.Status.HostIP
 		}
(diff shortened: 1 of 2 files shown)


sitole:
I will add a switch for pod/host ip address

packages/client-proxy/internal/service-discovery/k8s-service-discovery.go
```diff
@@ -0,0 +1,112 @@
...
+		sd.logger.Error("Failed to describe pods", zap.Error(err))
+		return
+	}
+
+	foundPods := make(map[string]string)
+	for _, pod := range list.Items {
+		ip := pod.Status.PodIP
+		if sd.hostIP {
+			ip = pod.Status.HostIP
+		}
```
dobrac: can you add a comment explaining the if?

packages/client-proxy/internal/service-discovery/k8s-service-discovery.go
```diff
@@ -0,0 +1,112 @@
...
+
+		key := fmt.Sprintf("%s:%d", ip, sd.port)
+		item := ServiceDiscoveryItem{
+			NodeIP:   ip,
```
dobrac, line 100: does this work across multiple nodes? I'm not 100% sure the pod IP is unique for the whole cluster

dobrac (changes requested):

dobrac:
just nits (for some reason Github review comment didn't work for me)

packages/client-proxy/internal/service-discovery/k8s-service-discovery.go
sitole: We are using GCP/AWS network providers that are using VPC IPs assigned to pods, so yes, they are unique.
They will probably always be unique, as you need to network between them.

Also you can still fallback to host ip if needed.

dobrac (approved):

PR #1020, opened 2025-08-20
Validate timestamps for clickhouse

jakubno:
Unifies and improves date validation logic for Clickhouse metrics endpoints


packages/api/internal/handlers/sandbox_metrics.go
```diff
@@ -46,9 +47,11 @@ func (a *APIStore) GetSandboxesSandboxIDMetrics(c *gin.Context, sandboxID string
 		return
 	}
 
-	// Validate time range parameters
-	if start.After(end) {
-		a.sendAPIStoreError(c, http.StatusBadRequest, "start time cannot be after end time")
+	// We don't want to change the start and end times anymore
+	start, end, err = utils.ValidateDates(nil, nil, start, end)
```
dobrac: can you do another helper function for this use case instead of passing `nil, nil`?

dobrac (changes requested):

PR #1021, opened 2025-08-20
Fix the cache key

djeebus:
`go.work.sum` is excluded from the repo since #869 , so the shortcut introduced in #1016 no longer works. This change looks up all go.mod/go.sum files without requiring full recursion.

### DIFF
--- a/.github/actions/build-packages/action.yml
+++ b/.github/actions/build-packages/action.yml
@@ -17,7 +17,7 @@ runs:
       uses: actions/cache@v4
       with:
         path: ${{ steps.go-cache-paths.outputs.go-build }}
-        key: ${{ runner.os }}-go-build-${{ hashFiles('go.work', 'go.work.sum') }}
+        key: ${{ runner.os }}-go-build-${{ hashFiles('go.work', 'go.work.sum', 'packages/*/go.mod', 'packages/*/go.sum', 'tests/*/go.mod', 'tests/*/go.sum') }}
 
     - name: Build Packages
       run: |


.github/actions/build-packages/action.yml
```diff
@@ -17,7 +17,7 @@ runs:
       uses: actions/cache@v4
       with:
         path: ${{ steps.go-cache-paths.outputs.go-build }}
-        key: ${{ runner.os }}-go-build-${{ hashFiles('go.work', 'go.work.sum') }}
+        key: ${{ runner.os }}-go-build-${{ hashFiles('go.work', 'go.work.sum', 'packages/*/go.mod', 'packages/*/go.sum', 'tests/*/go.mod', 'tests/*/go.sum') }}
```
jakubno, line 20: why is this better than the previous glob `**/go.mod` and `*/go.sum`?

djeebus: I was running into an issue with the nfs branch where it couldn't delve into some of the folders, since they were created by firecracker and weren't accessible (different group, permissions were 0600 or similar). Since it couldn't recurse those folders the key generation failed, and therefor so did the build. This prevents full recursion, and since we know where the go.mod files are, this seemed like a safer path to take. `go.work.sum` was a much safer choice though.

djeebus: You can see the error [here](https://github.com/e2b-dev/infra/actions/runs/17049363106/job/48333762026)

```
##[debug]Error: EACCES: permission denied, scandir '/home/runner/work/infra/infra/packages/orchestrator/.e2b-slab-cache'
##[debug]    at async Object.readdir (node:internal/fs/promises:952:18) {
##[debug]  errno: -13,
##[debug]  code: 'EACCES',
##[debug]  syscall: 'scandir',
##[debug]  path: '/home/runner/work/infra/infra/packages/orchestrator/.e2b-slab-cache'
##[debug]}
```

PR #1024, opened 2025-08-20
Centralize CI caching

djeebus:
if you ignore the go.work.sum file, this PR is +45/-79, which implies that using go.work.sum is a little easier to support than avoiding it.

This also includes a few cache tweaks to centralize initializing and cache golang dependencies and test results into a single action, and uses it more consistently. This reduces build times by a possible ~1-5 minutes, depending on changes made.

### DIFF
--- a/.github/actions/go-setup-cache/action.yml
+++ b/.github/actions/go-setup-cache/action.yml
@@ -23,7 +23,7 @@ runs:
             'packages/*/go.mod',
             'packages/*/go.mod.sum',
             'tests/*/go.mod',
-            'tests/*/go.mod.sum',
+            'tests/*/go.mod.sum'
           ) }}
         restore-keys: |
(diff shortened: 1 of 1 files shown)


djeebus:
I don't see anything in your link about them being host machine specific, although the last comment by a golang maintainer does imply that it should be checked in:

> But if you are checking in your go.work file it makes sense to check in the go.work.sum file for consistency for the same reason that a go.sum is checked in: to guarantee that others doing a build in your repo are getting the same results.

Changes to one package's go.mod file will change other packages in our current system. As an example, go to `packages/api/go.mod`, upgrade github.com/oapi-codegen/runtime to v1.1.2 and run `go work sync` and check `git status`.

Regardless, the go.work.sum reintroduction turned out to be the least important part of this PR. I've reverted those changes so we can get the performance improvements. Thanks for the review!

jakubno (approved):

PR #1027, opened 2025-08-21
Decrease the default of listed sandboxes

jakubno:
Eeducing the default and maximum limit for returned sandboxes from 1000 to 100.

### DIFF
--- a/packages/api/internal/handlers/sandboxes_list.go
+++ b/packages/api/internal/handlers/sandboxes_list.go
@@ -26,7 +26,7 @@ import (
 	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
 )
 
-const defaultSandboxListLimit int32 = 1000
+const defaultSandboxListLimit int32 = 100
 
 func (a *APIStore) getPausedSandboxes(ctx context.Context, teamID uuid.UUID, runningSandboxesIDs []string, metadataFilter *map[string]string, limit int32, cursorTime time.Time, cursorID string) ([]utils.PaginatedSandbox, error) {
 	// Apply limit + 1 to check if there are more results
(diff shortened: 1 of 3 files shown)


ValentaTomas:
Also mentioned here https://github.com/e2b-dev/E2B/blob/main/packages/js-sdk/src/sandbox/sandboxApi.ts#L101

jakubno:
> Also mentioned here https://github.com/e2b-dev/E2B/blob/main/packages/js-sdk/src/sandbox/sandboxApi.ts#L101

That's different repo

spec/openapi.yml
```diff
@@ -1225,8 +1225,8 @@ paths:
             type: integer
             format: int32
             minimum: 1
-            default: 1000
-            maximum: 1000
+            default: 100
+            maximum: 100
```
dobrac: is this soft or hard limit? Meaning if you specify more than 100, will it throw an error, or decrease the number to 100?

jakubno: Hard limit
```json
{
  "code": 400,
  "message": "validation error: error in openapi3filter.RequestError: parameter \"limit\" in query has an error: number must be at most 100"
}
```

dobrac: in that case it would be a breaking change for anyone who have specified more than 100 (after the Infra deployment)

packages/api/internal/handlers/sandboxes_list.go
```diff
@@ -121,6 +121,11 @@ func (a *APIStore) GetV2Sandboxes(c *gin.Context, params api.GetV2SandboxesParam
 		limit = *params.Limit
 	}
 
+	// Clip limit to max
+	if limit > defaultSandboxListLimit {
```
dobrac: nit: split the default to two variables (max, default)

dobrac (changes requested):

PR #1030, opened 2025-08-21
feat: allow passing login to Docker registries

dobrac:
### Summary
This PR introduces support for authenticating with private Docker registries when building templates from Docker images. Previously, users could only pull public images or had to rely on default Docker daemon credentials. With this change, users can now provide registry-specific authentication credentials directly through the API.

### Problem
Users need to build E2B templates from private Docker images stored in various container registries (AWS ECR, GCP Container Registry, Docker Hub, GitHub Container Registry, etc.). Without proper authentication support, they were unable to use private images, limiting the flexibility of the template build system.

### Solution
Implemented a flexible authentication system that supports multiple registry types:

#### 🔐 Supported Registry Types

1. **AWS ECR** (`type: "aws"`)
   - Authentication via AWS Access Key ID, Secret Access Key, and Region
   - Automatically handles ECR token generation

2. **GCP Container Registry** (`type: "gcp"`)  
   - Authentication via Service Account JSON
   - Supports both gcr.io and Artifact Registry

3. **General Registries** (`type: "registry"`)
   - Username/password authentication
   - Works with Docker Hub, GitHub Container Registry, GitLab Registry, and other standard registries

#### AWS ECR
```json
{
  "fromImage": "123456789.dkr.ecr.us-east-1.amazonaws.com/my-app:latest",
  "fromImageRegistry": {
    "type": "aws",
    "awsAccessKeyId": "AKIA...",
    "awsSecretAccessKey": "secret...",
    "awsRegion": "us-east-1"
  }
}
```

#### GCP Container Registry
```json
{
  "fromImage": "gcr.io/my-project/my-app:latest",
  "fromImageRegistry": {
    "type": "gcp",
    "serviceAccountJson": "{\"type\":\"service_account\",...}"
  }
}
```

#### Docker Hub / GitHub Container Registry / Others
```json
{
  "fromImage": "ghcr.io/myorg/my-app:latest",
  "fromImageRegistry": {
    "type": "registry",
    "username": "myusername",
    "password": "my-token"
  }
}
```


dobrac:
@cursor review

packages/api/internal/template-manager/create_template.go
```diff
@@ -87,6 +88,11 @@ func (tm *TemplateManager) CreateTemplate(
 		readyCmd = *readyCommand
 	}
 
+	imageRegistry, err := convertImageRegistry(fromImageRegistry)
```
sitole, line 91: Can we do version check so it will no be requested on older template manager not supporting it?

dobrac: do you mean template versions v1 or old template manager versions? For both, it won't be requested as this would be always nil (+ v1 templates go through different path)

dobrac: if you mean to throw error for requesting it on older template manager versions, then do we have any mechanism for the versioning already in place?

sitole: > if you mean to throw error for requesting it on older template manager versions, then do we have any mechanism for the versioning already in place?

If works correctly, we can skip it. I was just thinking if we should return proper error if not supported?

You should be able to get template version via clusterID and NodeID although its done only for clustered one. We are not getting version for local template manager.
```golang
c, _ := tm.edgePool.GetClusterById(*clusterID)
i, _ := c.GetTemplateBuilderByNodeID(*clusterNodeID)
i.ServiceVersion
```

dobrac: It will silently ignore it for outdated template manager and new SDK. This is not an issue for prod, but it might be relevant for BYOC.

Let's first unite the LocalTemplateManager implementation with the ClusterTemplateManager (this will allow us to have multiple template managers too). Then we have the ServiceVersion (which is otherwise missing) and then we can do proper feature/version guarded checks.

PR #1036, opened 2025-08-22
Make sure logs collector ip is named properly and its always just IPv4

sitole:
Needed for K8s support, where you can set env with host/pod ip, but you cannot adjust it with, for example, `http://` prefix.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -193,7 +193,7 @@ module "nomad" {
 
   # API
   api_machine_count                         = var.api_cluster_size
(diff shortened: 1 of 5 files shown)


jakubno:
I would still fix it as we use ip and expect address

sitole:
I opted for IP naming over address because it's used in sandbox iptables, where the IP address and subnet size are added, and we were always stripping the http prefix.

jakubno (approved):

PR #1041, opened 2025-08-25
Support multiple template builders

sitole:
Inject a nil UUID as the local cluster to the edge pool and utilize the Edge API for template managers' service discovery.
This is also an easy entry to using the Edge API for cluster nodes discovery.

- [x] Support for integration tests
- [x] Support for multiple builds node setup
- [x] Test if build node switch is gracefully done
- [x] Dynamic ENV variable for edge secret token and DNS target

https://linear.app/e2b/issue/E2B-1819/add-service-discovery-logic-for-template-manager-in-api

### DIFF
--- a/packages/api/internal/template-manager/template_manager.go
+++ b/packages/api/internal/template-manager/template_manager.go
@@ -175,7 +175,7 @@ func (tm *TemplateManager) DeleteBuild(ctx context.Context, t trace.Tracer, buil
 
 	client, err := tm.GetClusterBuildClient(clusterID, nodeID)
 	if err != nil {
-		return fmt.Errorf("failed to get builder edgeHttpClient: %w", err)
+		return fmt.Errorf("failed to get builder client: %w", err)
 	}
 
 	reqCtx := metadata.NewOutgoingContext(ctx, client.GRPC.Metadata)


packages/nomad/edge.hcl
```diff
@@ -94,7 +94,8 @@ job "client-proxy" {
 
         SERVICE_DISCOVERY_ORCHESTRATOR_PROVIDER             = "DNS"
         SERVICE_DISCOVERY_ORCHESTRATOR_DNS_RESOLVER_ADDRESS = "127.0.0.1:8600" // consul dns resolver
-        SERVICE_DISCOVERY_ORCHESTRATOR_DNS_QUERY            = "orchestrator.service.consul,template-manager.service.consul"
+        SERVICE_DISCOVERY_ORCHESTRATOR_DNS_QUERY            = "template-manager.service.consul"
+      //  SERVICE_DISCOVERY_ORCHESTRATOR_DNS_QUERY          = "orchestrator.service.consul,template-manager.service.consul"
```
sitole: We want to do service discovery only for template managers now

packages/api/internal/template-manager/template_manager.go
```diff
@@ -146,25 +126,12 @@ func (tm *TemplateManager) BuildsStatusPeriodicalSync(ctx context.Context) {
 	}
 }
 
-func (tm *TemplateManager) GetBuildClient(clusterID *uuid.UUID, nodeID string, placement bool) (*BuildClient, error) {
-	if clusterID == nil {
-		return tm.GetLocalBuildClient(placement)
-	} else {
-		return tm.GetClusterBuildClient(*clusterID, nodeID)
-	}
+func (tm *TemplateManager) GetBuildClient(clusterID uuid.UUID, nodeID string) (*BuildClient, error) {
```
dobrac: unnecessary now :yeaay:

packages/api/internal/handlers/template_layer_files_upload.go
```diff
@@ -37,14 +38,14 @@ func (a *APIStore) GetTemplatesTemplateIDFilesHash(c *gin.Context, templateID ap
 		return
 	}
 
-	nodeID, err := a.templateManager.GetAvailableBuildClient(ctx, templateDB.ClusterID)
+	nodeID, err := a.templateManager.GetAvailableBuildClient(ctx, utils.WithDefaultCluster(team.ClusterID))
```
dobrac: nit
```suggestion
	nodeID, err := a.templateManager.GetAvailableBuildClient(ctx, utils.WithDefaultCluster(templateDB.ClusterID))
```

packages/api/internal/edge/pool.go
```diff
@@ -46,7 +71,12 @@ func NewPool(ctx context.Context, tel *telemetry.Client, db *client.Client, trac
 		p.Close()
 	}()
 
-	store := poolSynchronizationStore{pool: p}
+	localCluster, err := takeLocalClusterConfig()
+	if err != nil {
+		return nil, err
+	}
+
```
sitole: When provided, we can append cluster used for local edge-based routing.

packages/api/internal/orchestrator/nodemanager/node.go
```diff
@@ -177,3 +176,7 @@ func (n *Node) Close() error {
 func (n *Node) GetClient(ctx context.Context) (*grpclient.GRPCClient, context.Context) {
 	return n.client, metadata.NewOutgoingContext(ctx, n.getClientMetadata())
 }
+
+func (n *Node) IsNomadManaged() bool {
+	return n.NomadNodeShortID != UnknownNomadNodeShortID
+}
```
sitole, line 182: The previous implementation was checking if clusterID is an empty UUID. Now the local cluster will gain an empty UUID, so we need to check if the node is owned by Nomad instead.

packages/api/internal/orchestrator/cache.go
```diff
@@ -59,26 +58,33 @@ func (o *Orchestrator) keepInSync(ctx context.Context, instanceCache *instance.I
...
+	if !skipSyncingWithNomad {
+		nomadSD, err := o.listNomadNodes(spanCtx)
+		if err != nil {
+			zap.L().Error("Error listing orchestrator nodes", zap.Error(err))
+			return
+		}
+
+		nomadNodes = nomadSD
+	}
+
```
sitole, line 80: For tests we want to skip nomad service discovery. Starting with an empty list to make as few changes as needed.

packages/api/internal/orchestrator/orchestrator.go
```diff
@@ -112,28 +111,17 @@ func New(
...
-			return nil, err
-		}
-	} else {
-		go o.keepInSync(ctx, cache)
-		go o.reportLongRunningSandboxes(ctx)
-	}
+	// For local development and testing, we skip the Nomad sync
+	// Local cluster is used for single-node setups instead
+	skipNomadSync := env.IsLocal()
+	go o.keepInSync(ctx, cache, skipNomadSync)
```
sitole, line 127: For tests, we are now using a local cluster instead, so there is no need to push a hardcoded node here.

packages/nomad/edge.hcl
dobrac: remove the comment if not relevant anymore

packages/api/internal/edge/pool.go
```diff
@@ -27,11 +32,31 @@ type Pool struct {
...
+		return nil, nil
+	}
+
+	clusterToken := os.Getenv(clusterTokenEnv)
+	if clusterToken == "" {
+		return nil, errors.New("no local cluster token provided")
+	}
+
+	return &queries.Cluster{
+		ID:                 uuid.Nil,
```
dobrac: this uuid.Nil should be same const as the default cluster

dobrac (changes requested):
Nice! Thank you!

Can you please also update the title?

packages/api/internal/template-manager/template_manager.go
sitole: nice, good catch!

dobrac:
> ### Bug: Missing IDs in DeleteBuild Struct
> The `CheckAndCancelConcurrentBuilds` function creates `DeleteBuild` structs without `ClusterID` and `NodeID`. The `DeleteBuild` struct now requires these fields, so they default to zero values, which likely causes build deletion to fail.
> 
> `packages/api/internal/handlers/template_start_build.go#L42-L47`
> https://github.com/e2b-dev/infra/blob/385d1b0de133a927837d1d853b71b895f3766812/packages/api/internal/handlers/template_start_build.go#L42-L47
> 
> [Fix in Cursor](https://cursor.com/open?data=eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6ImJ1Z2JvdC12MSJ9.eyJ2ZXJzaW9uIjoxLCJ0eXBlIjoiQlVHQk9UX0ZJWF9JTl9DVVJTT1IiLCJkYXRhIjp7InJlZGlzS2V5IjoiYnVnYm90OjE2Yjg4N2MxLTMwOTctNDVhZi1hMzgxLTg1NDdjZWQ1M2YyOCIsImVuY3J5cHRpb25LZXkiOiJISWU4aXBhdWw0RHZSUHhicXRDMER3eXBqUjZjUEFSVldfWEFRbDY5UDBnIiwiYnJhbmNoIjoic3VwcG9ydC1tdWx0aXBsZS10ZW1wbGF0ZS1tYW5hZ2VycyJ9LCJpYXQiOjE3NTY3Mzg4NDUsImV4cCI6MTc1NzM0MzY0NX0.Uf1UY-LuHduX8NrNBAZe5w4lwQR4JCJ7RE-hdM3NHMmC7U8HxTxrj0clTjjiOf-eXYKChOdrSfRGCsKt49pl7KNYv98psXqI96oANfJrzHCamyX2CO4kRH_4vca4gkmg0ikarp0yINgUKg3xZFrZAfFsxHAB3PZ8mnj2ECAil3rpZKHAHXiKYEVoGJSeYXojczsalv7zZ84La-dxvIenO81nMZstvbBZ0AP2Ae78j1fXRv0viHJhWRtse_TM2LCtrpyZ8naPMsrR42ygL5oLSqOJmdXLfm_J0SUBs4mDZ_AZDz2sA_FR5mFjIc_MzIP9MMtycDLdk8mrSRpgqFdL-A) [Fix in Web](https://cursor.com/agents?data=eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6ImJ1Z2JvdC12MSJ9.eyJ2ZXJzaW9uIjoxLCJ0eXBlIjoiQlVHQk9UX0ZJWF9JTl9XRUIiLCJkYXRhIjp7InJlZGlzS2V5IjoiYnVnYm90OjE2Yjg4N2MxLTMwOTctNDVhZi1hMzgxLTg1NDdjZWQ1M2YyOCIsImVuY3J5cHRpb25LZXkiOiJISWU4aXBhdWw0RHZSUHhicXRDMER3eXBqUjZjUEFSVldfWEFRbDY5UDBnIiwiYnJhbmNoIjoic3VwcG9ydC1tdWx0aXBsZS10ZW1wbGF0ZS1tYW5hZ2VycyIsInJlcG9Pd25lciI6ImUyYi1kZXYiLCJyZXBvTmFtZSI6ImluZnJhIiwicHJOdW1iZXIiOjEwNDEsImNvbW1pdFNoYSI6IjM4NWQxYjBkZTEzM2E5Mjc4MzdkMWQ4NTNiNzFiODk1ZjM3NjY4MTIifSwiaWF0IjoxNzU2NzM4ODQ1LCJleHAiOjE3NTczNDM2NDV9.JioL6ilDsKn1MKJQyWrA7yxuAETsiBWhiYQmthaoJGGTE68ArzngsLKt1iswIgg5bxAnGvvfYLu81stT4MemrZAHnGVchUiFLxBZ1ZyrryWHMygTTXBU-jDWhHLSfPKMK8xCNKyjqm1L4WDA-ZiLBJpfbYEfwCwZgRTwpaxeuzLQ3xZirQL4XFtaUvivxpXt2kixSNkoSmyjxCcCdqHFQhwcYknFaBPnBlgdG7ZMfU8VRyACzbDRIvIxeR-hkyFkROkNF9lYYwc4VKWALHBcTa2zUbz7ACXWwUpLR_UikBoLX_njy_mCOLlX6z7MPfszk7Ph4eSFG5a42OWFq-sQ1Q)

this looks relevant

sitole:
Issue with missing IDs during template build fixed. Will let it here and merge after the current batch is in prod.

PR #1043, opened 2025-08-25
Missing request metadata during node registration

sitole:
After node refactoring, the cluster node is not getting the gRPC metadata needed for proper request handling.
The initial call will fail as metadata is not present.

### DIFF
--- a/packages/api/internal/orchestrator/nodemanager/node.go
+++ b/packages/api/internal/orchestrator/nodemanager/node.go
@@ -109,7 +109,7 @@ func New(
 	return n, nil
 }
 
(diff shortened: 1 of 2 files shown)


packages/api/internal/orchestrator/nodemanager/node.go
```diff
@@ -149,7 +144,8 @@ func NewClusterNode(
 		},
 	}
 
-	nodeInfo, err := client.Info.ServiceInfo(ctx, &emptypb.Empty{})
+	reqCtx := metadata.NewOutgoingContext(ctx, conn.Metadata)
+	nodeInfo, err := conn.Client.Info.ServiceInfo(reqCtx, &emptypb.Empty{})
```
dobrac: can you use the `node.GetClient()` method instead?

dobrac (approved):

PR #1046, opened 2025-08-25
Add some logs to the cache provider

djeebus:


jakubno (approved):
could you make it all lowercase? 🙏🏻

ValentaTomas:
We will be also most probably removing the logs before deploying to prod.

PR #1047, opened 2025-08-25
Do not cache when building a template, only when starting a sandbox

djeebus:
This also cleans up logging and returns a struct rather than an interface in one place.

### DIFF
--- a/packages/orchestrator/internal/sandbox/template/cache.go
+++ b/packages/orchestrator/internal/sandbox/template/cache.go
@@ -101,11 +101,12 @@ func (c *Cache) GetTemplate(
 	kernelVersion,
 	firecrackerVersion string,
 	isSnapshot bool,
+	isBuilding bool,
 ) (Template, error) {
 	persistence := c.persistence
 	// Because of the template caching, if we enable the shared cache feature flag,
 	// it will start working only for new orchestrators or new builds.
-	if c.useNFSCache(isSnapshot) {
+	if c.useNFSCache(isBuilding, isSnapshot) {
 		zap.L().Info("using local template cache", zap.String("path", c.rootCachePath))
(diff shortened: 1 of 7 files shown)


packages/orchestrator/internal/sandbox/template/cache.go
```diff
@@ -192,7 +193,13 @@ func (c *Cache) AddSnapshot(
 	return nil
 }
 
-func (c *Cache) useNFSCache(isSnapshot bool) bool {
+func (c *Cache) useNFSCache(isBuilding bool, isSnapshot bool) bool {
+	if isBuilding {
+		// caching this layer doesn't speed up the next sandbox launch,
+		// as the previous template isn't used to load the oen that's being built.
```
0div:
```suggestion
		// as the previous template isn't used to load the one that's being built.
```

packages/orchestrator/internal/template/build/layer/create_sandbox.go
```diff
@@ -22,7 +22,9 @@ type CreateSandbox struct {
 	fcVersions fc.FirecrackerVersions
 }
 
-func NewCreateSandbox(config sandbox.Config, fcVersions fc.FirecrackerVersions) SandboxCreator {
+var _ SandboxCreator = (*CreateSandbox)(nil)
```
0div, line 25: what's this pattern?

0div (approved):

packages/orchestrator/internal/template/build/layer/create_sandbox.go
djeebus: It ensures that a struct implements an interface at compile time: https://stackoverflow.com/questions/10498547/ensure-a-type-implements-an-interface-at-compile-time-in-go

PR #1048, opened 2025-08-25
Introduce general product usage metrics

0div:
# Description 
Leverage ClickHouse to track usage of our product by teams. Follows Google Analytics' Category/Action/Label model.

Concretely, events would be stored in ClickHouse as follows:
```
┌─name──────┬─type───────────────────┬
│ timestamp │ DateTime64(9)          │
│ team_id   │ UUID                   │
│ category  │ LowCardinality(String) │
│ action    │ LowCardinality(String) │
│ label     │ String                 │
└───────────┴────────────────────────┴
```
And in the code:
```go
err = productUsageBatcher.Push(clickhouse.ProductUsage{
  Timestamp: time.Now().UTC(),
  TeamID:    teamID,
  Category:  string(clickhouse.ProductUsageCategoryArugsAPI),
  Action:    "get_events",
  Label:     metadaJSON
})
```

# Tasks
- [x] refactor clickhouse insert batcher to be generic
- [x] define product usage table schema 
- [x] write product usage queries 
- [x] implement insert product usage batcher


packages/clickhouse/pkg/batcher/product_usage_insert_batcher.go
```diff
@@ -0,0 +1,110 @@
...
+		return nil, err
+	}
+
+	b.Batcher = batcher
+
+	return b, nil
+}
+
+func (b *ProductUsageInsertBatcher) processInsertProductUsageBatch(events []clickhouse.ProductUsage) error {
+	ctx := context.Background()
```
djeebus, line 56: Can we refactor to bring a `context.Context` in from the parent function?

packages/clickhouse/pkg/batcher/product_usage_insert_batcher.go
```diff
@@ -0,0 +1,110 @@
...
+			event.Label,
+		)
+		if err != nil {
+			return fmt.Errorf("error appending product usage event to batch: %w", err)
+		}
+	}
+
+	err = batch.Send()
+	if err != nil {
+		return fmt.Errorf("error sending product usage events batch: %w", err)
```
djeebus: Can we add the number of events processed here? Might be useful in debugging errors.

packages/clickhouse/pkg/batcher/product_usage_insert_batcher.go
```diff
@@ -0,0 +1,110 @@
...
+	return nil
+}
+
+func (b *ProductUsageInsertBatcher) Push(event clickhouse.ProductUsage) error {
+	success, err := b.Batcher.Push(event)
+	if err != nil {
+		return err
+	}
+	if !success {
+		return errors.New("batcher queue is full")
```
djeebus: This should be turned into a global var

packages/clickhouse/pkg/batcher/product_usage_insert_batcher.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	return nil
+}
+
+func (b *ProductUsageInsertBatcher) Close(ctx context.Context) error {
+	stopErr := b.Batcher.Stop()
+	closeErr := b.conn.Close()
```
djeebus, line 97: If `b.conn` was passed in, is it ours to close?

0div: in this case, the insertBatcher.Close is called in orchestrator, not for the conn: https://github.com/e2b-dev/infra/blob/main/packages/orchestrator/main.go#L334-L343

packages/clickhouse/pkg/batcher/product_usage_insert_batcher.go
0div: This is to avoid having to use `WithoutCancel`

djeebus: Avoiding `WithoutCancel` isn't super valuable - as time goes on, we'll probably lean into using context for more useful things (preconfigured loggers, more tracing, etc), so getting into the habit of using context the way that golang expects you to use it is probably worth while. Still, it doesn't block anything now 👍

djeebus (approved):

packages/clickhouse/pkg/batcher/product_usage_insert_batcher.go
0div: I understand your point, the problem is that this function is passed at init of the batcher and is called during flushes (determined by the batcher algo), and the only point where we would pass a context is when calling `Start` which is a long running background routine.

PR #1057, opened 2025-08-26
Enforce max concurrent sandboxes per node in the orchestrator

jakubno:
Enforce max concurrent sandboxes per node in the orchestrator instead of API and increase retry limit

This is a next step for moving towards stateless API.

Starting 3 sandboxes concurrently on a node should be very improbable, so it's better to check during the placement instead of doing extra call to some centralized storage to get the number of concurrently starting sandboxes on the node.

### DIFF
--- a/packages/api/internal/orchestrator/placement/placement.go
+++ b/packages/api/internal/orchestrator/placement/placement.go
@@ -42,6 +42,10 @@ func PlaceSandbox(ctx context.Context, tracer trace.Tracer, algorithm Algorithm,
 		if preferredNode != nil {
 			node = preferredNode
 		} else {
+			if len(nodesExcluded) >= len(clusterNodes) {
+				return nil, fmt.Errorf("no nodes available")
+			}
+
 			node, err = algorithm.chooseNode(ctx, clusterNodes, nodesExcluded, nodemanager.SandboxResources{CPUs: sbxRequest.Sandbox.Vcpu, MiBMemory: sbxRequest.Sandbox.RamMb})
 			if err != nil {
 				return nil, err
--- a/packages/api/internal/orchestrator/placement/placement.go
+++ b/packages/api/internal/orchestrator/placement/placement.go
@@ -62,15 +62,17 @@ func PlaceSandbox(ctx context.Context, tracer trace.Tracer, algorithm Algorithm,
 		span.End()
 		if err != nil {
 			nodesExcluded[node.ID] = struct{}{}
-			node.PlacementMetrics.Fail(sbxRequest.Sandbox.SandboxId)
-			zap.L().Error("Failed to create sandbox", logger.WithSandboxID(sbxRequest.Sandbox.SandboxId), logger.WithNodeID(node.ID), zap.Int("attempt", attempt+1), zap.Error(utils.UnwrapGRPCError(err)))
-
-			node = nil
 
 			st, ok := status.FromError(err)
 			if !ok || st.Code() != codes.ResourceExhausted {
+				node.PlacementMetrics.Fail(sbxRequest.Sandbox.SandboxId)
+				zap.L().Error("Failed to create sandbox", logger.WithSandboxID(sbxRequest.Sandbox.SandboxId), logger.WithNodeID(node.ID), zap.Int("attempt", attempt+1), zap.Error(utils.UnwrapGRPCError(err)))
 				attempt++
+			} else {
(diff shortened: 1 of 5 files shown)


packages/api/internal/orchestrator/placement/placement_least_busy.go
```diff
@@ -63,11 +63,6 @@ func (a *LeastBusyAlgorithm) findLeastBusyNode(clusterNodes []*nodemanager.Node,
 			continue
 		}
 
-		// To prevent overloading the node
-		if node.PlacementMetrics.InProgressCount() > maxStartingInstancesPerNode {
-			continue
-		}
```
dobrac: won't this slowdown the start time when one node is much less used than the others as it will be "always" selected?

also, we still need to keep track of the in progress sandboxes for the cpuUsage, right?

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -45,6 +46,24 @@ func (s *server) Create(ctxConn context.Context, req *orchestrator.SandboxCreate
...
+	// Check if we've reached the max number of starting instances on this node
+	s.startingMu.Lock()
+	startingCount := s.startingSandboxes.Count()
+	if startingCount >= maxStartingInstancesPerNode {
+		s.startingMu.Unlock()
+		telemetry.ReportEvent(ctx, "too many starting sandboxes on node")
+		return nil, status.Errorf(codes.ResourceExhausted, "too many sandboxes starting on this node, please retry")
+	}
+	s.startingSandboxes.Insert(req.Sandbox.SandboxId, struct{}{})
+	s.startingMu.Unlock()
```
dobrac: why not use semaphore?

dobrac (changes requested):

packages/api/internal/orchestrator/placement/placement_least_busy.go
jakubno: This will prevent the race condition with comparison happening before the "reservation"

jakubno: Not sure about that yet about keeping track of the in progress as it will be race-y anyway. With our current stats and with keeping K reasonable small I think there's no need for that

packages/orchestrator/internal/server/main.go
```diff
@@ -37,6 +38,7 @@ type server struct {
 	persistence         storage.StorageProvider
 	featureFlags        *featureflags.Client
 	sandboxEventBatcher batcher.ClickhouseBatcher
+	startingSandboxes   semaphore.Weighted
```
dobrac:
```suggestion
	startingSandboxes   *semaphore.Weighted
```

(also update the creation to not dereference it)

packages/api/internal/orchestrator/placement/placement.go
```diff
@@ -62,6 +66,12 @@ func PlaceSandbox(ctx context.Context, tracer trace.Tracer, algorithm Algorithm,
 			zap.L().Error("Failed to create sandbox", logger.WithSandboxID(sbxRequest.Sandbox.SandboxId), logger.WithNodeID(node.ID), zap.Int("attempt", attempt+1), zap.Error(utils.UnwrapGRPCError(err)))
 
 			node = nil
+
+			st, ok := status.FromError(err)
+			if !ok || st.Code() != codes.ResourceExhausted {
+				attempt++
```
dobrac, line 74: could this infinite loop when all nodes are exhausted? There should be probably some escape

packages/api/internal/orchestrator/placement/placement_least_busy.go
dobrac: > This will prevent the race condition with comparison happening before the "reservation"

this might be a slightly different issue though

> Not sure about that yet about keeping track of the in progress as it will be race-y anyway

we can test it without it, but I wanted to stress now that if one node (e.g. newly started) is present, all request will go to that node and it will be requested for ALL new sandboxes and it will error out (causing the retry to another node -> slowdown in start time)

jakubno: not all, the chance the node will be in the selection is ((n-1) choose k-1) / (n choose k), which is k/n, so as the n >> k, it's fine

I reverted it for least busy, I was still working on it, I will put it in draft next time

packages/api/internal/orchestrator/placement/placement.go
jakubno: case <-ctx.Done():
			return nil, fmt.Errorf("request timed out during %d. attempt", attempt+1)

PR #1064, opened 2025-08-27
Refactor pausing / sandbox cache

jakubno:
# Description

- Fixes race condition in `GetSandbox` when the sandbox is pausing, the same for `ListSandboxes`
- Fixes some race conditions during pause
- Pause and kill now removes the sandbox without waiting for async cache hook
- Fixes bug with deprecated `autopause` on resume being propagated incorrectly

There are still probably some race conditions, but it would require adding locks, which would make this PR giant, so I would start with this refactor and add locks in the following PR.

### DIFF
--- a/packages/api/internal/cache/instance/instance.go
+++ b/packages/api/internal/cache/instance/instance.go
@@ -30,13 +30,6 @@ const (
 	StateKilled  State = "killed"
 )
 
-type OnEvictionType string
-
-const (
-	EvictionPause  OnEvictionType = "pause"
-	EvictionDelete OnEvictionType = "delete"
-)
-
 func NewInstanceInfo(
 	SandboxID string,
 	TemplateID string,


spec/openapi.yml
```diff
@@ -462,7 +462,6 @@ components:
         autoPause:
           type: boolean
           deprecated: true
-          default: false
```
jakubno: the field is deprecated and shouldn't be used, we want to know if the user explicitly asked for the action or not

packages/api/internal/orchestrator/pause_instance.go
```diff
@@ -22,16 +23,35 @@ func (ErrPauseQueueExhausted) Error() string {
...
+	return nil
+}
+
+func (o *Orchestrator) pauseSandbox(ctx context.Context, node *nodemanager.Node, sbx *instance.InstanceInfo) error {
+	// This is doubled if this is called from cache remove hook, but it's fine
+	o.dns.Remove(ctx, sbx.SandboxID, node.IPAddress)
+
 	snapshotConfig := &db.SnapshotInfo{
-		BaseTemplateID:      sbx.TemplateID,
+		BaseTemplateID:      sbx.BaseTemplateID,
```
jakubno: I think this was also a bug, we want to save the initial template, not the previous snapshot

packages/api/internal/orchestrator/orchestrator.go
```diff
@@ -211,8 +213,3 @@ func (o *Orchestrator) Close(ctx context.Context) error {
 
 	return errors.Join(errs...)
 }
-
-// WaitForPause waits for the instance to be paused and returns the node info where the instance was paused on.
-func (o *Orchestrator) WaitForPause(ctx context.Context, sandboxID string) (nodeID string, err error) {
```
jakubno, line 215: moved to instance itself

packages/api/internal/orchestrator/create_instance.go
```diff
@@ -235,9 +235,9 @@ func (o *Orchestrator) CreateSandbox(
 	if cacheErr != nil {
 		telemetry.ReportError(ctx, "error when adding instance to cache", cacheErr)
 
-		deleted := o.DeleteInstance(ctx, sbx.SandboxID, false)
-		if !deleted {
-			telemetry.ReportEvent(ctx, "instance wasn't found in cache when deleting")
+		err := o.RemoveInstance(ctx, instanceInfo, RemoveTypeKill)
```
jakubno: now it actually kills the sandbox

packages/api/internal/orchestrator/cache.go
```diff
@@ -227,75 +218,21 @@ func (o *Orchestrator) getDeleteInstanceFunction(
...
-		if info.AutoPause.Load() {
-			ct = ClosePause
-		} else {
-			ct = CloseDelete
+		removeType := RemoveTypeKill
+		if info.AutoPause {
+			removeType = RemoveTypePause
 		}
 
-		// Run in separate goroutine to not block sandbox deletion
```
jakubno, line 242: moved to `RemoveInstance` method

packages/api/internal/orchestrator/cache.go
```diff
@@ -227,75 +218,21 @@ func (o *Orchestrator) getDeleteInstanceFunction(
 		sbxlogger.I(info).Debug("Deleting sandbox from cache hook",
 			zap.Time("start_time", info.StartTime),
 			zap.Time("end_time", info.GetEndTime()),
-			zap.Bool("auto_pause", info.AutoPause.Load()),
 		)
 
-		defer o.instanceCache.UnmarkAsPausing(info)
```
jakubno, line 237: moved to the actual pausing, this was called even when there was no pausing

packages/api/internal/handlers/sandbox_resume.go
```diff
@@ -60,15 +60,23 @@ func (a *APIStore) PostSandboxesSandboxIDResume(c *gin.Context, sandboxID api.Sa
...
 	sbxCache, err := a.orchestrator.GetSandbox(sandboxID)
 	if err == nil {
-		zap.L().Debug("Sandbox is already running",
-			logger.WithSandboxID(sandboxID),
-			zap.Time("end_time", sbxCache.GetEndTime()),
-			zap.Time("start_time", sbxCache.StartTime),
-			zap.String("node_id", sbxCache.NodeID),
-		)
-		a.sendAPIStoreError(c, http.StatusConflict, fmt.Sprintf("Sandbox %s is already running", sandboxID))
+		if sbxCache.PauseStarted() {
```
jakubno: changed the logic little bit, now we wait for pause sooner, that way we can get the correct data ( the outdated if pause + resume at least once before) and not return any error.

packages/api/internal/handlers/sandbox_kill.go
```diff
@@ -84,6 +85,14 @@ func (a *APIStore) DeleteSandboxesSandboxID(
 
 	sbx, err := a.orchestrator.GetSandbox(sandboxID)
 	if err == nil {
+		if sbx.PauseStarted() {
```
jakubno: prevent race condition where snapshot could be persisted even after calling `kill`

packages/api/internal/handlers/sandbox_kill.go
```diff
@@ -93,16 +102,15 @@ func (a *APIStore) DeleteSandboxesSandboxID(
 		}
 
 		// remove running sandbox from the orchestrator
-		sandboxExists := a.orchestrator.DeleteInstance(ctx, sandboxID, false)
-		if !sandboxExists {
-			telemetry.ReportError(ctx, "sandbox not found", fmt.Errorf("sandbox '%s' not found", sandboxID), telemetry.WithSandboxID(sandboxID))
-			a.sendAPIStoreError(c, http.StatusNotFound, fmt.Sprintf("Error deleting sandbox - sandbox '%s' was not found", sandboxID))
-
+		err := a.orchestrator.RemoveInstance(ctx, sbx, orchestrator.RemoveTypeKill)
```
jakubno: this is actually killing the sandbox now (not only marking it for eviction)

packages/api/internal/handlers/sandbox_get.go
```diff
@@ -51,6 +51,11 @@ func (a *APIStore) GetSandboxesSandboxID(c *gin.Context, id string) {
 			return
 		}
 
+		state := api.Running
+		if info.PauseStarted() {
+			state = api.Paused
```
jakubno, line 59: if pausing already started, it will behave as paused sandbox - you can call connect and the sandbox will be resumed

packages/api/internal/cache/instance/sync.go
```diff
@@ -41,7 +41,7 @@ func (c *InstanceCache) KeepAliveFor(instanceID string, duration time.Duration,
 	}
 
 	if (time.Since(instance.StartTime)) > instance.MaxInstanceLength {
-		c.cache.Remove(instanceID)
+		instance.SetExpired()
```
jakubno, line 44: this is the same behavior as before but explicit

packages/api/internal/cache/instance/sync.go
```diff
@@ -73,7 +73,7 @@ func (c *InstanceCache) Sync(ctx context.Context, instances []*InstanceInfo, nod
 		}
 		_, found := instanceMap[item.SandboxID]
 		if !found {
-			c.cache.Remove(item.SandboxID)
+			item.SetExpired()
```
jakubno, line 77: this is the same behavior as before but explicit

packages/api/internal/cache/instance/operations.go
```diff
@@ -41,6 +41,22 @@ func (c *InstanceCache) Get(instanceID string) (*InstanceInfo, error) {
 	return item, nil
 }
 
+func (c *InstanceCache) Len() int {
```
jakubno: moved from `instance.go`

packages/api/internal/cache/instance/lifecycle_cache_test.go
```diff
@@ -74,21 +74,6 @@ func TestLifecycleCacheGet(t *testing.T) {
 	assert.Equal(t, 1, cache.Len())
 }
 
-func TestLifecycleCacheGetAndRemove(t *testing.T) {
```
jakubno, line 79: the method is removed

packages/api/internal/cache/instance/instance_cache.go
```diff
@@ -0,0 +1,66 @@
...
+	cache          *lifecycleCache[*InstanceInfo]
+	insertInstance func(data *InstanceInfo, created bool) error
+
+	sandboxCounter metric.Int64UpDownCounter
+	createdCounter metric.Int64Counter
+
+	mu sync.Mutex
+}
+
+func NewCache(
```
jakubno: moved from `instance.go`

packages/api/internal/cache/instance/instance.go
```diff
@@ -109,8 +122,10 @@ type InstanceInfo struct {
 	AllowInternetAccess *bool
 	NodeID              string
 	ClusterID           uuid.UUID
-	AutoPause           atomic.Bool
-	Pausing             *utils.SetOnce[string]
+	AutoPause           bool
```
jakubno, line 115: this field was misused to not only track if the sandbox autoPause but also to track if the eviction should pause the sandbox or kill it

packages/api/internal/cache/instance/instance.go
```diff
@@ -109,8 +122,10 @@ type InstanceInfo struct {
...
+	pausingState        pausingState
+	onEviction          EvictionType
+	pausing             *utils.SetOnce[struct{}]
```
jakubno: no need to store the string anymore, we are saving it to the db

packages/api/internal/cache/instance/instance.go
```diff
@@ -147,125 +148,44 @@ func (i *InstanceInfo) SetExpired() {
...
-	createdCounter metric.Int64Counter
+func (i *InstanceInfo) PauseStarted() bool {
+	i.mu.RLock()
+	defer i.mu.RUnlock()
 
-	mu sync.Mutex
+	return i.pausingState != pausingNotStarted
 }
 
-func NewCache(
```
jakubno, line 187: moved to separate file

packages/api/internal/cache/instance/instance.go
```diff
@@ -147,125 +148,44 @@ func (i *InstanceInfo) SetExpired() {
...
-				zap.L().Error("error inserting instance", zap.Error(err))
-			}
-		}()
+func (i *InstanceInfo) WaitForPausing(ctx context.Context) error {
+	if !i.PauseStarted() {
+		return fmt.Errorf("sandbox isn't pausing")
 	}
-}
 
-func (c *InstanceCache) MarkAsPausing(instanceInfo *InstanceInfo) {
```
jakubno, line 247: refactored this, now it all lives on the `InstanceInfo` and was basically moved into one boolean

packages/api/internal/cache/instance/lifecycle_cache.go
```diff
@@ -53,33 +53,26 @@ func (c *lifecycleCache[T]) Start(ctx context.Context) {
...
-						go func() {
-							c.onEvict(ctx, value)
-							c.evicting.Remove(key)
-						}()
-
-						return true
-					})
-
-					c.metrics.Evictions += 1
+					absent := c.evicting.SetIfAbsent(key, struct{}{})
```
jakubno: prevent remove from 2 different places

packages/api/internal/cache/instance/lifecycle_cache.go
```diff
@@ -53,33 +53,28 @@ func (c *lifecycleCache[T]) Start(ctx context.Context) {
...
-						if !exist {
-							return false
-						}
-
-						c.evicting.Set(key, struct{}{})
+					absent := c.evicting.SetIfAbsent(key, struct{}{})
+					if absent {
 						go func() {
 							c.onEvict(ctx, value)
+							c.items.Remove(key)
```
jakubno: remove after evict, we want to prevent situation where the sandbox is temporarily "lost". It's removed from cache, but appears after the snapshot info is saved to db

packages/api/internal/orchestrator/delete_instance.go
```diff
@@ -2,20 +2,94 @@ package orchestrator
...
+
+	return o.removeInstance(ctx, sandbox, removeType)
+}
+
+// removeInstance should be called from places where you already marked the sandbox as being removed
+func (o *Orchestrator) removeInstance(ctx context.Context, sandbox *instance.InstanceInfo, removeType RemoveType) error {
+	duration := time.Since(sandbox.StartTime).Seconds()
+	stopTime := time.Now()
+
+	// Run in separate goroutine to not block sandbox deletion
```
jakubno: Moved analytics here as this is the single point to remove any sandbox

packages/api/internal/handlers/sandbox_pause.go
```diff
@@ -57,17 +58,21 @@ func (a *APIStore) PostSandboxesSandboxIDPause(c *gin.Context, sandboxID api.San
...
-		return
-	}
+	// If the sandbox is already being paused, we just wait for it
+	if sbx.PauseStarted() {
+		err = sbx.WaitForPausing(ctx)
+		if err != nil {
+			telemetry.ReportError(ctx, "error when waiting for sandbox to pause", err)
+		}
+	} else {
+		err = a.orchestrator.RemoveInstance(ctx, sbx, orchestrator.RemoveTypePause)
```
jakubno: this will pause the sandbox directly (without any wait for instance cache eviction)

packages/api/internal/orchestrator/pause_instance.go
```diff
@@ -22,16 +22,24 @@ func (ErrPauseQueueExhausted) Error() string {
 	return "The pause queue is exhausted"
 }
 
-func (o *Orchestrator) PauseInstance(
-	ctx context.Context,
-	sbx *instance.InstanceInfo,
-	teamID uuid.UUID,
-) error {
+func (o *Orchestrator) pauseSandbox(ctx context.Context, node *nodemanager.Node, sbx *instance.InstanceInfo) (err error) {
```
jakubno, line 25: the only change should be, that handling the pausing state is done in here instead of the cache hook

packages/api/internal/handlers/sandbox_resume.go
```diff
@@ -58,17 +58,31 @@ func (a *APIStore) PostSandboxesSandboxIDResume(c *gin.Context, sandboxID api.Sa
...
+			err = sbxCache.WaitForStop(ctx)
+			if err != nil {
+				a.sendAPIStoreError(c, http.StatusInternalServerError, "Error when resuming sandbox")
+				return
+			}
+		case instance.StateShuttingDown, instance.StateKilled:
+			a.sendAPIStoreError(c, http.StatusBadRequest, fmt.Sprintf("Sandbox %s is already killed", sandboxID))
+			return
+		case instance.StateRunning:
+			a.sendAPIStoreError(c, http.StatusConflict, fmt.Sprintf("Sandbox %s is already running", sandboxID))
```
jakubno, line 72: I think we could return success state, small issue is we are returning `201`, which would be little bit weird? Not sure about adding /changing to`200` as it could be breaking change. In our SDK, it's handled and it doesn't require 201, so both option are fine

packages/api/internal/cache/instance/lifecycle_cache.go
```diff
@@ -53,85 +53,65 @@ func (c *lifecycleCache[T]) Start(ctx context.Context) {
...
-
-	return v, true
+	return value, true
 }
 
-func (c *lifecycleCache[T]) Remove(key string) bool {
-	_, ok := c.GetAndRemove(key)
-	return ok
+func (c *lifecycleCache[T]) Remove(key string) {
+	c.items.Remove(key)
```
dobrac: this doesn't call the onEvict callback, might cause inconsistent behavior

packages/api/internal/cache/instance/instance_cache.go
```diff
@@ -0,0 +1,66 @@
...
+	deleteInstance func(data *InstanceInfo) error,
+) *InstanceCache {
+	// We will need to either use Redis or Consul's KV for storing active sandboxes to keep everything in sync,
+	// right now we load them from Orchestrator
+	cache := newLifecycleCache[*InstanceInfo]()
+
+	meter := meterProvider.Meter("api.cache.sandbox")
+	sandboxCounter, err := telemetry.GetUpDownCounter(meter, telemetry.SandboxCountMeterName)
+	if err != nil {
+		zap.L().Error("error getting counter", zap.Error(err))
```
dobrac: is this soft fail by design?

packages/api/internal/orchestrator/create_instance.go
```diff
@@ -235,9 +235,9 @@ func (o *Orchestrator) CreateSandbox(
...
+		if err != nil {
+			telemetry.ReportEvent(ctx, fmt.Sprintf("Remove sandbox error after cache error: %s", err))
```
dobrac: you can report error instead

packages/api/internal/cache/instance/instance.go
dobrac: just for a context, we've changed the behavior a bit. In the initial design, the kill should have paused the sandbox, not kill it

packages/api/internal/cache/instance/instance.go
```diff
@@ -24,7 +19,15 @@ const (
 	InstanceAutoPauseDefault = false
 )
 
-var ErrPausingInstanceNotFound = errors.New("pausing instance not found")
+type State string
+
+const (
+	StateRunning      State = "running"
+	StatePausing      State = "pausing"
+	StateShuttingDown State = "shutting down"
```
dobrac:
```suggestion
	StateKilling State = "killing"
```

packages/api/internal/cache/instance/instance.go
```diff
@@ -147,125 +150,48 @@ func (i *InstanceInfo) SetExpired() {
...
-	c.pausing.RemoveCb(instanceInfo.SandboxID, func(key string, v *InstanceInfo, exists bool) bool {
-		if !exists {
-			return false
-		}
-
-		// Make sure it's the same instance and not a sandbox which has been already resumed
-		return v.ExecutionID == instanceInfo.ExecutionID
-	})
-}
+func (i *InstanceInfo) StopDone(err error, pause bool) {
```
dobrac: should we handle the pausing -> paused, killing -> killed a bit more safer - code wise? It would remove this boolean for a stop reason

packages/api/internal/cache/instance/lifecycle_cache.go
jakubno: that's on purpose, it's removed, not evicted

packages/api/internal/cache/instance/instance_cache.go
jakubno: I just copied it

packages/api/internal/cache/instance/lifecycle_cache.go
dobrac: then for example this `instanceCache.UpdateCounters(ctx, instanceInfo, -1, false)` will have invalid values

packages/api/internal/cache/instance/operations.go
```diff
@@ -98,18 +114,20 @@ func (c *InstanceCache) Add(ctx context.Context, instance *InstanceInfo, newlyCr
...
+// Remove the instance from the cache (no eviction callback).
+func (c *InstanceCache) Remove(instanceID string) {
+	defer c.cache.evicting.Remove(instanceID)
+	c.cache.Remove(instanceID)
+}
 
-		if pause {
-			c.MarkAsPausing(value)
-		}
+// StartRemoving marks the instance as being evicted to prevent
```
dobrac: to prevent what?

also, if you mark it as evicting like this, it will never finish evicting, right?

packages/api/internal/cache/instance/operations.go
```diff
@@ -98,18 +114,20 @@ func (c *InstanceCache) Add(ctx context.Context, instance *InstanceInfo, newlyCr
...
-// Delete the instance and remove it from the cache.
-func (c *InstanceCache) Delete(instanceID string, pause bool) bool {
-	value, found := c.cache.GetAndRemove(instanceID)
-	if found {
-		value.AutoPause.Store(pause)
+// Remove the instance from the cache (no eviction callback).
+func (c *InstanceCache) Remove(instanceID string) {
+	defer c.cache.evicting.Remove(instanceID)
+	c.cache.Remove(instanceID)
+}
```
dobrac: why do you need this behavior? It kinda goes against the callback meaning

packages/api/internal/handlers/sandbox_kill.go
dobrac: so the call is now synchronous (vs async before), right?

packages/api/internal/handlers/sandbox_resume.go
```diff
@@ -93,19 +107,6 @@ func (a *APIStore) PostSandboxesSandboxIDResume(c *gin.Context, sandboxID api.Sa
...
-		a.sendAPIStoreError(c, http.StatusInternalServerError, fmt.Sprintf("Error while pausing sandbox %s: %s", sandboxID, err))
-
-		return
-	}
-
-	if err == nil {
-		// If the pausing was in progress, prefer to restore on the node where the pausing happened.
-		nodeID = &pausedOnNode
-	}
-
```
dobrac, line 108: if the pausing is in progress now and you call resume, it won't await (or show you the pause error), but it will try to immediately resume -> and potentially fail?

dobrac (changes requested):

packages/api/internal/orchestrator/delete_instance.go
```diff
@@ -2,20 +2,99 @@ package orchestrator
...
-	zap.L().Debug("Delete instance from cache",
-		logger.WithSandboxID(sandboxID),
-		zap.Bool("pause", pause),
+	// Mark the sandbox as being removed to prevent it from being evicted by the cache or another call
+	err := o.instanceCache.StartRemoving(sandbox)
+	if err != nil {
+		return fmt.Errorf("failed to start removing sandbox '%s': %w", sandbox.SandboxID, err)
+	}
+
+	defer o.instanceCache.Remove(sandbox.SandboxID)
```
dobrac: i'm thinking if it wouldn't be better to separate the cache and the lifecycle handling from each other

packages/api/internal/handlers/sandbox_resume.go
jakubno: it will wait https://github.com/e2b-dev/infra/blob/6d32753f5a7aa6c1e9c0a650075b060caa685487/packages/api/internal/handlers/sandbox_resume.go#L65

packages/api/internal/orchestrator/delete_instance.go
jakubno: make sense, i will separate it

packages/api/internal/cache/instance/operations.go
jakubno: it will be removed manually

packages/api/internal/cache/instance/lifecycle_cache.go
jakubno: Handled little bit differently now

packages/api/internal/cache/instance/instance.go
jakubno: Removed completely the second argument

packages/api/internal/cache/instance/operations.go
jakubno: Separated eviction and the store

jakubno:
> ### Bug: Goroutine Closure Captures Last Loop Variable
> The goroutine launched within the `for` loop captures the `item` variable by reference. This classic Go closure bug means all goroutines will operate on the last `item` value from the loop, causing only one sandbox to be processed multiple times and leaving others unevicted.
> 
> `packages/api/internal/orchestrator/evictor/evict.go#L30-L45`
> [Fix in Cursor](https://cursor.com/open?data=eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6ImJ1Z2JvdC12MSJ9.eyJ2ZXJzaW9uIjoxLCJ0eXBlIjoiQlVHQk9UX0ZJWF9JTl9DVVJTT1IiLCJkYXRhIjp7InJlZGlzS2V5IjoiYnVnYm90OjVkMjFkOTc2LTNmN2UtNDRhMi05ZTU4LTg3NTJjNjQ1ZjIyYiIsImVuY3J5cHRpb25LZXkiOiJlV0trQ3hGWWx3YkxyZzd1VVlRajVfQXVGc3BORjR0OGVSLWIwRmVLT3JVIiwiYnJhbmNoIjoicmVmYWN0b3ItcGF1c2VkZWxldGUtZml4LWF1dG9wYXVzZS1idWctZTJiLTI5NTQifSwiaWF0IjoxNzU3MTYzMzgyLCJleHAiOjE3NTc3NjgxODJ9.C32lbwgqjAhuNjKjPDdTwvrm9uBqfIbUtLyAL9d8p0vT2IpzROky9UAbPtUd5cpjkg0NXwo2oNbKGVD6z-3L5hOre_QzJY9EKbr7Bnd5JobWWyo1U5J7T6iQwdUb0Bx2wiDhHI5pRikqcU97ErX8YBT-eOB6EhBlTm_skHlaHbtRvckgcw_CHzsPnFnem2XlF_CrGpMNmiyH_uyh-Lp8jhQWQj4i3kuuCRzZpt0SfYXONgXklGSOSkGYc69jVsaAXgbUPdUSQo3vQViKw6kuOcXb4fav7W4vAS1kA4PYIgvo1b7GhuvC-V_PX8phs0dhOiCWTme-o9gbIK7GS_8Kow) [Fix in Web](https://cursor.com/agents?data=eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6ImJ1Z2JvdC12MSJ9.eyJ2ZXJzaW9uIjoxLCJ0eXBlIjoiQlVHQk9UX0ZJWF9JTl9XRUIiLCJkYXRhIjp7InJlZGlzS2V5IjoiYnVnYm90OjVkMjFkOTc2LTNmN2UtNDRhMi05ZTU4LTg3NTJjNjQ1ZjIyYiIsImVuY3J5cHRpb25LZXkiOiJlV0trQ3hGWWx3YkxyZzd1VVlRajVfQXVGc3BORjR0OGVSLWIwRmVLT3JVIiwiYnJhbmNoIjoicmVmYWN0b3ItcGF1c2VkZWxldGUtZml4LWF1dG9wYXVzZS1idWctZTJiLTI5NTQiLCJyZXBvT3duZXIiOiJlMmItZGV2IiwicmVwb05hbWUiOiJpbmZyYSIsInByTnVtYmVyIjoxMDY0LCJjb21taXRTaGEiOiI3Njk1YTM5OTg4YTEwNGU3N2U4NjNiOWE3NWFhOTA1MTVmMmU4ZmI2In0sImlhdCI6MTc1NzE2MzM4MiwiZXhwIjoxNzU3NzY4MTgyfQ.eHK9Ufzi-cmYrQhJBWtEgidLxyZmz5FFKlvX-XEU5HsQ5qEec7V4CqfLf-VUtMN0G6zQiD105jLg-EpUyqq8OQ6tsuI3SwnQdc8oeP9FlhS66P3ElsRk-PzBTJXgzeu2IemPJzJjd9zyA4kRC-yy7Mkpk_q4-PiWxCSUBRbRVlcXfzLuTLi5zh03qVNSji_LthXx-IGOw0aO2mfQt8MuA9eKeUcRgc6UaAxFXGh4TcGdASOs_3JpYBZbz0zOdXz5Smp8a5DMXng2FOeKzPeFJoMTp2cw1UIhP1ZclFRxX1MV-99BE7gY2XiTUMAmK7SfnWAoyxM9DxHWhzoExW8x6Q)

this is go 1.24, this isn't a case anymore

PR #1067, opened 2025-08-27
Enable linters: errname and errorlint

djeebus:
https://github.com/Antonboom/errname
https://github.com/polyfloyd/go-errorlint


packages/api/internal/utils/error.go
```diff
@@ -97,42 +97,34 @@ func MultiErrorHandler(me openapi3.MultiError) error {
 
 	// Recreate logic from oapi-codegen/gin-middleware to handle the error
 	// Source: https://github.com/oapi-codegen/gin-middleware/blob/main/oapi_validate.go
-	switch e := err.(type) {
-	case *openapi3filter.RequestError:
```
dobrac: this code is copied, so it shouldn't be probably reformatted if possible

djeebus: It already looks nothing like the [original](https://github.com/oapi-codegen/gin-middleware/blob/main/oapi_validate.go#L156-L169). Other than the `e := err.(type)`, which breaks for wrapped errors, what else changed?

dobrac: only the `case *openapi3filter.SecurityRequirementsError:` is extended of additional handling, everything else stayed the same (maybe we could move it to a separate function to make it more clear)

this gin-middleware is unfortunately really fragile, but extremely important, so I'm not sure how much we want to interfere here

The long term plan is to rewrite this, maybe replace the gin-middleware for more robust library serving more to our needs

djeebus:
@dobrac okay, I copied the upstream implementation into ours, then pulled out our customizations into their own function. Check it out once when you get a chance?

dobrac (approved):

PR #1068, opened 2025-08-28
Refactor to remove embedded context

djeebus:
This relies on context cancellation at a parent level to cancel the goroutine, which is simpler and relies on less setup

### DIFF
--- a/packages/orchestrator/internal/template/build/builder.go
+++ b/packages/orchestrator/internal/template/build/builder.go
@@ -125,10 +125,17 @@ func (b *Builder) Build(ctx context.Context, template storage.TemplateFiles, con
 
 	isV1Build := config.FromImage == "" && config.FromTemplate == nil
 
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
 	postProcessor := writer.NewPostProcessor(ctx, logsWriter, isV1Build)
-	go postProcessor.Start()
 	defer func() {
-		postProcessor.Stop(ctx, e)
+		if e != nil {
+			postProcessor.Error(fmt.Sprintf("Build failed: %v", e))
+		} else {
+			postProcessor.Info(fmt.Sprintf("Build finished, took %s",
+				time.Since(startTime).Truncate(time.Second).String()))
+		}
 	}()
 
 	postProcessor.Info(fmt.Sprintf("Building template %s/%s", config.TemplateID, template.BuildID))
@@ -137,6 +144,7 @@ func (b *Builder) Build(ctx context.Context, template storage.TemplateFiles, con
 		if e == nil {
 			return
 		}
+
 		// Remove build files if build fails
 		removeErr := b.templateStorage.DeleteObjectsWithPrefix(ctx, template.BuildID)
 		if removeErr != nil {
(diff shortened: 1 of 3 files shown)


dobrac (changes requested):
Two things:

- this implementation could add additional "..." after the last message
- there is now missing explicit Stop step and it's implicit with the context. But, the goroutine shouldn't run after the last status message. Using context to manage lifecycle is not a good pattern, because it might be used by cleanups or other pending tasks still relevant to the execution, but not to the lifecycle

djeebus:
> Two things:
> 
> * this implementation could add additional "..." after the last message
> * there is now missing explicit Stop step and it's implicit with the context. But, the goroutine shouldn't run after the last status message. Using context to manage lifecycle is not a good pattern, because it might be used by cleanups or other pending tasks still relevant to the execution, but not to the lifecycle

Sorted both out; mind taking another look?

packages/orchestrator/internal/template/build/writer/postprocessor.go
```diff
@@ -1,101 +1,48 @@
...
-		Logger:         writer,
-		ctx:            ctx,
-		errChan:        make(chan error, 1),
-		stopCh:         make(chan struct{}, 1),
-		tickerInterval: tickerInterval,
-		ticker:         time.NewTicker(tickerInterval),
-	}
+	go pp.run()
+
+	return zapcore.RegisterHooks(core, pp.hook), func() { pp.done <- struct{}{} }
```
dobrac: nit: for the `done` function (pp.done <- struct{}{}) it might be nice to make it idempotent

packages/orchestrator/internal/template/build/builder.go
```diff
@@ -125,18 +126,31 @@ func (b *Builder) Build(ctx context.Context, template storage.TemplateFiles, con
 
 	isV1Build := config.FromImage == "" && config.FromTemplate == nil
 
-	postProcessor := writer.NewPostProcessor(ctx, logsWriter, isV1Build)
-	go postProcessor.Start()
+	done := func() {}
+	if isV1Build {
+		logsCore, done = writer.NewPostProcessor(5*time.Second, logsCore)
```
dobrac: make the `5*time.Second` a constant

packages/orchestrator/internal/template/build/builder.go
```diff
@@ -125,18 +126,31 @@ func (b *Builder) Build(ctx context.Context, template storage.TemplateFiles, con
...
-		postProcessor.Stop(ctx, e)
+		done()
+
+		if e != nil {
+			logger.Error(fmt.Sprintf("Build failed: %v", e))
+		} else {
+			logger.Info(fmt.Sprintf("Build finished, took %s",
+				time.Since(startTime).Truncate(time.Second).String()))
+		}
 	}()
```
dobrac: How about doing this instead? It will colocate done to the V1Build

```
logger := zap.New(logsCore)
defer func() {
	if e != nil {
		logger.Error("Build failed", zap.Error(e))
	} else {
		logger.Info("Build finished",
			zap.String("duration", time.Since(startTime).Truncate(time.Second).String()))
	}
}()

if isV1Build {
	logsCore, done := writer.NewPostProcessor(5*time.Second, logsCore)
	defer done()
	logger = zap.New(logsCore)
}
```

dobrac (changes requested):
I like this! Nice work 💪

PR #1069, opened 2025-08-28
Remove to refactor embedded storage ctx

djeebus:
This gets rid of a few, but not all, uses of the embedded ctx in the storage providers

### DIFF
--- a/packages/orchestrator/internal/sandbox/block/chunk.go
+++ b/packages/orchestrator/internal/sandbox/block/chunk.go
@@ -19,7 +19,7 @@ import (
 type Chunker struct {
 	ctx context.Context // nolint:containedctx // todo: refactor so this can be removed
 
-	base    io.ReaderAt
+	base    storage.ReaderAtCtx
 	cache   *Cache
 	metrics metrics.Metrics
 
@@ -33,7 +33,7 @@ func NewChunker(
 	ctx context.Context,
 	size,
 	blockSize int64,
-	base io.ReaderAt,
+	base storage.ReaderAtCtx,
 	cachePath string,
 	metrics metrics.Metrics,
 ) (*Chunker, error) {
@@ -163,7 +163,7 @@ func (c *Chunker) fetchToCache(off, length int64) error {
 					c.metrics.TotalBytesRetrievedMetric,
 					c.metrics.TotalRemoteReadsMetric,
 				)
-				readBytes, err := c.base.ReadAt(b, fetchOff)
+				readBytes, err := c.base.ReadAt(c.ctx, b, fetchOff)
 				if err != nil && !errors.Is(err, io.EOF) {
 					fetchSW.End(c.ctx, int64(readBytes),
 						attribute.String(result, resultTypeFailure),
--- a/packages/shared/pkg/storage/storage_cache.go
+++ b/packages/shared/pkg/storage/storage_cache.go
@@ -11,6 +11,8 @@ import (
 	"time"
 
 	"github.com/google/uuid"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
 	"go.uber.org/zap"
 )
 
@@ -67,7 +69,11 @@ type CachedFileObjectProvider struct {
 var _ StorageObjectProvider = (*CachedFileObjectProvider)(nil)
 
 // WriteTo is used for very small files and we can check agains their size to ensure the content is valid.
-func (c *CachedFileObjectProvider) WriteTo(dst io.Writer) (int64, error) {
+func (c *CachedFileObjectProvider) WriteTo(ctx context.Context, dst io.Writer) (int64, error) {
+	var err error
+	ctx, span := tracer.Start(ctx, "CachedFileObjectProvider.WriteAt")
+	defer span.End()
+
 	totalSize, err := c.Size()
 	if err != nil {
 		return 0, err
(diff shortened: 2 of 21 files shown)


packages/orchestrator/internal/sandbox/block/chunk.go
```diff
@@ -163,7 +163,7 @@ func (c *Chunker) fetchToCache(off, length int64) error {
 					c.metrics.TotalBytesRetrievedMetric,
 					c.metrics.TotalRemoteReadsMetric,
 				)
-				readBytes, err := c.base.ReadAt(b, fetchOff)
+				readBytes, err := c.base.ReadAt(c.ctx, b, fetchOff)
```
djeebus, line 158: not proud of this one, but baby steps!

packages/shared/pkg/storage/storage_cache.go
```diff
@@ -86,17 +88,20 @@ func (c CachedProvider) GetDetails() string {
...
 }
 
 var _ StorageObjectProvider = (*CachedFileObjectProvider)(nil)
 
-// WriteTo is used for very small files, and we can check against their size to ensure the content is valid.
-func (c *CachedFileObjectProvider) WriteTo(dst io.Writer) (int64, error) {
-	totalSize, err := c.Size()
+// WriteTo is used for very small files and we can check agains their size to ensure the content is valid.
+func (c *CachedFileObjectProvider) WriteTo(ctx context.Context, dst io.Writer) (int64, error) {
+	var err error
```
dobrac: why the error definition here?

packages/shared/pkg/storage/storage_cache.go
```diff
@@ -138,22 +143,31 @@ func (c *CachedFileObjectProvider) WriteTo(dst io.Writer) (int64, error) {
 			zap.Int64("actual", bytesWritten))
 	}
 
-	go c.writeFullFileToCache(fullCachePath, writer.Bytes())
+	go func(ctx context.Context) {
+		c.writeFullFileToCache(ctx, fullCachePath, writer.Bytes())
+	}(context.WithoutCancel(ctx))
```
dobrac: doing it this way is imo more transparent
```suggestion
	go func() {
		c.writeFullFileToCache(context.WithoutCancel(ctx), fullCachePath, writer.Bytes())
	}()
```

packages/shared/pkg/storage/storage_cache.go
```diff
@@ -138,22 +143,31 @@ func (c *CachedFileObjectProvider) WriteTo(dst io.Writer) (int64, error) {
...
-func (c *CachedFileObjectProvider) Write(src []byte) (int, error) {
-	return c.inner.Write(src)
+func (c *CachedFileObjectProvider) Write(ctx context.Context, src []byte) (int, error) {
+	return c.inner.Write(ctx, src)
 }
 
-func (c *CachedFileObjectProvider) ReadAt(buff []byte, offset int64) (int, error) {
-	if err := c.validateReadAtParams(int64(len(buff)), offset); err != nil {
+func (c *CachedFileObjectProvider) ReadAt(ctx context.Context, buff []byte, offset int64) (int, error) {
+	var err error
```
dobrac: same here

packages/shared/pkg/storage/storage_cache.go
```diff
@@ -173,12 +187,14 @@ func (c *CachedFileObjectProvider) ReadAt(buff []byte, offset int64) (int, error
...
-	readCount, err := c.inner.ReadAt(buff, offset)
+	readCount, err := c.inner.ReadAt(ctx, buff, offset)
 	if err != nil {
 		return 0, fmt.Errorf("failed to perform uncached read: %w", err)
 	}
 
-	go c.writeChunkToCache(offset, chunkPath, buff[:readCount])
+	go func(ctx context.Context) {
+		c.writeChunkToCache(ctx, offset, chunkPath, buff[:readCount])
+	}(context.WithoutCancel(ctx))
```
dobrac: same here

Makefile
```diff
@@ -279,6 +279,10 @@ lint:
 	@./scripts/golangci-lint-install.sh "2.1.6"
 	go work edit -json | jq -r '.Use[].DiskPath'  | xargs -I{} golangci-lint run {}/... --fix
 
+.PHONY: mocks
+mocks:
```
dobrac: maybe
```suggestion
.PHONY: generate-mocks
generate-mocks:
```

packages/shared/pkg/storage/storage_cache.go
djeebus: Oops, that was left over from a previous implementation, removing it.

packages/shared/pkg/storage/storage_cache.go
djeebus: I got used to writing it this way for a few reasons:
1. _all_ code in this goroutine uses the uncancellable context
2. the `context.WithoutCancel` call doesn't get hidden by being shifted to the right
3.  the function conforms to expected golang patterns: the first argument is `ctx context.Context`, which communicates to developers that this goroutine uses a ctx.
It's not important though, I'll swap to your implementation so we can get this merged. Happy to talk about it offline if you like!

dobrac: I understand the pov, I would still prefer passing it like this explicitly. We've already encountered one other place where it was confusing, as when you read the code, you need to first go at the bottom, and then at the top of the function

dobrac (approved):

PR #1072, opened 2025-08-28
Nodes sync race condition

sitole:
- (2b9091d9bb99d9cc6d8a0744b2688742bc705883) Fix issue when a long blocked run of `PoolExists/PoolUpdate` can cause a timeout for the whole sync run. This issue occurred when the orchestrator failed and the sync was blocked due to a timeout in the gRPC info service of one node, while the other node was simply hanging. 
- (c8540f72a74a1b14a078a12bb906dae089586301) Fixing race condition when orchestrator on same node restarter faster than next round of service discovery finishes, so api cluster sync pool never realized node disappeared and node de-registration was never triggered, leaving existing node with matching node id but outdated service instance id.

### DIFF
--- a/packages/shared/pkg/synchronization/synchronization.go
+++ b/packages/shared/pkg/synchronization/synchronization.go
@@ -104,15 +104,16 @@ func (s *Synchronize[SourceItem, PoolItem]) syncDiscovered(ctx context.Context, 
 	defer wg.Wait()
 
 	for _, item := range sourceItems {
-		// item already exists in the pool, skip it
-		if ok := s.store.PoolExists(ctx, item); ok {
-			continue
-		}
-
 		// initialize newly discovered item
 		wg.Add(1)
 		go func(item SourceItem) {
 			defer wg.Done()
+
+			// item already exists in the pool, skip it
+			if ok := s.store.PoolExists(ctx, item); ok {
+				return
+			}
+
 			s.store.PoolInsert(spanCtx, item)
 		}(item)
 	}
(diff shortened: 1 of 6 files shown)


packages/shared/pkg/synchronization/synchronization.go
```diff
@@ -104,15 +104,16 @@ func (s *Synchronize[SourceItem, PoolItem]) syncDiscovered(ctx context.Context,
...
-			continue
-		}
-
 		// initialize newly discovered item
 		wg.Add(1)
 		go func(item SourceItem) {
 			defer wg.Done()
+
+			// item already exists in the pool, skip it
+			if ok := s.store.PoolExists(ctx, item); ok {
```
dobrac, line 113: this needs to be now thread safe (just make sure it is)

packages/shared/pkg/synchronization/synchronization.go
```diff
@@ -126,16 +127,17 @@ func (s *Synchronize[SourceItem, PoolItem]) syncOutdated(ctx context.Context, so
...
-			s.store.PoolUpdate(ctx, poolItem)
-			continue
-		}
-
 		// remove the item that is no longer present in the source
 		wg.Add(1)
 		go func(poolItem PoolItem) {
 			defer wg.Done()
+
+			found := s.store.SourceExists(ctx, sourceItems, poolItem)
```
dobrac, line 135: this needs to be now thread safe (just make sure it is)

packages/shared/pkg/synchronization/synchronization.go
```diff
@@ -126,16 +127,17 @@ func (s *Synchronize[SourceItem, PoolItem]) syncOutdated(ctx context.Context, so
...
+			if found {
+				s.store.PoolUpdate(ctx, poolItem)
```
dobrac, line 137: this needs to be now thread safe (just make sure it is)

PR #1077, opened 2025-08-28
Add feature flags for placing algorithm

jakubno:
Add feature flags to choose and customize best of K placing algorithm

### DIFF
--- a/packages/api/internal/orchestrator/placement/placement.go
+++ b/packages/api/internal/orchestrator/placement/placement.go
@@ -33,6 +33,11 @@ func PlaceSandbox(ctx context.Context, tracer trace.Tracer, algorithm Algorithm,
 	nodesExcluded := make(map[string]struct{})
 	var err error
 
+	var node *nodemanager.Node
+	if preferredNode != nil {
+		node = preferredNode
+	}
+
 	attempt := 0
 	for attempt < maxRetries {
 		select {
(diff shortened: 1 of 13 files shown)


packages/api/internal/orchestrator/placement/placement_best_of_K.go
```diff
@@ -89,20 +89,27 @@ func (b *BestOfK) getConfig() BestOfKConfig {
...
+func (b *BestOfK) UpdateConfig(config BestOfKConfig) {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	b.config = config
+}
+
 // chooseNode selects the best node for placing a VM with the given quota
 func (b *BestOfK) chooseNode(_ context.Context, nodes []*nodemanager.Node, excludedNodes map[string]struct{}, resources nodemanager.SandboxResources) (bestNode *nodemanager.Node, err error) {
 	// Fix the config, we want to dynamically update it
 	config := b.getConfig()
```
jakubno, line 119: fix the config as it can change dynamically

packages/shared/pkg/feature-flags/client.go
jakubno, line 1: rewrite to same format as bools

packages/api/internal/orchestrator/placement/placement.go
```diff
@@ -32,6 +33,11 @@ func PlaceSandbox(ctx context.Context, tracer trace.Tracer, algorithm Algorithm,
 	nodesExcluded := make(map[string]struct{})
 	var err error
 
+	var node *nodemanager.Node
```
jakubno, line 36: there was a bug, if preferred node was passed, it was always used

packages/api/internal/orchestrator/placement/placement.go
```diff
@@ -71,9 +75,13 @@ func PlaceSandbox(ctx context.Context, tracer trace.Tracer, algorithm Algorithm,
 		err = node.SandboxCreate(ctx, sbxRequest)
 		span.End()
 		if err != nil {
+			if algorithm.excludeNode(err) {
```
jakubno, line 78: for best of K we want to remove the check for in progress sandboxes from API, this will also help with tests / small number of nodes and bursts

ValentaTomas (approved):

PR #1080, opened 2025-08-29
Add service instance id to `/nodes` endpoints

jakubno:
Add service instance id to `/nodes` endpoint


packages/api/internal/orchestrator/admin.go
```diff
@@ -58,18 +61,23 @@ func (o *Orchestrator) AdminNodes() []*api.Node {
 	return result
 }
 
-func (o *Orchestrator) AdminNodeDetail(nomadNodeShortID string) (*api.NodeDetail, error) {
-	n := o.GetNodeByNomadShortID(nomadNodeShortID)
+func (o *Orchestrator) AdminNodeDetail(clusterID uuid.UUID, nodeID string) (*api.NodeDetail, error) {
```
dobrac:
```suggestion
func (o *Orchestrator) AdminNodeDetail(clusterID uuid.UUID, nodeIDOrNomadNodeShortID string) (*api.NodeDetail, error) {
```

PR #1082, opened 2025-08-29
Fixed issue when comparing empty metadata jsonb with NULL column

sitole:


packages/db/queries/get_snapshots_with_cursor.sql
```diff
@@ -18,7 +18,11 @@ JOIN LATERAL (
 ) eb ON TRUE
 WHERE
     e.team_id = @team_id
-    AND s.metadata @> @metadata
+    AND (
+        -- When metadata arg is empty json, accept all as row metadata column can be empty json or NULL
+        -- And NULL does not match with empty json
+        @metadata = '{}'::jsonb OR s.metadata @> @metadata
+    )
```
sitole, line 25: We want to filter by metadata. When filter metadata is not set, skip metadata filtration so we will match snapshots with NULL instead of `{}`.

PR #1084, opened 2025-08-29
GCP upload using multipart

ValentaTomas:
GCP files upload in parallel using multipart instead of composite and gcloud Python cli.

More info on multipart vs composite uploads: https://www.beginswithdata.com/2024/02/01/google-cloud-storage-max-throughput/

### DIFF
--- a/packages/shared/pkg/storage/gcp_multipart.go
+++ b/packages/shared/pkg/storage/gcp_multipart.go
@@ -9,12 +9,14 @@ import (
 	"fmt"
 	"io"
 	"math"
+	"math/rand"
 	"net/http"
 	"os"
 	"sort"
 	"sync"
 	"time"
 
+	"github.com/hashicorp/go-retryablehttp"
 	"go.uber.org/zap"
 	"golang.org/x/oauth2/google"
 	"golang.org/x/sync/errgroup"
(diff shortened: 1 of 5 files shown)


packages/shared/pkg/storage/storage_google.go
```diff
@@ -217,61 +215,21 @@ func (g *GCPBucketStorageObjectProvider) WriteTo(dst io.Writer) (int64, error) {
 }
 
 func (g *GCPBucketStorageObjectProvider) WriteFromFileSystem(path string) error {
-	upload := func() error {
-		extraArgs := []string{}
-		if g.limiter != nil {
-			uploadLimiter := g.limiter.GCloudUploadLimiter()
```
k21, line 223: What was the context for using the upload limiter? I am asking since it has been removed and it doesn't seem to be used in the new code. If we were hitting some rate limits, it might get worse with parallel uploads.

packages/shared/pkg/storage/storage_google.go
```diff
@@ -217,61 +215,21 @@ func (g *GCPBucketStorageObjectProvider) WriteTo(dst io.Writer) (int64, error) {
...
-
-			continue
-		}
-
-		// Files was successfully uploaded
-		return nil
+	maxConcurrency := 10
+	if err := uploader.UploadFileInParallel(filePath, maxConcurrency); err != nil {
+		return fmt.Errorf("failed to upload file in parallel: %w", err)
 	}
```
k21, line 243: If retries were necessary previously, any flakiness will be more visible with multiple requests per upload, so some kind of retry mechanism would probably still be useful (ideally at the chunk level). Unless the reason why the retries were needed was the large file size?

packages/shared/pkg/storage/gcp_multipart.go
```diff
@@ -0,0 +1,290 @@
...
+	// Create channels for work distribution and results
+	jobs := make(chan int, numParts)
+	results := make(chan UploadResult, numParts)
+
+	// Start worker goroutines
+	var wg sync.WaitGroup
+	for i := 0; i < maxConcurrency; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
```
k21: Nit: `wg.Go` might be slightly more concise/less error-prone.

k21 (changes requested):
Did we see a significant improvement with the change?

PR #1086, opened 2025-08-29
Remove zerolog from orchestrator

djeebus:


packages/orchestrator/cmd/build-template/main.go
```diff
@@ -50,7 +49,7 @@ func main() {
 
 	err := buildTemplate(ctx, *kernelVersion, *fcVersion, *templateID, *buildID)
 	if err != nil {
-		log.Fatal().Err(err).Msg("error building template")
+		zap.L().Fatal("error building template", zap.Error(err))
```
dobrac: you can just use "log" here, or move the zap initialization sooner

dobrac (changes requested):

packages/orchestrator/cmd/build-template/main.go
djeebus: `log` here used to refer to `github.com/rs/zerolog/log`, which is literally the only place in orchestrator where this project is called. Any reason why we want to include that reference for this one line of code?

dobrac: you can use golang "log" [https://pkg.go.dev/log](https://pkg.go.dev/log)

dobrac (approved):

PR #1090, opened 2025-08-31
Fix sandbox cleanup for template build

dobrac:
Fix sandbox cleanup for template build


packages/orchestrator/internal/sandbox/uffd/handler.go
```diff
@@ -165,8 +165,8 @@ func (u *Uffd) Ready() chan struct{} {
 	return u.readyCh
 }
 
-func (u *Uffd) Exit() chan error {
-	return u.exitCh
+func (u *Uffd) Exit() *utils.SetOnce[struct{}] {
```
ValentaTomas: Non-blocking, but we could probably just return nil/error here.

ValentaTomas: Disregard—this is used as "Wait", right?

dobrac: > Disregard—this is used as "Wait", right?

yes

ValentaTomas (commented):
Do you need to call stop before close?

dobrac:
> Do you need to call stop before close?

You don't need it now (because it's already handled for you), but you still can if you want to call it sooner

packages/shared/pkg/utils/set_once.go
```diff
@@ -44,6 +44,13 @@ func (s *SetOnce[T]) SetError(err error) error {
 	return s.setResult(result[T]{err: err})
 }
 
+func (s *SetOnce[T]) SetResult(value T, err error) error {
```
ValentaTomas, line 47: Might make sense to use the `SetOnce[error]` here as that is the only value we care about passing (and the `_, error` is then actual error of setting?), or refactor so it allows passing arbitrary errors, but for the purpose of the fix it does not matter.

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -304,7 +304,7 @@ func (s *server) Delete(ctxConn context.Context, in *orchestrator.SandboxDeleteR
 	// Start the cleanup in a goroutine—the initial kill request should be send as the first thing in stop, and at this point you cannot route to the sandbox anymore.
 	// We don't wait for the whole cleanup to finish here.
 	go func() {
-		err := sbx.Stop(ctx)
+		err := sbx.Close(ctx)
```
ValentaTomas: Should this not be `Stop`? The `Close` will be called after the wait on the line 136 returns.

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -372,7 +372,7 @@ func (s *server) Pause(ctx context.Context, in *orchestrator.SandboxPauseRequest
 			ctx, childSpan := s.tracer.Start(ctx, "sandbox-pause-stop")
 			defer childSpan.End()
 
-			err := sbx.Stop(ctx)
+			err := sbx.Close(ctx)
```
ValentaTomas: Maybe the same as the case in `Delete`.

packages/shared/pkg/utils/set_once.go
dobrac: I don't follow here, can you show a code example please?

packages/orchestrator/internal/server/sandboxes.go
dobrac: I think it can be Stop after the refactor now

dobrac: but it doesn't matter that much, it will have the same effect. The Stop will just rely on the goroutine in Create, the Close is self-sustained

dobrac: which one do you prefer?

packages/shared/pkg/utils/set_once.go
ValentaTomas: Something like

```go
	// ...
	setOnce := NewSetOnce[error]()

	// u.exit.SetValue(errors.Join(handleErr, closeErr, fdExitErr))

	exitError, err := setOnce.Wait()

	// ...
```

or 

```go
// Create the setOnce by default with the not set error
func NewSetOnce[T any]() *SetOnce[T] {
	done := make(chan struct{})

	return &SetOnce[T]{
		Done: done,
		setDone: sync.OnceFunc(func() {
			close(done)
		}),
		res: &result[T]{err: ErrNotSet{}}, // <----------- not 100% sure if we don't need to init the `var zero T` somehow, but the empty value might work here properly
	}
}

// Keep the setters the same, as they overwrite the whole result
func (s *SetOnce[T]) SetValue(value T) error {
	return s.setResult(result[T]{value: value})
}

func (s *SetOnce[T]) SetError(err error) error {
	return s.setResult(result[T]{err: err})
}

// Simplify return result as now it either has value, default error value or new (might be nil) error value we set.
func (s *SetOnce[T]) Result() (T, error) {
	s.mu.RLock()
	defer s.mu.RUnlock()

	return s.res.value, s.res.err
}

// ...
// Use it as `u.exit.SetError(errors.Join(handleErr, closeErr, fdExitErr)` in your code

```

dobrac: yes, I can do that, will refactor then

ValentaTomas: On the other hand, the new `SetResult` method also makes sense, as the `value, error` pair can be in a lot of situations passed without knowing if the values are errorous or not.

I would just probably add the default `res: &result[T]{err: ErrNotSet{}}` either way.

dobrac: but you can't update the result, so if I add the default, I'll need to check the non-set state against the default

dobrac: btw, we're using the `NewSetOnce[struct{}]` in other place alredy

ValentaTomas: > but you can't update the result, so if I add the default

Why? Maybe I missed something.

dobrac: That's the way how the SetOnce is implemented, you can only set the value/error once

dobrac:
```
	// not set yet, so try to set it
		s.mu.Lock()
		defer s.mu.Unlock()

		if s.res != nil {
			return fmt.Errorf("value already set")
		}

		s.res = &r

		return nil
```

ValentaTomas: Ugh, don't like this at all now :D

dobrac: refactored to a wrapper ErrorOnce

packages/orchestrator/internal/server/sandboxes.go
ValentaTomas: Probably the Stop to keep (on this level) the `Wait`->`Close` in a single place.

dobrac: Updated

ValentaTomas (approved):

PR #1093, opened 2025-09-01
Set stdin to `/dev/null` for non pty command by default

jakubno:
This fixes issues the issue for claude code CLI tool

### DIFF
--- a/packages/envd/internal/services/spec/process/process.pb.go
+++ b/packages/envd/internal/services/spec/process/process.pb.go
@@ -344,7 +344,7 @@ type StartRequest struct {
 	Pty     *PTY           `protobuf:"bytes,2,opt,name=pty,proto3,oneof" json:"pty,omitempty"`
 	Tag     *string        `protobuf:"bytes,3,opt,name=tag,proto3,oneof" json:"tag,omitempty"`
 	// This is optional for backwards compatibility.
-	// We will default to false. Also new SDK versions will set this to false by default.
+	// We default to true. New SDK versions will set this to false by default.
 	Stdin *bool `protobuf:"varint,4,opt,name=stdin,proto3,oneof" json:"stdin,omitempty"`
 }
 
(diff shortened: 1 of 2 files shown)


packages/envd/internal/services/process/handler/handler.go
```diff
@@ -271,12 +271,16 @@ func New(
 			}
 		}()
 
-		stdin, err := cmd.StdinPipe()
-		if err != nil {
-			return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("error creating stdin pipe for command '%s': %w", cmd, err))
-		}
+		// For backwards compatibility we still set the stdin if not explicitly disabled
```
jakubno, line 274: in sdk we will explicitly set `false`

tests/integration/internal/utils/process.go
```diff
@@ -33,13 +33,16 @@ func ExecCommandAsRoot(tb testing.TB, ctx context.Context, sbx *api.Sandbox, env
...
 	tb.Helper()
 
+	f := false
 	req := connect.NewRequest(&process.StartRequest{
 		Process: &process.ProcessConfig{
 			Cmd:  command,
 			Args: args,
 			Cwd:  cwd,
 		},
+		Stdin: &f,
```
jakubno, line 43: this is the default SDK will use

mishushakov:
Please also make sure to test with older versions of the SDK.
Shouldn't be causing any protobuf issues as I am seeing that only request definition was changed, but still good to be sure 👍

packages/envd/spec/process/process.proto
```diff
@@ -49,6 +49,9 @@ message StartRequest {
     ProcessConfig process = 1;
     optional PTY pty = 2;
     optional string tag = 3;
+    // This is optional for backwards compatibility.
+    // We will default to false. Also new SDK versions will set this to false by default.
```
dobrac:
```suggestion
    // We default to true. New SDK versions will set this to false by default.
```

dobrac (changes requested):
small nit

dobrac (approved):

PR #1098, opened 2025-09-01
Sqlc is not able to get arg type when we are fist filteriing by empty jsonb object

sitole:


packages/db/queries/get_snapshots_with_cursor.sql
```diff
@@ -21,7 +21,7 @@ WHERE
     AND (
         -- When metadata arg is empty json, accept all as row metadata column can be empty json or NULL
         -- And NULL does not match with empty json
-        @metadata = '{}'::jsonb OR s.metadata @> @metadata
+        s.metadata @> @metadata OR @metadata = '{}'::jsonb
```
sitole, line 24: When using `@metadata = '{}'::jsonb OR s.metadata @> @metadata` sqlc is not able to properly resolve input type and will change sqlc jsonb struct to interface that is later not properly parsed, and the query fails.

jakubno (approved):

PR #1103, opened 2025-09-02
Fix panic when build not found because concurrent access

sitole:
The previous query was not run in a transaction, and when some builds were deleted/state changes were made later, filtering failed when accessing the array without builds.

Decided to use sqlc, so everything is now sql-native and the query runs in one transaction. Also, removing the additional part that was still using the deprecated ORM for accessing the database.


packages/db/queries/get_team_templates.sql
```diff
@@ -0,0 +1,28 @@
+-- name: GetTeamTemplates :many
+SELECT sqlc.embed(e),
+       eb.id as build_id, eb.vcpu as build_vcpu, eb.ram_mb as build_ram_mb, eb.total_disk_size_mb as build_total_disk_size_mb, eb.envd_version as build_envd_version, eb.firecracker_version as build_firecracker_version,
```
sitole, line 3: You cannot use `sqlc.embed(eb)`, because eb comes from `LEFT JOIN LATERAL` and sqlc cannot extract types from it natively.

dobrac:
can we still have the option to do migrate/down? I use it quite extensively for development

packages/api/internal/handlers/sandbox_create.go
```diff
@@ -232,12 +230,9 @@ func setTemplateNameMetric(c *gin.Context, aliases []string) {
...
-	if aliases == nil {
-		return ""
-	}
-	if len(*aliases) == 0 {
+func firstAlias(aliases []string) string {
+	if len(aliases) == 0 {
 		return ""
 	}
-	return (*aliases)[0]
+	return (aliases)[0]
```
jakubno:
```suggestion
	return aliases[0]
```

jakubno (approved):

PR #1106, opened 2025-09-02
Add multiple template managers in Nomad job

dobrac:
Add multiple template managers in the Nomad job. This is achieved by changing the job type from `service` to `system`. Unfortunately, service doesn't support the update strategy we need (first stop the current job, then run the new one, one by one across nodes).

### DIFF
--- a/packages/nomad/template-manager.hcl
+++ b/packages/nomad/template-manager.hcl
@@ -1,17 +1,20 @@
-job "template-manager" {
+job "template-manager-system" {
   datacenters = ["${gcp_zone}"]
+  type = "system"
   node_pool  = "build"
   priority = 70
 
 %{ if update_stanza }
   update {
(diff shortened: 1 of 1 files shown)


packages/nomad/template-manager.hcl
```diff
@@ -1,17 +1,20 @@
-job "template-manager" {
+job "template-manager-system" {
   datacenters = ["${gcp_zone}"]
+  type = "system"
   node_pool  = "build"
   priority = 70
 
 %{ if update_stanza }
   update {
```
jakubno, line 8: how does update work for system jobs?

dobrac: it should first stop the job on the node, wait for exit, start it again. It should do it one at the time (max_parallel=1)

Template manager is updatable compared to orchestrators, so that should be fine.

PR #1107, opened 2025-09-02
Networking prerequisites for sandbox special events endpoint

0div:
- [x] introduce `consts.go` file in orchestrator and add the sandbox event IP 
- [x] set event server addfress via `SANDBOX_EVENT_IP` env var and passed via MMDS to the sbx
- [x] add post-routing rule to catch requests to events domain

### DIFF
--- a/packages/envd/internal/api/init.go
+++ b/packages/envd/internal/api/init.go
@@ -52,7 +52,7 @@ func (a *API) PostInit(w http.ResponseWriter, r *http.Request) {
 		}
 
 		if initRequest.HyperloopIP != nil {
-			go a.SetupHyperloop(*initRequest.HyperloopIP)
+			a.SetupHyperloop(*initRequest.HyperloopIP)
 		}
 	}
 


packages/orchestrator/internal/consts.go
```diff
@@ -0,0 +1,15 @@
+package internal
+
+import (
+	"github.com/e2b-dev/infra/packages/shared/pkg/env"
+)
+
+const (
+	// Private IP don't leave the sandbox through the network bridge, so we use a reserved IP address for it:
+	// See TEST-NET-3 on https://en.wikipedia.org/wiki/Reserved_IP_addresses
+	defaultSandboxEventIP = "203.0.113.0"
```
sitole: Can we use IP from local link range? Same range used by metadata services in Firecracker/EC2. I think they are fitting better for what we are trying to do here.

https://en.m.wikipedia.org/wiki/Link-local_address

packages/orchestrator/internal/sandbox/network/network.go
```diff
@@ -214,6 +216,13 @@ func (s *Slot) CreateNetwork() error {
 		return fmt.Errorf("error creating postrouting rule: %w", err)
 	}
 
+	// Redirect traffic destined for event server
+	eventIP := internal.GetSandboxEventIP()
+	err = tables.Append("nat", "PREROUTING", "-i", s.VethName(), "-p", "tcp", "-d", eventIP, "--dport", "80", "-j", "REDIRECT", "--to-port", "5010")
```
sitole: Let’s make 5010 default constants and make it configurable optionally

sitole (changes requested):
Let’s name it somehow more general like hyperloop or something to allow re-use it for different stuff than just events. Using events.e2b.dev is okay as it’s just customer facing thing.

Let’s add some simple http server listener to orchestrator listening on port end serving for example sandbox ID back. Then integration tests for that so we know it works and will work in future.

packages/orchestrator/internal/template/build/core/rootfs/rootfs.go
```diff
@@ -207,7 +210,8 @@ ff00::	ip6-mcastprefix
 ff02::1	ip6-allnodes
 ff02::2	ip6-allrouters
 127.0.1.1	%s
-`, hostname)
+%s	%s
```
sitole: Are we certain we want to hardcode DNS and IP addresses in every template? I would prefer not to do.
cc @dobrac

dobrac: The only way this is ok is if these will never ever change, even with envd update

packages/orchestrator/internal/consts.go
0div: i'll try again but iirc it didn't work last time.

sitole: There is range that should not be used in local link address range. `The entire range may be used for this purpose, except for the first 256 and last 256 addresses (169.254.0.0/24 and 169.254.255.0/24)`.

Can you try with `169.254.1.1`?

sitole:
> ### Bug: NAT Rule Leak in Network Management
> The new iptables NAT PREROUTING rule, added in `CreateNetwork()` to redirect event proxy traffic, isn't removed in `RemoveNetwork()`. This leads to rule accumulation, causing resource leaks and potential networking issues.
> 
> `packages/orchestrator/internal/sandbox/network/network.go#L218-L228`
> https://github.com/e2b-dev/infra/blob/03daee432157d4345cb1fb6f8a6f5291de1bd9d5/packages/orchestrator/internal/sandbox/network/network.go#L218-L228
> 
> [Fix in Cursor](https://cursor.com/open?data=eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6ImJ1Z2JvdC12MSJ9.eyJ2ZXJzaW9uIjoxLCJ0eXBlIjoiQlVHQk9UX0ZJWF9JTl9DVVJTT1IiLCJkYXRhIjp7InJlZGlzS2V5IjoiYnVnYm90OjFhMzlhOTVhLWE3NzctNGI5Yi1iMDg4LTRjZDNlNDdjZGYzYiIsImVuY3J5cHRpb25LZXkiOiJEUXpWb3dYelRXSkMyMHp1QUVRX1c3d01oVXFBbGdrd2RycXVBbmZBX2JFIiwiYnJhbmNoIjoic2V0dXAtbmV0d29ya2luZy1pbi1zYW5kYm94LWZvci1ldmVudHMtZW5kcG9pbnQtZTJiLTI5NjEifSwiaWF0IjoxNzU3NDQ3NTgwLCJleHAiOjE3NTgwNTIzODB9.LkP4SmG3Q3gb-cKBIDrUuGknFxv8OjFlcs9NHE2ryTtb_RLeBd93oLppkOjD0sYhwzKwkt6bOmShh4VyMPW6EkNAOMe8MJNiU1PfYZ9Cgut1qy7p3W76L1H8UQ1vAnnNvT1XYkOi71UipZ7dhH0xj3k85fwlf9635orDUksr27Da3eSJsfImH86PyqtV7JIs9duY8kecYvEa2PugYL-UYGZlH-y5B1P_HftZxfRGgTy2aBh0pg6Ki1o_YbYxNocSaXvmC-fuAgQ0sJhiU8Eq2qaoFWzjL4dEN4kMZznDK9NBU0J1SvMATjSn6NpWWwfjenmicd0JITAOAu5KV239KA) [Fix in Web](https://cursor.com/agents?data=eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6ImJ1Z2JvdC12MSJ9.eyJ2ZXJzaW9uIjoxLCJ0eXBlIjoiQlVHQk9UX0ZJWF9JTl9XRUIiLCJkYXRhIjp7InJlZGlzS2V5IjoiYnVnYm90OjFhMzlhOTVhLWE3NzctNGI5Yi1iMDg4LTRjZDNlNDdjZGYzYiIsImVuY3J5cHRpb25LZXkiOiJEUXpWb3dYelRXSkMyMHp1QUVRX1c3d01oVXFBbGdrd2RycXVBbmZBX2JFIiwiYnJhbmNoIjoic2V0dXAtbmV0d29ya2luZy1pbi1zYW5kYm94LWZvci1ldmVudHMtZW5kcG9pbnQtZTJiLTI5NjEiLCJyZXBvT3duZXIiOiJlMmItZGV2IiwicmVwb05hbWUiOiJpbmZyYSIsInByTnVtYmVyIjoxMTA3LCJjb21taXRTaGEiOiIwM2RhZWU0MzIxNTdkNDM0NWNiMWZiNmY4YTZmNTI5MWRlMWJkOWQ1In0sImlhdCI6MTc1NzQ0NzU4MCwiZXhwIjoxNzU4MDUyMzgwfQ.gdgwK1L6pKG4TXy66Urj3Gn-8M9GtNxsQF58xvvFBVGXKBm89hhSlyvmdOasPzU7gZq4O5bSR0GFGewEh5oX1Dj430tX3-LeL864GKSgx0Ud__I5v_L00Ed6yrpLfHfkVD9p5Yyi5x4arqABl9_OdhNArR39rSpyMKK5oi7HQcY7nzUh9GOzsHyfPhdO2yZz8cIWql0yF1iGbdnpO4B24M4wNgTgJIOUdIdOgNniORnM5Rh6PrLdIAVOoyC5Itb4gXXZSKYeGOFP22IJWWibKnesADmVmd-gpZN4GN7ZvmEWz4ttbZI0F9hAp-crzReKp5zwcTiOgP3JJ047Rn9GSQ)

This looks like a relevant issue. You should delete that when removing the network.

sitole:
> Let’s name it something more general, like hyperloop or something, to allow re-use for different stuff than just events. Using events.e2b.dev is okay, as it’s a customer-facing feature.
> 
> Let’s add a simple HTTP server listener to the orchestrator listening on a port and serving, for example, the sandbox ID back. Then, let's do integration tests for that, so we know it works and will continue to work in the future.

@0div can you please respond/resolve comments in the previous review before we do another round?

packages/envd/internal/host/mmds.go
```diff
@@ -22,17 +22,19 @@ const (
...
 type MMDSOpts struct {
-	TraceID    string `json:"traceID"`
-	InstanceID string `json:"instanceID"`
-	EnvID      string `json:"envID"`
-	Address    string `json:"address"`
+	TraceID              string `json:"traceID"`
+	InstanceID           string `json:"instanceID"`
+	EnvID                string `json:"envID"`
+	LogsCollectorAddress string `json:"address"`
+	EventsAddress        string `json:"eventsAddress"`
```
sitole: Can `address` stay the same? Later, it should contain `http://169.254.1.1`, which points envd directly to the sandbox http server and can be reused for different stuff, so i would be pro not to name it `LogsCollectorAddress` and still stick with `Address`

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -450,6 +451,7 @@ func ResumeSandbox(
 			LogsCollectorAddress: fmt.Sprintf("http://%s", logsCollectorIP),
 			TraceId:              traceID,
 			TeamId:               runtime.TeamID,
+			EventsAddress:        internal.GetSandboxEventIP(),
```
sitole: Why is it called `eventsAddress` when it's IPv4 without any schema?

0div: Good point, i'll rename.

packages/envd/internal/events/events.go
```diff
@@ -0,0 +1,59 @@
+package events
```
sitole: Why is this file part of a "typo" commit (https://github.com/e2b-dev/infra/pull/1107/commits/fa6d9830d126264b983daa1af892b987b69b6a65)? I suppose it belongs to future implementation, revert it now.

packages/envd/internal/host/mmds.go
```diff
@@ -22,17 +22,19 @@ const (
...
-	Address    string `json:"address"`
+	TraceID      string `json:"traceID"`
+	InstanceID   string `json:"instanceID"`
+	EnvID        string `json:"envID"`
+	Address      string `json:"address"`
+	EventProxyIP string `json:"eventProxyIP"`
 }
 
-func (opts *MMDSOpts) Update(traceID, instanceID, envID, address string) {
+func (opts *MMDSOpts) Update(traceID, instanceID, envID, logsCollectorAddress, eventProxyIP string) {
```
sitole:
```suggestion
func (opts *MMDSOpts) Update(traceID, instanceID, envID, address, eventProxyIP string) {
```
for future us: let's rename it back to `address` when logs collector will live on sandbox http server

packages/envd/internal/host/mmds.go
```diff
@@ -22,17 +22,19 @@ const (
...
 type MMDSOpts struct {
-	TraceID    string `json:"traceID"`
-	InstanceID string `json:"instanceID"`
-	EnvID      string `json:"envID"`
-	Address    string `json:"address"`
+	TraceID      string `json:"traceID"`
+	InstanceID   string `json:"instanceID"`
+	EnvID        string `json:"envID"`
+	Address      string `json:"address"`
+	EventProxyIP string `json:"eventProxyIP"`
```
sitole:
```suggestion
	HyperloopIP string `json:"hyperloopIP"`
```

Ignore if you have a different approach, but it's just an IP address used for pointing to a sandbox server. We don't necessarily need to be stuck forever with `eventProxyIP` and can name it `hyperloopProxyIP` / `hyperloopIP`, so people can reuse it for different purposes.

packages/envd/internal/host/mmds.go
```diff
@@ -138,6 +140,7 @@ func PollForMMDSOpts(ctx context.Context, mmdsChan chan<- *MMDSOpts, envVars *ut
 
 			envVars.Store("E2B_SANDBOX_ID", mmdsOpts.InstanceID)
 			envVars.Store("E2B_TEMPLATE_ID", mmdsOpts.EnvID)
+			envVars.Store("E2B_EVENTS_ADDRESS", mmdsOpts.EventProxyIP)
```
sitole: It's called `E2B_EVENTS_ADDRESS` when you are putting IPv4 here. Let's set the full HTTP endpoint address here, including the `http://` prefix.

In the future, when the `address` received from MMDS is by Hyperloop One, we will use it here.

packages/orchestrator/internal/sandbox/network/network.go
```diff
@@ -214,6 +216,16 @@ func (s *Slot) CreateNetwork() error {
 		return fmt.Errorf("error creating postrouting rule: %w", err)
 	}
 
+	// Redirect traffic destined for event proxy server
```
sitole: Its still called `event proxy` (https://github.com/e2b-dev/infra/pull/1107#issuecomment-3284568557 point 1)

packages/orchestrator/internal/sandbox/network/network.go
```diff
@@ -245,6 +257,16 @@ func (s *Slot) RemoveNetwork() error {
 		if err != nil {
 			errs = append(errs, fmt.Errorf("error deleting host postrouting rule: %w", err))
 		}
+
+		// Delete event proxy redirect rule
```
sitole: Here too (https://github.com/e2b-dev/infra/pull/1107#issuecomment-3284568557 point 1)

packages/orchestrator/internal/consts.go
```diff
@@ -0,0 +1,20 @@
...
+	defaultEventProxyPort = "5010"
+)
+
+func GetSandboxEventIP() string {
+	return env.GetEnv("SANDBOX_EVENT_IP", defaultSandboxEventIP)
+}
+
+func GetEventProxyPort() string {
+	return env.GetEnv("SANDBOX_EVENT_PROXY_PORT", defaultEventProxyPort)
+}
```
sitole, line 20: Replace naming from event proxy here (https://github.com/e2b-dev/infra/pull/1107#issuecomment-3284568557 point 1).

sitole (changes requested):

packages/orchestrator/main.go
```diff
@@ -80,7 +84,12 @@
 		log.Fatalf("%d is larger than maximum possible proxy port %d", proxyPort, math.MaxInt16)
 	}
 
-	success := run(*port, *proxyPort)
+	hyperloopPort, err := strconv.ParseInt(internal.GetHyperloopProxyPort(), 10, 64)
+	if err != nil {
+		log.Fatalf("failed to get hyperloop proxy port: %v", err)
+	}
+
+	success := run(*port, *proxyPort, uint(hyperloopPort))
```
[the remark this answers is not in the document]
sitole: fixed with parsing as 32

packages/orchestrator/internal/sandbox/network/network.go
sitole: fixed

packages/orchestrator/internal/consts.go
sitole: fixed

packages/orchestrator/internal/sandbox/network/network.go
sitole: fixed

packages/envd/internal/host/mmds.go
sitole: fixed

packages/envd/internal/host/mmds.go
sitole: fixed

packages/envd/internal/host/mmds.go
sitole: fixed

packages/envd/internal/events/events.go
sitole: fixed

tests/integration/internal/tests/envd/hyperloop_test.go
```diff
@@ -0,0 +1,39 @@
...
+func TestAccessingHyperloopServer(t *testing.T) {
+	ctx, cancel := context.WithCancel(t.Context())
+	defer cancel()
+
+	client := setup.GetAPIClient()
+	sbx := utils.SetupSandboxWithCleanup(t, client, utils.WithTimeout(120))
+
+	envdClient := setup.GetEnvdClient(t, ctx)
+
+	err := utils.ExecCommand(t, ctx, sbx, envdClient, "/bin/bash", "-c", "curl -o output.txt http://203.0.113.0/me")
```
jakubno: shouldn't you rather use `events.e2b.dev` as this will break the IP address changes?

packages/orchestrator/main.go
```diff
@@ -442,6 +457,27 @@ func run(port, proxyPort uint) (success bool) {
...
+	g.Go(func() (err error) {
+		// this sets the error declared above so the function
+		// in the defer can check it.
+		hyperloopErr := hyperloopSrv.Start()
+		if hyperloopErr != nil {
+			hyperloopErr = fmt.Errorf("hyperloop server: %w", hyperloopErr)
+			zap.L().Error("hyperloop server error", zap.Error(hyperloopErr))
+
+			select {
+			case serviceError <- hyperloopErr:
```
jakubno, line 469: this will cause orchestrator to shutdown, not sure if hyperloop crash should do that

packages/envd/internal/host/network.go
```diff
@@ -0,0 +1,35 @@
...
+	}
+
+	// Otherwise, just append to the end
+	f, err := os.OpenFile("/etc/hosts", os.O_APPEND|os.O_WRONLY, 0o644)
+	if err != nil {
+		return fmt.Errorf("failed to open /etc/hosts: %w", err)
+	}
+	defer f.Close()
+
+	if _, err := f.WriteString("\n" + entry + "\n"); err != nil {
```
jakubno: this will probably break if we ever change the IP address as there will be two entries for the same address with different IPs

jakubno (changes requested):

tests/integration/internal/tests/envd/hyperloop_test.go
sitole: Ideally i would like to test both

packages/orchestrator/main.go
sitole: I think running the orchestrator without this is in an invalid state. Logger will stop working (in the future, when migrated) and events will stop coming.

packages/envd/internal/host/network.go
sitole: True, this host's management needs a major rewrite to be done properly

tests/integration/internal/tests/envd/hyperloop_test.go
sitole: We still want to use ip for logs etc. There is no much reason why use domain in envd itself

packages/envd/internal/host/network.go
sitole: now we are using lib for managing etc hosts files properly

packages/orchestrator/internal/sandbox/network/slot.go
```diff
@@ -74,6 +75,8 @@ type Slot struct {
 	hostIp   net.IP
 	hostNet  *net.IPNet
 	hostCIDR string
+
+	hyperloopIP string
```
sitole, line 79: Just to make it clearer that Hyperloop IP is only dependent on slot routing and theoretically can be different for each sandbox.

packages/envd/internal/api/init.go
```diff
@@ -61,3 +67,21 @@ func (a *API) PostInit(w http.ResponseWriter, r *http.Request) {
 
 	w.WriteHeader(http.StatusNoContent)
 }
+
+func (a *API) SetupHyperloop(address string) {
```
jakubno, line 71: 1) shouldn't we at least log that it was completed if it's async?

2) how long does it take? this should be in `µs` right? maybe we can have in sync

sitole: As its disk write time, it will probably be in milliseconds.
Sure, we can do that sync, but what should we do if it fails? Fail the whole init request?

jakubno: can you time it? 🙏🏻  I think failing silently is fine here

sitole: From testing, it takes something between `136.041µs` and `2.671696ms`. Mostly around 200µs.
It should be okay to let it sync.

jakubno (approved):

PR #1113, opened 2025-09-03
Fix base layer corrupted filesystem

dobrac:
Fix base layer corrupted filesystem. The issue was caused by a missing `sync` call to synchronize all file changes from memory before pausing the base layer. The changes need to be synchronized, because the finalize layer is always a sandbox creation (which doesn't reuse the memory -> like OS crash).

### DIFF
--- a/packages/orchestrator/internal/template/build/phases/finalize/builder.go
+++ b/packages/orchestrator/internal/template/build/phases/finalize/builder.go
@@ -126,8 +126,10 @@ func (ppb *PostProcessingBuilder) Build(
 
 	actionExecutor := layer.NewFunctionAction(ppb.postProcessingFn())
 
+	templateProvider := layer.NewCacheSourceTemplateProvider(sourceLayer.Metadata.Template)
+
 	finalLayer, err := ppb.layerExecutor.BuildLayer(ctx, layer.LayerBuildCommand{
-		SourceTemplate: sourceLayer.Metadata.Template,
+		SourceTemplate: templateProvider,
 		CurrentLayer:   currentLayer.Metadata,
 		Hash:           currentLayer.Hash,
 		UpdateEnvd:     sourceLayer.Cached,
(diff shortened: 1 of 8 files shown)


packages/orchestrator/internal/template/build/layer/resume_sandbox.go
```diff
@@ -15,13 +15,14 @@ import (
...
 
 var _ SandboxCreator = (*ResumeSandbox)(nil)
 
-func NewResumeSandbox(config sandbox.Config) *ResumeSandbox {
-	return &ResumeSandbox{config: config}
+func NewResumeSandbox(config sandbox.Config, timeout time.Duration) *ResumeSandbox {
+	return &ResumeSandbox{config: config, timeout: timeout}
 }
 
 func (f *ResumeSandbox) Sandbox(
```
jakubno: why is ResumeSandbox `f`?

packages/orchestrator/internal/template/build/phases/finalize/builder.go
```diff
@@ -23,6 +24,8 @@ import (
 	"github.com/e2b-dev/infra/packages/shared/pkg/storage"
 )
 
+const finalizeTimeout = 15 * time.Minute
```
jakubno: is there a reason why is it now 15 minutes? it was also hour if I am not mistaken

packages/orchestrator/internal/template/build/layer/resume_sandbox.go
dobrac: fixed

packages/orchestrator/internal/template/build/phases/finalize/builder.go
dobrac: Just because it should be fast. It was before 1 hour because I didn't set it tighter for the finalize layer. But the refactor for base allowed me to make it different.

dobrac: the longest (semi-valid) times from the past are around 4 minutes, but usually it's around 10 seconds with also the layer upload

dobrac: the configuration timeout is 5 minutes, ready command timeout is also 5 minutes, start command shouldn't be longer than ready command (as it's dictated by the ready command)

dobrac: updated the value so it's computed based on the two timeouts + some offset

PR #1114, opened 2025-09-04
Fix microVM clock drifts to the future

sitole:
Removing the Chrony service from template provisioning.
KVM clocks are used for every new template build and for every sandbox with envd version >= `0.2.11`.

Currently, there is an issue that templates with `0.2.11` versions that are already built can experience future time drifts as we are using Chrony and also KVM clock for clock sync. We cannot bump envd version and stop using KVM clock for this version, as we removed clock sync locks from envd.


ValentaTomas:
@sitole Are you sure the future time drifts are because of combination of kvm clock and chrony?

PR #1116, opened 2025-09-04
Rewrite query Get user from access token to sqlc

jakubno:
Refactors retrieving a user ID from an access token by replacing the previous ORM-based implementation with a direct SQL query using `sqlc`.


dobrac (approved):
Image

PR #1118, opened 2025-09-04
Limit concurrent builds

dobrac:
Limit concurrent builds. Sets the default to 20 for now


packages/api/internal/cache/templates/cache.go
```diff
@@ -257,3 +257,19 @@ func (c *TemplatesBuildCache) Get(ctx context.Context, buildID uuid.UUID, templa
 
 	return item.Value(), nil
 }
+
+// GetRunningBuildsForTeam returns all running builds for the given teamID
+// This is a simple implementation of concurrency limit
+// It does not guarantee that the limit is not exceeded, but it should be good enough for now (considering overall low number of total builds)
+func (c *TemplatesBuildCache) GetRunningBuildsForTeam(teamID uuid.UUID) ([]TemplateBuildInfo, error) {
```
jakubno: It looks like it can't return an error

jakubno (approved):
lgtm, one nit

packages/api/internal/cache/templates/cache.go
dobrac: fixed

PR #1122, opened 2025-09-04
Return total count for running sandboxes

jakubno:
Remove `X-Total-Items` as it's rather confusing
Add `X-Total-Running`


dobrac (changes requested):
Is this a breaking change? Also, can't rather properly implement the X-Total-Items?

dobrac:
> It wasn't used and not sure what it would be useful for. We would have to query all paused which would be currently a very expensive query

yes, we would have to query the count (you don't need the items). I would assume you might use it for when you do the pagination to know where is the end without reaching it

jakubno:
> > It wasn't used
> 
> how do you know this? Is the header not exposed to the user at all?

It  isn't documented, and we don't use it in our SDKs or dashboard. I highly doubt anyone would implement the list themselves, check that we return this header, and decide to use this header instead of the number of items in the body. If someone did use it, I suspect they would expect it to work differently, potentially causing a bug in their app.

dobrac (approved):

PR #1123, opened 2025-09-04
Support cpu platform and disk type configuration

djeebus:
This exposes a few fields to customize the cpu and disk type for the build and client nodes, but keeps the defaults the same.

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -62,6 +62,12 @@ variable "build_cluster_cache_disk_size_gb" {
   default     = 200
 }
 
+variable "build_cluster_cache_disk_type" {
+  description = "The GCE cache disk type for the build machines."
+  type        = string
+  default     = "pd-ssd"
+}
+
 variable "clickhouse_cluster_size" {
   type = number
 }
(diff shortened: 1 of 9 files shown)


variables.tf
```diff
@@ -377,3 +383,13 @@ variable "filestore_cache_capacity_gb" {
...
 }
+
+variable "build_min_cpu_platform" {
+  type    = string
+  default = "Intel Skylake"
+}
+
+variable "client_min_cpu_platform" {
+  type    = string
+  default = "Intel Skylake"
```
jakubno, line 395: The family must be the same for builder and client

packages/shared/scripts/e2b.toml
```diff
@@ -10,6 +10,7 @@
 # import { Sandbox } from 'e2b'
 # const sandbox = await Sandbox.create('base')
 
+team_id = "b2632569-10dd-41b1-9b2f-0b533bbc8f35"
```
jakubno: Remove this

jakubno (approved):

PR #1125, opened 2025-09-04
Change how we report `EAGAIN`s during UFFD serving

djeebus:
This line is 30% of the last 60 minutes of logs in production


packages/orchestrator/internal/sandbox/uffd/serve.go
```diff
@@ -108,9 +110,8 @@ outerLoop:
 			}
 
 			if err == syscall.EAGAIN {
-				logger.Debug("uffd: eagain error, going back to polling", zap.Error(err), zap.Int("read_bytes", n))
-
 				// Continue polling the fd.
+				time.Sleep(time.Millisecond)
```
jakubno: This isn't a log line remove

jakubno (changes requested):

packages/orchestrator/internal/sandbox/uffd/serve.go
djeebus: sorry, yup, this should be a different PR if it's worth doing. thanks for keeping me honest!

djeebus:
> Not sure about removing the log before we solve this, we identified it because of these logs, without them we have 0 observability

I don't think there's an actual bug here, it's working as designed, we're just spamming a lot of logs in a happy path.

djeebus:
talked to @ValentaTomas, we are going to log eagain counts when the goroutine ends AND each time that EAGAIN _stops_, and reset the counter when the EAGAIN stops being returned

PR #1128, opened 2025-09-05
Refactor template build cache to use sqlc

jakubno:
Also removed some unused code


packages/api/internal/cache/templates/cache.go
```diff
@@ -217,35 +216,31 @@ func (c *TemplatesBuildCache) Get(ctx context.Context, buildID uuid.UUID, templa
 	if item == nil {
 		zap.L().Debug("Template build info not found in cache, fetching from DB", logger.WithBuildID(buildID.String()))
 
-		envDB, envDBErr := c.db.GetEnv(ctx, templateID)
-		if envDBErr != nil {
-			if errors.Is(envDBErr, db.TemplateNotFoundError{}) {
+		template, err := c.db.GetTemplateBuildWithEnv(ctx, queries.GetTemplateBuildWithEnvParams{
```
jakubno: now we get both in query

packages/api/internal/handlers/template_build_status.go
```diff
@@ -36,11 +36,6 @@ func (a *APIStore) GetTemplatesTemplateIDBuildsBuildIDStatus(c *gin.Context, tem
 			return
 		}
 
-		if errors.Is(err, db.TemplateNotFoundError{}) {
```
jakubno, line 39: as there's only one query, there can be only one error

packages/api/internal/handlers/template_build_status.go
```diff
@@ -126,8 +121,8 @@ func getCorrespondingTemplateBuildStatus(s envbuild.Status) api.TemplateBuildSta
 	}
 }
 
-func getAPIReason(reason *schema.BuildReason) *api.BuildStatusReason {
-	if reason == nil {
+func getAPIReason(reason types.BuildReason) *api.BuildStatusReason {
+	if reason.Message == "" {
```
jakubno, line 125: not sure about this, should I make it `nil`-able? There are still some `{}` in DB, which won't be handled by making it `nil`-able

packages/api/internal/handlers/template_start_build.go
```diff
@@ -87,7 +87,7 @@ func (a *APIStore) PostTemplatesTemplateIDBuildsBuildID(c *gin.Context, template
 	telemetry.ReportEvent(ctx, "started environment build")
 
 	// Check if the user has access to the template, load the template with build info
-	templateBuildDB, err := a.sqlcDB.GetTemplateBuild(ctx, queries.GetTemplateBuildParams{
+	templateBuildDB, err := a.sqlcDB.GetTemplateBuildWithEnv(ctx, queries.GetTemplateBuildWithEnvParams{
```
jakubno: just a rename to properly track what the query do

packages/api/internal/cache/templates/cache.go
jakubno: also this was returning only builds in status 
```
envbuild.StatusWaiting
envbuild.StatusBuilding
```
which doesn't make sense for the template build status endpoint. We should return failed and succeeded builds as well

packages/db/migrations/20250905134524_fix_env_builds_non_null_fields.sql
```diff
@@ -0,0 +1,19 @@
+-- +goose Up
+-- +goose StatementBegin
+UPDATE public.env_builds
+    SET reason = '{}'::jsonb
+    WHERE reason IS NULL;
+
+ALTER TABLE public.env_builds
+    ALTER COLUMN env_id SET NOT NULL,
+    ALTER COLUMN reason SET DEFAULT '{}'::jsonb,
```
jakubno, line 9: now there's only possible 1 empty value - `{}`

packages/api/internal/handlers/template_start_build.go
dobrac: can you rename it to GetTemplateBuildWithTemplate? Because we don't refer to templates as `env` anymore in new code

packages/api/internal/orchestrator/pause_instance.go
```diff
@@ -71,7 +74,14 @@ func (o *Orchestrator) PauseInstance(
 		return fmt.Errorf("error pausing sandbox: %w", err)
 	}
 
-	err = o.dbClient.EnvBuildSetStatus(ctx, *envBuild.EnvID, envBuild.ID, envbuild.StatusSuccess, nil)
+	now := time.Now()
+	err = o.sqlcDB.UpdateEnvBuildStatus(ctx, queries.UpdateEnvBuildStatusParams{
```
dobrac, line 74: same here, rename to UpdateTemplateBuildStatus

packages/db/queries/update_template_build_status.sql
```diff
@@ -0,0 +1,6 @@
+-- name: UpdateEnvBuildStatus :exec
+UPDATE "public"."env_builds"
+SET status = @status,
+    finished_at = @finished_at,
+    reason = sqlc.narg(reason)
+WHERE id = @build_id AND env_id = @env_id;
```
dobrac:
```suggestion
WHERE id = @build_id AND env_id = @template_id;
```

packages/api/internal/handlers/template_build_status.go
dobrac: this is probably alright

other option would be to make the buildreason nullable and remove the {} from DB (these are I think artifacts of entgo handling nil as {})

dobrac (dismissed):
lgtm, just few nits

packages/api/internal/cache/templates/cache.go
```diff
@@ -217,35 +216,28 @@ func (c *TemplatesBuildCache) Get(ctx context.Context, buildID uuid.UUID, templa
...
-			}
-
-			return TemplateBuildInfo{}, fmt.Errorf("failed to get template '%s': %w", buildID, envDBErr)
-		}
-
-		// making sure associated template build really exists
-		envBuildDB, envBuildDBErr := c.db.GetEnvBuild(ctx, buildID)
-		if envBuildDBErr != nil {
-			if errors.Is(envBuildDBErr, db.TemplateBuildNotFoundError{}) {
+		template, err := c.db.GetTemplateBuildWithEnv(ctx, queries.GetTemplateBuildWithEnvParams{
```
sitole: Can we name it `result` or something else? Its confusing to take `template.Env` here.

packages/db/migrations/20250905134524_fix_env_builds_non_null_fields.sql
```diff
@@ -0,0 +1,19 @@
+-- +goose Up
+-- +goose StatementBegin
+UPDATE public.env_builds
+    SET reason = '{}'::jsonb
+    WHERE reason IS NULL;
+
+ALTER TABLE public.env_builds
+    ALTER COLUMN env_id SET NOT NULL,
```
sitole, line 8: Are we sure envd_id and reason cannot be NULL on some older builds?

packages/shared/pkg/models/event/sandbox_event.go
```diff
@@ -0,0 +1,19 @@
+package event
```
sitole, line 1: Why are sandbox events and webhooks commit reverse here?

jakubno: merge conflict, sorry

packages/db/migrations/20250905134524_fix_env_builds_non_null_fields.sql
jakubno: yes, env_id is always defined
reason can be null, but the default is set to `{}` which we are treating the same

sitole (approved):

packages/shared/pkg/models/event/sandbox_event.go
sitole: its still there

PR #1132, opened 2025-09-05
Include sandbox metadata in events

0div:
# Description 

Sandbox metadata is included in all lifecycle events (create, kill, pause, resume, update)

1. Create sbx with metadata:
```js
Sandbox.create('base', {
  metadata: {
    test_key: 'test_value',
  },
})
```
2. Register and receive webhook:
```json
{
  "event_category": "lifecycle",
  "event_data": {
    "sandbox_metadata": {
      "test_key": "test_value"
    }
  },
  "event_label": "kill",
  "sandbox_build_id": "f0872058-d08f-431c-a218-f3c5b4c52715",
  "sandbox_execution_id": "a8742a77-269e-4d95-a5a4-4795f0da6910",
  "sandbox_id": "i7qugrfqgh0wu3kt4os5m",
  "sandbox_team_id": "460355b3-4f64-48f9-9a16-4442817f79f5",
  "sandbox_template_id": "rki5dems9wqfm4r03t7g",
  "timestamp": "2025-09-05T03:50:11.181435712Z"
}
```

3. Same for event table:
```
SELECT
    sandbox_id,
    event_data
FROM sandbox_events
WHERE sandbox_id = 'im6lb0colfzijzwmfkjhh'

Query id: e36b5a7f-3650-49cf-a212-b8055016822f

   ┌─sandbox_id────────────┬─event_data──────────────────────────────────────┐
1. │ im6lb0colfzijzwmfkjhh │ {"sandbox_metadata":{"test_key": "test_value"}} │
```


packages/orchestrator/internal/events/sandbox_events_service.go
```diff
@@ -56,13 +57,7 @@ func (es *SandboxEventsService) HandleEvent(ctx context.Context, event event.San
 		return
 	}
 
-	// Create a new context without cancel, so we can pass it to the goroutines
-	// and not worry about the parent context being cancelled.
-	// This is important because we want to ensure that the goroutines are not cancelled
-	// when the parent context is cancelled.
-	childCtx := context.WithoutCancel(ctx)
```
ValentaTomas, line 63: To confirm—the `withoutCancel` is now on the level of sandboxEvent calls, right?

0div: correct, at the calling level like [here](https://github.com/e2b-dev/infra/pull/1132/files#diff-2ec9d602a87e26312ede1d987a89dcd8e1262a46d5c50d0b637d636646cdfa25R175)

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -214,20 +207,10 @@ func (s *server) Update(ctx context.Context, req *orchestrator.SandboxUpdateRequ
...
-	if err != nil {
-		sbxlogger.I(sbx).Error("error parsing team ID", zap.String("team_id", sbx.Runtime.TeamID), zap.Error(err))
-	}
-
-	buildId := ""
-	if sbx.APIStoredConfig != nil {
-		buildId = sbx.APIStoredConfig.BuildId
-	}
+	teamID, buildId, sbxMetadata := s.prepareSandboxEventData(sbx)
+	sbxMetadata["set_timeout"] = req.EndTime.AsTime().Format(time.RFC3339)
```
ValentaTomas: Is this `sbxMetadata` or actually event data?

0div: it is set as root key level in the map (not in `sandbox_metadata` value), but maybe the map variable name is confusing

packages/shared/pkg/events/event/sandbox_event.go
```diff
@@ -7,13 +7,13 @@ import (
...
-	EventData          string    `json:"event_data,omitempty"`
+	Timestamp          time.Time      `json:"timestamp"`
+	SandboxID          string         `json:"sandbox_id"`
+	SandboxExecutionID string         `json:"sandbox_execution_id"`
+	SandboxTemplateID  string         `json:"sandbox_template_id"`
+	SandboxBuildID     string         `json:"sandbox_build_id"`
+	SandboxTeamID      uuid.UUID      `json:"sandbox_team_id"`
+	EventCategory      string         `json:"event_category"`
+	EventLabel         string         `json:"event_label"`
+	EventData          map[string]any `json:"event_data,omitempty"`
```
ValentaTomas, line 18: The `map[string]any` gets serialized as map here? (We switched from string?)

0div: correct

ValentaTomas: I was confused as in `sandbox_event_service.go` are are passing a string there:
```
EventData:          sql.NullString{String: eventData, Valid: eventData != ""},
```

0div: it does for storing in postgres, yes.

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -446,8 +410,28 @@ func (s *server) Pause(ctx context.Context, in *orchestrator.SandboxPauseRequest
...
+// Extracts common data needed for sandbox events
+func (s *server) prepareSandboxEventData(sbx *sandbox.Sandbox) (uuid.UUID, string, map[string]any) {
+	teamID, err := uuid.Parse(sbx.Runtime.TeamID)
+	if err != nil {
+		sbxlogger.I(sbx).Error("error parsing team ID", zap.String("team_id", sbx.Runtime.TeamID), zap.Error(err))
+	}
+
+	buildId := ""
+	eventData := make(map[string]any)
+	if sbx.SandboxConfig != nil {
```
ValentaTomas: Can the sandbox even data be valid if the config is nil?

0div: it will be json encoded to `{}` if config is nil

packages/shared/pkg/utils/copy.go
```diff
@@ -0,0 +1,9 @@
+package utils
+
+func CopyMap[K comparable, V any](m map[K]V) map[K]V {
```
jakubno: this is just a shallow copy, please add a comment or improve the name to prevent confusion

ValentaTomas:
The only other thing I'm thinking about is if the sandbox metadata should be on the even data or separately:
```
  "event_data": {
    "sandbox_metadata": {
      "test_key": "test_value"
    }
  },
 ```
 
 vs
 
 ```
  "event_data": {
  },
  "sandbox_metadata": {
    "test_key": "test_value"
  }
 ```

File: packages/orchestrator/internal/server/sandboxes.go (hunk 2 above)
Comment (jakubno) [reply packages/orchestrator/internal/server/sandboxes.go]: agree with Tomas here, it should be named `eventData`

File: packages/shared/pkg/utils/copy.go (hunk 5 above)
Comment (0div) [reply packages/shared/pkg/utils/copy.go]: I updated the name, for now we only need shallow copy for sandbox metadata as it's `map[string]string`

File: packages/orchestrator/internal/sandbox/sandbox.go [hunk 6]
```diff
@@ -97,7 +97,7 @@ type Sandbox struct {
 
 	Checks *Checks
 
-	APIStoredConfig *orchestrator.SandboxConfig
```
jakubno, line 100: why have you changed this? Please also add a comment that this shouldn't be use and mark it as deprecated to prevent using this in future

0div: i simply thought it would be helpful to keep the naming consistent with its type and how it's named elsewhere, i got a bit confused looking it up at first so i thought it would help the next eng.

jakubno: As I said, it's on purpose. These data shouldn't be used anymore, we'll be removing them soon.

PR #1133, opened 2025-09-05
Write to cache when building a template

djeebus:


packages/api/internal/cache/templates/cache.go
```diff
@@ -74,6 +77,9 @@ func NewTemplateCache(db *sqlcdb.Client) *TemplateCache {
 }
 
 func (c *TemplateCache) Get(ctx context.Context, aliasOrEnvID string, teamID uuid.UUID, clusterID uuid.UUID, public bool) (*api.Template, *queries.EnvBuild, *api.APIError) {
+	ctx, span := tracer.Start(ctx, "TemplateCache.Get")
```
dobrac: one thing, not fully related, have you checked also the guidelines how to name the spans? I have found this: https://opentelemetry.io/blog/2025/how-to-name-your-spans/

packages/shared/pkg/storage/providers/storage_cache.go
```diff
@@ -147,15 +173,83 @@ func (c *CachedFileObjectProvider) WriteTo(ctx context.Context, dst io.Writer) (
...
+		trace.WithAttributes(attribute.String("path", path)))
+	defer span.End()
+
+	// write the file to the disk and the remote system at the same time.
+	// this opens the file twice, but the API makes it difficult to use a MultiWriter
+
+	var eg errgroup.Group
+
+	eg.Go(func() error {
+		return c.createCacheBlocksFromFile(ctx, path)
```
dobrac: I think it would be nice to wrap the error, so we know which goroutine it comes from if it happens

.github/workflows/pr-go-integrity.yml
```diff
@@ -9,11 +9,29 @@ jobs:
...
+      - uses: hashicorp/setup-terraform@v3
+        with:
+          terraform_version: "${{ env.TERRAFORM }}"
 
       - name: Run integrity check
-        run: STRICT_MODE=1 bash ./scripts/golang-dependencies-integrity.sh
+        run: make tidy
+
+      - name: Fix tracers
+        run: ./scripts/fix-tracers.sh
```
dobrac, line 27: can you move this to a separate PR please?

packages/shared/pkg/storage/providers/storage_cache.go
```diff
@@ -147,15 +173,83 @@ func (c *CachedFileObjectProvider) WriteTo(ctx context.Context, dst io.Writer) (
...
+	eg.Go(func() error {
+		return c.inner.WriteFromFileSystem(ctx, path)
+	})
+
+	return eg.Wait()
 }
 
 func (c *CachedFileObjectProvider) Write(ctx context.Context, src []byte) (int, error) {
-	return c.inner.Write(ctx, src)
+	var err error
```
dobrac: why this?

packages/shared/pkg/storage/providers/storage_cache.go
```diff
@@ -147,15 +173,83 @@ func (c *CachedFileObjectProvider) WriteTo(ctx context.Context, dst io.Writer) (
...
+	}
+
+	return num, nil
+}
+
+// writeCacheAndRemote simultaneously writes a full file to both local cache and the remote persistence store. It does
+// not need to worry about race conditions, as the files will only exist on the local machine, and can't be generated
+// in parallel on any other machines.
+func (c *CachedFileObjectProvider) writeCacheAndRemote(ctx context.Context, src []byte) (int, error) {
+	var err error
```
dobrac: same here

packages/shared/pkg/storage/providers/storage_cache.go
```diff
@@ -227,9 +323,107 @@ func (c *CachedFileObjectProvider) Size(ctx context.Context) (int64, error) {
...
+		if err := os.RemoveAll(c.path); ignoreFileMissingError(err) != nil {
+			zap.L().Error("error on cache delete", zap.String("path", c.path), zap.Error(err))
+		}
+	}()
+
 	return c.inner.Delete(ctx)
 }
 
+func (c *CachedFileObjectProvider) createCacheBlocksFromFile(ctx context.Context, inputPath string) error {
+	var err error
```
dobrac: here too

packages/shared/pkg/storage/providers/storage_cache.go
```diff
@@ -227,9 +323,107 @@ func (c *CachedFileObjectProvider) Size(ctx context.Context) (int64, error) {
...
+			})
+		}(offset, totalSize)
+	}
+	return errs.Wait()
+}
+
+// writeChunkFromFile writes a piece of a local file. It does not need to worry about race conditions, as it will only
+// be called when building templates, and templates cannot be built on multiple machines at the same time.x
+func (c *CachedFileObjectProvider) writeChunkFromFile(ctx context.Context, offset int64, totalSize int64, input *os.File) (string, error) {
+	var err error
```
dobrac: too

packages/api/internal/cache/templates/cache.go
djeebus: Ah, good advice! No, I hadn't seen that, but I'll clean up the spans so they follow this naming convention. Thanks!

packages/shared/pkg/storage/providers/factory.go
dobrac, line 1: not sure about the `main.go` filename if thats any convention. I've seen only main.go in executable (main program entry) contexts

packages/shared/pkg/storage/providers/storage_cache.go
```diff
@@ -90,15 +117,15 @@ func (c CachedProvider) GetDetails() string {
...
-var _ StorageObjectProvider = (*CachedFileObjectProvider)(nil)
+var _ storage.StorageObjectProvider = (*CachedFileObjectProvider)(nil)
 
-// WriteTo is used for very small files and we can check agains their size to ensure the content is valid.
-func (c *CachedFileObjectProvider) WriteTo(ctx context.Context, dst io.Writer) (int64, error) {
+// WriteTo is used for very small files, and we can check against their size to ensure the content is valid.
+func (c *CachedFileObjectProvider) WriteTo(ctx context.Context, dst io.Writer) (written int64, err error) {
 	ctx, span := tracer.Start(ctx, "CachedFileObjectProvider.WriteTo")
-	defer span.End()
+	defer endSpan(span, err)
```
dobrac: will this work and not be err == nil always?

packages/shared/pkg/storage/providers/storage_cache.go
```diff
@@ -146,24 +174,86 @@ func (c *CachedFileObjectProvider) WriteTo(ctx context.Context, dst io.Writer) (
...
+
+func (c *CachedFileObjectProvider) WriteFromFileSystem(ctx context.Context, path string) (err error) {
+	ctx, span := tracer.Start(ctx, "CachedFileObjectProvider.WriteFromFileSystem",
+		trace.WithAttributes(attribute.String("path", path)))
+	defer endSpan(span, err)
+
+	// write the file to the disk and the remote system at the same time.
+	// this opens the file twice, but the API makes it difficult to use a MultiWriter
+
+	go c.createCacheBlocksFromFile(ctx, path)
```
dobrac: is there any guarantee that the source file exists at the point when this goroutine is processing it?

packages/shared/pkg/storage/providers/storage_cache.go
```diff
@@ -146,24 +174,86 @@ func (c *CachedFileObjectProvider) WriteTo(ctx context.Context, dst io.Writer) (
...
+	defer endSpan(span, err)
+
+	size, err = c.inner.Write(ctx, src)
+	if err != nil {
+		return 0, fmt.Errorf("failed to remote write from byte array: %w", err)
+	}
+	if size != len(src) {
+		zap.L().Warn("remote write didn't match data length",
+			zap.Int("expected_size", len(src)),
+			zap.Int("actual_size", size))
```
dobrac: might be nice to include the rootpath too

packages/shared/pkg/storage/storage_cache.go
```diff
@@ -227,9 +319,107 @@ func (c *CachedFileObjectProvider) Size(ctx context.Context) (int64, error) {
...
+	defer cleanup("failed to close file", input)
+
+	stat, err := input.Stat()
+	if err != nil {
+		zap.L().Error("failed to stat input file", zap.String("path", inputPath), zap.Error(err))
+		return
+	}
+
+	totalSize := stat.Size()
+	var wg sync.WaitGroup
```
dobrac: how about using errorgroup instead which has included concurrency limit by default?

dobrac (dismissed):

packages/shared/pkg/storage/providers/factory.go
djeebus: I'll change it to "factory.go"

packages/shared/pkg/storage/providers/storage_cache.go
djeebus: Named returns get assigned when returning, so `err` will always be assigned the error that's being returned. Check this out:  https://go.dev/play/p/gIJnNXQAb7C

packages/shared/pkg/storage/providers/storage_cache.go
djeebus: By convention, this is called when a sandbox is being saved, but we never delete cached files unless the disk is filling up, at which point we delete old files. Regardless, if this fails, we log an error and move on. If the file is  requested again, we pull from cache.  Also worth noting that the remote write will take longer than the cache write, so we should be done long before the function exits.

packages/shared/pkg/storage/storage_cache.go
djeebus: errgroup will cancel the context and goroutines if an error is encountered, which we don't want. If a single chunk fails for some reason, we still want all of the rest of the chunks to write to disk.

packages/shared/pkg/storage/providers/storage_cache.go
djeebus: Whoops, removed this. It was leftover from a different method of recording errors at the end of a span. I like the new way better, and cleaned this up.

packages/shared/pkg/storage/providers/storage_cache.go
djeebus: Same as above

packages/shared/pkg/storage/providers/storage_cache.go
djeebus: Same as above

packages/shared/pkg/storage/providers/storage_cache.go
dobrac: This is not the same example. Check this out: https://go.dev/play/p/q4zjCGGhgTI

When variable is passed as an argument to a function, it’s value is from the time of defer code call, not the time of defer function execution. (Of course if its a pointer, then the target is the same.)

dobrac: Both of them are interfaces, so ti should be ok I guess. But please test it

djeebus: Yup, you're totally right. I thought I was making it simpler, I just broke it. Sad times. Thanks!

djeebus: Even with pointers/interfaces, my version was broken: https://go.dev/play/p/ZNdFlIrBWFj

packages/shared/pkg/storage/storage_cache.go
```diff
@@ -180,9 +237,65 @@ func (c *CachedFileObjectProvider) Size(ctx context.Context) (int64, error) {
...
+			// write to the cache file
+			filename := c.makeChunkFilename(int64(offset))
+			if err2 := os.WriteFile(filename, buf, cacheFilePermissions); err2 != nil {
+				safelyRemoveFile(filename)
+				zap.L().Warn("failed to write chunk file",
+					zap.String("filename", filename),
+					zap.Error(err2))
+			}
+		}(offset, buf)
+	}
```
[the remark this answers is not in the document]
djeebus: 1. the parent directory is created on `CachedProvider.OpenObject`
2. the `src` slice is not modified before this loop exits
3. ... the 3rd point is fair.

djeebus:
This is blocked by #1361 ; will refactor after that gets merged.

ValentaTomas:
My one concern here is that because some teams are building tens of thousands of templates, this might start ejecting things disproportionately.

packages/shared/pkg/storage/lock/atomic_file.go
```diff
@@ -0,0 +1,101 @@
...
+		if errors.Is(err, os.ErrExist) {
+			// Someone else created newPath first. Treat as success.
+			return nil
+		}
+
+		return err
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
djeebus: No interest in recovering the temp data, we'll get it next time.

Makefile
```diff
@@ -168,7 +168,7 @@ fmt:
 
 .PHONY: lint
 lint:
-	go work edit -json | jq -r '.Use[].DiskPath' | xargs -P 10 -I{} golangci-lint run {}/... --fix
+	go work edit -json | jq -r '.Use[].DiskPath' | xargs -P 4 -I{} golangci-lint run {}/... --fix
```
dobrac, line 171: just double checking if this was wanted

packages/orchestrator/internal/template/build/builder.go
```diff
@@ -213,21 +213,37 @@ func (b *Builder) Build(ctx context.Context, template storage.TemplateFiles, cfg
 	return runBuild(ctx, l, buildContext, b)
 }
 
+func (b *Builder) useNFSCache(ctx context.Context) (string, bool) {
+	if b.config.SharedChunkCacheDir == "" {
+		// can't enable cache if we don't have a cache path
+		return "", false
+	}
```
dobrac: should we move this below the `BuildingFeatureFlagName` flag and add warning? (I assume that is kinda not wanted state)

packages/shared/pkg/feature-flags/flags.go
```diff
@@ -74,6 +74,8 @@ var (
 	MetricsReadFlagName                = newBoolFlag("sandbox-metrics-read", env.IsDevelopment())
 	SnapshotFeatureFlagName            = newBoolFlag("use-nfs-for-snapshots", env.IsDevelopment())
 	TemplateFeatureFlagName            = newBoolFlag("use-nfs-for-templates", env.IsDevelopment())
+	WriteToCacheOnWrites               = newBoolFlag("write-to-cache-on-writes", false)
+	BuildingFeatureFlagName            = newBoolFlag("use-nfs-for-building-templates", env.IsDevelopment())
```
dobrac: nit: should we name it instead of `BuildingFeatureFlagName` something like `UseNFSForBuildingFlagName` ?

packages/shared/pkg/storage/storage_cache.go
```diff
@@ -35,25 +37,26 @@ var (
 		"Total bytes written to the cache",
 		"Total writes to the cache",
 	))
-	cacheHits = utils.Must(meter.Int64Counter("orchestrator.storage.cache.hits",
-		metric.WithDescription("total cache hits")))
-	cacheMisses = utils.Must(meter.Int64Counter("orchestrator.storage.cache.misses",
-		metric.WithDescription("total cache misses")))
```
dobrac, line 41: just note: please update after the dashboards too :)

packages/shared/pkg/storage/storage_cache.go
```diff
@@ -35,25 +37,26 @@ var (
...
-func NewCachedProvider(rootPath string, inner StorageProvider) *CachedProvider {
-	return &CachedProvider{rootPath: rootPath, inner: inner, chunkSize: MemoryChunkSize}
+func NewCachedProvider(rootPath string, inner StorageProvider, flags *featureflags.Client) *CachedProvider {
+	return &CachedProvider{rootPath: rootPath, inner: inner, chunkSize: MemoryChunkSize, flags: flags}
 }
 
 func (c CachedProvider) DeleteObjectsWithPrefix(ctx context.Context, prefix string) error {
+	go func(ctx context.Context) {
+		c.deleteObjectsWithPrefix(ctx, prefix)
+	}(context.WithoutCancel(ctx))
```
dobrac, line 62: could you add comment here explaining what it does and why it's run in a goroutine please?

packages/shared/pkg/storage/storage_cache.go
```diff
@@ -86,14 +89,30 @@ func (c CachedProvider) OpenSeekableObject(ctx context.Context, path string, obj
...
+			zap.String("path", fullPrefix),
+			zap.Error(err))
+	}
+}
+
+func cleanupCtx(ctx context.Context, msg string, fn func(ctx context.Context) error) {
+	if err := fn(ctx); err != nil {
+		logger.L().Warn(ctx, msg, zap.Error(err))
+	}
+}
```
dobrac: how about moving this helper function to utils - if you think it's useful (this is a second place where it's used)

packages/shared/pkg/storage/storage_cache_seekable.go
```diff
@@ -24,10 +29,17 @@ var (
 	ErrBufferTooLarge  = errors.New("buffer is too large")
 )
 
+const maxCacheWriterConcurrency = 10
```
dobrac: should we make this also a feature flag?

packages/shared/pkg/storage/storage_cache_seekable.go
```diff
@@ -24,10 +29,17 @@ var (
...
+
+type featureFlagsClient interface {
+	BoolFlag(ctx context.Context, flag featureflags.BoolFlag, ldctx ...ldcontext.Context) bool
+}
```
dobrac, line 51: isn't there already existing interface, or, do we need this at all? I think we have already tests where the feature flags client is used with offline storage and has also some flags set

packages/shared/pkg/storage/storage_cache_seekable.go
```diff
@@ -64,36 +81,84 @@ func (c CachedSeekableObjectProvider) ReadAt(ctx context.Context, buff []byte, o
...
 	if err != nil {
-		return 0, fmt.Errorf("failed to perform uncached read: %w", err)
+		return readCount, fmt.Errorf("failed to perform uncached read: %w", err)
 	}
 
-	go func() {
-		c.writeChunkToCache(context.WithoutCancel(ctx), offset, chunkPath, buff[:readCount])
-	}()
+	shadowBuff := make([]byte, readCount)
+	copy(shadowBuff, buff[:readCount])
```
dobrac, line 108: do you know if this is solving any specific issue you've encountered? Or is it just for ownership control

packages/shared/pkg/storage/storage_cache_seekable.go
```diff
@@ -64,36 +81,84 @@ func (c CachedSeekableObjectProvider) ReadAt(ctx context.Context, buff []byte, o
...
+
+	recordCacheRead(ctx, false, 8, cacheTypeSeekable, cacheOpSize)
 
 	return size, nil
 }
 
-func (c CachedSeekableObjectProvider) WriteFromFileSystem(ctx context.Context, path string) error {
-	return c.inner.WriteFromFileSystem(ctx, path)
+func (c CachedSeekableObjectProvider) WriteFromFileSystem(ctx context.Context, path string) (err error) {
+	ctx, span := tracer.Start(ctx, "CachedFileObjectProvider.WriteFromFileSystem",
```
dobrac: wdyt about keeping the convention of naming as `{verb} {object}`? (Based on OTEL blog post)

packages/shared/pkg/storage/storage_cache_metrics.go
```diff
@@ -0,0 +1,109 @@
...
+		errorType = "write-lock"
+	} else {
+		errorType = "write"
+	}
+
+	logger.L().Warn(ctx, "failed to write to cache",
+		zap.Error(err),
+		zap.String("cache_type", string(t)),
+		zap.String("op_type", string(op)),
+	)
```
dobrac, line 134: we probably don't want to log the lock.ErrLockAlreadyHeld errors (we previously weren't)

packages/shared/pkg/utils/move.go
```diff
@@ -0,0 +1,41 @@
...
+
+// AtomicMove tries to rename a file but will not replace the target if it already exists.
+// If the file already exists, the file will be deleted.
+func AtomicMove(oldPath, newPath string) error {
+	return atomicMove(oldPath, newPath, &osFns{})
+}
+
+func atomicMove(oldPath, newPath string, os fileOps) error {
+	if err := os.Link(oldPath, newPath); err != nil {
+		return fmt.Errorf("failed to create hard link: %w", err)
```
dobrac, line 30: this fails to remove the oldPath (it was previously removing it)

packages/shared/pkg/storage/storage_cache_seekable.go
```diff
@@ -197,47 +248,33 @@ func (c CachedSeekableObjectProvider) writeChunkToCache(ctx context.Context, off
...
 	if err != nil {
-		if errors.Is(err, lock.ErrLockAlreadyHeld) {
-			// Another process is already writing this chunk, so we can skip writing it ourselves
-			return
-		}
-
-		logger.L().Warn(ctx, "failed to acquire lock", zap.String("path", finalFilename), zap.Error(err))
-
-		return
+		return fmt.Errorf("failed to acquire lock for local size: %w", err)
```
dobrac, line 325: is the ignore of `if errors.Is(err, lock.ErrLockAlreadyHeld) {` here deliberately? If so, why is it different from the `writeChunkToCache`?

packages/shared/pkg/storage/storage_cache_seekable.go
```diff
@@ -253,20 +290,127 @@ func (c CachedSeekableObjectProvider) writeLocalSize(ctx context.Context, size i
...
+	if err != nil {
+		return 0, fmt.Errorf("failed to stat input file: %w", err)
 	}
 
-	if err := moveWithoutReplace(ctx, tempFilename, finalFilename); err != nil {
-		logger.L().Warn(ctx, "failed to move temp file",
-			zap.String("temp_path", tempFilename),
-			zap.String("final_path", finalFilename),
+	totalSize := stat.Size()
+	var wg sync.WaitGroup
```
dobrac: how about using error group with limit instead?

packages/shared/pkg/storage/storage_cache_seekable.go
```diff
@@ -253,20 +290,127 @@ func (c CachedSeekableObjectProvider) writeLocalSize(ctx context.Context, size i
...
+
+	workers := make(chan struct{}, maxCacheWriterConcurrency)
+	for offset := int64(0); offset < totalSize; offset += c.chunkSize {
```
dobrac: is it possible that the last chunk is smaller than chunkSize?

packages/shared/pkg/storage/storage_cache_seekable.go
```diff
@@ -253,20 +290,127 @@ func (c CachedSeekableObjectProvider) writeLocalSize(ctx context.Context, size i
...
+					zap.Error(err))
+			}
+		}(offset)
+	}
+	wg.Wait()
+
+	return totalSize, nil
+}
+
+type offsetReader struct {
```
dobrac: can you move this to a sperate file please?

packages/shared/pkg/storage/storage_cache_seekable.go
```diff
@@ -253,20 +290,127 @@ func (c CachedSeekableObjectProvider) writeLocalSize(ctx context.Context, size i
...
+
+	return
+}
+
+func newOffsetReader(file *os.File, offset int64) *offsetReader {
+	return &offsetReader{file, offset}
+}
+
+// writeChunkFromFile writes a piece of a local file. It does not need to worry about race conditions, as it will only
+// be called when building templates, and templates cannot be built on multiple machines at the same time.x
```
dobrac:
```suggestion
// be called when building templates, and templates cannot be built on multiple machines at the same time.
```

packages/shared/pkg/storage/storage_cache_seekable.go
```diff
@@ -253,20 +290,127 @@ func (c CachedSeekableObjectProvider) writeLocalSize(ctx context.Context, size i
...
+	r.offset += int64(n)
+
+	return
+}
+
+func newOffsetReader(file *os.File, offset int64) *offsetReader {
+	return &offsetReader{file, offset}
+}
+
+// writeChunkFromFile writes a piece of a local file. It does not need to worry about race conditions, as it will only
```
dobrac, line 397: can you please clarify how and why is this template building only related? Also if possible, I would worry about race conditions as building in parallel is to some extent possible.

packages/shared/pkg/storage/storage_cache_seekable.go
```diff
@@ -253,20 +290,127 @@ func (c CachedSeekableObjectProvider) writeLocalSize(ctx context.Context, size i
 
 	tempFilename := filepath.Join(c.path, fmt.Sprintf(".size.bin.%s", uuid.NewString()))
 
+	defer func() {
+		if err := os.Remove(tempFilename); err != nil && !os.IsNotExist(err) {
```
dobrac: can we use `safelyRemoveFile` instead? (same for other occurrences)

packages/shared/pkg/storage/storage_cache_object.go
```diff
@@ -36,132 +37,155 @@ func (c CachedObjectProvider) WriteTo(ctx context.Context, dst io.Writer) (n int
...
 		return bytesRead, nil
 	}
 
-	return c.readAndCacheFullRemoteFile(ctx, dst)
+	recordCacheReadError(ctx, cacheTypeObject, cacheOpWriteTo, err)
+
+	// This is semi-arbitrary. this code path is called for files that tend to be less than 1 MB (headers, metadata, etc),
+	// so 2 MB allows us to read the file without needing to allocate more memory, with some room for growth. If the
+	// file is larger than 2 MB, the buffer will grow, it just won't be as efficient WRT memory allocations.
+	const writeToInitialBufferSize = 2 * megabyte
```
dobrac, line 59: can you move this constant to the top level?

dobrac: also, do you think it would make sense to make it a feature flag?

packages/shared/pkg/storage/storage_cache_object.go
```diff
@@ -36,132 +37,155 @@ func (c CachedObjectProvider) WriteTo(ctx context.Context, dst io.Writer) (n int
...
+func (c CachedObjectProvider) Write(ctx context.Context, p []byte) (n int, e error) {
+	ctx, span := tracer.Start(ctx, "CachedFileObjectProvider.Write", trace.WithAttributes(attribute.Int("size", len(p))))
+	defer func() {
+		recordError(span, e)
+		span.End()
+	}()
+
+	var wg sync.WaitGroup
+
+	wg.Go(func() {
```
dobrac: do we want to await also this cache write when just doing any write? (afaik we don't do it for the seekable)

packages/shared/pkg/storage/storage_cache_object_test.go
```diff
@@ -0,0 +1,124 @@
...
+			RunAndReturn(func(_ context.Context, w io.Writer) (int64, error) {
+				num, err := w.Write(actualData)
+
+				return int64(num), err
+			})
+
+		c := CachedObjectProvider{path: cacheDir, inner: inner, chunkSize: 1024}
+
+		buff := bytes.NewBuffer(nil)
+		bytesRead, err := c.WriteTo(t.Context(), buff)
```
dobrac, line 84: can we please make a new PR and rename Write to Read? :D It's really confusing

packages/shared/pkg/storage/storage_cache_seekable.go
briann: This is intentional - RESTRICT prevents accidental deletion of workflows that have active experiments. The `delete_workflow()` service will need to be updated (in a future phase) to first check for/delete child experiments, providing explicit user confirmation before cascading deletes. This is safer for operational entities.

briann: Fixed: Changed to `ON DELETE CASCADE`. Deleting a parent workflow now automatically cascades to delete all experiment children - no separate cleanup needed in `delete_workflow()`.

packages/shared/pkg/feature-flags/flags.go
levb: another nit, `WriteToCacheOnWrites` should also be suffixed with `FlagName` for consistency?

packages/shared/pkg/storage/storage_cache.go
levb: 1/5 maybe `WarnIfFailed[Ctx]`? Also, maybe return the err? doesn't hurt?

packages/shared/pkg/storage/storage_cache_object.go
```diff
@@ -36,132 +37,155 @@ func (c CachedObjectProvider) WriteTo(ctx context.Context, dst io.Writer) (n int
...
+	recordCacheReadError(ctx, cacheTypeObject, cacheOpWriteTo, err)
+
+	// This is semi-arbitrary. this code path is called for files that tend to be less than 1 MB (headers, metadata, etc),
+	// so 2 MB allows us to read the file without needing to allocate more memory, with some room for growth. If the
+	// file is larger than 2 MB, the buffer will grow, it just won't be as efficient WRT memory allocations.
+	const writeToInitialBufferSize = 2 * megabyte
+
+	buffer := bytes.NewBuffer(make([]byte, 0, writeToInitialBufferSize))
+
+	if _, err := c.inner.WriteTo(ctx, buffer); ignoreEOF(err) != nil {
```
levb: 0/5 Should we consider fanning this operation out, so that it is parallel with `dst.Write`? Then wait for both to complete?

packages/shared/pkg/storage/storage_cache_object.go
levb: 0/5 suggest 
```go
if c.flags.BoolFlag(ctx, featureflags.WriteToCacheOnWrites) {
	wg.Go(...
```

packages/shared/pkg/storage/storage_cache_object.go
```diff
@@ -36,132 +37,155 @@ func (c CachedObjectProvider) WriteTo(ctx context.Context, dst io.Writer) (n int
...
+	})
+
+	wg.Wait()
+
+	return count, err
 }
 
 func (c CachedObjectProvider) WriteFromFileSystem(ctx context.Context, path string) error {
+	go func(ctx context.Context) {
+		if !c.flags.BoolFlag(ctx, featureflags.WriteToCacheOnWrites) {
```
levb: ditto, 0/5 move the if above `go`?

packages/shared/pkg/utils/move.go
levb: 1/5 reflect this in the test please?

packages/shared/pkg/storage/storage_cache_object_test.go
levb: Jokes aside, maybe we can group together the various _read_ functions (Read, WriteTo, etc.) and call them "read tests" (vs write of course)

packages/shared/pkg/storage/storage_cache_seekable.go
levb: OT: do we have a general guideline on adding new configurable values? const vs FF vs ENV/config? How does one decide?

packages/shared/pkg/storage/storage_cache_seekable.go
levb: The user buffer doesn't belong to the goroutine, so technically we shouldn't use it there. IMO the bigger question is whether the saving to the file could be done concurrently with reading the data from the source, but given that ReadAt returns a big slice we can't do that.

Makefile
djeebus: Yeah, the concurrency was set too high if nothing was cached; golangci-lint already makes use of all the processors it can, so between them it'd saturate your cores and cause contention. The end result was that P=10 made my laptop unusable while running and was actually slower than P=4:

On my laptop:
- P=4    cached = 2.7s, uncached = 30 seconds
- P=10 cached = 1.7s, uncached = 49 seconds

packages/shared/pkg/feature-flags/flags.go
djeebus: They're more than just flag *names* though, since they also hold the default value. I'll change them to `Flag`, and rename the rest for consistency.

packages/shared/pkg/storage/storage_cache.go
djeebus: The point of the helper is to make it clear that the error has been handled (logged) and does not need to be handled by the parent function. Eventually I'd like to enable the linter that requires that all returned errors are either handled, returned, or logged; returning the error would break that paradigm.

As an example, right now we have things like `defer file.Close()` in a bunch of places. `Close` returns an error, but we're not handling it at all. We should be logging it, and `defer cleanup(ctx, "failed to close file", file.Close)` would take care of that. If we returned the error, we'd prevent the linter from knowing that it's been logged.

On a side note, as a rule I think errors should either be logged OR returned, but never both. Too often that leads to logs like:

```
error occurred
tried to do something: error occurred
failed to perform action: tried to do something: error occurred
```

which is just noise; all we really need is that last line.

I can move it to the utils package.

djeebus: :rofl: looks like this is the third one I've created in our codebase; I'll consolidate those as well.

packages/shared/pkg/feature-flags/flags.go
djeebus: Hmm, OTOH .. is `Flag` redundant? They'll always be referenced via the package, so it's something like `flags.TemplateFeatureFlagName`. Isn't `flags.TemplateFeature` enough?

packages/shared/pkg/storage/storage_cache_seekable.go
djeebus: It's only needed to mock for tests; doesn't affect anything outside, and if the interface changes, the compiler will complain loudly. Doesn't seem valuable to share; maybe counter productive.

packages/shared/pkg/storage/storage_cache_seekable.go
djeebus: This is on the hot path, but we're currently in production doing this w/o the buffer and it *seems* fine? I could remove this if we feel good about that.

packages/shared/pkg/storage/storage_cache_seekable.go
djeebus: I'm on board; I think this PR predates my awareness of that blog post :sweat_smile:

packages/shared/pkg/utils/move.go
djeebus: There were some issues with callers deleting the files themselves, and reporting errors when the file didn't exist. I think that was caused by the anem; it's not clear that `AtomicMove` should be deleting the file. I'll find a better name that makes it clear that the source will be deleted, regardless of success or failure.

packages/shared/pkg/storage/storage_cache_seekable.go
djeebus: @levb not that I know of. I think we're still feeling this one out.

packages/shared/pkg/feature-flags/flags.go
levb: @djeebus 👍 with me, shorter is better; as long as it's consistent with the rest of the code base.

packages/shared/pkg/storage/storage_cache_seekable.go
djeebus: @dobrac I don't see the code you're talking about. Is the linked file wrong? Did I already change it?

@briann Wrong PR? :confused:

packages/shared/pkg/storage/storage_cache_seekable.go
djeebus: error groups cancel the context and stop spawning on the first error. we want to make sure every chunk gets to try to be written, regardless of temporary failures.

packages/shared/pkg/storage/storage_cache_seekable.go
djeebus: Very likely! The offset doesn't change though, and it's why we ignore `io.EOF` in lots of places, often with the `ignoreEOF` helper function.

packages/shared/pkg/storage/storage_cache_seekable.go
djeebus: :facepalm:

packages/shared/pkg/storage/storage_cache_seekable.go
djeebus:
```mermaid
graph TB
	subgraph sandbox.TemplateBuild
		TemplateBuild.Upload["Upload"]
		TemplateBuild.uploadMemfile["uploadMemfile"]
		TemplateBuild.uploadMetadata["uploadMetadata"]
		TemplateBuild.uploadSnapfile["uploadSnapfile"]
		TemplateBuild.uploadRootfs["uploadRootfs"]
	end

	subgraph storage.CachedSeekableObjectProvider
		CachedSeekableObjectProvider.WriteFromFileSystem["WriteFromFileSystem"]
		CachedSeekableObjectProvider.createCacheBlocksFromFile["createCacheBlocksFromFile"]
		CachedSeekableObjectProvider.writeChunkFromFile["writeChunkFromFile"]
	end

	CachedSeekableObjectProvider.createCacheBlocksFromFile --> CachedSeekableObjectProvider.writeChunkFromFile
	CachedSeekableObjectProvider.WriteFromFileSystem --> CachedSeekableObjectProvider.createCacheBlocksFromFile

	TemplateBuild.uploadMemfile --> CachedSeekableObjectProvider.WriteFromFileSystem
	TemplateBuild.uploadMetadata --> CachedSeekableObjectProvider.WriteFromFileSystem
	TemplateBuild.uploadRootfs --> CachedSeekableObjectProvider.WriteFromFileSystem
	TemplateBuild.uploadSnapfile --> CachedSeekableObjectProvider.WriteFromFileSystem

	TemplateBuild.Upload --> TemplateBuild.uploadMemfile
	TemplateBuild.Upload --> TemplateBuild.uploadMetadata
	TemplateBuild.Upload --> TemplateBuild.uploadRootfs
	TemplateBuild.Upload --> TemplateBuild.uploadSnapfile
```

It's my understanding that, once these files are written to the disk, they are never changed.

packages/shared/pkg/storage/storage_cache_seekable.go
djeebus: Modifying the code formerly known as `AtomicMove` has removed all this code :fireworks:

packages/shared/pkg/storage/storage_cache_object.go
djeebus: This constant is only ever used for this very specific function, in this very specific time. What is gained by increasing its scope, other than making it less clear what it's used for?

I don't think there's value in making this configurable at run time; unless we foresee having a subset of customers with giant metadata file, this is probably best left compiled in.

packages/shared/pkg/storage/storage_cache_object.go
djeebus: IIRC this was to appease a code review comment bot (not a great reason, I admit!). I don't believe there's a strong reason to wait for the local cache write to finish before returning. Furthermore, I suspect that the vast majority of time the local cache will accept writes much faster than GCS, and I don't think there's a race condition here possible, for similar reasons as the graph above; it's only used to upload immutable data. Happy to remove this and tell the code review bot to take a walk.

packages/shared/pkg/storage/storage_cache_object_test.go
djeebus: Haha, yeah, I've been wanting to do that since I got here. I can work on that after we get this PR merged in.

packages/shared/pkg/storage/storage_cache_object.go
djeebus: I wrote it this way to avoid a race condition dealing with buffers and offsets, but your comment made me look at it again, and I found a better way. It is now parallel!

Edit: I don't think there's a reason to make it parallel, the cache will only be used the *next* time that data is requested and the caller isn't interested in whether the cache write succeeded or not.

packages/shared/pkg/storage/storage_cache_object.go
```diff
@@ -24,144 +26,195 @@ type CachedObjectProvider struct {
...
+			count, err := c.writeFileToCache(
+				context.WithoutCancel(ctx),
+				bytes.NewReader(p),
+			)
+			if err != nil {
+				recordCacheWriteError(ctx, cacheTypeObject, cacheOpWrite, err)
+			} else {
+				recordCacheWrite(ctx, count, cacheTypeObject, cacheOpWrite)
+			}
+		})
```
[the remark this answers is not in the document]
djeebus: This function doesn't need to worry about callers mutating the data; it is considered immutable.

levb (approved):
LGTM from me, but @dobrac may have remaining issues?

packages/shared/pkg/storage/storage_cache_seekable.go
dobrac: it influences also the type in the code paths - not only in tests. We've already this in another tests, maybe we can unite?

```
func flagsWithMaxBuildCachePercentage(tb testing.TB, maxBuildCachePercentage int) *featureflags.Client {
	tb.Helper()

	flags, err := featureflags.NewClient()
	require.NoError(tb, err)

	tb.Cleanup(func() {
		defer func() {
			err := flags.Close(tb.Context())
			assert.NoError(tb, err)
		}()
	})

	featureflags.LaunchDarklyOfflineStore.Update(
		featureflags.LaunchDarklyOfflineStore.Flag(featureflags.BuildCacheMaxUsagePercentage.String()).ValueForAll(ldvalue.Int(maxBuildCachePercentage)),
	)

	return flags
}
```

packages/shared/pkg/storage/storage_cache_seekable.go
dobrac: let's leave it as that should be the correct code flow

packages/shared/pkg/storage/header/diff.go
```diff
@@ -24,15 +22,6 @@ var (
...
 )
 
-func WriteDiffWithTrace(ctx context.Context, source io.ReaderAt, blockSize int64, dirty *bitset.BitSet, diff io.Writer) (*DiffMetadata, error) {
-	_, childSpan := tracer.Start(ctx, "create-diff")
-	defer childSpan.End()
-	childSpan.SetAttributes(attribute.Int64("dirty.length", int64(dirty.Count())))
-	childSpan.SetAttributes(attribute.Int64("block.size", blockSize))
-
-	return writeDiff(source, blockSize, dirty, diff)
-}
```
dobrac, line 34: if doing cleanup, lets remove also the writeDiff function

packages/shared/pkg/storage/storage_cache_seekable.go
dobrac: Makes sense, can you please update the comment that there is no concurrent upload of the build layer rather than template? The same code path is called also in snapshot creation

packages/shared/pkg/storage/offset_reader.go
dobrac, line 1: nit, being this file just offset_reader is probably ok imo

packages/shared/pkg/storage/storage_cache_object.go
```diff
@@ -71,82 +73,96 @@ func (c CachedObjectProvider) WriteTo(ctx context.Context, dst io.Writer) (n int
...
-	}(context.WithoutCancel(ctx))
+			recordCacheWrite(ctx, count, cacheTypeObject, cacheOpWriteFromFileSystem)
+		})
+	}
 
 	return c.inner.WriteFromFileSystem(ctx, path)
 }
 
-func (c CachedObjectProvider) fullFilename() string {
+func (c *CachedObjectProvider) goCtx(ctx context.Context, fn func(context.Context)) {
```
dobrac: nit: can we include that it runs `WithoutCancel` in the name?

packages/shared/pkg/storage/storage_cache_seekable.go
```diff
@@ -168,18 +268,16 @@ func (c CachedSeekableObjectProvider) validateReadAtParams(buffSize, offset int6
...
-			// Another process is already writing this chunk, so we can skip writing it ourselves
-			return
-		}
-
-		logger.L().Warn(ctx, "failed to acquire lock", zap.String("path", chunkPath), zap.Error(err))
+		// failed to acquire lock, which is a different category of failure than "write failed"
+		recordCacheWriteError(ctx, cacheTypeSeekable, cacheOpReadAt, err)
 
-		return
+		return nil
```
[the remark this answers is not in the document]
djeebus: intentional

packages/shared/pkg/storage/storage_cache_seekable.go
```diff
@@ -193,51 +291,30 @@ func (c CachedSeekableObjectProvider) writeChunkToCache(ctx context.Context, off
...
+	}
 
-		return
+	if err := utils.RenameOrDeleteFile(ctx, tempPath, chunkPath); err != nil {
+		return fmt.Errorf("failed to rename temp file: %w", err)
 	}
 
 	writeTimer.End(ctx, int64(len(bytes)))
+
+	return nil
```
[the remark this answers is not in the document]
djeebus: intentional.

packages/shared/pkg/storage/gcp_multipart.go
```diff
@@ -157,10 +162,10 @@ func NewMultipartUploaderWithRetryConfig(ctx context.Context, bucketName, object
 	}, nil
 }
 
-func (m *MultipartUploader) InitiateUpload() (string, error) {
+func (m *MultipartUploader) InitiateUpload(ctx context.Context) (string, error) {
```
dobrac: can this be a separate PR please? (isn't it already here: https://github.com/e2b-dev/infra/pull/1671 ?)

packages/shared/pkg/storage/storage_cache_metrics.go
```diff
@@ -43,6 +44,14 @@ const (
 )
 
 func recordCacheRead(ctx context.Context, isHit bool, bytesRead int64, t cacheType, op cacheOp) {
+	span := trace.SpanFromContext(ctx)
+	span.SetAttributes(
```
dobrac, line 48: are you sure this isn't significantly slowing down the performance? (applies to most other added tracers, especially ones with attributes)

packages/shared/scripts/package.json
```diff
@@ -2,6 +2,6 @@
   "name": "base-template",
   "version": "1.0.0",
   "dependencies": {
-    "e2b": "^2.1.3"
+    "e2b": "^2.9.0"
```
dobrac, line 5: a bit unrelated change too

iac/provider-gcp/nomad/jobs/orchestrator.hcl
```diff
@@ -76,6 +76,7 @@ EOT
         OTEL_TRACING_PRINT           = "${otel_tracing_print}"
         LOGS_COLLECTOR_ADDRESS       = "${logs_collector_address}"
         ENVIRONMENT                  = "${environment}"
+        DOMAIN_NAME                  = "${domain_name}"
```
dobrac, line 79: the environment is already used for the same (similar?) purpose, could we use that instead?

packages/shared/scripts/package.json
djeebus: I can roll this back, but is there any reason to, other than purity? Is there any reason *not* to ensure that this script is run via the latest sdk?

packages/shared/pkg/storage/storage_cache_metrics.go
djeebus: Good point. I'll hide this behind a launchdarkly flag so we can enable it when we want to. It's been often useful, I'd like to keep it in some how, but there's no reason to do it for *every request*

packages/shared/pkg/storage/gcp_multipart.go
djeebus: It's already there, I'm waiting for that one to merge so we can merge this one.

iac/provider-gcp/nomad/jobs/orchestrator.hcl
djeebus: I don't want to have to create new environments for every single dev sandbox and deploy, and right now testing launch darkly flags is a little scary, as we have no way of saying "do this just on my cluster". This gives us that ability without having to manually do anything; it just works.

packages/shared/pkg/storage/gcp_multipart.go
jakubno: Merged 🚀

packages/shared/scripts/package.json
dobrac: > I can roll this back, but is there any reason to, other than purity?

nono, just that

> Is there any reason not to ensure that this script is run via the latest sdk?

NPM has a bug where if you publish a new version, the latest points to it, but the new version doest work for some time

iac/provider-gcp/nomad/jobs/orchestrator.hcl
dobrac: We use environments for this in production, you can create a new environment for you too, but I'm fine with keeping this

packages/shared/pkg/storage/lock/atomic_file_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+		f, err := OpenFile(t.Context(), filename)
+		require.NoError(t, err)
+		require.NotNil(t, f)
+
+		count, err := f.Write(expected)
+		require.NoError(t, err)
+		assert.Equal(t, len(expected), count)
+
+		_, err = os.Stat("test.bin")
```
jakubno:
```suggestion
		_, err = os.Stat(filename)
```

packages/shared/pkg/storage/lock/atomic_file_test.go
```diff
@@ -0,0 +1,102 @@
...
```
jakubno:
```suggestion
		_, err = os.Stat(filename)
```

packages/shared/pkg/storage/storage_cache.go
```diff
@@ -9,51 +9,59 @@ import (
...
+	rootPath string,
+	inner StorageProvider,
+	flags *featureflags.Client,
+) *CachedProvider {
+	cacheTracer := tracer
+
+	createCacheSpans := flags.BoolFlag(ctx, featureflags.CreateStorageCacheSpansFlag)
+	if !createCacheSpans {
+		cacheTracer = noop.NewTracerProvider().Tracer("github.com/e2b-dev/infra/packages/shared/pkg/storage")
+	}
```
jakubno, line 47: wouldn't it make sense to do it in `OpenObject`? Otherwise it will be set during start of the service and we can't change it without restarting

packages/shared/pkg/storage/storage_cache.go
```diff
@@ -108,23 +132,11 @@ func ignoreEOF(err error) error {
...
-	}()
-
-	if err := os.Link(oldPath, newPath); err != nil {
-		if errors.Is(err, os.ErrExist) {
-			// Someone else created newPath first. Treat as success.
-			return nil
-		}
-
-		return err
+func recordError(span trace.Span, err error) {
```
jakubno: i think this could be in a separate file

packages/shared/pkg/storage/storage.go
```diff
@@ -72,14 +72,18 @@ type WriterToCtx interface {
 	WriteTo(ctx context.Context, w io.Writer) (n int64, err error)
 }
 
+type WriteFromFSCtx interface {
```
jakubno: I would find `WriteFromFileSystemCtx` easier to read and as we don't need it in the code only in interfaces, I think doesn't matter that much that it's long

packages/shared/pkg/utils/errorcollector.go
```diff
@@ -0,0 +1,68 @@
...
+	"sync"
+)
+
+// ErrorCollector collects errors from multiple goroutines. This has a similar API to `errgroup.Group`,
+// except that it runs all goroutines, and returns all errors; it does not stop after the first error.
+type ErrorCollector struct {
+	done    chan struct{}
+	input   chan error
+	errs    []error
+	workers chan struct{}
```
jakubno: why not to use `semaphore.Weighted`?

packages/shared/pkg/storage/storage_cache_object.go
```diff
@@ -24,144 +25,192 @@ type CachedObjectProvider struct {
...
 
-	if _, err := c.inner.WriteTo(ctx, writer); ignoreEOF(err) != nil {
-		return 0, err
+	if c.flags.BoolFlag(ctx, featureflags.EnableWriteThroughCacheFlag) {
+		c.goCtxWithoutCancel(ctx, func(ctx context.Context) {
+			ctx, span := c.tracer.Start(ctx, "write data to cache")
+			defer span.End()
+
+			count, err := c.writeFileToCache(
+				context.WithoutCancel(ctx),
```
jakubno: do you need without cancel here as there's `goCtxWithoutCancel`

packages/shared/pkg/storage/storage_cache_object.go
```diff
@@ -24,144 +25,192 @@ type CachedObjectProvider struct {
 	path      string
 	chunkSize int64
 	inner     ObjectProvider
+	flags     featureFlagsClient
+	tracer    trace.Tracer
+
+	wg sync.WaitGroup
```
jakubno, line 31: wouldn't it make sense to wait somewhere (e.g. when the template is shutting down)
I am fine with doing this in separate PR though

packages/shared/pkg/storage/storage_cache_object.go
```diff
@@ -24,144 +25,192 @@ type CachedObjectProvider struct {
...
-	// This is semi-arbitrary. this code path is called for files that tend to be less than 1 MB (headers, metadata, etc),
-	// so 2 MB allows us to read the file without needing to allocate more memory, with some room for growth. If the
-	// file is larger than 2 MB, the buffer will grow, it just won't be as efficient WRT memory allocations.
-	const writeToInitialBufferSize = 2 * megabyte
-
-	writer := bytes.NewBuffer(make([]byte, 0, writeToInitialBufferSize))
+// Write pushes data to the wrapped object provider, and optionally pushes the data to a fast ephemeral cache as well.
+// `p` is considered immutable, and won't change if we access it after the function returns.
+func (c *CachedObjectProvider) Write(ctx context.Context, p []byte) (n int, e error) {
+	ctx, span := c.tracer.Start(ctx, "write data to object")
```
jakubno:
```suggestion
	ctx, span := c.tracer.Start(ctx, "write data to object storage")
```

packages/shared/pkg/storage/storage_cache_object.go
```diff
@@ -24,144 +25,192 @@ type CachedObjectProvider struct {
...
+		})
 	}
 
-	go func() {
-		c.writeFullFileToCache(context.WithoutCancel(ctx), writer.Bytes())
+	return c.inner.Write(ctx, p)
+}
+
+func (c *CachedObjectProvider) WriteFromFileSystem(ctx context.Context, path string) (e error) {
+	ctx, span := c.tracer.Start(ctx, "write from filesystem to object")
```
jakubno:
```suggestion
	ctx, span := c.tracer.Start(ctx, "write from filesystem to object storage")
```

packages/shared/pkg/storage/storage_cache_seekable.go
```diff
@@ -193,51 +295,34 @@ func (c CachedSeekableObjectProvider) writeChunkToCache(ctx context.Context, off
...
+func (c *CachedSeekableObjectProvider) writeLocalSize(ctx context.Context, size int64) error {
 	finalFilename := c.sizeFilename()
 
 	// Try to acquire lock for this chunk write to NFS cache
 	lockFile, err := lock.TryAcquireLock(ctx, finalFilename)
 	if err != nil {
-		if errors.Is(err, lock.ErrLockAlreadyHeld) {
-			// Another process is already writing this chunk, so we can skip writing it ourselves
-			return
-		}
```
jakubno, line 236: do we want to record this as an error? When a new template is build it "spam" errors, which are kinda expected

jakubno (changes requested):
few nits, otherwise lgtm

packages/shared/pkg/storage/storage_cache_object.go
dobrac: it should be actually {verb}_{object} 👀

packages/shared/pkg/storage/lock/atomic_file_test.go
djeebus: :angry: AI powered autocomplete is so quick! so easy! so sloppy!

packages/shared/pkg/storage/storage_cache.go
djeebus: We create a new provider every time we spawn a sandbox, and every time we build a template. Placing it here keeps it consistent for each request - it either uses cache or it doesn't, it can't change mid-execution.

packages/shared/pkg/storage/storage_cache_object.go
djeebus: It does make sense, but this would be a larger refactor, since the providers don't currently have any kind of `Close()` functionality. Worth doing though!

packages/shared/pkg/storage/storage_cache_object.go
```diff
@@ -24,144 +25,189 @@ type CachedObjectProvider struct {
...
+			defer span.End()
+
+			count, err := c.writeFileToCache(ctx, bytes.NewReader(p))
+			if err != nil {
+				recordError(span, err)
+				recordCacheWriteError(ctx, cacheTypeObject, cacheOpWrite, err)
+			} else {
+				recordCacheWrite(ctx, count, cacheTypeObject, cacheOpWrite)
+			}
+		})
```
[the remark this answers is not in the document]
djeebus: Currently doesn't happen, and it's an intentional choice to speed up writes. Not an issue.

PR #1134, opened 2025-09-05
Stop passing around a tracer

djeebus:
Now that we're setting the global tracer, we don't need to pass around the instance of one. This allows us to see what package a span was created in, as well as simplifying how we add traces.

### DIFF
--- a/packages/orchestrator/cmd/build-template/main.go
+++ b/packages/orchestrator/cmd/build-template/main.go
@@ -9,7 +9,6 @@ import (
 	"net/http"
 	"time"
 
-	"go.opentelemetry.io/otel"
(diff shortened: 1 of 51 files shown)


sitole:
Additionally, after reviewing some OTEL examples in Golang, I must say it makes sense to use a per-package tracer.

I also came across a discussion that, for testing purposes, they are defining tracer per package, but using from-struct so you can redefine it for unit tests later (https://github.com/open-telemetry/opentelemetry-go/discussions/4532#discussioncomment-7050813). But it's just curiosity about how to use both global and easily adjustable for unit tests, not something we need to use.

jakubno (approved):

PR #1144, opened 2025-09-10
Add process monitoring function with handlers 

0div:
# Description 
Implement a separate process monitoring background routine (i.e. not the one that tracks commands received via GRPC), by leveraging `gopsutil` to track and define how to handle new processes as they are running and exiting. This is the first step in sending process info to the events endpoint in the context of the internal sandbox events project. 

# Test

## Go test
```
[e2b-dev/infra/packages/envd] go test ./...
```

## Manual test
In `envd`'s `main.go`: 
```go
go host.MonitorProcesses(1*time.Second, func(processInfo *host.ProcessInfo) error {
	fmt.Printf("[HANDLE PROCESS] %+v\n", processInfo)
	return nil
})
```

```bash
[e2b-dev/infra/packages/envd] $ make build && make start-docker
 
 [...]
 
[HANDLE PROCESS] &{State:exited PID:92 Name:sleep Cmdline:/run/rosetta/rosetta /usr/bin/sleep sleep 10 CreateTime:1757462956600}
[HANDLE PROCESS] &{State:exited PID:88 Name:sleep Cmdline:/run/rosetta/rosetta /usr/bin/sleep sleep 10 CreateTime:1757462956420}
[HANDLE PROCESS] &{State:exited PID:91 Name:sleep Cmdline:/run/rosetta/rosetta /usr/bin/sleep sleep 10 CreateTime:1757462956590}
[HANDLE PROCESS] &{State:running PID:102 Name:bash Cmdline:/bin/bash CreateTime:1757463018860}
[HANDLE PROCESS] &{State:running PID:108 Name:sleep Cmdline:sleep 1 CreateTime:1757463032820}
[HANDLE PROCESS] &{State:exited PID:108 Name:sleep Cmdline:sleep 1 CreateTime:1757463032820}

$ docker exec -it ${ENDV_CONTAINER} /bin/bash
root@dda5af440389 sleep 1
````

### DIFF
--- a/packages/envd/internal/host/process_test.go
+++ b/packages/envd/internal/host/process_test.go
@@ -80,6 +80,6 @@ func TestMonitorProcesses(t *testing.T) {
 		// Test completed successfully
 		return
 	case <-time.After(15 * time.Second):
-		t.Fatal("Test took longer than 5 seconds to complete")
+		t.Fatal("Test took longer than 15 seconds to complete")
 	}
 }

### CODE REVIEW

File: packages/envd/internal/host/process.go [hunk 1]
```diff
@@ -0,0 +1,130 @@
...
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			currentPids, err := process.Pids()
+			if err != nil {
+				logger.Error().Err(err).Msg("Error getting current processes")
+				continue
+			}
```
sitole, line 88: Just a quick note:

Can we watch for processes spawned by envd (run, command methods committed from our SDK)? I think monitoring all PIDs, except for some initial ones, will not work as systemd / other daemons can periodically spawn and kill processes, so current pids will be blown by some irrelevant processes.

0div: I thought of doing that initially but it wouldn't account for the processes spawned outside of it, one of the goals being to detect anomalous behavior.

PR #1145, opened 2025-09-10
Refactor access tokens to sqlc

jakubno:
Refactor queries to SQLC


packages/db/queries/get_api_keys.sql
```diff
@@ -0,0 +1,15 @@
+-- name: GetTeamAPIKeysWithCreator :many
+SELECT
+    tak.id,
```
sitole, line 3: Image

packages/db/queries/get_api_keys.sql
jakubno, line 1: After the migration, we can make this a classic embed

packages/db/queries/create_access_token.sql
jakubno, line 1: after migration we can use `slqc.embed` for `RETURNING`

sitole (approved):

PR #1147, opened 2025-09-10
Fix for sandboxes with Ubuntu 24.04 source images

sitole:
Regression after [merging changes with Chrony removal](https://github.com/e2b-dev/infra/pull/1114) caused failing builds with Ubuntu 24.04 source images. There are some underlying changes in conditions for showing the firstboot wizard when the VM starts with fresh memory. It may be triggered by not correctly setting up the date/time/tz because of the missing Chrony service. It should be okay, because after that time is synced, and the finished build has the correct clock here.

This is causing the VM to be stuck on the firstboot wizard and envd wait timeouts, resulting in the VM launch in build failing.
Older versions we tested (E2B base image late, Ubuntu 22.04) worked without issues. This change should be safe to apply, as the first boot should never be used in our scenario.


sitole:
> Please add also tests for this

Tests added

jakubno (approved):

PR #1148, opened 2025-09-10
Skip evict if the sandbox is already being removed

jakubno:
Don't try to evict, if the sandbox is already being removed (it was only printing unnecessary errors)

### DIFF
--- a/packages/api/internal/orchestrator/evictor/evict.go
+++ b/packages/api/internal/orchestrator/evictor/evict.go
@@ -29,7 +29,7 @@ func (e *Evictor) Start(ctx context.Context) {
 			items := e.store.ExpiredItems()
 
 			for _, item := range items {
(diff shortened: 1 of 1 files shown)


packages/api/internal/orchestrator/evictor/evict.go
```diff
@@ -29,7 +29,7 @@ func (e *Evictor) Start(ctx context.Context) {
 			items := e.store.ExpiredItems()
 
 			for _, item := range items {
-				if item.IsExpired() {
+				if item.IsExpired() && item.GetState() == instance.StateRunning {
```
sitole, line 32: Can we maybe make it more sustainable/easy to search for and add `isRunning()`?

jakubno: I will improve this in following PRs

PR #1152, opened 2025-09-10
Nomad service discovery for Edge API

sitole:
- Implement Nomad service discovery
- Use Nomad service discovery for edge API
- Short service name env name prefixes
- (cleanup) Remove no longer used AWS service discovery


sitole:
The issue with the previous DNS resolution used for Consul DNS was that when Nomad jobs were rolling, it stopped not being output in DNS results. We still needed to receive it, as the rolling template manager was still running a build, and we required the ability to contact the service for future information.

sitole:
Tested on a case with multiple running template managers, one long build is running, and rollout is done in the meantime. The template manager with the build was propagated correctly to the API, and the build finished successfully. After that, the template manager rollout finished successfully.

PR #1154, opened 2025-09-10
Add some metadata to GCS errors

djeebus:


packages/shared/pkg/storage/storage_google.go
```diff
@@ -163,10 +167,10 @@ func (g *GCPBucketStorageObjectProvider) Size(ctx context.Context) (int64, error
 	attrs, err := g.handle.Attrs(ctx)
 	if err != nil {
 		if errors.Is(err, storage.ErrObjectNotExist) {
-			return 0, ErrObjectNotExist
+			err = ErrObjectNotExist // use ours instead of theirs
```
jakubno: is there a reason why not to return here? I think we're creating option for introducing bug later

PR #1155, opened 2025-09-10
Cache size calls

djeebus:


packages/shared/pkg/storage/storage_cache.go
```diff
@@ -328,25 +349,75 @@ func (c *CachedFileObjectProvider) readAtFromCache(chunkPath string, buff []byte
 	return count, err // return `err` in case it's io.EOF
 }
 
-func (c *CachedFileObjectProvider) copyFullFileFromCache(filePath string, buff []byte) (int64, error) {
+func (c *CachedFileObjectProvider) copyFullFileFromCache(dst io.Writer) (int64, bool) {
+	path := c.makeFullFilename()
```
jakubno: can this be just `fullFilename`? it isn't making anything

packages/shared/pkg/storage/storage_cache.go
```diff
@@ -220,10 +177,74 @@ func (c *CachedFileObjectProvider) validateReadAtParams(buffSize, offset int64)
 	return nil
 }
 
+func (c *CachedFileObjectProvider) makeSizeFilename() string {
```
jakubno:
```suggestion
func (c *CachedFileObjectProvider) sizeFilename() string {
```

packages/shared/pkg/storage/storage_cache.go
```diff
@@ -328,25 +349,75 @@ func (c *CachedFileObjectProvider) readAtFromCache(chunkPath string, buff []byte
...
+	tempWriter, err := os.Create(tempFname)
 	if err != nil {
-		return 0, fmt.Errorf("failed to read from chunk: %w", err)
+		zap.L().Error("failed to create temp full file, bailing on cache",
+			zap.Error(err),
+			zap.String("path", tempFname))
+		return c.inner.WriteTo(ctx, dst)
+	}
+
+	// write to both cache and destination simultaneously
```
jakubno: this will now slowdown the read, before it was written to the cache async

packages/shared/pkg/storage/storage_cache.go
```diff
@@ -220,23 +228,46 @@ func (c *CachedFileObjectProvider) validateReadAtParams(buffSize, offset int64)
...
+func (c *CachedFileObjectProvider) sizeFilename() string {
+	return filepath.Join(c.path, "size.txt")
 }
 
-func (c *CachedFileObjectProvider) Delete(ctx context.Context) error {
-	return c.inner.Delete(ctx)
+func (c *CachedFileObjectProvider) writeLocalSize(size int64) {
+	tempFilename := filepath.Join(c.path, fmt.Sprintf(".size.bin.%s", uuid.NewString()))
+
+	fp, err := os.OpenFile(tempFilename, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, cacheFilePermissions)
```
jakubno: why not to use `WriteFile`?

packages/shared/pkg/storage/storage_cache.go
```diff
@@ -328,25 +360,60 @@ func (c *CachedFileObjectProvider) readAtFromCache(chunkPath string, buff []byte
...
+	return count, true
+}
+
+const (
+	kilobyte = 1024
+	megabyte = 1024 * kilobyte
+)
+
+func (c *CachedFileObjectProvider) readAndCacheFullRemoteFile(ctx context.Context, dst io.Writer) (int64, error) {
+	const writeToInitialBufferSize = 2 * megabyte
```
jakubno, line 393: could you add a comment explaining this decision?

packages/shared/pkg/storage/storage_cache.go
```diff
@@ -204,6 +169,49 @@ var (
...
+	if err != nil {
+		zap.L().Warn("failed to read cached size, falling back to remote read",
+			zap.String("path", fname),
+			zap.Error(err))
+		return 0, false
+	}
+
+	size, err := strconv.ParseInt(string(content), 10, 64)
+	if err != nil {
+		zap.L().Warn("failed to parse cached size, falling back to remote read",
```
jakubno: this shouldn't happen right? it probably should an error

jakubno (changes requested):
looks good, some last nits.
Do you have some comparison, results of benchmark how this speeds up the starts?

packages/shared/pkg/storage/storage_cache.go
djeebus: IMHO we use warnings when something bad happens that the program can recover gracefully from, and an error is when the program cannot recover. In this case, failing to read the cached size is fine b/c we can fall back to the remote size. If you feel strongly though, I can change it.

jakubno: Not sure if we have any framework for that, my heuristics here: 
- errors as some unexpected state
- warnings are expected, but they shouldn't happen by default

PR #1159, opened 2025-09-11
API - Sandbox locking

jakubno:
This pull request refactors the sandbox store code to improve thread safety, encapsulate instance data, and simplify state management.

### DIFF
--- a/packages/api/internal/cache/instance/instance.go
+++ b/packages/api/internal/cache/instance/instance.go
@@ -33,7 +33,7 @@ var allowed = map[State]map[State]bool{
 	StatePaused:  {StateKilled: true},
 }
 
-func NewInstanceInfo(
+func NewSandbox(
 	sandboxID string,
 	templateID string,
 	clientID string,
@@ -57,39 +57,34 @@ func NewInstanceInfo(
 	envdAccessToken *string,
 	allowInternetAccess *bool,
 	baseTemplateID string,
-) *InstanceInfo {
-	instance := &InstanceInfo{
-		data: Data{
-			SandboxID:  sandboxID,
-			TemplateID: templateID,
-			ClientID:   clientID,
-			Alias:      alias,
+) Data {
+	return Data{
+		SandboxID:  sandboxID,
+		TemplateID: templateID,
+		ClientID:   clientID,
+		Alias:      alias,
 
-			ExecutionID:         executionID,
-			TeamID:              teamID,
-			BuildID:             buildID,
-			Metadata:            metadata,
-			MaxInstanceLength:   maxInstanceLength,
-			StartTime:           startTime,
-			EndTime:             endTime,
-			VCpu:                vcpu,
-			TotalDiskSizeMB:     totalDiskSizeMB,
-			RamMB:               ramMB,
-			KernelVersion:       kernelVersion,
-			FirecrackerVersion:  firecrackerVersion,
-			EnvdVersion:         envdVersion,
-			EnvdAccessToken:     envdAccessToken,
-			AllowInternetAccess: allowInternetAccess,
-			NodeID:              nodeID,
-			ClusterID:           clusterID,
-			AutoPause:           autoPause,
-			State:               StateRunning,
-			BaseTemplateID:      baseTemplateID,
-		},
-		mu: sync.RWMutex{},
+		ExecutionID:         executionID,
+		TeamID:              teamID,
+		BuildID:             buildID,
+		Metadata:            metadata,
+		MaxInstanceLength:   maxInstanceLength,
+		StartTime:           startTime,
+		EndTime:             endTime,
+		VCpu:                vcpu,
+		TotalDiskSizeMB:     totalDiskSizeMB,
+		RamMB:               ramMB,
+		KernelVersion:       kernelVersion,
+		FirecrackerVersion:  firecrackerVersion,
+		EnvdVersion:         envdVersion,
+		EnvdAccessToken:     envdAccessToken,
+		AllowInternetAccess: allowInternetAccess,
+		NodeID:              nodeID,
+		ClusterID:           clusterID,
+		AutoPause:           autoPause,
+		State:               StateRunning,
+		BaseTemplateID:      baseTemplateID,
 	}
-
-	return instance
 }
 
 type Data struct {
--- a/packages/api/internal/cache/instance/instance.go
+++ b/packages/api/internal/cache/instance/instance.go
@@ -124,7 +124,7 @@ type Data struct {
 type InstanceInfo struct {
 	data Data
 
-	transition *utils.SetOnce[struct{}]
+	transition *utils.ErrorOnce
 	mu         sync.RWMutex
 }
 
@@ -190,7 +190,7 @@ func (i *InstanceInfo) StartRemoving(ctx context.Context, stateAction StateActio
 		}
 
 		zap.L().Debug("State transition already in progress to the same state, waiting", logger.WithSandboxID(i.data.SandboxID), zap.String("state", string(newState)))
-		_, err = transition.WaitWithContext(ctx)
+		err = transition.WaitWithContext(ctx)
 		if err != nil {
 			zap.L().Error("State transition failed", logger.WithSandboxID(i.data.SandboxID), zap.String("state", string(newState)), zap.Error(err))
 			return false, nil, fmt.Errorf("sandbox is in failed state: %w", err)
@@ -218,14 +218,14 @@ func (i *InstanceInfo) StartRemoving(ctx context.Context, stateAction StateActio
 	}
 
 	i.data.State = newState
-	i.transition = utils.NewSetOnce[struct{}]()
+	i.transition = utils.NewErrorOnce()
 
 	callback = func(err error) {
 		zap.L().Debug("Transition complete", logger.WithSandboxID(i.data.SandboxID), zap.String("state", string(newState)), zap.Error(err))
 		i.mu.Lock()
 		defer i.mu.Unlock()
 
-		setErr := i.transition.SetResult(struct{}{}, err)
+		setErr := i.transition.SetError(err)
 		if err != nil {
 			// Keep the transition in place so the error stays
 			zap.L().Error("Failed to set transition result", logger.WithSandboxID(i.data.SandboxID), zap.Error(setErr))
@@ -248,6 +248,5 @@ func (i *InstanceInfo) WaitForStateChange(ctx context.Context) error {
 		return nil
 	}
 
-	_, err := transition.WaitWithContext(ctx)
-	return err
+	return transition.WaitWithContext(ctx)
 }
--- a/packages/api/internal/cache/instance/instance.go
+++ b/packages/api/internal/cache/instance/instance.go
@@ -2,7 +2,6 @@ package instance
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"sync"
 	"time"
@@ -10,8 +9,8 @@ import (
 	"github.com/google/uuid"
 	"go.uber.org/zap"
 
+	"github.com/e2b-dev/infra/packages/shared/pkg/logger"
 	sbxlogger "github.com/e2b-dev/infra/packages/shared/pkg/logger/sandbox"
-	"github.com/e2b-dev/infra/packages/shared/pkg/utils"
 )
 
 const (
@@ -24,11 +23,15 @@ type State string
 
 const (
 	StateRunning State = "running"
-	StatePausing State = "pausing"
-	StateKilling State = "killing"
 	StatePaused  State = "paused"
 	StateKilled  State = "killed"
+	StateFailed  State = "failed"
 )
+
+var allowed = map[State]map[State]bool{
+	StateRunning: {StatePaused: true, StateKilled: true, StateFailed: true},
+	StatePaused:  {StateKilled: true},
+}
 
 func NewInstanceInfo(
 	sandboxID string,
@@ -56,39 +59,40 @@ func NewInstanceInfo(
 	baseTemplateID string,
 ) *InstanceInfo {
 	instance := &InstanceInfo{
-		SandboxID:  sandboxID,
-		TemplateID: templateID,
-		ClientID:   clientID,
-		Alias:      alias,
+		data: Data{
+			SandboxID:  sandboxID,
+			TemplateID: templateID,
+			ClientID:   clientID,
+			Alias:      alias,
 
-		ExecutionID:         executionID,
-		TeamID:              teamID,
-		BuildID:             buildID,
-		Metadata:            metadata,
-		MaxInstanceLength:   maxInstanceLength,
-		StartTime:           startTime,
-		endTime:             endTime,
-		VCpu:                vcpu,
-		TotalDiskSizeMB:     totalDiskSizeMB,
-		RamMB:               ramMB,
-		KernelVersion:       kernelVersion,
-		FirecrackerVersion:  firecrackerVersion,
-		EnvdVersion:         envdVersion,
-		EnvdAccessToken:     envdAccessToken,
-		AllowInternetAccess: allowInternetAccess,
-		NodeID:              nodeID,
-		ClusterID:           clusterID,
-		AutoPause:           autoPause,
-		state:               StateRunning,
-		stopping:            utils.NewSetOnce[struct{}](),
-		BaseTemplateID:      baseTemplateID,
-		mu:                  sync.RWMutex{},
+			ExecutionID:         executionID,
+			TeamID:              teamID,
+			BuildID:             buildID,
+			Metadata:            metadata,
+			MaxInstanceLength:   maxInstanceLength,
+			StartTime:           startTime,
+			EndTime:             endTime,
+			VCpu:                vcpu,
+			TotalDiskSizeMB:     totalDiskSizeMB,
+			RamMB:               ramMB,
+			KernelVersion:       kernelVersion,
+			FirecrackerVersion:  firecrackerVersion,
+			EnvdVersion:         envdVersion,
+			EnvdAccessToken:     envdAccessToken,
+			AllowInternetAccess: allowInternetAccess,
+			NodeID:              nodeID,
+			ClusterID:           clusterID,
+			AutoPause:           autoPause,
+			State:               StateRunning,
+			BaseTemplateID:      baseTemplateID,
+		},
+		dataMu: sync.RWMutex{},
 	}
 
 	return instance
 }
 
-type InstanceInfo struct {
+type Data struct {
 	SandboxID  string
 	TemplateID string
 	ClientID   string
@@ -101,7 +105,7 @@ type InstanceInfo struct {
 	Metadata            map[string]string
 	MaxInstanceLength   time.Duration
 	StartTime           time.Time
-	endTime             time.Time
+	EndTime             time.Time
 	VCpu                int64
 	TotalDiskSizeMB     int64
 	RamMB               int64
@@ -114,13 +118,18 @@ type InstanceInfo struct {
 	ClusterID           uuid.UUID
 	AutoPause           bool
 
-	state State
-	mu    sync.RWMutex
-
-	stopping *utils.SetOnce[struct{}]
+	State  State
+	Reason error
 }
 
-func (i *InstanceInfo) LoggerMetadata() sbxlogger.SandboxMetadata {
+type InstanceInfo struct {
+	data Data
+
+	transition chan error
+	dataMu     sync.RWMutex
+}
+
+func (i Data) LoggerMetadata() sbxlogger.SandboxMetadata {
 	return sbxlogger.SandboxMetadata{
 		SandboxID:  i.SandboxID,
 		TemplateID: i.TemplateID,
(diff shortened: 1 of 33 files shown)


packages/api/internal/cache/instance/operations.go
```diff
@@ -19,52 +19,33 @@ const (
...
+
+	sbxlogger.I(sandbox.data).Debug("Adding sandbox to cache",
 		zap.Bool("newly_created", newlyCreated),
-		zap.Time("start_time", sandbox.StartTime),
-		zap.Time("end_time", sandbox.GetEndTime()),
+		zap.Time("start_time", sandbox.data.StartTime),
+		zap.Time("end_time", sandbox.data.EndTime),
 	)
 
-	if sandbox.SandboxID == "" {
```
jakubno, line 29: this isn't really needed, we create the instance via `NewInstanceInfo`, so the fields are always filled

packages/api/internal/cache/instance/sync.go
```diff
@@ -27,65 +26,62 @@ func getMaxAllowedTTL(now time.Time, startTime time.Time, duration, maxInstanceL
...
+	instance.mu.Lock()
+	defer instance.mu.Unlock()
+
+	endTime := instance.data.EndTime
 	if !allowShorter && endTime.After(now.Add(duration)) {
-		return instance, nil
+		return instance.data, nil
 	}
 
-	if (time.Since(instance.StartTime)) > instance.MaxInstanceLength {
```
jakubno, line 43: this can't happen

jakubno:
> ### Bug: API Violates Contract by Silently Ignoring Keep-Alive Limits
> The `KeepAliveFor` method no longer returns an HTTP 403 error when a sandbox exceeds its `MaxInstanceLength`. Instead, it now silently succeeds, effectively setting the sandbox's expiration to the current time. This change means API callers aren't informed when their keep-alive request fails due to the uptime limit, impacting API contract and resource protection.
> 
> `packages/api/internal/cache/instance/sync.go#L43-L49`
> https://github.com/e2b-dev/infra/blob/1a7d3e45e537dbe2c9237a458517b8b971b23407/packages/api/internal/cache/instance/sync.go#L43-L49
> 
> [Fix in Cursor](https://cursor.com/open?data=eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6ImJ1Z2JvdC12MSJ9.eyJ2ZXJzaW9uIjoxLCJ0eXBlIjoiQlVHQk9UX0ZJWF9JTl9DVVJTT1IiLCJkYXRhIjp7InJlZGlzS2V5IjoiYnVnYm90OmM3OGUzOGM5LTRkNjYtNGVkZS1iNjhiLTllMmU5OTMxMmRlOSIsImVuY3J5cHRpb25LZXkiOiJ0ZUUtRGtnSnJPMzNmM2ZlNEhLSkRESmdBUy1UenpfdjJCZExxTkM4NkdFIiwiYnJhbmNoIjoicmV3b3JrLXNhbmRib3gtbG9ja2luZy1pbi1hcGktZTJiLTI5NTEifSwiaWF0IjoxNzU4MTk4ODc1LCJleHAiOjE3NTg4MDM2NzV9.WBvx6gWZNMx1odV1Q5nHmkdWzYN6DXw-_epn7BQUbuZq1QTU4Fbkd4nLTOX6eB27xMgfyI3CIrmNjEPcbW1POweJxy03x58xLs7C8UWxhymsHXN4NsqPG7UQk7o2_X0D7qx461mb7EKxIc_AwvuFuz8wIkcr5ZSPOgdV-4vs2KudRdyNF0Q7QFVQq98zyW1HMdsoXjX10LRrNqV6fsG2WheDZCwc1GQW5MccVdOdi1GnzIf-LJKzfjSl-NGjlP2fzMaxhp8T51cT__yaJww1NmcE6U-2Qb-YJztpixb_ZpwD_momH2d_UO0YAAGbkKvbqVWd3fnFXu6gjtP7s_G7nQ) [Fix in Web](https://cursor.com/agents?data=eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6ImJ1Z2JvdC12MSJ9.eyJ2ZXJzaW9uIjoxLCJ0eXBlIjoiQlVHQk9UX0ZJWF9JTl9XRUIiLCJkYXRhIjp7InJlZGlzS2V5IjoiYnVnYm90OmM3OGUzOGM5LTRkNjYtNGVkZS1iNjhiLTllMmU5OTMxMmRlOSIsImVuY3J5cHRpb25LZXkiOiJ0ZUUtRGtnSnJPMzNmM2ZlNEhLSkRESmdBUy1UenpfdjJCZExxTkM4NkdFIiwiYnJhbmNoIjoicmV3b3JrLXNhbmRib3gtbG9ja2luZy1pbi1hcGktZTJiLTI5NTEiLCJyZXBvT3duZXIiOiJlMmItZGV2IiwicmVwb05hbWUiOiJpbmZyYSIsInByTnVtYmVyIjoxMTU5LCJjb21taXRTaGEiOiIxYTdkM2U0NWU1MzdkYmUyYzkyMzdhNDU4NTE3YjhiOTcxYjIzNDA3In0sImlhdCI6MTc1ODE5ODg3NSwiZXhwIjoxNzU4ODAzNjc1fQ.XYoOvxnqFFF3ogqTqFpZNgWJdJTgRKRKk5C6u-WhzWpHFjFdt0kG6m8fmWAfwLB-MUJHi12iD05tyUKjdHbR1pksQQJcgh6GYsisQGIY7_MMCH4l2YSa0G0o3CxI7_BvkKlICyf0GaB5wz7cTB8pb-jzPCSFOKn4rQw-f_TKbXSWTs0mbJ3mzXbiRSvBNggckjWSf0u_IiUctIHqpQiss5XFpihFvuRQCFjnWWvJTof3UssTAo5bB6P8hL8uVL5bCCk26TjrM_Qs9F4dG7WaZwFcMGgDM_KsaTAVP9z65HLxMbGm3_yVhRycjjFRchwKoT3lLawBiz5GYr6hTlhYow)

I will return the logic as it isn't relevant to the PR

packages/api/internal/cache/instance/instance_test.go
```diff
@@ -0,0 +1,495 @@
...
+// Test basic state transitions
+func TestStartRemoving_BasicTransitions(t *testing.T) {
+	tests := []struct {
+		name        string
+		fromState   State
+		removeType  RemoveType
+		expState    State
+		shouldError bool
+	}{
+		{"Running to Paused", StateRunning, RemoveTypePause, StatePaused, false},
```
dobrac: How about renaming "RemoveType" to "StateAction" or something similar

packages/api/internal/cache/instance/instance.go
```diff
@@ -114,125 +119,137 @@ type InstanceInfo struct {
...
 
-	// Mark the stop time
-	i.setExpired()
-
-	return nil
-}
+		zap.L().Debug("State transition already in progress to the same state, waiting", logger.WithSandboxID(i.data.SandboxID), zap.String("state", string(newState)))
+		_, err = transition.WaitWithContext(ctx)
+		if err != nil {
+			zap.L().Error("State transition failed", logger.WithSandboxID(i.data.SandboxID), zap.String("state", string(newState)), zap.Error(err))
```
dobrac: this log might be unnecessary (should be handled by the caller)

packages/api/internal/cache/instance/instance.go
```diff
@@ -114,125 +119,137 @@ type InstanceInfo struct {
...
-	mu    sync.RWMutex
+	State  State
+	Reason error
+}
 
-	stopping *utils.SetOnce[struct{}]
+type InstanceInfo struct {
+	data Data
+
+	transition *utils.SetOnce[struct{}]
```
dobrac: you can use ErrorOnce instead

packages/api/internal/cache/instance/instance.go
```diff
@@ -114,125 +119,137 @@ type InstanceInfo struct {
...
 
-func (i *InstanceInfo) WaitForStop(ctx context.Context) error {
-	if i.GetState() == StateRunning {
-		return fmt.Errorf("sandbox isn't stopping")
+		// If the transition is to the same state just wait
+		switch {
+		case currentState == newState:
+			return true, func(err error) {}, nil
+		case allowed[currentState][newState]:
+			return i.StartRemoving(ctx, removeType)
```
dobrac: am I correct that this should basically never happen?

packages/api/internal/cache/instance/operations.go
```diff
@@ -19,52 +19,33 @@ const (
...
+	endTime := sandbox.data.EndTime
 
-	if endTime.Sub(sandbox.StartTime) > sandbox.MaxInstanceLength {
-		sandbox.SetEndTime(sandbox.StartTime.Add(sandbox.MaxInstanceLength))
+	if endTime.Sub(sandbox.data.StartTime) > sandbox.data.MaxInstanceLength {
+		sandbox.data.EndTime = sandbox.data.StartTime.Add(sandbox.data.MaxInstanceLength)
 	}
 
-	c.items.SetIfAbsent(sandbox.SandboxID, sandbox)
+	c.items.SetIfAbsent(sandbox.data.SandboxID, sandbox)
```
dobrac: before it updated the end time, now it skips it altogether -> just want to make sure it's ok

I think the setIfAbsent might be actually an error state here (if it's not absent -> error)

packages/api/internal/cache/instance/operations.go
```diff
@@ -79,84 +60,66 @@ func (c *MemoryStore) Get(instanceID string, includeEvicting bool) (*InstanceInf
 		return nil, fmt.Errorf("instance \"%s\" doesn't exist", instanceID)
 	}
 
-	if item.IsExpired() && !includeEvicting {
+	if item.data.IsExpired() && !includeEvicting {
```
dobrac: this should be accessed through the getter

packages/api/internal/cache/instance/operations.go
```diff
@@ -19,52 +19,33 @@ const (
 )
 
 // Add the instance to the cache
-func (c *MemoryStore) Add(ctx context.Context, sandbox *InstanceInfo, newlyCreated bool) error {
-	sbxlogger.I(sandbox).Debug("Adding sandbox to cache",
+func (c *MemoryStore) Add(ctx context.Context, sandbox *InstanceInfo, newlyCreated bool) {
+	sandbox.mu.Lock()
+	defer sandbox.mu.Unlock()
```
dobrac: do you need the instanceInfo lock here? Wouldn't getting the Data suffice?

packages/api/internal/cache/instance/operations.go
```diff
@@ -79,84 +60,66 @@ func (c *MemoryStore) Get(instanceID string, includeEvicting bool) (*InstanceInf
...
-	for _, callback := range c.removeAsyncCallbacks {
-		go callback(context.WithoutCancel(ctx), sbx, removeType)
-	}
-	sbx.stopDone(err)
-	if err != nil {
-		return fmt.Errorf("error removing instance \"%s\": %w", instanceID, err)
-	}
-
-	return nil
+func (c *MemoryStore) Remove(instanceID string) {
```
dobrac: just rename please the instanceID to sandboxID (same to Exists and Get)

packages/api/internal/cache/instance/operations.go
```diff
@@ -79,84 +60,66 @@ func (c *MemoryStore) Get(instanceID string, includeEvicting bool) (*InstanceInf
...
 
 	return items
 }
 
-func (c *MemoryStore) ExpiredItems() []*InstanceInfo {
+func (c *MemoryStore) ItemsToEvict() []*InstanceInfo {
 	items := make([]*InstanceInfo, 0)
 	for _, item := range c.items.Items() {
-		if !item.IsExpired() {
+		if !item.data.IsExpired() {
```
dobrac: access through the getter

packages/api/internal/cache/instance/operations.go
```diff
@@ -79,84 +60,66 @@ func (c *MemoryStore) Get(instanceID string, includeEvicting bool) (*InstanceInf
...
 	return items
 }
 
-func (c *MemoryStore) ItemsByState(teamID *uuid.UUID, states []State) map[State][]*InstanceInfo {
-	items := make(map[State][]*InstanceInfo)
+func (c *MemoryStore) ItemsByState(teamID *uuid.UUID, states []State) map[State][]Data {
+	items := make(map[State][]Data)
 	for _, item := range c.items.Items() {
-		if teamID != nil && item.TeamID != *teamID {
+		if teamID != nil && item.data.TeamID != *teamID {
```
dobrac: data getter

packages/api/internal/cache/instance/reservation.go
```diff
@@ -48,10 +48,10 @@ func (r *ReservationCache) list(teamID uuid.UUID) (instanceIDs []string) {
 
 func (c *MemoryStore) list(teamID uuid.UUID) (instanceIDs []string) {
 	for _, value := range c.items.Items() {
-		currentTeamID := value.TeamID
+		currentTeamID := value.data.TeamID
```
dobrac: getter

packages/api/internal/cache/instance/instance.go
```diff
@@ -114,125 +119,137 @@ type InstanceInfo struct {
...
 
-	state State
-	mu    sync.RWMutex
+	State  State
+	Reason error
+}
 
-	stopping *utils.SetOnce[struct{}]
+type InstanceInfo struct {
+	data Data
```
dobrac: might be wise to rename the field to `_data` to signify invalid usage outside of getter

packages/api/internal/cache/instance/sync.go
```diff
@@ -27,65 +27,70 @@ func getMaxAllowedTTL(now time.Time, startTime time.Time, duration, maxInstanceL
...
+		return Data{}, &api.APIError{Code: http.StatusForbidden, ClientMsg: msg, Err: errors.New(msg)}
 	} else {
-		maxAllowedTTL := getMaxAllowedTTL(now, instance.StartTime, duration, instance.MaxInstanceLength)
-
-		newEndTime := now.Add(maxAllowedTTL)
-		instance.SetEndTime(newEndTime)
+		maxAllowedTTL := getMaxAllowedTTL(now, instance.data.StartTime, duration, instance.data.MaxInstanceLength)
+		instance.data.EndTime = now.Add(maxAllowedTTL)
+		zap.L().Debug("sandbox ttl updated", zap.String("sandboxID", instance.data.SandboxID), zap.Duration("ttl", maxAllowedTTL))
 	}
```
dobrac: can you move this to the instance, so the lock would be localized?

packages/api/internal/cache/instance/sync.go
```diff
@@ -27,65 +27,70 @@ func getMaxAllowedTTL(now time.Time, startTime time.Time, duration, maxInstanceL
...
+	return instance.data, nil
 }
 
 func (c *MemoryStore) Sync(ctx context.Context, instances []*InstanceInfo, nodeID string) {
 	instanceMap := make(map[string]*InstanceInfo)
 
 	// Use a map for faster lookup
 	for _, instance := range instances {
-		instanceMap[instance.SandboxID] = instance
+		instanceMap[instance.data.SandboxID] = instance
```
dobrac: getter

packages/api/internal/handlers/sandbox_get.go
```diff
@@ -91,6 +93,8 @@ func (a *APIStore) GetSandboxesSandboxID(c *gin.Context, id string) {
 		return
 	}
 
+	zap.L().Debug("Sandbox wasn't found in store", logger.WithSandboxID(sandboxId))
```
dobrac: is this debug message necessary?

packages/api/internal/handlers/sandbox_kill.go
```diff
@@ -82,67 +85,40 @@ func (a *APIStore) DeleteSandboxesSandboxID(
...
-		telemetry.ReportError(ctx, "snapshot for sandbox not found", fmt.Errorf("snapshot for sandbox '%s' not found", sandboxID), telemetry.WithSandboxID(sandboxID))
-		a.sendAPIStoreError(c, http.StatusNotFound, fmt.Sprintf("Error deleting sandbox - sandbox '%s' not found", sandboxID))
-
-		return
-	}
-
-	if deleteSnapshotErr != nil {
+	switch {
+	case errors.Is(deleteSnapshotErr, db.EnvNotFoundError{}):
+		zap.L().Debug("Snapshot for sandbox not found", logger.WithSandboxID(sandboxID))
```
dobrac: do we need this debug message?

packages/api/internal/handlers/sandbox_kill.go
```diff
@@ -82,67 +85,40 @@ func (a *APIStore) DeleteSandboxesSandboxID(
...
-			a.sendAPIStoreError(c, http.StatusInternalServerError, fmt.Sprintf("Error deleting sandbox: %s", err))
-
-			return
-		}
-
-		telemetry.ReportEvent(ctx, "deleted sandbox from orchestrator")
-
-		c.Status(http.StatusNoContent)
-
+	removed := false
```
dobrac: rename to something like "killedOrDeleted"

packages/api/internal/orchestrator/delete_instance.go
```diff
@@ -12,16 +13,85 @@ import (
...
 	defer childSpan.End()
 
-	// SandboxStore will remove the sandbox both from the store and from the orchestrator
-	return o.sandboxStore.Remove(ctx, sandbox.SandboxID, removeType)
+	sbx, err := o.GetSandbox(sandboxID, true)
+	if err != nil {
+		return ErrSandboxNotFound
+	}
+
+	if sbx.TeamID() != teamID {
```
dobrac: do we need to pass the ID this deep to handle authorization?

packages/api/internal/handlers/sandbox_resume.go
```diff
@@ -57,31 +57,39 @@ func (a *APIStore) PostSandboxesSandboxIDResume(c *gin.Context, sandboxID api.Sa
 	sandboxID = utils.ShortID(sandboxID)
 	sbxCache, err := a.orchestrator.GetSandbox(sandboxID, true)
 	if err == nil {
-		state := sbxCache.GetState()
-		switch state {
-		case instance.StatePaused, instance.StatePausing:
-			err = sbxCache.WaitForStop(ctx)
+		zap.L().Debug("Sandbox found in store", logger.WithSandboxID(sandboxID))
```
dobrac: is this necessary?

packages/api/internal/handlers/sandbox_resume.go
```diff
@@ -57,31 +57,39 @@ func (a *APIStore) PostSandboxesSandboxIDResume(c *gin.Context, sandboxID api.Sa
...
+
+		data := sbxCache.Data()
+		switch data.State {
+		case instance.StateFailed:
+			zap.L().Error("Sandbox is in failed state", logger.WithSandboxID(sandboxID), zap.Error(data.Reason))
+			a.sendAPIStoreError(c, http.StatusInternalServerError, "Sandbox can't be resumed, instance is in failed state")
```
dobrac: "Sandbox failed, it can't be resumed" (or something not mentioning state or instance)

packages/api/internal/handlers/sandbox_resume.go
```diff
@@ -57,31 +57,39 @@ func (a *APIStore) PostSandboxesSandboxIDResume(c *gin.Context, sandboxID api.Sa
...
-				zap.Time("end_time", sbxCache.GetEndTime()),
-				zap.Time("start_time", sbxCache.StartTime),
-				zap.String("node_id", sbxCache.NodeID),
+				zap.Time("end_time", data.EndTime),
+				zap.Time("start_time", data.StartTime),
+				zap.String("node_id", data.NodeID),
 			)
 
 			return
 		}
```
dobrac, line 87: missing default case, it will fall through to sandbox not found (even though it is)

packages/api/internal/handlers/sandboxes_list.go
```diff
@@ -167,6 +167,11 @@ func (a *APIStore) GetV2Sandboxes(c *gin.Context, params api.GetV2SandboxesParam
 	}
 
 	if slices.Contains(states, api.Paused) {
+		pausing := sandboxesInCache[instance.StatePaused]
+		for _, info := range pausing {
+			runningSandboxesIDs = append(runningSandboxesIDs, utils.ShortID(info.SandboxID))
+		}
```
dobrac, line 172: this is nice catch, but I think they should be appended already on line 150-152, as these are relevant for running sandboxes too

packages/api/internal/handlers/sandboxes_list.go
```diff
@@ -260,11 +265,16 @@ func snapshotsToPaginatedSandboxes(snapshots []queries.GetSnapshotsWithCursorRow
...
-func instanceInfoToPaginatedSandboxes(runningSandboxes []*instance.InstanceInfo) []utils.PaginatedSandbox {
+func instanceInfoToPaginatedSandboxes(runningSandboxes []instance.Data) []utils.PaginatedSandbox {
 	sandboxes := make([]utils.PaginatedSandbox, 0)
 
 	// Add running sandboxes to results
 	for _, info := range runningSandboxes {
+		state := api.Running
+		if info.State == instance.StatePaused {
+			state = api.Paused
+		}
```
dobrac, line 276: this is interesting, it actually misses the pausing state

but, I think sandbox that is pausing, should be in running state (because it's not yet paused). Or do you think otherwise?

packages/api/internal/orchestrator/delete_instance.go
```diff
@@ -12,16 +13,85 @@ import (
...
+		zap.L().Info("Sandbox was already in the process of being removed", logger.WithSandboxID(sandboxID), zap.String("state", string(sbx.State())))
+
+		return nil
+	}
+
+	sbx.SetExpired()
+	data := sbx.Data()
+
+	defer o.countersRemove(context.WithoutCancel(ctx), data, removeType)
+	defer o.analyticsRemove(context.WithoutCancel(ctx), data, removeType)
```
dobrac: these have been run previously asynchronously

packages/api/internal/orchestrator/delete_instance.go
```diff
@@ -12,16 +13,85 @@ import (
...
+	defer o.sandboxStore.Remove(sbx.SandboxID())
```
dobrac: data.SandboxID

packages/api/internal/orchestrator/delete_instance.go
```diff
@@ -12,16 +13,85 @@ import (
...
+	err = o.removeSandboxFromNode(ctx, data, removeType)
+	if err != nil {
+		zap.L().Error("Error pausing sandbox", zap.Error(err), logger.WithSandboxID(sbx.SandboxID()))
```
dobrac: data.SandboxID

packages/api/internal/orchestrator/delete_instance.go
```diff
@@ -12,16 +13,85 @@ import (
...
+		return ErrAccessForbidden
+	}
+
+	return o.removeSandbox(ctx, sbx, removeType)
 }
 
-// removeSandbox should be called from places where you already marked the sandbox as being removed
-func (o *Orchestrator) removeSandbox(ctx context.Context, sandbox *instance.InstanceInfo, removeType instance.RemoveType) error {
+func (o *Orchestrator) removeSandbox(ctx context.Context, sbx *instance.InstanceInfo, removeType instance.RemoveType) error {
+	_, childSpan := tracer.Start(ctx, "remove-sandbox")
```
dobrac: missing ctx (there is a context used later)

packages/api/internal/orchestrator/delete_instance.go
```diff
@@ -12,16 +13,85 @@ import (
...
+	defer childSpan.End()
+
+	sandboxID := sbx.SandboxID()
+	done, finish, err := sbx.StartRemoving(ctx, removeType)
```
dobrac: is the done more like "already in the state?"

packages/api/internal/orchestrator/delete_instance.go
```diff
@@ -12,16 +13,85 @@ import (
...
+				zap.L().Error("Sandbox is in failed state", logger.WithSandboxID(sandboxID), zap.Error(data.Reason))
+				return ErrSandboxOperationFailed
+			case instance.StateKilled:
+				zap.L().Info("Sandbox is already killed", logger.WithSandboxID(sandboxID))
+				return nil
+			default: // It shouldn't happen the sandbox ended in paused state
+				zap.L().Error("Error killing sandbox", zap.Error(err), logger.WithSandboxID(sandboxID))
+				return ErrSandboxOperationFailed
+			}
+		} else {
```
dobrac: is this a case for pause? If so, I would make it explicit and add default to handle invalid/not implemented RemoveType

dobrac (changes requested):
It looks good overall, I like the simplification. One thing I'm not 100% confident is the failed state -> that should be only really temporary, correct?

Also, we're missing the info about the transition states, which makes it slightly harder to follow the code, but otherwise I don't think it causes any issue. (If I understood it correctly, the states are: running -> paused (should this be called pausing?) -> not in the cache)

packages/api/internal/cache/instance/instance.go
jakubno: It can happen. Calling kill on pausing sandbox

packages/api/internal/cache/instance/operations.go
jakubno: Where did it update the end time? It was done on the "to be added" sandbox, but it wasn't added

I think it shouldn't happen and didn't want to change the logic here more than necessary.

dobrac: yeah, from the whole flow, it should have been "to be added" sandbox, so ok 👌

packages/api/internal/cache/instance/instance.go
dobrac: right, I see, ok. Interesting use case

packages/api/internal/handlers/sandboxes_list.go
jakubno: You can't access it and you can resume it, for the user it behaves as in the paused state, I will add a comment

dobrac: ok, sounds reasonable

packages/api/internal/cache/instance/instance.go
jakubno: It's valid use case, better than to return 500 as we've done before..  but I wouldn't say a very interesting one, the pause is then kinda useless 😅

dobrac: yes yes, it's exactly what we do in the tests

packages/api/internal/handlers/sandboxes_list.go
jakubno: What's that good for? I moved `runningSandboxIds` to the paused clause completely, they weren't used outside

dobrac: they're still present on line 150-152 though, otherwise you're correct, they're used only here

packages/api/internal/orchestrator/delete_instance.go
jakubno: Moved up the tree

packages/api/internal/handlers/sandbox_resume.go
jakubno: Completely removed now

packages/api/internal/handlers/sandbox_resume.go
jakubno: Added

packages/api/internal/handlers/sandboxes_list.go
jakubno: I moved, but it wasn't commited at that tiem

packages/api/internal/orchestrator/delete_instance.go
jakubno: Good catch 🙇🏻

packages/api/internal/orchestrator/delete_instance.go
jakubno: Added default

packages/api/internal/cache/instance/operations.go
```diff
@@ -4,77 +4,98 @@ import (
...
+	return c.items.Has(sandboxID)
 }
 
 // Get the item from the cache.
-func (c *MemoryStore) Get(instanceID string, includeEvicting bool) (*InstanceInfo, error) {
-	item, ok := c.items.Get(instanceID)
+func (c *MemoryStore) Get(sandboxID string) (*InstanceInfo, error) {
+	item, ok := c.items.Get(sandboxID)
+	if !ok {
+		return nil, fmt.Errorf("instance \"%s\" doesn't exist", sandboxID)
```
dobrac:
```suggestion
		return nil, fmt.Errorf("sandbox \"%s\" doesn't exist", sandboxID)
```

packages/api/internal/cache/instance/operations.go
```diff
@@ -4,77 +4,98 @@ import (
...
+
+	return item, nil
+}
+
+// GetData the item from the cache.
+func (c *MemoryStore) GetData(sandboxID string, includeEvicting bool) (Data, error) {
+	item, ok := c.items.Get(sandboxID)
 	if !ok {
-		return nil, fmt.Errorf("instance \"%s\" doesn't exist", instanceID)
+		return Data{}, fmt.Errorf("instance \"%s\" doesn't exist", sandboxID)
```
dobrac:
```suggestion
		return Data{}, fmt.Errorf("sandbox \"%s\" doesn't exist", sandboxID)
```

packages/api/internal/cache/instance/operations.go
```diff
@@ -4,77 +4,98 @@ import (
...
 	}
 
-	if item.data.IsExpired() && !includeEvicting {
-		return nil, fmt.Errorf("instance \"%s\" is being evicted", instanceID)
+	data := item.Data()
+
+	if data.IsExpired() && !includeEvicting {
+		return Data{}, fmt.Errorf("instance \"%s\" is being evicted", sandboxID)
```
dobrac:
```suggestion
		return Data{}, fmt.Errorf("sandbox \"%s\" is being evicted", sandboxID)
```

packages/api/internal/cache/instance/operations.go
```diff
@@ -4,77 +4,98 @@ import (
...
+	if endTime.Sub(sandbox.StartTime) > sandbox.MaxInstanceLength {
+		sandbox.EndTime = sandbox.StartTime.Add(sandbox.MaxInstanceLength)
 	}
 
-	c.items.SetIfAbsent(sandbox.data.SandboxID, sandbox)
+	added := c.items.SetIfAbsent(sandbox.SandboxID, &InstanceInfo{
+		_data:      sandbox,
+		transition: nil,
+		mu:         sync.RWMutex{},
+	})
```
dobrac: should we have a constructor for the instance info?

packages/api/internal/handlers/sandbox_create.go
```diff
@@ -42,7 +42,7 @@ var mostUsedTemplates = map[string]struct{}{
 func (a *APIStore) PostSandboxes(c *gin.Context) {
 	ctx := c.Request.Context()
 
-	// Get team from context, use TeamContextKey
+	// GetData team from context, use TeamContextKey
```
dobrac: why this change?

packages/api/internal/handlers/sandbox_pause.go
```diff
@@ -21,7 +21,7 @@ import (
 
 func (a *APIStore) PostSandboxesSandboxIDPause(c *gin.Context, sandboxID api.SandboxID) {
 	ctx := c.Request.Context()
-	// Get team from context, use TeamContextKey
+	// GetData team from context, use TeamContextKey
```
dobrac: it appears to be in more places, maybe incorrect replace

(please check all occurrences, I don't want to add too many comments)

packages/api/internal/cache/instance/instance.go
```diff
@@ -203,40 +199,41 @@ func (i *InstanceInfo) StartRemoving(ctx context.Context, removeType RemoveType)
...
-		}
-		err = i.transition.SetResult(struct{}{}, err)
-		if err != nil {
-			zap.L().Error("Failed to set transition result", logger.WithSandboxID(i.data.SandboxID), zap.Error(err))
+			// Keep the transition in place so the error stays
+			zap.L().Error("Failed to set transition result", logger.WithSandboxID(i.SandboxID()), zap.Error(setErr))
 		}
 
-		i.transition = nil
+		if err == nil {
```
dobrac: this could be just else, or better, exit early when err != nil

packages/api/internal/orchestrator/cache.go
```diff
@@ -19,15 +19,24 @@ var tracer = otel.Tracer("github.com/e2b-dev/infra/packages/api/internal/orchest
...
 	return item, nil
 }
 
+func (o *Orchestrator) GetSandboxData(sandboxID string, includeEvicting bool) (instance.Data, error) {
+	item, err := o.sandboxStore.GetData(sandboxID, includeEvicting)
+	if err != nil {
+		return instance.Data{}, fmt.Errorf("failed to get sandbox '%s': %w", sandboxID, err)
+	}
+
+	return item, nil
```
dobrac: this look a bit excessive, simply
`return o.sandboxStore.GetData(sandboxID, includeEvicting)`
might be enough

packages/api/internal/orchestrator/cache.go
```diff
@@ -19,15 +19,24 @@ var tracer = otel.Tracer("github.com/e2b-dev/infra/packages/api/internal/orchest
...
 
-func (o *Orchestrator) GetSandbox(sandboxID string, includeEvicting bool) (*instance.InstanceInfo, error) {
-	item, err := o.sandboxStore.Get(sandboxID, includeEvicting)
+func (o *Orchestrator) GetSandbox(sandboxID string) (*instance.InstanceInfo, error) {
+	item, err := o.sandboxStore.Get(sandboxID)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get sandbox '%s': %w", sandboxID, err)
 	}
 
 	return item, nil
```
dobrac: this look a bit excessive, simply
`return o.sandboxStore.Get(sandboxID)`
might be enough

packages/api/internal/orchestrator/delete_instance.go
```diff
@@ -13,79 +12,60 @@ import (
...
 
-	sbx.SetExpired()
-	data := sbx.Data()
-
-	defer o.countersRemove(context.WithoutCancel(ctx), data, removeType)
-	defer o.analyticsRemove(context.WithoutCancel(ctx), data, removeType)
-	defer o.sandboxStore.Remove(sbx.SandboxID())
-	err = o.removeSandboxFromNode(ctx, data, removeType)
+	defer func() { go o.countersRemove(context.WithoutCancel(ctx), sbx, stateAction) }()
+	defer func() { o.analyticsRemove(context.WithoutCancel(ctx), sbx, stateAction) }()
```
dobrac: is this still sync on purpose?

dobrac (changes requested):
Great work! It's so much better with the immutable data state. Just few nits missing

packages/api/internal/handlers/sandbox_create.go
jakubno: oops, renamed method and didn't noticed IDE changed some docstrings

packages/api/internal/orchestrator/delete_instance.go
jakubno: No 🙈

dobrac (approved):

PR #1164, opened 2025-09-12
Add fallback for getting the template builder

jakubno:
This was affecting old builds + snapshots

### DIFF
--- a/tests/integration/internal/main_test.go
+++ b/tests/integration/internal/main_test.go
@@ -42,7 +42,6 @@ func TestCacheTemplate(t *testing.T) {
 			t.Logf("Response error: %d %v", sbx.StatusCode(), string(sbx.Body))
 		default:
 			utils.TeardownSandbox(t, c, sbx.JSON201.SandboxID)
-
 		}
 	})
 }


packages/api/internal/template-manager/template_manager.go
```diff
@@ -175,7 +175,16 @@ func (tm *TemplateManager) DeleteBuild(ctx context.Context, buildID uuid.UUID, t
 
 	client, err := tm.GetClusterBuildClient(clusterID, nodeID)
 	if err != nil {
-		return fmt.Errorf("failed to get builder client: %w", err)
+		nodeID, err = tm.GetAvailableBuildClient(ctx, clusterID)
```
jakubno, line 181:
```suggestion
		// Note: the nodeID can be an orchestrator ID,
		// we may want to improve this later by adding the Delete method to Orchestrator as well.
		// This way we can remove the build (snapshot) from cache as well
		nodeID, err = tm.GetAvailableBuildClient(ctx, clusterID)
```

packages/api/internal/template-manager/template_manager.go
```diff
@@ -175,7 +175,19 @@ func (tm *TemplateManager) DeleteBuild(ctx context.Context, buildID uuid.UUID, t
 
 	client, err := tm.GetClusterBuildClient(clusterID, nodeID)
 	if err != nil {
-		return fmt.Errorf("failed to get builder client: %w", err)
+		// Note: the nodeID can be an orchestrator ID,
```
sitole: Can we add a note that when you want to delete a snapshot, it's actually the client node ID, not the builder node ID, so it will always fall back?

packages/api/internal/template-manager/template_manager.go
```diff
@@ -175,7 +175,19 @@ func (tm *TemplateManager) DeleteBuild(ctx context.Context, buildID uuid.UUID, t
...
+		// we may want to improve this later by adding the Delete method to Orchestrator as well.
```
jakubno:
```suggestion
		// nodeID can be an orchestrator ID, if the build is a snapshot.
		// We may want to improve this later by adding the Delete method to Orchestrator as well.
```

packages/api/internal/template-manager/template_manager.go
```diff
@@ -175,7 +175,19 @@ func (tm *TemplateManager) DeleteBuild(ctx context.Context, buildID uuid.UUID, t
 
 	client, err := tm.GetClusterBuildClient(clusterID, nodeID)
 	if err != nil {
-		return fmt.Errorf("failed to get builder client: %w", err)
+		// nodeID can be an orchestrator ID, if the build is a snapshot.
```
jakubno:
```suggestion
		// nodeID can be an orchestrator ID, if the build corresponds to a snapshot.
```

PR #1166, opened 2025-09-12
Use OTEL collector contrib for host metrics

sitole:
This pull request introduces the use of OTEL contrib host metrics. This allows us to utilize them and standardize monitoring for environments where Google Cloud and/or Nomad monitoring is not available.

Currently forwarded metrics
- Number of physical and logical cores (useful for clusters where we don't have same hosts with the same number of cpus)
- CPU avg load (1min and 5min), AVG across all CPU cores
- Network bytes IN/OUT for all interfaces (excluding `lo`, `veth-*` and `docker*` one)
- Memory available bytes, memory limit bytes
- Filesystem free, used, reserved (excluding `/dev/loop*`), dropping mount point, type, and mode
- For every host metric, there is `node.id` appended so we can distinguish between nodes. Maybe we should add `node.pool` too for environments where `node.id` cannot be easily used to distinguish node type.

### DIFF
--- a/packages/otel-collector/Makefile
+++ b/packages/otel-collector/Makefile
@@ -12,7 +12,7 @@ run:
 	clickhouse_username=$(CLICKHOUSE_USERNAME) \
 	clickhouse_port=$(CLICKHOUSE_PORT) \
 	clickhouse_database=$(CLICKHOUSE_DATABASE) \
-	envsubst < ../../iac/provider-gcp/nomad/configs/otel-collector.yaml > "./otel-collector.local.yaml"
+	envsubst < ./tests/otel-collector.yaml > "./otel-collector.local.yaml"
 
 	docker network create e2b || true
 	docker run --rm --name otel-collector \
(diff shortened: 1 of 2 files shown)


iac/provider-gcp/nomad/configs/otel-collector.yaml
```diff
@@ -18,6 +18,68 @@ receivers:
...
+            enabled: false
+          system.memory.utilization:
+            enabled: false
+
+      #  filter not needed stuff (mode, type)
+      #  filter /dev/loop*, maybe check fo other that are polluting
+      filesystem:
+        metrics:
+          system.filesystem.inodes.usage:
+            enabled: false
```
djeebus, line 80: inode usage is occasionally useful for projects that have lots of small files. for example, node projects w/ large node_packages folders

djeebus: although ... I suppose those would be inside the larger rootfs file, right? so probably not useful, nevermind!

iac/provider-gcp/nomad/configs/otel-collector.yaml
```diff
@@ -18,6 +18,68 @@ receivers:
...
+          system.filesystem.usage:
```
jakubno, line 81: isn't there some metric for total size?

PR #1167, opened 2025-09-12
Upgrade otel, merge 'require' blocks

djeebus:


djeebus:
By accident (or by design, since apparently `go mod tidy` sometimes makes them: https://github.com/golang/go/issues/56471). 

Turns out someone has made a tool that formats this file consistently. This is what it would look like if we started using it: https://github.com/e2b-dev/infra/pull/1168

jakubno:
From curiosity, what was the reason for the package update?

PR #1169, opened 2025-09-13
Remove embedded context from Dispatch, Checks, and DirectPathMount

djeebus:
This removes a few key blockers to improving tracing, and makes the linters a little happierrr

### DIFF
--- a/packages/orchestrator/internal/sandbox/checks.go
+++ b/packages/orchestrator/internal/sandbox/checks.go
@@ -22,7 +22,6 @@ const (
 type Checks struct {
 	sandbox *Sandbox
 
-	ctx       context.Context // nolint:containedctx // todo: refactor so this can be removed
 	cancelCtx context.CancelCauseFunc
 
 	healthy atomic.Bool
@@ -32,16 +31,10 @@ type Checks struct {
 
 var ErrChecksStopped = errors.New("checks stopped")
 
-func NewChecks(ctx context.Context, sandbox *Sandbox, useClickhouseMetrics bool) (*Checks, error) {
-	_, childSpan := tracer.Start(ctx, "checks-create")
-	defer childSpan.End()
-
+func NewChecks(sandbox *Sandbox, useClickhouseMetrics bool) (*Checks, error) {
 	// Create background context, passed ctx is from create/resume request and will be canceled after the request is processed.
-	ctx, cancel := context.WithCancelCause(context.Background())
 	h := &Checks{
 		sandbox:              sandbox,
-		ctx:                  ctx,
-		cancelCtx:            cancel,
 		healthy:              atomic.Bool{}, // defaults to `false`
 		UseClickhouseMetrics: useClickhouseMetrics,
 	}
@@ -52,35 +45,37 @@ func NewChecks(ctx context.Context, sandbox *Sandbox, useClickhouseMetrics bool)
 	return h, nil
 }
 
-func (c *Checks) Start() {
-	c.logHealth()
+func (c *Checks) Start(ctx context.Context) {
+	ctx, c.cancelCtx = context.WithCancelCause(ctx)
+
+	c.logHealth(ctx)
 }
 
 func (c *Checks) Stop() {
 	c.cancelCtx(ErrChecksStopped)
 }
 
-func (c *Checks) logHealth() {
+func (c *Checks) logHealth(ctx context.Context) {
 	healthTicker := time.NewTicker(healthCheckInterval)
 	defer func() {
 		healthTicker.Stop()
 	}()
 
 	// Get metrics and health status on sandbox startup
-	go c.Healthcheck(false)
+	go c.Healthcheck(ctx, false)
 
 	for {
 		select {
 		case <-healthTicker.C:
-			c.Healthcheck(false)
-		case <-c.ctx.Done():
+			c.Healthcheck(ctx, false)
+		case <-ctx.Done():
 			return
 		}
 	}
 }
 
-func (c *Checks) Healthcheck(alwaysReport bool) {
-	ok, err := c.GetHealth(healthCheckTimeout)
+func (c *Checks) Healthcheck(ctx context.Context, alwaysReport bool) {
+	ok, err := c.GetHealth(ctx, healthCheckTimeout)
 	// Sandbox stopped
 	if errors.Is(err, ErrChecksStopped) {
 		return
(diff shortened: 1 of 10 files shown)


packages/orchestrator/internal/sandbox/checks.go
```diff
@@ -52,35 +45,39 @@ func NewChecks(ctx context.Context, sandbox *Sandbox, useClickhouseMetrics bool)
 	return h, nil
 }
 
-func (c *Checks) Start() {
-	c.logHealth()
+func (c *Checks) Start(ctx context.Context) {
+	ctx, c.cancelCtx = context.WithCancelCause(ctx)
```
dobrac, line 49: the context was previously without parent cancel

packages/orchestrator/internal/sandbox/nbd/path_direct.go
```diff
@@ -102,13 +98,13 @@ func (d *DirectPathMount) Open(ctx context.Context) (retDeviceIndex uint32, err
 			}
 			server.Close()
 
-			dispatch := NewDispatch(d.ctx, serverc, d.Backend) // nolint:contextcheck // TODO: fix this later
+			dispatch := NewDispatch(serverc, d.Backend) // nolint:contextcheck // TODO: fix this later
```
dobrac: is the nolint still required here?

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -536,9 +538,9 @@ func ResumeSandbox(
 		return nil, fmt.Errorf("failed to wait for sandbox start: %w", err)
 	}
 
-	go sbx.Checks.Start() // nolint:contextcheck // TODO: fix this later
+	go sbx.Checks.Start(runCtx) // nolint:contextcheck // TODO: fix this later
```
dobrac: is the nolint still required here?

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -536,9 +538,9 @@ func ResumeSandbox(
...
 
-	go func() {
+	go func(ctx context.Context) {
```
dobrac: can we redo the go func(ctx context.Context) to go func(){} and pass the runCtx directly to the function body please?

(it will also uncover that double context.WithoutCancel wrap that happened)

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -301,7 +301,7 @@ func (s *server) Delete(ctxConn context.Context, in *orchestrator.SandboxDeleteR
 	s.sandboxes.Remove(in.SandboxId)
 
 	// Check health metrics before stopping the sandbox
-	sbx.Checks.Healthcheck(true) // nolint:contextcheck // TODO: fix this later
+	sbx.Checks.Healthcheck(ctx, true) // nolint:contextcheck // TODO: fix this later
```
dobrac: is the nolint still required here?

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -320,6 +321,8 @@ func ResumeSandbox(
 	ctx, childSpan := tracer.Start(ctx, "resume-sandbox")
 	defer childSpan.End()
 
+	runCtx := context.WithoutCancel(ctx)
```
dobrac, line 322: wondering if this shouldn't include a new root span to not pollute the resume-sandbox (same applies to the create-sandbox)

this might be united to some sandbox lifecycle root span together with:
`ctx, span := tracer.Start(context.WithoutCancel(ctx), "sandbox-exit-wait", trace.WithNewRoot())`

djeebus: My thinking was that the "resume-sandbox" trace would include everything that the grpc call was waiting on, and everything that could fail the grpc call. This PR is a piece of the over all effort though, so I can avoid submitting a 100-file 1000 line PR =/ This one shouldn't make anything _worse_, but the next one will make it much better.

dobrac: I agree, just note the runCtx context isn't used only in the critical path (where the grpc call is waiting on), but also in spawned goroutines

(where as before it used context.Background())

packages/orchestrator/internal/sandbox/block/chunk.go
```diff
@@ -17,7 +17,7 @@ import (
 )
 
 type Chunker struct {
-	ctx context.Context // nolint:containedctx // todo: refactor so this can be removed
+	ctx context.Context //nolint:containedctx // todo: refactor so this can be removed
```
dobrac: what happened here? (and lot of other places)

packages/orchestrator/internal/sandbox/sandbox.go
djeebus: It's the other way around. `runCtx` is *only* used for the spawned goroutines (except for fetching the template, since that can run across multiple sandboxes), and `ctx` is *only* used for the grpc critical path.

packages/orchestrator/internal/sandbox/block/chunk.go
djeebus: It's the `nolintlint` linter. I think the general golang practice is to have _no_ space between the `//` and the directive IF the directive is machine readable, so things like `//go:build`, `//go:generate`, etc. I think the linter is treating it like that. The benefit we get from the linter is that it removes `//nolint:<linter>` comments when they're no longer useful, so it's worth the trade off imho.

dobrac: ok, sounds reasonable

packages/orchestrator/internal/sandbox/sandbox.go
dobrac: yes, but the runCtx is based on `ctx` (as opposed to context.Background()), so it will include the traces from the ctx and not a new root

dobrac (approved):
Looks good. Just the one comment (https://github.com/e2b-dev/infra/pull/1169#discussion_r2355124687), but I don't want to block it any further.

packages/orchestrator/internal/sandbox/sandbox.go
djeebus: OH! Sorry, I see what you're saying now. I made this choice to avoid a scenario where we would need to end the span generated from `runCtx` early if we bailed in this function, but keep it going if we exited the function while the sandbox was still running. Let me see if I can straighten this out though.

djeebus:
Definitely better! More improvements coming!

PR #1170, opened 2025-09-14
Tweak setup to prepare for better concurrency handling

ValentaTomas:


packages/orchestrator/internal/sandbox/template/cache.go
```diff
@@ -29,7 +29,7 @@ const (
 
 	// buildCacheMaxUsedPercentage the maximum percentage of the cache disk storage
 	// that can be used before the cache starts evicting items.
-	buildCacheMaxUsedPercentage = 75.0
+	buildCacheMaxUsedPercentage = 85.0
```
ValentaTomas, line 32: Currently, the cache disk has cca 128 GiB free RAM. The biggest drop below that I saw was cca to 103. This change moves the free to 76.8, where the drop would be to 51.8.

iac/provider-gcp/nomad-cluster/scripts/start-client.sh
```diff
@@ -218,7 +218,7 @@ echo "- Huge page size: $hugepage_size_in_mib MiB"
 hugepages=$(($hugepages_ram / $hugepage_size_in_mib))
 
 # This percentage will be permanently allocated for huge pages and in monitoring it will be shown as used.
-base_hugepages_percentage=50
+base_hugepages_percentage=80
```
ValentaTomas, line 221: The RAM usage (excluding hugepages) appears to fluctuate between 170-190 GiB free RAM.

This change allocated an additional 96 GiB—the fluctuations should now be 74-94 GiB free RAM.

90-day worst case was 208 GiB RAM—that would be 56 GiB free with the allocation change.

This worst allocation also might (not 100% sure) include the overcommitted hugepages, so it is actually better.

packages/orchestrator/internal/sandbox/template/cache.go
ValentaTomas: I'm tweaking this, because I'm a little afraid of the cache/times with the adding of more nodes. Also it was wasteful.

jakubno (approved):
lgtm

PR #1190, opened 2025-09-17
Remove some more embedded contexts: Chunker, FileSystemStorageObjectProvider

djeebus:
The Chunker required pushing ctx through a lot of layers, but I think this gets us most of the way towards end-to-end traces.


packages/orchestrator/internal/sandbox/build/build.go
```diff
@@ -114,15 +115,15 @@ func (b *File) Slice(off, length int64) ([]byte, error) {
...
 	if err != nil {
 		return nil, fmt.Errorf("failed to get build: %w", err)
 	}
 
-	return build.Slice(mappedOffset, int64(b.header.Metadata.BlockSize))
+	return build.Slice(ctx, mappedOffset, int64(b.header.Metadata.BlockSize))
 }
 
-func (b *File) getBuild(buildID *uuid.UUID) (Diff, error) {
+func (b *File) getBuild(ctx context.Context, buildID *uuid.UUID) (Diff, error) {
```
dobrac: why is the context added here?

packages/orchestrator/internal/sandbox/nbd/dispatch.go
```diff
@@ -51,8 +53,13 @@ type Response struct {
...
 }
 
+type ReaderWriter interface {
+	Read(p []byte) (int, error)
+	Write(p []byte) (int, error)
+}
+
 type Dispatch struct {
-	fp               io.ReadWriteCloser
+	fp               ReaderWriter
```
dobrac: why not io.ReadWriter?

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -897,7 +897,7 @@ func serveMemory(
 		return nil, fmt.Errorf("failed to create uffd: %w", uffdErr)
 	}
 
-	uffdStartErr := fcUffd.Start(sandboxID)
+	uffdStartErr := fcUffd.Start(ctx, sandboxID)
```
dobrac, line 900: on line 892 is missing the span context

packages/orchestrator/internal/sandbox/build/build.go
djeebus: whoops, definitely doesn't need to be there. probably got added mid refactor.  will remove!

PR #1203, opened 2025-09-18
Add the intrange linter

djeebus:
https://github.com/ckaznocha/intrange

Worth noting that these were all fixed automatically!

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -31,6 +31,7 @@ linters:
     - importas
     - inamedparam
     - intrange
+    - mirror
     - misspell
     - nolintlint
     - reassign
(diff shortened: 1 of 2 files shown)


packages/clickhouse/pkg/batcher/batcher_test.go
```diff
@@ -137,10 +137,10 @@ func TestBatcherConcurrentPush(t *testing.T) {
 	}
 	var wg sync.WaitGroup
 	ss := uint32(0)
-	for i := 0; i < 10; i++ {
+	for i := range 10 {
```
jakubno:
```suggestion
	for range 10 {
```

PR #1205, opened 2025-09-18
Add the noctx linter

djeebus:
https://github.com/sonatard/noctx

Lots of changes here, but useful for tracing!


packages/orchestrator/internal/grpcserver/server.go
```diff
@@ -100,7 +100,7 @@ func (g *GRPCServer) GRPCServer() *grpc.Server {
 
 // Start launches
 func (g *GRPCServer) Start(ctx context.Context, port uint) error {
-	lis, err := net.Listen("tcp", fmt.Sprintf(":%d", port))
+	lis, err := new(net.ListenConfig).Listen(ctx, "tcp", fmt.Sprintf(":%d", port))
```
dobrac: lets declare a value instead of `new` (same applies to other places):

```
lc := net.ListenConfig{}
```

packages/orchestrator/internal/sandbox/fc/process.go
```diff
@@ -408,12 +408,12 @@ func getProcessState(pid int) (string, error) {
 	return state, nil
 }
 
-func (p *Process) Stop() error {
+func (p *Process) Stop(ctx context.Context) error {
```
dobrac, line 411: this should be non-cancelable context, or we should take care in the usage

packages/orchestrator/internal/sandbox/checks.go
```diff
@@ -16,7 +16,7 @@ var tracer = otel.Tracer("github.com/e2b-dev/infra/packages/orchestrator/interna
 
 const (
 	healthCheckInterval = 20 * time.Second
-	healthCheckTimeout  = 100 * time.Millisecond
+	healthCheckTimeout  = 250 * time.Millisecond
```
dobrac: this increase should be done in a separate PR (if it's necessary)

packages/orchestrator/internal/sandbox/fc/process.go
```diff
@@ -69,14 +69,14 @@ type Process struct {
 }
 
 func NewProcess(
-	ctx context.Context,
+	ctx, runCtx context.Context,
```
dobrac: passing these two contexts looks confusing to me (and unfortunately requires the knowledge of the usage)

maybe it would make sense to move the "unshare" command to configure (which has already the correct context)

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -586,7 +586,7 @@ func (s *Sandbox) Stop(ctx context.Context) error {
 	// Stop the health checks before stopping the sandbox
 	s.Checks.Stop()
 
-	fcStopErr := s.process.Stop()
+	fcStopErr := s.process.Stop(ctx)
```
dobrac, line 599: unused `sandbox-close` context that should be used

packages/orchestrator/internal/sandbox/checks.go
djeebus: Whoops, this should not have been done here. Maybe from a merge conflict? I'll revert it

packages/orchestrator/internal/sandbox/fc/process.go
djeebus: I suspect this could use a refactor to split background tasks into its own function from the foreground. Definitely confusing as it is!  Can we do that refactor in a follow-up PR, rather than adding to the complexity of this one?

dobrac: I'm alright with doing a refactor in a follow up PR, but can we make it a priority so this doesn't stay there for long?

djeebus: Definitely on board with that :+1:

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -578,15 +578,15 @@ func (s *Sandbox) Close(ctx context.Context) error {
 
 // Stop kills the sandbox.
 func (s *Sandbox) Stop(ctx context.Context) error {
```
dobrac, line 590: when called in `orchestrator/internal/server/sandboxes.go:304` we should make sure the context is not cancelled

dobrac (approved):

PR #1208, opened 2025-09-18
Add the protogetter linter

djeebus:
https://github.com/ghostiam/protogetter

### DIFF
--- a/packages/api/internal/orchestrator/nodemanager/sandboxes.go
+++ b/packages/api/internal/orchestrator/nodemanager/sandboxes.go
@@ -36,42 +36,42 @@ func (n *Node) GetSandboxes(ctx context.Context) ([]*instance.InstanceInfo, erro
 			return nil, fmt.Errorf("sandbox config is nil when listing sandboxes: %#v", sbx)
 		}
 
-		teamID, parseErr := uuid.Parse(config.TeamId)
+		teamID, parseErr := uuid.Parse(config.GetTeamId())
 		if parseErr != nil {
-			return nil, fmt.Errorf("failed to parse team ID '%s' for job: %w", config.TeamId, parseErr)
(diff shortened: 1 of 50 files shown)


packages/api/internal/orchestrator/nodemanager/sandboxes.go
```diff
@@ -36,42 +36,42 @@ func (n *Node) GetSandboxes(ctx context.Context) ([]instance.Data, error) {
...
 			sandboxesInfo,
 			instance.NewSandbox(
-				config.SandboxId,
-				config.TemplateId,
+				config.GetSandboxId(),
+				config.GetTemplateId(),
 				consts.ClientID,
-				config.Alias,
-				config.ExecutionId,
+				config.Alias, //nolint:protogetter // we need the nil check too
```
ValentaTomas, line 55: What is the plan for adding a check here? For example the `Alias` can still have valid value if it is not defined, how would we check for this?

djeebus: Right now `Alias` has three different states that it can be in:
- `nil` (nil pointer)
- `""` (empty string)
- `"some-value"` (non empty string)
Is there a difference between `nil` and `""`? If not, I propose we make `Alias` a `string` (instead of a `*string`) everywhere we can. Does that fix this?

djeebus: If there is a difference between `nil` and `""`, then we should leave this as-is.

ValentaTomas: I'm not 100% sure if we pass "" here already.
The alias "" should not be a valid value.

ValentaTomas: Let's merge and keep the comments there for now as it would require bigger changes across.

ValentaTomas (approved):

PR #1211, opened 2025-09-18
Enable the promlinter

djeebus:
https://github.com/yeya24/promlinter


jakubno (approved):

jakubno (changes requested):
Does it actually make sense to add it if we don't use Prometheus metrics anywhere?

djeebus:
Oh! Right, we use otel to push to Prometheus. Yup, ignore this 👍

PR #1213, opened 2025-09-19
feat: Add subdomain support for Cloudflare DNS

ken8203:
This PR extends Cloudflare DNS integration to properly handle subdomains, which was previously limited to root domains only.

### What's new
- Automatic subdomain detection and DNS record creation
- Support for both root domains and subdomains in the same configuration

### Why this matters

This improvement makes self-hosted setups much easier, as users can now use subdomains like `e2b.example.com` instead of needing separate dedicated domains.

Thanks for reviewing! Happy to address any feedback or questions.


jakubno:
Hey @ken8203, thanks a lot for the contribution! I’m sure many people will find this valuable. 

I created a new PR with a small refactor since I didn’t have permission to push to your fork. We'll merge it soon.

PR #1216, opened 2025-09-19
Add restart configurations across Nomad job files

jakubno:
Add restart policies to the jobs

### DIFF
--- a/iac/provider-gcp/nomad/jobs/docker-reverse-proxy.hcl
+++ b/iac/provider-gcp/nomad/jobs/docker-reverse-proxy.hcl
@@ -5,6 +5,15 @@ job "docker-reverse-proxy" {
   priority    = 85
 
   group "reverse-proxy" {
+    // Try to restart the task indefinitely
+    // Tries to restart every 5 seconds
+    restart {
+      interval         = "5s"
+      attempts         = 1
+      delay            = "5s"
+      mode             = "delay"
+    }
+
     network {
       port "${port_name}" {
         static = "${port_number}"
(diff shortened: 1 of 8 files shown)


iac/provider-gcp/nomad/jobs/docker-reverse-proxy.hcl
```diff
@@ -5,6 +5,15 @@ job "docker-reverse-proxy" {
   priority    = 85
 
   group "reverse-proxy" {
+    // Try to restart the task indefinitely
+    // Tries to restart every 5 seconds
+    restart {
+      interval         = "5s"
+      attempts         = 1
+      delay            = "5s"
```
dobrac, line 13: can you do exponential backoff?

dobrac (changes requested):

iac/provider-gcp/nomad/jobs/docker-reverse-proxy.hcl
dobrac: there is also: https://developer.hashicorp.com/nomad/docs/job-specification/reschedule

(although this doesn't work with system jobs)

jakubno: https://developer.hashicorp.com/nomad/docs/job-specification/restart#mode-values

dobrac (approved):
I've seen the options, but there is also a reschedule option with exponential backoff that might be useful for some of the task (and maybe instead of the restart).

PR #1217, opened 2025-09-19
Update Terraform Google provider to latest 6.X version

sitole:
- We no longer need google beta provider
- Newer version support interesting features such as secrets delete protection


sitole:
It should be safe to test Terraform init and plan to see all our environments. Until the apply is done, the state does not change, and you can safely downgrade the provider version on your clusters.

jakubno (approved):

PR #1222, opened 2025-09-19
Point dockerfile versions at arguments

djeebus:
This should make future upgrades more obvious


djeebus:
Cursor's suggestion is wrong this time, mostly. golang doesn't support pinning to patch-level versions of alpine images, so to reuse them we have to omit that. The minor-level versions of alpine are mutable, and get updated with later revisions of alpine, so we'll still get updates, they just won't be recorded in git.

PR #1223, opened 2025-09-19
Remove context from DiffStore

djeebus:


packages/orchestrator/internal/sandbox/build/cache.go
```diff
@@ -68,7 +66,7 @@ func NewDiffStore(ctx context.Context, cachePath string, ttl, delay time.Duratio
 	})
 
 	go cache.Start()
-	go ds.startDiskSpaceEviction(maxUsedPercentage)
+	go ds.startDiskSpaceEviction(ctx, maxUsedPercentage)
```
djeebus, line 69: This is run in a goroutine, but the context comes from `main`, so it makes sense that this should cancel when that context does. Also, it's the same context that it was using before this PR, just w/o the field on the struct.

dobrac (approved):

PR #1226, opened 2025-09-19
Enable the S1002 bool comparison check

djeebus:
https://staticcheck.dev/docs/checks#S1002


jakubno:
I purposefully disabled it, i don't see any added value, i may even prefer `value == false` over `!value`, it's easy to miss the `!`. 

Is there any rationale behind this rule?

PR #1231, opened 2025-09-22
Remove noisy log from API

jakubno:
Remove noisy log -we have a corresponding message in client proxy anyway


jakubno:
It should be fine, it doesn't matter much who does the request to redis

PR #1233, opened 2025-09-22
Hyperloop Gin server and logs forwarding

sitole:
- Initial setup for hyperloop server (gin, generated openapi schema, request validation...)
- Logs endpoint that received logs from sandbox and forwards them to Vector logs collector
- Removed need for `LOGS_COLLECTOR_PUBLIC_IP` in orchestrator, replaced with hyperloop


iac/provider-gcp/nomad-cluster/network/main.tf
```diff
@@ -98,6 +98,7 @@ locals {
 
 # ======== IP ADDRESSES ====================
 
+// todo: (2025-09-22): this can be removed when all orchestrator will be rolled with internal logs collector server
```
sitole, line 101: task for future IaC removal https://linear.app/e2b/issue/ENG-3029/remove-vector-public-load-balancer-from-iac

sitole:
> rr = hyperloopSrv.Shutdown(ctx)

Thank you, good bot!

spec/openapi-hyperloop.yml
```diff
@@ -0,0 +1,84 @@
...
+      description: Returns information about sandbox itself
+      responses:
+        "200":
+          description: Request was successful
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Me"
+
+  /:
```
jakubno: can you move it under `/logs`?

packages/orchestrator/main.go
```diff
@@ -527,6 +526,13 @@ func run(port, proxyPort, hyperloopPort uint) (success bool) {
 		}
 	}
 
+	zap.L().Info("Shutting down hyperloop server")
+	err = hyperloopSrv.Shutdown(closeCtx)
+	if err != nil {
+		zap.L().Error("error shutting down hyperloop server", zap.Error(err))
+		success = false
+	}
+
```
jakubno, line 535: can this also be a closer?

packages/orchestrator/internal/hyperloopserver/server.go
```diff
@@ -5,63 +5,52 @@ import (
...
+		Handler: engine,
+		Addr:    fmt.Sprintf("0.0.0.0:%d", port),
 
-func (h *HyperloopStore) Close(ctx context.Context) error {
-	return h.server.Shutdown(ctx)
-}
+		// Configure timeouts to be greater than the proxy timeouts.
+		ReadTimeout:  maxReadTimeout,
+		WriteTimeout: maxWriteTimeout,
+		IdleTimeout:  idleTimeout,
```
jakubno: I don't think this is relevant here

packages/orchestrator/internal/hyperloopserver/server.go
```diff
@@ -5,63 +5,52 @@ import (
...
+	store := handlers.NewHyperloopStore(logger, sandboxes, sandboxCollectorAddr)
+	swagger, err := api.GetSwagger()
+	if err != nil {
+		return nil, fmt.Errorf("error getting swagger spec: %w", err)
 	}
 
-	return store, nil
-}
+	engine := gin.New()
+	engine.Use(
```
jakubno, line 32: `limits.RequestSizeLimiter(maxUploadLimit)` could be useful here

spec/openapi-hyperloop.yml
sitole: Then we need to send it somehow to envd. Now we are using `address` parameter in the init request, and with adding back compatibility, we would need to send `/logs` here too, to make it not re-usable for different things.

Maybe we can do `/logs`, update in newer envds and do `/` just for back compatibility in http server?

packages/orchestrator/main.go
sitole: I am returning just the HTTP server (as we are doing with api' Gin), and there is no method that corresponds to the interface.  I was thinking about returning the entire store structure, but then you wouldn't have a pretty circular dependency (store -> gin -> server listener -> server listener stored in the store).

We can go with
- additional wrapper wrapping store + server, so we can impl close() method here
- store http.Server directly on store (will work but not pretty dependency imho) and implement close() here

sitole:
dd5cede98874d4103aef011126cb8535390e04cc is just wip for now. Moves url for sandbox logs to `/logs/push` with back compatibility for older envd. Re-do of metadata sync in envd.

PR #1234, opened 2025-09-22
Run lazy clock sync for longer running sandboxes

sitole:
We experienced issues due to the missing wall clock sync and virtualized CPUs' sandbox time drifts for sandboxes that run for multiple hours. Adding the Chrony service with a lazy sync interval should keep the clock synced for longer-running sandboxes.

### DIFF
--- a/packages/orchestrator/internal/template/build/phases/base/provision.sh
+++ b/packages/orchestrator/internal/template/build/phases/base/provision.sh
@@ -7,7 +7,7 @@ echo "Making configuration immutable"
 {{ .BusyBox }} chattr +i /etc/resolv.conf
 
 # Install required packages if not already installed
-PACKAGES="systemd systemd-sysv openssh-server sudo linuxptp socat curl"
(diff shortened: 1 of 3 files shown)


packages/orchestrator/internal/template/build/phases/base/provision.sh
```diff
@@ -39,6 +39,25 @@ echo "if [ -f ~/.bashrc ]; then source ~/.bashrc; fi; if [ -f ~/.profile ]; then
 echo "Remove root password"
 passwd -d root
 
+echo "Setting up chrony"
+mkdir -p /etc/chrony
+cat <<EOF >/etc/chrony/chrony.conf
+refclock PHC /dev/ptp0 poll 2 dpoll 2
```
sitole, line 45: Very relaxed sync interval

sitole:
Additionally, reverting to chrony as the default from Ubuntu 25.10 will be considered a replacement for `systemd-timesyncd` (https://documentation.ubuntu.com/server/how-to/networking/timedatectl-and-timesyncd/).

PR #1236, opened 2025-09-22
Add hit/miss metrics for template caching

djeebus:


ValentaTomas:
The `must` change could have been a separate PR.

packages/orchestrator/internal/sandbox/template/cache.go
```diff
@@ -245,3 +232,22 @@ func cleanDir(path string) error {
 
 	return nil
 }
+
+func (c *Cache) getPopulatedTemplate(ctx context.Context, storageTemplate *storageTemplate) Template {
```
ValentaTomas: What does "populated" mean in this context?

djeebus: Just looking for a word that implies "We have begun fetching data for it". "warm template" maybe? or "fetchedTemplate"?

ValentaTomas (approved):

PR #1239, opened 2025-09-23
Add team id to snapshots

jakubno:
# Description

Add `team_id` to `snapshots` to improve list query performance

This will be then used to optimize the list query #1240

### DIFF
--- a/packages/shared/pkg/models/event/sandbox_event.go
+++ b/packages/shared/pkg/models/event/sandbox_event.go
@@ -1,0 +1,19 @@
+package event
+
+import (
+	"time"
+
+	"github.com/google/uuid"
+)
+
+type SandboxEvent struct {
+	Timestamp          time.Time `json:"timestamp"`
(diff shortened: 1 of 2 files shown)


packages/shared/pkg/models/webhooks/webhooks.go
```diff
@@ -1,43 +0,0 @@
-package webhooks
```
sitole, line 1: this is not related, right?

packages/shared/pkg/models/event/sandbox_event.go
```diff
@@ -1,19 +0,0 @@
-package event
-
-import (
```
sitole, line 3: this also

packages/shared/pkg/models/webhooks/webhooks.go
jakubno: Okay, I will remove in separate PR

PR #1243, opened 2025-09-24
Remove load balancer infra for logs collector

sitole:
When https://github.com/e2b-dev/infra/commit/6662158659124c04517abbde37c5aa14263be42a is deployed, we no longer need an infrastructure provisioning load balancer for the logs collector.

### DIFF
--- a/iac/provider-gcp/nomad/variables.tf
+++ b/iac/provider-gcp/nomad/variables.tf
@@ -179,6 +179,10 @@ variable "logs_proxy_port" {
     name = string
     port = number
   })
+  default = {
+    name = "logs"
+    port = 30006
+  }
 }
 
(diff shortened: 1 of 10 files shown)


iac/provider-gcp/nomad/variables.tf
```diff
@@ -179,6 +179,10 @@ variable "logs_proxy_port" {
     name = string
     port = number
   })
+  default = {
```
sitole, line 186: This has been moved to the Nomad module with defaults, as it is still used as a port and port name in the Nomad job spec.

sitole:
> Requests are still coming to the load balancer even after the rollout.

Resolved in person. I think we can now move forward once we understand why this is happening.
Lmk if i should mark it as read  @jakubno

PR #1244, opened 2025-09-24
Expose only the stateless data from the sandbox store

jakubno:
Refactor to expose only stateless data and adapt naming for clarity.

### DIFF
--- a/packages/api/internal/cache/instance/instance.go
+++ b/packages/api/internal/cache/instance/instance.go
@@ -1,23 +1,19 @@
 package instance
 
 import (
-	"context"
-	"fmt"
 	"sync"
 	"time"
 
 	"github.com/google/uuid"
-	"go.uber.org/zap"
 
-	"github.com/e2b-dev/infra/packages/shared/pkg/logger"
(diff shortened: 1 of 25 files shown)


packages/api/internal/cache/instance/instance.go
```diff
@@ -1,23 +1,19 @@
...
-	"go.uber.org/zap"
 
-	"github.com/e2b-dev/infra/packages/shared/pkg/logger"
 	sbxlogger "github.com/e2b-dev/infra/packages/shared/pkg/logger/sandbox"
 	"github.com/e2b-dev/infra/packages/shared/pkg/utils"
 )
 
 const (
-	InstanceExpiration = time.Second * 15
+	SandboxExpiration = time.Second * 15
```
dobrac:
```suggestion
	SandboxTimeoutDefault = time.Second * 15
```

dobrac (approved):
looks good, one nit

I think file rename would be also nice, but lets do it in another PR so the changes are clear

jakubno:
> I think file rename would be also nice, but lets do it in another PR so the changes are clear

I will be separating the store from the data (so I can abstract the store) in the next PR, so there was no point in renaming now

PR #1247, opened 2025-09-24
Add build finalize traces

dobrac:
Add build finalize traces to improve observability

### DIFF
--- a/packages/orchestrator/internal/template/build/phases/finalize/builder.go
+++ b/packages/orchestrator/internal/template/build/phases/finalize/builder.go
@@ -162,6 +162,9 @@ func (ppb *PostProcessingBuilder) Build(
 
 func (ppb *PostProcessingBuilder) postProcessingFn(userLogger *zap.Logger) layer.FunctionActionFn {
 	return func(ctx context.Context, sbx *sandbox.Sandbox, meta metadata.Template) (cm metadata.Template, e error) {
+		ctx, span := tracer.Start(ctx, "post processing")
+		defer span.End()
+
 		defer func() {
 			if e != nil {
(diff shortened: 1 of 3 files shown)


packages/orchestrator/internal/template/build/phases/finalize/builder.go
```diff
@@ -162,6 +162,9 @@ func (ppb *PostProcessingBuilder) Build(
 
 func (ppb *PostProcessingBuilder) postProcessingFn(userLogger *zap.Logger) layer.FunctionActionFn {
 	return func(ctx context.Context, sbx *sandbox.Sandbox, meta metadata.Template) (cm metadata.Template, e error) {
+		ctx, span := tracer.Start(ctx, "post processing")
```
ValentaTomas: The `spanName` in other `tracer.Start` we have in the codebase uses `-` instead of spaces—do you think we should start using spaces or was this just a typo?

ValentaTomas (commented):
Waiting for the resolution of the added comment (will use the review comments in the future).

packages/orchestrator/internal/template/build/phases/finalize/builder.go
dobrac: it's not a typo, it's based on this: https://opentelemetry.io/blog/2025/how-to-name-your-spans/

but I'm open to any other suggestion that would bring some order to the naming

ValentaTomas: I'm okay with the format, I just wasn't sure if we consciously decided to switch, as the rest is still in the different format.

PR #1250, opened 2025-09-25
Refactor template request common methods

dobrac:
Refactor template request common methods (just moved around).


dobrac:
> Telemetry Misreports Errors After Build Failures

Valid issues, but not new here. They'll be fixed in another PR.

packages/api/internal/template/register_build.go
```diff
@@ -0,0 +1,366 @@
...
+}
+
+func RegisterBuild(
+	ctx context.Context,
+	templateBuildsCache *templatecache.TemplatesBuildCache,
+	db *db.DB,
+	sqlcDB *sqlcdb.Client,
+	data RegisterBuildData,
+) (*RegisterBuildResponse, *api.APIError) {
+	ctx, span := tracer.Start(ctx, "build-template-request")
```
ValentaTomas: I think we want to change the trace name after the function rename, since it was previously in a function named `BuildTemplate`.

dobrac: `register build` is ok?

ValentaTomas: Yes.

ValentaTomas (approved):

PR #1252, opened 2025-09-26
Replace os.Getenv calls with a config model

djeebus:
This centralizes configuration and removes the assumption that configuration will always come from environment variables, which makes testing easier.

### DIFF
--- a/packages/api/internal/cfg/model.go
+++ b/packages/api/internal/cfg/model.go
@@ -27,6 +27,9 @@ type Config struct {
 
 	SandboxAccessTokenHashSeed string `env:"SANDBOX_ACCESS_TOKEN_HASH_SEED"`
 
+	// SupabaseJWTSecrets is a list of secrets used to verify the Supabase JWT.
+	// More secrets are possible in the case of JWT secret rotation where we need to accept
+	// tokens signed with the old secret for some time.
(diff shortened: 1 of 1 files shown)


sitole:
Feels much cleaner to me.

Just in case errors occur if some required environment variables are missing, and doing value fallback for optional ones should be done directly in config parsing; if not, you can then use the value in different places, and you will need to write require/fallback logic multiple times.

jakubno:
I agree, it definitely looks better to me. It's much easier now to see what's really used and where.

@sitole, I'm leaving the decision on optional/required versus fallback logic up to you.

packages/api/internal/handlers/store.go
```diff
@@ -31,25 +30,18 @@ import (
...
 )
 
-var supabaseJWTSecretsString = strings.TrimSpace(os.Getenv("SUPABASE_JWT_SECRETS"))
-
 // minSupabaseJWTSecretLength is the minimum length of a secret used to verify the Supabase JWT.
 // This is a security measure to prevent the use of weak secrets (like empty).
 const minSupabaseJWTSecretLength = 16
 
-// supabaseJWTSecrets is a list of secrets used to verify the Supabase JWT.
-// More secrets are possible in the case of JWT secret rotation where we need to accept
```
dobrac, line 47: (just so it's not forgotten) this comment disappeared

PR #1254, opened 2025-09-26
Move the template spawn counter to sqlc

jakubno:
- Refactor the counter to use sqlc
- Pass context


packages/api/internal/utils/counter.go
```diff
@@ -68,13 +75,21 @@ func (t *TemplateSpawnCounter) flushCounters(dbClient *db.DB) {
...
 	}
 }
 
 func (t *TemplateSpawnCounter) Close() {
-	t.done <- true
+	select {
+	case t.done <- true:
+	default:
+		zap.L().Debug("template spawn counter already closed")
+	}
```
dobrac: this doesn't await the final processUpdates flush

otherwise 👍 for keeping the Close function

PR #1258, opened 2025-09-26
Refactor the Create/Resume functions to avoid globals/reduce args

djeebus:
This PR  creates a `Factory` struct, which does a few things:

- moves `networkPool` and `devicePool` to a central place, so they don't need to be passed into `CreateSandbox` and `ResumeSandbox` on every  invocation.
- moves the query for `MetricsWriteFlagName` into the Factory, simplifying the arguments.
- move the check of the `ALLOW_SANDBOX_INTERNET` env var into the `main` function, and out of the global scope.

### DIFF
--- a/packages/orchestrator/cmd/build-template/main.go
+++ b/packages/orchestrator/cmd/build-template/main.go
@@ -155,8 +155,12 @@ func buildTemplate(
 	if err != nil {
 		zap.L().Fatal("failed to create build metrics", zap.Error(err))
 	}
+
+	sandboxFactory := sandbox.NewFactory(networkPool, devicePool, featureFlags, true)
+
 	builder := build.NewBuilder(
 		logger,
+		sandboxFactory,
 		persistenceTemplate,
 		persistenceBuild,
 		artifactRegistry,
--- a/packages/orchestrator/internal/sandbox/sandbox.go
+++ b/packages/orchestrator/internal/sandbox/sandbox.go
@@ -12,7 +12,6 @@ import (
 	"go.opentelemetry.io/otel/trace"
 	"go.uber.org/zap"
 
-	globalconfig "github.com/e2b-dev/infra/packages/orchestrator/internal/config"
 	"github.com/e2b-dev/infra/packages/orchestrator/internal/sandbox/block"
 	"github.com/e2b-dev/infra/packages/orchestrator/internal/sandbox/build"
 	"github.com/e2b-dev/infra/packages/orchestrator/internal/sandbox/fc"
@@ -23,6 +22,7 @@ import (
 	"github.com/e2b-dev/infra/packages/orchestrator/internal/sandbox/uffd"
 	"github.com/e2b-dev/infra/packages/orchestrator/internal/template/metadata"
 	"github.com/e2b-dev/infra/packages/shared/pkg/env"
+	featureflags "github.com/e2b-dev/infra/packages/shared/pkg/feature-flags"
 	"github.com/e2b-dev/infra/packages/shared/pkg/grpc/orchestrator"
 	sbxlogger "github.com/e2b-dev/infra/packages/shared/pkg/logger/sandbox"
 	"github.com/e2b-dev/infra/packages/shared/pkg/storage"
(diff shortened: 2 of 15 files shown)


djeebus:
oh, the template server shouldn't have the hardcoded value, I'll fix that

packages/orchestrator/internal/template/build/layer/create_sandbox.go
```diff
@@ -65,10 +78,8 @@ func (cs *CreateSandbox) Sandbox(
 	}
 
 	// In case of a new sandbox, base template ID is now used as the potentially exported template base ID.
-	sbx, err := sandbox.CreateSandbox(
+	sbx, err := cs.sandboxFactory.CreateSandbox(
 		ctx,
-		layerExecutor.networkPool,
-		layerExecutor.devicePool,
```
jakubno, line 71: I think layer executor doesn't need those anymore

packages/orchestrator/internal/template/build/phases/base/builder.go
```diff
@@ -264,11 +266,9 @@ func (bb *BaseBuilder) buildLayerFromOCI(
 	// Create sandbox for building template
 	userLogger.Debug("Creating base sandbox template layer")
 
-	// TODO: Temporarily set this based on global config, should be removed later (it should be passed as a parameter in build)
```
jakubno, line 267: can you move the TODO? We're still not sending it as part of build

packages/orchestrator/internal/template/server/main.go
```diff
@@ -75,8 +77,17 @@ func New(
 	if err != nil {
 		return nil, fmt.Errorf("failed to create build metrics: %w", err)
 	}
+
+	featureFlags, err := featureflags.NewClient()
+	if err != nil {
+		return nil, fmt.Errorf("failed to create feature flags client: %w", err)
+	}
+
+	sandboxFactory := sandbox.NewFactory(networkPool, devicePool, featureFlags, defaultAllowSandboxInternet)
```
jakubno: do you need it here as well? Why not to use the one from `packages/orchestrator/main.go`

jakubno (changes requested):

packages/orchestrator/internal/template/build/layer/create_sandbox.go
djeebus: Good catch! Removed them and a bunch of other unused fields.

packages/orchestrator/internal/template/build/builder.go
```diff
@@ -189,30 +184,9 @@ func runBuild(
...
-		builder.networkPool,
-		builder.devicePool,
-		builder.templateCache,
-		builder.proxy,
-		builder.sandboxes,
-		builder.templateStorage,
-		builder.buildStorage,
-		index,
-	)
+	layerExecutor := layer.NewLayerExecutor(bc, builder.logger, builder.templateCache, builder.proxy, builder.sandboxes, builder.templateStorage, builder.buildStorage, index)
```
jakubno: can we keep it on separate lines? 🙏🏻  it's easier to review

packages/orchestrator/internal/template/build/builder.go
```diff
@@ -189,30 +184,9 @@ func runBuild(
...
-		builder.proxy,
-		builder.templateStorage,
-		builder.devicePool,
-		builder.networkPool,
-		builder.artifactRegistry,
-		layerExecutor,
-		index,
-		builder.metrics,
-	)
+	baseBuilder := base.New(bc, builder.logger, builder.proxy, builder.templateStorage, builder.artifactRegistry, layerExecutor, index, builder.metrics, builder.sandboxFactory)
```
jakubno: can we keep it on separate lines? 🙏🏻  it's easier to review

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -334,12 +352,13 @@ func ResumeSandbox(
 		}
 	}()
 
-	allowInternet := globalconfig.AllowSandboxInternet
+	// TODO: Temporarily set this based on global config, should be removed later (it should be passed as a parameter in build)
+	allowInternet := f.defaultAllowInternetAccess
```
jakubno: this isn't build

jakubno (changes requested):

jakubno (approved):

packages/orchestrator/cmd/build-template/main.go
```diff
@@ -155,13 +155,15 @@ func buildTemplate(
 	if err != nil {
 		zap.L().Fatal("failed to create build metrics", zap.Error(err))
 	}
+
+	sandboxFactory := sandbox.NewFactory(networkPool, devicePool, featureFlags, true)
```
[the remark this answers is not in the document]
jakubno: That's correct, we need to download packages in this stages

PR #1260, opened 2025-09-26
Fix concurrent hosts file write during sbx start

sitole:
Fixes the issue causing concurrent writing to the hosts file, as it sometimes occurs when we call init multiple times.

### DIFF
--- a/packages/envd/internal/api/store.go
+++ b/packages/envd/internal/api/store.go
@@ -3,6 +3,7 @@ package api
 import (
 	"encoding/json"
 	"net/http"
+	"sync"
 
 	"github.com/rs/zerolog"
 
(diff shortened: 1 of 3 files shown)


dobrac (changes requested):
can you increase envd version please?

packages/envd/internal/api/store.go
```diff
@@ -11,11 +12,12 @@ import (
...
-	logger      *zerolog.Logger
-	accessToken *string
-	envVars     *utils.Map[string, string]
-	mmdsChan    chan *host.MMDSOpts
+	isNotFC       bool
+	logger        *zerolog.Logger
+	accessToken   *string
+	envVars       *utils.Map[string, string]
+	mmdsChan      chan *host.MMDSOpts
+	hyperloopLock *sync.Mutex
```
dobrac: this shouldn't be a pointer

sitole: fixed, thx

dobrac (approved):

PR #1261, opened 2025-09-26
Host file parsing fails without trailing newline

djeebus:


packages/envd/internal/api/init.go
```diff
@@ -83,19 +85,37 @@ func (a *API) SetupHyperloop(address string) {
...
 	// Update /etc/hosts to point events.e2b.local to the hyperloop IP
 	// This will remove any existing entries for events.e2b.local first
 	hosts.AddHost(address, "events.e2b.local")
-	err = hosts.Save()
-	if err != nil {
-		a.logger.Error().Msgf("Failed to add events host entry: %v", err)
+
+	if err = os.WriteFile(outputPath, []byte(hosts.RenderHostsFile()), 0o644); err != nil {
+		return fmt.Errorf("failed to save hosts file: %w", err)
 	}
```
jakubno, line 167: wouldn't it be better to check if the entry is already there and update only in that case? I mean speed wise and probably more friendly to the system

packages/envd/internal/api/init.go
```diff
@@ -83,19 +85,37 @@ func (a *API) SetupHyperloop(address string) {
...
-		a.logger.Error().Msgf("Failed to create hosts: %v", err)
+	if err := rewriteHostsFile(address, "/etc/hosts", "/etc/hosts"); err != nil {
+		a.logger.Error().Err(err).Msg("failed to modify hosts file")
 		return
 	}
 
+	a.envVars.Store("E2B_EVENTS_ADDRESS", fmt.Sprintf("http://%s", address))
+}
+
+func rewriteHostsFile(address, inputPath, outputPath string) error {
```
jakubno: does it make sense a case where `input!=output`

packages/envd/internal/api/init_test.go
```diff
@@ -0,0 +1,49 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const perms = 0o700
```
jakubno: why do you use different permissions than in the actual implementation?

packages/envd/internal/api/init.go
```diff
@@ -83,19 +86,62 @@ func (a *API) SetupHyperloop(address string) {
...
 
-	hosts, err := txeh.NewHosts(&txeh.HostsConfig{ReadFilePath: "/etc/hosts", WriteFilePath: "/etc/hosts"})
-	if err != nil {
-		a.logger.Error().Msgf("Failed to create hosts: %v", err)
+	if err := rewriteHostsFile(address, "/etc/hosts"); err != nil {
+		a.logger.Error().Err(err).Msg("failed to modify hosts file")
 		return
 	}
 
+	a.envVars.Store("E2B_EVENTS_ADDRESS", fmt.Sprintf("http://%s", address))
```
jakubno: To prevent future problems
```suggestion
	if err := rewriteHostsFile(address, "/etc/hosts"); err != nil {
		a.logger.Error().Err(err).Msg("failed to modify hosts file")
	} else {
		a.envVars.Store("E2B_EVENTS_ADDRESS", fmt.Sprintf("http://%s", address))
	}
```

packages/envd/internal/api/init.go
```diff
@@ -83,19 +86,62 @@ func (a *API) SetupHyperloop(address string) {
...
+
+	return nil
+}
+
+var ErrEmptyAddress = errors.New("empty address")
+
+func getIPFamily(address string) (txeh.IPFamily, error) {
+	addressIP := net.ParseIP(address)
+	if addressIP == nil {
+		return txeh.IPFamilyV4, ErrEmptyAddress
```
jakubno: isn't this rather a invalid address?

packages/orchestrator/internal/sandbox/network/storage_kv.go
```diff
@@ -70,7 +70,7 @@ func (s *StorageKV) Acquire(_ context.Context) (*Slot, error) {
 	}
 
 	for randomTry := 1; randomTry <= 10; randomTry++ {
-		slotIdx := rand.Intn(s.slotsSize)
+		slotIdx := rand.Intn(s.slotsSize) + 1
```
jakubno: why have you changed this?

jakubno (changes requested):
Otherwise it looks good

packages/orchestrator/internal/sandbox/network/storage_kv.go
djeebus: This was a bug that cursor found, it was here before this PR. Slots are 1-based, but Intn returns 0-based. This brings the numbers into the proper range. Unless I misunderstood something?

djeebus: Moved it into #1317 so that it doesn't complicate this PR

jakubno (changes requested):
Last thing, can you increase envd versions 🙏🏻

jakubno (approved):

PR #1273, opened 2025-09-30
Implement best-of-k algorithm for deleting files

djeebus:
- randomly sample 10k files
- delete the oldest 1k files
- repeat until size is below threshold

also report progress when finding and deleting files, plus some refactor


packages/orchestrator/cmd/clean-nfs-cache/main.go
```diff
@@ -151,116 +158,103 @@ type results struct {
...
 
-func getDiskInfo(ctx context.Context, path string) (diskInfo, error) {
-	// Execute: df <path>
-	cmd := exec.CommandContext(ctx, "df", path)
-	out, err := cmd.CombinedOutput()
-	if err != nil {
-		return diskInfo{}, fmt.Errorf("df command failed: %w: %s", err, strings.TrimSpace(string(out)))
+func reportGetFilesProgress(usedFiles, dupeHits int) {
+	total := usedFiles + dupeHits
+	if total > 0 && total%100 != 0 {
```
jakubno: i think this should be 
```suggestion
	if total > 0 && total%100 == 0 {
```

packages/orchestrator/cmd/clean-nfs-cache/pkg/cache.go
```diff
@@ -0,0 +1,81 @@
...
+				return "", ErrEmptyDir
+			}
+			continue
+		}
+
+		item := items[rand.Intn(len(items))]
+		if !item.isDir {
+			return item.path, nil
+		}
+	}
```
jakubno, line 57: I think if all items are dir, then there's infinite cycle

packages/orchestrator/cmd/clean-nfs-cache/pkg/cache.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+func (c *ListingCache) Decache(path string) {
+	dirName := filepath.Dir(path)
+	delete(c.cache, dirName)
+}
+
+func (c *ListingCache) GetRandomFile() (string, error) {
+	history := []string{c.root}
```
jakubno: you are never adding to the history, seems like a bug

jakubno (changes requested):

packages/orchestrator/cmd/clean-nfs-cache/pkg/atime_osx.go
jakubno, line 1: btw this doesn't work

packages/orchestrator/cmd/clean-nfs-cache/main.go
djeebus: Nope, I only want to report something on every 100th file action

packages/orchestrator/cmd/clean-nfs-cache/pkg/cache.go
djeebus: oh, fair point. okay, let me see what I can do here.

packages/orchestrator/cmd/clean-nfs-cache/pkg/atime_osx.go
djeebus: What doesn't work? I've only got a linux laptop now, but when I coded it up the first time it seemed to function on my macbook. Are you getting an error?

djeebus: This PR did rename a struct, move it into a new package, and expose the fields. It looks like the osx file didn't get the changes, so I've added them there. I assume that's what you were talking about?

PR #1281, opened 2025-10-01
Decrease the number of files scanned per loop

jakubno:


djeebus:
Closing in favor of #1282 , which builds on this and adds another performance improvement.

PR #1285, opened 2025-10-02
Test harness

djeebus:
simple script to generate times for resuming and pausing sandboxes


packages/orchestrator/internal/template/build/core/oci/oci.go
```diff
@@ -237,6 +238,8 @@ func unpackRootfs(ctx context.Context, logger *zap.Logger, srcImage containerreg
 		return fmt.Errorf("while mounting overlayfs with layers: %w", err)
 	}
 	defer func() {
+		syscall.Sync()
```
ValentaTomas: Is this related to the benchmarks, or it is a general bug?

djeebus: That shouldn't have been added in the first place, and was removed a while ago.

PR #1293, opened 2025-10-03
Use only the latest request to envd

jakubno:
Right after the start the envd may not be responsible, we retry until the request succeeds, the problem is the sandbox receives all of these requests, even though they should be cancelled. We set the time based on the request, so if we receive multiple requests it can happen the time jump backwards, this should prevent this

---

> [!NOTE]
> Process init data only from the newest request, adding locking and an AtomicMax utility, with a refactor into SetData and proper 409 on access token conflicts.
> 
> - **API**:
>   - `PostInit`:
>     - Guard updates with `initLock` and apply only if `Timestamp` is newer via `lastSetTime.SetToGreater`.
>     - Delegate mutations to new `SetData`; return `409` on `ErrAccessTokenAlreadySet`.
>   - `SetData` (new): sets system time, env vars, access token, and triggers Hyperloop setup.
>   - Imports `zerolog`; define `ErrAccessTokenAlreadySet`.
> - **Store**:
>   - `API` struct: add `lastSetTime` (`utils.AtomicMax`) and `initLock`; initialize `lastSetTime` in `New`.
> - **Utils**:
>   - Add `utils/atomic.go`: `AtomicMax` with `SetToGreater`.
>   - Add tests in `utils/atomic_test.go`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit dc91242ca0864dd30cb8dcaf23da3d8203bcf75a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/envd/internal/api/init.go
+++ b/packages/envd/internal/api/init.go
@@ -40,8 +40,7 @@ func (a *API) PostInit(w http.ResponseWriter, r *http.Request) {
 		defer a.initLock.Unlock()
 
 		// Update data only if the request is newer or if there's no timestamp at all
-		if initRequest.Timestamp == nil ||
-			a.lastSetTime.SetToGreater(initRequest.Timestamp.UnixNano()) {
+		if initRequest.Timestamp == nil || a.lastSetTime.SetToGreater(initRequest.Timestamp.UnixNano()) {
 			err = a.SetData(logger, initRequest)
 			if err != nil {
 				switch {


packages/envd/internal/api/init.go
```diff
@@ -33,37 +34,16 @@ func (a *API) PostInit(w http.ResponseWriter, r *http.Request) {
...
-				a.envVars.Store(key, value)
-			}
-		}
-
-		if initRequest.AccessToken != nil {
-			if a.accessToken != nil && *initRequest.AccessToken != *a.accessToken {
-				logger.Error().Msg("Access token is already set and cannot be changed")
-				w.WriteHeader(http.StatusConflict)
+				logger.Error().Msgf("Failed to set data: %v", err)
+				w.WriteHeader(http.StatusBadRequest)
```
jakubno: I've changed it to `BadRequest` here, I can probably add a check and revert the initial behavior back to `Conflict`, but not sure if it make sense

packages/envd/internal/utils/atomic.go
```diff
@@ -0,0 +1,26 @@
...
+type AtomicMax struct {
+	val int64
+	mu  sync.Mutex
+}
+
+func NewAtomicMax() *AtomicMax {
+	return &AtomicMax{}
+}
+
+func (a *AtomicMax) Compare(newValue int64) bool {
```
dobrac: Rename to CompareAndSwap

packages/envd/internal/api/init.go
```diff
@@ -33,37 +36,21 @@ func (a *API) PostInit(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 
-		if initRequest.Timestamp != nil {
-			logger.Debug().Msgf("Setting sandbox start time to: %v", *initRequest.Timestamp)
-			ts := unix.NsecToTimespec(initRequest.Timestamp.UnixNano())
-			err = unix.ClockSettime(unix.CLOCK_REALTIME, &ts)
+		// Update data only if the request is newer or if there's no timestamp at all
+		if initRequest.Timestamp == nil ||
```
dobrac: I think you need a lock to prevent concurrent body processing, even if the last timestamp is older than current request timestamp

packages/envd/internal/utils/atomic.go
jakubno: CompareAndSwap has typically different signature, I renamed to SetToGreater

packages/envd/internal/api/init.go
```diff
@@ -33,37 +36,24 @@ func (a *API) PostInit(w http.ResponseWriter, r *http.Request) {
...
+		a.initLock.Lock()
+		defer a.initLock.Unlock()
 
-		if initRequest.AccessToken != nil {
-			if a.accessToken != nil && *initRequest.AccessToken != *a.accessToken {
-				logger.Error().Msg("Access token is already set and cannot be changed")
-				w.WriteHeader(http.StatusConflict)
+		// Update data only if the request is newer or if there's no timestamp at all
+		if initRequest.Timestamp == nil ||
+			(initRequest.Timestamp != nil && a.lastSetTime.SetToGreater(initRequest.Timestamp.UnixNano())) {
```
dobrac: you can simplify the if to:

if initRequest.Timestamp == nil || a.lastSetTime.SetToGreater(initRequest.Timestamp.UnixNano()) {

dobrac (changes requested):
one nit and gtg

dobrac (approved):

PR #1298, opened 2025-10-06
Send sandbox routing data to shared Redis storage via catalog impl

sitole:
- Adjustments in sandbox catalog impl to lower memory cache ttl, removed global lock
- Logic that was currently handling storing data in Redis is now also using sandbox catalog impl


packages/client-proxy/internal/edge/handlers/sandbox-catalog-create-entry.go
```diff
@@ -20,16 +20,24 @@ func (a *APIStore) V1SandboxCatalogCreate(c *gin.Context) {
...
 		return
 	}
 
 	_, span := tracer.Start(ctx, "create-sandbox-catalog-entry-handler")
 	defer span.End()
 
+	o, ok := a.orchestratorPool.GetOrchestrator(body.OrchestratorID)
+	if !ok {
+		a.sendAPIStoreError(c, http.StatusBadRequest, "Orchestrator not found")
+		telemetry.ReportError(ctx, "orchestrator not found", err)
```
jakubno:
```suggestion
		telemetry.ReportError(ctx, "orchestrator not found", nil)
```

There's no error or create a new one with fmt.Errorf

packages/shared/pkg/sandbox-catalog/catalog_redis.go
jakubno, line 1: can you prefix the key? 🙏🏻

packages/client-proxy/internal/edge/sandboxes/catalog_redis.go
jakubno, line 58: can we at least log the actual error 🙏🏻

packages/shared/pkg/sandbox-catalog/catalog_redis.go
jakubno: Can you return this error only on redis.Nil?

packages/shared/pkg/sandbox-catalog/catalog_redis.go
```diff
@@ -15,35 +14,28 @@ import (
 	"github.com/e2b-dev/infra/packages/shared/pkg/logger"
 )
 
-var tracer = otel.Tracer("github.com/e2b-dev/infra/packages/client-proxy/internal/edge/sandboxes")
+var tracer = otel.Tracer("github.com/e2b-dev/infra/packages/shared/pkg/sandbox-catalog")
```
jakubno: i think it should be in catalog if it's shared

jakubno (changes requested):

jakubno (approved):

PR #1313, opened 2025-10-07
Remove dns infrastructure for sandbox routing

sitole:
- Removed logic for DNS-based routing in client proxy
- Removes DNS server with legacy Redis backup storage
- Refactors use of sandbox routing catalog in api


sitole:
We can deploy this when all running sandboxes are registered in the Redis catalog.

packages/api/internal/orchestrator/orchestrator.go
```diff
@@ -77,13 +78,11 @@ func New(
 		return nil, err
 	}
 
-	dnsServer := dns.New(redisClient)
-
-	if env.IsLocal() {
-		zap.L().Info("Running locally, skipping starting DNS server")
+	var routingCatalog e2bcatalog.SandboxesCatalog
+	if redisClient != nil && !reflect.ValueOf(redisClient).IsNil() {
```
jakubno: can you try if this is still needed?

sitole: works without reflext, != nill with match that properly

PR #1314, opened 2025-10-07
Implement ingress for easier services exposing

sitole:
- Deploys Traefik v3.5 as a Nomad job with configuration for auto-discovery of Nomad services.
- Ingress load balancer that will handle traffic for services behind Traefik.
- Move "additional domains" to the security store so they can be taken from other IaC projects and correctly propagate DNS for each domain that should be routed with Traefik

---

> [!NOTE]
> Deploys Traefik-based ingress with a managed HTTPS load balancer and moves additional domain routing to Secret Manager, adding ingress vars and wiring across modules.
> 
> - **Ingress**:
>   - Deploy Traefik (`nomad/jobs/ingress.hcl`) via `nomad_job.ingress` with configurable `count`, CPU/memory, and ports.
>   - Add GCP HTTPS LB for ingress (`nomad-cluster/network/ingress.tf`): health check, backend service, URL map, target HTTPS proxy, global IP, forwarding rule; reference cert map.
>   - Open firewall and expose named port for `ingress` (`nomad-cluster/network/main.tf`, `nodepool-api.tf`).
> - **Domains configuration**:
>   - Create Secret Manager secret `routing-domains` and initial version (`init/main.tf`); output `routing_domains_secret_name` (`init/outputs.tf`).
>   - Read `routing-domains` and merge with env `ADDITIONAL_DOMAINS` to form `local.additional_domains` (`main.tf`); pass to cluster.
> - **Variables & plumbing**:
>   - Introduce `ingress_port` (with defaults) and `ingress_count` (`variables.tf`); surface through modules (`main.tf`, `nomad/variables.tf`, `nomad-cluster/variables.tf`, `nomad-cluster/network/variables.tf`).
>   - Wire `ingress_port` to network, firewall, and instance group named ports; pass `ingress_count` to Nomad job.
>   - Makefile: add `INGRESS_COUNT` to Terraform var propagation.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8d137bc61f18604a1f6e00523943e35fe1e22b01. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/iac/provider-gcp/nomad/jobs/ingress.hcl
+++ b/iac/provider-gcp/nomad/jobs/ingress.hcl
@@ -46,8 +46,6 @@ job "ingress" {
     task "ingress" {
       driver = "docker"
 
-      # If we need more than 30s we will need to update the max_kill_timeout in nomad
-      # https://developer.hashicorp.com/nomad/docs/configuration/client#max_kill_timeout
       %{ if update_stanza }
         kill_timeout = "24h"
       %{ endif }


iac/provider-gcp/nomad-cluster/network/ingress.tf
```diff
@@ -0,0 +1,58 @@
...
+
+  load_balancing_scheme = "EXTERNAL_MANAGED"
+  locality_lb_policy    = "ROUND_ROBIN"
+
+  backend {
+    group = var.api_instance_group
+  }
+}
+
+resource "google_compute_url_map" "ingress" {
```
djeebus, line 34: Can we get a better name here? I think we now have two load balancers, "ingress" and "orch_map", neither of their names help understand what they do. Maybe "traefik" and "direct"?

sitole: I like ingress as its common name and it describes what it is. Yes, "orch_map" is, in my opinion, a mistake, as it no longer makes sense. Ideally, I would like to transition away from the current load balancer once the migration is complete/rename it to something like "ingress-sandboxes" or a similar name to distinguish better.

I don't like to call it Traefik, as we can switch the ingress backend at any time in the future, but I'm okay with you coming up with a better name.

dobrac: a note here, if we want to rename the `orch_map` to `ingress-sandboxes`, maybe we should name this ingress something like `ingress-api` or `ingress-management` or `ingress-services`

iac/provider-gcp/nomad/jobs/ingress.hcl
```diff
@@ -0,0 +1,97 @@
...
+        timeout  = "3s"
+        port     = "${ingress_port}"
+      }
+    }
+
+    task "ingress" {
+      driver = "docker"
+
+      # If we need more than 30s we will need to update the max_kill_timeout in nomad
+      # https://developer.hashicorp.com/nomad/docs/configuration/client#max_kill_timeout
```
dobrac: is this comment up to date?

iac/provider-gcp/nomad/main.tf
```diff
@@ -68,6 +68,27 @@ resource "docker_image" "db_migrator_image" {
   platform      = "linux/amd64/v8"
 }
 
+resource "nomad_job" "ingress" {
+  jobspec = templatefile("${path.module}/jobs/ingress.hcl",
+    {
+      count         = 1
```
dobrac: we should do at least 2 or more to be able to update

iac/provider-gcp/nomad/main.tf
```diff
@@ -68,6 +68,27 @@ resource "docker_image" "db_migrator_image" {
...
+      update_stanza = var.api_machine_count > 1
+      cpu_count     = 1
+      memory_mb     = 512
```
dobrac, line 77: are these enough for our traffic?

iac/provider-gcp/Makefile
```diff
@@ -34,7 +34,6 @@ tf_vars := 	TF_VAR_environment=$(TERRAFORM_ENVIRONMENT) \
 	$(call tfvar, GCP_REGION) \
 	$(call tfvar, GCP_ZONE) \
 	$(call tfvar, DOMAIN_NAME) \
-	$(call tfvar, ADDITIONAL_DOMAINS) \
```
dobrac, line 39: note here, this is a breaking change and should be handled with proper migration path

sitole: Do we have any? This type transformation @jakubno we handles manually.

sitole: Maybe I can push attribute before this pr

dobrac: can we maybe do the secret default value from the env var? Is that a bad idea?

iac/provider-gcp/nomad-cluster/network/ingress.tf
sitole: I still don’t like that we would need two load balancers just because we cannot filter sandbox traffic. Will look into again tomorrow

Yep, we can rename ingress to something else. Iam not sure about management/api as we can use it for something different in future. Ingress services sounds okay to me.

dobrac: I though it's actually quite nice to have separate LBs for user's sandbox traffic and our services traffic (different limitations, limits, HTTP support, etc), but *maybe* it's unnecessary

iac/provider-gcp/Makefile
sitole: Imho during migration period we can merge server + tf var lists.

iac/provider-gcp/nomad-cluster/network/ingress.tf
sitole: Ideally, we should be able to match sandbox traffic to different rules (now it's catch-all fallback) so we can apply different limits/armor rules to them, then we don't need to have different LBs.

For supporting newer versions of HTTP, etc, we can still relatively easily migrate everything, and I'm not sure if we would need some special LB that cannot handle both sandbox and services traffic.

iac/provider-gcp/nomad/main.tf
sitole: Yes, more than enough for now. When we proceed and start routing sandbox traffic, we need to revisit this, with memory ideally set to 1024 min and 2048 max.

iac/provider-gcp/Makefile
sitole: Added compatibility during the migration period that will accept both env and secret and merge them together.

iac/provider-gcp/nomad-cluster/network/ingress.tf
sitole: Okay, discovered that GCP Armor policy rule allows you to filter based on host regex, so we can use one shared backend and apply dynamic rules based on the domain there. This will solve our issue with needing two load balancers. I would stick with `ingress` naming, as after migration is completed, we can remove `orch-map` as outdated. 

Example of a regexp that can catch sandbox traffic and apply rate limiting. In the same way, we can create rules for API limiting and other related restrictions. The good thing is that this only appends rules to already existing security policy, so we can push rules even from a private monorepo that will handle block/rate limit for services that are not open source.

```terraform
resource "google_compute_security_policy_rule" "sandbox-throttling-ip" {
  security_policy = google_compute_security_policy.default["session"].name
  action          = "throttle"
  priority        = "500"

  match {
    expr {
      expression = <<-EOT
request.headers["host"].matches("^(?i)[0-9]+-[a-z0-9-]+\\.e2b-jirka\\.dev$")
EOT
    }
  }

  rate_limit_options {
    conform_action = "allow"
    exceed_action  = "deny(429)"

    enforce_on_key = ""

    enforce_on_key_configs {
      enforce_on_key_type = "IP"
    }

    rate_limit_threshold {
      count        = 40000
      interval_sec = 60
    }
  }

  description = "Requests to sandboxes from IP address"
}
```

sitole: Internal docs that will contain all info related to migration and current state -> https://www.notion.so/e2bdev/Ingress-Migration-288b8c296873807a8264f1615602d11d

PR #1315, opened 2025-10-07
Fix build status error check condition

dobrac:
Fix build status error check condition


ValentaTomas (approved):

djeebus:
Can we get a test for this?

dobrac:
Actually, there issue is different 🤦

PR #1324, opened 2025-10-08
Fix concurrent resumes

jakubno:
> [!NOTE]
> Serialize concurrent sandbox start/resume via reservation result waiting; return API-safe sandbox; propagate cluster sandbox domain; add tests.
> 
> - **Sandbox orchestration**:
>   - Serialize concurrent starts/resumes by enhancing `Reserve` to return a result-aware release callback and tracking `StartResult`; on `AlreadyBeingStartedError`, wait for the first start to finish and return that sandbox.
>   - Change `CreateSandbox` to return internal `sandbox.Sandbox`; convert to API model via new `Sandbox.ToAPISandbox()` in handlers.
>   - Implement `APIError` as `error` with `Error()`.
> - **Domain propagation**:
>   - Add `Node.SandboxDomain` and include domain in `sandbox.NewSandbox`; pass cluster `SandboxDomain` to nodes and surface as `Sandbox.Domain`.
> - **Logging**:
>   - Include `SandboxID` in secure token error logs during resume.
> - **Tests**:
>   - Add integration test for concurrent resumes ensuring all callers see a running sandbox and one 201 response.
>   - Update memory reservation tests for new release signature.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 63bcd3f5e63d31dd509e112301695ea24d5c7b34. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api/internal/sandbox/store/memory/reservation.go
+++ b/packages/api/internal/sandbox/store/memory/reservation.go
@@ -26,11 +26,11 @@ func NewReservationCache() *ReservationCache {
 	}
 }
 
-func (r *ReservationCache) insertIfAbsent(sandboxID string, team uuid.UUID, start *utils.SetOnce[sandbox.Sandbox]) bool {
+func (r *ReservationCache) insertIfAbsent(sandboxID string, team uuid.UUID, startResult *utils.SetOnce[sandbox.Sandbox]) bool {
 	return r.reservations.InsertIfAbsent(sandboxID, &Reservation{
 		team:        team,
 		sandboxID:   sandboxID,
-		startResult: start,
+		startResult: startResult,
 	})
 }
 


packages/api/internal/orchestrator/create_instance.go
```diff
@@ -66,7 +66,29 @@ func (o *Orchestrator) CreateSandbox(
...
+						Code:      http.StatusInternalServerError,
+						ClientMsg: "Error waiting for sandbox to start",
+						Err:       err,
+					}
+				}
+
+				return sbxData.ToAPISandbox(), nil
+			}
+
+			// TODO: Handle this error better
```
jakubno, line 90: this should happen only if in the first get in sandbox_resume, the sandbox isn't running (it's resuming) and now it's running. But this is handled well in SDK.

packages/api/internal/orchestrator/create_instance.go
```diff
@@ -83,7 +105,14 @@ func (o *Orchestrator) CreateSandbox(
...
 
 	telemetry.ReportEvent(ctx, "Reserved sandbox for team")
-	defer releaseTeamSandboxReservation()
+	var instanceInfo sandbox.Sandbox
+	defer func() {
+		if apiErr != nil {
+			releaseTeamSandboxReservation(instanceInfo, apiErr)
+		} else {
+			releaseTeamSandboxReservation(instanceInfo, nil)
+		}
```
dobrac, line 114: the if here is probably redundant

packages/api/internal/orchestrator/create_instance.go
```diff
@@ -83,7 +105,14 @@ func (o *Orchestrator) CreateSandbox(
 	}
 
 	telemetry.ReportEvent(ctx, "Reserved sandbox for team")
-	defer releaseTeamSandboxReservation()
+	var instanceInfo sandbox.Sandbox
+	defer func() {
+		if apiErr != nil {
+			releaseTeamSandboxReservation(instanceInfo, apiErr)
```
dobrac: how about using the sbx variable instead of the instanceInfo to better symbolize, that either one of them is always set (sbx vs apiErr)

(you can then remove the var instanceInfo definition altogether)

packages/api/internal/orchestrator/nodemanager/sandboxes.go
```diff
@@ -13,7 +13,7 @@ import (
 	"github.com/e2b-dev/infra/packages/shared/pkg/consts"
 )
 
-func (n *Node) GetSandboxes(ctx context.Context) ([]sandbox.Sandbox, error) {
+func (n *Node) GetSandboxes(ctx context.Context, sandboxDomain *string) ([]sandbox.Sandbox, error) {
```
dobrac: the sandbox domain is not specific per sandbox, but per cluster. Hence I think it might be better to handle it the same as the clusterID (or just read it from the cluster when required, rather than passing around)

```
	cluster, ok := o.clusters.GetClusterById(*team.Team.ClusterID)
		if !ok {
			return nil, &api.APIError{
				Code:      http.StatusInternalServerError,
				ClientMsg: "Error while looking for sandbox cluster information",
				Err:       fmt.Errorf("cannot access cluster %s associated with team id %s that spawned sandbox %s", *team.Team.ClusterID, team.Team.ID, sandboxID),
			}
		}

		sbxDomain = cluster.SandboxDomain
```

packages/api/internal/orchestrator/nodemanager/sync.go
```diff
@@ -13,7 +13,7 @@ import (
 
 const syncMaxRetries = 4
 
-func (n *Node) Sync(ctx context.Context, instanceCache *memory.Store) {
+func (n *Node) Sync(ctx context.Context, instanceCache *memory.Store, sandboxDomain *string) {
```
dobrac: (same here, I think we can get it from the node.ClusterID)

packages/api/internal/orchestrator/create_instance.go
```diff
@@ -66,7 +66,29 @@ func (o *Orchestrator) CreateSandbox(
 				Err: fmt.Errorf("team '%s' has reached the maximum number of instances (%d)", team.Team.ID, team.Tier.ConcurrentInstances),
 			}
 		case errors.As(err, &alreadyErr):
-			zap.L().Warn("sandbox already being started", logger.WithSandboxID(sandboxID), zap.Error(err))
+			zap.L().Info("sandbox already being started", logger.WithSandboxID(sandboxID), zap.Error(err))
+			if alreadyErr.Start != nil {
+				sbxData, err := alreadyErr.Start.WaitWithContext(ctx)
+				if err != nil {
+					zap.L().Error("Error waiting for sandbox to start", zap.Error(err), logger.WithSandboxID(sandboxID))
```
dobrac: to prevent polluting the logs, the context timeout is probably a warning, not error

packages/api/internal/sandbox/store/memory/reservation.go
```diff
@@ -57,37 +62,59 @@ func (s *Store) list(teamID uuid.UUID) (sandboxIDs []string) {
...
 	}
 
-	return func() {
+	return func(sbx sandbox.Sandbox, err error) {
+		var setErr error
+		if err != nil {
+			setErr = start.SetError(err)
+		} else {
+			setErr = start.SetValue(sbx)
+		}
```
dobrac: you can use start.SetResult instead

packages/api/internal/sandbox/store/memory/reservation.go
```diff
@@ -2,14 +2,18 @@ package memory
...
 	"github.com/e2b-dev/infra/packages/api/internal/sandbox"
+	"github.com/e2b-dev/infra/packages/shared/pkg/logger"
 	"github.com/e2b-dev/infra/packages/shared/pkg/smap"
+	"github.com/e2b-dev/infra/packages/shared/pkg/utils"
 )
 
 type Reservation struct {
 	sandboxID string
 	team      uuid.UUID
+	start     *utils.SetOnce[sandbox.Sandbox]
```
dobrac: maybe rename this to `result` or `startResult`?

packages/api/internal/orchestrator/create_instance.go
jakubno: it isn't 😢  that's just go. Typed nil pointer isn't nil

packages/api/internal/orchestrator/nodemanager/sandboxes.go
jakubno: But I don't have access to cluster nor orchestrator from node

PR #1325, opened 2025-10-08
Stop sending team and trace ids to envd

sitole:
- Envd is now not expecting `traceID` in mmds, `teamID` was removed in the past.
- Actually, stop these two fields from sending from the orchestrator to the sandbox MMDs server.
- Cleanup orchestrator sbx metadata previously used for two similar tasks (orchestrator sbx logger and mmds metadata).

Older versions of envd are safe, as the mmds struct will not find teamID/traceID and will fall back to an empty string. 
The Trace ID is not used at all, and the Team ID is overridden, as the Hyperloop proxy server handles sandbox logs forwarding.


packages/orchestrator/internal/sandbox/fc/process.go
```diff
@@ -324,19 +324,13 @@ func (p *Process) Create(
...
 }
 
-func (p *Process) Resume(
-	ctx context.Context,
-	mmdsMetadata *MmdsMetadata,
-	uffdSocketPath string,
-	snapfile template.File,
-	uffdReady chan struct{},
-) error {
+func (p *Process) Resume(ctx context.Context, sbxMetadata sbxlogger.SandboxMetadata, uffdSocketPath string, snapfile template.File, uffdReady chan struct{}, slot *network.Slot) error {
```
jakubno: can you leave it multiline?

PR #1350, opened 2025-10-13
Add orchestrator proxy retries to handle port forwarding delay

Code42Cate:
**Problem**
When a process binds to localhost inside a sandbox, envd scans every 1s to detect it and starts socat to forward the port to the host IP (169.254.0.21). This creates a race condition where HTTP requests can arrive before forwarding is active, causing immediate connection failures.

**Solution**
Added retry logic in the proxy's DialContext with linear backoff:
- 5 retry attempts with backoff: 100ms, 200ms, 300ms, 400ms
- Total retry window: ~1000ms (covers envd's 1s scan interval)
- The amount of retries is configurable. Only the orchestrator proxy has 5 retries, the client proxy has 1

**Behavior**
Best case: Port already forwarded → immediate success (0ms)
Average: Request arrives mid-scan → 1-2 retries, ~300ms delay
Worst case (success): Request arrives just before scan → 4-5 retries, ~1000ms delay
Worst case (failure): Port doesn't exist → ~1000ms then error

**Testing**
Added TestProxyRetriesOnDelayedBackendStartup that simulates a 300ms backend startup delay. Test passes with retries, fails without them. You can verify that behavior by reducing the attempts to 1. The tests will then fail

---

> [!NOTE]
> Adds linear-backoff connection retries to proxy DialContext (5 for orchestrator, 1 for client) and updates pool/proxy APIs and tests, including a delayed-backend startup test.
> 
> - **Proxy Core**:
>   - Add `MaxConnectionAttempts` and constants `ClientProxyRetries` (1) and `SandboxProxyRetries` (5).
>   - Update `reverseproxy.New(port, maxConnectionAttempts, idleTimeout, getDestination)` and pass attempts to pool.
> - **Pool/Client**:
>   - Implement retry loop in `pool/client.go` `DialContext` with linear backoff (100–400ms) and context-aware exits.
>   - Extend `pool.New(maxClientConns, maxConnectionAttempts, idleTimeout)` and propagate attempts in `ProxyPool.Get` to `newProxyClient`.
> - **Orchestrator/Client Proxies**:
>   - Wire retries: orchestrator uses `SandboxProxyRetries`; client uses `ClientProxyRetries`.
> - **Tests**:
>   - Add `TestProxyRetriesOnDelayedBackendStartup` to validate retry behavior.
>   - Update test helpers to pass retry constant; replace several error checks with `require.NoError`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 95393d624a151af1525cb1d21f500c3b2fc85d9e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/shared/pkg/proxy/proxy_test.go
+++ b/packages/shared/pkg/proxy/proxy_test.go
@@ -418,13 +418,6 @@ func TestProxyRetriesOnDelayedBackendStartup(t *testing.T) {
 
 	// Start backend after a delay (simulating envd port forwarding)
 	go func() {
-		defer func() {
-			// Ensure channel is always closed to prevent test hangs
-			if len(backendReady) == 0 {
-				backendReady <- backendResult{nil, errors.New("goroutine exited without sending result")}
-			}
-		}()
-
 		// Wait 300ms before starting the backend (should succeed on retry 2 or 3)
 		time.Sleep(300 * time.Millisecond)
 


packages/client-proxy/internal/proxy/proxy.go
```diff
@@ -52,6 +52,8 @@ func catalogResolution(ctx context.Context, sandboxId string, c catalog.Sandboxe
 func NewClientProxy(meterProvider metric.MeterProvider, serviceName string, port uint16, catalog catalog.SandboxesCatalog) (*reverseproxy.Proxy, error) {
 	proxy := reverseproxy.New(
 		port,
+		// Retries that are needed to handle port forwarding delays in sandbox envd are handled by the orchestrator proxy
+		1,
```
Code42Cate: I personally have no issues with this magic number, if you want that in a constant thats understandable though. Same in the other proxy!

packages/shared/pkg/proxy/proxy_test.go
```diff
@@ -430,3 +431,99 @@ func TestProxyDoesNotReuseConnectionsWhenBackendChanges(t *testing.T) {
...
+	}()
+
+	// Make request - this should retry and eventually succeed
+	proxyURL := fmt.Sprintf("http://127.0.0.1:%d/hello", port)
+	start := time.Now()
+
+	resp, err := httpGet(t, proxyURL)
+	if err != nil {
+		t.Fatalf("failed to GET from proxy: %v", err)
+	}
```
djeebus: testify can simplify your life:

```suggestion
	require.NoError(t, err)
```

You can avoid if statements and coming up with fancy messages, and let the line of code do the talking. Lots of other places in this file you can do this too.

Code42Cate: changed all in this file

PR #1361, opened 2025-10-16
Split seekable files from non-seekable files

djeebus:
This allows us to cache, compress, etc.

- Add a new `OpenSeekableObject` method to the storage provider that is used to read seekable files
- Remove the `ReadAt` method from the `ObjectProvider` interface (only used by `SeekableObjectProvider`).
- Remove the `Delete` method from the `ObjectProvider` interface (never used).
- Add a `SeekableObjectProvider` interface.
- Add new `ObjectType` and `SeekableObjectType` types that defines which file type is being passed in, in order to help the compiler prevent mistakes

---

> [!NOTE]
> Introduces separate seekable/non-seekable storage interfaces with typed OpenObject/OpenSeekableObject, updates providers, caching, orchestrator code, mocks, and tests accordingly.
> 
> - **Storage API**:
>   - Split interfaces into `ObjectProvider` (non-seekable) and `SeekableObjectProvider` (seekable); removed `ReadAt`/`Delete` from `ObjectProvider`, added `Exists`.
>   - Added `ObjectType` and `SeekableObjectType` enums; updated `StorageProvider` with `OpenObject(ctx, path, ObjectType)` and `OpenSeekableObject(ctx, path, SeekableObjectType)`.
>   - Refactored caching: new `CachedObjectProvider` and `CachedSeekableObjectProvider`; moved chunked read/size logic to seekable cache; reorganized cache helpers.
> - **Providers (GCP/AWS/FS)**:
>   - Implement new interfaces and methods; added `Exists`; adjusted `OpenObject`/`OpenSeekableObject` signatures.
> - **Orchestrator**:
>   - Updated commands and template code to pass object kinds and use `OpenSeekableObject` for memfile/rootfs and `OpenObject` for headers/metadata/snapfile.
>   - `storage_diff`: constructor now returns `(*StorageDiff, error)`; added diff→object type mapping; use `OpenSeekableObject`.
>   - Template build/upload paths now typed; server upload checks `Exists` instead of `Size`.
> - **Tests & Mocks**:
>   - Replaced old mocks with `mockobjectprovider` and `mockseekableobjectprovider`; updated unit tests for new interfaces and behaviors.
> - **Config**:
>   - `.mockery.yaml` updated to generate mocks for new interfaces.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 77c3db50ff555b100c49571a44bc19772ae263ee. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/orchestrator/internal/template/server/upload_layer_files_template.go
+++ b/packages/orchestrator/internal/template/server/upload_layer_files_template.go
@@ -13,7 +13,7 @@ import (
 const signedUrlExpiration = time.Minute * 30
 
 func (s *ServerStore) InitLayerFileUpload(ctx context.Context, in *templatemanager.InitLayerFileUploadRequest) (*templatemanager.InitLayerFileUploadResponse, error) {
-	_, childSpan := tracer.Start(ctx, "template-create")
+	ctx, childSpan := tracer.Start(ctx, "template-create")
 	defer childSpan.End()
 
 	// default to scope by template ID
@@ -23,7 +23,7 @@ func (s *ServerStore) InitLayerFileUpload(ctx context.Context, in *templatemanag
 	}
 
 	path := paths.GetLayerFilesCachePath(cacheScope, in.GetHash())
-	obj, err := s.buildStorage.OpenObject(ctx, path, storage.LayerObjectType)
+	obj, err := s.buildStorage.OpenObject(ctx, path, storage.BuildLayerFileObjectType)
 	if err != nil {
 		return nil, fmt.Errorf("failed to open layer files cache: %w", err)
 	}
--- a/packages/shared/pkg/storage/storage.go
+++ b/packages/shared/pkg/storage/storage.go
@@ -52,7 +52,7 @@ const (
 	RootFSHeaderObjectType
 	SnapfileObjectType
 	MetadataObjectType
-	LayerObjectType
+	BuildLayerFileObjectType
 	LayerMetadataObjectType
 )
 
--- a/packages/orchestrator/internal/template/build/commands/copy.go
+++ b/packages/orchestrator/internal/template/build/commands/copy.go
@@ -80,7 +80,7 @@ func (c *Copy) Execute(
 	}
 
 	// 1) Download the layer tar file from the storage to the local filesystem
-	obj, err := c.FilesStorage.OpenObject(ctx, paths.GetLayerFilesCachePath(c.CacheScope, step.GetFilesHash()), storage.LayerObjectType)
+	obj, err := c.FilesStorage.OpenObject(ctx, paths.GetLayerFilesCachePath(c.CacheScope, step.GetFilesHash()), storage.BuildLayerFileObjectType)
 	if err != nil {
 		return metadata.Context{}, fmt.Errorf("failed to open files object from storage: %w", err)
 	}


packages/orchestrator/cmd/inspect-header/main.go
```diff
@@ -38,7 +39,7 @@ func main() {
 		log.Fatalf("failed to get storage provider: %s", err)
 	}
 
-	obj, err := storage.OpenObject(ctx, storagePath)
+	obj, err := storage.OpenObject(ctx, storagePath, fileType)
```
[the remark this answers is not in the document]
djeebus: The previous `log.Fatalf` prevents it from being unassigned by the time it gets here.

packages/orchestrator/cmd/inspect-header/main.go
```diff
@@ -22,6 +22,7 @@ func main() {
 	}
 
 	var storagePath string
+	var objectType storage.ObjectType
```
dobrac, line 25: this var is probably not initialized based on the file kind

packages/orchestrator/internal/sandbox/build/storage_diff.go
```diff
@@ -40,17 +42,30 @@ func newStorageDiff(
...
+}
+
+func storageObjectType(diffType DiffType) storage.SeekableObjectType {
+	switch diffType {
+	case Memfile:
+		return storage.MemfileObjectType
+	case Rootfs:
+		return storage.RootFSObjectType
+	default:
+		panic(fmt.Sprintf("unknown diff type: %s", diffType))
```
dobrac: I would rather prefer returning an error instead of the panic

packages/orchestrator/internal/sandbox/template/storage.go
```diff
@@ -23,6 +23,28 @@ type Storage struct {
...
 }
 
+func storageHeaderObjectType(diffType build.DiffType) storage.ObjectType {
+	switch diffType {
+	case build.Memfile:
+		return storage.MemfileHeaderObjectType
+	case build.Rootfs:
+		return storage.RootFSHeaderObjectType
+	default:
+		panic(fmt.Sprintf("unknown diff type: %s", diffType))
```
dobrac: same here and also in the other mapper, return an error or some default type

packages/shared/pkg/storage/storage.go
```diff
@@ -52,15 +76,27 @@ type ReaderAtCtx interface {
 	ReadAt(ctx context.Context, p []byte, off int64) (n int, err error)
 }
 
-type StorageObjectProvider interface {
+type ObjectProvider interface {
```
dobrac, line 79: how do we read non-seekable files?

dobrac: why is there missing Size() for the non-seekable files?

packages/shared/pkg/storage/storage.go
```diff
@@ -52,15 +76,27 @@ type ReaderAtCtx interface {
...
+
+	// read
 	WriterToCtx
-	ReaderAtCtx
 
+	// utility
+	Exists(ctx context.Context) (bool, error)
+}
+
+type SeekableObjectProvider interface {
```
dobrac, line 91: why is SeekableObjectProvider not a superset of the object provider?

packages/shared/pkg/storage/storage.go
```diff
@@ -33,10 +36,31 @@ const (
...
+
+	RootFSObjectType       SeekableObjectType = "rootfs"
+	RootFSHeaderObjectType ObjectType         = "rootfs-header"
+
+	SnapfileObjectType ObjectType = "snapfile"
+
+	MetadataObjectType ObjectType = "metadata"
+
+	LayerObjectType         ObjectType = "layer"
+	LayerMetadataObjectType ObjectType = "layer-metadata"
```
dobrac: I think the type should be based on the feature set, not the file name

packages/orchestrator/internal/sandbox/template_build.go
```diff
@@ -105,7 +105,7 @@ func (t *TemplateBuild) uploadRootfs(ctx context.Context, rootfsPath string) err
 
 // Snap-file is small enough so we don't use composite upload.
 func (t *TemplateBuild) uploadSnapfile(ctx context.Context, path string) error {
-	object, err := t.persistence.OpenObject(ctx, t.files.StorageSnapfilePath())
+	object, err := t.persistence.OpenSeekableObject(ctx, t.files.StorageSnapfilePath(), storage.SnapfileName)
```
dobrac: snapfile is probably not seekable

packages/shared/pkg/storage/storage_aws.go
```diff
@@ -246,6 +257,11 @@ func (a *AWSBucketStorageObjectProvider) Size(ctx context.Context) (int64, error
 	return *resp.ContentLength, nil
 }
 
+func (a *AWSBucketStorageObjectProvider) Exists(ctx context.Context) (bool, error) {
+	_, err := a.Size(ctx)
+	return err == nil, err
+}
```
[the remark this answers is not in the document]
ValentaTomas: Missing the fix in GCP + maybe implement not via Size.

packages/shared/pkg/storage/storage.go
djeebus: `WriteTo` is actually `ReadFrom`. The names are awkward, but that's the method we call to read from the remote and write to a local. We don't have Size here because ... we don't need it! We never call it.

packages/shared/pkg/storage/storage.go
djeebus: The functionality we need from it is different. 
- SeekableObjectProvider gets written in parallel chunks, then gets read 4 MB at a time. 
- ObjectProvider gets written without parallelism (small files), then gets written to disk without chunking
There's really no overlap between the two, which is why we can do this.

packages/orchestrator/internal/sandbox/template_build.go
djeebus: :facepalm:

packages/shared/pkg/storage/storage.go
djeebus: The purpose of this PR is to provide some compile-time safety to the different types of files. When we added caching, we wrote some files to the cache as a full file, then tried to read them back via chunking. This is trying to prevent that and similar problems in the future.

packages/orchestrator/internal/sandbox/template_build.go
djeebus: Using strings as typed constants doesn't prevent compilation. Swapping to ints so that this actually prevents us from doing bad things.

packages/shared/pkg/storage/storage.go
dobrac: I see, thats confusing, thx!

packages/orchestrator/internal/template/server/upload_layer_files_template.go
```diff
@@ -22,7 +23,7 @@ func (s *ServerStore) InitLayerFileUpload(ctx context.Context, in *templatemanag
 	}
 
 	path := paths.GetLayerFilesCachePath(cacheScope, in.GetHash())
-	obj, err := s.buildStorage.OpenObject(ctx, path)
+	obj, err := s.buildStorage.OpenObject(ctx, path, storage.LayerObjectType)
```
dobrac: let's rename the `LayerObjectType` to "BuildLayerFileObjectType` or `LayerFileObjectType`

packages/orchestrator/internal/template/server/upload_layer_files_template.go
dobrac: I know it's not related to this PR, but can you please fix this missing context? 🙏

PR #1362, opened 2025-10-16
Extract goroutine management and clean up into a separate package

djeebus:
This allows us to test and standardize how we start, monitor, cancel, and clean up our goroutines. This PR only adds it to orchestrator, but we should add this to api, client-proxy, and the rest as well.

---

> [!NOTE]
> Adds a shared supervisor to manage background tasks and cleanup, and migrates orchestrator to use it instead of custom errgroup/signal/closer logic.
> 
> - **Shared Library**:
>   - **`packages/shared/pkg/supervisor`**: New task supervisor with `Task`, `Options`, and `Run` to start background tasks, handle OS signals, cancel via context, and run reverse-order cleanups; includes tests.
> - **Orchestrator**:
>   - **`packages/orchestrator/main.go`**: Replace `errgroup`/signal/`closer` scaffolding with a `[]supervisor.Task` for telemetry, loggers (with `cleanupLogger`), feature flags, limiter, Redis/pubsub, sandbox observer, proxy, NBD device pool, network pool, hyperloop server, cmux, HTTP, and gRPC servers, plus drain steps; invoke `supervisor.Run` for lifecycle and error handling.
>   - Minor cleanup of unused imports and logs.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f09069b8b91014268b1fc636f2354907d5409e44. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/shared/pkg/supervisor/main.go
```diff
@@ -0,0 +1,166 @@
...
+type Runner struct {
+	tasks  []Task
+	logger *zap.Logger
+
+	mu          sync.Mutex
+	started     bool
+	closed      bool
+	doneCh      chan taskResult
+	wg          sync.WaitGroup
+	cancelFunc  context.CancelFunc
```
dobrac: Not a fan of having this context on the struct. Also, the cancelation is incorrect and differs from the initial implementation drastically

packages/shared/pkg/supervisor/main.go
```diff
@@ -0,0 +1,166 @@
...
+		}
+
+		logger.Info("task finished", zap.Error(err))
+	}(task)
+}
+
+func (s *Runner) Close(ctx context.Context) error {
+	// Idempotent: if Run already performed cleanup, nothing to do
+	s.mu.Lock()
+	if s.cleanupsRun {
```
dobrac: can be done using atomic boolean

packages/shared/pkg/supervisor/main.go
```diff
@@ -0,0 +1,166 @@
...
+)
+
+type Runner struct {
+	tasks  []Task
+	logger *zap.Logger
+
+	mu          sync.Mutex
+	started     bool
+	closed      bool
+	doneCh      chan taskResult
```
dobrac: how about using the setonce instead? It could then handle multiple variables on this struct

packages/shared/pkg/supervisor/main.go
```diff
@@ -0,0 +1,166 @@
...
+	"syscall"
+
+	"go.uber.org/zap"
+)
+
+type Runner struct {
+	tasks  []Task
+	logger *zap.Logger
+
+	mu          sync.Mutex
```
dobrac: do we even need the mutex?

dobrac (changes requested):

packages/shared/pkg/supervisor/main.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	baseLogger.Info("starting tasks", zap.Int("count", len(opts.Tasks)))
+	for _, task := range opts.Tasks {
+		startTask(ctx, task)
+	}
+
+	// Setup OS signal listener
+	sigs := make(chan os.Signal, 1)
+	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
```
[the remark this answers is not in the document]
djeebus: This is intentional, USR1 was just for testing

packages/orchestrator/main.go
```diff
@@ -463,89 +441,88 @@ func run(config cfg.Config) (success bool) {
...
-	if err := g.Wait(); err != nil && !errors.As(err, &sde) {
-		zap.L().Error("service group error", zap.Error(err))
-		success = false
+	// Wait for the shutdown signal or if some service fails
+	if err := supervisor.Run(ctx, supervisor.Options{
+		ForceStop: config.ForceStop,
+		Tasks:     tasks,
+		Logger:    globalLogger,
+	}); err != nil {
+		zap.L().Error("failed to run tasks", zap.Error(err))
```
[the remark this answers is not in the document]
djeebus: Good point, we can remove all that now.

packages/shared/pkg/supervisor/main.go
djeebus: Nope, more AI fueled fun. I'll refactor down to a method so we can remove all the extra locks, channels, etc.

packages/shared/pkg/supervisor/main.go
djeebus: Much cleaner now, check out the latest version?

packages/shared/pkg/supervisor/main.go
djeebus: Much cleaner now, check out the latest version?

packages/shared/pkg/supervisor/main.go
djeebus: Much cleaner now, check out the latest version?

packages/orchestrator/main.go
```diff
@@ -463,89 +426,89 @@ func run(config cfg.Config) (success bool) {
...
-		clog := globalLogger.With(zap.String("service", closer.name), zap.Bool("forced", config.ForceStop))
-		clog.Info("closing")
-		if err := closer.close(closeCtx); err != nil {
-			clog.Error("error during shutdown", zap.Error(err))
-			success = false
-		}
+		tasks = append(tasks, supervisor.Task{
+			Name:    "template manager drain",
+			Cleanup: tmpl.Wait,
+		})
```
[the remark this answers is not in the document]
djeebus: This is intentional, the orchestrator and template manager are not actually related to each other.

packages/orchestrator/main.go
```diff
@@ -463,89 +426,89 @@ func run(config cfg.Config) (success bool) {
...
+		},
+	})
 
-	closeCtx, cancelCloseCtx := context.WithCancel(context.Background())
-	defer cancelCloseCtx()
-	if config.ForceStop {
-		cancelCloseCtx()
-	}
+	tasks = append(tasks, supervisor.Task{
+		Name: "orchestrator drain",
```
dobrac, line 462: this needs to happen as the first action when shutting down the orchestrator or the template manager

packages/orchestrator/main.go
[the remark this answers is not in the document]
dobrac: this needs to happen AFTER the orchestrator draining status

packages/orchestrator/main.go
```diff
@@ -463,89 +426,89 @@ func run(config cfg.Config) (success bool) {
...
-	zap.L().Info("Waiting for services to finish")
-	var sde serviceDoneError
-	if err := g.Wait(); err != nil && !errors.As(err, &sde) {
-		zap.L().Error("service group error", zap.Error(err))
-		success = false
+	// Wait for the shutdown signal or if some service fails
+	err = supervisor.Run(ctx, supervisor.Options{
+		ForceStop: config.ForceStop,
+		Tasks:     tasks,
+		Logger:    globalLogger,
```
dobrac, line 490: this is a bit strange, the logger is closed by the supervisor, but it's also used there

packages/shared/pkg/supervisor/main.go
```diff
@@ -0,0 +1,132 @@
...
+
+	baseLogger := opts.Logger.Named("supervisor")
+
+	var wg sync.WaitGroup
+	doneCh := make(chan taskResult, len(opts.Tasks))
+
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	startTask := func(ctx context.Context, task Task) {
```
dobrac: can this function be extracted outside of the body?

packages/shared/pkg/supervisor/main.go
```diff
@@ -0,0 +1,132 @@
...
+
+type Task struct {
+	Name       string
+	Background func(context.Context) error
+	Cleanup    func(context.Context) error
+}
+
+type Options struct {
+	ForceStop bool
+	Tasks     []Task
```
dobrac, line 45: Shouldn't this be a []*Task to prevent copy of the Task functions? (but maybe it's ok, not sure now)

packages/shared/pkg/supervisor/main.go
```diff
@@ -0,0 +1,132 @@
...
+	if reason != nil {
+		errs = append(errs, reason)
+	}
+
+	for index := len(opts.Tasks) - 1; index >= 0; index-- {
+		task := opts.Tasks[index]
+		taskLogger := baseLogger.Named(task.Name)
+		if task.Cleanup != nil {
+			taskLogger.Info("running cleanup")
+			if err := task.Cleanup(ctx); err != nil {
```
dobrac: the close context should be different from the background tasks context

dobrac (changes requested):

packages/orchestrator/main.go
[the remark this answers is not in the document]
djeebus: Template manager shouldn't have anything to do with the orchestrator, as they are launched conditionally - you can run orchestrator, template manager, or both. How does template manager rely on orchestrator?

packages/shared/pkg/supervisor/main.go
djeebus: `Task` doesn't have any mutable fields, so copying them around is okay.

PR #1365, opened 2025-10-16
Speedup docker build with cached deps, remove need for dot folders

sitole:
- Docker build context is now set to the parent directory, so we can access the whole `packages` directory.
- Moved golang package cache download before source code copy, so it will be cached always, and codebase change will not re-trigger layer cache refresh, making build much faster when you are just writing code without adding deps.

---

> [!NOTE]
> Refactors Dockerfiles and Makefiles to cache Go modules and build from parent directory, removing dot-folder copies and updating paths/binaries accordingly.
> 
> - **Dockerfiles**:
>   - **Cached deps first**: Move `go mod download` before source copy in `packages/*/Dockerfile` (`api`, `client-proxy`, `db`, `docker-reverse-proxy`, `orchestrator`) to leverage layer caching.
>   - **Parent context + paths**: Replace `.shared/.db/.clickhouse` copies with direct `./shared`, `./db`, `./clickhouse`, `./<service>` paths; adjust `WORKDIR`s and `COPY` sources accordingly.
>   - **Binaries/entrypoints**: Normalize output locations (e.g., `db` migrator to `./bin/migrator`) and ENTRYPOINT arrays; update final image COPY paths.
> - **Makefiles**:
>   - **Build context change**: Update `build-and-upload` (and `build`) targets to use `docker build ... -f ./Dockerfile ..`, removing temporary dot-folder creation/cleanup across `api`, `client-proxy`, `db`, `docker-reverse-proxy`, `orchestrator`.
>   - Minor path updates to match new Dockerfile structure.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8d3aac5042fb95eb033422c213e6aa9206f841b1. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/db/Dockerfile
+++ b/packages/db/Dockerfile
@@ -19,6 +19,7 @@ WORKDIR /build
 COPY ./shared/pkg ./shared/pkg
 
 COPY ./db/scripts/migrator.go ./db/scripts/migrator.go
+COPY ./db/migrations/ ./db/migrations
 
 WORKDIR /build/db
 
@@ -28,5 +29,6 @@ RUN chmod +x ./bin/migrator
 FROM alpine:${ALPINE_VERSION}
 
 COPY --from=builder /build/db/bin/migrator .
+COPY --from=builder  /build/db/migrations ./migrations
 
 ENTRYPOINT ["./migrator"]


packages/api/Dockerfile
```diff
@@ -5,38 +5,38 @@ FROM golang:${GOLANG_VERSION}-alpine${ALPINE_VERSION} AS builder
...
+WORKDIR /build
 
-# API
-WORKDIR /build/api
+COPY ./clickhouse/pkg ./clickhouse/pkg
+COPY ./shared/pkg ./shared/pkg
+COPY ./db/client ./db/client
+COPY ./db/dberrors ./db/dberrors
+COPY ./db/queries ./db/queries
+COPY ./db/types ./db/types
```
jakubno, line 33: can you move this to `/pkg` in a next PR? 🙏🏻

sitole: Yep

PR #1368, opened 2025-10-16
Don't assume an orchestrator port number, but pull it from nomad

djeebus:
This keeps support for assuming an orchestrator port number, but only as a fallback. This allows us to place orchestrators where ever we want, with whatever port number we want, and the API will find them.

---

> [!NOTE]
> API now discovers orchestrator IP/port from Nomad allocation resources (label `grpc`), with a configurable default fallback; Nomad job defines labeled ports and env wired from Nomad.
> 
> - **Orchestrator discovery (API)**:
>   - Switch from listing Nomad nodes to listing allocations with filter `ClientStatus == "running" and JobID contains "orchestrator-"`.
>   - Add health gating via `isHealthy` and resolve address by scanning `AllocatedResources` for port label `o.portLabel` (fallback to `o.defaultPort`).
>   - Build `NomadServiceDiscovery` from allocation IP/port; logs when missing/found.
>   - Add tests for `isHealthy` and port resolution across reserved/dynamic/shared networks in `client_test.go`.
> - **Configuration**:
>   - Introduce `Config.DefaultOrchestratorPort` (deprecated fallback) and `Config.OrchestratorPortLabel` (default `grpc`).
>   - Thread `defaultPort` and `portLabel` into `Orchestrator`.
> - **Nomad job (`iac/provider-gcp/nomad/jobs/orchestrator.hcl`)**:
>   - Define `network { mode = "host" }` with labeled ports `grpc` and `proxy`; services reference labels.
>   - Set `GRPC_PORT`/`PROXY_PORT` from `$${NOMAD_PORT_grpc}`/`$${NOMAD_PORT_proxy}`.
> - **Cleanup**:
>   - Remove `OrchestratorPort` from `packages/shared/pkg/consts/sandboxes.go`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit bd759470d1de05cc2fc8dfc9098be2994450c53f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api/internal/orchestrator/client.go
+++ b/packages/api/internal/orchestrator/client.go
@@ -97,7 +97,7 @@ func (o *Orchestrator) listNomadNodes(ctx context.Context) ([]nodemanager.NomadS
 			continue
 		}
 
-		ip, port, ok := o.findPortInAllocation(alloc, "grpc")
+		ip, port, ok := o.findPortInAllocation(alloc)
 		if !ok {
 			zap.L().Warn("Cannot find port in allocation",
 				zap.String("allocation_id", alloc.ID), zap.String("port_name", "grpc"))
@@ -144,10 +144,18 @@ func isHealthy(alloc *nomadapi.AllocationListStub) bool {
 	return *alloc.DeploymentStatus.Healthy
 }
 
-func (o *Orchestrator) findPortInAllocation(allocation *nomadapi.AllocationListStub, portLabel string) (string, int, bool) {
+func (o *Orchestrator) findPortInAllocation(allocation *nomadapi.AllocationListStub) (string, int, bool) {
+	if allocation == nil {
+		return "", 0, false
+	}
+
+	if allocation.AllocatedResources == nil {
+		return "", 0, false
+	}
+
 	for _, task := range allocation.AllocatedResources.Tasks {
 		for _, network := range task.Networks {
-			host, port, ok := o.findPortInNetwork(network, portLabel)
+			host, port, ok := o.findPortInNetwork(network)
 			if ok {
 				return host, port, true
 			}
@@ -155,7 +163,7 @@ func (o *Orchestrator) findPortInAllocation(allocation *nomadapi.AllocationListS
 	}
 
 	for _, net := range allocation.AllocatedResources.Shared.Networks {
-		host, port, ok := o.findPortInNetwork(net, portLabel)
+		host, port, ok := o.findPortInNetwork(net)
 		if ok {
 			return host, port, true
 		}
@@ -164,9 +172,9 @@ func (o *Orchestrator) findPortInAllocation(allocation *nomadapi.AllocationListS
 	return "", 0, false
 }
 
-func (o *Orchestrator) findPortInNetwork(net *nomadapi.NetworkResource, portLabel string) (string, int, bool) {
+func (o *Orchestrator) findPortInNetwork(net *nomadapi.NetworkResource) (string, int, bool) {
 	for _, port := range net.ReservedPorts {
-		if port.Label == portLabel {
+		if port.Label == o.portLabel {
 			return net.IP, port.Value, true
 		}
 
@@ -176,7 +184,7 @@ func (o *Orchestrator) findPortInNetwork(net *nomadapi.NetworkResource, portLabe
 	}
 
 	for _, port := range net.DynamicPorts {
-		if port.Label == portLabel {
+		if port.Label == o.portLabel {
 			return net.IP, port.Value, true
 		}
 
--- a/packages/api/internal/orchestrator/orchestrator.go
+++ b/packages/api/internal/orchestrator/orchestrator.go
@@ -38,6 +38,7 @@ type Orchestrator struct {
 	// Deprecated: Nomad knows which port the orchestrator is listening on. Keep this
 	// around temporarily until all nomad jobs have a port labeled "grpc", then this can be removed.
 	defaultPort int
+	portLabel   string
 
 	httpClient              *http.Client
 	nomadClient             *nomadapi.Client
@@ -128,6 +129,7 @@ func New(
 		sqlcDB:             sqlcDB,
 		tel:                tel,
 		clusters:           clusters,
+		portLabel:          config.OrchestratorPortLabel,
 		defaultPort:        config.DefaultOrchestratorPort,
 
 		sandboxCounter: sandboxCounter,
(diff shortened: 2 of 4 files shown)


packages/api/internal/orchestrator/client.go
```diff
@@ -78,26 +78,84 @@ func (o *Orchestrator) listNomadNodes(ctx context.Context) ([]nodemanager.NomadS
...
+			return net.IP, port.Value, true
+		}
+
+		if port.Value == o.defaultPort {
+			return net.IP, o.defaultPort, true
+		}
+	}
+
+	return "", 0, false
+}
```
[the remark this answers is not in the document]
djeebus: This is intentional; if we don't find a port that matches the default one, we assume that we've found something that isn't an orchestrator job and move on.

.editorconfig
```diff
@@ -0,0 +1,2 @@
+[*.hcl]
+indent_size = 2
```
sitole, line 2: Shouldn’t this be enforced by terraform fmt?

packages/api/internal/orchestrator/orchestrator.go
```diff
@@ -35,6 +35,10 @@ const statusLogInterval = time.Second * 20
 var ErrNodeNotFound = errors.New("node not found")
 
 type Orchestrator struct {
+	// Deprecated: Nomad knows which port the orchestrator is listening on. Keep this
```
sitole, line 38: We also need to track proxy port. It will ve needed for sandbox traffic routing done by client proxy.

Data are send via Redis sandbox catalog done in API. We are now sending just IP to orchestrator job

.editorconfig
djeebus: This file is to inform IDEs and text editors about our preferred format. Additionally, `terraform fmt` only formats `.tf` files, not `.hcl`, and this file is effectively just a template file as far as terraform is concerned.

packages/api/internal/orchestrator/orchestrator.go
djeebus: Definitely! I was hoping to keep this PR smaller by omitting those changes, but I can extend this to include those changes as well.

.editorconfig
sitole: True! Did not realize `.hcl` is for Nomad job specs.

packages/api/internal/orchestrator/client.go
```diff
@@ -78,26 +78,84 @@ func (o *Orchestrator) listNomadNodes(ctx context.Context) ([]nodemanager.NomadS
 	defer listSpan.End()
 
 	options := &nomadapi.QueryOptions{
-		// TODO: Use variable for node pool name ("default")
-		Filter: "Status == \"ready\" and NodePool == \"default\"",
+		Filter: `ClientStatus == "running" and JobID contains "orchestrator-"`,
+		Params: map[string]string{"resources": "true"},
 	}
-	nomadNodes, _, err := o.nomadClient.Nodes().List(options.WithContext(ctx))
+	nomadAllocations, _, err := o.nomadClient.Allocations().List(options.WithContext(ctx))
```
ValentaTomas, line 87: We need to add only healthly orchestrators.

packages/api/internal/orchestrator/client.go
```diff
@@ -88,10 +91,17 @@ func (o *Orchestrator) listNomadNodes(ctx context.Context) ([]nodemanager.NomadS
 
 	result := make([]nodemanager.NomadServiceDiscovery, 0, len(nomadAllocations))
 	for _, alloc := range nomadAllocations {
+		if !isHealthy(alloc) {
+			zap.L().Info("Skipping unhealthy allocation", zap.String("allocation_id", alloc.ID))
```
sitole: This can spam a lot in prod, right? Can we make it debug at least?

sitole: Same for `isHealthy` function

packages/api/internal/orchestrator/orchestrator.go
sitole: Feel free to skip it for now. Let's add a dependent task so we will not forget about it.
It will also need some testing in edge service discovery (that does orchestrator discovery in BYOC) to add it there also.

packages/api/internal/orchestrator/client.go
djeebus: This info might, I'll reduce it to debug. 

The ones inside `isHealthy` shouldn't ever be triggered:

- `alloc` should never be nil
- all orchestrator allocations should have a deployment status
- all orchestrator deployment statuses should have a health check

If any of the above assumptions are incorrect, I really want to know about it. so I think it's worth keeping. If it turns out to be valueless spam, we can definitely swap it to debug or remove it completely.

packages/api/internal/orchestrator/client.go
```diff
@@ -81,26 +81,113 @@ func (o *Orchestrator) listNomadNodes(ctx context.Context) ([]nodemanager.NomadS
...
 	options := &nomadapi.QueryOptions{
-		// TODO: Use variable for node pool name ("default")
-		Filter: "Status == \"ready\" and NodePool == \"default\"",
+		Filter: `ClientStatus == "running" and JobID contains "orchestrator-"`,
+		Params: map[string]string{"resources": "true"},
 	}
-	nomadNodes, _, err := o.nomadClient.Nodes().List(options.WithContext(ctx))
+	nomadAllocations, _, err := o.nomadClient.Allocations().List(options.WithContext(ctx))
 	if err != nil {
 		return nil, err
```
ValentaTomas, line 89: The list of allocations might be paginated, so we should check if we are actually listing everything.

ValentaTomas: By default it seems not paginated though.

ValentaTomas (approved):

packages/api/internal/orchestrator/client.go
```diff
@@ -81,26 +81,91 @@ func (o *Orchestrator) listNomadNodes(ctx context.Context) ([]nodemanager.NomadS
...
 
-	result := make([]nodemanager.NomadServiceDiscovery, 0, len(nomadNodes))
-	for _, n := range nomadNodes {
+	result := make([]nodemanager.NomadServiceDiscovery, 0, len(nomadAllocations))
+	for _, alloc := range nomadAllocations {
+		if alloc.DeploymentStatus.Healthy != nil && !*alloc.DeploymentStatus.Healthy {
+			zap.L().Info("Skipping unhealthy allocation", zap.String("allocation_id", alloc.ID))
+
+			continue
+		}
```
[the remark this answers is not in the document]
djeebus: might have been true at some point, but not anymore

PR #1369, opened 2025-10-17
Refactor sandbox events structure

sitole:
- Added version number and unified event type string to each event, so handling is easier
- Back compatibility migrator so events API can start serving new event formats, even data is still send in the old format
- Unified events naming in events pkg, removed the duplicated mapping from the clickhouse pkg
- Clickhouse migrations that add event type and version
- When the events api will adopts a new events format handling, we can stop sending event label, category, and add back compatibility in the events api on the fly, and clickhouse columns can be dropped.

---

> [!NOTE]
> Adds versioned, dot-namespaced sandbox event types with migration helpers, updates ClickHouse schema and batchers, and refactors orchestrator to publish/use the new event model.
> 
> - **Events (shared)**:
>   - Add `events.SandboxEvent` with `version` and `type` fields and constants for lifecycle events; deprecate `event_category`/`event_label` (kept for back-compat).
>   - Add `LegacySandboxEventMigrationMapping` and event pair helpers to translate between legacy and new formats.
>   - Remove old `event.SandboxEvent` type and lifecycle/webhooks enums/metadata; keep `webhooks.DeriveKey` only.
> - **ClickHouse**:
>   - Migration adds `type` and `version` columns to `sandbox_events`/`sandbox_events_local`.
>   - Update insert batcher and `SandboxEvent` model to handle `type`/`version`.
> - **Orchestrator**:
>   - Switch to `events` package, update pubsub generics to `struct{}` metadata, and require `version`/`type` in validation.
>   - Publish lifecycle events using new `events.*EventPair` (populate both new `type/version` and legacy fields); adjust logging.
> - **API/Config**:
>   - Remove `DNS_PORT` from API config.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 008c6a17bf9e04e52ed5df91636c628317b51a11. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/shared/pkg/events/migration_test.go
+++ b/packages/shared/pkg/events/migration_test.go
@@ -13,6 +13,7 @@ func TestLegacySandboxEventMigrationMapping(t *testing.T) {
 			in: SandboxEvent{
 				EventCategory: "lifecycle",
 				EventLabel:    "create",
+				Version:       "v1",
 			},
 			want: SandboxEvent{
 				EventCategory: "lifecycle",
@@ -25,6 +26,7 @@ func TestLegacySandboxEventMigrationMapping(t *testing.T) {
 			in: SandboxEvent{
 				EventCategory: "lifecycle",
 				EventLabel:    "pause",
+				Version:       "v1",
 			},
 			want: SandboxEvent{
 				EventCategory: "lifecycle",
@@ -37,6 +39,7 @@ func TestLegacySandboxEventMigrationMapping(t *testing.T) {
 			in: SandboxEvent{
 				EventCategory: "lifecycle",
 				EventLabel:    "kill",
+				Version:       "v1",
 			},
 			want: SandboxEvent{
 				EventCategory: "lifecycle",
@@ -49,6 +52,7 @@ func TestLegacySandboxEventMigrationMapping(t *testing.T) {
 			in: SandboxEvent{
 				EventCategory: "lifecycle",
 				EventLabel:    "resume",
+				Version:       "v1",
 			},
 			want: SandboxEvent{
 				EventCategory: "lifecycle",
@@ -61,6 +65,7 @@ func TestLegacySandboxEventMigrationMapping(t *testing.T) {
 			in: SandboxEvent{
 				EventCategory: "lifecycle",
 				EventLabel:    "update",
+				Version:       "v1",
 			},
 			want: SandboxEvent{
 				EventCategory: "lifecycle",


packages/clickhouse/pkg/sandbox_events.go
```diff
@@ -9,37 +9,20 @@ import (
...
-	SandboxID          string         `ch:"sandbox_id"`
-	SandboxExecutionID string         `ch:"sandbox_execution_id"`
-	SandboxTemplateID  string         `ch:"sandbox_template_id"`
-	SandboxBuildID     string         `ch:"sandbox_build_id"`
-	SandboxTeamID      uuid.UUID      `ch:"sandbox_team_id"`
-	EventCategory      string         `ch:"event_category"`
-	EventLabel         string         `ch:"event_label"`
-	EventData          sql.NullString `ch:"event_data"`
+	Version string `ch:"version"`
+	Type    string `ch:"type"`
```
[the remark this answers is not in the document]
sitole: True, i will change it to `type` in clickhouse

packages/orchestrator/internal/events/sandbox_events_service.go
```diff
@@ -30,14 +30,14 @@ func (e *MissingEventFieldError) Error() string {
 // as well as persisting using a ClickHouse Batcher
 type SandboxEventsService struct {
 	featureFlags *featureflags.Client
-	pubsub       pubsub.PubSub[event.SandboxEvent, webhooks.SandboxWebhooksMetaData]
```
sitole, line 33: We still need same pub sub impl for internal use during migration period, then we can remove all webhooks metadata related code from there.

packages/api/internal/cfg/model.go
```diff
@@ -10,8 +10,6 @@ type Config struct {
 
 	ClickhouseConnectionString string `env:"CLICKHOUSE_CONNECTION_STRING"`
 
-	DNSPort uint16 `env:"DNS_PORT" envDefault:"5353"`
-
```
dobrac, line 14: not related to this PR, but I assume it's otherwise unused

packages/clickhouse/migrations/20251017213616_sandbox_events_sync_global.sql
```diff
@@ -0,0 +1,13 @@
+-- +goose Up
+-- +goose StatementBegin
+ALTER TABLE sandbox_events
```
dobrac: can these two migrations be just one?

packages/clickhouse/migrations/20251017213615_sandbox_events_version_and_type_fields.sql
```diff
@@ -0,0 +1,13 @@
+-- +goose Up
+-- +goose StatementBegin
+ALTER TABLE sandbox_events_local
+    ADD COLUMN type LowCardinality(String) CODEC (ZSTD(1)) AFTER event_label,
+    ADD COLUMN version String CODEC (ZSTD(1)) AFTER type;
```
dobrac: wdyt about specifying the version Default value?

packages/shared/pkg/events/sandbox.go
```diff
@@ -0,0 +1,44 @@
...
+	SandboxUpdatedEvent,
+}
+
+type SandboxEvent struct {
+	Version string `json:"version"`
+	Type    string `json:"type"`
+
+	// Deprecated: for new events use event field with dot syntax
+	EventCategory string `json:"event_category"`
+	// Deprecated: for new events use event field with dot syntax
```
dobrac, line 36:
```suggestion
	// Deprecated: for new events use type field with dot syntax
	EventCategory string `json:"event_category"`
	// Deprecated: for new events use type field with dot syntax
```

packages/shared/pkg/events/migration.go
```diff
@@ -0,0 +1,62 @@
...
+	Type:           SandboxKilledEvent,
+	LegacyCategory: "lifecycle",
+	LegacyLabel:    "kill",
+}
+
+// LegacySandboxEventMigrationMapping works for senders back compatibility and converting old event types to new ones
+// We will receive old event just with event category and label, so we need to map them to new event types that
+// are using new dot namespaced syntax for event names
+func LegacySandboxEventMigrationMapping(e SandboxEvent) SandboxEvent {
+	e.Version = StructureVersionV1
```
dobrac: isn't this technically version 2 already?

packages/shared/pkg/events/sandbox.go
```diff
@@ -0,0 +1,44 @@
+package events
+
+import (
+	"time"
+
+	"github.com/google/uuid"
+)
+
+const StructureVersionV1 = "v1"
```
dobrac: just double checking, is this fully compatible with the semver package versioning handling?

packages/clickhouse/migrations/20251017213616_sandbox_events_sync_global.sql
sitole: They cannot. Clickhouse will allow only one alert at once. I tried to do that

packages/api/internal/cfg/model.go
sitole: yep, i just discovered so removed it. It was forgotten removal from pr where we removed dns server from api

packages/shared/pkg/events/sandbox.go
sitole: it should be correct from semver perspective, will check if we will handle that correctly.

packages/shared/pkg/events/migration.go
sitole: Previously there was no version at all, but yes, makes sense.. will update

packages/shared/pkg/events/sandbox.go
sitole: Checked, and we already have tests for major versions. Works with this.

sitole:
> > maybe I've missed it, but is there any migration of the existing data in ClickHouse?
> 
> Can do it, good point

Since rollout will take a longer time and orchestrators will send old date, I would do it after all orchestrators are rolled out and our events api will be in production. Then we can stop sending.

Related issues that will follow:
- https://linear.app/e2b/issue/ENG-3203/stop-sending-event-label-and-event-category-to-clickhouse
- https://linear.app/e2b/issue/ENG-3204/events-database-migration-path

packages/clickhouse/migrations/20251017213616_sandbox_events_sync_global.sql
sitole: Found out they can be one file, but every statement must be separated with a comment line.

packages/shared/pkg/events/migration.go
```diff
@@ -41,20 +41,46 @@ var SandboxKilledEventPair = SandboxEventType{
...
-		case "resume":
-			e.Type = SandboxResumedEventPair.Type
-		case "update":
-			e.Type = SandboxUpdatedEventPair.Type
-		case "kill":
-			e.Type = SandboxKilledEventPair.Type
+	// Older events structures did not have version set, so we need to set it to v1 first
+	if e.Version == "" {
+		e.Version = StructureVersionV1
+	}
```
dobrac: isn't this case solved by using the "default" in the migration?

sitole: true, removed

packages/shared/pkg/events/migration.go
```diff
@@ -0,0 +1,83 @@
...
+			e.EventCategory = SandboxUpdatedEventPair.LegacyCategory
+			e.EventLabel = SandboxUpdatedEventPair.LegacyLabel
+		case SandboxKilledEvent:
+			e.EventCategory = SandboxKilledEventPair.LegacyCategory
+			e.EventLabel = SandboxKilledEventPair.LegacyLabel
+		}
+	}
+
+	return e
+}
```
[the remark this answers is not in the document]
sitole: Migration function will be used only on environments with ClickHouse database migrations, adding a default version field with back compatibility.

dobrac (approved):

PR #1377, opened 2025-10-18
fix: orchestrator/template manager shutdown exit code

dobrac:
Fix orchestrator/template manager shutdown exit code by properly handling exit errors and improving logger messages observability.

### DIFF
--- a/packages/orchestrator/main.go
+++ b/packages/orchestrator/main.go
@@ -144,7 +144,7 @@ func run(config cfg.Config) (success bool) {
 	}))
 	defer func(l *zap.Logger) {
 		err := l.Sync()
-		if err != nil && !errors.Is(err, syscall.EINVAL) {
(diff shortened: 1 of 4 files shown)


packages/orchestrator/main.go
```diff
@@ -194,15 +180,30 @@ func run(config cfg.Config) (success bool) {
...
+	// defer waiting on the group so that this runs even when
+	// there's a panic.
+	defer func(g *errgroup.Group) {
+		err := g.Wait()
+		var sde serviceDoneError
+		if err != nil && !errors.As(err, &sde) {
+			globalLogger.Error("error while waiting for services to finish", zap.Error(err))
+			success = false
+		}
+	}(&g)
```
dobrac, line 205: just moved down so it can use the global logger

PR #1381, opened 2025-10-19
fix: template build COPY user handling

dobrac:
This fixes the following usage:

```
 .copy("folder", "/project/workspace/linux/", {
    user: "root"
  })
```

---

> [!NOTE]
> Runs COPY extraction/move as root and applies ownership/permissions (including safe symlink handling) via a new arg parser and script, with comprehensive tests added.
> 
> - **Build Orchestrator – `commands/copy.go`**:
>   - Run file transfer and extraction as `root` using `Context.WithUser("root")` and root-owned tmp copy.
>   - Add `parseCopyArgs` to normalize `source`, `target`, `owner` (defaulting to `user:user`), and optional `permissions`; strip globs via `doublestar`.
>   - Enhance copy script to `chown`/`chmod` before move:
>     - Symlinks: `chown -h` and move without `chmod`.
>     - Files: `chown` and optional `chmod`.
>     - Directories: recursive `chown`/`chmod`, then move contents (including hidden files).
>   - Remove post-move `chown`/`chmod`; handled inside script. Improve error message for unknown entry types.
> - **Tests – `commands/copy_test.go`**:
>   - Add extensive tests for arg parsing and script behavior across files, dirs, symlinks, hidden files, nested paths, and permissions.
> - **Metadata – `template_metadata.go`**:
>   - Add `Context.WithUser(user string)` helper.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a2929c4f801d2a3a499c4a8b4bace06892854182. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/orchestrator/internal/template/build/commands/copy_test.go
+++ b/packages/orchestrator/internal/template/build/commands/copy_test.go
@@ -163,6 +163,224 @@ type testCase struct {
 	expectedPaths map[string]string
 }
 
+func TestParseCopyArgs(t *testing.T) {
+	tests := []struct {
+		name          string
+		args          []string
+		defaultUser   string
+		expected      *copyArgs
+		expectedError string
+	}{
+		{
+			name:        "minimum_valid_arguments",
+			args:        []string{"/local/path", "/container/path"},
+			defaultUser: "user",
+			expected: &copyArgs{
+				SourcePath:  "/local/path",
+				TargetPath:  "/container/path",
+				Owner:       "user:user",
+				Permissions: "",
+			},
+		},
+		{
+			name:        "with_owner_specified",
+			args:        []string{"/local/path", "/container/path", "root"},
+			defaultUser: "user",
+			expected: &copyArgs{
+				SourcePath:  "/local/path",
+				TargetPath:  "/container/path",
+				Owner:       "root:root",
+				Permissions: "",
+			},
+		},
+		{
+			name:        "owner_with_group",
+			args:        []string{"/local/path", "/container/path", "www-data:www-data"},
+			defaultUser: "user",
+			expected: &copyArgs{
+				SourcePath:  "/local/path",
+				TargetPath:  "/container/path",
+				Owner:       "www-data:www-data",
+				Permissions: "",
+			},
+		},
+		{
+			name:        "owner_with_different_group",
+			args:        []string{"/local/path", "/container/path", "user:staff"},
+			defaultUser: "defaultuser",
+			expected: &copyArgs{
+				SourcePath:  "/local/path",
+				TargetPath:  "/container/path",
+				Owner:       "user:staff",
+				Permissions: "",
+			},
+		},
+		{
+			name:        "empty_owner_uses_default",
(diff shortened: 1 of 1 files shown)


packages/orchestrator/internal/template/build/commands/copy.go
```diff
@@ -164,15 +184,36 @@ func (c *Copy) Execute(
...
+
+	// Determine the target owner
+	owner := fmt.Sprintf("%s:%s", cmdMetadata.User, cmdMetadata.User)
+	if len(args) >= 3 && args[2] != "" {
+		// Assumes the format of chown: user:group or just user
+		owner = args[2]
+		if !strings.Contains(owner, ":") {
+			owner = fmt.Sprintf("%s:%s", owner, owner)
+		}
+	}
```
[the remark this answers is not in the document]
dobrac: the user is never empty, and also there was a bug before

ValentaTomas:
Is the recursive chmod/chown ok speed-wise? In the previous build provision.sh, wasn't a similar permission change relatively slow?

dobrac:
> Is the recursive chmod/chown ok speed-wise? In the previous build provision.sh, wasn't a similar permission change relatively slow?

Totally fine. The issue previously was that it was run in Docker and Docker is slow for this because it needs to copy all files to a new layer, which we don't need to do. (e.g. for the Code Interpreter it went from 120 seconds to basically 0 seconds)

packages/orchestrator/internal/template/build/commands/copy.go
```diff
@@ -164,15 +184,36 @@ func (c *Copy) Execute(
...
+
+	// Get optional permissions
+	permissions := ""
+	if len(args) >= 4 {
+		permissions = args[3]
+	}
+
```
sitole: Don't we want to extract this logic and unit test it? Can get really messy and confusing with more arguments coming.

packages/orchestrator/internal/template/build/commands/copy.go
```diff
@@ -57,6 +59,24 @@ if [ -z "$entry" ]; then
  exit 1
 fi
 
+{{- if .Owner }}
```
sitole: Isn't owner known always?

packages/orchestrator/internal/template/build/commands/copy.go
dobrac: probably yes

packages/orchestrator/internal/template/build/commands/copy.go
dobrac: yes, it is, will update

packages/orchestrator/internal/template/build/commands/copy.go
dobrac: added tests for the script + moved the argument parsing to a separate function

sitole (commented):
Lgtm!

Image

sitole (approved):

PR #1406, opened 2025-10-27
feat: return templates that are building for the first time

dobrac:
> [!NOTE]
> Adds buildStatus to Template API and updates listing to include templates with the latest build (including in-progress).
> 
> - **API/schema**:
>   - Add `buildStatus` to `Template` and define reusable `TemplateBuildStatus` enum; reference it in `TemplateBuild.status`.
>   - Regenerate OpenAPI spec and client/server models (`spec.gen.go`, `types.gen.go`, test models).
> - **DB**:
>   - Update `get_team_templates` to return `eb.status AS build_status` and remove `status = 'uploaded'` filter to select the latest build.
> - **Handlers**:
>   - Populate `Template.buildStatus` in `GetTemplates` from query result.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f359e5326fc6940ae2fde0a9be49ce342abd8d1c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/api/internal/handlers/templates_list.go
```diff
@@ -85,6 +85,7 @@ func (a *APIStore) GetTemplates(c *gin.Context, params api.GetTemplatesParams) {
 			LastSpawnedAt: item.Env.LastSpawnedAt,
 			SpawnCount:    item.Env.SpawnCount,
 			BuildCount:    item.Env.BuildCount,
+			BuildStatus:   api.TemplateBuildStatus(item.BuildStatus),
```
jakubno, line 88: use `getCorrespondingTemplateBuildStatus`

dobrac: the getCorrespondingTemplateBuildStatus is for entgo, also, as the item.BuildStatus is string, it wouldn't be type safe anyway, so I think we should keep it just like this for now unless we introduce enum field in the DB

PR #1408, opened 2025-10-27
feat: add list of builds per template

dobrac:
## Summary

Adds the ability to list all builds for a template via the `GET /templates/{templateID}` endpoint.

## Changes

- **Response schema**:
  - Template fields (templateID, public, aliases, timestamps, etc.)
  - `builds` array containing complete build history
- **Each build** contains build metadata:
  - `buildID` - unique identifier for the build
  - `status` - current build status
  - `createdAt`, `updatedAt`, `finishedAt` - timestamps
  - `cpuCount`, `memoryMB`, `diskSizeMB` - resource configuration
  - `envdVersion` - version of the envd used

---

> [!NOTE]
> Adds GET `/templates/{templateID}` to return a template with its paginated build history and introduces shared cursor-based pagination, updating schemas, handlers, DB queries, and clients.
> 
> - **API/Spec**:
>   - New endpoint `GET /templates/{templateID}` with `nextToken`/`limit` to return `TemplateWithBuilds` (template + paginated `builds`).
>   - Introduce reusable pagination params (`paginationNextToken`, `paginationLimit`) and apply to `v2/sandboxes`.
>   - Split build schemas: `TemplateBuild` (metadata) vs `TemplateBuildInfo` (status/logs); update build-status endpoint to return `TemplateBuildInfo`.
> - **Handlers**:
>   - Implement `GetTemplatesTemplateID` to authorize, fetch template + builds via cursor, and set `X-Next-Token`.
>   - Refactor `GetV2Sandboxes` to use new pagination utility.
> - **DB**:
>   - Add `GetTemplateByIDWithAliases` and `GetTemplateBuilds` (cursor on `(created_at, id)`), with generated sqlc code.
> - **Utils**:
>   - New generic pagination helper (`NewPagination`, `ProcessResultsWithHeader`) and export `MaxSandboxID`; add `CastPtr`.
> - **Types/Clients**:
>   - `TemplateBuild` now includes `buildID (uuid)`, timestamps, resources; logging moved to `TemplateBuildInfo`.
>   - Regenerate API types/clients/tests to reflect new endpoint, params, and schemas.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6dbcfb8ce096fa095682e1abc9e7bc9424cf179b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/db/queries/get_template_with_builds.sql
```diff
@@ -0,0 +1,5 @@
+-- name: GetTemplateBuilds :many
+SELECT eb.*
+FROM public.env_builds eb
+WHERE eb.env_id = sqlc.arg(template_id)
+ORDER BY eb.created_at DESC;
```
sitole: Can we add some limit here?

packages/api/internal/handlers/template_get.go
```diff
@@ -0,0 +1,83 @@
...
+		telemetry.ReportCriticalError(ctx, "error when getting team and tier", apiErr.Err)
+
+		return
+	}
+
+	telemetry.SetAttributes(ctx,
+		telemetry.WithTeamID(team.ID.String()),
+	)
+
+	template, err := a.sqlcDB.GetTemplateByIDWithAliases(ctx, templateID)
```
sitole, line 41: It's expected that this will allow you to return snapshots?

packages/db/queries/get_template_with_builds.sql
dobrac: thats a great point, although we don't have currently a standardized way to do pagination. This will most probably require larger refactor, so I'm changing this PR to draft a will create a new one that will refactor existing pagination places

packages/api/internal/handlers/template_get.go
dobrac: no, thx!

packages/db/queries/get_template_with_builds.sql
dobrac: refactored

packages/db/queries/get_template_with_builds.sql
```diff
@@ -0,0 +1,7 @@
+-- name: GetTemplateBuilds :many
+SELECT eb.*
+FROM public.env_builds eb
+WHERE eb.env_id = sqlc.arg(template_id)
+  AND (eb.created_at, eb.id::text) < (@cursor_time, @cursor_id::text)
```
sitole, line 5: I think use of uuidv7 can can help here a lot

sitole (approved):

packages/db/queries/get_template_with_builds.sql
dobrac: maybe, but only if we're generating them all from the database. Otherwise it's not guaranteed that the uuidv7 generated in one place is always preceeding uuid generated in another afaik

sitole: Yep, but as it's part of B-tree, I think it will still benefit from only local arrangements (because you insert newer UUID and then older) than reworking the whole tree because of UUIDv4 use.

PR #1409, opened 2025-10-27
Refactor UFFD

ValentaTomas:
This should be only a refactor now—I tried to minimize all functional changes.

---

> [!NOTE]
> Refactors UFFD into a dedicated userfaultfd type with fd helpers, replaces old mapping with a tested memory mapping, and updates integration/tests accordingly.
> 
> - **Userfaultfd refactor**:
>   - Introduce `userfaultfd.Userfaultfd` with `Serve`/`Close`, encapsulating polling, pagefault handling, and missing-page copy logic.
>   - Add `userfaultfd/fd.go` wrapping UFFD syscalls/constants; remove old `constants.go` and `syscalls.go`.
>   - Move eagain logging to `userfaultfd` and streamline read/poll loop.
> - **Memory mapping module**:
>   - Replace `mapping` package with `memory` module: `Region`, `Mapping`, `GetOffset`, and `AddressNotFoundError`.
>   - Add comprehensive tests in `memory/mapping_test.go`.
> - **Orchestrator integration**:
>   - Update `uffd.Uffd` to parse region mappings, construct `memory.Mapping`, create `userfaultfd.Userfaultfd`, and call `Serve`.
>   - Remove legacy `Serve` implementation and mapping types; rename buffers (`regionMappings*`).
> - **Tests**:
>   - Migrate tests to `userfaultfd` package; update cross-process helpers to new fd API and mapping; adjust helpers and missing/missing_write tests.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ac9a4b19b228be7cc4b75e28c7e3b755a72051a9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/orchestrator/internal/sandbox/uffd/userfaultfd/missing_test.go
```diff
@@ -0,0 +1,225 @@
...
+			}
+
+			expectedAccessedOffsets := getOperationsOffsets(tt.operations, operationModeRead|operationModeWrite)
+			assert.Equal(t, expectedAccessedOffsets, h.getAccessedOffsets(), "checking which pages were faulted")
+		})
+	}
+}
+
+func TestParallelMissing(t *testing.T) {
+	// TODO: At around 10k+ parallel operations the test often freezes.
```
ValentaTomas: The serving loop should be essentially identical to the loop we now have in production, so I think this problem might be also in production—from checking this I don't think this is in our code, it seems just like the uffd/pipe stops serving things when we do a lot of parallel requests from the other side.

packages/orchestrator/internal/sandbox/uffd/userfaultfd/missing_write_test.go
```diff
@@ -0,0 +1,230 @@
...
+			}
+
+			expectedAccessedOffsets := getOperationsOffsets(tt.operations, operationModeRead|operationModeWrite)
+			assert.Equal(t, expectedAccessedOffsets, h.getAccessedOffsets(), "checking which pages were faulted")
+		})
+	}
+}
+
+func TestParallelMissingWrite(t *testing.T) {
+	// TODO: At around 10k+ parallel operations the test often freezes.
```
ValentaTomas: Same as for the parallel missing.

packages/orchestrator/internal/sandbox/uffd/userfaultfd/serve.go
```diff
@@ -0,0 +1,207 @@
...
+	onFailure func() error,
+	addr uintptr,
+	offset int64,
+	pagesize uint64,
+) error {
+	if _, ok := u.missingRequests.Load(offset); ok {
+		return nil
+	}
+
+	u.missingRequests.Store(offset, struct{}{})
```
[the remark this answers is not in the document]
ValentaTomas: We are keeping the current functionality, but this might be a concern as outlined in comments.

packages/orchestrator/internal/sandbox/uffd/userfaultfd/serve.go
```diff
@@ -0,0 +1,207 @@
...
+		if flags == 0 {
+			err := u.handleMissing(ctx, fdExit.SignalExit, addr, offset, pagesize)
+			if err != nil {
+				return fmt.Errorf("failed to handle missing: %w", err)
+			}
+
+			continue
+		}
+
+		return fmt.Errorf("unexpected event type: %d, closing uffd", flags)
```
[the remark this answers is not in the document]
ValentaTomas: There should be no other types of faults at this point and if they are, handling them via copy is not correct.

packages/orchestrator/internal/sandbox/uffd/memory/mapping_test.go
```diff
@@ -0,0 +1,168 @@
...
+		hostVirtAddr   uintptr
+		expectedOffset int64
+		expectedSize   uint64
+		expectError    bool
+	}{
+		{
+			name:           "valid address in first region",
+			hostVirtAddr:   0x1500,
+			expectedOffset: 0x5500, // 0x5000 + (0x1500 - 0x1000)
+			expectedSize:   4096,
```
djeebus: if we're going to use hex, can we be consistent?


```suggestion
			hostVirtAddr:   0x1500,
			expectedOffset: 0x5500, // 0x5000 + (0x1500 - 0x1000)
			expectedSize:   0x1000,
```

djeebus: Same for below

packages/orchestrator/internal/sandbox/uffd/memory/mapping_test.go
```diff
@@ -0,0 +1,168 @@
...
+		},
+		{
+			name:         "address after last region",
+			hostVirtAddr: 0x7000,
+			expectError:  true,
+		},
+		{
+			name:         "address in gap between regions",
+			hostVirtAddr: 0x4000,
+			expectError:  true,
```
djeebus: Can we verify what the error was, and not just that there's an error and it's not nil?

packages/orchestrator/internal/sandbox/uffd/uffd.go
```diff
@@ -27,40 +27,39 @@ var tracer = otel.Tracer("github.com/e2b-dev/infra/packages/orchestrator/interna
...
+	fdExit     *fdexit.FdExit
+	lis        *net.UnixListener
 	socketPath string
+	memfile    *block.TrackedSliceDevice
+	handler    utils.SetOnce[*userfaultfd.Userfaultfd]
 }
 
 var _ MemoryBackend = (*Uffd)(nil)
 
 func New(memfile block.ReadonlyDevice, socketPath string, blockSize int64) (*Uffd, error) {
```
djeebus, line 45: can we get some tests covering this struct?

ValentaTomas: Theoretically, yes, but this whole interaction is FC-specific (essentially passing fd + their struct), so not sure if that is worth it.

packages/orchestrator/internal/sandbox/uffd/memory/mapping_test.go
ValentaTomas: Currently, we don't return a specific error if the address cannot be resolved; the whole FC is essentially unrecoverable anyway.

packages/orchestrator/internal/sandbox/uffd/userfaultfd/serve.go
```diff
@@ -0,0 +1,207 @@
...
+			}
+
+			u.logger.Error("uffd: read error", zap.Error(err))
+
+			return fmt.Errorf("failed to read: %w", err)
+		}
+
+		msg := *(*UffdMsg)(unsafe.Pointer(&buf[0]))
+		if GetMsgEvent(&msg) != UFFD_EVENT_PAGEFAULT {
+			u.logger.Error("UFFD serve unexpected event type", zap.Any("event_type", GetMsgEvent(&msg)))
```
djeebus:
```suggestion
		if msgEvent := GetMsgEvent(&msg); msgEvent != UFFD_EVENT_PAGEFAULT {
			u.logger.Error("UFFD serve unexpected event type", zap.Any("event_type", msgEvent))
```

packages/orchestrator/internal/sandbox/uffd/userfaultfd/constants.go
```diff
@@ -104,14 +97,6 @@ func GetMsgArg(msg *UffdMsg) [24]byte {
...
-
-func IsWritePageFault(pagefault *UffdPagefault) bool {
-	return pagefault.flags&UFFD_PAGEFAULT_FLAG_WRITE != 0
-}
-
-func IsWriteProtectPageFault(pagefault *UffdPagefault) bool {
-	return pagefault.flags&UFFD_PAGEFAULT_FLAG_WP != 0
+func GetPagefaultAddress(pagefault *UffdPagefault) uintptr {
+	return uintptr(pagefault.address)
 }
```
djeebus: These are only used internally, unexport them:


```suggestion
func getPagefaultAddress(pagefault *UffdPagefault) uintptr {
	return uintptr(pagefault.address)
}
```

packages/orchestrator/internal/sandbox/uffd/userfaultfd/syscalls.go
```diff
@@ -40,10 +44,10 @@ func ConfigureApi(fd uintptr, pagesize uint64) error {
 // mode: UFFDIO_REGISTER_MODE_WP|UFFDIO_REGISTER_MODE_MISSING
 // This is already called by the FC, but only with the UFFDIO_REGISTER_MODE_MISSING
 // We need to call it with UFFDIO_REGISTER_MODE_WP when we use both missing and wp
-func Register(fd uintptr, addr uintptr, size uint64, mode CULong) error {
+func (u *Userfaultfd) Register(addr uintptr, size uint64, mode CULong) error {
```
djeebus: Looks like this function is only called in the test. Can we move it and the `UffdioRegister` struct into a _test.go file?

packages/orchestrator/internal/sandbox/uffd/userfaultfd/serve.go
```diff
@@ -0,0 +1,207 @@
...
+
+			return ErrUnexpectedEventType
+		}
+
+		arg := GetMsgArg(&msg)
```
djeebus: Let's unexport this, since it's only used in the package:


```suggestion
		arg := getMsgArg(&msg)
```

packages/orchestrator/internal/sandbox/uffd/userfaultfd/serve.go
```diff
@@ -0,0 +1,207 @@
...
+		pagefault := (*(*UffdPagefault)(unsafe.Pointer(&arg[0])))
+		flags := pagefault.flags
+
+		addr := GetPagefaultAddress(&pagefault)
```
djeebus: Let's unexport this, since it's only used in the package:


```suggestion
		addr := getPagefaultAddress(&pagefault)
```

packages/orchestrator/internal/sandbox/uffd/uffd.go
ValentaTomas: Not adding this for now.

packages/orchestrator/internal/sandbox/uffd/userfaultfd/syscalls.go
ValentaTomas: Not sure if splitting the methods that all operate on the same type between normal and test files won't be confusing?

packages/orchestrator/internal/sandbox/uffd/memory/mapping.go
```diff
@@ -0,0 +1,32 @@
...
+type AddressNotFoundError struct {
+	hostVirtAddr uintptr
+}
+
+func (e AddressNotFoundError) Error() string {
+	return fmt.Sprintf("address %d not found in any mapping", e.hostVirtAddr)
+}
+
+type Mapping struct {
+	Regions []Region
```
djeebus, line 16: Unexport this, as it's only used internally:


```suggestion
	regions []Region
```

packages/orchestrator/internal/sandbox/uffd/memory/region.go
```diff
@@ -0,0 +1,23 @@
+package memory
+
+// Region is a mapping of a region of memory of the guest to a region of memory on the host.
+// The serialization is based on the Firecracker UFFD protocol communication.
```
djeebus, line 4: Is there a link to the documentation that we can add here?

packages/orchestrator/internal/sandbox/uffd/memory/region.go
```diff
@@ -0,0 +1,23 @@
...
+type Region struct {
+	BaseHostVirtAddr uintptr `json:"base_host_virt_addr"`
+	Size             uintptr `json:"size"`
+	Offset           uintptr `json:"offset"`
+	// This is actually in bytes.
```
djeebus: 😱

packages/orchestrator/internal/sandbox/uffd/memory/region.go
```diff
@@ -0,0 +1,23 @@
...
+	// This field is deprecated in the newer version of the Firecracer with a new field `page_size`.
```
djeebus:
```suggestion
	// This field is deprecated in the newer version of the Firecracker with a new field `page_size`.
```

packages/orchestrator/internal/sandbox/uffd/userfaultfd/cross_process_helpers_test.go
```diff
@@ -208,16 +207,45 @@ func crossProcessServe() error {
...
+	uffdFd := uffdFile.Fd()
 
 	contentFile := os.NewFile(uintptr(4), "content")
 	defer contentFile.Close()
 
-	offsetsFile := os.NewFile(uintptr(5), "offsets")
+	content, err := io.ReadAll(contentFile)
+	if err != nil {
+		return fmt.Errorf("exit reading content: %w", err)
+	}
```
djeebus, line 218: This is executed via `GO_TEST_HELPER_PROCESS`, but it still has access to `*testing.T`. Can we use it to `require.NoError(t, err)`? What's that look like if it fails?

packages/orchestrator/internal/sandbox/uffd/memory/mapping.go
ValentaTomas: Will use in followup PR.

packages/orchestrator/internal/sandbox/uffd/userfaultfd/cross_process_helpers_test.go
```diff
@@ -208,16 +207,45 @@ func crossProcessServe() error {
...
 
-	readyFile := os.NewFile(uintptr(6), "ready")
+	pageSize, err := strconv.Atoi(os.Getenv("GO_MMAP_PAGE_SIZE"))
```
djeebus: `strconv.ParseInt(os.Getenv("GO_MMAP_PAGE_SIZE"), 10, 64)` lets you avoid a conversion, but is slightly more verbose. Up to you.

packages/orchestrator/internal/sandbox/uffd/userfaultfd/cross_process_helpers_test.go
```diff
@@ -208,16 +207,45 @@ func crossProcessServe() error {
...
+	m := memory.NewMapping([]memory.Region{
+		{
+			BaseHostVirtAddr: memoryStart,
+			Size:             uintptr(len(content)),
+			Offset:           0,
+			PageSize:         uintptr(pageSize),
+		},
+	})
+
+	exitUffd := make(chan struct{}, 1)
```
djeebus, line 236: probably need a `defer close(exitUffd)`

packages/orchestrator/internal/sandbox/uffd/userfaultfd/cross_process_helpers_test.go
```diff
@@ -332,6 +332,8 @@ func crossProcessServe() error {
 	signal.Notify(exitSignal, syscall.SIGUSR1)
 	defer signal.Stop(exitSignal)
 
+	readyFile := os.NewFile(uintptr(6), "ready")
```
djeebus, line 336: The FD offsets are ugly, but I'm not sure there's a simple way to document "this is the input index + 3" =/ if you can think of one, it'd be really good to have here!

packages/orchestrator/internal/sandbox/uffd/uffd.go
```diff
@@ -27,40 +27,39 @@ var tracer = otel.Tracer("github.com/e2b-dev/infra/packages/orchestrator/interna
...
-	trackedMemfile, err := block.NewTrackedSliceDevice(blockSize, memfile)
+	fdExit, err := fdexit.New()
 	if err != nil {
-		return nil, fmt.Errorf("failed to create tracked slice device: %w", err)
+		return nil, fmt.Errorf("failed to create fd exit: %w", err)
 	}
 
-	fdExit, err := fdexit.New()
+	trackedMemfile, err := block.NewTrackedSliceDevice(blockSize, memfile)
```
jakubno: looks like it can't return an error

ValentaTomas: Won't be touching this one yet—removing this altogether in the next PR.

packages/orchestrator/internal/sandbox/uffd/userfaultfd/cross_process_helpers_test.go
ValentaTomas: I opted out of that—it does not start as regular tests and the *testing.T was strange.

PR #1411, opened 2025-10-27
Support local ephemeral nvme disks

djeebus:
Local benchmarks indicate that this drop base sandbox latency ~20-30ms

Downsides:
- Live migrations will probably get slower for longer.

---

> [!NOTE]
> Replaces PD cache disks with configurable local NVMe SSDs for client/build nodes, provisioning them via startup script and wiring new disk-count variables through Terraform.
> 
> - **Compute templates (client/build)**:
>   - Replace PD cache disks with dynamic `local-ssd` NVMe scratch disks (`375GB`) using `dynamic "disk"` and `range(var.*_cluster_cache_disk_count)`.
>   - Set boot disk `auto_delete` on client; keep root disk sizing via `build_cluster_root_disk_size_gb`.
> - **Startup script (`scripts/start-client.sh`)**:
>   - Partition local NVMe disks and assemble RAID0 (`mdadm`) when multiple; persist config.
>   - Format XFS, add to `/etc/fstab`, mount at `/orchestrator`; create `sandbox`, `template`, `build` dirs.
>   - Add 100G swap, persist swap and sysctl tuning; remove sudo usage; minor robustness (retry loops).
>   - Template var `LOCAL_CACHE_DISK_COUNT` passed from Terraform.
> - **Terraform variables/wiring**:
>   - Add `build_cluster_cache_disk_count` and `client_cluster_cache_disk_count` (with validation) in root and module; pass through in `main.tf`.
>   - Remove cache disk size/type vars from `nomad-cluster` module (PD-based settings).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4e008cd711f877203c5acd2326220495b0dbd6d9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/iac/provider-gcp/nomad-cluster/nodepool-client.tf
+++ b/iac/provider-gcp/nomad-cluster/nodepool-client.tf
@@ -1,7 +1,4 @@
 locals {
-  cache_disk_count = 3   # todo: make configurable
-  cache_disk_size  = 375 # todo: make configurable
-
   client_pool_name = "${var.prefix}${var.client_cluster_name}"
   client_startup_script = templatefile("${path.module}/scripts/start-client.sh", {
     CLUSTER_TAG_NAME             = var.cluster_tag_name
@@ -25,7 +22,7 @@ locals {
     USE_FILESTORE_CACHE          = var.filestore_cache_enabled
     NODE_POOL                    = var.orchestrator_node_pool
     BASE_HUGEPAGES_PERCENTAGE    = var.orchestrator_base_hugepages_percentage
-    LOCAL_CACHE_DISK_COUNT       = local.cache_disk_count
+    LOCAL_CACHE_DISK_COUNT       = var.client_cluster_cache_disk_count
   })
 }
 
@@ -143,12 +140,12 @@ resource "google_compute_instance_template" "client" {
   }
 
   dynamic "disk" {
-    for_each = [for n in range(local.cache_disk_count) : {}]
+    for_each = [for n in range(var.client_cluster_cache_disk_count) : {}]
 
     content {
       auto_delete  = true
       boot         = false
-      disk_size_gb = local.cache_disk_size
+      disk_size_gb = 375
       interface    = "NVME"
       disk_type    = "local-ssd"
       type         = "SCRATCH"
--- a/iac/provider-gcp/nomad-cluster/scripts/start-client.sh
+++ b/iac/provider-gcp/nomad-cluster/scripts/start-client.sh
@@ -16,93 +16,85 @@ set -x
 exec > >(tee /var/log/user-data.log | logger -t user-data -s 2>/dev/console) 2>&1
 
 # Add cache disk for orchestrator and swapfile
-%{ if LOCAL_CACHE_DISK_COUNT > 0 }
 for i in {0..${ LOCAL_CACHE_DISK_COUNT - 1 }}; do
   dev_path="/dev/disk/by-id/google-local-nvme-ssd-$i"
-  part_path="$dev_path-part1"
   echo "partitioning drive #$i"
-  sudo parted --script $dev_path \
+  parted --script $dev_path \
       mklabel gpt \
       mkpart primary 0% 100% \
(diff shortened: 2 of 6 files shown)


iac/provider-gcp/nomad-cluster/nodepool-client.tf
```diff
@@ -1,4 +1,7 @@
 locals {
+  cache_disk_count = 3   # todo: make configurable
+  cache_disk_size  = 375 # todo: make configurable
```
ValentaTomas: Local SSD size is not effectively configurable, right?

iac/provider-gcp/nomad-cluster/nodepool-client.tf
```diff
@@ -131,18 +135,24 @@ resource "google_compute_instance_template" "client" {
   }
 
   disk {
+    auto_delete  = true
```
ValentaTomas, line 135: Do we want this?

iac/provider-gcp/nomad-cluster/scripts/start-client.sh
```diff
@@ -16,8 +16,45 @@ set -x
...
+
+if [ ! -b "$DISK" ]; then
+  echo "failed to create raid array"
+  exit 99
+fi
+
+echo "persisting array configuration"
+sudo mdadm --detail --scan --verbose | sudo tee -a /etc/mdadm/mdadm.conf
+%{ else }
 DISK="/dev/disk/by-id/google-persistent-disk-1"
```
ValentaTomas: Is this relevant anymore?

iac/provider-gcp/nomad-cluster/nodepool-client.tf
djeebus: I believe that's currently true, but it probably won't be true forever. I'll definitely set it as a default though.

iac/provider-gcp/nomad-cluster/nodepool-client.tf
djeebus: No reason to keep a boot or cache disk around after we delete the VM instance, right?

iac/provider-gcp/nomad-cluster/scripts/start-client.sh
djeebus: I wasn't sure if we want to commit to this, or a/b test this in production. I can assume we'll commit and remove the old stuff.

iac/provider-gcp/nomad-cluster/nodepool-client.tf
ValentaTomas: I think only Z3 machines have this configurable, for everything else it is always 375 per disk.

iac/provider-gcp/nomad-cluster/nodepool-client.tf
ValentaTomas: Maybe in some strange debugging circumstances, but agree here.
Does this mean we are accumulating the disk somewhere right now?

djeebus: No, turns out `auto_delete` is set to `true` by default. I can remove this if we don't want it be clear.

iac/provider-gcp/nomad-cluster/scripts/start-client.sh
```diff
@@ -16,8 +16,45 @@ set -x
 exec > >(tee /var/log/user-data.log | logger -t user-data -s 2>/dev/console) 2>&1
 
 # Add cache disk for orchestrator and swapfile
-# TODO: Parametrize this
+%{ if LOCAL_CACHE_DISK_COUNT > 0 }
+for i in {0..${ LOCAL_CACHE_DISK_COUNT - 1 }}; do
```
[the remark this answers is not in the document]
ValentaTomas: This is actually a hcl template, so it should work.

ValentaTomas (approved):

PR #1414, opened 2025-10-28
`RST` sandbox network connections on sandbox close

ValentaTomas:
TODO:
- [x] Don't return errors that are expected when closing connections
- [x] Check if RST correctly propagates to client/error in client


packages/orchestrator/internal/template/build/layer/layer_executor.go
```diff
@@ -82,7 +82,12 @@ func (lb *LayerExecutor) BuildLayer(
 	lb.sandboxes.Insert(sbx)
 	defer func() {
 		lb.sandboxes.Remove(sbx.Runtime.SandboxID)
-		lb.proxy.RemoveFromPool(sbx.Runtime.ExecutionID)
+
+		closeErr := lb.proxy.RemoveFromPool(sbx.Runtime.ExecutionID)
+		if closeErr != nil {
+			// Errors here will be from forcefully closing the connections, so we can ignore them—they will at worst timeout on their own.
+			lb.logger.Warn("errors when manually closing connections to sandbox", zap.Error(closeErr))
```
ValentaTomas, line 89: @dobrac I didn't want to completely discard the closing errors—is the `lb.logger` the correct logger to use if I don't want to expose these logs to users (during build)?

packages/shared/pkg/proxy/tracking/connection.go
```diff
@@ -1,23 +1,57 @@
...
+}
+
+func (c *Connection) Reset() error {
+	var errs []error
+
+	// This forces the connection to close with RST.
+	err := c.Conn.(*net.TCPConn).SetLinger(0)
+	if err != nil {
+		errs = append(errs, err)
 	}
```
djeebus, line 47: any idea how we can test that this is doing what we want it to do?

ValentaTomas: It might always be partially flaky if the connection is closed from somewhere else, but will look into it.

packages/orchestrator/internal/template/build/layer/layer_executor.go
dobrac: yes, that should be correct

packages/shared/pkg/proxy/tracking/connection.go
ValentaTomas: I had to instrument it a bit more, but achieved this.

ValentaTomas:
> Will be checking the failed integration tests.

On the rerun they went through, will still check (https://github.com/e2b-dev/infra/actions/runs/18963434009/job/54155528873?pr=1414).

packages/shared/pkg/proxy/tracking/connection.go
```diff
@@ -1,23 +1,57 @@
...
 	counter.Add(1)
 
-	return &Connection{
+	c := &Connection{
 		Conn:    conn,
 		counter: counter,
+		m:       m,
+	}
+
+	if m != nil {
```
djeebus, line 31: do we want `m` to ever be nil? seems like we could simplify this a little by always tracking connections

djeebus (approved):

packages/shared/pkg/proxy/tracking/connection.go
ValentaTomas: Yes, we don't track this for the incoming conn at all as we never have a use case with cancelling these.

PR #1425, opened 2025-10-30
Remove double logging of requests in client proxy and cleanup logging fields in proxy logs

ValentaTomas:


packages/client-proxy/internal/proxy/proxy.go
```diff
@@ -77,17 +81,22 @@ func NewClientProxy(meterProvider metric.MeterProvider, serviceName string, port
 				return nil, reverseproxy.NewErrSandboxNotFound(sandboxId)
 			}
 
-			logger.Debug("Proxying request", zap.String("node_ip", nodeIP))
```
ValentaTomas, line 80: We stopped logging here, because both proxies log by default via https://github.com/e2b-dev/infra/blob/eb971d7a2959cfe5a61766ef3c31ad7880beeabb/packages/shared/pkg/proxy/handler.go#L86

packages/client-proxy/internal/proxy/proxy.go
```diff
@@ -77,17 +81,22 @@ func NewClientProxy(meterProvider metric.MeterProvider, serviceName string, port
...
+			url := &url.URL{
+				Scheme: "http",
+				Host:   fmt.Sprintf("%s:%d", nodeIP, orchestratorProxyPort),
+			}
+
+			logger = logger.With(
+				zap.String("target_hostname", url.Hostname()),
```
ValentaTomas, line 90: The node ip is included as this now.

packages/client-proxy/internal/proxy/proxy.go
```diff
@@ -62,10 +62,14 @@ func NewClientProxy(meterProvider metric.MeterProvider, serviceName string, port
 			}
 
 			logger := zap.L().With(
-				zap.String("host", r.Host),
+				zap.String("origin_host", r.Host),
```
ValentaTomas, line 65: This was confusing as `host_name` and `host_ip` are automatically included fields for the host the proxy is running on.

packages/orchestrator/internal/proxy/proxy.go
```diff
@@ -50,6 +50,19 @@ func NewSandboxProxy(meterProvider metric.MeterProvider, port uint16, sandboxes
 				Host:   fmt.Sprintf("%s:%d", sbx.Slot.HostIPString(), port),
 			}
 
+			logger := zap.L().With(
```
ValentaTomas, line 53: We added mostly the same fields we add in the client proxy 👌

djeebus (approved):

PR #1430, opened 2025-11-02
Fix GCP Consul DNS resolution with two-layer protection

tomassrnka:
- Add dpkg-divert to permanently block gce-resolved.conf in Packer image
  - Configure Consul as recursive DNS with GCE forwarder in startup script
  - Remove routing domains approach (Domains=~consul) which doesn't work reliably
  - Restart systemd-resolved after Consul starts to prevent marking DNS as unreachable

_Note: This works only for GCE deployment and does not take into account AWS / BYOC. The IP address is the same in AWS, so actually it should work, but it's just not generic enough at this moment_

---

> [!NOTE]
> On GCE, block gce-resolved.conf at image build and start Consul with a dynamically fetched GCE DNS recursor, reconfiguring systemd-resolved after Consul is up so Consul handles all DNS.
> 
> - **GCP Image (Packer)**:
>   - Add `dpkg-divert` to block `gce-resolved.conf` (`/etc/systemd/resolved.conf.d/gce-resolved.conf`) to avoid DNS conflicts with Consul.
> - **Startup Script (`start-client.sh`)**:
>   - Configure `systemd-resolved` to use only `127.0.0.1:8600`; remove routing domains and disable GCE's `gce-resolved.conf`.
>   - Dynamically fetch GCE DNS from metadata and pass it to Consul via `--recursor`.
>   - Start Consul first, wait for DNS port readiness, then restart `systemd-resolved`; add DNS readiness checks and cache flush.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9fbd28e24c3d53b18e54d934bdd9a8a6e49e29a5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/iac/provider-gcp/nomad-cluster/scripts/start-client.sh
+++ b/iac/provider-gcp/nomad-cluster/scripts/start-client.sh
@@ -304,10 +304,22 @@ echo "[Configuring systemd-resolved for Consul DNS]"
 echo "- Restarting systemd-resolved to apply Consul DNS config"
 systemctl restart systemd-resolved
 echo "- Waiting for systemd-resolved to settle"
-sleep 5
+
+# Give Consul a moment to start its DNS server on port 8600
+echo "- Waiting for Systemd-resolved to start..."
+for i in {1..10}; do
+  if host google.com 2>/dev/null; then
+    echo "- DNS resolving is ready (attempt $i/10)"
+    break
+  fi
+  if [ $i -eq 10 ]; then
+    echo "- ERROR: Systemd-resolved not responding after 10 seconds, exiting..."
+    exit 1
+  fi
+  sleep 1
+done
 echo "- Flushing DNS caches"
 resolvectl flush-caches
-echo "- systemd-resolved configured to use Consul DNS for all queries"
 
 /opt/nomad/bin/run-nomad.sh --client --consul-token "${CONSUL_TOKEN}" --node-pool "${NODE_POOL}" &
 


sitole:
> Note: This works only for GCE deployment and does not take into account AWS / BYOC. The IP address is the same in AWS, so actually it should work, but it's just not generic enough at this moment

Provision script here is GPC specific already so we can merge it without IFs for different clouds. Later when merging configuration related on BYOC and multi-cloud support we can resolve it.

iac/provider-gcp/nomad-cluster/scripts/start-client.sh
```diff
@@ -260,13 +267,36 @@ overcommitment_hugepages=$(remove_decimal $overcommitment_hugepages)
...
     --consul-token "${CONSUL_TOKEN}" \
     --cluster-tag-name "${CLUSTER_TAG_NAME}" \
     --enable-gossip-encryption \
     --gossip-encryption-key "${CONSUL_GOSSIP_ENCRYPTION_KEY}" \
-    --dns-request-token "${CONSUL_DNS_REQUEST_TOKEN}" &
+    --dns-request-token "${CONSUL_DNS_REQUEST_TOKEN}" \
+    --recursor "$${GCE_DNS}" &
+
+# Give Consul a moment to start its DNS server on port 8600
+sleep 3
```
jakubno: Please put some loop which will check if the server is reachable. Otherwise it's race-y and this will cause problems as we've seen before. The script should fail if the port 8600 isn't accessible in some timeout, that will prevent nomad to start and it will mark VM unhealthy after a while

jakubno (changes requested):

iac/provider-gcp/nomad-cluster/scripts/start-client.sh
```diff
@@ -260,13 +267,47 @@ overcommitment_hugepages=$(remove_decimal $overcommitment_hugepages)
...
     --enable-gossip-encryption \
     --gossip-encryption-key "${CONSUL_GOSSIP_ENCRYPTION_KEY}" \
-    --dns-request-token "${CONSUL_DNS_REQUEST_TOKEN}" &
+    --dns-request-token "${CONSUL_DNS_REQUEST_TOKEN}" \
+    --recursor "$${GCE_DNS}" &
+
+# Give Consul a moment to start its DNS server on port 8600
+echo "- Waiting for Consul DNS to start on port 8600..."
+for i in {1..10}; do
+  if nc -z 127.0.0.1 8600 2>/dev/null; then
```
djeebus, line 288: `nslookup` returns non-zero if the lookup fails, could use that for a more meaningful check

```suggestion
  if ! nslookup google.com; then
```

iac/provider-gcp/nomad-cluster/scripts/start-client.sh
```diff
@@ -260,13 +267,47 @@ overcommitment_hugepages=$(remove_decimal $overcommitment_hugepages)
...
+
+# Now restart systemd-resolved to apply Consul DNS configuration
+# This must happen AFTER Consul starts, otherwise systemd-resolved marks 127.0.0.1:8600 as unreachable
+# Consul DNS (127.0.0.1:8600) is the ONLY DNS server configured in systemd-resolved
+# Consul handles ALL queries: .consul directly, everything else via recursor to GCE DNS
+echo "[Configuring systemd-resolved for Consul DNS]"
+echo "- Restarting systemd-resolved to apply Consul DNS config"
+systemctl restart systemd-resolved
+echo "- Waiting for systemd-resolved to settle"
+sleep 5
```
djeebus: Could use `nslookup` here as well to wait for it to successfully resolve names

jakubno (approved):

PR #1432, opened 2025-11-03
Refactor dirty page tracking

ValentaTomas:
> [!NOTE]
> Refactors dirty-page tracking to a new block Tracker, updates UFFD to unregister and return tracked pages via Disable(ctx), and wires changes through sandbox, build diffs, and tests.
> 
> - **Memory/UFFD**:
>   - Replace internal missing/dirty tracking with `block.Tracker`; add `Disable(ctx) (*block.Tracker, error)` and remove direct `Dirty()` from `MemoryBackend`.
>   - `uffd.New` drops block-size param; `Uffd.Disable` now unregisters mappings (adds `UFFDIO_UNREGISTER`) and returns tracker; internal `dirty(ctx)` helper added.
>   - Enforce page-size consistency in `userfaultfd.NewUserfaultfdFromFd`; limit in-flight requests to `4096`.
> - **Block**:
>   - Add `Tracker` with `Add/Has/Reset/Clone/Offsets/BitSet/BlockSize`; `Slicer` now requires `BlockSize()`.
>   - Remove old tracked slice device; new `bitsetOffsets` and iter-based offsets.
> - **Sandbox Integration**:
>   - Use `memory.Disable(ctx)` to get dirty pages; pass `dirty.BitSet()` to memory diff; update `serveMemory` to new `uffd.New` signature.
> - **Build/Diff**:
>   - Add `BlockSize()` to `Diff` implementations (`NoDiff`, `localDiff`, `StorageDiff`).
> - **Tests/Utils**:
>   - Add `tracker_test.go`; revise cross-process test to stream `Tracker.Offsets()`; add `utils.TransformTo` iterator helper; test `MemorySlicer` now implements `BlockSize()`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8dc51d0425c77fc172fc6fb8c6901aeeb690fbb3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/orchestrator/internal/sandbox/uffd/noop.go
```diff
@@ -23,23 +24,23 @@ var _ MemoryBackend = (*NoopMemory)(nil)
...
 
-func (m *NoopMemory) Disable() error {
+func (m *NoopMemory) Disable(context.Context) error {
 	return nil
 }
 
-func (m *NoopMemory) Dirty() *bitset.BitSet {
-	return m.dirty
+func (m *NoopMemory) Dirty(context.Context) (*block.Tracker, error) {
+	return m.dirty.Clone(), nil
```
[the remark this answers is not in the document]
ValentaTomas: The code lines are not correct, but the reported error is.

packages/orchestrator/internal/sandbox/uffd/userfaultfd/userfaultfd.go
```diff
@@ -192,12 +191,11 @@ func (u *Userfaultfd) handleMissing(
 	offset int64,
 	pagesize uint64,
 ) error {
-	if _, ok := u.missingRequests.Load(offset); ok {
+	if !u.missingRequests.Add(offset) {
+		// Page is already mapped
```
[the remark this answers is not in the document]
ValentaTomas: This is the same problem we already have with the way we tracked the dirty pages. If the copy fails then it would be the same problem anyway.

The planned solution here is to pause the sandbox, wait for all requests in flight to actually copy, then notify the orchestrator that it can proceed.

packages/shared/pkg/utils/settle_counter.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+// Wait waits for the counter to be the settle value.
+func (w *SettleCounter) Wait(ctx context.Context) error {
+	// Ensure we can break out of this Wait when the context is done.
+	go func() {
+		<-ctx.Done()
+
+		w.cond.Broadcast()
+	}()
```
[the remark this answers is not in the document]
ValentaTomas: Didn't know `context.AfterFunc` was a thing.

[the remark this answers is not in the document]
djeebus: Why not just use a `sync.Waitgroup`?

packages/shared/pkg/utils/settle_counter.go
```diff
@@ -0,0 +1,67 @@
...
+		cond:        *sync.NewCond(&sync.Mutex{}),
+		settleValue: 0,
+	}
+}
+
+func (w *SettleCounter) add(delta int64) {
+	if w.counter.Add(delta) == w.settleValue {
+		w.cond.Broadcast()
+	}
+}
```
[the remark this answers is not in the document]
ValentaTomas: Got rid of the settle counter, as we might not need to wait this way.

packages/orchestrator/internal/sandbox/uffd/uffd.go
```diff
@@ -187,10 +184,20 @@ func (u *Uffd) Exit() *utils.ErrorOnce {
...
 
-func (u *Uffd) Disable() error {
-	return u.memfile.Disable()
+func (u *Uffd) Disable(ctx context.Context) error {
+	uffd, err := u.handler.WaitWithContext(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get uffd: %w", err)
+	}
+
+	return uffd.Unregister()
```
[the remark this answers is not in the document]
ValentaTomas: This should not happen as we are calling the unregister only after the pause.

packages/orchestrator/internal/sandbox/uffd/userfaultfd/userfaultfd.go
```diff
@@ -192,13 +200,16 @@
...
-	u.missingRequests.Store(offset, struct{}{})
+	// We don't skip the already mapped pages, because if the memory is swappable the page *might* under some conditions be mapped out.
 
 	u.wg.Go(func() error {
+		// Add() must be called inside the goroutine to ensure Remove() runs via defer
+		// even if the errgroup is cancelled or the goroutine returns early.
+		// This check protects us agains race condition between marking the request as missing and accessing the missingRequests tracker.
+		// The Firecracker pause should return only after the requested memory is faulted in, so we don't need to guard the pagefault from the moment it is created.
+		u.settleRequests.Add()
+		defer u.settleRequests.Remove()
```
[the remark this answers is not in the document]
ValentaTomas: Dirty must be called only after calling pause, otherwise we are not able to properly synchronize.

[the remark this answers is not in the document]
ValentaTomas: I'll add clarifying comment.

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -706,6 +707,10 @@ func (s *Sandbox) Pause(
 	// Close the file even if an error occurs
 	defer memfile.Close()
 
+	if err := s.memory.Disable(ctx); err != nil {
+		return nil, fmt.Errorf("failed to disable uffd: %w", err)
+	}
```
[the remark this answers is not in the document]
ValentaTomas: At the time of calling of dirty, all page fault requests should be processed (FC pause returned)—we are sychronizing only because we want to missingRequest to be marked after the page copy finished.

packages/orchestrator/internal/sandbox/uffd/userfaultfd/userfaultfd.go
```diff
@@ -25,22 +27,29 @@ type Userfaultfd struct {
...
 		src:             src,
-		missingRequests: sync.Map{},
+		missingRequests: block.NewTracker(pagesize),
 		ma:              m,
 		logger:          logger,
-	}, nil
+	}
+
+	// By default this was unlimited.
+	u.wg.SetLimit(maxRequestsInProgress)
```
ValentaTomas, line 63: Now that we don't skip previously faulted pages we add at least some boundaries to the concurrency.

ValentaTomas: In some brief tests, adding a limit actually improved the handling at high concurrency.

packages/orchestrator/internal/sandbox/uffd/userfaultfd/userfaultfd.go
```diff
@@ -192,13 +201,16 @@ func (u *Userfaultfd) handleMissing(
 	offset int64,
 	pagesize uint64,
 ) error {
-	if _, ok := u.missingRequests.Load(offset); ok {
-		return nil
-	}
-
-	u.missingRequests.Store(offset, struct{}{})
+	// We don't skip the already mapped pages, because if the memory is swappable the page *might* under some conditions be mapped out.
```
ValentaTomas: For hugepages this should not be a problem, but might theoretically happen to normal pages with swap—I'm switching this back, because I'm not sure and cannot find 100% evidence on the contrary.

packages/orchestrator/internal/sandbox/uffd/userfaultfd/userfaultfd.go
```diff
@@ -25,22 +27,29 @@ type Userfaultfd struct {
...
 		fd:              uffdFd(fd),
 		src:             src,
-		missingRequests: sync.Map{},
+		missingRequests: block.NewTracker(pagesize),
 		ma:              m,
 		logger:          logger,
-	}, nil
+	}
+
+	// By default this was unlimited.
```
dobrac, line 60: nit: I would switch the comment in this PR page and in the code :D

packages/orchestrator/internal/sandbox/uffd/userfaultfd/userfaultfd.go
dobrac: can you add this to the code comment please?

packages/orchestrator/internal/sandbox/uffd/userfaultfd/userfaultfd.go
```diff
@@ -192,13 +201,16 @@ func (u *Userfaultfd) handleMissing(
...
 ) error {
-	if _, ok := u.missingRequests.Load(offset); ok {
-		return nil
-	}
-
-	u.missingRequests.Store(offset, struct{}{})
+	// We don't skip the already mapped pages, because if the memory is swappable the page *might* under some conditions be mapped out.
 
 	u.wg.Go(func() error {
+		// Add() must be called inside the goroutine to ensure Remove() runs via defer
```
dobrac: which remove? Can you please link it? 🙏

packages/orchestrator/internal/sandbox/uffd/userfaultfd/userfaultfd.go
```diff
@@ -25,22 +27,29 @@ type Userfaultfd struct {
...
 
 	wg errgroup.Group
 
 	logger *zap.Logger
 }
 
 // NewUserfaultfdFromFd creates a new userfaultfd instance with optional configuration.
-func NewUserfaultfdFromFd(fd uintptr, src block.Slicer, m *memory.Mapping, logger *zap.Logger) (*Userfaultfd, error) {
-	return &Userfaultfd{
+func NewUserfaultfdFromFd(fd uintptr, src block.Slicer, m *memory.Mapping, pagesize int64, logger *zap.Logger) (*Userfaultfd, error) {
```
dobrac: you should change the slicer interface rather than adding the pagesize, as the pagesize is directly related to the slicer implementation

packages/orchestrator/internal/sandbox/uffd/userfaultfd/userfaultfd.go
```diff
@@ -235,8 +247,30 @@ func (u *Userfaultfd) handleMissing(
...
+			return fmt.Errorf("failed to unregister: %w", err)
+		}
+	}
+
+	return nil
+}
+
+func (u *Userfaultfd) Dirty() *block.Tracker {
+	// This will be at worst cancelled when the uffd is closed.
+	u.settleRequests.Lock()
```
dobrac, line 286: can we be sure that at this point all uffd goroutines with the read lock already started and any new one is not started afterwards?

can we maybe use better locking mechanism here?

packages/orchestrator/internal/sandbox/uffd/userfaultfd/fd.go
```diff
@@ -66,6 +67,13 @@ func newUffdioAPI(api, features CULong) UffdioAPI {
...
+func newUffdioRange(start, length CULong) UffdioRange {
+	return UffdioRange{
+		start: start,
+		len:   length,
+	}
+}
+
 func newUffdioRegister(start, length, mode CULong) UffdioRegister {
 	return UffdioRegister{
 		_range: UffdioRange{
```
dobrac: should the UffdioRange be used also here (probably also other places)

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -706,6 +707,10 @@ func (s *Sandbox) Pause(
 	// Close the file even if an error occurs
 	defer memfile.Close()
 
+	if err := s.memory.Disable(ctx); err != nil {
```
dobrac: is there a reason why the Disable is called after the AcquireTmpMemfile lock? 

if not, can we maybe merge the Dirty and Disable calls to just one (exposed one, internally, there can be two)?

packages/orchestrator/internal/sandbox/block/tracker_test.go
```diff
@@ -0,0 +1,109 @@
...
+func TestTracker_AddAndHas(t *testing.T) {
+	const pageSize = 4096
+	tr := NewTracker(pageSize)
+
+	offset := int64(pageSize * 4)
+
+	// Initially should not be marked
+	if tr.Has(offset) {
+		t.Errorf("Expected offset %d not to be marked initially", offset)
+	}
```
djeebus: maybe a little easier to understand?

```suggestion
	hasOffset := tr.Has(offset)
	assert.False(t, hasOffset)
```

djeebus: Similar comment to the next block.

djeebus: And the rest of this file.

packages/orchestrator/internal/sandbox/block/tracker_test.go
```diff
@@ -0,0 +1,109 @@
...
+		tr.Add(o)
+	}
+
+	// Verify all offsets are marked
+	for _, o := range offsets {
+		if !tr.Has(o) {
+			t.Errorf("Expected offset %d to be marked", o)
+		}
+	}
+}
```
djeebus, line 89: Should probably test that other offsets are _not_ marked.

packages/orchestrator/internal/sandbox/block/tracker.go
```diff
@@ -1,66 +1,83 @@
...
+func (t *Tracker) Offsets() iter.Seq[int64] {
+	return bitsetOffsets(t.BitSet(), t.BlockSize())
+}
 
-	return t.dirty.Clone()
+func bitsetOffsets(b *bitset.BitSet, blockSize int64) iter.Seq[int64] {
+	return utils.TransformTo(b.EachSet(), func(idx uint) int64 {
+		return header.BlockOffset(int64(idx), blockSize)
+	})
 }
```
djeebus, line 83: Can we get a test for Offsets as well?

packages/shared/pkg/utils/iter.go
```diff
@@ -0,0 +1,13 @@
+package utils
+
+import "iter"
+
+func TransformTo[S, T any](iterator iter.Seq[S], f func(S) T) iter.Seq[T] {
```
djeebus, line 5: `Map` is probably a more expected name here

ValentaTomas: We already have `Map` here, so I opted to use transform for iterators.

packages/orchestrator/internal/sandbox/sandbox.go
ValentaTomas: I wanted to call it at a latest possible moment before it is required—right before `CreateSnapshot`.

Merge of the methods could be possible, but I wanted to separate them, because in the next step (forked FC+clone), we will be calling only `Dirty`, as the disable (unregister) is a non-reversible action.

packages/orchestrator/internal/sandbox/uffd/userfaultfd/fd.go
ValentaTomas: @dobrac You mean the constructor, right?

ValentaTomas: Switched to the constructor, not finding this anywhere else.

packages/orchestrator/internal/sandbox/uffd/userfaultfd/userfaultfd.go
ValentaTomas: The only requirement is that this is called after FC API pause (I have a comment about this on the `Uffd` method).
All the uffd requests should be handled by now (after the FC pause returned) and we are effectively only waiting for the info to be added to trackers (but still need to guard this).

I originally added a "settle counter", doing this via `sync.Cond`, but the implementation itself was way bigger and in the end it felt more error prone, as what the RWLock is guarding agains here is really the only thing we need to look for.

ValentaTomas: > any new one is not started afterwards?

Dived into FC and talked with Tomas, was also suspicious of this :D
The FC, KMV should be handling this, otherwise we won't be essentially able to handle this at all too.

packages/orchestrator/internal/sandbox/uffd/userfaultfd/userfaultfd.go
ValentaTomas: I think these are actually separate—the uffd will later also track writes, which have nothing to do with the slicer.

I'll refactor, though, as there is some ambiguity with the passed pagesize and the pagesize we get from regions.

packages/orchestrator/internal/sandbox/uffd/userfaultfd/fd.go
dobrac: sorry, yes, the constructor

packages/orchestrator/internal/sandbox/uffd/userfaultfd/userfaultfd.go
ValentaTomas: Ok, very good note, pushed.

packages/orchestrator/internal/sandbox/uffd/userfaultfd/userfaultfd.go
dobrac: Do you have any guarantee that the goroutine is actually started? In general the goroutine doesn’t have to be run yet when the function returns (although, the probability is small, so might be hard to manifest)

ValentaTomas: Yes—if the goroutine was not started, the copy was not run, so the FC API pause should not have finished.
The FC API is the synchronization primitive—after it runs the memory should be be "settled", otherwise even then normal fc memory dump snapshots would be racy.

packages/orchestrator/internal/sandbox/uffd/userfaultfd/userfaultfd.go
```diff
@@ -0,0 +1,286 @@
...
+	// Intentionally not using defer—we just need to write lock and write unlock the mutex, to be sure all the RLock calls are released.
+	u.settleRequests.Unlock() //nolint:staticcheck
+
+	return u.missingRequests.Clone()
+}
```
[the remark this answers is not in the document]
ValentaTomas: Kinda—not needed, but added for clarity.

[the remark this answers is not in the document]
ValentaTomas: This should not affect us in the pause flow, but changing it to be more explicit.

packages/orchestrator/internal/sandbox/uffd/userfaultfd/userfaultfd.go
```diff
@@ -196,13 +217,14 @@ func (u *Userfaultfd) handleMissing(
...
-
-	u.missingRequests.Store(offset, struct{}{})
-
 	u.wg.Go(func() error {
+		// The RLock must be called inside the goroutine to ensure RUnlock runs via defer,
+		// even if the errgroup is cancelled or the goroutine returns early.
+		// This check protects us against race condition between marking the request as missing and accessing the missingRequests tracker.
+		// The Firecracker pause should return only after the requested memory is faulted in, so we don't need to guard the pagefault from the moment it is created.
+		u.settleRequests.RLock()
+		defer u.settleRequests.RUnlock()
```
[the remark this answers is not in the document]
ValentaTomas: All the read goroutines must have been started before we call Dirty as we can only call dirty after we get response from pause and that should happen only after all copy syscalls in the goroutine return.

packages/orchestrator/internal/sandbox/uffd/userfaultfd/userfaultfd.go
```diff
@@ -239,8 +261,31 @@ func (u *Userfaultfd) handleMissing(
 			return fmt.Errorf("failed uffdio copy %w", joinedErr)
 		}
 
+		// Add the offset to the missing requests tracker.
+		u.missingRequests.Add(offset)
+
```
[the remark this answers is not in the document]
ValentaTomas: Again.

packages/orchestrator/internal/sandbox/uffd/userfaultfd/userfaultfd.go
```diff
@@ -239,8 +261,31 @@ func (u *Userfaultfd) handleMissing(
...
+func (u *Userfaultfd) Dirty() *block.Tracker {
+	// This will be at worst cancelled when the uffd is closed.
+	u.settleRequests.Lock()
+	// The locking here would work even without using defer (just lock-then-unlock the mutex), but at this point let's make it lock to the clone,
+	// so it is consistent even if there is a another uffd call after.
+	defer u.settleRequests.Unlock()
+
+	return u.missingRequests.Clone()
+}
```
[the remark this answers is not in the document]
ValentaTomas: Slightly valid, but it will be cancelled by the sandbox end. We might refactor in the future.

packages/orchestrator/internal/sandbox/uffd/userfaultfd/userfaultfd.go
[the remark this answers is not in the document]
ValentaTomas: Yes, but there might be no guarantee that the page won't need to be copied again if it was requested, as it might have been swapped out.

packages/orchestrator/internal/sandbox/sandbox.go
ValentaTomas: For now I added the dirty to the Disable return, to make it more clear.

PR #1439, opened 2025-11-04
chore: move copy script to a file

dobrac:
> [!NOTE]
> Replaces the inline copy script template with an embedded `copy_script.sh` file and parses it via `go:embed`.
> 
> - **Build templates (`commands/copy.go`)**:
>   - Replace inline `copyScriptTemplate` with `//go:embed`-loaded `copy_script.sh` and parse it as the template.
> - **Scripts**:
>   - Add `packages/orchestrator/internal/template/build/commands/copy_script.sh` containing the copy logic (workdir resolution, ownership/permissions, symlink/file/dir handling).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 036501a6b9a62425e56e92533ce23725b4e4b50b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/orchestrator/internal/template/build/commands/copy_script.sh
+++ b/packages/orchestrator/internal/template/build/commands/copy_script.sh
@@ -1,0 +1,68 @@
+#!/bin/bash
+
+targetPath="{{ .TargetPath }}"
+sourcePath="{{ .SourcePath }}"
+owner="{{ .Owner }}"
+permissions="{{ .Permissions }}"
+
+workdir="{{ .Workdir }}"
+user="{{ .User }}"
+
+# Fill the workdir with user home directory if empty
+if [ -z "${workdir}" ]; then
+    # Use the user's home directory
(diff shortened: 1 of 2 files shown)


packages/orchestrator/internal/template/build/commands/copy_script.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
```
djeebus, line 1: Using `env` is probably more portable, if you want to make another change here

```suggestion
#!/usr/bin/env bash
```

dobrac: Thats interesting! I would keep it as for now, but we can reevaluate if it causes any issues, e.g. if we want to support alpine and other linux distros

PR #1447, opened 2025-11-06
feat: add sandbox network out configuration

dobrac:
Add sandbox network out traffic configuration. The usage would look then something like this:

```ts
const sbx = await Sandbox.create({
  network: {
    allowOut: ['8.8.8.8'],
    blockOut: ['8.8.0.0/16'],
  },
})
```

---

> [!NOTE]
> Adds configurable egress allow/deny for sandboxes, wires it through API → orchestrator, enforces via firewall, and persists in snapshots; updates proto/spec and tests.
> 
> - **API**:
>   - Add `SandboxNetworkConfig` (`allowOut`, `denyOut`) to spec and generated types; include `network` in `NewSandbox`. Clarify `allow_internet_access` semantics (equivalent to `denyOut: ["0.0.0.0/0"]` when false).
>   - Handlers pass `network` to orchestrator on create/resume/connect.
> - **Orchestrator**:
>   - Accept and translate network config; if `allowInternetAccess` is false, set `deny 0.0.0.0/0`.
>   - Update network pool/slot to configure per-sandbox egress; firewall refactor to allow/deny sets, allow-list precedence, support "block all", default-allow hyperloop IP, and validate private ranges cannot be allowed.
>   - Extend gRPC/proto with `SandboxNetworkConfig` and egress fields.
> - **DB**:
>   - Add `snapshots.config` (JSONB) via migration; define `PausedSandboxConfig` and `SandboxNetworkConfig` types; store network config on pause; update queries/models.
> - **Tests**:
>   - Integration tests for egress allow/deny, CIDR handling, precedence, persistence across pause/resume, and hyperloop accessibility with blocked internet.
>   - Unit tests for firewall address validation.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 738177c7885e42ef8da1d64bbf37870b06815d2f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/orchestrator/internal/sandbox/network/firewall.go
+++ b/packages/orchestrator/internal/sandbox/network/firewall.go
@@ -17,7 +17,7 @@ const (
 	allInternetTrafficAddress = "0.0.0.0/0"
 )
 
-var blockedRanges = []string{
+var deniedRanges = []string{
 	"10.0.0.0/8",
 	"169.254.0.0/16",
 	"192.168.0.0/16",
@@ -30,12 +30,14 @@ type Firewall struct {
 	chain *nftables.Chain
 
 	blockInternetTraffic bool
-	blockSet             set.Set
+	denySet              set.Set
 	allowSet             set.Set
 	tapInterface         string
+
+	allowedRanges []string
 }
 
-func NewFirewall(tapIf string) (*Firewall, error) {
+func NewFirewall(tapIf string, hyperloopIP string) (*Firewall, error) {
 	conn, err := nftables.New(nftables.AsLasting())
 	if err != nil {
 		return nil, fmt.Errorf("new nftables conn: %w", err)
@@ -55,10 +57,10 @@ func NewFirewall(tapIf string) (*Firewall, error) {
 		Policy:   &acceptPolicy,
 	})
 
-	// Create block-set and allow-set
-	blockSet, err := set.New(conn, table, "filtered_blocklist", nftables.TypeIPAddr)
+	// Create deny-set and allow-set
+	denySet, err := set.New(conn, table, "filtered_denylist", nftables.TypeIPAddr)
 	if err != nil {
-		return nil, fmt.Errorf("new block set: %w", err)
+		return nil, fmt.Errorf("new deny set: %w", err)
 	}
 	allowSet, err := set.New(conn, table, "filtered_allowlist", nftables.TypeIPAddr)
 	if err != nil {
@@ -70,9 +72,10 @@ func NewFirewall(tapIf string) (*Firewall, error) {
 		table:                table,
 		chain:                chain,
 		blockInternetTraffic: false,
-		blockSet:             blockSet,
+		denySet:              denySet,
 		allowSet:             allowSet,
 		tapInterface:         tapIf,
+		allowedRanges:        []string{fmt.Sprintf("%s/32", hyperloopIP)},
 	}
 
 	// Add firewall rules to the chain
@@ -83,7 +86,7 @@ func NewFirewall(tapIf string) (*Firewall, error) {
 	// Populate the sets with initial data
 	err = fw.ResetAllCustom()
 	if err != nil {
-		return nil, fmt.Errorf("error while configuring initial block set: %w", err)
+		return nil, fmt.Errorf("error while configuring initial deny set: %w", err)
 	}
 
 	return fw, nil
@@ -133,12 +136,12 @@ func (fw *Firewall) installRules() error {
 		),
 	})
 
-	// Drop anything in blockSet
+	// Drop anything in denySet
 	fw.conn.AddRule(&nftables.Rule{
 		Table: fw.table, Chain: fw.chain,
 		Exprs: append(ifaceMatch,
 			expressions.IPv4DestinationAddress(1),
-			expressions.IPSetLookUp(fw.blockSet.Set(), 1),
+			expressions.IPSetLookUp(fw.denySet.Set(), 1),
 			expressions.Drop(),
 		),
 	})
@@ -150,11 +153,11 @@ func (fw *Firewall) installRules() error {
 	return nil
 }
 
-// AddDeniedCIDR adds a single CIDR to the block set at runtime.
+// AddDeniedCIDR adds a single CIDR to the deny set at runtime.
 func (fw *Firewall) AddDeniedCIDR(cidr string) error {
 	if fw.blockInternetTraffic {
-		// If internet is blocked, we don't need to add any other addresses to the block set.
-		// Because 0.0.0.0/0 is not valid IP per GoLang, we can't add new addresses to the block set.
+		// If internet is denied, we don't need to add any other addresses to the deny set.
+		// Because 0.0.0.0/0 is not valid IP per GoLang, we can't add new addresses to the deny set.
 		return nil
 	}
 
@@ -162,7 +165,7 @@ func (fw *Firewall) AddDeniedCIDR(cidr string) error {
 	if cidr == allInternetTrafficAddress {
 		fw.blockInternetTraffic = true
 
-		fw.conn.FlushSet(fw.blockSet.Set())
+		fw.conn.FlushSet(fw.denySet.Set())
 
 		toAppend := []nftables.SetElement{
 			{Key: netip.MustParseAddr("0.0.0.0").AsSlice()},
@@ -172,11 +175,11 @@ func (fw *Firewall) AddDeniedCIDR(cidr string) error {
 			},
 		}
 
-		if err := fw.conn.SetAddElements(fw.blockSet.Set(), toAppend); err != nil {
-			return fmt.Errorf("add elements to block set: %w", err)
+		if err := fw.conn.SetAddElements(fw.denySet.Set(), toAppend); err != nil {
+			return fmt.Errorf("add elements to denied set: %w", err)
 		}
 	} else {
-		current, err := fw.blockSet.Elements(fw.conn)
+		current, err := fw.denySet.Elements(fw.conn)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


packages/orchestrator/internal/sandbox/network/slot.go
```diff
@@ -248,14 +249,13 @@ func (s *Slot) CloseFirewall() error {
...
-func (s *Slot) ConfigureInternet(ctx context.Context, allowInternet bool) (e error) {
+func (s *Slot) ConfigureInternet(ctx context.Context, network *orchestrator.SandboxNetworkConfig) (e error) {
 	_, span := tracer.Start(ctx, "slot-internet-configure", trace.WithAttributes(
 		attribute.String("namespace_id", s.NamespaceID()),
-		attribute.Bool("allow_internet", allowInternet),
 	))
 	defer span.End()
 
-	if allowInternet {
+	if network == nil || len(network.GetEgress().GetAllowedAddresses()) == 0 && len(network.GetEgress().GetBlockedAddresses()) == 0 {
```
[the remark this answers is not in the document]
dobrac: its ok

packages/api/internal/handlers/sandbox.go
```diff
@@ -37,6 +38,7 @@ func (a *APIStore) startSandbox(
 	autoPause bool,
 	envdAccessToken *string,
 	allowInternetAccess *bool,
+	network *types.SandboxNetworkConfig,
```
sitole, line 41: Lets merge `allowInternetAccess` with network struct

packages/api/internal/orchestrator/create_instance.go
```diff
@@ -13,25 +13,50 @@ import (
...
+
+	// Copy network configuration if provided
+	if network != nil && network.Egress != nil {
+		orchNetwork.Egress.AllowedAddresses = network.Egress.AllowedAddresses
+		orchNetwork.Egress.BlockedAddresses = network.Egress.BlockedAddresses
+	}
+
+	// Handle the case where internet access is explicitly disabled
+	// This should be applied after copying the network config to preserve allowed addresses
+	if allowInternetAccess != nil && !*allowInternetAccess {
```
sitole, line 61: I think we should return error when both network egress allow/block lists and allow internet are provided. It can bring unexpected behavior for users.

packages/api/internal/orchestrator/create_instance.go
```diff
@@ -138,6 +164,8 @@ func (o *Orchestrator) CreateSandbox(
 		sbxDomain = cluster.SandboxDomain
 	}
 
+	orchNetwork := buildNetworkConfig(network, allowInternetAccess)
```
sitole:
```suggestion
	sbxNetwork := buildNetworkConfig(network, allowInternetAccess)
```

packages/api/internal/orchestrator/nodemanager/sandboxes.go
```diff
@@ -46,6 +47,13 @@ func (n *Node) GetSandboxes(ctx context.Context) ([]sandbox.Sandbox, error) {
 			return nil, fmt.Errorf("failed to parse build ID '%s' for job: %w", config.GetBuildId(), parseErr)
 		}
 
+		network := &types.SandboxNetworkConfig{
+			Egress: &types.SandboxNetworkEgressConfig{
+				AllowedAddresses: config.GetNetwork().GetEgress().GetAllowedAddresses(),
```
sitole: Are we sure this call chain will work correctly when communicating with orchestrators that are not yet supporting network configuration?

packages/orchestrator/internal/sandbox/network/firewall.go
```diff
@@ -146,9 +151,17 @@ func (fw *Firewall) installRules() error {
 }
 
 // AddBlockedIP adds a single CIDR to the block set at runtime.
-func (fw *Firewall) AddBlockedIP(cidr string) error {
+func (fw *Firewall) AddBlockedIP(address string) error {
```
sitole: We switched to inserting blocked IPs and not using CIDRs, which seems cleaner to me.

sitole: Function comment still refers to CIDRs

packages/orchestrator/internal/sandbox/network/firewall.go
```diff
@@ -231,10 +254,50 @@ func (fw *Firewall) ResetBlockedCustom() error {
...
+	addressData, err := set.AddressStringsToSetData([]string{address})
+	if err != nil {
+		return err
+	}
+
+	if len(addressData) == 0 {
+		return fmt.Errorf("address %s is not a valid IP address", address)
+	}
+
+	// Convert single IP address to prefix for comparison
```
sitole: I would prefer to do this on api side during the request and then refer just with CIDR and forget about the single IP address.

packages/orchestrator/internal/sandbox/network/slot.go
sitole: Lets use local var for network.GetEgress() to make IF nicer

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -99,6 +102,25 @@ func (s *Server) Create(ctx context.Context, req *orchestrator.SandboxCreateRequ
 		return nil, fmt.Errorf("failed to get template snapshot data: %w", err)
 	}
 
+	// Clone the network config to avoid modifying the original request
+	network := proto.CloneOf(req.GetSandbox().GetNetwork())
+
+	// TODO: Temporarily set this based on global config, should be removed later
+	//  (it should be passed network config from API)
```
sitole, line 110: Add linear task link as comment here

sitole (changes requested):

packages/api/internal/orchestrator/nodemanager/sandboxes.go
dobrac: yes: https://go.dev/play/p/TbywUVUKQ_4

packages/orchestrator/internal/sandbox/network/firewall.go
dobrac: you can insert:
- One IP
- Cidr
- IP range (1.1.1.0-1.1.1.1)

packages/orchestrator/internal/sandbox/network/firewall.go
dobrac: do you think it makes sense to do the IP/CIDR/Range parsing ourselves when the library does it already?

packages/api/internal/handlers/sandbox.go
dobrac: let's do this only in the database for now, I still need three states for the allowInternetAccess: true, false, nil

packages/api/internal/orchestrator/create_instance.go
dobrac: The allowOut takes precedence over blockOut

also, internet is defined by default, allowInternetAccess false does the same as blockOut=[0.0.0.0/0]

sitole:
Btw @dobrac can we check with deny all hyperloop server is still working?

packages/orchestrator/internal/sandbox/network/firewall.go
sitole: Can we make it cidr only to simplify thinking around? I think api should move all three types (i dont like IP range tho) to CIDR format.

packages/orchestrator/internal/sandbox/network/firewall.go
sitole: We can use library, iam okay with it

packages/api/internal/orchestrator/nodemanager/sandboxes.go
sitole: Thing was if we will not receive nil. It not, that is okay.

dobrac: we'll receive nil for the list, but nil is a valid empty list

packages/api/internal/handlers/sandbox.go
dobrac: let's remove it after we remove the default internet access allow in the orchestrator

packages/orchestrator/internal/sandbox/network/firewall.go
dobrac: removed support for IP ranges, passing now addresses only as CIDR blocks to the orchestrator. In the DB, they're saved as user provided

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -99,6 +102,26 @@ func (s *Server) Create(ctx context.Context, req *orchestrator.SandboxCreateRequ
...
+	}
+	if !allowInternet {
+		if network == nil {
+			network = &orchestrator.SandboxNetworkConfig{}
+		}
+		if network.GetEgress() == nil {
+			network.Egress = &orchestrator.SandboxNetworkEgressConfig{}
+		}
+		network.Egress.DeniedCidrs = []string{internetBlockCIDR}
+	}
```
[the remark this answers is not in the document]
sitole: I think here we should fallback to `s.config.AllowSandboxInternet` only when customers configuration is not provided at all (both allow internet access flag and egress internet config).

cc @dobrac

spec/openapi.yml
```diff
@@ -475,7 +489,11 @@ components:
           description: Secure all system communication with sandbox
         allow_internet_access:
```
sitole, line 490: Can we deprecate this?

packages/orchestrator/internal/server/sandboxes.go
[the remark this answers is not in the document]
dobrac: thats desired

spec/openapi.yml
dobrac: I think we might benefit in copying the SDK functionality in the API more. Let's keep it for now and revisit it again later after some feedback on the network setup DX

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -170,7 +170,8 @@ func (f *Factory) CreateSandbox(
 	apiConfigToStore *orchestrator.SandboxConfig,
 ) (s *Sandbox, e error) {
 	ctx, span := tracer.Start(ctx, "create sandbox")
-	defer func() { endSpan(span, e) }()
+	defer span.End()
+	defer handleSpanError(span, &e)
```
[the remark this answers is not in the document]
dobrac: no

PR #1457, opened 2025-11-07
Remove response and request content from logs

jakubno:
> [!NOTE]
> Filters gRPC request/response payloads from logs and adds a debug log when evicting sandboxes.
> 
> - **Logging**:
>   - **gRPC logger (`packages/shared/pkg/logger/grpc.go`)**:
>     - Ignore `grpc.request.content` and `grpc.response.content` fields to avoid logging payloads.
>     - Safer key handling and field construction; skip non-string keys and excluded fields.
> - **Orchestrator**:
>   - **Evictor (`packages/api/internal/orchestrator/evictor/evict.go`)**: Add debug log when evicting a sandbox with `sandbox_id` and `state_action`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b599e9d2ec723fc2d766968a8737503c7a76394e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/shared/pkg/logger/grpc.go
+++ b/packages/shared/pkg/logger/grpc.go
@@ -28,7 +28,7 @@ func GRPCLogger(l *zap.Logger) logging.Logger {
 			"grpc.code":        "-",
 		}
 
-		fieldsLength := 0
+		fieldsCount := 0
 		for i := 0; i < len(fields)-1; i += 2 {
 			key, ok := fields[i].(string)
 			if !ok {
@@ -38,7 +38,7 @@ func GRPCLogger(l *zap.Logger) logging.Logger {
 			if _, ok := ignoredFields[key]; ok {
 				continue
 			}
-			fieldsLength++
+			fieldsCount++
 
 			value := fields[i+1]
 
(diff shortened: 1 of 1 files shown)


ValentaTomas (approved):

packages/shared/pkg/logger/grpc.go
```diff
@@ -24,23 +45,28 @@ func GRPCLogger(l *zap.Logger) logging.Logger {
 		}
 
 		for i := 0; i < len(fields)-1; i += 2 {
-			key := fields[i]
+			key := fields[i].(string)
```
djeebus: To prevent edge case panic catastrophe, we should check for success:


```suggestion
			key, ok := fields[i].(string)
			if !ok {
				l.Warn("field is not a string", zap.Any(fields[i]))
				continue
			}
```

djeebus: Oops, I see that this is the same loop as above. Ignore!

djeebus: Sorry, now that the logic is simplified, this matters again. We need to make sure we don't panic if somehow a key is not a string.


```suggestion
			key, ok := fields[i].(string)
			if !ok {
				l.Warn("field is not a string", zap.Any(fields[i]))
				continue
			}
```

PR #1474, opened 2025-11-11
Allow optional sandbox ingress traffic auth 

sitole:
- Adds ability to specify `allowPublicAccess: false` during sandbox creation and access token needed for sbx traffic.
- Configuration is persisted when sandbox is resumed.
- Envd port is always allowed as there is different auth mechanism for it.
- Sandbox traffic token can be re-constructed from master hash key and sandbox id so we dont need to store in db.

---

> [!NOTE]
> Adds optional private sandbox ingress that requires a per-sandbox traffic access token, enforced by the proxy, with API/proto updates and tests.
> 
> - **API/Schema**:
>   - Add `allowPublicTraffic` to `SandboxNetworkConfig` and expose `trafficAccessToken` on `Sandbox` responses.
>   - Regenerate OpenAPI spec and generated types.
> - **Handlers**:
>   - On create, build `network.ingress` and validate: disallow non-public traffic unless `secure` (envd) is enabled; generate `EnvdAccessToken` via new generator.
>   - On get, use new generator for envd token when secure.
> - **Token generation**:
>   - Replace `EnvdAccessTokenGenerator` with unified `AccessTokenGenerator` supporting `GenerateEnvdAccessToken` and `GenerateTrafficAccessToken`.
>   - Wire generator through `APIStore` and `Orchestrator`.
> - **Orchestrator**:
>   - Extend network config with `ingress.traffic_access_token` (proto + generated pb).
>   - Produce `trafficAccessToken` when `AllowPublicAccess` is false; pass through to runtime sandbox and network config.
>   - Node manager maps ingress to `AllowPublicAccess`.
> - **Proxy**:
>   - Enforce `x-e2b-traffic-access-token` for sandboxes with private ingress; add specific errors and browser-friendly HTML templates for missing/invalid token.
>   - Improve error handling types.
> - **DB types**:
>   - Add `SandboxNetworkIngressConfig` and ingress to `SandboxNetworkConfig` (persisted for pause/resume).
> - **Tests**:
>   - Add integration tests for traffic token enforcement (missing/invalid/valid) and sandbox-not-found page.
>   - Add test asserting 400 when disabling public traffic without `secure` envd flag.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c6889f7cc6c1d7a1d503ef18c7341c19da994ae1. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/orchestrator/internal/proxy/proxy.go
+++ b/packages/orchestrator/internal/proxy/proxy.go
@@ -61,7 +61,7 @@ func NewSandboxProxy(meterProvider metric.MeterProvider, port uint16, sandboxes 
 				if accessTokenRaw == "" {
 					return nil, reverseproxy.NewErrMissingTrafficAccessToken(sandboxId, trafficAccessTokenHeader)
 				} else if accessTokenRaw != *accessToken {
-					return nil, reverseproxy.NewErrInvalidTrafficAccessToken(sandboxId)
+					return nil, reverseproxy.NewErrInvalidTrafficAccessToken(sandboxId, trafficAccessTokenHeader)
 				}
 			}
 
--- a/packages/shared/pkg/proxy/handler.go
+++ b/packages/shared/pkg/proxy/handler.go
@@ -80,7 +80,7 @@ func handler(p *pool.ProxyPool, getDestination func(r *http.Request) (*pool.Dest
 			zap.L().Warn("traffic access token is invalid", zap.String("host", r.Host))
 
 			err = template.
-				NewTrafficAccessTokenInvalidHeader(trafficInvalidTokenErr.SandboxId, r.Host).
+				NewTrafficAccessTokenInvalidHeader(trafficInvalidTokenErr.SandboxId, r.Host, trafficInvalidTokenErr.Header).
 				HandleError(w, r)
 			if err != nil {
 				zap.L().Error("failed to handle traffic invalid traffic access token header error", zap.Error(err), logger.WithSandboxID(trafficInvalidTokenErr.SandboxId))
--- a/packages/shared/pkg/proxy/errors.go
+++ b/packages/shared/pkg/proxy/errors.go
@@ -51,14 +51,16 @@ func NewErrMissingTrafficAccessToken(sandboxId, header string) *MissingTrafficAc
 
 type InvalidTrafficAccessTokenError struct {
 	SandboxId string
+	Header    string
 }
 
 func (e InvalidTrafficAccessTokenError) Error() string {
 	return "invalid traffic access token"
 }
 
-func NewErrInvalidTrafficAccessToken(sandboxId string) *InvalidTrafficAccessTokenError {
+func NewErrInvalidTrafficAccessToken(sandboxId string, header string) *InvalidTrafficAccessTokenError {
 	return &InvalidTrafficAccessTokenError{
 		SandboxId: sandboxId,
+		Header:    header,
 	}
 }
--- a/packages/shared/pkg/proxy/template/browser_traffic_access_token_invalid_error.html
+++ b/packages/shared/pkg/proxy/template/browser_traffic_access_token_invalid_error.html
@@ -10,7 +10,7 @@
     <img alt="Logo" class="logo" src="https://hebbkx1anhila5yf.public.blob.vercel-storage.com/Symbol%20Gradient-Kr5pnWlK3ZhzBcRGf6Am4cNbJvY1Ge.svg">
     <div class="error-header">
         <h1 class="error-title">Invalid Traffic Access Token</h1>
-        <p class="error-message">The sandbox <span class="highlight" id="sandbox-id">{{.SandboxId}}</span> enforces traffic access token. Token header is invalid.</p>
+        <p class="error-message">The sandbox <span class="highlight" id="sandbox-id">{{.SandboxId}}</span> enforces traffic access token. Token header <span class="highlight">"{{.Header}}"</span> is invalid.</p>
     </div>
 </main>
 </body>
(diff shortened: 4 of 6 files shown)


packages/orchestrator/internal/proxy/proxy.go
```diff
@@ -48,6 +51,22 @@ func NewSandboxProxy(meterProvider metric.MeterProvider, port uint16, sandboxes
...
 			}
 
+			var accessToken *string = nil
+			if net := sbx.Config.Network; net != nil && net.GetIngress() != nil {
+				accessToken = net.GetIngress().TrafficAccessToken
+			}
+
+			// Handle traffic access token validation.
+			// We are skipping envd port as it has its own access validation mechanism.
+			if accessToken != nil && int64(port) != consts.DefaultEnvdServerPort {
```
sitole: We can avoid checking for envd port, but i think it will make SDK handling easier.
Maybe we should enforce `secure: true` flag for sandboxes that are run with `allowPublicTraffic: false` to make it consistent.

sitole: Not sure how this can be easily handled for things such as code interpreter that communicates on different port.

packages/shared/pkg/proxy/handler.go
```diff
@@ -58,6 +58,20 @@ func handler(p *pool.ProxyPool, getDestination func(r *http.Request) (*pool.Dest
...
+			return
+		}
+
+		if errors.Is(err, ErrInvalidTrafficAccessToken) {
+			zap.L().Warn("traffic access token is invalid", zap.String("host", r.Host))
+			http.Error(w, "Sandbox is secured with traffic access token. Provided access token is invalid.", http.StatusForbidden)
+
+			return
+		}
+
```
sitole: Should we make pretty error pages here as we are doing with sandbox not found and port closed?

packages/api/internal/orchestrator/nodemanager/sandboxes.go
```diff
@@ -47,7 +47,11 @@ func (n *Node) GetSandboxes(ctx context.Context) ([]sandbox.Sandbox, error) {
 			return nil, fmt.Errorf("failed to parse build ID '%s' for job: %w", config.GetBuildId(), parseErr)
 		}
 
+		networkTrafficAccessToken := config.GetNetwork().GetIngress().TrafficAccessToken
 		network := &types.SandboxNetworkConfig{
+			Ingress: &types.SandboxNetworkIngressConfig{
+				AllowPublicAccess: networkTrafficAccessToken == nil,
+			},
```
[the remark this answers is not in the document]
sitole: This is using gRPC structs that returns empty struct when not defined.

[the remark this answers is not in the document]
sitole: This is using gRPC structs that returns empty struct when not defined.

spec/openapi.yml
```diff
@@ -220,6 +220,10 @@ components:
     SandboxNetworkConfig:
       type: object
       properties:
+        allowPublicAccess:
```
jakubno: shouldn't we reverse it to name it so we can default to false?

packages/shared/pkg/proxy/handler.go
jakubno: I think it would make sense at least for missing, if you try to open it in the browser

packages/orchestrator/internal/proxy/proxy.go
jakubno: what would be the reason to not check for this for envd?

packages/shared/pkg/proxy/handler.go
```diff
@@ -58,6 +58,20 @@ func handler(p *pool.ProxyPool, getDestination func(r *http.Request) (*pool.Dest
 			return
 		}
 
+		if errors.Is(err, ErrMissingTrafficAccessToken) {
+			zap.L().Warn("traffic access token header is missing", zap.String("host", r.Host))
+			http.Error(w, "Sandbox is secured with traffic access token. Access token header is missing", http.StatusForbidden)
```
jakubno: can you specify the header?

jakubno (changes requested):

packages/orchestrator/internal/proxy/proxy.go
sitole: 1. Double auth, that means two headers
2. Not working envd file upload and download URLs
3. Double auth logic in SDKs

sitole: For context, if we will add ingress port firewall we will do also exception for envd port, right?

packages/shared/pkg/proxy/handler.go
sitole: Yep, will do

spec/openapi.yml
sitole: Okay so `denyPublicTraffic` with default false?

sitole:
Changes added, only unresolved thing now is https://github.com/e2b-dev/infra/pull/1474#discussion_r2518161319. @jakubno

spec/openapi.yml
jakubno: actually let's stick with `allow` to prevent double negation. Sorry 😞

packages/orchestrator/internal/proxy/proxy.go
jakubno: > Not working envd file upload and download URLs

That's a good point

> Double auth logic in SDKs

I think we should enforce `secure: true`, otherwise it doesn't make sense and it can be weird security issue

jakubno: > For context, if we will add ingress port firewall we will do also exception for envd port, right?

You want to have different limits for e.g. API and sandbox traffic, if that's what you've been asking

sitole: Okay, so after sync we decided to enforce `secure: true` and remove envd from proxy logic exception as it will be more unified (this will require to introduce header also in SDKs).

jakubno (approved):

PR #1475, opened 2025-11-11
Add lock for NFS write

jakubno:
Check that the lock works as expected (logs were removed afterwards)
image

Left is with locking (cache NFS was cleaned and orchestrator redeployed, so both caches were clean)

image
image

*Without the lock the same load timed out.*

---

> [!NOTE]
> Introduce a file-based lock with TTL and integrate it into cache writes to prevent concurrent NFS writes; add comprehensive tests.
> 
> - **Storage / Locking**:
>   - **New `packages/shared/pkg/storage/lock`**: Implement file-based lock with TTL cleanup (`TryAcquireLock`, `ReleaseLock`, `ErrLockAlreadyHeld`).
>   - **Tests**: Add concurrency, staleness, and path consistency tests for the lock.
> - **Cache Providers**:
>   - **`CachedObjectProvider`**: Wrap full-file cache writes with lock acquisition; skip on `ErrLockAlreadyHeld`; add warnings.
>   - **`CachedSeekableObjectProvider`**:
>     - Lock around chunk writes in `writeChunkToCache`.
>     - Lock around size writes in `writeLocalSize`.
>     - Skip writes if lock already held; emit warnings on lock errors.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit cc765fcf10faa81fbada1b9b6b9321253c962ea4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/shared/pkg/storage/lock/file_lock_test.go
+++ b/packages/shared/pkg/storage/lock/file_lock_test.go
@@ -120,7 +120,6 @@ func TestConcurrentLockAcquisition(t *testing.T) {
 			case errors.Is(err, ErrLockAlreadyHeld):
 				results <- result{alreadyHeld: true}
 			case err != nil:
-				time.Sleep(50 * time.Millisecond)
 				results <- result{err: err}
 			default:
 				_ = ReleaseLock(file)


ValentaTomas:
Alternative—if we limit the concurrency of writing things to cache per orchestrator, can the usage reasonably flatten (until we have too many orchestrators)?

packages/shared/pkg/storage/lock/file_lock_test.go
```diff
@@ -0,0 +1,174 @@
...
+	// Try to acquire lock concurrently from multiple goroutines
+	for range routines {
+		go func() {
+			file, err := TryAcquireLock(testPath)
+
+			switch {
+			case errors.Is(err, ErrLockAlreadyHeld):
+				results <- result{alreadyHeld: true}
+			case err != nil:
+				time.Sleep(50 * time.Millisecond)
```
dobrac: why is here the sleep, but not for ErrLockAlreadyHeld?

jakubno:
```suggestion
```

jakubno: Didn't make any sense, it was in go routine and had no effect on the result

PR #1479, opened 2025-11-12
feat: add mask request host

dobrac:
- [x] Requires https://github.com/e2b-dev/infra/pull/1474 to be merged first

---

> [!NOTE]
> Adds `maskRequestHost` to sandbox network config and proxies non-envd traffic with a masked Host (with `${PORT}` substitution), including validation, propagation, and tests.
> 
> - **API**:
>   - Add `network.maskRequestHost` to `NewSandbox` handling with ASCII validation; accept host or host:port (port optional).
>   - Propagate to internal `types.SandboxNetworkConfig` and pass through to orchestrator creation.
> - **Orchestrator/Proxy**:
>   - Extend proto (`SandboxNetworkIngressConfig.mask_request_host`) and generated stubs.
>   - Proxy masks non-envd requests: sets `X-Forwarded-Host` and overrides `Host`; supports `${PORT}` placeholder; logs `mask_request_host`.
>   - Add `MaskRequestHost` to proxy pool `Destination` and rewrite logic.
> - **Spec/Types**:
>   - Update OpenAPI (`spec/openapi.yml`) and generated client/server types to include `maskRequestHost`.
> - **Tests**:
>   - Unit test for proxy host masking behavior.
>   - Integration tests for `maskRequestHost` API param (valid/invalid), traffic access token paths, sandbox-not-found/closed-port; refactor request helpers.
> - **Deps/Chores**:
>   - Add `golang.org/x/net` to API module; minor go.sum tidy.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit bf6cbd024d926c4825d2628f2cde6508a7173237. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api/internal/handlers/sandbox_create.go
+++ b/packages/api/internal/handlers/sandbox_create.go
@@ -173,10 +173,17 @@ func (a *APIStore) PostSandboxes(c *gin.Context) {
 				return
 			}
 
-			_, err = idna.Display.ToUnicode(hostname)
+			host, err := idna.Display.ToASCII(hostname)
 			if err != nil {
 				telemetry.ReportError(ctx, "error when parsing mask request host", err, telemetry.WithSandboxID(sandboxID))
 				a.sendAPIStoreError(c, http.StatusBadRequest, fmt.Sprintf("Invalid mask request host: %s", err))
+
+				return
+			}
+
+			if !strings.EqualFold(host, hostname) {
+				telemetry.ReportError(ctx, "mask request host is not ASCII", nil, telemetry.WithSandboxID(sandboxID), attribute.String("mask_request_host", hostname), attribute.String("mask_request_host_ascii", host))
+				a.sendAPIStoreError(c, http.StatusBadRequest, fmt.Sprintf("Mask request host '%s' is not ASCII. Please use ASCII characters only.", hostname))
 
 				return
 			}


packages/api/internal/orchestrator/nodemanager/sandboxes.go
```diff
@@ -51,6 +51,7 @@ func (n *Node) GetSandboxes(ctx context.Context) ([]sandbox.Sandbox, error) {
 		network := &types.SandboxNetworkConfig{
 			Ingress: &types.SandboxNetworkIngressConfig{
 				DenyPublicAccess: networkTrafficAccessToken != nil,
+				MaskRequestHost:  config.GetNetwork().GetIngress().MaskRequestHost,
```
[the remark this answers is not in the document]
dobrac: not true

packages/shared/pkg/proxy/pool/client.go
```diff
@@ -103,8 +103,15 @@ func newProxyClient(
 			}
 
 			r.SetURL(t.Url)
-			// We are **not** using SetXForwarded() because servers can sometimes modify the content-location header to be http which might break some customer services.
-			r.Out.Host = r.In.Host
+
+			if t.MaskRequestHost != nil {
+				// Mask the request host to bypass source host protections.
+				r.Out.Header.Set("X-Forwarded-Host", r.In.Host)
+				r.Out.Host = *t.MaskRequestHost
```
djeebus, line 110: Are we certain that we want to change the ip/port to connect to, or do we just want the `Host` http request header to change?

dobrac: just the http request header to change

djeebus: Changing the `Host` field doesn't just set the header, I believe it changes where you send the request as well. I'll see if I can verify that.

dobrac: I think the `req.URL` (or `r.SetURL(url)`) is what controls where the request goes, the `r.Out.Host` is just the header

djeebus: Huh. Turns out that, when writing the http request, it pulls the host [from the struct](https://github.com/golang/go/blob/master/src/net/http/request.go#L605) and then writes it [out of band](https://github.com/golang/go/blob/master/src/net/http/request.go#L679-L685). I'll push the test on to your branch, but it just confirms that the code works 👍

djeebus: One suggestion: instead of using `X-Forwarded-Host`, we might want to consider using more [widely accepted headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers#proxies)

dobrac: `X-Forwarded-Host` is actually [widely accepted header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Forwarded)

spec/openapi.yml
```diff
@@ -234,6 +234,9 @@ components:
           description: List of denied CIDR blocks or IP addresses for egress traffic
           items:
             type: string
+        maskRequestHost:
+          type: string
+          description: Specify all sandbox requests host mask.
```
jakubno:
```suggestion
          description: Specify host mask which will be used for all sandbox requests 
```

packages/api/internal/handlers/sandbox_create.go
```diff
@@ -156,13 +159,37 @@ func (a *APIStore) PostSandboxes(c *gin.Context) {
 		envdAccessToken = &accessToken
 	}
 
+	var maskRequestHost *string = nil
+	if body.Network != nil && body.Network.MaskRequestHost != nil {
```
jakubno: wouldn't it make sense to do it inside the already present `if body.Network != nil` block?

packages/api/internal/handlers/sandbox_create.go
```diff
@@ -156,13 +159,37 @@ func (a *APIStore) PostSandboxes(c *gin.Context) {
...
+
+		hostname, _, err := splitHostPortOptional(host)
+		if err != nil {
+			telemetry.ReportError(ctx, "error when splitting mask request host", err, telemetry.WithSandboxID(sandboxID))
+			a.sendAPIStoreError(c, http.StatusBadRequest, fmt.Sprintf("Invalid mask request host: %s", err))
+
+			return
+		}
+
+		_, err = idna.Display.ToUnicode(hostname)
```
jakubno: shouldn't we enforce the hostname to be in Punycode?

packages/orchestrator/internal/proxy/proxy.go
```diff
@@ -67,6 +71,13 @@ func NewSandboxProxy(meterProvider metric.MeterProvider, port uint16, sandboxes
 				}
 			}
 
+			// Handle request host masking only for non-envd traffic.
+			var maskRequestHost *string = nil
+			if h := sbx.Config.Network.GetIngress().GetMaskRequestHost(); isNonEnvdTraffic && h != "" {
+				h = strings.ReplaceAll(h, pool.MaskRequestHostPortPlaceholder, strconv.FormatUint(port, 10))
```
jakubno, line 77: How did the placeholder get to the host?

packages/api/internal/handlers/sandbox_create.go
dobrac: the host will be automatically translated to ASCII in the reverse proxy. I think we can allow users to define the hostname in unicode, but of course we can also require the ASCII format directly

packages/orchestrator/internal/proxy/proxy.go
dobrac: the user can supply ${PORT} to allow the host to not be always like "localhost:3000", but it can be "localhost:${PORT}", where the port is automatically resolved to the requested port number on demand

jakubno (approved):

PR #1485, opened 2025-11-13
fix: slot return on network fail and improve validation

dobrac:
> [!NOTE]
> Centralizes sandbox network helpers, refactors firewall to predefined/user allow/deny sets with 0.0.0.0/0 handling, validates maskRequestHost, and returns slots on network config failure with expanded integration tests and deps updates.
> 
> - **API**:
>   - Validate `network.maskRequestHost` (IDNA/ASCII, optional port) and surface clear `400` errors.
>   - Build network config using shared helpers; use `AllInternetTrafficCIDR` when `allowInternetAccess=false`.
>   - Replace secure token error logging with telemetry.
> - **Orchestrator**:
>   - Return network slot to pool when internet configuration fails in `Pool.Get`.
>   - Refactor firewall:
>     - Separate predefined allow/deny vs user allow/deny sets; enforce rule order (allow→deny).
>     - Reset functions renamed to `ResetAllSets`, `ResetAllowedSets`, `ResetDeniedSets`.
>     - Proper handling of `0.0.0.0/0` and incremental set updates.
>   - Use shared CIDR conversion and `AllInternetTrafficCIDR` in server create path and instance builder.
> - **Shared**:
>   - New `sandbox-network` package with `AllInternetTrafficCIDR`, default denied ranges, and address→CIDR helpers.
> - **Tests**:
>   - Add/extend egress firewall integration tests (allow/block IPs & CIDRs, precedence, persistence, 0.0.0.0/0 cases).
>   - Add unit tests for CIDR helper; remove obsolete firewall allow validation test.
> - **Dependencies**:
>   - Add/update networking and logging libs (e.g., `ngrok/firewall_toolkit`, `google/nftables`, `mdlayher/*`, `zerolog`) in `api`/`shared` modules.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 73a2bd22b393e19f88d8dc0b7e16caba69ceb276. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/orchestrator/internal/sandbox/network/firewall.go
+++ b/packages/orchestrator/internal/sandbox/network/firewall.go
@@ -266,6 +266,7 @@ func addCIDRToSet(conn *nftables.Conn, ipset set.Set, cidr string) error {
 		return err
 	}
 
+	// The checked range is 0.0.0.0 to 255.255.255.254, because when 255.255.255.255 is added, it's then requested as 255.255.255.254.
 	if len(current) == 1 && current[0].AddressRangeStart == netip.MustParseAddr("0.0.0.0") && current[0].AddressRangeEnd == netip.MustParseAddr("255.255.255.254") {
 		// Because 0.0.0.0/0 is not valid IP per GoLang, we can't add new addresses to the set.
 		return nil


tests/integration/internal/tests/api/sandboxes/sandbox_network_out_test.go
```diff
@@ -228,31 +228,6 @@ func TestEgressFirewallEmptyConfig(t *testing.T) {
 	require.NoError(t, err, "Expected curl to succeed with empty allowOut list")
 }
 
-// TestEgressFirewallAllowAll tests that 0.0.0.0/0 allows all traffic
-func TestEgressFirewallAllowAll(t *testing.T) {
```
jakubno, line 232: why is this removed?

packages/shared/pkg/sandbox-network/firewall.go
```diff
@@ -0,0 +1,70 @@
...
+	"169.254.0.0/16",
+	"192.168.0.0/16",
+	"172.16.0.0/12",
+}
+
+var deniedSandboxData = utils.Must(set.AddressStringsToSetData(DeniedSandboxCIDRs))
+
+// CanAllowCIDR checks if the address is in the default denied ranges.
+func CanAllowCIDR(cidr string) error {
+	if cidr == AllInternetTrafficCIDR {
```
jakubno: not a huge fan of this, I would suspect adding 0.0.0.0/0 would work. It moves some if handling to the user.

packages/shared/pkg/sandbox-network/firewall.go
```diff
@@ -0,0 +1,70 @@
...
+	if err != nil {
+		return err
+	}
+
+	if len(addressData) == 0 {
+		return fmt.Errorf("address %s is not a valid IP address", cidr)
+	}
+
+	for _, deniedRange := range deniedSandboxData {
+		if deniedRange.Prefix.Overlaps(addressData[0].Prefix) {
```
jakubno: what happens if the there's overlap, does allow or deny have preference, does it fail or it's undefined?

tests/integration/internal/tests/api/sandboxes/sandbox_network_out_test.go
dobrac: The test was moved to the blocked ranges when trying to allowOut

packages/shared/pkg/sandbox-network/firewall.go
dobrac: allow has always preference, hence you can't allow any IP from the blocked ranges

packages/shared/pkg/sandbox-network/firewall.go
dobrac: so if there is 0.0.0.0/0, it would behave exactly the same as if you didn't supply any limits at all

I think I can implement it that way if you think it's better

dobrac: updated it so that 0.0.0.0/0 clears all network restrictions

packages/api/internal/orchestrator/create_instance.go
```diff
@@ -67,10 +53,18 @@ func buildNetworkConfig(network *types.SandboxNetworkConfig, allowInternetAccess
...
-		orchNetwork.Egress.DeniedCidrs = []string{internetBlockCIDR}
+		orchNetwork.Egress.DeniedCidrs = []string{sandbox_network.AllInternetTrafficCIDR}
+	}
+
+	if slices.Contains(orchNetwork.GetEgress().GetAllowedCidrs(), sandbox_network.AllInternetTrafficCIDR) {
+		// If all traffic is allowed, clear the allowed and denied lists
+		// Internet access is enabled by default.
+		orchNetwork.Egress.AllowedCidrs = nil
+		orchNetwork.Egress.DeniedCidrs = nil
+		aia = ut.ToPtr(true)
```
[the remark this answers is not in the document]
dobrac: yeah, this is now confusing 😢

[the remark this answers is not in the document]
dobrac: or maybe it's kinda okay-ish. Just that the allowOut has always precedence even in this case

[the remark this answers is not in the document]
dobrac: what is confusing now a bit, is that anything like 0.0.0.0/1 will fail, because that will hit the blocked ranges

tests/integration/internal/tests/api/sandboxes/sandbox_network_out_test.go
jakubno: what even should happen here?

packages/api/internal/orchestrator/create_instance.go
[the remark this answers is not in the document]
dobrac: fixed now by allows allowing it

packages/orchestrator/internal/sandbox/network/firewall.go
```diff
@@ -155,148 +187,117 @@ func (fw *Firewall) installRules() error {
...
+	// 0.0.0.0/0 is not valid IP per GoLang, so we handle it as a special case
+	if cidr == sandbox_network.AllInternetTrafficCIDR {
+		conn.FlushSet(ipset.Set())
+
+		toAppend := []nftables.SetElement{
+			{
+				Key: netip.MustParseAddr("0.0.0.0").AsSlice(),
+			},
+			{
+				Key:         netip.MustParseAddr("255.255.255.255").AsSlice(),
```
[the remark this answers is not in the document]
dobrac: this is correct, when 255.255.255.255 is added, it's then requested as 255.255.255.254

jakubno (approved):

PR #1511, opened 2025-11-19
Add CPU info to the service info

jakubno:
> [!NOTE]
> Expose machine CPU metadata (arch/family/model/name) from orchestrators via gRPC ServiceInfo and surface it in API Node/NodeDetail; update proto/OpenAPI, generated code, and node manager/admin to collect and return it.
> 
> - **API/Spec**:
>   - Add `MachineInfo` schema (cpuArchitecture/cpuFamily/cpuModel/cpuModelName) and include it in `Node` and `NodeDetail`.
>   - Regenerate OpenAPI client/server code (`spec.gen.go`, `types.gen.go`, tests models).
> - **gRPC/Proto**:
>   - Introduce `MachineInfo` message and add `machine_info` to `ServiceInfoResponse`.
>   - Regenerate protobufs.
> - **Orchestrator Service**:
>   - Detect machine CPU info (`machineinfo.Detect`) at startup and store in `ServiceInfo`.
>   - Return machine info in `InfoService.ServiceInfo` (with converter).
> - **Node Manager/Admin**:
>   - Store machine info in `nodemanager.Node`; update on sync and creation.
>   - Include `machineInfo` in admin `AdminNodes` and `AdminNodeDetail` responses.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a2ce0f44c8a9cf3c5acf86b7a34696a31710a480. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/orchestrator/internal/service/service_info.go
+++ b/packages/orchestrator/internal/service/service_info.go
@@ -126,7 +126,6 @@ func convertMachineInfo(machineInfo machineinfo.MachineInfo) *orchestratorinfo.M
 	}
 }
 
-
 func (s *Server) ServiceStatusOverride(ctx context.Context, req *orchestratorinfo.ServiceStatusChangeRequest) (*emptypb.Empty, error) {
 	logger.L().Info(ctx, "service status override request received", zap.String("status", req.GetServiceStatus().String()))
 	s.info.SetStatus(ctx, req.GetServiceStatus())


packages/orchestrator/info.proto
```diff
@@ -34,6 +34,7 @@ message ServiceInfoResponse {
   ServiceInfoStatus service_status = 51;
   repeated ServiceInfoRole service_roles = 52;
   google.protobuf.Timestamp service_startup = 53;
+  string cpu_platform = 54;
```
sitole: Can we maybe put it into struct "machine" or something? i think there will be more values needed now or in near future and would be nice to have them namespaced

sitole: And same in api orchestrator info struct

packages/shared/pkg/platform/platform.go
```diff
@@ -0,0 +1,59 @@
...
+	// AMD platforms (newest to oldest)
+	AMDGenoa  CpuFamily = "AMD Genoa"
+	AMDMilan  CpuFamily = "AMD Milan"
+	AMDRome   CpuFamily = "AMD Rome"
+	AMDNaples CpuFamily = "AMD Naples"
+	AMDEPYC   CpuFamily = "AMD EPYC"
+
+	// AWS Graviton (ARM)
+	AWSGraviton3 CpuFamily = "AWS Graviton3"
+	AWSGraviton2 CpuFamily = "AWS Graviton2"
```
sitole: If there is preparation, maybe we should return to platform architecture too.

packages/api/internal/edge/cluster_instances.go
```diff
@@ -68,6 +70,13 @@ func (n *ClusterInstance) GetStatus() infogrpc.ServiceInfoStatus {
 	return n.status
 }
 
+func (n *ClusterInstance) GetMachineInfo() *infogrpc.MachineInfo {
```
sitole: Is there a reason for the mutex lock here? I think we can remove it from the status getter too.

packages/api/internal/edge/cluster_instances.go
```diff
@@ -144,8 +153,9 @@ func (d clusterSynchronizationStore) PoolInsert(_ context.Context, item api.Clus
 		ServiceVersionCommit: item.ServiceVersionCommit,
 
 		// initial values before first sync
-		status: infogrpc.ServiceInfoStatus_Unhealthy,
-		roles:  make([]infogrpc.ServiceInfoRole, 0),
+		status:      infogrpc.ServiceInfoStatus_Unhealthy,
+		roles:       make([]infogrpc.ServiceInfoRole, 0),
+		machineInfo: nil,
```
sitole: Isn't the empty struct option here?

packages/orchestrator/internal/service/machineinfo/main.go
```diff
@@ -0,0 +1,31 @@
...
+)
+
+type MachineInfo struct {
+	Family string
+	Model  string
+	Arch   string
+}
+
+func Detect() (MachineInfo, error) {
+	info, err := cpu.Info()
```
sitole, line 19: Will this work on both AWS and GCP instances? At least for AWS they have Graviton processors that are not explicitly mentioned here https://github.com/shirou/gopsutil/blob/master/cpu/cpu_linux.go

sitole (changes requested):

sitole:
Just few nits

packages/api/internal/edge/cluster_instances.go
jakubno: What would be the reason to pass empty struct here?

sitole: You are essentially doing the same with gRPC, where missing values are treated as empty strings. Removing the pointer here eliminates the need to handle a nil state in all places where machine info is used.

packages/orchestrator/internal/service/machineinfo/main.go
jakubno: Model was corresponding to Graviton and we now have ModelName which should contain that information

packages/orchestrator/internal/service/machineinfo/main.go
```diff
@@ -0,0 +1,39 @@
...
+	if err != nil {
+		return MachineInfo{}, fmt.Errorf("failed to get CPU info: %w", err)
+	}
+
+	if len(info) > 0 {
+		if info[0].Family == "" || info[0].Model == "" {
+			return MachineInfo{}, fmt.Errorf("unable to detect CPU platform from CPU info: %+v", info[0])
+		}
```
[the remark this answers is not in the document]
jakubno: We'll fix once relevant

PR #1514, opened 2025-11-20
Remove context from the constructor of build.DiffStore and template.Cache

djeebus:
> [!NOTE]
> Removes context from `template.NewCache` and `build.NewDiffStore` constructors and introduces explicit `Start`/`Stop` lifecycle methods, updating internal logic and all call sites/tests accordingly.
> 
> - **Core refactor**:
>   - `internal/sandbox/template/cache.go`:
>     - Change `NewCache(config, flags, persistence, metrics)` (remove `ctx`).
>     - Add `Cache.Start(ctx)` and `Cache.Stop()`; start underlying `ttlcache` and `buildStore` here.
>     - Eviction callback now receives `ctx`.
>   - `internal/sandbox/build/cache.go` (`DiffStore`):
>     - Change `NewDiffStore(config, flags, cachePath, ttl, delay)` (remove `ctx`).
>     - Replace `close chan` with `cancel func`; store `config` and `flags`.
>     - Add `Start(ctx)` to run `cache.Start()` and disk-space eviction; `Close()` uses `cancel()`.
>     - Remove `s.close` branch in eviction loop.
> - **Call site updates**:
>   - Start/stop lifecycle wired in `packages/orchestrator/main.go`, `cmd/build-template/main.go`, `benchmark_test.go`, and NBD test utils (call `Start(ctx)` and `Stop/Close`).
> - **Tests**:
>   - `internal/sandbox/build/cache_test.go`: updated to new constructors; call `store.Start(t.Context())`; cleanup via `store.Close()`; race tests adjusted.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 53b1c86324d556ef8cdcd18c3ea869e5bd7d449d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/orchestrator/main.go
+++ b/packages/orchestrator/main.go
@@ -288,6 +288,7 @@ func run(config cfg.Config) (success bool) {
 	templateCache.Start(ctx)
 	closers = append(closers, closer{"template cache", func(context.Context) error {
 		templateCache.Stop()
+
 		return nil
 	}})
 


packages/orchestrator/internal/sandbox/build/cache_test.go
```diff
@@ -285,7 +269,6 @@ func TestDiffStoreOldestFromCache(t *testing.T) {
 	ttl := 60 * time.Second
 	delay := 4 * time.Second
 	store, err := NewDiffStore(
-		t.Context(),
 		c,
```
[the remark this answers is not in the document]
djeebus: This is intentional - it's not necessary to have the background processes running for this test, as it causes race conditions

packages/orchestrator/internal/sandbox/build/cache.go
```diff
@@ -57,7 +58,9 @@ func NewDiffStore(
 	ds := &DiffStore{
 		cachePath: cachePath,
 		cache:     cache,
-		close:     make(chan struct{}),
```
dobrac, line 60: I would prefer to keep using either the channel or switch to the ErrorOnce instead

packages/orchestrator/internal/sandbox/nbd/testutils/template_rootfs.go
```diff
@@ -93,6 +92,8 @@ func TemplateRootfs(ctx context.Context, buildID string) (*BuildDevice, *Cleaner
 		return nil, &cleaner, fmt.Errorf("failed to create diff store: %w", err)
 	}
 
+	store.Start(ctx)
```
dobrac, line 95: there should be probably a cleanup, or not?

packages/orchestrator/internal/sandbox/template/cache.go
```diff
@@ -103,6 +99,17 @@ func NewCache(
 	}, nil
 }
 
+func (c *Cache) Start(ctx context.Context) {
+	c.buildStore.Start(ctx)
+
+	go c.cache.Start()
+}
+
+func (c *Cache) Shutdown() {
```
dobrac: should we call it Close instead because of the buildStore.Close call?

packages/orchestrator/main.go
```diff
@@ -281,10 +281,12 @@ func run(config cfg.Config) (success bool) {
 		zap.L().Fatal("failed to create metrics provider", zap.Error(err))
 	}
 
-	templateCache, err := template.NewCache(ctx, config, featureFlags, persistence, blockMetrics)
+	templateCache, err := template.NewCache(config, featureFlags, persistence, blockMetrics)
 	if err != nil {
 		zap.L().Fatal("failed to create template cache", zap.Error(err))
 	}
+	templateCache.Start(ctx)
+	defer templateCache.Shutdown()
```
dobrac: this should be probably in closers instead

packages/orchestrator/internal/sandbox/template/cache.go
djeebus: What about `Stop`, since that's the opposite of `Start`?

dobrac: also fine I think

packages/orchestrator/internal/sandbox/template/cache.go
```diff
@@ -63,7 +62,7 @@ func NewCache(
 		ttlcache.WithTTL[string, Template](templateExpiration),
 	)
 
-	cache.OnEviction(func(_ context.Context, _ ttlcache.EvictionReason, item *ttlcache.Item[string, Template]) {
+	cache.OnEviction(func(ctx context.Context, _ ttlcache.EvictionReason, item *ttlcache.Item[string, Template]) {
```
dobrac, line 65: not fully confident about this change. Maybe we would need then to call the template.Close(ctx) with context.WithoutCancel

packages/orchestrator/internal/sandbox/nbd/testutils/template_rootfs.go
djeebus: It's on line 104, it was already there.

dobrac: ups, missed that one, thx

packages/orchestrator/internal/sandbox/template/cache.go
djeebus: It looks like, when this function is called, it's called with a context created when this eviction function was first registered. Cancellation only occurs when the return value of this function is called, which it never is.

```golang
func (c *Cache[K, V]) OnEviction(fn func(context.Context, EvictionReason, *Item[K, V])) func() {
	var (
		wg          sync.WaitGroup
		ctx, cancel = context.WithCancel(context.Background())
	)

	c.events.eviction.mu.Lock()
	id := c.events.eviction.nextID
	c.events.eviction.fns[id] = func(r EvictionReason, item *Item[K, V]) {
		wg.Add(1)
		go func() {
			fn(ctx, r, item)
			wg.Done()
		}()
	}
	c.events.eviction.nextID++
	c.events.eviction.mu.Unlock()

	return func() {
		cancel()

		c.events.eviction.mu.Lock()
		delete(c.events.eviction.fns, id)
		c.events.eviction.mu.Unlock()

		wg.Wait()
	}
}```

dobrac: oh, this is nice find! thx!

PR #1524, opened 2025-11-20
feat: add template versioning using tags

dobrac:
This PR introduces the public API for managing **template versions** via **Tags**. It allows users to build templates with multiple aliases/tags and manage those tags post-build. It also updates the `Sandbox.create` method to support the `template:tag` identifier syntax.

#### 1. Sandbox Initialization

Updated `Sandbox.create` to accept a scoped string identifier.

* **Format:** `template-name:tag-name`
* **Default:** If no tag is provided, the backend defaults to `latest`.

#### 2. Template Building

Implemented a clean separation between the template alias and its associated tags during the build process.

* Supports passing a single string or an array/list of tags.
* Ensures that a specific build artifact can be referenced by multiple identifiers immediately.

#### 3. Template Tagging

Added a utility to assign tags to existing builds.

* **Syntax:** `Template.tag("template:tag", tags)`

---

#### JavaScript/Typescript

```ts
// 1. Create sandbox from tag
const sbx = await Sandbox.create("web-server:v1.0.2");

// 2. Build with multiple tags
await Template.build(tmpl, ["web-server:v1.1.0", "web-server:stable"]);

// 3. Update tags for existing build
await Template.assignTag("web-server:v1.1.0", "web-server:production");
// or
await Template.assignTag("web-server:v1.1.0", ["web-server:production"]);

```

#### Python

```python
# 1. Create sandbox from tag
sbx = Sandbox.create("web-server:v1.0.2")

# 2. Build with multiple tags
Template.build(
    config, 
    names=["web-server:v1.1.0", "web-server:stable"]
)

# 3. Update tags for existing build
Template.assign_tag("web-server:v1.1.0", "web-server:production")
# or
Template.assign_tag("web-server:v1.1.0", ["web-server:production"])

```

---

> [!NOTE]
> Introduces first-class template versioning via tags and `template:tag` identifiers across the API surface.
> 
> - **New endpoints:** `POST /templates/tags` to assign tags and `DELETE /templates/tags/{name}` to remove them; OpenAPI, server routes, and generated clients updated
> - **Build request changes:** `TemplateBuildRequestV3` now supports `names` (alias:tag list); `alias` deprecated; build flow propagates alias and tags
> - **Tag-aware usage:** Handlers (sandbox create, template delete/update, build v2/v3, deprecated paths) now parse `template:tag`; health flag switched to atomic
> - **Caching:** Template cache keys include `templateID:tag`; add targeted and all-tags invalidation
> - **Database:** Migration adds `env_build_assignments` (env↔build↔tag), updates queries to join via assignments; snapshot upsert creates assignment; new CRUD for tag assignments
> - **Tests:** Integration and DB tests for tagging, sandbox creation with tags, and snapshot behavior; specs and codegen updated accordingly
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 143316a6152d64f883750faf5d5c2a7fdfc58866. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api/internal/handlers/template_tags.go
+++ b/packages/api/internal/handlers/template_tags.go
@@ -169,7 +169,7 @@ func (a *APIStore) PostTemplatesTags(c *gin.Context) {
 	a.posthog.CreateAnalyticsTeamEvent(ctx, team.ID.String(), "assigned template tag",
 		properties.
 			Set("environment", template.ID).
-			Set("tags", maps.Keys(tags)),
+			Set("tags", slices.Collect(maps.Keys(tags))),
 	)
 
 	logger.L().Info(ctx, "Assigned template tag",


packages/db/migrations/20251120180603_allow_m_n_builds_with_tags.sql
```diff
@@ -0,0 +1,89 @@
+-- +goose Up
+-- +goose StatementBegin
+-- 1. Create the new env_build_assignments table
+CREATE TABLE env_build_assignments (
+    id SERIAL PRIMARY KEY,
```
sitole: Do we want to use PostgreSQL's serial as a primary key? I feel like there are a few things speaking against using serial now (table data consistency, possible permissions issues, managing sequence restarts, and data replication issues).

dobrac: changed to UUIDs

packages/api/internal/template/register_build.go
```diff
@@ -301,6 +295,44 @@ func RegisterBuild(
...
 
+	// Add default tag if no tags are present
+	if len(data.Tags) == 0 {
+		data.Tags = []string{id.DefaultTag}
+	} else {
+		// TODO: Remove this once the migration is deployed
+		err = client.DeleteTriggerTemplateBuildAssignment(ctx, queries.DeleteTriggerTemplateBuildAssignmentParams{
+			TemplateID: data.TemplateID,
+			BuildID:    buildID,
+			Tag:        id.DefaultTag,
```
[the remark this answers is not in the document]
dobrac: this is wanted, no tag isn't latest

spec/openapi.yml
```diff
@@ -877,12 +890,15 @@ components:
           description: Whether the step should be forced to run regardless of the cache
 
     TemplateBuildRequestV3:
-      required:
-        - alias
       properties:
         alias:
```
sitole, line 888: Isn't an alias duplicated now that we have names in use? If yes, then it would make sense to mark it as deprecated, prefer the names, and fail the request when both an alias and names are provided.

spec/openapi.yml
```diff
@@ -2428,6 +2470,61 @@ paths:
         "500":
           $ref: "#/components/responses/500"
 
+  /templates/tags/{name}:
```
sitole: Shouldn't this be `/templates/{name}/tags`? It makes sense to me that you are accessing the template `{name}` and doing something with its tags.

The current path with `/templates/tags/{name}` feels more like I'm already operating on a tag with `{name}`.

sitole: For delete request `DELETE /templates/tags/{name}` makes sense.

packages/db/migrations/20251218160000_allow_m_n_builds_with_tags.sql
```diff
@@ -0,0 +1,128 @@
...
+    RETURN p_value::uuid;
+EXCEPTION WHEN invalid_text_representation THEN
+    RETURN NULL;
+END;
+$$ LANGUAGE plpgsql IMMUTABLE;
+-- +goose StatementEnd
+
+-- +goose StatementBegin
+-- 1. Create the new env_build_assignments table
+CREATE TABLE env_build_assignments (
```
sitole: What about `env_build_tag_assignments`? Feels more intuitive to me.

packages/db/migrations/20251218160000_allow_m_n_builds_with_tags.sql
```diff
@@ -0,0 +1,128 @@
...
+    FOR EACH ROW EXECUTE FUNCTION validate_assignment_source_takeover();
+-- +goose StatementEnd
+
+-- +goose StatementBegin
+-- 3. Legacy Sync Trigger function
+CREATE OR REPLACE FUNCTION sync_env_build_assignment()
+RETURNS TRIGGER AS $$
+BEGIN
+    IF NEW.env_id IS NOT NULL THEN
+        IF TG_OP = 'INSERT' OR (TG_OP = 'UPDATE' AND (OLD.env_id IS NULL OR OLD.env_id != NEW.env_id)) THEN
```
sitole: Why do we need to execute this during UPDATE too? When does someone manually change the template/build IDs?

How can OLD.env_id be NULL during UPDATE?

packages/db/migrations/20251218160000_allow_m_n_builds_with_tags.sql
```diff
@@ -0,0 +1,128 @@
...
+-- 4. Create trigger to automatically sync changes
+CREATE TRIGGER trigger_sync_env_build_assignment
+    AFTER INSERT OR UPDATE ON env_builds
+    FOR EACH ROW
+    EXECUTE FUNCTION sync_env_build_assignment();
+-- +goose StatementEnd
+
+-- +goose StatementBegin
+-- 5. Migrate existing data
+INSERT INTO env_build_assignments (env_id, build_id, tag, source, created_at)
```
sitole: With the amount of data we have, cannot this timeout/exhaust instance?

packages/db/queries/templates/create_template_build_assignment.sql
```diff
@@ -0,0 +1,5 @@
+-- name: CreateTemplateBuildAssignment :exec
```
sitole, line 1: Shouldn't this and the following be CreateTemplateBuildTag, DeleteTemplateBuildTag, and the following?

packages/db/queries/templates/delete_template_build_assignment.sql
```diff
@@ -0,0 +1,10 @@
+-- name: DeleteTemplateBuildAssignment :exec
```
sitole: Is this name correct? This efficiently deletes the tag from all builds of a specific template. I think the name of the query is not transparent in this case and should be named something like `DeleteTemplateTag`

packages/api/internal/template/register_build.go
```diff
@@ -301,6 +295,44 @@ func RegisterBuild(
 		telemetry.ReportEvent(ctx, "inserted alias", attribute.String("env.alias", alias))
 	}
 
+	// Add default tag if no tags are present
+	if len(data.Tags) == 0 {
+		data.Tags = []string{id.DefaultTag}
+	} else {
+		// TODO: Remove this once the migration is deployed
```
sitole: Lets add linear issue link here

packages/api/internal/handlers/template_request_build_v3.go
```diff
@@ -48,17 +51,54 @@ func requestTemplateBuild(ctx context.Context, c *gin.Context, a *APIStore, body
 		return nil
 	}
 
+	names := utils.DerefOrDefault(body.Names, nil)
+	if body.Alias != nil {
+		names = slices.Concat(names, []string{*body.Alias})
+	}
+
+	if len(names) == 0 {
```
sitole, line 68: Why are names required now?

spec/openapi.yml
dobrac: changed to POST /templates/tags and pass the name in body

packages/db/migrations/20251218160000_allow_m_n_builds_with_tags.sql
dobrac: the env_build_assignments is meant instead of the env_id column to support the edge for binding envs with builds in general

packages/db/migrations/20251218160000_allow_m_n_builds_with_tags.sql
dobrac: > How can OLD.env_id be NULL during UPDATE?

Removed

> Why do we need to execute this during UPDATE too

no really, the trigger will be removed after the migration so it's ok to keep it there

packages/api/internal/handlers/template_request_build_v3.go
dobrac: they have been required previously as the alias was required

dobrac:
> Few questions
> 
> 1. Are we okay and aware that when I specify template alias `bob` and `alice`, and then the alias for template `bob` called `foo I will actually be able to retrieve it via `bob:foo`or`alice:foo`, is this expected?
> 2. When I use the flow you designed in `TestTemplateTagAssignFromSourceTag` - I think at some point I would love to reference not just template tag but specific template builds. This way i will be able to revert my production to any build in the past. Now I can only revert to versions that are somehow tagged, is that right?
> 3. In openspec for sandbox creation we have `TemplateID` that can actually be template id, alias or now id/alias with tag. What about updating the spec description or introducing a new field and deprecating the old one?

1. yes

2. you can reference them already, the select query already searches by the buildID

3. I agree with renaming, let's do that in another PR as it will change a lot of code

packages/db/queries/templates/create_template_build_assignment.sql
dobrac: I think this one is ok

packages/db/migrations/20251218160000_allow_m_n_builds_with_tags.sql
```diff
@@ -97,28 +100,83 @@ $$ LANGUAGE plpgsql;
...
+        SELECT id INTO current_max_id
+        FROM (
+            SELECT id FROM env_builds 
+            WHERE env_id IS NOT NULL 
+            AND id > last_id 
+            ORDER BY id 
+            LIMIT batch_size
+        ) sub
+        ORDER BY id DESC
+        LIMIT 1;
```
sitole: Will this work when our IDs are not sorted?
If you execute this, it will return the IDs of A1...A30, but during execution, the UUID at A10 will be stored, we will miss that, right?

dobrac: fair, I'll sort it by the creation date

sitole: Yeah, hopefully this will work. We can also add a temporary column marking the row as migrated.

dobrac: It should be idempotent, so it's possible to run it multiple times

packages/api/internal/handlers/template_tags.go
```diff
@@ -0,0 +1,275 @@
...
+			BuildID:    buildID,
+			Tag:        tag,
+		})
+		if err != nil {
+			telemetry.ReportCriticalError(ctx, "error when creating tag assignment", err)
+			a.sendAPIStoreError(c, http.StatusInternalServerError, "Error creating tag assignment")
+
+			return
+		}
+	}
```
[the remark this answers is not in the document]
dobrac: desired

PR #1525, opened 2025-11-20
Add the 'early-return' linter

djeebus:
Also tweak some integration tests to try and make them less flaky. Success?

Demonstration of why the test changes are necessary: https://go.dev/play/p/f2B4sJgHvmr

---

> [!NOTE]
> Enable the 'early-return' linter and refactor Redis init, NBD dispatch, and file-lock logic to use early returns and simpler control flow.
> 
> - **Lint/config**:
>   - Enable revive `early-return` rule in `.golangci.yml`.
> - **Client Proxy (`packages/client-proxy/main.go`)**:
>   - Simplify Redis init error handling: only error on non-`ErrRedisDisabled`; otherwise warn and fallback to in-memory catalog.
> - **Orchestrator NBD (`packages/orchestrator/internal/sandbox/nbd/dispatch.go`)**:
>   - Flatten header parsing/dispatch loop with early breaks; refactor command handling.
>   - Change shutdown checks to early-return and adjust `pendingResponses` accounting.
> - **Storage lock (`packages/shared/pkg/storage/lock/file_lock.go`)**:
>   - Early-return when lock is within TTL; remove stale lock files before acquiring.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit fbe41a8da62ad40069747802b214ee5dea887369. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -113,7 +113,6 @@ linters:
         - { disabled: true, name: context-as-argument }
         - { disabled: true, name: cyclomatic }
         - { disabled: true, name: deep-exit }
-        - { disabled: true, name: early-return }
         - { disabled: true, name: enforce-switch-style }
         - { disabled: true, name: exported }
         - { disabled: true, name: flag-parameter } # todo: enable this
--- a/packages/client-proxy/main.go
+++ b/packages/client-proxy/main.go
@@ -136,14 +136,14 @@ func run() int {
 		}()
 		catalog = e2bcatalog.NewRedisSandboxesCatalog(redisClient)
 	} else {
-		if errors.Is(err, factories.ErrRedisDisabled) {
-			l.Warn(ctx, "Redis environment variable is not set, will fallback to in-memory sandboxes catalog that works only with one instance setup")
-			catalog = e2bcatalog.NewMemorySandboxesCatalog()
-		} else {
+		if !errors.Is(err, factories.ErrRedisDisabled) {
 			l.Error(ctx, "Failed to create redis client", zap.Error(err))
 
 			return 1
 		}
+
+		l.Warn(ctx, "Redis environment variable is not set, will fallback to in-memory sandboxes catalog that works only with one instance setup")
+		catalog = e2bcatalog.NewMemorySandboxesCatalog()
 	}
 
 	orchestrators := e2borchestrators.NewOrchestratorsPool(ctx, l, tel.TracerProvider, tel.MeterProvider, orchestratorsSD)
--- a/packages/orchestrator/internal/sandbox/nbd/dispatch.go
+++ b/packages/orchestrator/internal/sandbox/nbd/dispatch.go
@@ -149,79 +149,79 @@ func (d *Dispatch) Handle(ctx context.Context) error {
 			}
 
 			// Make sure we have a complete header
-			if wp-rp >= 28 {
-				// We can read the neader...
-
-				header := buffer[rp : rp+28]
-				request.Magic = binary.BigEndian.Uint32(header)
-				request.Type = binary.BigEndian.Uint32(header[4:8])
-				request.Handle = binary.BigEndian.Uint64(header[8:16])
-				request.From = binary.BigEndian.Uint64(header[16:24])
-				request.Length = binary.BigEndian.Uint32(header[24:28])
-
-				if request.Magic != NBDRequestMagic {
-					return fmt.Errorf("received invalid MAGIC")
-				}
-
-				switch request.Type {
-				case NBDCmdDisconnect:
-					return nil // All done
-				case NBDCmdFlush:
-					return fmt.Errorf("not supported: Flush")
-				case NBDCmdRead:
-					rp += 28
-					err := d.cmdRead(ctx, request.Handle, request.From, request.Length)
-					if err != nil {
-						return err
-					}
-				case NBDCmdWrite:
-					rp += 28
-
-					if request.Length > dispatchMaxWriteBufferSize {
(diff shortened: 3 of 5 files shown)


tests/integration/internal/tests/api/sandboxes/sandbox_list_test.go
```diff
@@ -167,23 +166,15 @@ func TestSandboxListPausing(t *testing.T) {
...
-	wg.Go(func() error {
-		pauseSandboxResponse, err := c.PostSandboxesSandboxIDPauseWithResponse(t.Context(), sandboxID, setup.WithAPIKey())
-		if err != nil {
-			return err
-		}
-
-		if pauseSandboxResponse.StatusCode() != http.StatusNoContent {
-			return fmt.Errorf("expected status code %d, got %d", http.StatusNoContent, pauseSandboxResponse.StatusCode())
-		}
+	pauseSandboxResponse, err := c.PostSandboxesSandboxIDPauseWithResponse(t.Context(), sandboxID, setup.WithAPIKey())
```
jakubno: Now the test is doing something completely different.

Before, we were testing if the sandbox is still present in the sandbox list **during** the pausing; now you test if the sandbox is in the list **after** the pause.

jakubno (dismissed):

tests/integration/internal/tests/api/sandboxes/sandbox_list_test.go
jakubno: I fixed it in #1529

jakubno:
Feel free to cancel my review, I can't do it when there's merge conflict

tests/integration/internal/tests/api/sandboxes/sandbox_list_test.go
djeebus: Ah, good point! I reverted that change to go back to the wait group.

There were a few other issues with the test. 
1. Sometimes it takes [longer than 10 seconds](https://github.com/e2b-dev/infra/actions/runs/19554049580/job/55992349918?pr=1525#step:7:649) to pause a sandbox, so that timeout needed to be increased
2. Using `require.*` inside of `require.Eventually` will cause the test to fail, even if the function eventually passes. Swapping to `EventuallyWithT` fixes this. See the [go playground example](https://go.dev/play/p/K_hc1BhkKf9).

tests/integration/internal/tests/api/sandboxes/sandbox_auto_pause_test.go
```diff
@@ -72,14 +72,20 @@ func TestSandboxAutoPauseResumePersisted(t *testing.T) {
...
-	require.Eventually(t, func() bool {
-		res, err := c.GetSandboxesSandboxIDWithResponse(t.Context(), sbxId, setup.WithAPIKey())
-		require.NoError(t, err)
-		require.Equal(t, http.StatusOK, res.StatusCode())
-		require.NotNil(t, res.JSON200)
-
-		return res.JSON200.State == api.Paused
-	}, 10*time.Second, 10*time.Millisecond, "Sandbox is not paused")
+	ctx := t.Context()
+	require.EventuallyWithT(t, func(t *assert.CollectT) {
```
jakubno: this changes the test. It shouldn't ever return an error or any response other than 200. If it does, it should fail

tests/integration/internal/tests/api/sandboxes/sandbox_list_test.go
```diff
@@ -181,32 +183,35 @@ func TestSandboxListPausing(t *testing.T) {
 		return nil
 	})
 
-	require.Eventually(t, func() bool {
+	require.EventuallyWithT(t, func(t *assert.CollectT) {
```
jakubno: same here
this changes the test. It shouldn't ever return an error or any response other than 200. If it does, it should fail

tests/integration/internal/tests/envd/auth_test.go
```diff
@@ -140,33 +141,31 @@ func TestAccessAuthorizedPathWithResumedSandboxWithValidAccessToken(t *testing.T
...
-	fileResponse, err := envdClient.HTTPClient.GetFilesWithResponse(
-		ctx,
-		&envdapi.GetFilesParams{Path: &filePath, Username: sharedUtils.ToPtr("user")},
-		setup.WithSandbox(sbx.JSON201.SandboxID),
-		setup.WithEnvdAccessToken(*sbxMeta.EnvdAccessToken),
-	)
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.EventuallyWithT(t, func(t *assert.CollectT) {
```
jakubno: why eventually here?

jakubno (dismissed):

tests/integration/internal/tests/api/sandboxes/sandbox_auto_pause_test.go
djeebus: The test still does that. 

`require.EventuallyWithT` calls the function and collects all the asserts. If any assert was collected, it waits the `tick` duration and tries again, until it hits the `waitFor` duration. If the function still hasn't been called without failed asserts, then the `EventuallyWithT` fails and reports all the asserts collected during the final function call.

The swap from `require` to `assert` needs some explanation of testify's internals. The way that `require` works is by recording the assertion failure and then calling `t.FailNow()` which called `runtime.Goexit`; this exits the `EventulalyWithT` and prevents it from looping again. Check the demonstration above. If it still doesn't make sense, let's talk about it.

tests/integration/internal/tests/envd/auth_test.go
djeebus: I ran into a race condition here, this resolved it.

PR #1526, opened 2025-11-21
fix: don't require node id before template build starts

dobrac:
> [!NOTE]
> Defers selecting/setting the builder node until build start and makes `env_builds.cluster_node_id` nullable, updating handlers, cache, template manager, models, queries, and tests.
> 
> - **DB/Schema**:
>   - Make `env_builds.cluster_node_id` nullable (migration) and update Ent schema/models accordingly.
>   - Adjust SQL/queries (`UpdateTemplateBuild`) to set `cluster_node_id`; expose as `*string` in generated types.
>   - Snapshot `origin_node_id` switched to `*string` in queries and tests.
> - **API/Handlers**:
>   - Remove builder node from `template.RegisterBuildData`; node is chosen at build start (v1/v2) and persisted via `UpdateTemplateBuild`.
>   - Handle nil `ClusterNodeID` when canceling/deleting builds and when killing sandboxes.
>   - Build status/logs: guard `NodeID` being nil before fetching logs; early-return for `waiting`.
>   - Return `503 ServiceUnavailable` for unavailable build client in relevant endpoints.
> - **Template Manager**:
>   - `CreateTemplate` triggers background `BuildStatusSync` with optional node ID; `BuildStatusSync` now accepts `*string` and errors if missing after start.
>   - Template build cache stores `NodeID` as `*string` and propagates through `TemplateBuildInfo`.
> - **Orchestrator**:
>   - Use pointer for `OriginNodeID` when pausing/snapshotting sandboxes.
> - **Tests**:
>   - Update snapshot tests to use pointer `OriginNodeID` and validate new behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e5ef954f8bbd3bd9ecfc2db37a52f727eb8070c5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api/internal/handlers/deprecated_template_start_build.go
+++ b/packages/api/internal/handlers/deprecated_template_start_build.go
@@ -153,7 +153,7 @@ func (a *APIStore) PostTemplatesTemplateIDBuildsBuildID(c *gin.Context, template
 
 	builderNodeID, err := a.templateManager.GetAvailableBuildClient(ctx, apiutils.WithClusterFallback(team.ClusterID))
 	if err != nil {
-		a.sendAPIStoreError(c, http.StatusInternalServerError, "Error when getting available build client")
+		a.sendAPIStoreError(c, http.StatusServiceUnavailable, "Error when getting available build client")
 		telemetry.ReportCriticalError(ctx, "error when getting available build client", err, telemetry.WithTemplateID(templateID))
 
 		return
--- a/packages/api/internal/handlers/template_layer_files_upload.go
+++ b/packages/api/internal/handlers/template_layer_files_upload.go
@@ -44,7 +44,7 @@ func (a *APIStore) GetTemplatesTemplateIDFilesHash(c *gin.Context, templateID ap
 	nodeID, err := a.templateManager.GetAvailableBuildClient(ctx, utils.WithClusterFallback(templateDB.ClusterID))
 	if err != nil {
 		telemetry.ReportCriticalError(ctx, "error when getting available build client", err, telemetry.WithTemplateID(templateID))
-		a.sendAPIStoreError(c, http.StatusInternalServerError, "Error when getting available build client")
+		a.sendAPIStoreError(c, http.StatusServiceUnavailable, "Error when getting available build client")
 
 		return
 	}
--- a/packages/api/internal/handlers/template_start_build_v2.go
+++ b/packages/api/internal/handlers/template_start_build_v2.go
@@ -130,7 +130,7 @@ func (a *APIStore) PostV2TemplatesTemplateIDBuildsBuildID(c *gin.Context, templa
 
 	builderNodeID, err := a.templateManager.GetAvailableBuildClient(ctx, apiutils.WithClusterFallback(team.ClusterID))
 	if err != nil {
-		a.sendAPIStoreError(c, http.StatusInternalServerError, "Error when getting available build client")
+		a.sendAPIStoreError(c, http.StatusServiceUnavailable, "Error when getting available build client")
 		telemetry.ReportCriticalError(ctx, "error when getting available build client", err, telemetry.WithTemplateID(templateID))
 
 		return


packages/api/internal/handlers/deprecated_template_start_build.go
```diff
@@ -36,18 +36,27 @@ func (a *APIStore) CheckAndCancelConcurrentBuilds(ctx context.Context, templateI
 	}
 
 	// make sure there is no other build in progress for the same template
-	if len(concurrentlyRunningBuilds) > 0 {
-		buildIDs := utils.Map(concurrentlyRunningBuilds, func(b queries.EnvBuild) templatemanager.DeleteBuild {
+	if len(concurrentBuilds) > 0 {
+		concurrentRunningBuilds := utils.Filter(concurrentBuilds, func(b queries.EnvBuild) bool {
+			return b.Status == envbuild.StatusBuilding.String()
```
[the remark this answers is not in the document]
dobrac: this comment doesn't make sense at all

packages/api/internal/template-manager/template_status.go
```diff
@@ -56,6 +56,10 @@ func (tm *TemplateManager) BuildStatusSync(ctx context.Context, buildID uuid.UUI
 		return nil
 	}
 
+	if nodeID == nil {
```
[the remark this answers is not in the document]
dobrac: not true

packages/db/migrations/20251121101953_build_node_id_nullable.sql
```diff
@@ -0,0 +1,9 @@
+-- +goose Up
+-- +goose StatementBegin
+ALTER TABLE "public"."env_builds" ALTER COLUMN cluster_node_id DROP NOT NULL;
```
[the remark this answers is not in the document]
dobrac: this is not true

packages/api/internal/handlers/template_build_status.go
```diff
@@ -113,18 +105,29 @@ func (a *APIStore) GetTemplatesTemplateIDBuildsBuildIDStatus(c *gin.Context, tem
...
 	}
 
-	for _, entry := range cli.GetLogs(ctx, templateID, buildID, offset, apiToLogLevel(params.Level)) {
-		if legacyLogs {
-			lgs = append(lgs, fmt.Sprintf("[%s] %s\n", entry.Timestamp.Format(time.RFC3339), entry.Message))
+	nodeID := buildInfo.NodeID
+	if nodeID != nil {
+		cli, err := a.templateManager.GetClusterBuildClient(utils.WithClusterFallback(team.ClusterID), *nodeID)
+		if err != nil {
+			telemetry.ReportError(ctx, "error when getting build client", err, telemetry.WithTemplateID(templateID), telemetry.WithBuildID(buildID))
```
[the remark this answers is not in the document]
dobrac: the cache should be invalidated on build start

packages/api/internal/handlers/deprecated_template_start_build.go
```diff
@@ -36,18 +36,27 @@ func (a *APIStore) CheckAndCancelConcurrentBuilds(ctx context.Context, templateI
...
+		})
+		buildIDs := utils.Map(concurrentRunningBuilds, func(b queries.EnvBuild) templatemanager.DeleteBuild {
+			clusterNodeID := "non-existent"
+			if b.ClusterNodeID != nil {
+				clusterNodeID = *b.ClusterNodeID
+			}
```
sitole: Can we use something similar as `const UnknownNomadNodeShortID = "unknown"` so we don't need to use a pointer here?

packages/db/migrations/20251121101953_build_node_id_nullable.sql
```diff
@@ -0,0 +1,9 @@
...
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
+ALTER TABLE "public"."env_builds" ALTER COLUMN cluster_node_id SET NOT NULL;
```
sitole, line 9: Shouldn't we update fields that are null?

packages/api/internal/handlers/deprecated_template_start_build.go
sitole: If there is unknown node id we should skip it

dobrac: refactored

dobrac: refactored

packages/api/internal/handlers/deprecated_template_start_build.go
```diff
@@ -152,7 +153,7 @@ func (a *APIStore) PostTemplatesTemplateIDBuildsBuildID(c *gin.Context, template
 
 	builderNodeID, err := a.templateManager.GetAvailableBuildClient(ctx, apiutils.WithClusterFallback(team.ClusterID))
 	if err != nil {
-		a.sendAPIStoreError(c, http.StatusBadRequest, fmt.Sprintf("Error when getting available build client: %s", err))
+		a.sendAPIStoreError(c, http.StatusInternalServerError, "Error when getting available build client")
```
sitole: We talked about this with @jakubno, and it's possible the customer cluster doesn't have available builder nodes, so I'm not sure returning 5xx here is okay, but I agree that a bad request can be confusing too.

dobrac: changed to 503

sitole (approved):

PR #1543, opened 2025-11-25
chore: refactor rootfs direct provider

dobrac:
Refactor the provisioning to not use symlink for rootfs. Fixes also improper file descriptor cleanup.

---

> [!NOTE]
> Reworks rootfs direct provider to use mmap and a shared DiffMetadataBuilder for diff export, removes symlink provisioning, and tightens resource/FD handling across components.
> 
> - **Rootfs (Direct provider)**:
>   - Replace `block.Cache` usage with mmap-backed file (`mmap-go`) and explicit flush/unmap lifecycle.
>   - Implement `exportToDiff` reading blocks from disk and building diffs via `header.NewDiffMetadataBuilder`.
>   - Add close coordination with `closed` flag; ensure mmap unmap on export/close.
> - **Shared Header**:
>   - Introduce `DiffMetadataBuilder` with `Process` and `Build` to unify dirty/empty block detection and writing.
> - **Block**:
>   - `Cache.ExportToDiff` now uses `DiffMetadataBuilder`; remove manual bitset handling (and `MarkAllAsDirty`).
>   - `Local` improves error handling on init; simplify `Close` (no removal side effects).
> - **Template Build**:
>   - Remove rootfs provisioning symlink; use actual `rootfsPath`.
>   - `phases.Run` now accepts a logger and logs per phase; builder invocation updated.
> - **Sandbox**:
>   - Close memfile immediately after snapshot to release lock before shutdown.
> - **Socket**:
>   - Include `context.Cause` in wait cancellation error.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7072d9b20503809103170fa258a1901fc871fc2b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/orchestrator/internal/sandbox/rootfs/direct.go
+++ b/packages/orchestrator/internal/sandbox/rootfs/direct.go
@@ -27,7 +27,7 @@ type DirectProvider struct {
 	// TODO: Remove when the snapshot flow is improved
 	finishedOperations chan struct{}
 	// TODO: Remove when the snapshot flow is improved
-	exporting atomic.Bool
+	closed atomic.Bool
 
 	mmap *mmap.MMap
 }
@@ -75,7 +75,7 @@ func (o *DirectProvider) ExportDiff(
 	ctx, childSpan := tracer.Start(ctx, "direct-provider-export")
 	defer childSpan.End()
 
-	if !o.exporting.CompareAndSwap(false, true) {
+	if !o.closed.CompareAndSwap(false, true) {
 		return nil, fmt.Errorf("direct provider close is already in progress")
 	}
 
@@ -114,7 +114,7 @@ func (o *DirectProvider) ExportDiff(
 func (o *DirectProvider) Close(ctx context.Context) error {
 	o.finishedOperations <- struct{}{}
 
-	if !o.exporting.CompareAndSwap(false, true) {
+	if !o.closed.CompareAndSwap(false, true) {
 		return nil
 	}
 


packages/orchestrator/internal/sandbox/build/build.go
```diff
@@ -71,6 +71,14 @@ func (b *File) ReadAt(ctx context.Context, p []byte, off int64) (n int, err erro
 		// We will use this to handle base builds that are already diffs.
 		// The passed slice p must start as empty, otherwise we would need to copy the empty values there.
 		if *buildID == uuid.Nil {
+			isEmpty, err := header.IsEmptyBlock(p[n:int64(n)+readLength], int64(b.header.Metadata.BlockSize))
```
dobrac: this might not be needed, it's just check to be sure that the block is empty

packages/orchestrator/internal/sandbox/rootfs/nbd.go
```diff
@@ -48,15 +53,79 @@ func NewNBDProvider(rootfs block.ReadonlyDevice, cachePath string, devicePool *n
...
+			l.Error(ctx, "rootfs block checksum mismatch nbd",
+				zap.Int("block_index", blockIndex),
+				zap.Int64("block_offset", blockOffset),
+				zap.String("block_checksum", hex.EncodeToString(blockChecksum[:])),
+				zap.String("block_checksum_expected", hex.EncodeToString(blockChecksumExpected[:])),
+			)
+			wrongCount++
+			if wrongCount > 10 {
+				return fmt.Errorf("too many block checksum mismatches")
+			}
```
jakubno: this is just a optimization?

packages/orchestrator/internal/sandbox/rootfs/nbd.go
```diff
@@ -48,15 +53,79 @@ func NewNBDProvider(rootfs block.ReadonlyDevice, cachePath string, devicePool *n
...
+		}
+	}
+
+	if !bytes.Equal(checksums.Checksum[:], headerChecksums.Checksum[:]) {
```
jakubno: do you even need to do this if you do block checksums check?

packages/orchestrator/internal/sandbox/rootfs/direct.go
```diff
@@ -99,13 +140,78 @@ func (o *DirectProvider) Close(ctx context.Context) error {
...
+		}
+	}
+
+	m, err := builder.Build()
+	if err != nil {
+		return nil, fmt.Errorf("error building diff metadata: %w", err)
+	}
+
+	return m, nil
+}
```
jakubno: You can do the same as in the Cache, also `Build()` can't return error, so that can be also simplified
```suggestion
	return builder.Build()
}
```

packages/orchestrator/internal/sandbox/rootfs/nbd.go
dobrac: yes

packages/orchestrator/internal/sandbox/rootfs/nbd.go
dobrac: not really because I do also the length check

packages/orchestrator/internal/sandbox/build/build.go
dobrac: removed

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -707,6 +708,12 @@ func (s *Sandbox) Shutdown(ctx context.Context) error {
 		return fmt.Errorf("error creating snapshot: %w", err)
 	}
 
+	// Close the memfile right after the snapshot to release the lock.
+	err = memfile.Close()
+	if err != nil {
+		return fmt.Errorf("error closing memfile: %w", err)
+	}
```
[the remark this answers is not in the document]
dobrac: not an issue, desired

packages/shared/pkg/storage/header/metadata.go
```diff
@@ -106,3 +107,55 @@ func (d *DiffMetadata) ToDiffHeader(
...
+	b.dirty.Set(uint(blockIdx))
+	n, err := out.Write(block)
+	if err != nil {
+		logger.L().Error(ctx, "error writing to out", zap.Error(err))
+
+		return err
+	}
+
+	if int64(n) != b.blockSize {
+		return fmt.Errorf("short write: %d != %d", int64(n), b.blockSize)
```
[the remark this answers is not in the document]
dobrac: this is expected

packages/orchestrator/internal/sandbox/block/local.go
```diff
@@ -69,10 +73,6 @@ func (d *Local) BlockSize() int64 {
 }
 
 func (d *Local) Close() (e error) {
-	defer func() {
-		e = errors.Join(e, os.Remove(d.path))
```
jakubno, line 73: how will the rootfs and memfile get removed now if there's some error e.g. in `base/builder.go`

packages/orchestrator/internal/sandbox/rootfs/direct.go
```diff
@@ -54,15 +71,16 @@ func (o *DirectProvider) ExportDiff(
...
+) (*header.DiffMetadata, error) {
 	ctx, childSpan := tracer.Start(ctx, "direct-provider-export")
 	defer childSpan.End()
 
 	o.exporting.CompareAndSwap(false, true)
 
 	defer func() {
-		if e != nil {
-			e = errors.Join(e, o.cache.Close())
+		err := o.mmap.Unmap()
```
jakubno, line 83: aren't you calling `Unmap()` twice? Once in here and the second in `Close()` as `Close()` has to be called in export to `ExportToDiff()` to finish

packages/orchestrator/internal/sandbox/block/local.go
dobrac: using this: https://github.com/e2b-dev/infra/blob/7e3ebcc8c38661fc2de5d0a66b81d50a0dcfc487/packages/orchestrator/internal/template/build/phases/base/builder.go#L170

basically I wanted to move the control scope up

packages/orchestrator/internal/sandbox/rootfs/direct.go
dobrac: It's really bad code flow - I shouldn't because of this `o.exporting.CompareAndSwap(false, true)`

jakubno: Isn't there some race condition if you would call Pause/Kill at the same time? API shouldn't allow it. Not sure if that's something where we want to move the responsibility to API

dobrac: I think there is, we don't have any locks for the sandbox operations in the orchestrator in general

dobrac: added check also for export so that it can't be called twice

packages/orchestrator/internal/sandbox/rootfs/direct.go
```diff
@@ -95,17 +114,60 @@ func (o *DirectProvider) ExportDiff(
 func (o *DirectProvider) Close(ctx context.Context) error {
 	o.finishedOperations <- struct{}{}
 
-	if !o.exporting.CompareAndSwap(false, true) {
+	if !o.closed.CompareAndSwap(false, true) {
 		return nil
 	}
 
-	return errors.Join(o.sync(ctx), o.cache.Close())
+	return errors.Join(o.sync(ctx), o.mmap.Unmap())
```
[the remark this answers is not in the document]
dobrac: good point, not an issue right now

PR #1553, opened 2025-11-30
Add ability to change FC IO engine

ValentaTomas:
> [!NOTE]
> Firecracker rootfs drive IO engine is now configurable via LaunchDarkly (BuildIoEngine) by wiring the feature-flags client through the orchestrator and adding string-flag support.
> 
> - **Firecracker/orchestrator**:
>   - `apiClient` now holds `featureFlags`; `setRootfsDrive` reads `featureflags.BuildIoEngine` to set `Drive.IoEngine` (with fallback logging).
>   - `newApiClient` and `NewProcess` accept `featureFlags`; `sandbox.Factory` passes it in both Create and Resume paths.
> - **Feature flags**:
>   - Add `StringFlag` type and `Client.StringFlag` API; define `BuildIoEngine` with default `"Async"`.
> - **Template Build**:
>   - Set LaunchDarkly contexts (template, team) at build start.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a0e4077b3fab7fc68b161879f27d13f138a9b730. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/orchestrator/internal/sandbox/fc/client.go
+++ b/packages/orchestrator/internal/sandbox/fc/client.go
@@ -6,28 +6,34 @@ import (
 
 	"github.com/firecracker-microvm/firecracker-go-sdk"
 	"github.com/go-openapi/strfmt"
+	"go.uber.org/zap"
 
 	"github.com/e2b-dev/infra/packages/orchestrator/internal/sandbox/socket"
 	"github.com/e2b-dev/infra/packages/orchestrator/internal/sandbox/template"
 	"github.com/e2b-dev/infra/packages/shared/pkg/fc/client"
 	"github.com/e2b-dev/infra/packages/shared/pkg/fc/client/operations"
 	"github.com/e2b-dev/infra/packages/shared/pkg/fc/models"
+	featureflags "github.com/e2b-dev/infra/packages/shared/pkg/feature-flags"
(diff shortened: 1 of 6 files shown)


packages/orchestrator/internal/sandbox/fc/client.go
```diff
@@ -178,7 +184,15 @@ func (c *apiClient) setBootSource(ctx context.Context, kernelArgs string, kernel
 
 func (c *apiClient) setRootfsDrive(ctx context.Context, rootfsPath string) error {
 	rootfs := "rootfs"
-	ioEngine := "Async"
+
+	ioEngine, err := c.featureFlags.StringFlag(
```
jakubno, line 188: i am thinking if it isn't safer to do it as a boolean flag rather than passing a string from launch darkly. Or we should do some validation. A typo would break the flow. Is it case sensitive?

dobrac: You're right, but I think this is a temporary solution and the LaunchDarkly variations definitions prevent you from using incorrect value (unless the definition itself is incorrect).

So, I thinks it's as of now

PR #1557, opened 2025-12-02
Transition from Docker registry images to Google Artifact Registry tags for image references

jakubno:
- Eliminates the need for a local Docker client during Terraform runs. This improves portability and avoids failures on systems where Docker isn’t available or configured.
- Corrects an issue where Terraform would incorrectly resolve a locally cached :latest image. Terraform now consistently pulls/uses the intended remote image instead of relying on stale local state.

---

> [!NOTE]
> Switch Terraform image references to Google Artifact Registry data sources (dropping Docker provider/resources) and relocate Docker auth from the composite action to the build workflow.
> 
> - **Infrastructure (Terraform/Nomad)**:
>   - Replace Docker provider/resources with `google_artifact_registry_docker_image` data sources in `nomad/images.tf` and update Nomad jobs to use `.self_link` for `api`, `db-migrator`, `docker-reverse-proxy`, `client-proxy`, and `clickhouse-migrator` images.
>   - Remove Docker and Grafana providers/config from `main.tf`; simplify required providers.
> - **CI/CD**:
>   - Remove Docker setup from `./.github/actions/deploy-setup/action.yml`.
>   - Add Docker auth/setup step to `.github/workflows/build-and-upload-job.yml`.
> - **Lockfile**:
>   - Drop `grafana/grafana` and `kreuzwerker/docker` providers from `iac/provider-gcp/.terraform.lock.hcl`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b78d4a5ba85820281a9f70af9f979dc9a2edc611. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/iac/provider-gcp/nomad/main.tf
+++ b/iac/provider-gcp/nomad/main.tf
@@ -48,36 +48,6 @@ data "google_secret_manager_secret_version" "redis_tls_ca_base64" {
 }
 
 
-data "google_artifact_registry_docker_image" "api_image" {
-  location      = var.gcp_region
-  repository_id = var.orchestration_repository_name
-  image_name    = "api:latest"
-}
-
-data "google_artifact_registry_docker_image" "db_migrator_image" {
-  location      = var.gcp_region
-  image_name    = "db-migrator:latest"
-  repository_id = var.orchestration_repository_name
-}
-
-data "google_artifact_registry_docker_image" "docker_reverse_proxy_image" {
-  location      = var.gcp_region
-  image_name    = "docker-reverse-proxy:latest"
-  repository_id = var.orchestration_repository_name
-}
-
-data "google_artifact_registry_docker_image" "client_proxy_image" {
-  location      = var.gcp_region
-  image_name    = "client-proxy:latest"
-  repository_id = var.orchestration_repository_name
-}
-
-data "google_artifact_registry_docker_image" "clickhouse_migrator_image" {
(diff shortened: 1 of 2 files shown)


iac/provider-gcp/nomad/main.tf
```diff
@@ -56,24 +48,71 @@ data "google_secret_manager_secret_version" "redis_tls_ca_base64" {
...
+  location      = var.gcp_region
+  package_name  = "docker-reverse-proxy"
+  repository_id = var.orchestration_repository_name
+  tag_name      = "latest"
 }
 
-resource "docker_image" "api_image" {
-  name          = data.docker_registry_image.api_image.name
-  pull_triggers = [data.docker_registry_image.api_image.sha256_digest]
-  platform      = "linux/amd64/v8"
```
dobrac, line 66: is the platform not required anymore? I know we have only the correct images there for now, but not sure if that would be the case always

dobrac (approved):
This is very nice change 🙌

iac/provider-gcp/nomad/main.tf
jakubno: We can't really check without pulling the image. If somehow the image is uploaded with incorrect platform the job just won't start, so it should be fine

PR #1572, opened 2025-12-04
Improved: clean-nfs-cache performance

levb:
- Added global caching of directories and files; not memory-bounded yet.
- Added concurrent handling for finding random candidates, stat-ing and deleting files
- Trim some empty folders (only if empty immediately after scanning)
- Integrated with launchdarkly for feature flag, see https://app.launchdarkly.com/projects/default/flags/clean-nfs-cache-experimental

Results from a run on my dev environment from a snapshot-restored identical filesystem:
Old version: 
- "files": 23962,
- "bytes": 100003299328,
- "files_per_second": 102.68987202964917,
- "bytes_per_second": 428567148.42396384,

New version with (randomly-picked) concurrency `"max_concurrent_stat": 32, "max_concurrent_scan": 16, "max_concurrent_delete": 4`:
- "del_files": 24300,
- "empty_dirs": 6,
- "bytes": 101403889664,
- "files_per_second": 443.1544835696119,
- "bytes_per_second": 1849283471.4403222,

### DIFF
--- a/packages/orchestrator/cmd/clean-nfs-cache/ex/clean.go
+++ b/packages/orchestrator/cmd/clean-nfs-cache/ex/clean.go
@@ -232,10 +232,7 @@ func (c *Cleaner) Clean(ctx context.Context) error {
 			)
 
 			// reinsert the "younger" candidates back into the directory tree
-			c.timeit(ctx,
-				fmt.Sprintf("reinsert %v candidates", len(batch[c.DeleteN:])), func() {
-					c.reinsertCandidates(batch[c.DeleteN:])
-				})
+			c.reinsertCandidates(batch[c.DeleteN:])
 
 			total := uint64(0)
 			for _, toDelete := range batch[:c.DeleteN] {
--- a/packages/orchestrator/cmd/clean-nfs-cache/ex/scan.go
+++ b/packages/orchestrator/cmd/clean-nfs-cache/ex/scan.go
@@ -21,11 +21,7 @@ func (c *Cleaner) Scanner(ctx context.Context, candidateCh chan<- *Candidate, er
 		case <-ctx.Done():
 			return
 		default:
-			var candidate *Candidate
-			var err error
-			c.timeit(ctx, "find candidate", func() {
-				candidate, err = c.FindCandidate(ctx)
-			})
+			candidate, err := c.FindCandidate(ctx)
 
 			switch {
 			case err == nil:


packages/orchestrator/cmd/clean-nfs-cache/main2.go
```diff
@@ -0,0 +1,183 @@
...
+		return
+	}
+
+	if c.RemoveC.Load() == 0 {
+		log.Info(ctx, "no files deleted")
+
+		return
+	}
+
+	mean, sd := standardDeviation(c.DeletedAges)
```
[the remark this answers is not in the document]
levb: Not really relevant since everything is already finished by then

packages/orchestrator/cmd/clean-nfs-cache/main2.go
```diff
@@ -0,0 +1,183 @@
...
+			if log != nil {
+				log.Error(ctx, "NFS cache cleaner failed", zap.Error(err))
+			} else {
+				fmt.Println("NFS cache cleaner failed:", err)
+			}
+			os.Exit(1)
+		}
+	}()
+
+	opts, log, err = preRun(ctx)
```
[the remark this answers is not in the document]
levb: added defer log.Sync()

packages/orchestrator/cmd/clean-nfs-cache/ex/clean.go
```diff
@@ -0,0 +1,365 @@
...
+	ErrUsage      = errors.New("usage: clean-nfs-cache <path> [<options>]")
+)
+
+func NewCleaner(opts Options, log logger.Logger) *Cleaner {
+	c := &Cleaner{
+		Options:  opts,
+		Logger:   log,
+		mu:       sync.RWMutex{},
+		basePath: filepath.Dir(opts.Path),
+		cacheRoot: Dir{
```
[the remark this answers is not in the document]
levb: Understood, will be discussed in the review.

packages/orchestrator/cmd/clean-nfs-cache/ex/clean.go
```diff
@@ -0,0 +1,365 @@
...
+	}
+	return c
+}
+
+func (c *Cleaner) Clean(ctx context.Context) error {
+	if err := c.validateOptions(); err != nil {
+		return err
+	}
+
+	errCh := make(chan error)
```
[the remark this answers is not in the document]
levb: This is not a possibility. Once the event loop is entered, the only way out is to drain, which closes `quitCh`. I rearranged the code a little, and renamed the channels to make it more intention-revealing.

packages/orchestrator/cmd/clean-nfs-cache/ex/clean.go
```diff
@@ -0,0 +1,365 @@
...
+
+			c.Info(ctx, "selected batch",
+				zap.Int("count", len(batch)),
+				zap.Int64("oldest_atime_unix", batch[0].ATimeUnix),
+				zap.Int64("newest_atime_unix", batch[len(batch)-1].ATimeUnix),
+			)
+
+			// reinsert the "younger" candidates back into the directory tree
+			c.timeit(ctx, "reinserting candidates", func() {
+				c.reinsertCandidates(batch[c.DeleteN:])
```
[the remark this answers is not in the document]
levb: There is already
```go
	if c.BatchN < c.DeleteN {
		return errors.New("files-per-loop must be >= deletions-per-loop")
	}
```

packages/orchestrator/cmd/clean-nfs-cache/ex/clean.go
```diff
@@ -0,0 +1,375 @@
...
+	TargetDiskUsagePercent float64
+	OtelCollectorEndpoint  string
+}
+
+type Counters struct {
+	SeenDirC  atomic.Uint32
+	SeenFileC atomic.Uint64
+
+	DeletedBytes atomic.Uint64
+	DeletedAges  []time.Duration
```
levb: @djeebus We are not doing much with this, we could just maintain the running average without expensing the memory. Or, implement a log-scale histogram to keep track of the deleted ages?

(ditto for btime, not currently using it at all)

packages/orchestrator/cmd/clean-nfs-cache/ex/scan.go
```diff
@@ -0,0 +1,101 @@
...
+
+	if i < len(d.Dirs) {
+		return nil, d.Dirs[i], nil
+	}
+
+	// file needs to be unlinked before it's returned
+	f := d.Files[len(d.Files)-1]
+	d.Files = d.Files[:len(d.Files)-1]
+
+	return &f, nil, nil
```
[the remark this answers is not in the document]
levb: The comment was not exactly true, but I added tests and refactored to make it cleaner

packages/shared/pkg/feature-flags/client.go
```diff
@@ -65,6 +66,19 @@ func (c *Client) BoolFlag(ctx context.Context, flag BoolFlag, contexts ...ldcont
...
+		return flag.fallback, fmt.Errorf("LaunchDarkly client is not initialized")
+	}
+
+	v, err := c.ld.JSONVariationCtx(ctx, flag.name, mergeContexts(ctx, contexts), flag.fallback)
+	if err != nil {
+		return v, fmt.Errorf("error evaluating %s: %w", flag, err)
+	}
+
+	return v, nil
+}
```
djeebus, line 80: Might be cleaner and more type-safe to pass in a pointer to a struct, then populate it using some Unmarshal method. More code, maybe, but if we use this in more places, probably useful. Definitely not necessary though!

packages/orchestrator/cmd/clean-nfs-cache/ex/clean.go
```diff
@@ -0,0 +1,374 @@
...
+	StatxC      atomic.Int64
+	RemoveC     atomic.Int64
+	RemoveDirC  atomic.Int64
+}
+
+const (
+	initial = iota
+	scanning
+	scanned
+)
```
djeebus, line 65: best practice is to prefix this with something to make it clear that they belong together. maybe something like `statusInitial`, `statusScanning`, etc?

packages/orchestrator/cmd/clean-nfs-cache/ex/clean.go
```diff
@@ -0,0 +1,374 @@
...
+	}
+	if c.MaxConcurrentScan <= 0 {
+		return errors.New("max-concurrent-scan must be > 0")
+	}
+	if c.MaxConcurrentDelete <= 0 {
+		return errors.New("max-concurrent-delete must be > 0")
+	}
+
+	return nil
+}
```
djeebus, line 144: It'd probably be more user-friendly to batch this up with `errors.Join`:

```golang
func demoFunc() error { 
	var errs []error
	if err := badThing(); err != nil {
		errs = append(errs, err)
	}
	if err := otherThing(); err != nil { 
		errs = append(errs, err)
	}
	return errors.Join(errs...)
}
```

`errors.Join` is smart enough to return a single error if that's all there is, to wrap them in a group if there's more than one, and return `nil` if there are no errors. It also ignores `nil` errors, which can clean up the code even more. For example, the above example could be rewritten a little simpler:

```golang
func demoFunc() error { 
	errs := []error {
		badThing(),
		otherThing(),
	}
	return errors.Join(errs...)
}
```

Not necessary though!

packages/orchestrator/cmd/clean-nfs-cache/ex/clean.go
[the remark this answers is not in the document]
djeebus: I think it's correct that `errCh` is never closed.

packages/orchestrator/cmd/clean-nfs-cache/ex/clean.go
```diff
@@ -0,0 +1,374 @@
...
+	}
+
+	ctx, cancel := context.WithCancel(ctx)
+
+	errCh := make(chan error)
+	candidateCh := make(chan *Candidate, c.MaxConcurrentScan*2)
+	deleteCh := make(chan *Candidate, c.MaxConcurrentDelete*2)
+	drainedCh := make(chan struct{})
+	cleanShutdown := &sync.WaitGroup{}
+	c.statRequestCh = make(chan *statReq)
```
djeebus, line 159: Is there value in making this channel buffered?

packages/orchestrator/cmd/clean-nfs-cache/main2.go
```diff
@@ -0,0 +1,196 @@
...
+
+	"github.com/e2b-dev/infra/packages/orchestrator/cmd/clean-nfs-cache/ex"
+	"github.com/e2b-dev/infra/packages/orchestrator/cmd/clean-nfs-cache/pkg"
+	"github.com/e2b-dev/infra/packages/shared/pkg/env"
+	featureflags "github.com/e2b-dev/infra/packages/shared/pkg/feature-flags"
+	"github.com/e2b-dev/infra/packages/shared/pkg/logger"
+	"github.com/e2b-dev/infra/packages/shared/pkg/utils"
+)
+
+func main() {
```
djeebus: Can you move things around such that the `main` function is in `main.go`, and the old version is either moved somewhere (maybe just an `old` sub package?) or deleted altogether?

packages/orchestrator/cmd/clean-nfs-cache/ex/clean.go
```diff
@@ -0,0 +1,374 @@
...
+		return err
+	}
+
+	ctx, cancel := context.WithCancel(ctx)
+
+	errCh := make(chan error)
+	candidateCh := make(chan *Candidate, c.MaxConcurrentScan*2)
+	deleteCh := make(chan *Candidate, c.MaxConcurrentDelete*2)
+	drainedCh := make(chan struct{})
+	cleanShutdown := &sync.WaitGroup{}
```
djeebus: This is named a little awkwardly. It contains a counter that, when it reaches zero, we can then cleanly shut down, right? Assuming yes, maybe we can rename it to `runningGoroutines` or `procsInProgress`, something like that. The code makes it clear that `drainedCh` is `Wait`ing on that number to reduce to zero, but the above change would make it clear what that number represents.

packages/orchestrator/cmd/clean-nfs-cache/ex/clean.go
```diff
@@ -0,0 +1,374 @@
...
+
+func (c *Cleaner) timeit(ctx context.Context, message string, fn func()) {
+	start := time.Now()
+	fn()
+	done := time.Since(start).Round(time.Millisecond)
+
+	c.Debug(ctx, message, zap.Duration("duration", done))
+}
+
+func CreateTestDir(path string, nDirs int, nFiles int, fsize int) {
```
djeebus, line 351: This should probably be moved to a `_test.go` file

packages/orchestrator/cmd/clean-nfs-cache/ex/delete.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func (c *Cleaner) deleteFile(ctx context.Context, candidate *Candidate) {
+	// Best-effort: get current metadata to detect atime changes or if file is gone
+	meta, err := c.stat(candidate.FullPath)
+	c.DeleteAttemptC.Add(1)
+
+	switch {
+	case err != nil:
+		c.DeleteAlreadyGoneC.Add(1)
```
djeebus: Not sure if it's worth the effort, but you might consider checking `if os.IsNotExist(err)`, and only adding to the counter when appropriate, logging otherwise. we might find some interesting errors (stale file handles, io timeout, etc).

packages/orchestrator/cmd/clean-nfs-cache/ex/scan_test.go
```diff
@@ -0,0 +1,82 @@
...
+	defer dfsub.Close()
+
+	sub, err = c.scanDir(ctx, []*Dir{c.root, sub})
+	require.NoError(t, err)
+	require.True(t, sub.IsScanned())
+	require.False(t, sub.IsEmpty())
+	require.NotEmpty(t, sub.Files)
+}
+
+func TestRandomSubdirOrOldestFile(t *testing.T) {
```
djeebus, line 59: Can we also add a few more tests for:
- returns a directory
- returns nothing

packages/orchestrator/cmd/clean-nfs-cache/ex/stat.go
```diff
@@ -0,0 +1,179 @@
...
+		case req := <-c.statRequestCh:
+			f, err := c.statInDir(req.df, req.name)
+			req.f = f
+			req.err = err
+			req.response <- req
+		}
+	}
+}
+
+func (c *Cleaner) scanDir(ctx context.Context, path []*Dir) (out *Dir, err error) {
```
djeebus: This should probably be in `scan.go`?

packages/orchestrator/cmd/clean-nfs-cache/ex/stat.go
```diff
@@ -0,0 +1,179 @@
...
+	d.Dirs = dirs
+	d.Files = files
+	d.sort()
+	d.state = scanned
+	d.mu.Unlock()
+
+	return d, nil
+}
+
+func (c *Cleaner) removeEmptyDir(ctx context.Context, path []*Dir) {
```
djeebus: This should probably be in `scan.go` as well.

packages/orchestrator/cmd/clean-nfs-cache/ex/stat_osx.go
djeebus, line 1: Might consider renaming this to `stat_osx.go`, since it's implementing features only used in `stat.go`?

packages/orchestrator/cmd/clean-nfs-cache/ex/stat_linux.go
djeebus, line 1: Might consider renaming this to `stat_linux.go`, since it's implementing features only used in `stat.go`?

packages/orchestrator/cmd/clean-nfs-cache/ex/stat.go
```diff
@@ -0,0 +1,179 @@
...
+	case scanning:
+		d.mu.Unlock()
+
+		return nil, ErrBusy
+
+	default:
+		// continue
+	}
+	d.state = scanning
+	d.mu.Unlock()
```
djeebus: might be cleaner to stick this into something like `d.getState()` to hide the details of the locking when reading it.

packages/orchestrator/cmd/clean-nfs-cache/ex/clean.go
djeebus: OK, reading a bit more, it seems like a better prefix might be `dirStatusUnscanned`, `dirStatusScanning`, etc.

packages/orchestrator/cmd/clean-nfs-cache/ex/clean.go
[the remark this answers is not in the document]
levb: Added a memory constraint (2Gb) at the Nomad level

packages/orchestrator/cmd/clean-nfs-cache/ex/stat.go
levb: Need to update the state inside the lock here, 0/5 that something like `examineState(func changeState)` would be less intention-revealing.

packages/orchestrator/cmd/clean-nfs-cache/ex/stat.go
```diff
@@ -0,0 +1,164 @@
...
+
+	absPath := c.abs(path, "")
```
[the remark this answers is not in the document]
levb: Fixed

packages/orchestrator/cmd/clean-nfs-cache/ex/clean.go
levb: It needs to be exposed for `compare_test.go` which is 1 level up; I do not know if one can export function between tests in packages

packages/orchestrator/cmd/clean-nfs-cache/main2.go
levb: Just did, was minimizing the diffs before; consolidated is cleaner.

packages/orchestrator/cmd/clean-nfs-cache/ex/clean.go
levb: I'd argue the opposite: if there is no obvious use case for the buffer (and therefore not an obvious buffer _size_), stick with the unbuffered channel. I changed the other 2 to be unbuffered (I added the buffer when I was debugging the deadlocks)

packages/shared/pkg/feature-flags/client.go
levb: I'd like to avoid it for now; specifically, 1/5 returning the flags value allows for easier testing of whether specific fields were set.

packages/orchestrator/cmd/clean-nfs-cache/ex/clean.go
levb: I can see a small improvement in testing times when I use buffered channels, but I think it's just because of the warm up which is serialized with unbuffered.

packages/orchestrator/cmd/clean-nfs-cache/ex/clean.go
djeebus: Works for me! I only added it to log an estimate of how useful this is, and how aggressive we're being with it; histogram would be just fine.

packages/orchestrator/cmd/clean-nfs-cache/ex/clean.go
djeebus: You can't :cry: the typical pattern for this scenario is to put them into a package called `testutils`; you can check this project for examples.

PR #1574, opened 2025-12-04
feat(iac): add memory metric for client autoscaler

ken8203:
## Summary

- Add memory utilization metric to client node pool autoscaler
- Make CPU and memory autoscaling thresholds configurable via environment variables

## Motivation

During self-hosting, we observed that sandboxes consume significant memory. Relying solely on CPU utilization for autoscaling decisions can result in nodes running out of memory before CPU-based scaling triggers. This change adds memory-based scaling to ensure the cluster scales out appropriately under memory pressure.

Thanks for reviewing! Happy to address any feedback or questions.

### DIFF
--- a/iac/provider-gcp/nomad-cluster/nodepool-client.tf
+++ b/iac/provider-gcp/nomad-cluster/nodepool-client.tf
@@ -62,11 +62,14 @@ resource "google_compute_region_autoscaler" "client" {
       target = var.client_cluster_autoscaling_cpu_target
     }
 
-    metric {
-      name   = "agent.googleapis.com/memory/percent_used"
-      type   = "GAUGE"
-      filter = "resource.type = \"gce_instance\""
-      target = var.client_cluster_autoscaling_memory_target
+    dynamic "metric" {
+      for_each = var.client_cluster_autoscaling_memory_target < 100 ? [1] : []
+      content {
+        name   = "agent.googleapis.com/memory/percent_used"
+        type   = "GAUGE"
+        filter = "resource.type = \"gce_instance\" AND metric.labels.state = \"used\""
+        target = var.client_cluster_autoscaling_memory_target
+      }
     }
   }
 }


ken8203:
> Hey @ken8203, thanks for sumbitting the PR — it looks good overall. One thing to keep in mind: we preallocate part of the memory to huge pages, and that portion is counted as utilized. The current default is 80%. Could you bump the default autoscaling threshold to something higher than 80%?
> 
> Thanks!

Oh, I didn't notice the base huge pages. Thanks for the reminder!

Fixed it in be0153c.

iac/provider-gcp/nomad-cluster/nodepool-client.tf
```diff
@@ -59,7 +59,14 @@ resource "google_compute_region_autoscaler" "client" {
...
-      target = 0.6
+      target = var.client_cluster_autoscaling_cpu_target
+    }
+
+    metric {
+      name   = "agent.googleapis.com/memory/percent_used"
+      type   = "GAUGE"
+      filter = "resource.type = \"gce_instance\""
+      target = var.client_cluster_autoscaling_memory_target
     }
```
jakubno: One small improvement, if you don't mind. This will allow to turn it off, if you set it to 100% or higher
```suggestion
    dynamic "metric" {
      for_each = var.client_cluster_autoscaling_memory_target < 100 ? [1] : []
      content {
        name   = "agent.googleapis.com/memory/percent_used"
        type   = "GAUGE"
        filter = "resource.type = \"gce_instance\""
        target = var.client_cluster_autoscaling_memory_target
      }
    }
```

iac/provider-gcp/nomad-cluster/nodepool-client.tf
```diff
@@ -59,7 +59,14 @@ resource "google_compute_region_autoscaler" "client" {
...
 
     cpu_utilization {
-      target = 0.6
+      target = var.client_cluster_autoscaling_cpu_target
+    }
+
+    metric {
+      name   = "agent.googleapis.com/memory/percent_used"
+      type   = "GAUGE"
+      filter = "resource.type = \"gce_instance\""
```
jakubno: Also we should compare it only to used memory
```suggestion
      filter = "resource.type = \"gce_instance\" AND metric.labels.state = \"used\""
```

iac/provider-gcp/nomad-cluster/nodepool-client.tf
ken8203: Of course, it makes sense!

iac/provider-gcp/nomad-cluster/nodepool-client.tf
ken8203: 👍🏽

jakubno (approved):

jakubno:
Thanks, @ken8203 ! Great contribution.

PR #1580, opened 2025-12-04
Put envd, ptys, socats, and commands into their own cgroups

djeebus:
This configures memory reservations on the envd service, and puts internal processes into cgroups based on their process type. Most importantly, this keeps envd responsive when the system is saturated.

## envd.service cgroup:
the envd process. manages and handles communication into the sandbox.
- memory.min = 50 megabytes. envd seems to use ~10 MB of memory, so this should be plenty
- memory.low = 100 megabytes. if available memory is less than this, try to free up more memory
- cpu.weight = 1000. this gives it 10x the priority of an average process, and 20x background procs launched by the user. this isn't a reservation, it's just priority. if envd doesn't need the cpu cycles, they go to other processes.

## pty cgroup:
any process that defines a pty configuration. assumed to be interactive, so it gets higher priority.
- cpu.weight = 200

## socat cgroup:
forwarded ports. shouldn't require much cpu, but is used for interaction, so it gets higher priority.
- cpu.weight = 150

## user cgroup:
every other command launched by the user. the lowest priority, as it has the highest chance of interrupting envd.
- cpu.weight = 50
- memory.high = 7/8 of the total sandbox

---

> [!NOTE]
> Adds a cgroup v2 manager and assigns envd, PTY, socat, and user processes to dedicated cgroups with CPU/memory settings; integrated into process handling, port forwarding, and orchestrator templates.
> 
> - **Cgroups (new)**:
>   - Add `cgroups` package with cgroup v2 manager (`Cgroup2Manager`), options, noop manager, and tests.
> - **envd runtime**:
>   - `main.go`: create and configure cgroup manager (flags: `--cgroup-root`), set per-type CPU/memory (`pty`, `socat`, `user`), pass to process service and port forwarder; bump `Version` to `0.4.3`.
>   - `port/forward.go`: run `socat` in `ProcessTypeSocat` cgroup via `SysProcAttr` FD.
>   - `services/process`: plumb cgroup manager; `handler.New` sets `SysProcAttr{UseCgroupFD,CgroupFD}` and selects proc type (`pty` vs `user`).
> - **Orchestrator**:
>   - Update `envd.service.tpl` with cgroup delegation and resource reservations (`Delegate=yes`, `MemoryMin/Low`, `CPUWeight`).
>   - Bump `github.com/opencontainers/runtime-spec` to `v1.3.0`; adjust rootfs test accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4d795e681c7515fdc0ad87ac79642c94d204586f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/envd/internal/services/cgroups/cgroup2_test.go
+++ b/packages/envd/internal/services/cgroups/cgroup2_test.go
@@ -166,7 +166,7 @@ func startProcess(t *testing.T, m *Cgroup2Manager, pt ProcessType) *exec.Cmd {
 func waitForProcess(t *testing.T, cmd *exec.Cmd, timeout time.Duration) error {
 	t.Helper()
 
-	done := make(chan error)
+	done := make(chan error, 1)
 
 	go func() {
 		defer close(done)


packages/orchestrator/internal/sandbox/fc/process.go
```diff
@@ -260,6 +260,7 @@ func (p *Process) Create(
 
 		"reboot":           "k",
 		"pci":              "off",
+		"psi":              "1",
```
ValentaTomas: What does this do?

djeebus: This turns on [pressure stall information](https://docs.kernel.org/accounting/psi.html), which is what systemd-oomd kill processes when they're under memory pressure. We compile it into the kernel, but disable it by default.

ValentaTomas: It is possible that for some templates the psi can be off and the new envd config with oom can be enabled? (I don't think so, but haven't though about it in depth).

djeebus: It's possible, if the templates are created from old templates ... I think. I'll have to dig more into the order-of-operations when building templates. If that situation does happen, it means that we won't have access to memory stall information, which won't allow cgroups to recognize that a process is struggling to find memory. We'll still be able to keep envd responsive, and the process will still be killed if it tries to use too much memory, which I think are still wins.

packages/api/internal/auth/middleware.go
```diff
@@ -80,7 +80,7 @@ func (a *commonAuthenticator[T]) Authenticate(ctx context.Context, input *openap
 	headerKey, err := a.getHeaderKeysFromRequest(input.RequestValidationInput.Request)
 	if err != nil {
 		err = fmt.Errorf("%s: %w", a.errorMessage, err)
-		trace.SpanFromContext(ctx).RecordError(err, trace.WithStackTrace(true))
```
djeebus, line 83: Can we log fewer stack traces? Do they provide value? They seem to just take up space. I can bring this back if we find them useful.

PR #1597, opened 2025-12-09
feat: add hostname based egress firewall

dobrac:
Add possibility to allow egress traffic only to specific hostnames. This filtering is done only for  (TCP) TLS and HTTP traffic. The current implementation doesn't support ECH (encrypted TLS negotiation) and QUIC/HTTP3.

The implementation uses user space proxy in Go, so every TCP connection goes kernel -> user space -> kernel, but it allows the SNI/HTTP sniffing.

Domains are supported only in `allowOut`, because the use case for `denyOut` not clear - could be easily bypassed. Also, as HTTP3 (UDP based) is not supported, it could be bypassed even on the protocol level.

The example usage is the following:
```ts
 const sbx = await Sandbox.create({
    network: {
      allowOut: ["google.com"],
      denyOut: [ALL_TRAFFIC],
    },
  });
```

```py
sandbox = Sandbox.create(
    network={
        "allow_out": ["google.com"],
        "deny_out": [ALL_TRAFFIC]
    }
)
```

Wildcards for subdomains are also supported:
```ts
 const sbx = await Sandbox.create({
    network: {
      allowOut: ["*.mydomain.com"],
      denyOut: [ALL_TRAFFIC],
    },
  });
```

```py
sandbox = Sandbox.create(
    network={
        "allow_out": ["*.mydomain.com"],
        "deny_out": [ALL_TRAFFIC]
    }
)
```

---

> [!NOTE]
> Add hostname-based egress filtering for TCP (HTTP/TLS) via a user-space proxy and wire it through API validation, orchestrator networking, telemetry, and tests.
> 
> - **API**:
>   - Validate `network.allowOut` domains require `denyOut` to include `ALL_TRAFFIC` (`0.0.0.0/0`); validate CIDRs/IPs; add clear client error.
>   - Split `AllowedAddresses` into `allowed_cidrs` and `allowed_domains`; auto-allow default nameserver when domains present.
>   - Add unit tests for network config validation.
> - **Orchestrator**:
>   - Introduce TCP egress firewall proxy (hostname/SNI/Host filtering) using `inet.af/tcpproxy`; add metrics and startup/shutdown wiring.
>   - Refactor nftables/iptables flow: mark allowed traffic, redirect unmarked TCP to proxy; add configurable TCP firewall port.
>   - Extend proto with `SandboxNetworkEgressConfig.allowed_domains`; plumb through create flow and node manager (recombine addresses).
>   - Templating: resolv.conf nameserver is configurable.
>   - Hyperloop handlers use `sandboxes.GetByHostPort` for sandbox lookup.
> - **Tests**:
>   - Add comprehensive integration tests for egress rules (IP/CIDR/domain, wildcards, resume, HTTP/HTTPS/UDP) and unit tests for domain matching/egress decisions.
> - **Telemetry/Deps**:
>   - Add TCP firewall metrics (connections, errors, decisions, durations) and `inet.af/tcpproxy` dependency.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8297f2b79bb1c4348ca971de57b436af34ce98bb. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api/internal/handlers/sandbox_create.go
+++ b/packages/api/internal/handlers/sandbox_create.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"slices"
 	"strings"
 	"time"
 
@@ -36,7 +37,6 @@ const (
 
 	// Network validation error messages
 	ErrMsgDomainsRequireBlockAll = "When specifying allowed domains in allow out, you must include 'ALL_TRAFFIC' in deny out to block all other traffic."
-	ErrMsgCIDRsRequireDenyOut    = "When specifying allowed CIDRs in allow out, you must include intersecting denied CIDRs."
 )
 
 // mostUsedTemplates is a map of the most used template aliases.
@@ -340,83 +340,24 @@ func validateNetworkConfig(network *api.SandboxNetworkConfig) *api.APIError {
 		}
 	}
 
-	// Validate that AllowOut rules are meaningful (require corresponding DenyOut)
+	// Validate that allow out rules have corresponding deny out rules
 	allowOut := sharedUtils.DerefOrDefault(network.AllowOut, nil)
 	if len(allowOut) > 0 {
-		allowedCIDRs, allowedDomains := sandbox_network.ParseAddressesAndDomains(allowOut)
+		_, allowedDomains := sandbox_network.ParseAddressesAndDomains(allowOut)
 
 		// Check if DenyOut contains block-all CIDR
-		hasBlockAll := false
-		for _, cidr := range denyOut {
-			if cidr == sandbox_network.AllInternetTrafficCIDR {
-				hasBlockAll = true
-
-				break
-			}
-		}
+		hasBlockAll := slices.Contains(denyOut, sandbox_network.AllInternetTrafficCIDR)
 
 		// When specifying domains, require block-all CIDR in DenyOut
 		// Without this, domain filtering is meaningless (traffic is allowed by default)
 		if len(allowedDomains) > 0 && !hasBlockAll {
 			return &api.APIError{
 				Code:      http.StatusBadRequest,
-				Err:       fmt.Errorf("allow out contains domains but deny out is missing 0.0.0.0/0"),
+				Err:       fmt.Errorf("allow out contains domains but deny out is missing 0.0.0.0/0 (ALL_TRAFFIC)"),
 				ClientMsg: ErrMsgDomainsRequireBlockAll,
-			}
-		}
-
-		// When specifying CIDRs in AllowOut, require that each allowed CIDR is covered by denied CIDRs
-		// Without denied CIDRs that cover the allowed range, allowlist is meaningless
-		if len(allowedCIDRs) > 0 {
-			if len(denyOut) == 0 {
-				return &api.APIError{
-					Code:      http.StatusBadRequest,
-					Err:       fmt.Errorf("allow out contains CIDRs but deny_out is empty"),
-					ClientMsg: ErrMsgCIDRsRequireDenyOut,
-				}
-			}
-
-			// If block-all is present, all allowed CIDRs are covered
-			if !hasBlockAll {
-				// Check each allowed CIDR has intersection with denied CIDRs
-				for _, allowedCIDR := range allowedCIDRs {
-					if !cidrIntersectsAny(allowedCIDR, denyOut) {
-						return &api.APIError{
-							Code:      http.StatusBadRequest,
-							Err:       fmt.Errorf("allowed CIDR %s has no intersection with any denied CIDR", allowedCIDR),
-							ClientMsg: fmt.Sprintf("Allowed CIDR '%s' is not covered by any denied CIDR. The allowed CIDR must intersect with at least one denied CIDR to be meaningful.", allowedCIDR),
-						}
-					}
-				}
 			}
 		}
 	}
 
 	return nil
-}
-
-// cidrIntersectsAny checks if a CIDR (or IP) intersects with any CIDR in the list.
-// Two CIDRs intersect if one contains the network address of the other.
-func cidrIntersectsAny(cidr string, cidrList []string) bool {
-	// Parse the CIDR we're checking (convert IP to /32 CIDR if needed)
-	cidrStr := sandbox_network.AddressStringToCIDR(cidr)
-	_, cidrNet, err := net.ParseCIDR(cidrStr)
-	if err != nil {
-		return false
-	}
-
-	for _, deniedCIDR := range cidrList {
-		deniedStr := sandbox_network.AddressStringToCIDR(deniedCIDR)
-		_, deniedNet, err := net.ParseCIDR(deniedStr)
-		if err != nil {
-			continue
-		}
-
-		// Check if either CIDR contains the other's network address
-		if deniedNet.Contains(cidrNet.IP) || cidrNet.Contains(deniedNet.IP) {
-			return true
-		}
-	}
-
-	return false
 }
--- a/packages/api/internal/handlers/sandbox_create_test.go
+++ b/packages/api/internal/handlers/sandbox_create_test.go
@@ -89,13 +89,11 @@ func TestValidateNetworkConfig(t *testing.T) {
 		},
 		// CIDR validation tests
 		{
-			name: "allow_out with CIDR requires deny_out",
+			name: "allow_out with CIDR without deny_out is valid",
 			network: &api.SandboxNetworkConfig{
 				AllowOut: &[]string{"10.0.0.0/8"},
 			},
-			wantErr:    true,
-			wantCode:   http.StatusBadRequest,
-			wantErrMsg: ErrMsgCIDRsRequireDenyOut,
+			wantErr: false,
 		},
 		{
 			name: "allow_out with CIDR and deny_out block-all is valid",
@@ -106,13 +104,11 @@ func TestValidateNetworkConfig(t *testing.T) {
 			wantErr: false,
 		},
 		{
-			name: "allow_out with IP requires deny_out",
+			name: "allow_out with IP without deny_out is valid",
 			network: &api.SandboxNetworkConfig{
 				AllowOut: &[]string{"8.8.8.8"},
 			},
-			wantErr:    true,
-			wantCode:   http.StatusBadRequest,
-			wantErrMsg: ErrMsgCIDRsRequireDenyOut,
+			wantErr: false,
 		},
 		{
 			name: "allow_out with IP and deny_out block-all is valid",
@@ -124,14 +120,12 @@ func TestValidateNetworkConfig(t *testing.T) {
 		},
 		// CIDR intersection validation tests
 		{
-			name: "allow_out CIDR not covered by deny_out CIDR is invalid",
+			name: "allow_out CIDR not covered by deny_out CIDR is valid (no intersection check)",
 			network: &api.SandboxNetworkConfig{
 				AllowOut: &[]string{"10.0.0.0/8"},
-				DenyOut:  &[]string{"192.168.0.0/16"}, // No intersection
+				DenyOut:  &[]string{"192.168.0.0/16"}, // No intersection, but still valid
 			},
-			wantErr:    true,
-			wantCode:   http.StatusBadRequest,
-			wantErrMsg: "Allowed CIDR '10.0.0.0/8' is not covered by any denied CIDR. The allowed CIDR must intersect with at least one denied CIDR to be meaningful.",
+			wantErr: false,
 		},
 		{
 			name: "allow_out CIDR covered by intersecting deny_out CIDR is valid",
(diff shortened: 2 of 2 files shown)


packages/orchestrator/internal/sandbox/network/proxy/proxy.go
```diff
@@ -0,0 +1,284 @@
...
+	if err != nil {
+		return false, err
+	}
+
+	if net := sbx.Config.Network; net != nil {
+		for _, domain := range net.GetEgress().GetAllowedDomains() {
+			if strings.EqualFold(domain, hostname) {
+				return true, nil
+			}
+			if strings.HasPrefix(domain, "*.") {
```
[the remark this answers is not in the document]
dobrac: how is *. case sensitive?

packages/orchestrator/internal/tcpfirewall/handlers.go
```diff
@@ -0,0 +1,182 @@
...
+	// Get hostname from tcpproxy's wrapped connection
+	var hostname string
+	if tc, ok := conn.(*tcpproxy.Conn); ok {
+		hostname = tc.HostName
+	}
+
+	if hostname == "" {
+		// No hostname found, this is the case e.g. for https://1.1.1.1 like requests
+		logger.Debug(ctx, "No hostname found, ignoring hostname based filter", zap.String("source_addr", sourceAddr))
+	}
```
djeebus: This seems like a pretty big security flaw; if they want to egress to google.com and that domain is blocked, all they need to do is `nslookup google.com` and then swap the host name for an address, right?

djeebus: Ah, unless requests are intended to match BOTH allowed domains and allowed IPs/CIDRs. OK, I'm on board, but a confirmation that I'm reading this correctly would be appreciated 🙏

packages/orchestrator/internal/tcpfirewall/handlers.go
```diff
@@ -0,0 +1,182 @@
...
+	dp.HandleConn(conn)
+}
+
+// isEgressAllowed checks if egress is allowed based on domain and CIDR rules.
+// Priority order:
+//  1. Allow domain / Allow CIDR (if either matches → allow)
+//  2. Deny domain / Deny CIDR (if either matches → deny)
+//  3. Default: allow
+func isEgressAllowed(sbx *sandbox.Sandbox, hostname string, ip net.IP) (bool, error) {
```
djeebus: This function is really important, and we'd benefit from having tests covering it all.

packages/orchestrator/internal/tcpfirewall/handlers.go
```diff
@@ -0,0 +1,182 @@
...
+		}
+	}
+
+	// Default: allow all traffic.
+	return true, nil
+}
+
+// matchDomain checks if a hostname matches a domain pattern.
+// Patterns can be exact matches, wildcards (*), or suffix wildcards (*.example.com).
+func matchDomain(hostname, pattern string) bool {
```
djeebus, line 158: This would be a good function to cover in tests.

packages/orchestrator/internal/tcpfirewall/handlers.go
```diff
@@ -0,0 +1,182 @@
...
+	switch {
+	case strings.EqualFold(pattern, hostname):
+		return true
+	case strings.EqualFold(pattern, "*"):
+		return true
+	case strings.HasPrefix(pattern, "*."):
+		suffix := pattern[1:]
+		if strings.HasSuffix(strings.ToLower(hostname), strings.ToLower(suffix)) {
+			return true
+		}
```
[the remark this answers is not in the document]
djeebus: This feels like a false positive, but would be good to have tests to prove this.

packages/orchestrator/internal/tcpfirewall/handlers.go
dobrac: You can't block domains, only allow them when CIDRs are blocked. So you can only allow list the domain when all the underlying IPs are denied -> accessing the IP directly will be blocked.

tests/integration/internal/tests/api/sandboxes/sandbox_network_out_test.go
```diff
@@ -1,55 +1,130 @@
...
+
+	return networkTestTemplateID
+}
+
+// =============================================================================
+// Test helper functions for network egress assertions
+// =============================================================================
+
+// assertHTTPRequest asserts that an HTTP/HTTPS request to the given URL succeeds or fails based on expectSuccess
+func assertHTTPRequest(t *testing.T, ctx context.Context, sbx *api.Sandbox, envdClient *setup.EnvdClient, url string, expectSuccess bool, msg string) {
```
djeebus: It would probably be a slightly cleaner interface to have two functions: `assertSuccessfulHTTPRequest` and `assertBlockedHTTPRequest`; there'd be fewer conditionals in the helper(s), and it would be easier to read the tests.

tests/integration/internal/tests/api/sandboxes/sandbox_network_out_test.go
```diff
@@ -1,55 +1,130 @@
...
+	if expectSuccess {
+		require.NoError(t, err, msg)
+	} else {
+		require.Error(t, err, msg)
+		require.Contains(t, err.Error(), "failed with exit code", "Expected connection failure message")
+	}
+}
+
+// assertDNSQuery asserts that a DNS query to the given server succeeds or fails based on expectSuccess
+func assertDNSQuery(t *testing.T, ctx context.Context, sbx *api.Sandbox, envdClient *setup.EnvdClient, dnsServer, domain string, expectSuccess bool, msg string) {
```
djeebus: Same as above, would probably be cleaner to have two functions - `assertSuccessfulDNSQuery` and `assertFailedDNSQuery`, or similar.

packages/api/internal/orchestrator/nodemanager/sandboxes.go
```diff
@@ -65,8 +65,10 @@ func (n *Node) GetSandboxes(ctx context.Context) ([]sandbox.Sandbox, error) {
 			}
 
 			if egress := config.GetNetwork().GetEgress(); egress != nil {
+				// Combine allowed CIDRs and domains back into AllowedAddresses
+				allowedAddresses := append(egress.GetAllowedCidrs(), egress.GetAllowedDomains()...)
```
jakubno: do you need to merge it? you split it right after

packages/orchestrator/internal/tcpfirewall/handlers.go
```diff
@@ -0,0 +1,182 @@
...
+func domainHandler(ctx context.Context, conn net.Conn, dstIP net.IP, dstPort int, sbx *sandbox.Sandbox, logger logger.Logger) {
+	sourceAddr := conn.RemoteAddr().String()
+
+	// Get hostname from tcpproxy's wrapped connection
+	var hostname string
+	if tc, ok := conn.(*tcpproxy.Conn); ok {
+		hostname = tc.HostName
+	}
+
+	if hostname == "" {
```
jakubno: Not sure what the constant is for tbh
```suggestion
	if hostname == noHostnameValue {
```

packages/orchestrator/internal/tcpfirewall/handlers.go
```diff
@@ -0,0 +1,182 @@
...
+	if hostname != noHostnameValue {
+		for _, domain := range egress.GetAllowedDomains() {
+			if matchDomain(hostname, domain) {
+				return true, nil // Explicitly allowed by domain
+			}
+		}
+	}
+
+	// Priority 1: Check allowed CIDRs
+	for _, cidr := range egress.GetAllowedCidrs() {
```
jakubno, line 129: Wouldn't it make sense to return false, nil, if AllowedCidrs are non empty and there no match?

packages/orchestrator/internal/tcpfirewall/handlers.go
```diff
@@ -0,0 +1,182 @@
...
+		zap.String("source_addr", sourceAddr),
+		zap.String("upstream_addr", upstreamAddr),
+	)
+
+	proxy(conn, upstreamAddr)
+}
+
+// proxy proxies the connection to the upstream address.
+func proxy(conn net.Conn, upstreamAddr string) {
+	dp := &tcpproxy.DialProxy{
```
jakubno, line 93: Looks like this should be very lightweight, so it should be fine

packages/orchestrator/internal/tcpfirewall/handlers.go
dobrac: added

packages/orchestrator/internal/tcpfirewall/handlers.go
dobrac: added

packages/api/internal/orchestrator/nodemanager/sandboxes.go
dobrac: yes, this is for the case when the sandbox is lost in the API (e.g. a restart) and we load it from the orchestrator

packages/orchestrator/internal/tcpfirewall/handlers.go
dobrac: added

packages/orchestrator/internal/tcpfirewall/handlers.go
dobrac: no, because by default the internet is allowed

packages/shared/pkg/sandbox-network/firewall.go
```diff
@@ -42,3 +45,29 @@ func AddressStringsToCIDRs(addressStrings []string) []string {
...
+	// Check if it's a valid IP address
+	if ip := net.ParseIP(s); ip != nil {
+		return true
+	}
+
+	// Check if it's a valid CIDR
+	_, _, err := net.ParseCIDR(s)
+
+	return err == nil
+}
```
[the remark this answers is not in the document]
dobrac: IPv6s are not supported

packages/orchestrator/internal/tcpfirewall/handlers_test.go
```diff
@@ -0,0 +1,420 @@
...
+			hostname: "example.com",
+			ip:       net.ParseIP("1.2.3.4"),
+			want:     true,
+		},
+		{
+			// No denied CIDRs configured, so traffic is allowed by default.
+			name: "default allow: no denied CIDRs means all traffic allowed",
+			network: &orchestrator.SandboxNetworkConfig{
+				Egress: &orchestrator.SandboxNetworkEgressConfig{
+					AllowedDomains: []string{"other.com"},
```
jakubno: not sure if this should be allowed, the config is in a sense invalid. We should required block all cidrs in a case when you specify domain. Similarly we should probably validate that if you specify CIDR, there should be non empty intersection of denied CIRR

 There's should be at least validation on API level, maybe having a validation also on orchestartor level wouldn't hurt

packages/orchestrator/internal/tcpfirewall/handlers_test.go
```diff
@@ -0,0 +1,420 @@
...
+		// ---------------------------------------------------------------------
+		{
+			name: "bypass: broad allowed CIDR bypasses specific denied CIDR",
+			// Warning: A broad allow will bypass a more specific deny.
+			// IP 10.1.1.1 matches allowed 10.0.0.0/8 -> bypass, never checks denied.
+			network: &orchestrator.SandboxNetworkConfig{
+				Egress: &orchestrator.SandboxNetworkEgressConfig{
+					AllowedCidrs: []string{"10.0.0.0/8"},
+					DeniedCidrs:  []string{"10.1.1.1/32"},
+				},
```
jakubno, line 254: oaky, for test, but as mentioned above, this is imho invalid config

dobrac: there is a non-empty intersection of the denied CIDRs in this case

packages/orchestrator/internal/tcpfirewall/handlers_test.go
dobrac: > not sure if this should be allowed, the config is in a sense invalid. We should required block all cidrs in a case when you specify domain. Similarly we should probably validate that if you specify CIDR, there should be non empty intersection of denied CIRR

fair point. I'm not confident about saying that all CIDRs should be blocked in general is enough, but we can start with that.

> having a validation also on orchestartor level wouldn't hurt

I don't want to duplicate the validation logic as it doesn't add much benefit and slows down updates. (As the "invalid" values don't break anything, they just don't do anything.)

dobrac: added the validation to the API for both:
- allow out domain requires ALL_TRAFFIC blocked
- allow out must intersect with at least one deny out cidr

jakubno (approved):

PR #1602, opened 2025-12-10
Add client and builder cluster configuration via JSON

jakubno:
For migration run `make migrate-build-cluster`

---

> [!NOTE]
> Modernizes cluster configuration and Terraform structure.
> 
> - Introduces JSON-based `BUILD_CLUSTERS_CONFIG` and `CLIENT_CLUSTERS_CONFIG` with examples in `.env.template`; removes legacy vars
> - Adds reusable `nomad-cluster/worker-cluster` module and instantiates via `for_each` for build/client clusters; deletes old nodepool files
> - Adds state migration (`moved` blocks) and Make targets (`migrate-clusters-terraform`, env migration helper) for transition
> - Updates `iac/provider-gcp` Makefile to pass new TF vars; trims obsolete tfvar wiring
> - Wires new configs through `main.tf`; sets template manager update flag based on cluster sizes
> - Sets default hugepages in locals; preserves existing network/filestore integration and boot disk type vars
> - Minor: ensures gcloud ADC login in provider-login
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit daa7c6762c97dc290bf0c3714de561a04d2f9ba3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/iac/provider-gcp/Makefile
+++ b/iac/provider-gcp/Makefile
@@ -161,5 +161,11 @@ provider-login:
 migrate-clusters-terraform:
 	@ printf "Planning Terraform for env: `tput setaf 2``tput bold`$(ENV)`tput sgr0`\n\n"
 	@ $(TF) fmt -recursive
-	@ $(tf_vars) $(TF) plan -out=.tfplan.$(ENV) -compact-warnings -target=module.cluster.module.build_cluster -target=module.cluster.module.client_cluster  -target="module.cluster.google_compute_health_check.client_nomad_check" -target="module.cluster.google_compute_instance_template.client" -target="module.cluster.google_compute_region_instance_group_manager.client_pool"
+	@ $(tf_vars) $(TF) plan -out=.tfplan.$(ENV) -compact-warnings \
+		 -target="module.cluster.module.build_cluster" \
+		 -target="module.cluster.module.client_cluster" \
+		 -target="module.cluster.google_compute_health_check.client_nomad_check" \
+		 -target="module.cluster.google_compute_instance_template.client" \
+		 -target="module.cluster.google_compute_region_instance_group_manager.client_pool" \
+		 -target="module.cluster.google_compute_region_autoscaler.client"
 


iac/provider-gcp/variables.tf
```diff
@@ -20,34 +20,6 @@ variable "server_machine_type" {
...
-
-variable "client_cluster_autoscaling_cpu_target" {
-  description = "Target CPU utilization for client autoscaling (0.0-1.0)"
-  type        = number
-  default     = 0.6
-}
-
-variable "client_cluster_autoscaling_memory_target" {
-  # Note: This must be higher than orchestrator_base_hugepages_percentage (default 80%)
-  # because preallocated hugepages are counted as used memory in monitoring.
```
dobrac, line 41: This note should be still somewhere

iac/provider-gcp/Makefile
```diff
@@ -14,13 +14,7 @@ $(if $(value $(strip $(1))), TF_VAR_$(shell echo $(strip $(1)) | tr A-Z a-z)=$($
 endef
 
 tf_vars := 	TF_VAR_environment=$(TERRAFORM_ENVIRONMENT) \
-	$(call tfvar, CLIENT_MACHINE_TYPE) \
-	$(call tfvar, CLIENT_CLUSTER_SIZE) \
-	$(call tfvar, CLIENT_CLUSTER_SIZE_MAX) \
-	$(call tfvar, CLIENT_CLUSTER_AUTOSCALING_CPU_TARGET) \
-	$(call tfvar, CLIENT_CLUSTER_AUTOSCALING_MEMORY_TARGET) \
-	$(call tfvar, CLIENT_CLUSTER_CACHE_DISK_SIZE_GB) \
```
dobrac, line 23: will these still work when removed here?

.env.template
```diff
@@ -13,21 +13,11 @@ TERRAFORM_ENVIRONMENT=
...
-# Number of 375 GB NVME disks for client cache (default: 3)
-CLIENT_CLUSTER_CACHE_DISK_COUNT=
-# Target CPU utilization for autoscaling (0.0-1.0), e.g. 0.6
-CLIENT_CLUSTER_AUTOSCALING_CPU_TARGET=
-# Target memory utilization percentage for autoscaling (0-100), e.g. 85
-# Note: This must be higher than orchestrator_base_hugepages_percentage (default 80%)
-# because preallocated hugepages are counted as used memory in monitoring.
-CLIENT_CLUSTER_AUTOSCALING_MEMORY_TARGET=
+# Provide all client cluster configuration as a single JSON object
+# If provided, this overrides all individual CLIENT_* variables below
```
dobrac: is this line still true?

iac/provider-gcp/nomad-cluster/scripts/start-client.sh
```diff
@@ -43,6 +44,11 @@ mdadm --detail --scan --verbose | tee -a /etc/mdadm/mdadm.conf
 %{ else }
 DISK="/dev/disk/by-id/google-local-nvme-ssd-0-part1"
 %{ endif }
+%{ else }
+# Add cache disk for orchestrator and swapfile
+# TODO: Parametrize this
+DISK="/dev/disk/by-id/google-persistent-disk-1"
```
dobrac: where does this change come from?

iac/provider-gcp/nomad-cluster/worker-cluster/nodepool.tf
```diff
@@ -145,23 +146,38 @@ resource "google_compute_instance_template" "client" {
...
+  dynamic "disk" {
+    for_each = [for n in range(var.cache_disks.disk_type != "local-ssd" ? 1 : 0) : {}]
+    content {
+      auto_delete  = true
+      boot         = false
+      type         = "PERSISTENT"
+      disk_size_gb = var.cache_disks.size_gb
+      disk_type    = var.cache_disks.disk_type
+    }
+  }
```
dobrac, line 204: is this a new definition, or has it been defined elsewhere before?

iac/provider-gcp/variables.tf
```diff
@@ -502,23 +462,45 @@ variable "remote_repository_enabled" {
...
+      "disk_type": "pd-ssd",
+      "size_gb": 100
+    },
+    "cache_disk": {
+      "disk_type": "local-ssd",
+      "size_gb": 375,
+      "count": 3
+    }
+  }
+]
```
dobrac, line 474: is this the default? I'm also missing some comments with explanation what some of the fields do, how should they be configured

iac/provider-gcp/nomad-cluster/scripts/start-client.sh
jakubno: As this is not more configurable, this is needed for persistent disk to work. But could be probably in separate PR. I will simplify it

.env.template
```diff
@@ -13,21 +13,37 @@ TERRAFORM_ENVIRONMENT=
...
+#     },
+#     "cache_disks": {
+#         "disk_type": "local-ssd",  // Cache disk type (local-ssd or pd-ssd)
+#         "size_gb": 375,  // Cache disk size in GB
+#         "count": 1  // Number of cache disks
+#     }
+# }
+# Build cluster configuration
+# Example:
+# BUILD_CLUSTER_CONFIG_JSON='{"cluster_size": 1, "machine":{"type":"n1-standard-8","min_cpu_platform":"Intel Skylake"}, "autoscaler": {"size_min": 1, "size_max": 1, "memory_target": 100, "cpu_target": 0.7}, "boot_disk":{"disk_type":"pd-ssd","size_gb":100}, "cache_disks":{"disk_type":"local-ssd","size_gb":375,"count":1}}'
```
dobrac: where is the default configuration/values defined? If it should be this, can we put it directly as the env value or rather, in the terraform value default?

.env.template
```diff
@@ -13,21 +13,37 @@ TERRAFORM_ENVIRONMENT=
...
+BUILD_CLUSTER_CONFIG_JSON=
```
dobrac: why are there multiple client clusters but only one build cluster?

.env.template
jakubno: I wanted people to fill it in explicitly, otherwise it could lead to some breaking changes

.env.template
jakubno: I will convert it to support multiple of them

.env.template
dobrac: if you put it to the env example, it won't be breaking. We can then move it later to the terraform (or somewhere else)

jakubno: That's the idea, the example is there. Or did you mean putting it already as a value

dobrac (approved):

PR #1609, opened 2025-12-12
Add some safety when closing orchestrator and template manager

djeebus:
Also hide some "errors" that aren't really errors.

---

> [!NOTE]
> Adds context-driven shutdown across orchestrator/template-manager, stop-all for sandboxes/builds, 24h kill timeouts, context-aware pools/storage, and a new utils.Wait with tests.
> 
> - **Orchestrator / Shutdown & Signals**:
>   - Refactor main shutdown flow: context-based waiting, ignore benign errors, structured service start/stop, and parallel close steps.
>   - Replace `FORCE_STOP` usage with `Config.StopSandboxesOnExit` (`env:"FORCE_STOP"`) to optionally kill all sandboxes on exit.
>   - Add `sleepCtx`, helpers to ignore canceled/invalid-arg/service-done errors.
> - **Sandbox**:
>   - **Factory**: track active sandboxes via `sync.WaitGroup`; new `Wait(ctx)` to block until all exit; ensure decrement via cleanup.
>   - **Map**: add `StopAll(ctx)` to gracefully stop all sandboxes.
>   - **Cleanup**: introduce named cleaners with logging; prioritize and report failures.
>   - Reduce noisy logs and improve envd init retry behavior.
> - **Pools & Network**:
>   - Make NBD and network pool `Populate(ctx)` return errors and respect context/done channels; fix slot cleanup paths.
>   - Change network `Storage.Release` to `Release(ctx, *Slot)`; implement in KV/Local/Memory with context support.
> - **Template Manager**:
>   - Server now waits via `utils.Wait(ctx, &wg)` and logs clean shutdown; add `StopAllBuilds()`.
>   - **Build Cache**: add `StopAll()` to mark running builds as cancelled.
> - **gRPC/HTTP**:
>   - Simplify gRPC server interceptors (comment out logging), keep recovery/OTEL handlers.
>   - cmux/http/grpc servers: context-aware start/stop, consistent error handling.
> - **Infra / Config**:
>   - **Nomad jobs**: set `kill_timeout = "24h"` for `orchestrator` and `template-manager` tasks.
>   - `.editorconfig`: add `*.hcl` indent rules.
> - **Utilities & Tests**:
>   - New `utils.Wait(ctx, *sync.WaitGroup)` with comprehensive tests; sandbox factory wait test added.
> - **Scripts/Deps**:
>   - `packages/shared/scripts`: run `npm install`, add `dotenv`, bump `e2b` and related deps.
>   - `go.mod`: add `go.uber.org/atomic`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8d10e4a0888f462a043508c5a86a0d979a3ee77d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/orchestrator/internal/sandbox/sandbox.go
+++ b/packages/orchestrator/internal/sandbox/sandbox.go
@@ -208,7 +208,7 @@ func (f *Factory) CreateSandbox(
 		}
 	}()
 
-	cleanup.AddNoContext(ctx, func() error {
+	cleanup.AddNoContext(ctx, "reduce counter", func() error {
 		f.subtractSandbox()
 
 		return nil
@@ -221,7 +221,7 @@ func (f *Factory) CreateSandbox(
 	}()
 
 	sandboxFiles := template.Files().NewSandboxFiles(runtime.SandboxID)
-	cleanup.Add(ctx, cleanupFiles(f.config, sandboxFiles))
+	cleanup.Add(ctx, "clean up files", cleanupFiles(f.config, sandboxFiles))
 
 	rootFS, err := template.Rootfs()
 	if err != nil {
@@ -246,7 +246,7 @@ func (f *Factory) CreateSandbox(
 	if err != nil {
 		return nil, fmt.Errorf("failed to create rootfs overlay: %w", err)
 	}
-	cleanup.Add(ctx, rootfsProvider.Close)
+	cleanup.Add(ctx, "close rootfs provider", rootfsProvider.Close)
 	go func() {
 		runErr := rootfsProvider.Start(execCtx)
 		if runErr != nil {
@@ -344,7 +344,7 @@ func (f *Factory) CreateSandbox(
 	sbx.Checks = NewChecks(sbx, false)
 
 	// Stop the sandbox first if it is still running, otherwise do nothing
-	cleanup.AddPriority(ctx, sbx.Stop)
+	cleanup.AddPriority(ctx, "stop sandbox", sbx.Stop)
 
 	go func() {
 		defer execSpan.End()
@@ -401,7 +401,7 @@ func (f *Factory) ResumeSandbox(
 			execSpan.End()
 		}
 	}()
-	cleanup.AddNoContext(ctx, func() error {
+	cleanup.AddNoContext(ctx, "subtract sandbox", func() error {
 		f.subtractSandbox()
 
 		return nil
@@ -414,7 +414,7 @@ func (f *Factory) ResumeSandbox(
 	}()
 
 	sandboxFiles := t.Files().NewSandboxFiles(runtime.SandboxID)
-	cleanup.Add(ctx, cleanupFiles(f.config, sandboxFiles))
+	cleanup.Add(ctx, "clean up files", cleanupFiles(f.config, sandboxFiles))
 
 	telemetry.ReportEvent(ctx, "created sandbox files")
 
@@ -434,7 +434,7 @@ func (f *Factory) ResumeSandbox(
 		return nil, fmt.Errorf("failed to create rootfs overlay: %w", err)
 	}
 
-	cleanup.Add(ctx, rootfsOverlay.Close)
+	cleanup.Add(ctx, "close rootfs overlay", rootfsOverlay.Close)
 
 	telemetry.ReportEvent(ctx, "created rootfs overlay")
 
@@ -588,7 +588,7 @@ func (f *Factory) ResumeSandbox(
 	// This is to prevent race condition of reporting unhealthy sandbox
 	sbx.Checks = NewChecks(sbx, useClickhouseMetrics)
 
-	cleanup.AddPriority(ctx, func(ctx context.Context) error {
+	cleanup.AddPriority(ctx, "stop sandbox", func(ctx context.Context) error {
 		// Stop the sandbox first if it is still running, otherwise do nothing
 		return sbx.Stop(ctx)
 	})
@@ -775,7 +775,7 @@ func (s *Sandbox) Pause(
 	if err != nil {
 		return nil, fmt.Errorf("failed to get template files: %w", err)
 	}
-	cleanup.AddNoContext(ctx, snapshotTemplateFiles.Close)
+	cleanup.AddNoContext(ctx, "close snapshot template files", snapshotTemplateFiles.Close)
 
 	buildID, err := uuid.Parse(snapshotTemplateFiles.BuildID)
 	if err != nil {
@@ -799,7 +799,7 @@ func (s *Sandbox) Pause(
 
 	// Snapfile is not closed as it's returned and cached for later use (like resume)
 	snapfile := template.NewLocalFileLink(snapshotTemplateFiles.CacheSnapfilePath())
-	cleanup.AddNoContext(ctx, snapfile.Close)
+	cleanup.AddNoContext(ctx, "close snapfile", snapfile.Close)
 	// Memfile is also closed on diff creation processing
 	/* The process of snapshotting memory is as follows:
 	1. Pause FC via API
@@ -852,7 +852,7 @@ func (s *Sandbox) Pause(
 	if err != nil {
 		return nil, fmt.Errorf("error while post processing: %w", err)
 	}
-	cleanup.AddNoContext(ctx, memfileDiff.Close)
+	cleanup.AddNoContext(ctx, "close memfile diff", memfileDiff.Close)
 
 	rootfsDiff, rootfsDiffHeader, err := pauseProcessRootfs(
 		ctx,
@@ -867,10 +867,10 @@ func (s *Sandbox) Pause(
 	if err != nil {
 		return nil, fmt.Errorf("error while post processing: %w", err)
 	}
-	cleanup.AddNoContext(ctx, rootfsDiff.Close)
+	cleanup.AddNoContext(ctx, "close rootfs diff", rootfsDiff.Close)
 
 	metadataFileLink := template.NewLocalFileLink(snapshotTemplateFiles.CacheMetadataPath())
-	cleanup.AddNoContext(ctx, metadataFileLink.Close)
+	cleanup.AddNoContext(ctx, "close metadata file link", metadataFileLink.Close)
 
 	err = m.ToFile(metadataFileLink.Path())
 	if err != nil {
@@ -992,7 +992,7 @@ func getNetworkSlotAsync(
 			return
 		}
 
-		cleanup.Add(ctx, func(ctx context.Context) error {
+		cleanup.Add(ctx, "return network slot", func(ctx context.Context) error {
 			ctx, span := tracer.Start(ctx, "network-slot-clean")
 			defer span.End()
 
@@ -1035,7 +1035,7 @@ func serveMemory(
 
 	telemetry.ReportEvent(ctx, "started uffd")
 
-	cleanup.Add(ctx, func(ctx context.Context) error {
+	cleanup.Add(ctx, "stop uffd", func(ctx context.Context) error {
 		_, span := tracer.Start(ctx, "uffd-stop")
 		defer span.End()
 
--- a/packages/orchestrator/internal/template/server/main.go
+++ b/packages/orchestrator/internal/template/server/main.go
@@ -42,7 +42,7 @@ type ServerStore struct {
 	templateStorage   storage.StorageProvider
 	buildStorage      storage.StorageProvider
 
-	wg   *sync.WaitGroup // wait group for running builds
+	wg   sync.WaitGroup // wait group for running builds
 	info *service.ServiceInfo
 
 	closers []closeable
@@ -124,7 +124,6 @@ func New(
 		templateStorage:   templatePersistence,
 		buildStorage:      buildPersistance,
 		info:              info,
-		wg:                &sync.WaitGroup{},
 		closers:           closers,
 	}
 
@@ -148,7 +147,11 @@ func (s *ServerStore) Close(ctx context.Context) error {
 
 func (s *ServerStore) Wait(ctx context.Context) error {
 	s.logger.Info(ctx, "Waiting for all build jobs to finish")
-	defer s.logger.Info(ctx, "Template build queue cleaned")
+	defer s.logger.Info(ctx, "Template build queue is empty")
 
-	return utils.Wait(ctx, s.wg)
+	return utils.Wait(ctx, &s.wg)
+}
+
+func (s *ServerStore) StopAllBuilds() {
+	s.buildCache.StopAll()
 }
--- a/packages/orchestrator/main.go
+++ b/packages/orchestrator/main.go
@@ -16,7 +16,6 @@ import (
 
 	"github.com/google/uuid"
 	"github.com/soheilhy/cmux"
-	"go.uber.org/atomic"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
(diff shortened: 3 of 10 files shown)


packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -584,6 +612,7 @@ func (f *Factory) ResumeSandbox(
 	go sbx.Checks.Start(execCtx)
 
 	go func() {
+		defer f.subtractSandbox()
```
dobrac: the sbx cleanup might not be called at this point, so the graceful shutdown might be invalid

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -163,6 +165,17 @@ func NewFactory(
 	}
 }
 
+// Wait for all sandboxes to exit
+func (f *Factory) Wait(ctx context.Context) error {
+	l := logger.L()
+	l.Info(ctx, "Waiting for all sandboxes to exit")
+	defer l.Info(ctx, "All sandboxes exited")
+
+	f.wg.Wait()
```
dobrac: force restart implementation is missing

packages/orchestrator/internal/template/server/main.go
```diff
@@ -153,20 +152,13 @@ func (s *ServerStore) Close(ctx context.Context) error {
 }
 
 func (s *ServerStore) Wait(ctx context.Context) error {
-	select {
-	case <-ctx.Done():
```
dobrac, line 157: some kind of force restart implementation should be still there, at least for development

packages/orchestrator/main.go
```diff
@@ -544,7 +548,7 @@ func run(config cfg.Config) (success bool) {
 		logger.L().Error(ctx, "Service error", zap.Error(serviceErr))
 	}
 
-	closeCtx, cancelCloseCtx := context.WithCancel(context.Background())
+	closeCtx, cancelCloseCtx := context.WithTimeout(context.Background(), 24*time.Hour)
```
dobrac: the 24 hours should be probably somewhere else. Also, we don't want to handle max sandbox lifetime from the orchestrator

djeebus: - "somewhere else"? I'm not sure what you mean here.
- "max sandbox lifetime". How does the orchestrator know how long we need to wait before exiting? Do we just want to wait for the sandbox to quit gracefully, no matter how long that takes?

packages/orchestrator/main.go
```diff
@@ -555,10 +559,12 @@ func run(config cfg.Config) (success bool) {
 	if serviceInfo.GetStatus() == orchestratorinfo.ServiceInfoStatus_Healthy {
 		serviceInfo.SetStatus(ctx, orchestratorinfo.ServiceInfoStatus_Draining)
 
-		// Wait for draining state to propagate to all consumers
-		if !env.IsLocal() {
-			time.Sleep(15 * time.Second)
-		}
+		time.Sleep(15 * time.Second)
```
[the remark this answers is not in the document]
djeebus: Local development still requires some kind of sleep when quitting to ensure that deregistration happens.

packages/orchestrator/internal/template/server/main.go
```diff
@@ -153,20 +152,17 @@ func (s *ServerStore) Close(ctx context.Context) error {
...
-		if !env.IsLocal() {
-			s.logger.Info(ctx, "Waiting for consumers to check build status")
-			time.Sleep(15 * time.Second)
-		}
+	s.logger.Info(ctx, "Waiting for all build jobs to finish")
+	s.wg.Wait()
 
-		s.logger.Info(ctx, "Template build queue cleaned")
+	s.logger.Info(ctx, "Waiting for consumers to check build status")
+	time.Sleep(15 * time.Second)
```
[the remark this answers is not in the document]
djeebus: Local development still requires some kind of sleep when quitting to ensure that clean up happens.

packages/orchestrator/internal/sandbox/sandbox.go
djeebus: I'm not sure what you mean here. "cleanup"? Can you say more?

packages/orchestrator/main.go
```diff
@@ -538,37 +545,65 @@ func run(config cfg.Config) (success bool) {
...
+			err := tmpl.Wait(closeCtx)
+			if err != nil {
+				logger.L().Error(ctx, "error while waiting for template manager to drain", zap.Error(err))
+				success = false
+			}
+
+			logger.L().Info(ctx, "Waiting for consumers to check build status")
+			sleepCtx(closeCtx, 15*time.Second)
+		}
+	})
```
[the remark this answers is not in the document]
djeebus: The end result is the same; success can only be set to false, and it doesn't matter which goroutine does it.

packages/orchestrator/internal/sandbox/network/pool.go
```diff
@@ -124,7 +124,12 @@ func (p *Pool) Populate(ctx context.Context) {
 			}
 
 			newSlotsAvailableCounter.Add(ctx, 1)
-			p.newSlots <- slot
+
+			select {
+			case p.newSlots <- slot:
+			case <-ctx.Done():
+				return ctx.Err()
+			}
```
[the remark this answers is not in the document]
djeebus: This would be a force-quit situation, don't think we're worried about resources left around.

packages/orchestrator/internal/sandbox/sandbox.go
dobrac: there is a `cleanup` field on the sandbox, which is triggered on `sbx.Close` method call

packages/orchestrator/main.go
dobrac: > "somewhere else"? I'm not sure what you mean here.

there is (atm) a max time you wait for a kill in Nomad, so I suppose there should be "one" source of truth of this interval. Also in general, the max 24 hour limit is enforced from the API, so having it also on the orchestrator level might be forgotten when we do y change in the future, e.g. from 24 to 48.

> "max sandbox lifetime". How does the orchestrator know how long we need to wait before exiting? Do we just want to wait for the sandbox to quit gracefully, no matter how long that takes?

This is quite tricky, but if the API works correctly, then maybe yes=`no matter how long that takes`? This is most probably also dependent on the Nomad value, which is now afaik set to 24h max

djeebus: Got it, that all makes sense to me. I've removed the time limits, mind taking another look when you get a chance?

packages/orchestrator/internal/sandbox/network/pool.go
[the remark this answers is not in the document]
dobrac: can we clean that one slot regardless?

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -163,6 +165,32 @@ func NewFactory(
...
+		return ctx.Err()
+	}
+
+	l.Info(ctx, "Waiting for all sandboxes to exit")
+	defer l.Info(ctx, "All sandboxes exited")
+
+	done := make(chan struct{}, 1)
+
+	go func() {
+		f.wg.Wait()
```
dobrac: we should block allowing new sandboxes to be started at this point, otherwise the wait group could fail

packages/orchestrator/main.go
```diff
@@ -259,6 +260,8 @@ func run(config cfg.Config) (success bool) {
 				// or if the error is already sent
 			}
 
+			l.Info(ctx, "service stopped")
```
dobrac, line 263: nit: can we say which service stopped?

packages/orchestrator/internal/template/server/main.go
```diff
@@ -135,38 +133,26 @@ func New(
...
-			s.logger.Info(ctx, "Waiting for consumers to check build status")
-			time.Sleep(15 * time.Second)
-		}
+	if err := ctx.Err(); err != nil {
+		return err
+	}
 
-		s.logger.Info(ctx, "Template build queue cleaned")
+	s.logger.Info(ctx, "Waiting for all build jobs to finish")
+	s.wg.Wait()
```
dobrac: this ignores the context cancelation after the wait has been executed (we're not using it that way rn, but e.g. the sandboxes will exit)

(I know it didn't before)

packages/orchestrator/internal/template/server/main.go
```diff
@@ -135,38 +133,26 @@ func New(
...
-	select {
-	case <-ctx.Done():
-		return errors.New("force exit, not waiting for builds to finish")
-	default:
-		s.logger.Info(ctx, "Waiting for all build jobs to finish")
-		s.wg.Wait()
-
-		if !env.IsLocal() {
-			s.logger.Info(ctx, "Waiting for consumers to check build status")
-			time.Sleep(15 * time.Second)
```
dobrac, line 165: this looks to be missing

packages/orchestrator/main.go
```diff
@@ -563,31 +566,59 @@ func run(config cfg.Config) (success bool) {
...
-		if !env.IsLocal() {
-			time.Sleep(15 * time.Second)
+	closeWg.Go(func() {
+		// Mark service draining if not already.
+		// If service stats was previously changed via API, we don't want to override it.
+		if serviceInfo.GetStatus() == orchestratorinfo.ServiceInfoStatus_Healthy {
+			serviceInfo.SetStatus(ctx, orchestratorinfo.ServiceInfoStatus_Draining)
+
+			logger.L().Info(ctx, "Waiting for api to read orchestrator status")
+			sleepCtx(closeCtx, 15*time.Second)
```
dobrac: you need to wait before you await in the sandbox factory, as new sandboxes can be still scheduled in that time (same for templates)

packages/orchestrator/internal/sandbox/sandbox.go
djeebus: The way this is implemented, the caller sets the status to Draining before calling Wait, which *should* prevent new sandboxes from being scheduled. I don't think we want to block sandboxes from being scheduled in multiple places, for the same reason we don't want to set a max time limit in multiple places - if implementation changes, we don't want to hunt down all the places we blocked, closed, etc.

packages/orchestrator/main.go
djeebus: Line 248 adds `zap.String("server", name")` to the logger, which this line inherits.

packages/orchestrator/internal/sandbox/sandbox.go
dobrac: I think it might be worth it here though as it is a limitation of using the WaitGroup - you can't add new tasks to the WaitGroup after you do Wait, otherwise it may panic

iac/provider-gcp/nomad/jobs/template-manager.hcl
dobrac, line 77: should we also remove this if the kill_timeout is now 24h?

packages/orchestrator/internal/sandbox/network/pool.go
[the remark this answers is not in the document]
dobrac: if the context is done, the slot is still nto cleared, correct?

packages/orchestrator/main.go
```diff
@@ -563,31 +566,59 @@ func run(config cfg.Config) (success bool) {
...
 
-	// Mark service draining if not already.
-	// If service stats was previously changed via API, we don't want to override it.
-	if serviceInfo.GetStatus() == orchestratorinfo.ServiceInfoStatus_Healthy {
-		serviceInfo.SetStatus(ctx, orchestratorinfo.ServiceInfoStatus_Draining)
+	// if a signal is received again, cancel the context and force quit
+	go func() {
+		<-sigs
+		logger.L().Info(ctx, "Force shutdown signal received")
+		cancel()
```
dobrac: do we need to cancel this context?

packages/orchestrator/internal/sandbox/sandbox.go
djeebus: I don't think it panics in that scenario. I set this playground up, let me know if it looks like what you expected: https://go.dev/play/p/bDPHf1ejwGs

packages/orchestrator/main.go
djeebus: Assuming that a) the window is quite small, and b) the sync.WaitGroup doesn't panic in that scenario (see reply to other comment), is this still worth complicating the solution to prevent?

iac/provider-gcp/nomad/jobs/template-manager.hcl
djeebus: I think that's there to prevent single-node clusters from having to wait too long to restart their allocation. I'm not totally sure. They don't seem to server an important purpose anymore, but I'd like input from @jakubno and @ValentaTomas before we remove them.

packages/orchestrator/main.go
djeebus: (for both this and the previous comment) My theory on "force shutdown" was that, if we were in that scenario, we want everything to bail immediately - A dirty shutdown that leaves files and processes is exactly what we want, and cancelling the context gets us what we want. It effectively simulates setting FORCE_QUIT=true at runtime.

packages/orchestrator/main.go
```diff
@@ -563,31 +566,59 @@ func run(config cfg.Config) (success bool) {
...
+	closeWg.Go(func() {
+		if err := sandboxFactory.Wait(closeCtx); err != nil {
+			logger.L().Error(ctx, "error while waiting for sandbox factory to drain", zap.Error(err))
 			success = false
 		}
-	}
+
+		logger.L().Info(ctx, "Waiting for sandbox clients to close connections")
+		sleepCtx(closeCtx, 15*time.Second)
+	})
```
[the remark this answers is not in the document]
djeebus: This is exactly what we want. Since we're likely to be waiting for quite a while for the orchestrator to shut down, one more sandbox isn't a big deal. It won't deadlock, since the new sandbox will add/subtract to the counter exactly the same way that the rest of them will.

packages/orchestrator/main.go
dobrac: The initial idea behind adding FORCE_QUIT was to not wait for the sbxs/template builds when developing locally, but still clean everything cleanly

packages/orchestrator/internal/sandbox/sandbox.go
dobrac: It's not that easy to simulate, but we've hit this already few times. Here are the links to the source code:
- https://github.com/golang/go/blob/ad91f5d241f3b8e85dc866d0087c3f13af96ef33/src/sync/waitgroup.go#L121C3-L121C69
- https://github.com/golang/go/blob/ad91f5d241f3b8e85dc866d0087c3f13af96ef33/src/sync/waitgroup.go#L213

Here is the PR:
- https://github.com/e2b-dev/infra/pull/1341

packages/orchestrator/main.go
dobrac: if the b) is true, then you need to be sure at least that the 15 seconds pass before the await for both template and sbxs exists

packages/orchestrator/main.go
djeebus: oooh. definitely not my understanding, I'll clean that up

ValentaTomas:
Can we fix the lint error + resolve the bot convos, so it is clear that this can be reviewed again?

PR #1614, opened 2025-12-12
Simplify the int feature flag method

djeebus:
> [!NOTE]
> Replaces IntFlag’s (int, error) API with a non-error int-returning version, updates all callers, and adjusts/renames related flags and defaults (e.g., Clickhouse batcher, Envd timeout).
> 
> - **Feature Flags**:
>   - **API change**: `Client.IntFlag` now returns `int` (no error); logs on failure and uses fallback.
>   - **Defaults/Names**: Tweaks Clickhouse batcher defaults (`ClickhouseBatcherMaxBatchSize=100`, `MaxDelay=1000ms`, `QueueSize=1000`); renames `EnvdInitTimeoutSeconds` to `EnvdInitTimeoutMilliseconds`.
> - **Call Site Updates**:
>   - Orchestrator: `getBestOfKConfig` and sandbox limit checks use `IntFlag` without error handling.
>   - Clickhouse events: batcher options (`MaxBatchSize`, `MaxDelay`, `QueueSize`) sourced via new `IntFlag`.
>   - Orchestrator sandbox: `GetEnvdInitRequestTimeout` reads `EnvdInitTimeoutMilliseconds` and converts to `time.Millisecond`.
>   - Build/Template:
>     - Build cache eviction reads `BuildCacheMaxUsagePercentage` without error.
>     - Provision hashing compares `BuildProvisionVersion` to `Fallback()` without error handling.
>   - Limits: GCloud max tasks and concurrent upload limit use new `IntFlag`; periodic semaphore update simplified.
> - **Logging**:
>   - Adds warnings when LaunchDarkly is uninitialized or flag evaluation fails in `IntFlag`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d10728ecae6db547a72acf100384f0fa0dd724b6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/orchestrator/internal/sandbox/sandbox.go
+++ b/packages/orchestrator/internal/sandbox/sandbox.go
@@ -1085,7 +1085,7 @@ func (s *Sandbox) WaitForEnvd(
 }
 
 func (f *Factory) GetEnvdInitRequestTimeout(ctx context.Context) time.Duration {
-	envdInitRequestTimeoutMs := f.featureFlags.IntFlag(ctx, featureflags.EnvdInitTimeoutSeconds)
+	envdInitRequestTimeoutMs := f.featureFlags.IntFlag(ctx, featureflags.EnvdInitTimeoutMilliseconds)
 
 	return time.Duration(envdInitRequestTimeoutMs) * time.Millisecond
 }
--- a/packages/orchestrator/internal/template/build/phases/base/hash.go
+++ b/packages/orchestrator/internal/template/build/phases/base/hash.go
@@ -38,7 +38,7 @@ func (bb *BaseBuilder) Hash(ctx context.Context, _ phases.LayerResult) (string, 
 		featureflags.TemplateContext(bb.Config.TemplateID),
 		featureflags.TeamContext(bb.Config.TeamID),
 		// for dev environments (fallback value), use the provision script hash
-	); val != 0 {
+	); val != featureflags.BuildProvisionVersion.Fallback() {
 		provisionVersion = strconv.FormatInt(int64(val), 10)
 	}
 
--- a/packages/shared/pkg/feature-flags/flags.go
+++ b/packages/shared/pkg/feature-flags/flags.go
@@ -111,7 +111,7 @@ var (
 	BestOfKMaxOvercommit          = newIntFlag("best-of-k-max-overcommit", 400)              // Default R=4 (stored as percentage, max over-commit ratio)
 	BestOfKAlpha                  = newIntFlag("best-of-k-alpha", 50)                        // Default Alpha=0.5 (stored as percentage for int flag, current usage weight)
 	PubsubQueueChannelSize        = newIntFlag("pubsub-queue-channel-size", 8*1024)          // size of the channel buffer used to queue incoming sandbox events
-	EnvdInitTimeoutSeconds        = newIntFlag("envd-init-request-timeout-milliseconds", 50) // Timeout for envd init request in milliseconds
+	EnvdInitTimeoutMilliseconds   = newIntFlag("envd-init-request-timeout-milliseconds", 50) // Timeout for envd init request in milliseconds
 
 	// BuildCacheMaxUsagePercentage the maximum percentage of the cache disk storage
 	// that can be used before the cache starts evicting items.
--- a/packages/orchestrator/internal/sandbox/build/cache.go
+++ b/packages/orchestrator/internal/sandbox/build/cache.go
@@ -165,7 +165,7 @@ func (s *DiffStore) startDiskSpaceEviction(
 			used := int64(dUsed) - pUsed
 			percentage := float64(used) / float64(dTotal) * 100
 
-			threshold := 100 // max value, won't ever be used
+			threshold := featureflags.BuildCacheMaxUsagePercentage.Fallback()
 			// When multiple services (template manager, orchestrator) are defined, take the lowest threshold
 			// to ensure we don't exceed any of the set limits
 			for _, s := range services {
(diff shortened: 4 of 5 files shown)


packages/shared/pkg/feature-flags/flags.go
```diff
@@ -104,9 +104,9 @@ var (
 	MaxSandboxesPerNode           = newIntFlag("max-sandboxes-per-node", 200)
 	GcloudConcurrentUploadLimit   = newIntFlag("gcloud-concurrent-upload-limit", 8)
 	GcloudMaxTasks                = newIntFlag("gcloud-max-tasks", 16)
-	ClickhouseBatcherMaxBatchSize = newIntFlag("clickhouse-batcher-max-batch-size", 64*1024) // 65536
-	ClickhouseBatcherMaxDelay     = newIntFlag("clickhouse-batcher-max-delay", 100)          // 100ms in milliseconds
-	ClickhouseBatcherQueueSize    = newIntFlag("clickhouse-batcher-queue-size", 8*1024)      // 8192
+	ClickhouseBatcherMaxBatchSize = newIntFlag("clickhouse-batcher-max-batch-size", 100)
```
ValentaTomas, line 107: Why are the values changed?

packages/orchestrator/internal/template/build/phases/base/hash.go
```diff
@@ -32,13 +32,13 @@ func (bb *BaseBuilder) Hash(ctx context.Context, _ phases.LayerResult) (string,
...
-	if val, err := bb.featureFlags.IntFlag(
+	if val := bb.featureFlags.IntFlag(
 		ctx,
 		featureflags.BuildProvisionVersion,
 		featureflags.TemplateContext(bb.Config.TemplateID),
 		featureflags.TeamContext(bb.Config.TeamID),
-	// for dev environments (fallback value), use the provision script hash
-	); val != featureflags.BuildProvisionVersion.Fallback() && err == nil {
+		// for dev environments (fallback value), use the provision script hash
+	); val != 0 {
```
ValentaTomas: Is it ok to suddenly not check agains fallback?

packages/shared/pkg/feature-flags/flags.go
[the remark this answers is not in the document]
djeebus: This is how the code worked; if it returned an error (ie, it used the fallback) then it would change the value to 100. Strange, but consistent.

djeebus: If you look at the code, we always overwrote fallback values with these other values. This makes the other values the real fallbacks, this change just puts them where they should be.

packages/orchestrator/internal/template/build/phases/base/hash.go
djeebus: The end result was the same, but the bot did point out that now it's implicit, and explicit is better, so I've reverted this.

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -1085,10 +1085,7 @@ func (s *Sandbox) WaitForEnvd(
 }
 
 func (f *Factory) GetEnvdInitRequestTimeout(ctx context.Context) time.Duration {
-	envdInitRequestTimeoutMs, err := f.featureFlags.IntFlag(ctx, featureflags.EnvdInitTimeoutSeconds)
-	if err != nil {
-		logger.L().Warn(ctx, "failed to get envd timeout from feature flag, using default", zap.Error(err))
-	}
+	envdInitRequestTimeoutMs := f.featureFlags.IntFlag(ctx, featureflags.EnvdInitTimeoutSeconds)
```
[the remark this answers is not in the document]
djeebus: The feature flag was just poorly named, I've changed it.

packages/shared/pkg/feature-flags/flags.go
ValentaTomas: Can you please point where the overwrite happened? I somehow cannot find it.

djeebus: Sure, can do!
- [ClickhouseBatcherMaxBatchSize](https://github.com/e2b-dev/infra/blob/4d5d712bab72e899b0d4dbf67c52a5eb42675f86/packages/clickhouse/pkg/events/delivery.go#L51-L54)
- [ClickhouseBatcherMaxDelay](https://github.com/e2b-dev/infra/blob/4d5d712bab72e899b0d4dbf67c52a5eb42675f86/packages/clickhouse/pkg/events/delivery.go#L56-L59)
- [ClickhouseBatcherQueueSize](https://github.com/e2b-dev/infra/blob/4d5d712bab72e899b0d4dbf67c52a5eb42675f86/packages/clickhouse/pkg/events/delivery.go#L61-L64)

packages/orchestrator/internal/template/build/phases/base/hash.go
```diff
@@ -32,13 +32,13 @@ func (bb *BaseBuilder) Hash(ctx context.Context, _ phases.LayerResult) (string,
 	}
 
 	provisionVersion := provisionScriptFile
-	if val, err := bb.featureFlags.IntFlag(
```
sitole, line 35: Let's drop inline if here and assign provisionVersion in a separate if, just checking val is not a fallback.

djeebus: The difficulty here is in making it clean. The end result that we actually want is a string, not a number. So we end up with something like this (in pseudocode):

```golang
var text string
number := flags.IntFlag(myFlag)
if number == myFlag.Fallback() {
	text = strconv.FormatInt(int64(number), 10)
} else {
	text = provisionScriptFile
}
```

I'm not sure this is more understandable than:

```golang
text := provisionScriptFile
if number := flags.IntFlag(myFlag); number != myFlag.Fallback() {
	text = strconv.FormatInt(int64(number), 10)
}
```

wdyt?

sitole: Let's stick with inline if version. Makes sense.

PR #1625, opened 2025-12-17
Make edge api deployment optional

sitole:
- Edge API is used for remote clusters only.
- Bring back Loki sbx and build logs directly to api so we don't need to rely on edge.
- Try to share as much as possible across api and edge api that share sbx/build logs and sbx metrics query functions.
- Move local nodes discovery to the cluster so we have a unified interface.
- To simplify the codebase, we now have a gRPC connection for each node; before, cluster nodes shared just one.
- Now we don't need to manually get the call context for cluster gRPC calls. Metadata are stored in the connection itself.


iac/provider-gcp/nomad/jobs/api.hcl
```diff
@@ -96,9 +97,6 @@ job "api" {
         REDIS_TLS_CA_BASE64            = "${redis_tls_ca_base64}"
         SANDBOX_ACCESS_TOKEN_HASH_SEED = "${sandbox_access_token_hash_seed}"
 
-        LOCAL_CLUSTER_ENDPOINT = "${local_cluster_endpoint}"
-        LOCAL_CLUSTER_TOKEN    = "${local_cluster_token}"
```
sitole, line 100: It's not needed as we are not using the edge interface for the local cluster.

packages/api/internal/analytics_collector/client.go
```diff
@@ -19,39 +19,35 @@ type Analytics struct {
 }
 
 func NewAnalytics(ctx context.Context, host, grpcAPIKey string) (*Analytics, error) {
-	var client AnalyticsCollectorClient
```
sitole, line 22: This fixes the issue where, when the host is empty, the client is not nil and methods actually try to call implemented functions.

packages/api/internal/orchestrator/nodemanager/metadata.go
```diff
@@ -33,44 +33,49 @@ func (n *Node) Metadata() NodeMetadata {
...
+}
+
+func appendMetadataCtx(ctx context.Context, md metadata.MD) context.Context {
+	args := make([]string, 0, len(md)*2)
+	for k, v := range md {
+		args = append(args, k, v[0])
+	}
 
-	return metadata.NewOutgoingContext(ctx, metadata.Join(n.getClientMetadata(), md))
+	return metadata.AppendToOutgoingContext(ctx, args...)
```
sitole, line 80: We are now using `AppendToOutgoingContext` method that only appends metadata, the previous `NewOutgoingContext` cleared metadata set during connection creation (auth and service instance id).

packages/api/internal/orchestrator/nodemanager/node.go
```diff
@@ -164,27 +96,16 @@ func NewClusterNode(ctx context.Context, client *grpclient.GRPCClient, clusterID
...
-		logger.L().Info(ctx, "Closing local node", logger.WithNodeID(n.ID))
-		err := n.client.Close()
-		if err != nil {
-			logger.L().Error(ctx, "Error closing connection to node", zap.Error(err), logger.WithNodeID(n.ID))
-		}
-	} else {
-		logger.L().Info(ctx, "Closing cluster node", logger.WithNodeID(n.ID), logger.WithClusterID(n.ClusterID))
-		// We are not closing grpc connection, because it is shared between all cluster nodes, and it's handled by the cluster
-	}
+func (n *Node) Close(_ context.Context) error {
```
sitole, line 95: Connection is now closed by the cluster instances pool on removal. It's closed in the pool, even though the node is no longer in the orchestrator pool; the connection should still be available to the cluster pool as a parent.

packages/shared/pkg/sandbox-catalog/catalog.go
```diff
@@ -9,17 +9,18 @@ import (
 )
 
 type SandboxInfo struct {
-	OrchestratorID string `json:"orchestrator_id"`
-	OrchestratorIP string `json:"orchestrator_ip"` // used only for cases where orchestrator is not registered in edge pool
-	ExecutionID    string `json:"execution_id"`
+	OrchestratorID        string `json:"orchestrator_id"`
+	OrchestratorIP        string `json:"orchestrator_ip"` // used only for cases where orchestrator is not registered in edge pool
+	OrchestratorProxyPort uint16 `json:"orchestrator_proxy_port"`
```
sitole: We are now forwarding the orchestrator proxy port. This prepares the orchestrator to run on different ports, as planned for multiple orchestrators on a single node.

packages/shared/pkg/logs/loki/provider.go
```diff
@@ -10,22 +10,20 @@ import (
...
-	"github.com/e2b-dev/infra/packages/shared/pkg/logs/logsloki"
 	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
 )
 
 type LokiQueryProvider struct {
 	client *loki.DefaultClient
 }
 
-func NewLokiQueryProvider(config cfg.Config) (*LokiQueryProvider, error) {
+func NewLokiQueryProvider(lokiURL string, lokiUser string, lokiPassword string) (*LokiQueryProvider, error) {
```
sitole, line 22: This is out of edge as it's now shared between edge and core api for builds and sbx logs query.

packages/shared/pkg/feature-flags/flags.go
```diff
@@ -70,13 +70,12 @@ func newBoolFlag(name string, fallback bool) BoolFlag {
 }
 
 var (
-	MetricsWriteFlagName               = newBoolFlag("sandbox-metrics-write", env.IsDevelopment())
-	MetricsReadFlagName                = newBoolFlag("sandbox-metrics-read", env.IsDevelopment())
-	SnapshotFeatureFlagName            = newBoolFlag("use-nfs-for-snapshots", env.IsDevelopment())
-	TemplateFeatureFlagName            = newBoolFlag("use-nfs-for-templates", env.IsDevelopment())
-	BestOfKCanFit                      = newBoolFlag("best-of-k-can-fit", true)
-	BestOfKTooManyStarting             = newBoolFlag("best-of-k-too-many-starting", false)
-	EdgeProvidedSandboxMetricsFlagName = newBoolFlag("edge-provided-sandbox-metrics", false)
```
sitole, line 79: Removed `EdgeProvidedSandboxMetricsFlagName`. The following changes do not require an updated edge api in our infrastructure, as it's used only for remote one.

packages/shared/pkg/clusters/discovery/nomad.go
```diff
@@ -0,0 +1,74 @@
...
+
+var filter = fmt.Sprintf(
+	"ClientStatus == \"running\" and ((TaskGroup == \"%s\" and JobID contains \"%s\") or (TaskGroup == \"%s\" and JobID contains \"%s\"))",
+	templateManagersTaskGroup,
+	templateManagerJobPrefix,
+	orchestratorsTaskGroup,
+	orchestratorJobPrefix,
+)
+
+func ListOrchestratorAndTemplateBuilderAllocations(ctx context.Context, client *nomadapi.Client) ([]Allocation, error) {
```
sitole, line 35: Nomad service discovery now uses allocations (for API 1, we previously used node discovery). This is preparation for the deployment of multiple orchestrators on one node - we can take a port for each allocation separately.

sitole: This discovery is not done in a way that it's shared for the edge api and the core api.

packages/db/queries/health.sql
```diff
@@ -0,0 +1,2 @@
+-- name: Health :one
```
sitole: Simple query that pings the database and verifies the pool of connections is actually ready.

packages/client-proxy/internal/proxy/proxy.go
```diff
@@ -80,22 +77,29 @@ func NewClientProxy(meterProvider metric.MeterProvider, serviceName string, port
...
+			// https://linear.app/e2b/issue/ENG-3398
+			proxyPort := entry.OrchestratorProxyPort
+			if proxyPort == 0 {
+				proxyPort = consts.OrchestratorProxyPort
+			}
+
+			proxyUrl := &url.URL{
 				Scheme: "http",
-				Host:   fmt.Sprintf("%s:%d", nodeIP, orchestratorProxyPort),
+				Host:   fmt.Sprintf("%s:%d", entry.OrchestratorIP, proxyPort),
```
sitole: Client proxy is now ready to proxy traffic to the orch running on non-standard ports.

packages/client-proxy/internal/edge/handlers/sandbox-metrics.go
```diff
@@ -67,38 +65,3 @@ func (a *APIStore) V1SandboxMetrics(c *gin.Context, sandboxID string, params api
 
 	c.JSON(http.StatusOK, apiMetrics)
 }
-
-func getSandboxStartEndTime(ctx context.Context, a *APIStore, sandboxID string, params api.V1SandboxMetricsParams) (time.Time, time.Time, error) {
```
sitole, line 71: Moved shared clickhouse sbx metrics utils to its own repo, as it's now used in two services.

packages/clickhouse/pkg/utils/sandbox.go
```diff
@@ -0,0 +1,47 @@
...
+	"fmt"
+	"time"
+
+	"go.uber.org/zap"
+
+	clickhouse "github.com/e2b-dev/infra/packages/clickhouse/pkg"
+	"github.com/e2b-dev/infra/packages/shared/pkg/logger"
+)
+
+func GetSandboxStartEndTime(ctx context.Context, clickhouseStore clickhouse.SandboxQueriesProvider, teamID, sandboxID string, qStart *int64, qEnd *int64) (time.Time, time.Time, error) {
```
sitole, line 14: Moved as it's now used in two services.

packages/api/internal/orchestrator/lifecycle.go
```diff
@@ -22,17 +22,23 @@ func (o *Orchestrator) addToNode(ctx context.Context, sandbox sandbox.Sandbox, _
 	} else {
 		node.AddSandbox(sandbox)
 
+		// Add routing record to only for local nodes
+		if !node.IsLocal() {
+			return
+		}
```
sitole, line 28: Store routing catalog record only for nodes that runs in our cluster

PR #1629, opened 2025-12-18
Execute Postgres query before using pg pool to limit wait time later

sitole:
An easy solution that uses an already-created but not yet initialized sqlcdb client for fast queries.
This forces the client to initialize the requested pool of connections before they are actually used for the first time.

This should fix the issue where, sometimes, the pool of connections takes too long to create, and some initial background job requests fail because the context times out. Ideally, we should force a warming pool during client creation, but I didn't see an option for it.


sitole:
For future context, it looks like it's caused by a stuck TCP connection hanging for too long.
image

PR #1636, opened 2025-12-20
Implement warm pool base struct

djeebus:
Once this looks good, we should use it for the nbd device pool as well.

---

> [!NOTE]
> Introduces a reusable, telemetry‑instrumented pooling primitive and adopts it for network slots in the orchestrator.
> 
> - **Add `utils.WarmPool`** with OTEL metrics, lifecycle handling, and tests; provide `Item`/`ItemFactory` interfaces and generated mocks
> - **Refactor orchestrator network pool** to `utils.WarmPool` (`network.Pool` now wraps a warm pool); add `network.Config` and `Slot.String()`
> - **Config/infra:** expose pool sizing via `NETWORK_SLOTS_REUSE_POOL_SIZE` and `NETWORK_SLOTS_FRESH_POOL_SIZE`; set these in Nomad jobs (`orchestrator.hcl`, `template-manager.hcl`)
> - **Call site updates:** switch constructors in benchmark and build-template; main orchestrator uses new pool
> - **Tooling:** add `scripts/fix-meters.sh` and improve `fix-tracers.sh`; run both in `make fmt`; CI workflow now relies on `make fmt` (drops explicit tracer fix step)
> - Minor go.mod/sum updates across packages to support new telemetry/logging deps
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9d960af8774ba23fccba914880f592dc10647c83. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/shared/pkg/utils/warmpool.go
```diff
@@ -0,0 +1,297 @@
...
+	}
+}
+
+// Return an item to the pool.
+// - puts the item in `reusableItems` to get priority when being returned.
+// - destroy the item and return if the pool has already been closed.
+// - if the item fails to be pushed back into the channel for any reason, destroy it.
+func (wp *WarmPool[T]) Return(ctx context.Context, item T) {
+	wp.wg.Go(func() {
+		if err := wp.isClosed(ctx); err != nil {
```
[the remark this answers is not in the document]
djeebus: That's why we have the select statement, to avoid the panic. I don't think this is a real thing.

packages/shared/pkg/utils/warmpool.go
```diff
@@ -0,0 +1,297 @@
...
+		return s, nil
+	}
+}
+
+// Return an item to the pool.
+// - puts the item in `reusableItems` to get priority when being returned.
+// - destroy the item and return if the pool has already been closed.
+// - if the item fails to be pushed back into the channel for any reason, destroy it.
+func (wp *WarmPool[T]) Return(ctx context.Context, item T) {
+	wp.wg.Go(func() {
```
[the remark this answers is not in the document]
djeebus: It does. Poor claude, living in ... early 2024

packages/shared/pkg/utils/warmpool.go
```diff
@@ -0,0 +1,297 @@
...
+		if err := wp.isClosed(ctx); err != nil {
+			wp.returnedCounter.Add(ctx, 1, withFailureAttr())
```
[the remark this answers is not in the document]
djeebus: Try using a more current version of golang.

packages/orchestrator/internal/sandbox/nbd/pool.go
```diff
@@ -27,8 +27,9 @@ const (
 	devicePoolCloseReleaseTimeout = 10 * time.Minute
 )
 
+var meter = otel.Meter("github.com/e2b-dev/infra/packages/orchestrator/internal/sandbox/nbd")
```
jakubno: why?

scripts/fix-meters.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+
+# Script to find and replace otel.Tracer declarations with file paths
+# Usage: ./replace_tracer.sh [directory]
```
jakubno:
```suggestion
# Usage: ./replace_meter.sh [directory]
```

scripts/fix-meters.sh
```diff
@@ -0,0 +1,47 @@
...
+SEARCH_DIR="packages/"
+
+# Counter for processed files
+processed=0
+skipped=0
+
+readonly prefix="github.com/e2b-dev/infra/packages"
+readonly search_pattern='var meter = otel\.Meter(.*'
+
+echo "Starting tracer replacement in directory: $SEARCH_DIR"
```
jakubno:
```suggestion
echo "Starting meter replacement in directory: $SEARCH_DIR"
```

scripts/fix-meters.sh
jakubno, line 1: shouldn't this be in linting workflow?

packages/shared/pkg/utils/warmpool.go
```diff
@@ -0,0 +1,321 @@
...
+
+	for {
+		if err := wp.isClosed(ctx); err != nil {
+			return ignoreClosed(err)
+		}
+
+		item, err := wp.acquisition.Create(ctx)
+		if err != nil {
+			wp.createdCounter.Add(ctx, 1, metric.WithAttributes(failureAttr()))
+			time.Sleep(time.Second) // wait a bit before retrying
```
jakubno: Isn't this too much?

packages/shared/pkg/utils/warmpool.go
```diff
@@ -0,0 +1,321 @@
...
+		case <-wp.done:
+			wp.beginDestroy(ctx, item, "populate while closed")
+
+			return nil
+		case <-ctx.Done():
+			wp.beginDestroy(ctx, item, "populate while context canceled")
+
+			return ctx.Err()
+		case wp.freshItems <- item:
+		}
```
jakubno: is this really needed? looks like micro optimization

packages/shared/pkg/utils/warmpool.go
```diff
@@ -0,0 +1,321 @@
...
+}
+
+// Get an item from the pool.
+// - First try to find a reusable item.
+// - If no reusable items are available, return an item from the fresh channel.
+// - If the context has been canceled or the `Close` method has been called, return with an error.
+func (wp *WarmPool[T]) Get(ctx context.Context) (T, error) {
+	var item T
+
+	recordSuccess := func(source string, attrs ...attribute.KeyValue) {
```
jakubno: Why inline?

Can you also make `source` and `reason` a typed string?

packages/shared/pkg/utils/warmpool.go
```diff
@@ -0,0 +1,321 @@
...
+		return item, ctx.Err()
+	case <-wp.done:
+		recordFailure("closed")
+
+		return item, ErrClosed
+	case s := <-wp.reusableItems:
+		telemetry.ReportEvent(ctx, fmt.Sprintf("reused %s", wp.name))
+		recordSuccess("used")
+
+		return s, nil
```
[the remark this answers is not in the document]
jakubno: This seems valid

packages/shared/pkg/utils/warmpool.go
```diff
@@ -0,0 +1,321 @@
...
+	}
+}
+
+// Return an item to the pool.
+// - puts the item in `reusableItems` to get priority when being returned.
+// - destroy the item and return if the pool has already been closed.
+// - if the item fails to be pushed back into the channel for any reason, destroy it.
+func (wp *WarmPool[T]) Return(ctx context.Context, item T) {
+	wp.wg.Go(func() {
+		recordSuccess := func(attrs ...attribute.KeyValue) {
```
jakubno: isn't this a same function as above?

packages/shared/pkg/utils/warmpool.go
```diff
@@ -0,0 +1,321 @@
...
+
+		// try to push item in to reusable pool first
+		select {
+		case wp.reusableItems <- item:
+			telemetry.ReportEvent(ctx, fmt.Sprintf("returned %s", wp.name))
+			recordSuccess()
+
+			return
+		default:
+		}
```
jakubno: this seems to be kinda redundant

packages/shared/pkg/utils/warmpool.go
```diff
@@ -0,0 +1,321 @@
...
+
+		// if that didn't work, keep trying to push, but listen for failures as well
```
jakubno, line 238: this is a little bit different behavior than before. This will cause that if there's a spike of `s` items, reusable has capacity `k` and currently you are using `c`, you will always keep the same amount and there will be 
```
s - k - c
```
waiting routines

jakubno: I would put some timeout, if you can't return it in time, then discard?

packages/shared/pkg/utils/warmpool.go
```diff
@@ -0,0 +1,321 @@
...
+
+// Close destroys all items and prevents the pool from creating/returning anymore.
+func (wp *WarmPool[T]) Close(ctx context.Context) error {
+	logger.L().Info(ctx, "Closing pool")
+
+	if err := wp.isClosed(ctx); err != nil {
+		return err
+	}
+
+	wp.doneOnce.Do(func() {
```
jakubno, line 264: nit: why do you close freshItems in different function? Why not to keep it consistent?

packages/shared/pkg/utils/warmpool.go
```diff
@@ -0,0 +1,321 @@
...
+			telemetry.ReportEvent(ctx, fmt.Sprintf("returned %s", wp.name))
+			recordSuccess()
+
+			return
+		default:
+		}
+
+		// if that didn't work, keep trying to push, but listen for failures as well
+		select {
+		case wp.reusableItems <- item:
```
jakubno, line 240:
```suggestion
		case wp.reusableItems, ok <- item:
```

jakubno (changes requested):

packages/orchestrator/internal/sandbox/nbd/pool.go
djeebus: Only so that the script would pick it up and correct the path.

Edit: I'll see if I can fix the regex so this isn't necessary.

packages/shared/pkg/utils/warmpool.go
djeebus: Probably, wasn't sure what was reasonable. Mostly I'm just trying to avoid spinning a cpu for no reason. 100ms? This is only going to be hit when we're trying to populate the warm pool AND we fail to create a new item, and at some point that won't be necessary. Once the warm pool is full and the reused count is reasonable for keeping up with demand, the code will create one new one and then park while trying to push it into an already full channel.

packages/shared/pkg/utils/warmpool.go
djeebus: Exactly to avoid the need to create safety through types and constants. If we keep it internal to `WarmPool.Get`, it's very clear what it's scope is and what uses it. If we move it external, now we have to worry about misuse and abuse of this function.

packages/shared/pkg/utils/warmpool.go
djeebus: Slightly different attributes. I've simplified it a little bit now, should be more obvious.

packages/shared/pkg/utils/warmpool.go
djeebus: It's safe to read from a closed channel, but you'll panic if you write to a closed channel or close a channel twice. As such, golang encourages you to close channels near the writer, not the reader, and only when you're done writing to it. The `freshItems` channel gives us an easy place to do that - when the loop has exited, nothing will ever write to the channel again. `reusableItems` is odd, as its writer doesn't write in a loop, so it's awkward to close it within that channel; we have to resort to closing it in another function, then protecting it by selecting from both the write and a read to `done`.

packages/shared/pkg/utils/warmpool.go
djeebus: Not sure what you're getting at here?

packages/shared/pkg/utils/warmpool_test.go
```diff
@@ -0,0 +1,498 @@
...
+		time.Sleep(10 * time.Millisecond)
+
+		cancel()
+
+		<-done
+
+		item, ok := <-wp.freshItems
+		if ok {
+			assert.False(t, ok, "fresh channel should be closed, but got %s instead", item.Key)
+		}
```
jakubno:
```suggestion
		assert.False(t, ok, "fresh channel should be closed, but got %s instead", item.Key)
```

packages/shared/pkg/utils/warmpool_test.go
```diff
@@ -0,0 +1,498 @@
...
+
+		time.Sleep(10 * time.Millisecond)
+
+		cancel()
+
+		<-done
+	})
+}
+
+func TestWarmPool_Get(t *testing.T) {
```
jakubno, line 272: I would also add a test that gets from reuse first

packages/shared/pkg/utils/warmpool_test.go
```diff
@@ -0,0 +1,498 @@
...
+		)
+
+		err := wp.Close(t.Context())
+		require.NoError(t, err)
+
+		wp.Return(t.Context(), &testItem{})
+		assert.Empty(t, wp.reusableItems) // ensure the item did not make it to the reuse pool
+	})
+
+	t.Run("return ends in error when context already canceled", func(t *testing.T) {
```
jakubno: there's no error
```suggestion
	t.Run("return ends when context already canceled", func(t *testing.T) {
```

packages/shared/pkg/utils/warmpool_test.go
```diff
@@ -0,0 +1,498 @@
...
+		ctx, cancel := context.WithCancel(t.Context())
+		cancel()
+
+		wp.Return(ctx, &testItem{})
+		assert.Empty(t, wp.reusableItems) // ensure the item did not make it to the reuse pool
+	})
+
+	t.Run("return returns an item to the reuse pool", func(t *testing.T) {
+		f := NewMockItemFactory[*testItem](t)
+		f.EXPECT().Destroy(mock.Anything, mock.Anything).Return(nil)
```
jakubno: could you assert Destroy is not called

packages/shared/pkg/utils/warmpool_test.go
```diff
@@ -0,0 +1,498 @@
...
+		t.Cleanup(closePool(t, wp))
+
+		wp.Return(t.Context(), &testItem{})
+
+		wp.wg.Wait()
+
+		assert.Len(t, wp.reusableItems, 1) // ensure the item did not make it to the reuse pool
+	})
+
+	t.Run("return waits if the pool is already full", func(t *testing.T) {
```
jakubno: it seems it doesn't wait (anymore?)

packages/shared/pkg/utils/warmpool.go
jakubno: True, but if it's already full it won't try to create anyway.. I think it doesn't matter that much anyway

packages/shared/pkg/utils/warmpool.go
jakubno: I think it would be nice to quickly see what are the possible values

jakubno: It also prevents the metric dimension skyrocket

packages/shared/pkg/utils/warmpool.go
```diff
@@ -0,0 +1,325 @@
...
+		wp.getCounter.Add(ctx, 1, metric.WithAttributes(attrs...))
+	}
+	recordFailure := func(reason string, attrs ...attribute.KeyValue) {
+		attrs = append(attrs, failureAttr(), reasonAttr(reason))
+		wp.getCounter.Add(ctx, 1, metric.WithAttributes(attrs...))
+	}
+
+	// early check to bail
+	if err := wp.isClosed(ctx); err != nil {
+		recordFailure("closed")
```
jakubno: or canceled

packages/shared/pkg/utils/warmpool.go
```diff
@@ -0,0 +1,325 @@
...
+	// get from the reusable pool first
+	select {
+	case <-ctx.Done():
+		recordFailure("canceled")
+
+		return item, ctx.Err()
+	case <-wp.done:
+		recordFailure("closed")
+
+		return item, ErrClosed
```
jakubno: do we need to check here? we already check above and there's default branch

packages/shared/pkg/utils/warmpool.go
jakubno: sorry, incorrect place

packages/shared/pkg/utils/warmpool.go
jakubno: We have 3 different failures for `closed`

packages/shared/pkg/utils/warmpool.go
jakubno: You should check the `wg.done` before running `wg.Go`; otherwise, you have no guarantee there won't be a new Return called, please add test for that (long running Return)  +  Close and these `Return`s concurrently

jakubno: Good point, the mechanism in Populate is much cleaner and easier to get right

jakubno (changes requested):

packages/shared/pkg/utils/warmpool_test.go
djeebus: lol, thought I got all the AI nonsense. that'll teach me to trust auto complete =P I suppose it's _technically_ correct, but 100% ugly. thanks!

packages/shared/pkg/utils/warmpool_test.go
djeebus: Good point. We need the `Destroy` to close the pool, but I can avoid configuring it before it's necessary; if the test tries to call Destroy before clean up, it'll fail.

packages/shared/pkg/utils/warmpool.go
```diff
@@ -0,0 +1,329 @@
...
+	"go.opentelemetry.io/otel/metric"
+	"go.uber.org/zap"
+
+	"github.com/e2b-dev/infra/packages/shared/pkg/logger"
+	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
+)
+
+var meter = otel.Meter("github.com/e2b-dev/infra/packages/shared/pkg/utils")
+
+var ErrClosed = errors.New("cannot read from a closed pool")
```
levb: 0/5 `ErrPoolClosed`

packages/shared/pkg/utils/warmpool.go
```diff
@@ -0,0 +1,329 @@
...
+		case <-ctx.Done():
+			wp.beginDestroy(ctx, item, "return failed due to canceled context")
+			recordFailure("canceled")
+
+		case <-wp.done:
+			wp.beginDestroy(ctx, item, "return failed due to closed pool")
+			recordFailure("closed")
+		case <-time.After(time.Minute * 5):
+			wp.beginDestroy(ctx, item, "return failed due to closed pool")
+			recordFailure("return timeout")
```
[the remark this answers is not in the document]
levb: What cursor said

packages/shared/pkg/utils/warmpool.go
```diff
@@ -0,0 +1,329 @@
...
+
+		for item := range wp.reusableItems {
+			wp.destroy(ctx, item, "close and destroy reusable items")
+		}
+	})
+
+	return err
+}
+
+func (wp *WarmPool[T]) beginDestroy(ctx context.Context, item T, reason string) {
```
levb: nit: 0/5 rename to `destroyAsync` or even just 'goDestory`?

packages/shared/pkg/utils/warmpool.go
levb: Q: what prevents `reusableItems` from overfilling? The timeout? Sorry if I missed it.

packages/shared/pkg/utils/warmpool.go
```diff
@@ -0,0 +1,316 @@
...
+		close(wp.reusableItems)
+
+		for item := range wp.reusableItems {
+			wp.destroy(ctx, item, operationClose, reasonCleanupReusable)
+		}
+
+		// closing this channel is done in Populate
+		for item := range wp.freshItems {
+			wp.destroy(ctx, item, operationClose, reasonCleanupFresh)
+		}
```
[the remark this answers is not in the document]
djeebus: This used to be true, but isn't anymore.

packages/shared/pkg/utils/warmpool.go
djeebus: `destroyAsync` works for me :+1:

packages/shared/pkg/utils/warmpool.go
djeebus: - line 64 defines the max size for the channel: `reusableItems: make(chan T, maxPoolSize),`
- a channel cannot *over* fill, it'll simply hang while waiting for space in the channel
- to prevent it from hanging forever, we add `case <- time.After(timeout):` into the `select`, which will be chosen after `timeout` has passed.

packages/shared/pkg/utils/warmpool.go
djeebus: Ugh. Okay, nope, this wasn't an option here. I moved it into the `Close`.

It works perfectly fine if you call `Populate` before calling `Close`. If you don't, you end up with a deadlock while trying to empty the channel, as the channel never got closed. Safer to close everything in `Close`.

PR #1638, opened 2025-12-22
Make edge api optional for OSS deployment

sitole:
This pull request introduces a change that makes our API able to work and be deployed without the need for an edge API.
The main change is migrating queries for sandbox metrics/logs and build logs under the cluster resource provider, implemented as an interface backed by local (Loki, ClickHouse) and remote (edge API) implementations.

The second change is the use of local service discovery, which is now working only for template builders (orchestrators are still discovered in the old way to minimize changes in this pr).

The last change is that each instance now maintains its own gRPC connection. Before cluster instances, there was one shared connection handled by the cluster and reused by all instances. This was fine, but it forced us to provide metadata for each gRPC call, including the auth token and service instance ID metadata required by the gRPC proxy running on the edge API. Now, this is not needed, and each connection already holds this metadata, so we don't need to pass it in the context.

A change to connections was needed: now that we are not using the gRPC proxy for local cluster instances (template builders), we cannot use a shared connection, since it always points to different instances.

After this is deployed, the API can work without the edge API. It's no longer needed even in integration tests.

### How To Review
I tried to make changes between commits as small as possible, only changing the needed stuff, and then describing it in the commit description. Each commit should be self-contained and should compile without issues. You can go one by one to make the overall review faster and easier for both sides. Some commits do a temporary job to pipe stuff together before the final solution is in place.

I am happy to receive feedback on this, as we want to do more of it.

---

> [!NOTE]
> Enables deployments/tests without `edge` by abstracting cluster I/O and discovery.
> 
> - Adds `clusters` package: per-instance gRPC connections, local/remote service discovery, instance/cluster sync, and a `ClusterResource` interface with local (ClickHouse/Loki) and remote (Edge API) implementations for sandbox metrics/logs and template build logs
> - Updates config to require `LOKI_URL` (optional `LOKI_USER`/`LOKI_PASSWORD`); removes use of `LOCAL_CLUSTER_ENDPOINT`/`LOCAL_CLUSTER_TOKEN`
> - Infra: propagate `LOKI_URL` to Nomad jobs and Terraform locals; CI stops starting Edge service and removes related env vars
> - API wiring moved from shared connection to instance-scoped clients; new log query windowing and unified log source selection with fallbacks
> - go.mod: add Loki and related deps
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 914d08b5e2da036f7d6c7809b5756ad0aa70c14e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api/internal/clusters/resources_local.go
+++ b/packages/api/internal/clusters/resources_local.go
@@ -152,53 +152,17 @@ func (l *LocalClusterResourceProvider) GetBuildLogs(
 	direction api.LogsDirection,
 	source *api.LogsSource,
 ) ([]logs.LogEntry, error) {
-	ctx, span := tracer.Start(ctx, "get build-logs")
-	defer span.End()
-
+	// Use shared implementation with Loki as the persistent log backend
 	start, end := logQueryWindow(cursor, direction)
 
-	var sources []logSourceFunc
-
-	if nodeID != nil && logCheckSourceType(source, api.LogsSourceTemporary) {
-		instance, found := l.instances.Get(*nodeID)
-		if found {
-			sourceCallback := logsFromBuilderInstance(ctx, instance, templateID, buildID, offset, limit, level, start, end, direction)
-			sources = append(sources, sourceCallback)
-		} else {
-			logger.L().Warn(
-				ctx, "Node instance not found for build logs, falling back to other sources",
-				logger.WithNodeID(*nodeID),
-				logger.WithTemplateID(templateID),
-				logger.WithBuildID(buildID),
-			)
-		}
+	lokiDirection := defaultDirection
+	if direction == api.LogsDirectionBackward {
+		lokiDirection = logproto.BACKWARD
 	}
 
-	// Fetch logs from Loki backend (explicit persistent source or default)
-	if logCheckSourceType(source, api.LogsSourcePersistent) {
-		lokiDirection := defaultDirection
-		if direction == api.LogsDirectionBackward {
-			lokiDirection = logproto.BACKWARD
-		}
+	persistentFetcher := l.logsFromLocalLoki(ctx, templateID, buildID, start, end, int(limit), offset, level, lokiDirection)
 
-		sourceCallback := l.logsFromLocalLoki(ctx, templateID, buildID, start, end, int(limit), offset, level, lokiDirection)
-		sources = append(sources, sourceCallback)
-	}
-
-	// Iterate through sources and return the first successful fetch,
-	// it depends on nodeID and source what sources are available here.
-	for _, sourceFetch := range sources {
-		entries, err := sourceFetch()
-		if err != nil {
-			logger.L().Warn(ctx, "Error fetching build logs", logger.WithTemplateID(templateID), logger.WithBuildID(buildID), zap.Error(err))
-
-			continue
-		}
-
-		return entries, nil
-	}
-
-	return nil, nil
+	return getBuildLogsWithSources(ctx, l.instances, nodeID, templateID, buildID, offset, limit, level, cursor, direction, source, persistentFetcher)
 }
 
 func (l *LocalClusterResourceProvider) logsFromLocalLoki(ctx context.Context, templateID string, buildID string, start time.Time, end time.Time, limit int, offset int32, level *logs.LogLevel, direction logproto.Direction) logSourceFunc {
--- a/packages/api/internal/clusters/resources_remote.go
+++ b/packages/api/internal/clusters/resources_remote.go
@@ -7,11 +7,8 @@ import (
 	"net/http"
 	"time"
 
-	"go.uber.org/zap"
-
 	"github.com/e2b-dev/infra/packages/api/internal/api"
 	edgeapi "github.com/e2b-dev/infra/packages/shared/pkg/http/edge"
-	"github.com/e2b-dev/infra/packages/shared/pkg/logger"
 	"github.com/e2b-dev/infra/packages/shared/pkg/logs"
 	"github.com/e2b-dev/infra/packages/shared/pkg/smap"
 	"github.com/e2b-dev/infra/packages/shared/pkg/utils"
@@ -147,47 +144,11 @@ func (r *ClusterResourceProviderImpl) GetBuildLogs(
 	direction api.LogsDirection,
 	source *api.LogsSource,
 ) ([]logs.LogEntry, error) {
-	ctx, span := tracer.Start(ctx, "get build-logs")
-	defer span.End()
-
+	// Use shared implementation with Edge API as the persistent log backend
 	start, end := logQueryWindow(cursor, direction)
+	persistentFetcher := r.getBuildLogsFromEdge(ctx, templateID, buildID, offset, limit, level, start, end, direction)
 
-	var sources []logSourceFunc
-
-	if nodeID != nil && logCheckSourceType(source, api.LogsSourceTemporary) {
-		instance, found := r.instances.Get(*nodeID)
-		if found {
-			sourceCallback := logsFromBuilderInstance(ctx, instance, templateID, buildID, offset, limit, level, start, end, direction)
-			sources = append(sources, sourceCallback)
-		} else {
-			logger.L().Warn(
-				ctx, "Node instance not found for build logs, falling back to other sources",
-				logger.WithNodeID(*nodeID),
-				logger.WithTemplateID(templateID),
-				logger.WithBuildID(buildID),
-			)
-		}
-	}
-
-	if logCheckSourceType(source, api.LogsSourcePersistent) {
-		sourceCallback := r.getBuildLogsFromEdge(ctx, templateID, buildID, offset, limit, level, start, end, direction)
-		sources = append(sources, sourceCallback)
-	}
-
-	// Iterate through sources and return the first successful fetch,
-	// it depends on nodeID and source what sources are available here.
-	for _, sourceFetch := range sources {
-		entries, err := sourceFetch()
-		if err != nil {
-			logger.L().Warn(ctx, "Error fetching build logs", logger.WithTemplateID(templateID), logger.WithBuildID(buildID), zap.Error(err))
-
-			continue
-		}
-
-		return entries, nil
-	}
-
-	return nil, nil
+	return getBuildLogsWithSources(ctx, r.instances, nodeID, templateID, buildID, offset, limit, level, cursor, direction, source, persistentFetcher)
 }
 
 func (r *ClusterResourceProviderImpl) getBuildLogsFromEdge(ctx context.Context, templateID string, buildID string, offset int32, limit int32, level *logs.LogLevel, start time.Time, end time.Time, direction api.LogsDirection) logSourceFunc {
--- a/packages/api/internal/clusters/resources.go
+++ b/packages/api/internal/clusters/resources.go
@@ -12,6 +12,7 @@ import (
 	edgeapi "github.com/e2b-dev/infra/packages/shared/pkg/http/edge"
 	"github.com/e2b-dev/infra/packages/shared/pkg/logger"
 	"github.com/e2b-dev/infra/packages/shared/pkg/logs"
+	"github.com/e2b-dev/infra/packages/shared/pkg/smap"
 	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
 	"github.com/e2b-dev/infra/packages/shared/pkg/utils"
 )
@@ -110,4 +111,72 @@ func logsFromBuilderInstance(ctx context.Context, instance *Instance, templateID
 
 		return entries, nil
 	}
+}
+
+// getBuildLogsWithSources implements the shared logic for fetching build logs from multiple sources.
+// This function extracts the common pattern used by both local and remote cluster resource providers,
+// avoiding code duplication between the two implementations.
+//
+// The function tries log sources in order based on availability and configuration:
+// 1. Temporary logs from the builder instance (if nodeID is provided and source allows)
+// 2. Persistent logs from backend storage (strategy provided by caller)
+//
+// It returns the first successful result, logging warnings for any failures encountered.
+// This unified approach ensures consistent behavior and makes maintenance easier by centralizing
+// the source selection and fallback logic.
+func getBuildLogsWithSources(
+	ctx context.Context,
+	instances *smap.Map[*Instance],
+	nodeID *string,
+	templateID string,
+	buildID string,
+	offset int32,
+	limit int32,
+	level *logs.LogLevel,
+	cursor *time.Time,
+	direction api.LogsDirection,
+	source *api.LogsSource,
+	persistentLogFetcher logSourceFunc, // Backend-specific strategy for persistent logs (Loki for local, Edge API for remote)
+) ([]logs.LogEntry, error) {
+	ctx, span := tracer.Start(ctx, "get build-logs")
+	defer span.End()
+
+	start, end := logQueryWindow(cursor, direction)
+
+	var sources []logSourceFunc
+
+	// Handle temporary logs from builder instance
+	if nodeID != nil && logCheckSourceType(source, api.LogsSourceTemporary) {
+		instance, found := instances.Get(*nodeID)
+		if found {
+			sourceCallback := logsFromBuilderInstance(ctx, instance, templateID, buildID, offset, limit, level, start, end, direction)
+			sources = append(sources, sourceCallback)
+		} else {
+			logger.L().Warn(
+				ctx, "Node instance not found for build logs, falling back to other sources",
+				logger.WithNodeID(*nodeID),
+				logger.WithTemplateID(templateID),
+				logger.WithBuildID(buildID),
+			)
+		}
+	}
+
+	// Handle persistent logs (backend-specific implementation provided by caller)
+	if logCheckSourceType(source, api.LogsSourcePersistent) {
+		sources = append(sources, persistentLogFetcher)
+	}
+
+	// Iterate through sources and return the first successful fetch
+	for _, sourceFetch := range sources {
+		entries, err := sourceFetch()
+		if err != nil {
+			logger.L().Warn(ctx, "Error fetching build logs", logger.WithTemplateID(templateID), logger.WithBuildID(buildID), zap.Error(err))
+
+			continue
+		}
+
+		return entries, nil
+	}
+
+	return nil, nil
 }


packages/api/internal/clusters/instance.go
```diff
@@ -0,0 +1,147 @@
...
+)
+
+type Instance struct {
+	// Identifier that uniquely identifies the instance so it will not be registered multiple times.
+	// Depending on service discovery used, it can be combination of different parameters, what service discovery gives us.
+	UniqueIdentifier string
+
+	ClusterID  uuid.UUID
+	NodeID     string
+	InstanceID string
```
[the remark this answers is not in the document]
sitole: It's okay as instance ID is under one word long, so it's safe to access concurrently.

packages/api/internal/orchestrator/lifecycle.go
```diff
@@ -19,8 +19,8 @@ func (o *Orchestrator) addSandboxToRoutingTable(ctx context.Context, sandbox san
 		info := e2bcatalog.SandboxInfo{
 			OrchestratorID: node.Metadata().ServiceInstanceID,
 			OrchestratorIP: node.IPAddress,
-			ExecutionID:    sandbox.ExecutionID,
 
+			SandboxExecutionID:      sandbox.ExecutionID,
```
dobrac: isn't everything in the SandboxInfo struct about `sandbox` if not specified otherwise?

packages/api/internal/clusters/cluster.go
```diff
@@ -80,11 +80,14 @@ func NewCluster(ctx context.Context, tel *telemetry.Client, endpoint string, end
...
 
+	instances := smap.New[*ClusterInstance]()
+
 	c := &Cluster{
 		ID:            clusterID,
 		SandboxDomain: sandboxDomain,
 
-		instances:  smap.New[*ClusterInstance](),
+		resources:  newRemoteClusterResourceProvider(instances, httpClient),
+		instances:  instances,
```
dobrac: this is quite strange ownership of the instances list

packages/api/internal/template-manager/logs.go
```diff
@@ -1,74 +0,0 @@
...
-		// Return the first non-error logs, the providers are ordered by most up-to-date data
-		return l
-	}
-
-	return nil
-}
-
-func isSourceType(source *api.LogsSource, sourceType api.LogsSource) bool {
-	return source == nil || *source == sourceType
-}
```
dobrac, line 74: where did this handling go?

packages/api/internal/clusters/cluster_instances.go
```diff
@@ -156,10 +156,10 @@ func (d clusterSynchronizationStore) PoolInsert(ctx context.Context, item api.Cl
 		logger.WithServiceInstanceID(item.ServiceInstanceID),
 	)
 
-	instance := &ClusterInstance{
+	instance := &Instance{
 		NodeID: item.NodeID,
 
-		ServiceInstanceID:    item.ServiceInstanceID,
+		InstanceID:           item.ServiceInstanceID,
```
dobrac: can we keep it as serviceinstanceID? Or maybe rename Instance to ServiceInstance?

packages/api/internal/clusters/instance.go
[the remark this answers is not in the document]
dobrac: can you please link why is this really safe in the long run?

packages/api/internal/clusters/instances_sync.go
```diff
@@ -13,11 +13,11 @@ import (
 )
 
 // SynchronizationStore defines methods for synchronizing cluster instances
-type clusterSynchronizationStore struct {
+type instancesSyncStore struct {
```
dobrac, line 16: why is it not a cluster store anymore? Also, outdated comment

packages/api/internal/clusters/instances_sync.go
```diff
@@ -69,26 +70,53 @@ func (d instancesSyncStore) PoolInsert(ctx context.Context, item api.ClusterOrch
...
+	if err != nil {
+		closeErr := instance.Close()
+		if closeErr != nil {
+			logger.L().Error(ctx, "Failed to close cluster instance after sync failure",
+				zap.Error(closeErr),
+				logger.WithClusterID(d.cluster.ID),
+				logger.WithNodeID(instance.NodeID),
+				logger.WithServiceInstanceID(instance.InstanceID),
+			)
+		}
```
dobrac: this is a duplicit logic with the one in "newInstance". Can we unite it?

packages/api/internal/clusters/instance.go
```diff
@@ -17,36 +29,119 @@ type Instance struct {
...
+	ctx context.Context,
+	tel *telemetry.Client,
+	clusterAuth *instanceAuthorization,
+	clusterID uuid.UUID,
+	nodeID string,
+	instanceID string,
+	connAddr string,
+	connTls bool,
+) (*Instance, error) {
+	conn, err := createConnection(tel, clusterAuth, connAddr, connTls)
```
dobrac: does this handle automatic reconnects when the connection drops?

dobrac: isn't it quite wasteful to keep a connection for each serviceinstance? This number will be quite soon in thousands.

packages/api/internal/clusters/discovery/local.go
```diff
@@ -1 +1,60 @@
...
+
+func NewLocalDiscovery(clusterID uuid.UUID, nomad *nomadapi.Client) Discovery {
+	return &LocalServiceDiscovery{
+		nomad:     nomad,
+		clusterID: clusterID,
+	}
+}
+
+func (sd *LocalServiceDiscovery) Query(ctx context.Context) ([]Item, error) {
+	_, span := tracer.Start(ctx, "query-local-cluster-nodes", trace.WithAttributes(telemetry.WithClusterID(sd.clusterID)))
```
dobrac: missing context usage

packages/api/internal/clusters/discovery/remote.go
```diff
@@ -0,0 +1,56 @@
...
+
+func NewRemoteServiceDiscovery(clusterID uuid.UUID, client *api.ClientWithResponses) Discovery {
+	return &RemoteServiceDiscovery{
+		clusterID: clusterID,
+		client:    client,
+	}
+}
+
+func (sd *RemoteServiceDiscovery) Query(ctx context.Context) ([]Item, error) {
+	_, span := tracer.Start(ctx, "query-remote-cluster-nodes", trace.WithAttributes(telemetry.WithClusterID(sd.clusterID)))
```
dobrac: missing context usage

packages/api/internal/clusters/resources_local.go
```diff
@@ -0,0 +1,225 @@
...
+
+func (l *LocalClusterResourceProvider) GetBuildLogs(ctx context.Context, nodeID *string, templateID string, buildID string, offset int32, limit int32, level *logs.LogLevel, cursor *time.Time, direction api.LogsDirection, source *api.LogsSource) ([]logs.LogEntry, error) {
+	start, end := logQueryWindow(cursor, direction)
+
+	lokiDirection := defaultDirection
+	if direction == api.LogsDirectionBackward {
+		lokiDirection = logproto.BACKWARD
+	}
+
+	// todo
```
dobrac: todo?

dobrac (commented):
First batch of comments

Questions so far:
- how will we handle the sandbox/build logs in the BYOC and no access to Loki in the API?
- isn't now the edge only a grpc proxy?
- why not use the same cluster discovery for local as for remote cluster?
- not fully confident about the scaling performance of handling too many connections. Do you have a plan how to deal with that?

sitole:
> how will we handle the sandbox/build logs in the BYOC and no access to Loki in the API?

This PR introduces a remote resource provider. Each BYOC deployment exposes an edge API for sandbox logs, metrics, etc.

> isn't now the edge only a grpc proxy?

For this PR, the edge service (part of the client proxy) stays where it is now. In the following PRs, the edge API and edge gRPC proxy will be completely removed from the open-source repository, allowing us to implement closed-source features available only to enterprise (BYOC) clients.

> why not use the same cluster discovery for local as for remote cluster?

What do you mean here? For the local cluster, we are using a local Nomad server to do discovery. For remote clusters, we use the service discovery endpoint exposed by the edge API because we don't have access to the clients' Nomad deployments, which live in their VPCs without our direct access.

This is why there are two implementations (local and remote) for cluster instance discovery.

> not fully confident about the scaling performance of handling too many connections. Do you have a plan for how to deal with that?

I think we are now far from any real limits we could hit. At some point, we will need to implement a pool of connections for each call, and this pool will be smaller than the number of instances, so we will benefit from it.

We will fall into the same situation, too, as with the previous implementation: one connection shared across all instances. When the instance count per cluster is higher, we will hit the connection's throughput, and we will be forced to implement a pool otherwise.

I prefer to use a connection per instance as it simplifies all thinking, makes the mental model exactly the same as with a connection used for local nodes and code is cleaner. 

@dobrac

packages/api/internal/clusters/instance.go
sitole: We are doing the same with local nodes now. More here - https://github.com/e2b-dev/infra/pull/1638#issuecomment-3687047451

sitole: Yes, it does the same as our connection for the local oirch instances, it only handles optional TLS and auth middleware.

packages/api/internal/clusters/instances_sync.go
sitole: I think it's been a little confusing with the naming since the initial implementation. It's *instances* sync store, not *cluster*. Also, it's now more obvious since we removed the direct dependency on the cluster struct.

Agree with the comment, will fix it.

packages/api/internal/template-manager/logs.go
sitole: Moved under package handling for cluster resource logic and named `logCheckSourceType`.

https://github.com/e2b-dev/infra/blob/41045e8bcad9bd6ee4b36a21f505a021e6625ba3/packages/api/internal/clusters/resources.go

packages/api/internal/orchestrator/lifecycle.go
sitole: Agree, I will remove the prefix from `SandboxStartedAt ` and `SandboxMaxLengthInHours` too.

packages/api/internal/clusters/cluster.go
sitole: Do you have something concrete to worry about here?
It's used internally and not exported outside the cluster package, so I wouldn't worry about any unwanted access. 

If only worry is pushing instances to `newRemoteClusterResourceProvider`, we can replace it with a fetch callback that returns only the instances' connections (used for direct template builds).

packages/api/internal/clusters/cluster_instances.go
sitole: Actually, I think we need a lock here. I know we internally realized we dont need lock for smaller types (in our case lot of booleans/small numbers) as they are smaller than machine word and they can be safely accessed in parallel - https://go.dev/ref/mem#restrictions.

This is, of course, not the case, as the value size is a UUID (16 bytes), making it ish 2x bigger.
I will add a lock here.

packages/api/internal/clusters/resources_local.go
sitole: It's already outdated.

packages/api/internal/clusters/instance.go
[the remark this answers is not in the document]
sitole: Actually, I think we need a lock here. I know we internally realized we dont need lock for smaller types (in our case lot of booleans/small numbers) as they are smaller than machine word and they can be safely accessed in parallel - https://go.dev/ref/mem#restrictions.

This is, of course, not the case, as the value size is a UUID (16 bytes), making it ish 2x bigger.
I will add a lock here.

packages/api/internal/clusters/resources_local.go
```diff
@@ -0,0 +1,188 @@
...
+
+		entries, err := l.queryLogsProvider.QueryBuildLogs(ctx, templateID, buildID, start, end, int(limit), offset, level, lokiDirection)
+		if err != nil {
+			return nil, fmt.Errorf("error when fetching build logs: %w", err)
+		}
+
+		return entries, nil
+	}
+
+	return nil, nil
```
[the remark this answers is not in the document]
sitole: This is expected.

sitole:
Test from mail

packages/api/internal/clusters/resources_local.go
```diff
@@ -0,0 +1,188 @@
...
+}
```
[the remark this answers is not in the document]
sitole: True, thanks for the notice.

In https://github.com/e2b-dev/infra/pull/1638/commits/fa2b8fb78a43a9134a56bde2da191fa30b54b22f, I re-worked how logs are fetched and added back the previous style with a callback, so it simplifies logic for cases where the source is not defined, and we want to check all available sources.

sitole:
In https://github.com/e2b-dev/infra/pull/1638/commits/fa2b8fb78a43a9134a56bde2da191fa30b54b22f, I re-worked how logs are fetched and added back the previous style with a callback, so it simplifies logic for cases where the source is not defined, and we want to check all available sources.

packages/api/internal/clusters/cluster.go
```diff
@@ -0,0 +1,227 @@
...
+}
+
+func (c *Cluster) GetTemplateBuilderByNodeID(nodeID string) (*Instance, error) {
+	instance, found := c.instances.Get(nodeID)
+	if !found {
+		return nil, ErrTemplateBuilderNotFound
+	}
+
+	if info := instance.GetInfo(); info.Status == infogrpc.ServiceInfoStatus_Unhealthy || !info.IsBuilder {
+		return nil, ErrTemplateBuilderNotFound
```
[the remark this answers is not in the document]
sitole: This looks correct to me. `GetAvailableTemplateBuilder` should return a healthy one as we want to work with it later, and work takes some time. With `GetAvailableTemplateBuilder`, we can return draining as it's only used for short-term work (getting logs, status of the build), and in some cases, the builder can be in the draining stage already.

packages/api/internal/clusters/instances_sync.go
```diff
@@ -0,0 +1,128 @@
...
+
+		return
+	}
+
+	d.instances.Insert(item.NodeID, instance)
+}
+
+func (d instancesSyncStore) PoolUpdate(ctx context.Context, instance *Instance) {
+	_ = d.tryToSyncInstance(ctx, instance)
+}
```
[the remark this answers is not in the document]
sitole: Nice catch.

This can theoretically happen when the network connection is broken, when the orchestrator is removed prematurely and we miss draining, when the unhealthy stage is missed, or when Nomad still discovers this allocation while we are in a race condition between service discovery calls.

packages/api/internal/clusters/resources_local.go
```diff
@@ -0,0 +1,213 @@
...
+		if err != nil {
+			logger.L().Warn(ctx, "Error fetching build logs", logger.WithTemplateID(templateID), logger.WithBuildID(buildID), zap.Error(err))
+
+			continue
+		}
+
+		return entries, nil
+	}
+
+	return nil, nil
```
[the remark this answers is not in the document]
sitole: We actually did this previously https://github.com/e2b-dev/infra/blob/38ec66b4cea0f3d890424e92a7228b6dd56e8d0a/packages/api/internal/template-manager/logs.go#L69

iac/provider-gcp/nomad/main.tf
```diff
@@ -157,7 +156,7 @@ resource "nomad_job" "client_proxy" {
       redis_cluster_url   = local.redis_cluster_url
       redis_tls_ca_base64 = trimspace(data.google_secret_manager_secret_version.redis_tls_ca_base64.secret_data)
 
-      loki_url                     = "http://loki.service.consul:${var.loki_service_port.port}"
+      loki_url                     = local.loki_url
```
dobrac, line 159: does the client proxy still need loki?

packages/api/internal/clusters/resources_remote.go
```diff
@@ -0,0 +1,233 @@
...
+			Level:     api.LogLevel(row.Level),
+			Message:   row.Message,
+			Fields:    row.Fields,
+		}
+	}
+
+	return api.SandboxLogs{Logs: l, LogEntries: le}, nil
+}
+
+func (r *ClusterResourceProviderImpl) GetBuildLogs(
```
dobrac, line 135: the implementations of build and sandbox logs are unfortunately almost the same for local and remote. This will be difficult to maintain. Could we rethink this?

packages/client-proxy/internal/service-discovery/static-service-discovery.go
```diff
@@ -3,21 +3,21 @@ package service_discovery
...
 type StaticServiceDiscovery struct {
-	items []ServiceDiscoveryItem
+	items []DiscoveredInstance
 }
 
 func NewStaticServiceDiscovery(results []string, port uint16) *StaticServiceDiscovery {
-	items := make([]ServiceDiscoveryItem, 0)
+	items := make([]DiscoveredInstance, 0)
 
 	for _, result := range results {
```
dobrac: we could do `utils.Map(` instead

packages/api/internal/template-manager/template_manager.go
```diff
@@ -59,7 +58,7 @@ const (
 
 func New(
 	sqlcDB *sqlcdb.Client,
-	edgePool *edge.Pool,
+	edgePool *clusters.Pool,
```
dobrac:
```suggestion
	clusters *clusters.Pool,
```
(+ formatting)

packages/shared/pkg/consts/sandboxes.go
```diff
@@ -9,4 +14,4 @@ const NodeIDLength = 8
 // We don't want to use some obviously dummy value such as empty zeros, because for users it will look like something is wrong with the sandbox id
 const ClientID = "6532622b"
 
-var OrchestratorPort = os.Getenv("ORCHESTRATOR_PORT")
+var OrchestratorApiPort = uint16(utils.Must(strconv.ParseUint(env.GetEnv("ORCHESTRATOR_PORT", "5008"), 10, 16)))
```
dobrac:
```suggestion
var OrchestratorAPIPort = uint16(utils.Must(strconv.ParseUint(env.GetEnv("ORCHESTRATOR_PORT", "5008"), 10, 16)))
```

packages/api/internal/orchestrator/nodemanager/node.go
```diff
@@ -41,8 +39,8 @@ type Node struct {
 	IPAddress     string
 	SandboxDomain *string
 
-	client *grpclient.GRPCClient
-	status api.NodeStatus
+	connection *clusters.GRPCClient
```
dobrac: shouldn't this be "client"?

packages/api/internal/orchestrator/nodemanager/node.go
```diff
@@ -167,22 +166,21 @@ func NewClusterNode(ctx context.Context, client *grpclient.GRPCClient, clusterID
 func (n *Node) Close(ctx context.Context) error {
 	if n.IsNomadManaged() {
```
dobrac, line 167: this might be nicer to solve with interface and two implementations, but feel free to do in another PR

packages/api/internal/clusters/cluster.go
```diff
@@ -0,0 +1,234 @@
...
+	endpoint string,
+	endpointTLS bool,
+	secret string,
+	clusterID uuid.UUID,
+	sandboxDomain *string,
+) (*Cluster, error) {
+	endpointBaseUrl := fmt.Sprintf("http://%s", endpoint)
+	if endpointTLS {
+		endpointBaseUrl = fmt.Sprintf("https://%s", endpoint)
+	}
```
dobrac: changed in main to:

```
// generate the full endpoint URL
	scheme := "http"
	if endpointTLS {
		scheme = "https"
	}
	endpointBaseUrl := fmt.Sprintf("%s://%s", scheme, endpoint)
```

dobrac (commented):
is there any specific order for the production deployed?

iac/provider-gcp/nomad/main.tf
sitole: Now yes, removing edge api - part that needs loki will be next step after this is merged.

packages/api/internal/clusters/resources_remote.go
sitole: I unified logic under one function that accepts a persistent logs provider - the only difference between these two implementations.  https://github.com/e2b-dev/infra/pull/1638/commits/914d08b5e2da036f7d6c7809b5756ad0aa70c14e

PR #1665, opened 2025-12-31
Add build-node-info flag to prefer a machine type for builds

jakubno:
The feature flag has following attributes:
- team_id
- template_id

---

> [!NOTE]
> Implements feature-flagged node preference for template builds based on machine CPU attributes.
> 
> - Adds `preferred-build-node` JSON flag (`BuildNodeInfo`) and `machineinfo.MachineInfo` with JSON parsing (`FromLDValue`) and compatibility checks
> - Updates `edge.Cluster.GetAvailableTemplateBuilder(ctx, info)` to filter healthy builder nodes by CPU compatibility when `info` is provided
> - `TemplateManager.GetAvailableBuildClient` reads the flag (scoped to `cluster` context) to select a compatible builder; falls back to any builder if none found
> - Build handlers (`deprecated_template_start_build.go`, `template_start_build_v2.go`) set LaunchDarkly `template` context before build coordination
> - Persists selected node's machine info into the build record as before
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f7530ab6101be29367e4ff26d2c4dfdd6a6c6c82. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api/internal/template-manager/template_manager.go
+++ b/packages/api/internal/template-manager/template_manager.go
@@ -124,6 +124,8 @@ func (tm *TemplateManager) GetAvailableBuildClient(ctx context.Context, clusterI
 	if err != nil {
 		if errors.Is(err, edge.ErrAvailableTemplateBuilderNotFound) {
 			// Fallback to any template builder
+			logger.L().Warn(ctx, "No available template builder found with the specified machine info, falling back to any available template builder", zap.String("clusterID", clusterID.String()))
+
 			builder, err = cluster.GetAvailableTemplateBuilder(ctx, machineinfo.MachineInfo{})
 			if err != nil {
 				return nil, fmt.Errorf("failed to get any available template builder for cluster '%s': %w", clusterID, err)


packages/api/internal/edge/cluster.go
```diff
@@ -152,6 +153,10 @@ func (c *Cluster) GetAvailableTemplateBuilder(ctx context.Context) (*ClusterInst
 			continue
 		}
 
+		if info.CPUModel != "" && !info.IsCompatibleWith(instance.machineInfo) {
```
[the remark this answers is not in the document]
jakubno: Info is a struct (not a pointer)

packages/api/internal/template-manager/template_manager.go
```diff
@@ -114,8 +115,20 @@ func (tm *TemplateManager) GetAvailableBuildClient(ctx context.Context, clusterI
 		return nil, fmt.Errorf("cluster with ID '%s' not found", clusterID)
 	}
 
-	builder, err := cluster.GetAvailableTemplateBuilder(ctx)
+	nodeInfoJSON := tm.featureFlags.JSONFlag(ctx, featureflags.BuildNodeInfo)
+	nodeInfo := machineinfo.FromLDValue(ctx, nodeInfoJSON)
+	builder, err := cluster.GetAvailableTemplateBuilder(ctx, nodeInfo)
 	if err != nil {
+		if errors.Is(err, edge.ErrAvailableTemplateBuilderNotFound) {
+			// Fallback to any template builder
```
dobrac, line 126: why the fallback?

Also, shouldn't this be set somehow per cluster?

dobrac (changes requested):

packages/api/internal/template-manager/template_manager.go
jakubno: Let's say, we want to use Type 1, but there was some issue and we still have Type 2, it's probably better to use Type 2 for the build rather than fail the build, isn't it? 

> Also, shouldn't this be set somehow per cluster?

I added cluster context

dobrac: > Let's say, we want to use Type 1, but there was some issue and we still have Type 2, it's probably better to use Type 2 for the build rather than fail the build, isn't it?

Lets add at least warning or error when this happens

dobrac (approved):

PR #1676, opened 2026-01-04
refactor: return 404 for missing cluster, optimize locks and preallocations

yaninyzwitty:
- Changed GetClusterSandboxMetrics to return http.StatusNotFound
- Optimized ClusterInstance syncInstance by reducing lock hold time:
  build status, roles, and machineInfo outside of the mutex and lock
  only for assignment. Improves concurrency and performance.
- Added slice/map preallocation where applicable for SandboxMetrics
  and sandboxes map to reduce allocations and improve efficiency.

### DIFF
--- a/packages/api/internal/edge/cluster_instances.go
+++ b/packages/api/internal/edge/cluster_instances.go
@@ -48,9 +48,9 @@ func (c *Cluster) syncInstance(ctx context.Context, instance *ClusterInstance) {
 
 	// we are taking service info directly from the instance to avoid timing delays in service discovery
 	reqCtx := metadata.NewOutgoingContext(ctx, grpc.Metadata)
-	info, grpcError := grpc.Client.Info.ServiceInfo(reqCtx, &emptypb.Empty{})
+	info, err := grpc.Client.Info.ServiceInfo(reqCtx, &emptypb.Empty{})
 
-	err := utils.UnwrapGRPCError(grpcError)
+	err = utils.UnwrapGRPCError(err)
 	if err != nil {
 		logger.L().Error(ctx, "Failed to get instance info",
 			zap.Error(err),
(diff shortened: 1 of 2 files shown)


packages/api/internal/edge/cluster_instances.go
```diff
@@ -48,9 +48,9 @@ func (c *Cluster) syncInstance(ctx context.Context, instance *ClusterInstance) {
 
 	// we are taking service info directly from the instance to avoid timing delays in service discovery
 	reqCtx := metadata.NewOutgoingContext(ctx, grpc.Metadata)
-	info, err := grpc.Client.Info.ServiceInfo(reqCtx, &emptypb.Empty{})
+	info, grpcError := grpc.Client.Info.ServiceInfo(reqCtx, &emptypb.Empty{})
 
-	err = utils.UnwrapGRPCError(err)
+	err := utils.UnwrapGRPCError(grpcError)
```
jakubno: what was the reason for renaming it? Is there some advantage?

packages/api/internal/edge/cluster_instances.go
```diff
@@ -62,12 +62,19 @@ func (c *Cluster) syncInstance(ctx context.Context, instance *ClusterInstance) {
...
 	}
 
-	instance.mutex.Lock()
-	defer instance.mutex.Unlock()
+	// build the data object first to minimize lock time
+	// machineInfo.FromGrpcInfo might be expensive
+
+	infosStatus := info.GetServiceStatus()
+	roles := info.GetServiceRoles()
+	machineInfo := machineinfo.FromGRPCInfo(info.GetMachineInfo())
```
jakubno: no need to do this as those operations aren't expensive, also please keep `defer *.Unlock()`, as the product evolves there's lower chance to end up with accidental deadlock when you return early without unlocking

packages/api/internal/edge/metrics.go
```diff
@@ -11,11 +11,23 @@ import (
...
+		DiskTotal:     m.DiskTotal,
+		DiskUsed:      m.DiskUsed,
+	}
+}
 func GetClusterSandboxMetrics(ctx context.Context, pool *Pool, sandboxID string, teamID string, clusterID uuid.UUID, qStart *int64, qEnd *int64) ([]api.SandboxMetric, *api.APIError) {
 	cluster, ok := pool.GetClusterById(clusterID)
 	if !ok {
 		return nil, &api.APIError{
-			Code:      http.StatusInternalServerError,
+			Code:      http.StatusNotFound,
```
jakubno, line 18: we want to return 500 here as it isn't a user error (the data are correct, but something unexpected happened on the server)

packages/api/internal/edge/metrics.go
```diff
@@ -74,7 +79,7 @@ func GetClusterSandboxListMetrics(ctx context.Context, pool *Pool, teamID string
...
```
jakubno: same here

packages/api/internal/edge/cluster_instances.go
yaninyzwitty: I will revert and keep edgeMetricToApi and the preallocation. I’ll update the PR.

packages/api/internal/edge/metrics.go
yaninyzwitty: I will revert most of the changes I made, according to your request

yaninyzwitty: if anything is required, tell me

PR #1678, opened 2026-01-05
Transparently check if template alias exists

sitole:
We want to add separated template that accepts only template alias (not
id) and transparently shows difference between 404 (you are owner and
template does not exists) and 403 (you are not owner but template
exists).

---

> [!NOTE]
> Adds a dedicated endpoint to resolve templates by alias and wires it through the server interface and router.
> 
> - New `GET /templates/aliases/{alias}` route with `GetTemplatesAliasesAlias` handler and middleware in `api.gen.go`
> - Router registration for `/templates/aliases/:alias`
> - Makefile `generate` now also produces `internal/api/client.gen.go`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8fc7c16baa9ebcf5d2342ac8749e7ab2af8cbb8f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


spec/openapi.yml
```diff
@@ -2525,6 +2525,33 @@ paths:
...
+          schema:
+            type: string
+            description: Template alias
+      responses:
+        "200":
+          description: Successfully queried template by alias
+        "400":
+          $ref: "#/components/responses/400"
+        "403":
+          $ref: "#/components/responses/401"
```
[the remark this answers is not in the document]
sitole: thx!

sitole:
In d2e792c3a750d27dcb39c6cde05a688bb13c0e52 i added test containers for api handlers. Implemented the same way we are using for our internal rest services to make testing easier.

We can later clean up and unify testing utilities, but it would be a bigger refactor than this PR itself, as now the utils are only in test packages that cannot access the private code needed for testing api handlers.

packages/api/internal/handlers/template_alias.go
```diff
@@ -0,0 +1,48 @@
...
+		a.sendAPIStoreError(c, apiErr.Code, apiErr.ClientMsg)
+		telemetry.ReportCriticalError(ctx, "error when getting team and tier", apiErr.Err)
+
+		return
+	}
+
+	result, err := a.sqlcDB.GetTemplateAliasByAlias(ctx, alias)
+	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			a.sendAPIStoreError(c, http.StatusNotFound, "Template alias not found")
```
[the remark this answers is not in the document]
sitole: We have here test for that.

packages/api/Makefile
```diff
@@ -16,6 +16,7 @@ generate:
 	$(codegen) -old-config-style -generate gin --package api $(openapi) > internal/api/api.gen.go
 	$(codegen) -old-config-style -generate types --package api $(openapi) > internal/api/types.gen.go
 	$(codegen) -old-config-style -generate spec --package api $(openapi) > internal/api/spec.gen.go
+	$(codegen) -old-config-style -generate client --package api $(openapi) > internal/api/client.gen.go
```
jakubno, line 19: why?

spec/openapi.yml
```diff
@@ -2525,6 +2525,33 @@ paths:
...
+      parameters:
+        - name: alias
+          in: path
+          required: true
+          schema:
+            type: string
+            description: Template alias
+      responses:
+        "200":
+          description: Successfully queried template by alias
```
jakubno, line 2557: maybe return template it belongs to? or return 204

packages/api/Makefile
sitole: I need client methods for tests to parse the response body from the handler.

spec/openapi.yml
sitole: I will return template ID and team ID

packages/api/internal/handlers/template_alias.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/gin-gonic/gin"
+
+	"github.com/e2b-dev/infra/packages/api/internal/api"
+	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
+)
+
+func (a *APIStore) GetTemplatesAliasesAlias(c *gin.Context, alias string) {
+	ctx := c.Request.Context()
+
+	team, apiErr := a.GetTeam(ctx, c, nil)
```
jakubno, line 17: this may not work as intended (it will after removing access token)

spec/openapi.yml
```diff
@@ -2525,6 +2537,37 @@ paths:
...
           $ref: "#/components/responses/500"
 
+  /templates/aliases/{alias}:
+    get:
+      description: Check if template with given alias exists
+      tags: [templates]
+      security:
+        - AccessTokenAuth: []
+        - ApiKeyAuth: []
+        - Supabase1TokenAuth: []
```
jakubno, line 2546:
```suggestion
        - Supabase1TokenAuth: []
          Supabase2TeamAuth: []
```

spec/openapi.yml
```diff
@@ -837,6 +837,18 @@ components:
           items:
             $ref: "#/components/schemas/TemplateBuild"
 
+    TemplateQuick:
```
jakubno:
```suggestion
    TemplateAliasResponse:
```

spec/openapi.yml
jakubno: also i would remove access token

sitole: Can you add context here, please? We have different combinations of a lot of template-related endpoints.

PR #1683, opened 2026-01-06
fix: template manager rolling deployments

dobrac:
Use the job type `service` instead of `system` to properly do rolling updates with health checks. In order to sync the number of instances with the nodes count, there is a custom nomad APM plugin exposing the number of nodes in a node_pool. The nomad autoscaler then syncs the count to the number of template manager instances.

**Migration** path without downtime:
- run `make migrate-template-manager-detach` to remove existing template-manager from the terraform state
- `make build-and-upload/nomad-nodepool-apm`
- `make plan`
- `make apply`
- go to Nomad and stop allocations in the template-manager-system job one by one (ensure the allocation finished draining before proceeding to a next one)
- the new job should start the allocations as soon as the port is freed

---

> [!NOTE]
> Implements rolling deployments for `template-manager` by converting the Nomad job from `system` to `service` and introducing autoscaling tied to node pool size.
> 
> - Replaces `template-manager-system` with `job "template-manager"` (service) including `update` config, `distinct_hosts` constraint, health checks, and a scaling policy driven by a new APM metric; preserves count via `scripts/get-nomad-job-count.sh`
> - Adds Nomad Autoscaler job (`jobs/nomad-autoscaler.hcl`) configured to load a custom APM plugin and expose health; wires it in Terraform with conditional creation
> - Introduces `packages/nomad-nodepool-apm` (plugin code, Dockerfile, Makefile, README) and adds build/upload targets and `go.work` entry; Terraform fetches the plugin from GCS with checksum
> - Adds migration target to detach old `template-manager-system` from Terraform state
> - Updates Nomad API dependency across modules; minor log writer fix in orchestrator (`filtered_writer.go`)
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3677c4d863219315690b0b695b7ac6fc1f029545. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/nomad-nodepool-apm/plugin/plugin.go
+++ b/packages/nomad-nodepool-apm/plugin/plugin.go
@@ -129,7 +129,7 @@ func (p *NodePoolPlugin) Query(query string, _ sdk.TimeRange) (sdk.TimestampedMe
 	// Count only nodes that are ready and eligible for scheduling
 	readyCount := 0
 	for _, node := range nodes {
-		if node.Status == "ready" && node.SchedulingEligibility == "eligible" {
+		if node.Status == api.NodeStatusReady && node.SchedulingEligibility == api.NodeSchedulingEligible {
 			readyCount++
 		}
 	}


iac/provider-gcp/nomad/jobs/template-manager.hcl
```diff
@@ -1,16 +1,49 @@
...
-  type = "system"
+job "template-manager" {
+  type = "service"
   node_pool  = "${node_pool}"
-  priority = 70
+  priority = 75
+
+  group "template-manager" {
+    # Count is fetched from current Nomad state to preserve autoscaler-managed value
+    count = ${current_count}
```
sitole, line 8: Do we need to specify if it's autoscaled?

dobrac: yes, otherwise it will go to 1. I know, ridiculous, tested.

dobrac: linking the issues:

https://github.com/hashicorp/nomad/issues/19845
https://github.com/hashicorp/terraform-provider-nomad/issues/420

sitole: Oh my god 🤦‍♂️

> yes, otherwise it will go to 1. I know, ridiculous, tested.

This means it will not just set "count = 1" and apply it, but will actually try to kill all allocations except that one, or autoscaling will catch it and correct the value?

sitole: Thinking if there can be some race condition where we will actually set the wrong count.

dobrac: > This means it will not just set "count = 1" and apply it, but will actually try to kill all allocations except that one, or autoscaling will catch it and correct the value?

it will descale everything to 1 + restart the one as an update (make it completely unavailable) and then the autoscaler will kick in and scale it back

> Thinking if there can be some race condition where we will actually set the wrong count.

There is a min=2 which should prevent no template manager being available even if you really mess up. Otherwise I think only for the brief amount of time between request for the count and apply of the job (we don't await full job deployment).

The good thing is, that even scaling down is fully graceful (sends SIGTERM and awaits the exit), so no traffic should be interrupted.

sitole: Thank you for more context.

packages/orchestrator/internal/template/build/writer/filtered_writer.go
```diff
@@ -43,7 +43,7 @@ func (w *PrefixFilteredWriter) writeLine(line []byte) (remaining []byte) {
 	if w.buff.Len() == 0 {
 		w.log(line)
 
-		return
+		return remaining
```
jakubno, line 46: should this be here?

iac/provider-gcp/nomad/jobs/template-manager.hcl
```diff
@@ -1,16 +1,49 @@
...
 %{ if update_stanza }
-  update {
-    max_parallel      = 1    # Update only 1 node at a time
-  }
-%{ endif }
+    # Scaling policy to match node count in the pool
+    # Uses the nomad-nodepool APM plugin
+    scaling {
+      enabled = true
+      min     = 2
```
jakubno, line 21: will this work well if you have only 1 node?

iac/provider-gcp/nomad/scripts/get-nomad-job-count.sh
jakubno, line 1: Why Nomad, why😢

packages/nomad-nodepool-apm/plugin/plugin.go
```diff
@@ -0,0 +1,162 @@
...
+		Filter: fmt.Sprintf(`NodePool == "%s"`, escapedNodePool),
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to list nodes in pool %q: %w", nodePool, err)
+	}
+
+	// Count only nodes that are ready and eligible for scheduling
+	readyCount := 0
+	for _, node := range nodes {
+		if node.Status == "ready" && node.SchedulingEligibility == "eligible" {
```
jakubno: - so marking node ineligible will try to kill trigger scale down. Will it correctly target the one which is ineligible and is that actually desired
- similar for draining, what happens if I start draining a node, will it count correctly
- will the terraform script work correctly for these situations?

jakubno:
```suggestion
		if node.Status == api.NodeStatusReady && node.SchedulingEligibility == api.NodeSchedulingEligible {
```

packages/orchestrator/internal/template/build/writer/filtered_writer.go
dobrac: it's a linter thing

iac/provider-gcp/nomad/jobs/template-manager.hcl
dobrac: yes and no, if you have only 1 node, no auto scaler is enabled at all

packages/nomad-nodepool-apm/plugin/plugin.go
dobrac: > so marking node ineligible will try to kill trigger scale down. Will it correctly target the one which is ineligible and is that actually desired

yes, it will scale down. No, it will not stop the correct one if you don't drain the node (I know a bit weird, suggestions are welcome)

> similar for draining, what happens if I start draining a node, will it count correctly

if you drain it, it will stop the correct one and the the autoscale will be == to number of eligible nodes

> will the terraform script work correctly for these situations?

terraform reads it from the nomad job count, so it's unrelated to these scenarios

dobrac: > yes, it will scale down. No, it will not stop the correct one if you don't drain the node (I know a bit weird, suggestions are welcome)

If you then drain the job on the non-eligible node, it will be rescheduled automatically on the eligible nodes (to keep the count)

jakubno: So we should always drain and we shouldn't use ineligible, do i get it right?

jakubno (approved):

packages/nomad-nodepool-apm/plugin/plugin.go
dobrac: I think so for the template managers. But even if you accidentally make it ineligible, nothing terrible happens. I think this is still better, than scaling to the number of nodes regardless of availability and then let it being stuck on not enough available nodes to place the job (when you actually want to drain the node).

If we decide to change it, it should be quite easy though

PR #1684, opened 2026-01-06
Fixed: flaky TestCleanDeletesTwoFiles

levb:
Due to concurrency, sometimes clean-nfs-cache was cleaning 1 more batch than expected by the test. Adjusted the expectations, the code behavior is correct.

### DIFF
--- a/packages/orchestrator/cmd/clean-nfs-cache/ex/clean_test.go
+++ b/packages/orchestrator/cmd/clean-nfs-cache/ex/clean_test.go
@@ -41,7 +41,7 @@ func TestDirSort(t *testing.T) {
 	require.Equal(t, int64(100), d.Files[2].ATimeUnix)
 }
 
-func TestCleanDeletesTwoFiles(t *testing.T) {
+func TestCleanDeletesOldestFiles(t *testing.T) {
 	root := t.TempDir()
 	defer os.RemoveAll(root)
 


packages/orchestrator/cmd/clean-nfs-cache/ex/clean_test.go
```diff
@@ -114,12 +114,14 @@ func TestCleanDeletesTwoFiles(t *testing.T) {
...
 	require.GreaterOrEqual(t, len(deleted), 2)
 
-	// The two files must have been some combination of 7 and 8 (from whichever folder)
+	// The two files must have been some combination of 7 and 8 (from whichever
+	// folder) Because of concurrency, sometimes we may pick an extra batch of
+	// candidates, so include 6 as well.
 	for _, d := range deleted {
-		require.Regexp(t, `.+/file(7|8)\.txt`, d)
+		require.Regexp(t, `.+/file(6|7|8)\.txt`, d)
 	}
```
jakubno, line 125: Wouldn't something like that be safer?
```go
	// The two files must have been some combination of 7 and 8 (from whichever folder)
	deletedFiles := make(map[string]struct{})
	for _, d := range deleted {
		file := filepath.Base(d)
		deletedFiles[file] = struct{}{}
	}

	require.Contains(t, deletedFiles, "file7.txt")
	require.Contains(t, deletedFiles, "file8.txt")
```

jakubno (changes requested):

packages/orchestrator/cmd/clean-nfs-cache/ex/clean_test.go
levb: It's longer but I don't see it as a stronger assertion. The point of the test is that none of the more recent files  (0-5) are deleted. That 6s get picked up on occasion (we drain the delete queue once the target is reached) - was by design.

levb: 1/5  can rename the test to avoid the confusion

packages/orchestrator/cmd/clean-nfs-cache/ex/clean_test.go
```diff
@@ -114,12 +114,14 @@ func TestCleanDeletesTwoFiles(t *testing.T) {
 		}
 	}
 
-	// Expect at least 2 deletions
+	// Expect at least 2 deletions, it can be more due to concurrency.
```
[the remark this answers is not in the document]
levb: the extra deletions were by design, the test was not adequate

PR #1686, opened 2026-01-06
fix(analytics): guard long-running sandboxes and set proper report timeout

yaninyzwitty:
- Add check for zero StartTime to avoid misclassifying sandboxes as long-running
- Use a single `now` timestamp for consistent evaluation within each tick
- Replace `syncAnalyticsTime` with `reportTimeout` for analytics request context
- Preallocate slices safely and append IDs instead of using full-length slice
- Improves correctness, testability, and avoids sending empty or duplicate analytics events

### DIFF
--- a/packages/api/internal/orchestrator/analytics.go
+++ b/packages/api/internal/orchestrator/analytics.go
@@ -60,11 +60,11 @@ func sendAnalyticsForLongRunningSandboxes(ctx context.Context, analytics *analyt
 	childCtx, cancel := context.WithTimeout(ctx, reportTimeout)
 	defer cancel()
 
-	instanceIds := make([]string, 0, len(instances))
-	executionIds := make([]string, 0, len(instances))
-	for _, i := range instances {
-		instanceIds = append(instanceIds, i.SandboxID)
-		executionIds = append(executionIds, i.ExecutionID)
+	instanceIds := make([]string, len(instances))
+	executionIds := make([]string, len(instances))
+	for i, instance := range instances {
+		instanceIds[i] = instance.SandboxID
+		executionIds[i] = instance.ExecutionID
 	}
 
 	_, err := analytics.RunningInstances(childCtx,


packages/api/internal/orchestrator/analytics.go
```diff
@@ -38,8 +38,12 @@ func (o *Orchestrator) reportLongRunningSandboxes(ctx context.Context) {
 		case <-ticker.C:
 			sandboxes := o.sandboxStore.Items(nil, []sandbox.State{sandbox.StateRunning})
 			longRunningSandboxes := make([]sandbox.Sandbox, 0, len(sandboxes))
+			now := time.Now()
 			for _, sandbox := range sandboxes {
-				if time.Since(sandbox.StartTime) > oldSandboxThreshold {
+				if sandbox.StartTime.IsZero() {
+					continue
+				}
+				if now.Sub(sandbox.StartTime) > oldSandboxThreshold {
```
jakubno: can we keep 
```
time.Since(sandbox.StartTime) > oldSandboxThreshold
```

I think it's more readable and the improvement is negligible

packages/api/internal/orchestrator/analytics.go
```diff
@@ -38,8 +38,12 @@ func (o *Orchestrator) reportLongRunningSandboxes(ctx context.Context) {
 		case <-ticker.C:
 			sandboxes := o.sandboxStore.Items(nil, []sandbox.State{sandbox.StateRunning})
 			longRunningSandboxes := make([]sandbox.Sandbox, 0, len(sandboxes))
+			now := time.Now()
 			for _, sandbox := range sandboxes {
-				if time.Since(sandbox.StartTime) > oldSandboxThreshold {
+				if sandbox.StartTime.IsZero() {
+					continue
+				}
```
jakubno: this check isn't really needed

packages/api/internal/orchestrator/analytics.go
```diff
@@ -57,14 +57,14 @@ func sendAnalyticsForLongRunningSandboxes(ctx context.Context, analytics *analyt
...
+	childCtx, cancel := context.WithTimeout(ctx, reportTimeout)
 	defer cancel()
 
-	instanceIds := make([]string, len(instances))
-	executionIds := make([]string, len(instances))
-	for idx, i := range instances {
-		instanceIds[idx] = i.SandboxID
-		executionIds[idx] = i.ExecutionID
+	instanceIds := make([]string, 0, len(instances))
+	executionIds := make([]string, 0, len(instances))
```
jakubno: We know the length will be exactly instances, so it's seems to be safe to preallocate and do it as it is now. What was the reason behind this change?

jakubno (changes requested):

packages/api/internal/orchestrator/analytics.go
yaninyzwitty: Well ok — given your  current logic always appends exactly one entry per instance, this change doesn’t add much value. I will revert it to keep the code simpler.

PR #1711, opened 2026-01-12
Use teamID from headers in template request

jakubno:
> [!NOTE]
> Enables team selection via header for template operations and deprecates body `teamID` in new template requests.
> 
> - Adds `Supabase2TeamAuth` security to template endpoints (`/templates`, `/v2/templates`, `/v3/templates`, builds/logs/status/files routes) and wires `Supabase2TeamAuthScopes` in generated Gin middleware
> - Marks `teamID` as deprecated in `TemplateBuildRequestV2` and `TemplateBuildRequestV3` (OpenAPI, generated types, and test models); updates embedded swagger accordingly
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f41379670728b6a200fcbc9a0dd6e5e5e2b6ffb6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


spec/openapi.yml
```diff
@@ -907,6 +907,7 @@ components:
           items:
             type: string
         teamID:
+          deprecated: true
```
sitole, line 910: They are already not used, can we delete them?

jakubno: I think we should keep the public API backward compatible

sitole: Backward compatible, yes, but it will be compatible even if we remove that.
Gin JSON parser will not complain when the `teamID` field is unused in the request. For a user who already does not make sense, as the team from the auth token is used.

Or am I missing something here?

PR #1714, opened 2026-01-13
fix(edge): avoid int overflow when serializing sandbox max lifetime

yaninyzwitty:
Use strconv.FormatInt / ParseInt instead of Itoa/Atoi to prevent
silent truncation when values exceed math.MaxInt or on 32-bit systems.


jakubno:
I think this is little bit unnecessary, I don't think there's any reason to set sandbox life to bigger number than the max int32 (2^31-1) hours

PR #1723, opened 2026-01-13
Refactor storage API for clarity 

levb:
In preparation for adding compression

### DIFF
--- a/packages/shared/pkg/storage/storage_fs_test.go
+++ b/packages/shared/pkg/storage/storage_fs_test.go
@@ -44,7 +44,7 @@ func TestOpenObject_Write_Exists_WriteTo(t *testing.T) {
 	require.Equal(t, contents, data)
 }
 
-func TestWriteFromFileSystem(t *testing.T) {
+func TestFSPut(t *testing.T) {
 	t.Parallel()
 	p := newTempProvider(t)
 	ctx := t.Context()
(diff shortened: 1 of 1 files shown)


packages/orchestrator/internal/sandbox/build/storage_diff.go
```diff
@@ -155,6 +155,10 @@ func (b *StorageDiff) FileSize() (int64, error) {
 	return c.FileSize()
 }
 
+func (b *StorageDiff) Size(ctx context.Context) (int64, error) {
+	return b.FileSize()
```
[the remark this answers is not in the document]
levb: It's probably best to return an error here, this method exists only to satisfy BlobReader, and is not really called. I did not want to create a yet-another interface without Size()

packages/shared/pkg/storage/storage_aws.go
```diff
@@ -26,26 +26,26 @@ const (
...
 	client        *s3.Client
 	presignClient *s3.PresignClient
 	bucketName    string
 }
 
-var _ StorageProvider = (*AWSBucketStorageProvider)(nil)
+var _ StorageProvider = (*awsStorage)(nil)
 
-type AWSBucketStorageObjectProvider struct {
+type awsObject struct {
```
djeebus, line 37: I don't think hiding these is the right call. We use GCP right now in experiments like `copy-build`, but only b/c we're currently primarily GCP. If that were to change, we'd need this exposed in order to perform those experiments. I don't think it makes this package much simpler to hide this. Not a big deal either way though, as this isn't expected to be a publicly consumed package in general.

packages/shared/pkg/storage/storage_cache.go
```diff
@@ -22,7 +22,7 @@ const (
 	cacheDirPermissions  = 0o700
 )
 
-type CachedProvider struct {
+type cache struct {
```
djeebus, line 25: Similar to the above. Also, in general, the golang community would prefer that you [accept interfaces, but return structs](https://duncanleung.com/go-idiom-accept-interfaces-return-types/)  << not authoritative, but the best I could find

PR #1777, opened 2026-01-27
Add support for ".terraform.${ENV}.tfvars" files

djeebus:
> [!NOTE]
> **Low Risk**
> Low risk Makefile/IaC tooling change; main impact is plans/destroys/imports may now pick up additional variables when an env-specific tfvars file exists.
> 
> **Overview**
> Adds support for env-specific Terraform variable files by teaching the GCP provider Makefile to auto-include `./.terraform.${ENV}.tfvars` (when present) via `-var-file` for `plan`, `destroy`, targeted plan commands, `plan-without-jobs`, and `import`.
> 
> Updates `.gitignore` to ignore all `*.tfvars` files (instead of only `terraform.tfvars`) to prevent committing per-env variable files.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 832c1180b0e82a6c6cd6d5ec1d640841cafc67d7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/iac/provider-gcp/Makefile
+++ b/iac/provider-gcp/Makefile
@@ -98,7 +98,7 @@ switch:
 .PHONY: destroy
 destroy:
 	@ printf "Destroying Terraform for env: `tput setaf 2``tput bold`$(ENV)`tput sgr0`\n\n"
-	@ $(tf_vars) $(TF) destroy
+	@ $(tf_vars) $(TF) destroy $(TF_VAR_FILE_ARG)
 
 .PHONY: plan
 plan:


iac/provider-gcp/Makefile
```diff
@@ -4,6 +4,8 @@ ENV_FILE := ../../.env.${ENV}
 -include ${ENV_FILE}
 
 TF := $(shell which terraform)
+TF_VAR_FILE := ./.terraform.${ENV}.tfvars
```
sitole, line 7: Do we want tfvars in iac/provider or in the project root, along with other env files?

The context-related iac/provider folder makes more sense to me, but managing multi-env setups will be more complicated, as you need to manage both the .env and .tfenvs files, which are in different directions.

djeebus: Our current setup (putting the "root" of the terraform files inside of nested folders) makes it difficult for IDEs to help with auto complete. By putting it in the same folder as main, the IDE can help out and we don't have to do odd things like "-var-file ../../.terraform.tfvars". It also puts it clearly where they belong, in case you have both (for example) aws and gcp infra. Our current setup would require that both files be in the same place. I'd prefer it colocated with the actual .tf files, but it's not critical; I'll move it if you feel strongly about keeping them in the root.

djeebus: Long term, I'd like to see if we can remove the `.env.<name>` files entirely, but I'm not sure if that's possible. Using tfvars as a source of truth for terraform makes sense, but I know we're using the files for other things, and I'm not convinced yet that we'll be able to remove them entirely.

sitole: Yep, the point about IDEs' autocomplete is valid. Same, that it should be as close to real Terraform as possible.

> Long term, I'd like to see if we can remove the .env.<name> files entirely

Yes please :D

PR #1799, opened 2026-01-28
feat: snapshots

dobrac:
## Summary

- Add snapshot API endpoints to create persistent snapshots from running sandboxes
- Implement `Checkpoint` gRPC call that atomically pauses, snapshots, and resumes sandboxes as new Firecracker instances
- Add database schema to track snapshot source and origin sandbox
- Support named snapshots with alias/namespace/tag for versioned snapshot management

## Description

This PR introduces a Snapshots feature that allows users to create persistent point-in-time snapshots from running sandboxes. The sandbox is briefly paused during the checkpoint, then a new Firecracker instance is booted from the snapshot with the same sandbox ID — from the API consumer's perspective the sandbox remains available, though there is a brief period of downtime during the operation.

### New API Endpoints

- **POST `/sandboxes/{sandboxID}/snapshots`** - Create a snapshot from a running sandbox (supports optional `name` with `namespace/alias:tag` syntax)
- **GET `/snapshots`** - List all snapshots for the team (with optional `sandboxID` filter and cursor pagination)

### Checkpoint Flow

1. API receives snapshot request for a running sandbox
2. Sandbox state transitions to `snapshotting` (internal state, prevents concurrent operations)
3. Orchestrator calls `Checkpoint` on the node (acquires starting-sandboxes semaphore to limit concurrency)
4. Node pauses the original VM, creates snapshot, starts uploading to storage
5. A **new** Firecracker instance is booted from the snapshot (new `ExecutionID`, same sandbox ID)
6. The original VM is stopped in the background
7. Build status updated to `uploaded`, sandbox transitions back to `running`

### Template Deletion

Template deletion now rejects with 400 if the template has existing snapshots referencing it. Build artifact cleanup is deferred to future GC (`[ENG-3477]`).

### Database Changes

- New `snapshot_templates` table tracking `env_id`, `sandbox_id`, and `created_at`
- Updated template queries to include `source` and `source_sandbox_id` fields

---

> [!NOTE]
> **High Risk**
> Touches core sandbox lifecycle/state transitions, introduces new gRPC behavior that pauses/resumes live sandboxes, and adds new persistence/schema paths; failures could lead to leaked sandboxes or inconsistent state if edge cases aren’t covered.
> 
> **Overview**
> Adds first-class **snapshot templates**: an API to create a persistent snapshot from a running sandbox (optionally named via team-scoped alias/tag) and to list snapshots with cursor pagination. This introduces a new `snapshotting` sandbox state with transient transition semantics (and conflict responses for pause/resume during snapshotting), plus an orchestrator `Checkpoint` gRPC that pauses, snapshots, resumes into a new Firecracker lifecycle while keeping stable sandbox identity, and uploads snapshot/prefetch metadata. It also adds DB support (`snapshot_templates` table and indexes), broadens template lookups to include `snapshot_template` sources, updates deletion to cascade DB records without removing build artifacts, and tweaks routing/proxy connection pooling to use a per-VM `LifecycleID` to avoid resume/cleanup races.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 65011db88870874b5dcd4640803a9631505d1dbd. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api/internal/sandbox/storage/redis/state_change_test.go
+++ b/packages/api/internal/sandbox/storage/redis/state_change_test.go
@@ -13,7 +13,7 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/e2b-dev/infra/packages/api/internal/sandbox"
-	"github.com/e2b-dev/infra/packages/shared/pkg/redis"
+	redis_utils "github.com/e2b-dev/infra/packages/shared/pkg/redis"
 )
 
 func setupTestStorage(t *testing.T) (*Storage, redis.UniversalClient) {
--- a/tests/integration/internal/tests/api/sandboxes/snapshot_template_test.go
+++ b/tests/integration/internal/tests/api/sandboxes/snapshot_template_test.go
@@ -262,10 +262,10 @@ func TestSnapshotTemplateCreateSandbox(t *testing.T) {
 
 // waitForSnapshotting polls the database until a build with status 'snapshotting'
 // appears for the given sandbox, or the timeout expires.
-func waitForSnapshotting(t *testing.T, db *setup.Database, sandboxID string, timeout time.Duration) {
+func waitForSnapshotting(t *testing.T, db *setup.Database, sandboxID string) {
 	t.Helper()
 
-	deadline := time.Now().Add(timeout)
+	deadline := time.Now().Add(30 * time.Second)
 	for time.Now().Before(deadline) {
 		var count int
 
@@ -338,7 +338,7 @@ func TestSnapshotTemplateConcurrentOperations(t *testing.T) {
 		sbx := utils.SetupSandboxWithCleanup(t, c, utils.WithAutoPause(false))
 
 		snapshotDone := startSnapshotInBackground(t, c, sbx.SandboxID)
-		waitForSnapshotting(t, db, sbx.SandboxID, 30*time.Second)
+		waitForSnapshotting(t, db, sbx.SandboxID)
 
 		pauseResp, err := c.PostSandboxesSandboxIDPauseWithResponse(t.Context(), sbx.SandboxID, setup.WithAPIKey())
 		require.NoError(t, err)
@@ -353,7 +353,7 @@ func TestSnapshotTemplateConcurrentOperations(t *testing.T) {
 		sbx := utils.SetupSandboxWithCleanup(t, c, utils.WithAutoPause(false))
 
 		snapshotDone := startSnapshotInBackground(t, c, sbx.SandboxID)
-		waitForSnapshotting(t, db, sbx.SandboxID, 30*time.Second)
+		waitForSnapshotting(t, db, sbx.SandboxID)
 
 		killResp, err := c.DeleteSandboxesSandboxIDWithResponse(t.Context(), sbx.SandboxID, setup.WithAPIKey())
 		require.NoError(t, err)
@@ -369,7 +369,7 @@ func TestSnapshotTemplateConcurrentOperations(t *testing.T) {
 		sbx := utils.SetupSandboxWithCleanup(t, c, utils.WithAutoPause(false))
 
 		snapshotDone := startSnapshotInBackground(t, c, sbx.SandboxID)
-		waitForSnapshotting(t, db, sbx.SandboxID, 30*time.Second)
+		waitForSnapshotting(t, db, sbx.SandboxID)
 
 		resumeResp, err := c.PostSandboxesSandboxIDResumeWithResponse(
 			t.Context(),
@@ -389,7 +389,7 @@ func TestSnapshotTemplateConcurrentOperations(t *testing.T) {
 		sbx := utils.SetupSandboxWithCleanup(t, c, utils.WithAutoPause(false))
 
 		snapshotDone := startSnapshotInBackground(t, c, sbx.SandboxID)
-		waitForSnapshotting(t, db, sbx.SandboxID, 30*time.Second)
+		waitForSnapshotting(t, db, sbx.SandboxID)
 
 		resp, err := c.PostSandboxesSandboxIDSnapshotsWithResponse(
 			t.Context(),


packages/api/internal/handlers/snapshot_create.go
```diff
@@ -0,0 +1,205 @@
...
+		return
+	}
+
+	if sbx.State != sandbox.StateRunning {
+		a.sendAPIStoreError(c, http.StatusConflict, fmt.Sprintf("Sandbox '%s' is not running", sandboxID))
+
+		return
+	}
+
+	telemetry.ReportEvent(ctx, "Creating snapshot")
```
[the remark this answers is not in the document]
dobrac: Fixed. Added quota enforcement with `maxSnapshotsPerTeam = 100` constant. The handler now calls `CountTeamSnapshots` before creating a new snapshot and returns 403 Forbidden when the limit is reached.

```go
snapshotCount, err := a.sqlcDB.CountTeamSnapshots(ctx, teamID)
if err != nil {
    // handle error
}
if snapshotCount >= maxSnapshotsPerTeam {
    a.sendAPIStoreError(c, http.StatusForbidden, 
        fmt.Sprintf("Snapshot limit reached. Maximum %d snapshots per team", maxSnapshotsPerTeam))
    return
}
```

packages/api/internal/handlers/snapshot_create.go
```diff
@@ -0,0 +1,205 @@
...
+		a.tryResumeSandbox(ctx, sandboxID, teamInfo)
+		a.sendAPIStoreError(c, http.StatusInternalServerError, "Error creating snapshot")
+
+		return
+	}
+
+	resumeErr := a.resumeSandboxAfterSnapshot(ctx, sandboxID, teamInfo, lastSnapshot)
+	if resumeErr != nil {
+		logger.L().Error(ctx, "Error resuming sandbox after snapshot", zap.Error(resumeErr.Err), logger.WithSandboxID(sandboxID))
+		a.sendAPIStoreError(c, resumeErr.Code, fmt.Sprintf("Snapshot created but failed to resume sandbox: %s", resumeErr.ClientMsg))
```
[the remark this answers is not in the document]
dobrac: Fixed. Changed the error handling to return the created snapshot info (HTTP 201) even when resume fails. This prevents users from creating duplicate snapshots on retry since they will see the snapshot was already created.

```go
snapshotInfo := buildSnapshotInfo(...)

resumeErr := a.resumeSandboxAfterSnapshot(ctx, sandboxID, teamInfo, lastSnapshot)
if resumeErr != nil {
    logger.L().Error(ctx, "Error resuming sandbox after snapshot", ...)
    telemetry.ReportError(ctx, "Snapshot created but failed to resume sandbox", resumeErr.Err)
    c.JSON(http.StatusCreated, snapshotInfo)  // Still return 201 with snapshot info
    return
}

c.JSON(http.StatusCreated, snapshotInfo)
```

The error is logged and reported to telemetry for monitoring.

packages/api/internal/handlers/snapshot_delete.go
```diff
@@ -0,0 +1,59 @@
...
+
+			return
+		}
+		logger.L().Error(ctx, "Error getting snapshot", zap.Error(err), zap.String("snapshot_id", snapshotID))
+		a.sendAPIStoreError(c, http.StatusInternalServerError, "Error deleting snapshot")
+
+		return
+	}
+
+	err = a.sqlcDB.DeleteSnapshotTemplate(ctx, queries.DeleteSnapshotTemplateParams{
```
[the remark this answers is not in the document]
dobrac: Fixed. Added storage cleanup similar to template deletion. The handler now:

1. Gets exclusive builds using `GetExclusiveBuildsForTemplateDeletion`
2. Deletes the DB record
3. Asynchronously cleans up storage artifacts using `templateManager.DeleteBuilds`

```go
builds, err := a.sqlcDB.GetExclusiveBuildsForTemplateDeletion(ctx, snapshotID)
// ... delete from DB ...
go a.deleteSnapshotBuilds(context.WithoutCancel(ctx), snapshotID, builds, teamInfo.Team.ClusterID)
a.templateCache.InvalidateAllTags(snapshotID)
```

The async cleanup follows the same pattern used in `sandbox_kill.go` for snapshot build deletion.

packages/db/queries/snapshots/list_team_snapshots.sql
```diff
@@ -0,0 +1,41 @@
...
+) eb ON TRUE
+LEFT JOIN LATERAL (
+    SELECT ARRAY_AGG(alias ORDER BY alias) AS aliases
+    FROM "public"."env_aliases"
+    WHERE env_id = e.id
+) ea ON TRUE
+WHERE e.team_id = @team_id
+AND e.source = 'snapshot'
+AND (
+    @source_sandbox_id::text IS NULL
```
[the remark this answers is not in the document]
dobrac: Fixed. Changed the SQL condition from `@source_sandbox_id::text IS NULL` to `@source_sandbox_id = ''` to properly handle empty string filtering.

The Go handler passes empty string when no filter is provided, so checking for empty string is the correct approach:

```sql
WHERE e.team_id = @team_id
AND e.source = 'snapshot'
AND (
    @source_sandbox_id = '' 
    OR e.source_sandbox_id = @source_sandbox_id
)
```

packages/api/internal/handlers/snapshot_create.go
```diff
@@ -0,0 +1,205 @@
...
+		return
+	}
+
+	if sbx.TeamID != teamID {
+		a.sendAPIStoreError(c, http.StatusForbidden, fmt.Sprintf("You don't have access to sandbox '%s'", sandboxID))
+
+		return
+	}
+
+	if sbx.State != sandbox.StateRunning {
```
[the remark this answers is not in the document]
dobrac: Addressed. The existing state check combined with the sandbox state machine provides protection:

1. The `StateRunning` check ensures only running sandboxes can be snapshotted
2. When `RemoveSandbox` is called with `StateActionPause`, the sandbox transitions to `StatePausing`
3. Any concurrent request will fail the `sbx.State != sandbox.StateRunning` check since the sandbox is now in `StatePausing` state

Updated the error message to be clearer:
```go
if sbx.State != sandbox.StateRunning {
    a.sendAPIStoreError(c, http.StatusConflict, 
        fmt.Sprintf("Sandbox '%s' is not running or is already being snapshotted", sandboxID))
    return
}
```

The state transitions (`Running → Pausing → Paused`) are atomic in Redis, preventing race conditions at the storage layer.

packages/api/internal/handlers/snapshot_create.go
```diff
@@ -0,0 +1,205 @@
...
+
+			return
+		}
+		telemetry.ReportError(ctx, "Error pausing sandbox for snapshot", err)
+		a.sendAPIStoreError(c, http.StatusInternalServerError, "Error creating snapshot")
+
+		return
+	}
+
+	lastSnapshot, err := a.sqlcDB.GetLastSnapshot(ctx, sandboxID)
```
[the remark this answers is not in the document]
dobrac: Acknowledged. This is a valid concern about the async nature of the pause operation.

The current flow relies on `RemoveSandbox` completing the pause synchronously before `GetLastSnapshot` is called. Looking at the orchestrator code, `RemoveSandbox` with `StateActionPause` does wait for the snapshot to be written to the database before returning.

However, if there are edge cases where this assumption fails, potential improvements would be:
1. Add polling/retry logic for `GetLastSnapshot` with exponential backoff
2. Return a "pending" status and let clients poll for completion
3. Add a synchronous confirmation channel from the orchestrator

This would require architectural changes to the pause flow. Flagging for follow-up.

packages/api/internal/handlers/snapshot_create.go
```diff
@@ -0,0 +1,205 @@
...
+	if err != nil {
+		logger.L().Error(ctx, "Error getting snapshot after pause", zap.Error(err), logger.WithSandboxID(sandboxID))
+		a.tryResumeSandbox(ctx, sandboxID, teamInfo)
```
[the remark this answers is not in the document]
dobrac: Partially addressed. The `tryResumeSandbox` function handles the recovery case, but you are correct that if the snapshot was never created in the DB, the sandbox could be left in an inconsistent state.

The fix for the data leak issue (returning 201 even on resume failure) helps here - if snapshot creation succeeded but resume failed, the sandbox is paused and the user gets the snapshot info. They can then manually resume it.

For the case where `GetLastSnapshot` fails entirely, the sandbox may indeed be stuck in a paused state. A more robust solution would require:
1. A background reconciliation job to detect and recover orphaned sandboxes
2. Or adding explicit state tracking for "snapshot in progress" that can be rolled back

Flagging for potential future improvement.

packages/api/internal/handlers/snapshot_create.go
```diff
@@ -0,0 +1,205 @@
...
+		TeamID:             teamID,
+		SourceSandboxID:    &sandboxID,
+		BaseTemplateID:     &snap.BaseEnvID,
+		Vcpu:               build.Vcpu,
+		RamMb:              build.RamMb,
+		FreeDiskSizeMb:     build.FreeDiskSizeMb,
+		KernelVersion:      build.KernelVersion,
+		FirecrackerVersion: build.FirecrackerVersion,
+		EnvdVersion:        build.EnvdVersion,
+		Status:             "success",
```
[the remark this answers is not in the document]
dobrac: Acknowledged. The `status: "success"` is set immediately because the snapshot files (memory state, disk image) are created synchronously by the orchestrator during the pause operation before control returns to the API handler.

However, you raise a valid point about eventual consistency if snapshot files need to be replicated to GCS or other nodes. The current architecture assumes the snapshot is usable from the origin node immediately.

To fully address this, we could:
1. Add a "pending" → "ready" status flow with verification
2. Add a health check endpoint for snapshot availability
3. Document the expected availability latency

This is an architectural consideration for high-availability deployments. Flagging for documentation/follow-up.

spec/openapi.yml
```diff
@@ -2289,6 +2311,73 @@ paths:
         "404":
           $ref: "#/components/responses/404"
 
+  /sandboxes/{sandboxID}/snapshots:
```
sitole, line 2332:
```suggestion
  /sandboxes/{sandboxID}/snapshot:
```

We already have `/sandboxes/{sandboxID}/pause`  and `/sandboxes/{sandboxID}/resume`

packages/db/migrations/20260211120000_add_snapshot_templates.sql
```diff
@@ -0,0 +1,26 @@
+-- +goose Up
+-- +goose NO TRANSACTION
+
+-- +goose StatementBegin
+-- Create snapshot_templates table to track which sandbox a snapshot template was created from
+CREATE TABLE IF NOT EXISTS "public"."snapshot_templates" (
+    env_id text NOT NULL PRIMARY KEY REFERENCES envs(id) ON DELETE CASCADE,
+    sandbox_id text NOT NULL,
+    created_at timestamp with time zone DEFAULT now()
+);
```
sitole: Should we maybe add a team here?

packages/api/internal/handlers/snapshot_template_create.go
```diff
@@ -0,0 +1,140 @@
...
+	telemetry.SetAttributes(ctx,
+		telemetry.WithTeamID(teamID.String()),
+		telemetry.WithSandboxID(sandboxID),
+	)
+
+	sandboxID = utils.ShortID(sandboxID)
+
+	// Parse optional request body
+	var body api.PostSandboxesSandboxIDSnapshotsJSONRequestBody
+	_ = c.ShouldBindJSON(&body)
```
sitole: Why now `utils.ParseBody`?

packages/db/queries/snapshots/list_team_snapshot_templates.sql
```diff
@@ -0,0 +1,41 @@
...
+    eb.id as build_id,
+    eb.vcpu,
+    eb.ram_mb,
+    eb.total_disk_size_mb,
+    eb.envd_version,
+    eb.kernel_version,
+    eb.firecracker_version,
+    eb.cluster_node_id,
+    COALESCE(ea.names, ARRAY[]::text[])::text[] AS names
+FROM "public"."envs" e
```
sitole, line 18: Looks like to get snapshots, you need to iterate over all team envs first, as there is a missing INDEX source. Does it make sense to optimize by doing INDEX(team_id, source)?

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -459,34 +552,112 @@ func (s *Server) Pause(ctx context.Context, in *orchestrator.SandboxPauseRequest
...
+}
+
+// acquireSandboxForSnapshot locks the pause mutex, retrieves the sandbox, removes it from the map,
+// and unlocks. Returns the sandbox for snapshotting or an error if not found.
+func (s *Server) acquireSandboxForSnapshot(ctx context.Context, sandboxID string) (*sandbox.Sandbox, error) {
+	s.pauseMu.Lock()
+
+	sbx, ok := s.sandboxes.Get(sandboxID)
+	if !ok {
+		s.pauseMu.Unlock()
```
sitole: Lets do defer unlock instead

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -497,26 +668,35 @@ func (s *Server) Pause(ctx context.Context, in *orchestrator.SandboxPauseRequest
...
+			sbxlogger.I(sbx).Warn(ctx, "failed to upload prefetch metadata", zap.Error(err))
+
+			return
+		}
+
+		s.templateCache.Invalidate(meta.Template.BuildID)
+
+		sbxlogger.I(sbx).Info(ctx, "prefetch mapping uploaded",
+			zap.Int("block_count", prefetchMapping.Count()))
+	}(context.WithoutCancel(ctx))
```
sitole: Let's do `ctx = context.WithoutCancel(ctx)` and remove goroutine param pass

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -400,52 +374,171 @@ func (s *Server) Pause(ctx context.Context, in *orchestrator.SandboxPauseRequest
...
+
+	// Setup lifecycle for the resumed sandbox
+	s.setupSandboxLifecycle(ctx, resumedSbx)
+
+	// Upload prefetch mapping in background
+	if prefetchErr == nil {
+		s.uploadPrefetchMappingAsync(ctx, resumedSbx, meta, prefetchData)
+	}
+
+	s.publishSandboxEvent(ctx, resumedSbx, "sandbox.checkpointed")
```
sitole: Use the event type constant instead of a hardcoded string.

packages/orchestrator/internal/template/metadata/prefetch.go
```diff
@@ -0,0 +1,70 @@
...
+	"fmt"
+	"slices"
+
+	"github.com/e2b-dev/infra/packages/orchestrator/internal/sandbox/block"
+	"github.com/e2b-dev/infra/packages/shared/pkg/storage"
+)
+
+// PrefetchEntriesToMapping converts a slice of PrefetchBlockEntry to MemoryPrefetchMapping.
+// Entries are sorted by access order. Returns nil if empty.
+func PrefetchEntriesToMapping(entries []block.PrefetchBlockEntry, blockSize int64) *MemoryPrefetchMapping {
```
sitole, line 15: This is just moved, right?

packages/api/internal/handlers/template_delete.go
```diff
@@ -83,29 +76,6 @@ func (a *APIStore) DeleteTemplatesTemplateID(c *gin.Context, aliasOrTemplateID a
...
-
-	// Delete all builds.
-	err = a.templateManager.DeleteBuilds(ctx, buildIds)
-	if err != nil {
-		telemetry.ReportCriticalError(ctx, "error when deleting template files from storage", err)
-	} else {
-		telemetry.ReportEvent(ctx, "deleted template from storage")
-	}
-
 	a.templateCache.InvalidateAllTags(templateID)
```
sitole, line 79: Lets sync on how we will communicate this

spec/openapi.yml
dobrac: the snapshots here is in the meaning of a resource, not an action

packages/api/internal/handlers/snapshot_template_create.go
dobrac: reworked

packages/db/queries/snapshots/list_team_snapshot_templates.sql
dobrac: added

packages/orchestrator/internal/template/metadata/prefetch.go
dobrac: yes

packages/orchestrator/internal/server/sandboxes.go
dobrac: updated

packages/orchestrator/internal/server/sandboxes.go
dobrac: updated

packages/orchestrator/internal/server/sandboxes.go
dobrac: updated

packages/api/internal/handlers/snapshot_template_create.go
```diff
@@ -0,0 +1,144 @@
...
+			opts.Tag = *tag
+		}
+
+		// Resolve alias using the cache — same pattern as template builds
+		aliasInfo, err := a.templateCache.ResolveAlias(ctx, identifier, teamInfo.Slug)
+		switch {
+		case err == nil && aliasInfo.TeamID == teamID:
+			// Alias exists and is owned by this team — reuse the template
+			opts.ExistingTemplateID = &aliasInfo.TemplateID
+		case err == nil || errors.Is(err, templatecache.ErrTemplateNotFound):
```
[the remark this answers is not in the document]
dobrac: this is correct

packages/shared/pkg/events/sandbox.go
```diff
@@ -12,11 +12,12 @@ const (
...
-	SandboxKilledEvent  = "sandbox.lifecycle.killed"
-	SandboxPausedEvent  = "sandbox.lifecycle.paused"
-	SandboxResumedEvent = "sandbox.lifecycle.resumed"
-	SandboxUpdatedEvent = "sandbox.lifecycle.updated"
+	SandboxCreatedEvent      = "sandbox.lifecycle.created"
+	SandboxKilledEvent       = "sandbox.lifecycle.killed"
+	SandboxPausedEvent       = "sandbox.lifecycle.paused"
+	SandboxResumedEvent      = "sandbox.lifecycle.resumed"
+	SandboxUpdatedEvent      = "sandbox.lifecycle.updated"
+	SandboxCheckpointedEvent = "sandbox.lifecycle.checkpointed"
```
sitole, line 20: Please create an issue to verify that the event and webhooks work correctly with the new event type (this is the first additional event type since we introduced them). We also need to update the docs.

A can do that, just assign it to me :ty:

dobrac: issue created

packages/db/queries/snapshots/list_team_snapshot_templates.sql
```diff
@@ -0,0 +1,41 @@
...
+    SELECT b.*
+    FROM "public"."env_build_assignments" ba
+    JOIN "public"."env_builds" b ON b.id = ba.build_id
+    WHERE ba.env_id = e.id AND b.status IN ('success', 'uploaded', 'ready')
+    ORDER BY ba.created_at DESC
+    LIMIT 1
+) eb ON TRUE
+LEFT JOIN LATERAL (
+    SELECT ARRAY_AGG(CASE WHEN namespace IS NOT NULL THEN namespace || '/' || alias ELSE alias END ORDER BY alias) AS names
+    FROM "public"."env_aliases"
```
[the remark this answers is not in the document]
dobrac: its okay for now. I might improve this a bit in another PR before GA

packages/api/internal/handlers/snapshot_template_create.go
```diff
@@ -0,0 +1,144 @@
...
+		telemetry.ReportCriticalError(ctx, "Error creating snapshot template", err, telemetry.WithSandboxID(sandboxID))
+		a.sendAPIStoreError(c, http.StatusInternalServerError, "Error creating snapshot template")
+
+		return
+	}
+
+	// Invalidate cached tombstone if a new alias was created
+	if opts.Alias != nil && opts.Namespace != nil && opts.ExistingTemplateID == nil {
+		a.templateCache.InvalidateAlias(opts.Namespace, *opts.Alias)
+	}
```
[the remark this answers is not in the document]
dobrac: will be addressed in another PR

PR #1801, opened 2026-01-28
Add S3-style multi-part upload mechanism for large file parallel upload

mishushakov:
> [!NOTE]
> **Medium Risk**
> Introduces new file-upload endpoints and in-memory session tracking with concurrent writes/renames, which can impact disk usage and correctness under race conditions if edge cases are missed.
> 
> **Overview**
> Adds an S3-style multipart upload flow to `envd` so large files can be uploaded in parallel: a client can `init` an upload session (with limits on sessions/size/parts), `PUT` individual parts directly into a preallocated temp file, then either `complete` to atomically rename into place or `abort` to clean up. The OpenAPI spec and generated clients/servers are updated accordingly (including new 409/429/404 error responses), and a comprehensive test suite is added to cover ordering, missing parts, abort/complete behavior, and resource-limit enforcement.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1005e97c3b583edf9668cc1f2a2d87ea464bf93c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/envd/internal/api/multipart_upload.go
+++ b/packages/envd/internal/api/multipart_upload.go
@@ -382,6 +382,12 @@ func (a *API) PutFilesUploadUploadId(w http.ResponseWriter, r *http.Request, upl
 		}
 	}()
 
+	// Limit the request body to expectedSize+1 so the server does not buffer
+	// an arbitrarily large oversized body. The +1 allows the trailing-byte
+	// check below to detect excess data without triggering MaxBytesError
+	// during io.CopyN itself (which reads exactly expectedSize bytes).
+	r.Body = http.MaxBytesReader(w, r.Body, expectedSize+1)
+
 	// Stream the part data directly to the file at offset without buffering the
 	// entire part in memory. OffsetWriter + CopyN uses a small internal buffer
 	// (~32KB) instead of reading the full part into a single allocation.


packages/envd/internal/api/multipart_upload.go
```diff
@@ -0,0 +1,463 @@
...
+	sessionCount := len(a.uploads)
+	a.uploadsLock.RUnlock()
+	if sessionCount >= maxUploadSessions {
+		a.logger.Error().Str(string(logs.OperationIDKey), operationID).Int("maxSessions", maxUploadSessions).Msg("too many concurrent upload sessions")
+		jsonError(w, http.StatusTooManyRequests, fmt.Errorf("too many concurrent upload sessions (max %d)", maxUploadSessions))
+
+		return
+	}
+
+	// Validate signing if needed
```
mishushakov: not sure about this part

packages/envd/spec/envd.yaml
```diff
@@ -125,6 +125,144 @@ paths:
...
+            type: string
+      responses:
+        "200":
+          description: Upload completed successfully
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/MultipartUploadComplete"
+        "404":
+          $ref: "#/components/responses/UploadNotFound"
```
djeebus, line 278: Might also need a `UploadIncomplete` error; 400?

packages/envd/spec/envd.yaml
```diff
@@ -125,6 +125,144 @@ paths:
...
+              format: binary
+      responses:
+        "200":
+          description: Part uploaded successfully
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/MultipartUploadPart"
+        "400":
+          $ref: "#/components/responses/InvalidPath"
```
djeebus, line 222: What does "InvalidPath" mean? We also probably need a "PartAlreadyUploadedError", and/or "PartUploadInProgress"

packages/envd/internal/api/store.go
```diff
@@ -23,15 +42,53 @@ type API struct {
...
+						s.DestFile.Close()
+						if err := os.Remove(s.FilePath); err != nil && !os.IsNotExist(err) {
+							a.logger.Warn().Err(err).Str("filePath", s.FilePath).Msg("failed to cleanup expired upload file")
+						}
+					}(session)
+					a.logger.Info().Str("uploadId", uploadID).Msg("cleaned up expired multipart upload session")
+				}
+			}
+		}
+		a.uploadsLock.Unlock()
```
djeebus: it might read better if this were in a different function. then the ticket would just be:

```
for range ticker.C {
	a.checkUploads()
}
```

and you could use defer to ensure lock/unlock

packages/envd/internal/api/store.go
```diff
@@ -23,15 +42,53 @@ type API struct {
...
+			if now.Sub(session.CreatedAt) > uploadSessionTTL {
+				// Mark as completed to prevent races
+				if session.completed.CompareAndSwap(false, true) {
+					delete(a.uploads, uploadID)
+					// Close file handle and remove file in background
+					go func(s *MultipartUploadSession) {
+						s.DestFile.Close()
+						if err := os.Remove(s.FilePath); err != nil && !os.IsNotExist(err) {
+							a.logger.Warn().Err(err).Str("filePath", s.FilePath).Msg("failed to cleanup expired upload file")
+						}
```
djeebus: dunno if you'd prefer this, but sometimes it's easier to read:

```
if err := os.Remove(s.FilePath); ignoreNotExist(err) != nil {
	a.logger.Warn(...)
}
```

```
func ignoreNotExist(err error) error {
	if os.NotExist(err) {
		return nil
	}

	return err
}
```

packages/envd/internal/api/multipart_upload.go
```diff
@@ -0,0 +1,471 @@
...
+	}
+
+	// Create upload ID
+	uploadID := uuid.New().String()
+
+	// Calculate number of parts
+	numParts := int((body.TotalSize + body.PartSize - 1) / body.PartSize)
+	if numParts == 0 && body.TotalSize == 0 {
+		numParts = 0 // Empty file, no parts needed
+	}
```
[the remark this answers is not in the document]
djeebus: This is a pretty valid point. We should probably have a minimum part size, or maximum parts, or both.

packages/envd/internal/api/multipart_upload.go
```diff
@@ -0,0 +1,471 @@
...
+		destFile.Close()
+		os.Remove(filePath)
+		a.logger.Error().Err(err).Str(string(logs.OperationIDKey), operationID).Msg("error changing file ownership")
+		jsonError(w, http.StatusInternalServerError, fmt.Errorf("error changing file ownership: %w", err))
+
+		return
+	}
+
+	// Create upload ID
+	uploadID := uuid.New().String()
```
djeebus: Pretty sure there's a uuid.NewString()

packages/envd/internal/api/multipart_upload.go
```diff
@@ -0,0 +1,471 @@
...
+
+	operationID := logs.AssignOperationID()
+
+	// Validate uploadId is a valid UUID to prevent path traversal
+	if _, err := uuid.Parse(uploadId); err != nil {
+		a.logger.Error().Err(err).Str(string(logs.OperationIDKey), operationID).Str("uploadId", uploadId).Msg("invalid upload ID format")
+		jsonError(w, http.StatusBadRequest, fmt.Errorf("invalid upload ID format: must be a valid UUID"))
+
+		return
+	}
```
djeebus: Don't think you need this, as `session, exists := a.uploads[uploadId]` does a better job of validating whether it's a real uploadId or not.

packages/envd/internal/api/multipart_upload.go
```diff
@@ -0,0 +1,471 @@
...
+
+	// Check if session is already being completed/aborted
+	if session.completed.Load() {
+		a.logger.Error().Str(string(logs.OperationIDKey), operationID).Str("uploadId", uploadId).Msg("upload session is already completing")
+		jsonError(w, http.StatusConflict, fmt.Errorf("upload session is already completing or aborted"))
+
+		return
+	}
```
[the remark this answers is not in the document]
djeebus: This feels important.

packages/envd/internal/api/multipart_upload.go
```diff
@@ -0,0 +1,471 @@
...
+
+	partNumber := params.Part
```
djeebus: Can we make `Part` a `uint` so that parsing will fail, and we can avoid some checks?

packages/envd/internal/api/store.go
```diff
@@ -12,6 +15,22 @@ import (
...
 )
 
+// MultipartUploadSession tracks an in-progress multipart upload
+type MultipartUploadSession struct {
+	UploadID     string
+	FilePath     string   // Final destination path
+	DestFile     *os.File // Open file handle for direct writes
+	TotalSize    int64    // Total expected file size
+	PartSize     int64    // Size of each part (except possibly last)
+	NumParts     int      // Total number of expected parts
```
djeebus: We could make these uints to avoid some failure modes.

packages/envd/internal/api/multipart_upload.go
mishushakov: I think I can just add minimum: 0 to the spec instead, it should validate automatically

packages/envd/internal/api/store.go
```diff
@@ -37,17 +65,59 @@ type API struct {
...
+	defer ticker.Stop()
+
+	for range ticker.C {
+		a.removeExpiredSessions()
+	}
+}
+
+func (a *API) removeExpiredSessions() {
+	a.uploadsLock.Lock()
+	defer a.uploadsLock.Unlock()
```
mishushakov: I am bit worried that it would enable race condition where I cannot do upload while the flush is happening.
am thinking maybe we can remove the expired uploads logic altogether?

packages/envd/internal/api/multipart_upload.go
```diff
@@ -0,0 +1,428 @@
...
+		FilePath:     filePath,
+		DestFile:     destFile,
+		TotalSize:    body.TotalSize,
+		PartSize:     body.PartSize,
+		NumParts:     numParts,
+		UID:          uid,
+		GID:          gid,
+		PartsWritten: make(map[uint]bool),
+		CreatedAt:    time.Now(),
+	}
```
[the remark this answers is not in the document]
mishushakov: I think it should be done by openapi middleware

mishushakov:
@cursor review
@claude review

packages/envd/internal/api/multipart_upload.go
```diff
@@ -0,0 +1,555 @@
...
+		Msg("multipart upload initialized")
+
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
+	if err := json.NewEncoder(w).Encode(MultipartUploadInit{
+		UploadId: uploadID,
+	}); err != nil {
+		a.logger.Error().Err(err).Str(string(logs.OperationIDKey), operationID).Msg("failed to encode response")
+	}
+}
```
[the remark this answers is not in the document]
mishushakov: this is alright, they should remain incomplete

packages/envd/internal/api/multipart_upload.go
```diff
@@ -0,0 +1,569 @@
...
+
+	if err := os.Rename(session.TempPath, session.FilePath); err != nil {
+		if rmErr := ignoreNotExist(os.Remove(session.TempPath)); rmErr != nil {
+			a.logger.Warn().Err(rmErr).Str(string(logs.OperationIDKey), operationID).Msg("failed to remove temp file after rename error")
+		}
+		a.logger.Error().Err(err).Str(string(logs.OperationIDKey), operationID).Msg("error renaming temp file to destination")
+		jsonError(w, http.StatusInternalServerError, fmt.Errorf("error renaming temp file to destination: %w", err))
+
+		return
+	}
```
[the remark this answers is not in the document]
mishushakov: I think it's fine - it's just a failed upload in that case

mishushakov:
@claude review

PR #1804, opened 2026-01-29
Add volume CRUD and API usage

djeebus:
This implements everything up to actually mounting them in the sandbox

---

> [!NOTE]
> **Medium Risk**
> Adds new API surface and database writes plus orchestrator filesystem operations (mkdir/rm) controlled by config and feature flags; incorrect configuration or validation gaps could impact data integrity or delete unintended paths.
> 
> **Overview**
> Adds first-class **persistent volume** support end-to-end: a new `volumes` table and API endpoints to create/list/get/delete volumes (guarded by a feature flag), plus orchestrator gRPC support to create/delete the underlying volume directories from configured mount roots. Sandbox creation now accepts `volumeMounts`, validates requested mounts against the team’s volumes (name existence, unique/clean absolute paths) and forwards them through the API→orchestrator request chain, while local/CI configs are updated to provide a default volume type and test mount path.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit fea80f53c7d4177ec2b592fce6f2391ab63de458. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/envd/internal/api/upload.go
+++ b/packages/envd/internal/api/upload.go
@@ -148,6 +148,36 @@ func resolvePath(part *multipart.Part, paths *UploadSuccess, u *user.User, defau
 	return filePath, nil
 }
 
+func (a *API) handlePart(r *http.Request, part *multipart.Part, paths UploadSuccess, u *user.User, uid, gid int, operationID string, params PostFilesParams) (*EntryInfo, int, error) {
+	defer part.Close()
+
+	if part.FormName() != "file" {
+		return nil, http.StatusOK, nil
+	}
+
+	filePath, err := resolvePath(part, &paths, u, a.defaults.Workdir, params)
+	if err != nil {
+		return nil, http.StatusBadRequest, err
+	}
+
+	logger := a.logger.
+		With().
+		Str(string(logs.OperationIDKey), operationID).
+		Str("event_type", "file_processing").
+		Logger()
+
+	status, err := processFile(r, filePath, part, uid, gid, logger)
+	if err != nil {
+		return nil, status, err
+	}
+
+	return &EntryInfo{
+		Path: filePath,
+		Name: filepath.Base(filePath),
+		Type: File,
+	}, http.StatusOK, nil
+}
+
 func (a *API) PostFiles(w http.ResponseWriter, r *http.Request, params PostFilesParams) {
 	defer r.Body.Close()
 
@@ -237,38 +267,18 @@ func (a *API) PostFiles(w http.ResponseWriter, r *http.Request, params PostFiles
 			break
 		}
 
-		if part.FormName() == "file" {
-			filePath, err := resolvePath(part, &paths, u, a.defaults.Workdir, params)
-			if err != nil {
-				errorCode = http.StatusBadRequest
-				errMsg = err
-				jsonError(w, errorCode, errMsg)
+		entry, status, err := a.handlePart(r, part, paths, u, uid, gid, operationID, params)
+		if err != nil {
+			errorCode = status
+			errMsg = err
+			jsonError(w, errorCode, errMsg)
 
-				return
-			}
-
-			logger := a.logger.
-				With().
-				Str(string(logs.OperationIDKey), operationID).
-				Str("event_type", "file_processing").
-				Logger()
-			status, err := processFile(r, filePath, part, uid, gid, logger)
-			if err != nil {
-				errorCode = status
-				errMsg = err
-				jsonError(w, errorCode, errMsg)
-
-				return
-			}
-
-			paths = append(paths, EntryInfo{
-				Path: filePath,
-				Name: filepath.Base(filePath),
-				Type: File,
-			})
+			return
 		}
 
-		part.Close()
+		if entry != nil {
+			paths = append(paths, *entry)
+		}
 	}
 
 	data, err := json.Marshal(paths)
--- a/packages/envd/internal/services/process/handler/handler.go
+++ b/packages/envd/internal/services/process/handler/handler.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"os/exec"
 	"os/user"
+	"strconv"
 	"sync"
 	"syscall"
 
@@ -78,16 +79,26 @@ func New(
 		return nil, connect.NewError(connect.CodeInternal, err)
 	}
 
+	groups := []uint32{gid}
+	if gids, err := user.GroupIds(); err != nil {
+		logger.Warn().Err(err).Str("user", user.Username).Msg("failed to get supplementary groups")
+	} else {
+		for _, g := range gids {
+			if parsed, err := strconv.ParseUint(g, 10, 32); err == nil {
+				groups = append(groups, uint32(parsed))
+			}
+		}
+	}
+
 	cgroupFD, ok := cgroupManager.GetFileDescriptor(getProcType(req))
 
 	cmd.SysProcAttr = &syscall.SysProcAttr{
 		UseCgroupFD: ok,
 		CgroupFD:    cgroupFD,
 		Credential: &syscall.Credential{
-			Uid:         uid,
-			Gid:         gid,
-			Groups:      []uint32{gid},
-			NoSetGroups: true,
+			Uid:    uid,
+			Gid:    gid,
+			Groups: groups,
 		},
 	}
 
--- a/packages/orchestrator/cmd/clean-nfs-cache/main.go
+++ b/packages/orchestrator/cmd/clean-nfs-cache/main.go
@@ -35,22 +35,24 @@ func main() {
 	var err error
 	var opts cleaner.Options
 
+	var lp telemetry.LogProvider
+	opts, log, lp, err = preRun(ctx)
+	if err != nil {
+		fmt.Println("NFS cache cleaner failed:", err)
+		if lp != nil {
+			lp.Shutdown(ctx)
+		}
+		os.Exit(1)
+	}
+
 	defer func() {
 		if err != nil {
-			if log != nil {
-				log.Error(ctx, "NFS cache cleaner failed", zap.Error(err))
-			} else {
-				fmt.Println("NFS cache cleaner failed:", err)
-			}
-			os.Exit(1)
+			log.Error(ctx, "NFS cache cleaner failed", zap.Error(err))
+			defer os.Exit(1)
 		}
+		log.Sync()
+		lp.Shutdown(ctx)
 	}()
-
-	opts, log, err = preRun(ctx)
-	if err != nil {
-		return
-	}
-	defer log.Sync()
 
 	start := time.Now()
 	log.Info(ctx, "starting",
@@ -102,7 +104,7 @@ func main() {
 		zap.Duration("std_deviation", sd.Round(time.Second)))
 }
 
-func preRun(ctx context.Context) (cleaner.Options, logger.Logger, error) {
+func preRun(ctx context.Context) (cleaner.Options, logger.Logger, telemetry.LogProvider, error) {
 	var opts cleaner.Options
 	var featureFlagPresent bool
 
@@ -121,18 +123,18 @@ func preRun(ctx context.Context) (cleaner.Options, logger.Logger, error) {
 
 	args := os.Args[1:] // skip the command name
 	if err := flags.Parse(args); err != nil {
-		return opts, nil, fmt.Errorf("could not parse flags: %w", err)
+		return opts, nil, nil, fmt.Errorf("could not parse flags: %w", err)
 	}
 
 	args = flags.Args()
 	if len(args) != 1 {
-		return opts, nil, ErrUsage
+		return opts, nil, nil, ErrUsage
 	}
 	opts.Path = args[0]
 
 	ffc, err := featureflags.NewClient()
 	if err != nil {
-		return opts, nil, err
+		return opts, nil, nil, err
 	}
 	defer ffc.Close(ctx)
 
@@ -161,10 +163,13 @@ func preRun(ctx context.Context) (cleaner.Options, logger.Logger, error) {
 	}
 
 	var cores []zapcore.Core
+	logProvider := telemetry.NewNoopLogProvider()
 	if opts.OtelCollectorEndpoint != "" {
-		otelCore, err := newOtelCore(ctx, opts.OtelCollectorEndpoint)
+		var otelCore zapcore.Core
+		var err error
+		otelCore, logProvider, err = newOtelCore(ctx, opts.OtelCollectorEndpoint)
 		if err != nil {
-			return opts, nil, fmt.Errorf("failed to create otel logger: %w", err)
+			return opts, nil, nil, fmt.Errorf("failed to create otel logger: %w", err)
 		}
 		cores = append(cores, otelCore)
 	}
@@ -188,7 +193,11 @@ func preRun(ctx context.Context) (cleaner.Options, logger.Logger, error) {
 			diskInfo, err = cleaner.GetDiskInfo(ctx, opts.Path)
 		})
 		if err != nil {
-			return opts, nil, fmt.Errorf("could not get disk info: %w", err)
+			if logProvider != nil {
+				logProvider.Shutdown(ctx)
+			}
+
+			return opts, nil, nil, fmt.Errorf("could not get disk info: %w", err)
 		}
 		targetDiskUsage := uint64(opts.TargetDiskUsagePercent / 100 * float64(diskInfo.Total))
 		if uint64(diskInfo.Used) > targetDiskUsage {
@@ -196,29 +205,25 @@ func preRun(ctx context.Context) (cleaner.Options, logger.Logger, error) {
 		}
 	}
 
-	return opts, l, nil
+	return opts, l, logProvider, nil
 }
 
-func newOtelCore(ctx context.Context, endpoint string) (zapcore.Core, error) {
+func newOtelCore(ctx context.Context, endpoint string) (zapcore.Core, telemetry.LogProvider, error) {
 	nodeID := env.GetNodeID()
 	serviceInstanceID := uuid.NewString()
 
 	resource, err := telemetry.GetResource(ctx, nodeID, serviceName, commitSHA, serviceVersion, serviceInstanceID)
 	if err != nil {
-		return nil, fmt.Errorf("failed to create resource: %w", err)
+		return nil, nil, fmt.Errorf("failed to create resource: %w", err)
 	}
 
-	logsExporter, err := telemetry.NewLogExporter(ctx,
-		otlploggrpc.WithEndpoint(endpoint),
-	)
+	logProvider, err := telemetry.NewLogProvider(ctx, resource, otlploggrpc.WithEndpoint(endpoint))
 	if err != nil {
-		return nil, fmt.Errorf("failed to create logs exporter: %w", err)
+		return nil, nil, err
 	}
+	otelCore := logger.GetOTELCore(logProvider, serviceName)
 
-	loggerProvider := telemetry.NewLogProvider(logsExporter, resource)
-	otelCore := logger.GetOTELCore(loggerProvider, serviceName)
-
-	return otelCore, nil
+	return otelCore, logProvider, nil
 }
 
 func standardDeviation(accessed []time.Duration) (mean, stddev time.Duration) {
--- a/packages/shared/pkg/telemetry/logs.go
+++ b/packages/shared/pkg/telemetry/logs.go
@@ -6,44 +6,42 @@ import (
 
 	"go.opentelemetry.io/otel/exporters/otlp/otlplog/otlploggrpc"
 	"go.opentelemetry.io/otel/log"
+	nooplog "go.opentelemetry.io/otel/log/noop"
 	sdklog "go.opentelemetry.io/otel/sdk/log"
 	"go.opentelemetry.io/otel/sdk/resource"
 	"google.golang.org/grpc/encoding/gzip"
 )
 
-type noopLogExporter struct{}
+// LogProvider extends log.LoggerProvider with a Shutdown method
+// so the batch processor and exporter are properly flushed on exit.
+type LogProvider interface {
+	log.LoggerProvider
+	Shutdown(ctx context.Context) error
+}
 
-func (noopLogExporter) Export(context.Context, []sdklog.Record) error { return nil }
+type noopLogProvider struct{ nooplog.LoggerProvider }
 
-func (noopLogExporter) Shutdown(context.Context) error { return nil }
+func (noopLogProvider) Shutdown(context.Context) error { return nil }
 
-func (noopLogExporter) ForceFlush(context.Context) error { return nil }
+func NewNoopLogProvider() LogProvider { return noopLogProvider{} }
 
-func NewLogExporter(ctx context.Context, extraOption ...otlploggrpc.Option) (sdklog.Exporter, error) {
+func NewLogProvider(ctx context.Context, res *resource.Resource, extraOpts ...otlploggrpc.Option) (LogProvider, error) {
 	opts := []otlploggrpc.Option{
 		otlploggrpc.WithInsecure(),
 		otlploggrpc.WithEndpoint(otelCollectorGRPCEndpoint),
 		otlploggrpc.WithCompressor(gzip.Name),
 	}
-	opts = append(opts, extraOption...)
+	opts = append(opts, extraOpts...)
 
-	logsExporter, err := otlploggrpc.New(
-		ctx,
-		opts...,
-	)
+	exporter, err := otlploggrpc.New(ctx, opts...)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create logs exporter: %w", err)
 	}
 
-	return logsExporter, nil
-}
-
-func NewLogProvider(logsExporter sdklog.Exporter, res *resource.Resource) log.LoggerProvider {
-	logsProcessor := sdklog.NewBatchProcessor(logsExporter)
-	logsProvider := sdklog.NewLoggerProvider(
+	p := sdklog.NewLoggerProvider(
 		sdklog.WithResource(res),
-		sdklog.WithProcessor(logsProcessor),
+		sdklog.WithProcessor(sdklog.NewBatchProcessor(exporter)),
 	)
 
-	return logsProvider
+	return p, nil
 }
--- a/packages/shared/pkg/telemetry/main.go
+++ b/packages/shared/pkg/telemetry/main.go
@@ -8,12 +8,9 @@ import (
 
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetricgrpc"
-	"go.opentelemetry.io/otel/log"
-	noopLogs "go.opentelemetry.io/otel/log/noop"
 	"go.opentelemetry.io/otel/metric"
 	noopMetric "go.opentelemetry.io/otel/metric/noop"
 	"go.opentelemetry.io/otel/propagation"
-	sdklog "go.opentelemetry.io/otel/sdk/log"
 	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
 	sdktrace "go.opentelemetry.io/otel/sdk/trace"
 	"go.opentelemetry.io/otel/trace"
@@ -28,8 +25,7 @@ type Client struct {
 	SpanExporter    sdktrace.SpanExporter
 	TracerProvider  trace.TracerProvider
 	TracePropagator propagation.TextMapPropagator
-	LogsExporter    sdklog.Exporter
-	LogsProvider    log.LoggerProvider
+	LogsProvider    LogProvider
 }
 
 func New(ctx context.Context, nodeID, serviceName, serviceCommit, serviceVersion, serviceInstanceID string) (*Client, error) {
@@ -66,12 +62,10 @@ func New(ctx context.Context, nodeID, serviceName, serviceCommit, serviceVersion
 	otel.SetMeterProvider(meterProvider)
 
 	// Setup logging
-	logsExporter, err := NewLogExporter(ctx)
+	logProvider, err := NewLogProvider(ctx, res)
 	if err != nil {
-		return nil, fmt.Errorf("failed to create logs exporter: %w", err)
+		return nil, fmt.Errorf("failed to create log provider: %w", err)
 	}
-
-	logsProvider := NewLogProvider(logsExporter, res)
 
 	// Setup tracing
 	spanExporter, err := NewSpanExporter(ctx)
@@ -92,8 +86,7 @@ func New(ctx context.Context, nodeID, serviceName, serviceCommit, serviceVersion
 		SpanExporter:    spanExporter,
 		TracerProvider:  tracerProvider,
 		TracePropagator: propagator,
-		LogsExporter:    logsExporter,
-		LogsProvider:    logsProvider,
+		LogsProvider:    logProvider,
 	}, nil
 }
 
@@ -109,8 +102,8 @@ func (t *Client) Shutdown(ctx context.Context) error {
 			errs = append(errs, err)
 		}
 	}
-	if t.LogsExporter != nil {
-		if err := t.LogsExporter.Shutdown(ctx); err != nil {
+	if t.LogsProvider != nil {
+		if err := t.LogsProvider.Shutdown(ctx); err != nil {
 			errs = append(errs, err)
 		}
 	}
@@ -125,7 +118,6 @@ func NewNoopClient() *Client {
 		SpanExporter:    &noopSpanExporter{},
 		TracerProvider:  noopTrace.NewTracerProvider(),
 		TracePropagator: propagation.NewCompositeTextMapPropagator(),
-		LogsExporter:    &noopLogExporter{},
-		LogsProvider:    noopLogs.NewLoggerProvider(),
+		LogsProvider:    NewNoopLogProvider(),
 	}
 }


packages/api/internal/handlers/volume_create.go
```diff
@@ -0,0 +1,101 @@
...
+
+	telemetry.ReportEvent(ctx, "Parsed body")
+
+	// validate body
+	if !isValidVolumeName(body.Name) {
+		a.sendAPIStoreError(c, http.StatusBadRequest, "Invalid volume name")
+		telemetry.ReportError(ctx, "invalid volume name", nil)
+
+		return
+	}
```
sitole, line 63: Not hard to resolve this. I'm just thinking about how we will need to set volumes for different clusters, where each cluster defines its own paths.

Would you mind moving this validation to api/clusters?
For local, we can use API envs; for remote, deny by default, as not supported?

packages/api/internal/handlers/volume_update.go
```diff
@@ -0,0 +1,90 @@
...
+
+		c.JSON(http.StatusOK, result)
+	}
+}
+
+var validVolumeNameRegex = regexp.MustCompile(`^[a-zA-Z0-9_-]+$`)
+
+func isValidVolumeName(name string) bool {
+	return validVolumeNameRegex.MatchString(name)
+}
```
sitole: Can we add tests for this?

packages/api/internal/orchestrator/create_instance.go
```diff
@@ -333,3 +355,70 @@ func (o *Orchestrator) CreateSandbox(
...
+func (o *Orchestrator) convertVolumeMounts(
+	ctx context.Context,
+	teamID uuid.UUID,
+	volumeMounts []api.SandboxVolumeMount,
+) ([]*orchestrator.SandboxVolumeMount, error) {
+	// dedupe mounted volumes
+	uniqueVolumeNames := make(map[string]struct{}, len(volumeMounts))
+	for _, v := range volumeMounts {
+		uniqueVolumeNames[v.Name] = struct{}{}
+	}
```
sitole: You are assuming the array is always `len(volumeMounts)`, and nothing is deduplicated from `volumeMounts`

packages/db/queries/volumes/volumes.sql
```diff
@@ -0,0 +1,21 @@
...
+VALUES (@team_id, @volume_type, @name)
+RETURNING *;
+
+-- name: GetVolume :one
+SELECT * FROM volumes WHERE id = @volume_id AND team_id = @team_id;
+
+-- name: GetVolumesByName :many
+SELECT * FROM volumes WHERE team_id = @team_id AND name IN (
+    SELECT UNNEST(@volume_names::text[])
+);
```
sitole, line 12: How can you have a team with multiple volumes all named the same?

packages/envd/spec/envd.yaml
```diff
@@ -284,3 +288,15 @@ components:
...
           description: Total disk space in bytes
+    VolumeMount:
+      type: object
+      description: Volume
+      additionalProperties: false
+      properties:
+        nfs_target:
+          type: string
+        path:
+          type: string
```
sitole, line 300: It's not part of this PR, but how will restricted access work here?

packages/api/internal/handlers/volume_create.go
djeebus: You're thinking clusters will advertise which volume types they support, and the API will simply query the cluster? That makes sense, and simplifies the configuration. 

Does that also imply that volumes will have to be cluster-specific? I'll need to add a column to the database as well then, correct?

packages/api/internal/orchestrator/create_instance.go
djeebus: Not sure I understand you. If you're talking about `len(volumeMounts)`, that's an upper bound on the capacity of `uniqueVolumeNames`, but it does not set its length. If it ends up being smaller due to deduplication, that's fine. It sacrifices some memory to avoid multiple allocation attempts, which are more expensive than one slightly larger request.

packages/db/queries/volumes/volumes.sql
djeebus: You cannot. Volume names must be unique across a team.

packages/envd/spec/envd.yaml
djeebus: The nfs server that it connects to will chroot the requests to the requested volume; other volumes cannot be accessed without mounting, and the mount request will validate that they are allowed to access the volume.

packages/db/queries/volumes/volumes.sql
sitole: Sorry i read `@volume_names` as @volume_name`.  Okay, makes sense now.

packages/api/internal/handlers/volume_create.go
sitole: > You're thinking clusters will advertise which volume types they support, and the API will simply query the cluster? That makes sense, and simplifies the configuration.

Yes, I was thinking about it this way.

> Does that also imply that volumes will have to be cluster-specific? I'll need to add a column to the database as well then, correct?

I don't think you need to add a cluster there. We're assuming the team is assigned to the cluster, and that assignment can't change, so you have a non-direct connection to the cluster via the team.

packages/envd/spec/envd.yaml
sitole: Nice, just interested - do we have this somewhere in progress?

spec/openapi.yml
```diff
@@ -2893,3 +2948,88 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Volume"
+        "401":
+          $ref: "#/components/responses/401"
+        "404":
+          $ref: "#/components/responses/404"
+        "500":
+          $ref: "#/components/responses/500"
```
mishushakov: not sure if we want to update the volume

spec/openapi.yml
```diff
@@ -2893,3 +2948,88 @@ paths:
...
+              schema:
+                $ref: "#/components/schemas/Volume"
+        "400":
+          $ref: "#/components/responses/400"
+        "401":
+          $ref: "#/components/responses/401"
+        "500":
+          $ref: "#/components/responses/500"
+
+  /volumes/{volumeID}:
```
mishushakov, line 3052: there's no delete?

mishushakov:
i find the persistent volume types weird, because it is unclear which are available and how each works. i think if we only support one (nfs bucket), maybe we should remove the code? wdyt?

packages/api/internal/cfg/model.go
```diff
@@ -56,5 +64,24 @@ func Parse() (Config, error) {
...
+		for name, path := range config.PersistentVolumeTypes {
+			info, err := os.Stat(path)
+			if err != nil {
+				return config, fmt.Errorf("failed to stat persistent volume mount %q (%q): %w", name, path, err)
+			}
+
+			if !info.IsDir() {
+				return config, fmt.Errorf("persistent volume mount %q (%q) is not a directory", name, path)
+			}
+		}
```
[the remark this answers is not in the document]
djeebus: out of scope

spec/openapi.yml
djeebus: Fair, I'll remove it. if we end up adding something meaningful to update (size limit, for example) later, we can always add this back.

spec/openapi.yml
djeebus: It's little dangerous to implement now (what do we do with existing files? make them delete them first? Lock while we delete so no new files get added? require admin permissions? only from the UI?) so I just punted on it. We can always remove it manually if they need it done, and when we think this through, we can implement it later.

packages/api/internal/handlers/volume_create.go
```diff
@@ -0,0 +1,129 @@
...
+	if !ok {
+		telemetry.ReportCriticalError(ctx, fmt.Sprintf("cluster with ID '%s' not found", team.ClusterID), nil)
+		a.sendAPIStoreError(c, http.StatusInternalServerError, fmt.Sprintf("cluster with id %s not found", team.ClusterID))
+
+		return
+	}
+
+	if volumeTypes, err := cluster.GetResources().GetVolumeTypes(ctx); err != nil {
+		a.sendAPIStoreError(c, http.StatusInternalServerError, "failed to get volume types for cluster")
+		telemetry.ReportCriticalError(ctx, "failed to get volume types for cluster", nil)
```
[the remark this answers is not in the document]
sitole: Yep, lets fill it with err

packages/api/internal/clusters/cluster.go
```diff
@@ -40,7 +41,7 @@ type Cluster struct {
 
 	instances       *smap.Map[*Instance]
 	synchronization *synchronization.Synchronize[discovery.Item, *Instance]
-	resources       ClusterResource
+	Resources       ClusterResource
```
sitole: Can we help create a cluster with resources instead of making it available outside the module? If not, then remove the getter and use it directly everywhere.

spec/openapi.yml
mishushakov: i am thinking if you've got a use-case for ephemeral volumes you'd need a way to delete them after every run

djeebus: You and @ValentaTomas convinced me, I brought back volume deletion

packages/api/internal/handlers/sandbox_create.go
```diff
@@ -217,6 +244,140 @@ func (a *APIStore) PostSandboxes(c *gin.Context) {
...
+	return results
+}
+
+var ErrVolumeMountsDisabled = errors.New("volume mounts are not enabled")
+
+type featureFlagsClient interface {
+	BoolFlag(ctx context.Context, flagName featureflags.BoolFlag, contexts ...ldcontext.Context) bool
+}
+
+func convertVolumeMounts(
```
levb: nit: change the name to reveal that the intention is to get Orchestrator-specific volume mounts?

packages/api/internal/handlers/sandbox_create.go
```diff
@@ -217,6 +244,140 @@ func (a *APIStore) PostSandboxes(c *gin.Context) {
 	c.JSON(http.StatusCreated, &sbx)
 }
 
+func toSet[TItem any, TKey comparable](items []TItem, fn func(TItem) TKey) map[TKey]struct{} {
```
levb: 0/5 real nit-picking, but since it returns a go map, maybe `toMap` or `asMap`

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -234,6 +236,20 @@ func (s *Server) Create(ctx context.Context, req *orchestrator.SandboxCreateRequ
 	}, nil
 }
 
+func convertVolumeMounts(volumeMounts []*orchestrator.SandboxVolumeMount) []sandbox.VolumeMountConfig {
```
levb: ditto, the opposite of the otehr `convertVolumeMounts`?

packages/shared/pkg/feature-flags/context.go
```diff
@@ -103,7 +113,10 @@ func mergeContexts(ctx context.Context, contexts []ldcontext.Context) ldcontext.
 
 	contexts = flattenContexts(contexts)
 
-	contexts = removeUndefined(contexts)
+	contexts = filterContexts(contexts,
```
levb: what prompted this change, curious, does not seem related directly to the PR?

packages/api/internal/handlers/sandbox_create.go
djeebus: I was being too clever here. Too many lines of code to do a simple thing. I'll clean it up :saluting_face:

packages/shared/pkg/feature-flags/context.go
djeebus: Oh, good callout. This should be added to a different PR. Launch darkly does not appreciate when we create kinds with no name, which it turns out we do a lot. This was too avoid the logged warnings, but it doesn't need to be here. Thanks for keeping me honest!

packages/api/internal/handlers/sandbox_create.go
```diff
@@ -218,6 +248,179 @@ func (a *APIStore) PostSandboxes(c *gin.Context) {
...
+			continue
+		}
+		usedPaths[v.Path] = struct{}{}
+
+		results = append(results, &orchestrator.SandboxVolumeMount{
+			Id:   actualVolume.ID.String(),
+			Path: v.Path,
+			Type: actualVolume.VolumeType,
+		})
+	}
```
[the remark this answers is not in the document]
djeebus: Mounting the same volume in two different places is perfectly acceptable.

packages/api/internal/handlers/volume_delete.go
```diff
@@ -0,0 +1,66 @@
...
+		a.sendAPIStoreError(c, http.StatusInternalServerError, fmt.Sprintf("failed to delete data for volume %q", volume.ID.String()))
+
+		return
+	}
+
+	if err := tx.Commit(ctx); err != nil {
+		telemetry.ReportCriticalError(ctx, "error when committing transaction", err)
+		a.sendAPIStoreError(c, http.StatusInternalServerError, "Error committing transaction")
+
+		return
```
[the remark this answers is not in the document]
djeebus: Chicken or egg? Happy for a human to tell me to swap this.

packages/api/internal/handlers/volume_delete.go
```diff
@@ -0,0 +1,66 @@
...
+	}
```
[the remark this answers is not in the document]
djeebus: Still don't care unless a human says otherwise.

packages/api/internal/handlers/volume_delete.go
```diff
@@ -0,0 +1,66 @@
...
+	c.Status(http.StatusNoContent)
+}
```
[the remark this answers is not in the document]
djeebus: This is true, but we're going to ignore it for v0.

packages/api/internal/clusters/cluster.go
```diff
@@ -234,3 +265,44 @@ func (c *Cluster) GetOrchestrators() []*Instance {
 func (c *Cluster) GetResources() ClusterResource {
 	return c.resources
 }
+
+var ErrNoOrchestratorFound = errors.New("no orchestrator found")
+
+func (c *Cluster) DeleteVolume(ctx context.Context, volume queries.Volume) error {
+	instance, ok := c.getRandomInstance(func(instance InstanceInfo, _ machineinfo.MachineInfo) bool {
+		return instance.IsOrchestrator
+	})
```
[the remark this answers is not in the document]
djeebus: On volume create, if orchestrator fails to create the volume, then we roll back and the volume disappears from the db.

On volume delete, if orchestrator fails to delete the volume, then we have orphan files and a log that tells us to go clean them up. This seems fine.

Prove me wrong.

packages/api/internal/clusters/cluster.go
```diff
@@ -190,34 +208,53 @@ func (c *Cluster) GetByServiceInstanceID(serviceInstanceID string) (*Instance, b
 	return nil, false
 }
 
-func (c *Cluster) GetAvailableTemplateBuilder(ctx context.Context, expectedInfo machineinfo.MachineInfo) (*Instance, error) {
-	_, span := tracer.Start(ctx, "template-builder-get-available-instance")
-	span.SetAttributes(telemetry.WithClusterID(c.ID))
-	defer span.End()
-
+func (c *Cluster) getRandomHealthyInstance(isMatch func(InstanceInfo, machineinfo.MachineInfo) bool) (*Instance, bool) {
```
mishushakov: is this a safe change?

packages/api/internal/handlers/sandbox_create.go
```diff
@@ -218,6 +233,140 @@ func (a *APIStore) PostSandboxes(c *gin.Context) {
 	c.JSON(http.StatusCreated, &sbx)
 }
 
+func dedupeVolumeNames(items []api.SandboxVolumeMount) []string {
```
mishushakov, line 238: this probably something you could move to utils and handle with generics nicely

packages/api/internal/handlers/sandbox_create.go
```diff
@@ -218,6 +233,140 @@ func (a *APIStore) PostSandboxes(c *gin.Context) {
...
+
+func (im InvalidVolumeMountsError) Error() string {
+	var errs []string
+
+	for _, mount := range im.InvalidMounts {
+		errs = append(errs, fmt.Sprintf("\t- volume mount #%d: %s", mount.Index, mount.Reason))
+	}
+
+	return fmt.Sprintf("invalid mounts:\n%s", strings.Join(errs, "\n"))
+}
```
mishushakov, line 275: not sure if this logic belongs here

packages/api/internal/handlers/sandbox_create.go
```diff
@@ -218,6 +233,140 @@ func (a *APIStore) PostSandboxes(c *gin.Context) {
...
+			Name: actualVolume.Name,
+		})
+	}
+
+	if len(invalidVolumeMounts) > 0 {
+		return nil, InvalidVolumeMountsError{InvalidMounts: invalidVolumeMounts}
+	}
+
+	return results, nil
+}
```
mishushakov: probably move this function elsewhere for clarity

tests/integration/internal/tests/api/sandboxes/sandbox_list_test.go
```diff
@@ -119,7 +119,7 @@ func TestSandboxListRunning_NoMetadata(t *testing.T) {
 		State: &[]api.SandboxState{api.Running},
 	}, setup.WithAPIKey())
 	require.NoError(t, err)
-	require.Equal(t, http.StatusOK, listResponse.StatusCode())
+	require.Equal(t, http.StatusOK, listResponse.StatusCode(), string(listResponse.Body))
```
mishushakov, line 122: what's this for? you want to check that response body is not empty?

packages/orchestrator/orchestrator.proto
```diff
@@ -122,3 +131,24 @@ service SandboxService {
...
+  string volume_id = 1;
+  string volume_type = 2;
+  string team_id = 3;
+}
+
+message VolumeCreateResponse {}
+
+message VolumeDeleteRequest {
+  string volume_id = 1;
+  string volume_type = 2;
```
mishushakov, line 145: any reason we need volume_type for delete request?

packages/orchestrator/orchestrator.proto
```diff
@@ -46,6 +46,15 @@ message SandboxConfig {
...
 
   optional SandboxNetworkConfig network = 22;
+
+  repeated SandboxVolumeMount volumeMounts = 23;
+}
+
+message SandboxVolumeMount {
+  string id = 1;
+  string path = 2;
+  string type = 3;
```
mishushakov, line 56: do we need the type here?

packages/orchestrator/internal/volumes/service.go
```diff
@@ -0,0 +1,85 @@
...
+
+	if !tryParseUUID(volumeID) {
+		return "", status.Newf(codes.InvalidArgument, "invalid volume ID %q", volumeID)
+	}
+
+	volumePath := filepath.Join(volPath, teamID, volumeID)
+	rel, err := filepath.Rel(volPath, volumePath)
+	if err != nil || strings.HasPrefix(rel, ".."+string(filepath.Separator)) || rel == ".." {
+		return "", status.Newf(codes.PermissionDenied, "volume path %s is not under %s", volumePath, volPath)
+	}
```
mishushakov: can we make this bit prettier, what does it check?

packages/orchestrator/internal/cfg/model.go
```diff
@@ -92,6 +94,18 @@ func Parse() (Config, error) {
...
+		for name, path := range config.PersistentVolumeMounts {
+			path = filepath.Clean(path)
+
+			if err := os.MkdirAll(path, 0o700); err != nil {
+				return config, fmt.Errorf("failed to mkdir on persistent volume mount %q (%q): %w", name, path, err)
+			}
+
+			config.PersistentVolumeMounts[name] = path // store the cleaned path
+		}
+	}
```
mishushakov, line 111: what does this do? isn't it duplicate of what we're doing somewhere else already?

packages/api/internal/handlers/volume_util.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/gin-gonic/gin"
+	"github.com/google/uuid"
+
+	"github.com/e2b-dev/infra/packages/api/internal/db/types"
+	"github.com/e2b-dev/infra/packages/db/pkg/dberrors"
+	"github.com/e2b-dev/infra/packages/db/queries"
+	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
+)
+
+func (a *APIStore) getVolume(c *gin.Context, volumeID string) (queries.Volume, *types.Team, bool) {
```
mishushakov, line 20: should probably be

queries.Volume, *types.Team, err

packages/api/internal/handlers/volume_delete.go
```diff
@@ -0,0 +1,51 @@
...
+		a.sendAPIStoreError(c, http.StatusInternalServerError, "Error recording volume deletion")
+
+		return
+	}
+
+	go func(ctx context.Context) {
+		// if this fails, we can clean it up later
+		if err := cluster.DeleteVolume(ctx, volume); err != nil {
+			telemetry.ReportCriticalError(ctx, fmt.Sprintf("failed to delete data for volume %q", volume.ID.String()), err)
+		}
```
mishushakov: what if it was created but was never mounted on the node?

packages/api/internal/handlers/volume_create.go
```diff
@@ -0,0 +1,152 @@
...
+		a.sendAPIStoreError(c, http.StatusBadRequest, "Invalid volume name")
+		telemetry.ReportError(ctx, "invalid volume name", nil)
+
+		return
+	}
+
+	ctx = feature_flags.AddToContext(ctx, feature_flags.VolumeContext(body.Name))
+
+	volumeType := a.getVolumeType(ctx)
+	if volumeType == "" {
```
mishushakov, line 68: this should be a const, maybe BetaVolumeType or V1VolumeType

packages/api/internal/handlers/volume_create.go
```diff
@@ -0,0 +1,152 @@
...
+		Id:   volume.ID.String(),
+		Name: volume.Name,
+	}
+
+	c.JSON(http.StatusCreated, result)
+}
+
+func (a *APIStore) getVolumeType(ctx context.Context) string {
+	volumeType := a.featureFlags.StringFlag(ctx, feature_flags.DefaultPersistentVolumeType)
+	if volumeType == "" {
```
mishushakov, line 138: same as above

packages/api/internal/handlers/volume_create.go
```diff
@@ -0,0 +1,152 @@
...
+		volumeType = a.config.DefaultPersistentVolumeType
+	}
+
+	return volumeType
+}
+
+var validVolumeNameRegex = regexp.MustCompile(`^[a-zA-Z0-9_-]+$`)
```
mishushakov, line 145: you could add this pattern to openapi for validation?

packages/api/internal/handlers/sandbox_create.go
```diff
@@ -218,6 +233,140 @@ func (a *APIStore) PostSandboxes(c *gin.Context) {
...
+	var errs []string
+
+	for _, mount := range im.InvalidMounts {
+		errs = append(errs, fmt.Sprintf("\t- volume mount #%d: %s", mount.Index, mount.Reason))
+	}
+
+	return fmt.Sprintf("invalid mounts:\n%s", strings.Join(errs, "\n"))
+}
+
+func createOrchestratorVolumeMounts(
```
mishushakov: this doesn't actually create anything, does it?

packages/api/internal/handlers/sandbox_create.go
```diff
@@ -218,6 +233,140 @@ func (a *APIStore) PostSandboxes(c *gin.Context) {
...
+	}
+
+	if filepath.Clean(path) != path {
+		return "path must not contain any '.' or '..' components", false
+	}
+
+	return "", true
+}
+
+func getDBVolumesMap(ctx context.Context, sqlcDB *sqlcdb.Client, teamID uuid.UUID, volumeMounts []api.SandboxVolumeMount) (map[string]queries.Volume, error) {
```
mishushakov, line 355: i feel like this should live somewhere under db package not here

sitole: imho, not needed; we have a lot of helper functions in the API, as it's the only place where they're actually used.

tests/integration/internal/tests/api/sandboxes/sandbox_list_test.go
sitole: if status is not OK it will print error that is body

packages/api/internal/clusters/cluster.go
mishushakov: i'm referring to the overall change, including this:

```go
if machineInfo := instance.GetMachineInfo(); expectedInfo.CPUModel != "" && !expectedInfo.IsCompatibleWith(machineInfo) {
```

```go
if isMatch != nil && !isMatch(instance.GetInfo(), instance.GetMachineInfo()) {
```

and

```go
if info := instance.GetInfo(); info.Status != infogrpc.ServiceInfoStatus_Healthy || !info.IsBuilder {
```

```go
if instance.status != infogrpc.ServiceInfoStatus_Healthy {
```

packages/api/internal/handlers/volume_delete.go
```diff
@@ -0,0 +1,51 @@
...
+func (a *APIStore) DeleteVolumesVolumeID(c *gin.Context, volumeID api.VolumeID) {
+	ctx := c.Request.Context()
+
+	volume, team, ok := a.getVolume(c, volumeID)
+	if !ok {
+		return
+	}
+
+	clusterID := utils.WithClusterFallback(team.ClusterID)
+	cluster, ok := a.clusters.GetClusterById(clusterID)
```
sitole: todo

packages/db/migrations/20260204185039_volumes.sql
```diff
@@ -0,0 +1,38 @@
...
+CREATE POLICY "Allow selection for users that are in the team"
+    ON "public"."volumes"
+    AS PERMISSIVE
+    FOR SELECT
+    TO authenticated
+    USING ((auth.uid() IN ( SELECT users_teams.user_id
+                            FROM users_teams
+                            WHERE (users_teams.team_id = volumes.team_id))));
+
+-- +goose StatementEnd
```
sitole: Lets remove this one, we agreed that we are not using it now.

tests/integration/internal/tests/api/volumes/crud_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/e2b-dev/infra/tests/integration/internal/api"
+	"github.com/e2b-dev/infra/tests/integration/internal/setup"
+)
+
+func TestVolumeRoundTrip(t *testing.T) {
```
sitole, line 15: If it's possible, can we use test containers (probably with an orch call mockup) and move them to the API?

packages/api/internal/clusters/cluster.go
djeebus: Turns out cluster doesn't return orchestrators, it returns template managers. definitely doesn't work, I'll revert it.

packages/api/internal/handlers/sandbox_create.go
djeebus: It's only ever used when creating sandboxes with volumes, where else would it go? I'm open to suggestions!

packages/orchestrator/orchestrator.proto
djeebus: It helps to generate the path that needs to be deleted.

1. Orchestrator has a map of volume types that it supports, and the host mount that they correspond with. For example: "cheap=/mnt/drobo; fast=/mnt/filestore"
2. The volume root is looked up by passing the volume type passed in the request to that map, and getting the root
3. The volume is placed in `<volume type root>/<team id>/<volume id>`

packages/orchestrator/orchestrator.proto
djeebus: Same as above; it's necessary to figure out where the volume root is on the host.

packages/orchestrator/internal/volumes/service.go
djeebus: It was to prevent path traversal attacks, but now that we're not using volume names in paths, this isn't a concern; I'll remove it.

packages/orchestrator/internal/cfg/model.go
djeebus: Not a duplicate, but I realize this only makes sense when testing. We should just make sure we have access to the path, we shouldn't be creating it. I'll change it.

packages/api/internal/handlers/volume_util.go
djeebus: Why?

packages/api/internal/handlers/volume_delete.go
djeebus: That's a very real concern. Deploying new volume types will have to be done after mounting, and removing old volume types will have to be done after migrating everyone off of the old one.

packages/api/internal/handlers/volume_create.go
djeebus: All volumes have to have a volume type, empty volume types cannot be created.

packages/api/internal/handlers/volume_create.go
djeebus: All volumes have to have a volume type, empty volume types cannot be created.

packages/api/internal/handlers/volume_create.go
djeebus: That's a good idea! I'd like to validate it here as well though, just in case a bug is introduced into our openapi validator.

packages/api/internal/handlers/sandbox_create.go
djeebus: It creates structs, but not database rows, file paths, etc. How about `convertAPIVolumesToOrchestratorVolumes`?

tests/integration/internal/tests/api/volumes/crud_test.go
djeebus: I wanted to test everything (authentication, file access, network, config, etc), and this does all of that (and pointed out a few bugs that I had missed, and code that I had misunderstood). Is there a reason why we want this test to do less? I also plan on adding to it when we implement the nfs proxy and test reads and writes as well.

packages/api/internal/handlers/volume_util.go
```diff
@@ -0,0 +1,98 @@
...
+		if node.Status() != api.NodeStatusReady {
+			continue
+		}
+
+		c, ctx := node.GetClient(ctx)
+
+		// todo: retry on transient errors such as "failed to connect" or "read timeout", etc.
+
+		return fn(ctx, c)
+	}
```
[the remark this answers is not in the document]
djeebus: @sitole now I see why you did the shuffle :grin:

packages/api/internal/handlers/sandbox_create.go
djeebus: There's value in that, there's a lot of places we could use a common Set implementation, but I'd like to get this PR merged before any more potentially controversial refactors.

packages/api/internal/handlers/sandbox_create.go
djeebus: This is only used by CreateSandbox, where else would we move it?

packages/api/internal/handlers/sandbox_create.go
mishushakov: i think it's better yes

packages/api/internal/handlers/volume_create.go
mishushakov: sorry i misread that part, can we make it a const VolumeTypeNull ?

packages/api/internal/handlers/volume_util.go
mishushakov: i think for utility method i'd rather have both the values and an error and then when you call the method you'd still get the error value (for debugging at least) and log it/return an API error

packages/api/internal/handlers/sandbox_create.go
mishushakov: idk maybe sandbox_volume_utils - my concern is primarily readability of the code

packages/api/internal/handlers/sandbox_create.go
mishushakov: i was i think more confused that we mix types and functions and vars everywhere in the file - maybe we should have sandbox_volume_utils

packages/api/internal/handlers/sandbox_create.go
mishushakov: you could move to sandbox_volume_utils (see my other comment)

mishushakov:
can we allow empty array and make it no-op? or do you want to be strict?

packages/api/internal/handlers/volume_util.go
djeebus: The error gets logged and returned as an API error. To refactor this would require returning, not just the error, but either appropriate http status codes (as there are many) or custom errors (so the calling code can decide which http status code to return). An extra ~50 lines of code so that we can avoid returning a bool seems unnecessary, and the end result is exactly the same: the error is logged and returned via http.

packages/api/internal/handlers/sandbox_create.go
djeebus: I don't think there's value in that, if the purpose is to hide the code. Having that code in the only file that calls the code makes its scope very clear. Refactoring is safe and obvious. When you pull the code out into another file, it becomes less obvious what refactoring it would affect, and hides its scope.

I think adding a shared `Set` struct could consolidate a lot of code across our code base, but that's out of scope for this PR.

packages/api/internal/handlers/sandbox_create.go
djeebus: This logic isn't used for anything outside of this file, and is tightly coupled with the "Create Sandbox" API call. Moving it out would obscure its scope and make refactoring more difficult.

packages/api/internal/handlers/sandbox_create.go
djeebus: Keeping the code within this file makes it easier to understand the scope, which is really important. "utils" files tend to end up being a dumpster of unrelated code that doesn't get the attention it deserves, but b/c they're shared, people get nervous about improving it (for fear of breaking something that some unknown consumer needed), so it's best to avoid those if you don't need them.

As an example, let's say we refactored `dedupeVolumeNames` to use generics, and moved it into a shared folder that made it look as though it was used by lots of consumers. Then we find out that the original use case had a bug - we have to omit empty strings from the output. Currently we can just hack it in the code that's only used by "create sandbox", and we're fine. If we centralize it, we need to make sure that everyone is okay with omitting default values from the output set. Alternately, we take the easy way out and loop through the set it "create sandbox" before using it and remove the empty values. Either way, we've turned a 2 line change into an hour or two long chore through our code base, with potential side effects and negative consequences.

packages/api/internal/handlers/volume_create.go
djeebus: It's not a nil volume type though, it's unset, which is the same as `""`. There's generally no reason to set a default value to a const; for example, you wouldn't do `for i := StartIndexConst; i < len(slice); i++` right? `i := 0` is perfectly fine, and more readable than the const, as it removes a layer of abstraction.

packages/api/internal/handlers/sandbox_create.go
```diff
@@ -182,14 +174,36 @@ func (a *APIStore) PostSandboxes(c *gin.Context) {
...
 			return
 		}
 	}
 
+	sbxVolumeMounts, err := convertAPIVolumesToOrchestratorVolumes(
+		ctx, a.sqlcDB, a.featureFlags, teamInfo.ID, apiVolumeMounts,
+	)
+	if err != nil {
+		if errors.Is(err, ErrVolumeMountsDisabled) {
+			a.sendAPIStoreError(c, http.StatusForbidden, "Volume mounts are not enabled.")
```
sitole: Isn this just 400?

packages/api/internal/handlers/sandbox_create.go
```diff
@@ -182,14 +174,36 @@ func (a *APIStore) PostSandboxes(c *gin.Context) {
...
+		}
+
+		var vne InvalidVolumeMountsError
+		if errors.As(err, &vne) {
+			a.sendAPIStoreError(c, http.StatusBadRequest, vne.Error())
+
+			return
+		}
+
+		a.sendAPIStoreError(c, http.StatusInternalServerError, fmt.Errorf("failed to convert volume mounts: %w", err).Error())
```
sitole: Can we manually fill error message / let `convertAPIVolumesToOrchestratorVolumes` to return api error? We dont want to forward `err` to client messages directly.

packages/api/internal/handlers/sandbox_create.go
sitole: Just move it to `volume_util.go` wdyt?

packages/api/internal/handlers/sandbox_create.go
sitole: Lets move it to volume util file. It will be easier to find

sitole: Okay i see why its there, we already put there lot of functions. I think we should stop with that or this file will grow really fast.

sitole:
LGTM, just nits. Feel free to merge.

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -234,6 +236,20 @@ func (s *Server) Create(ctx context.Context, req *orchestrator.SandboxCreateRequ
...
+	for _, v := range volumeMounts {
+		results = append(results, sandbox.VolumeMountConfig{
+			ID:   v.GetId(),
+			Path: v.GetPath(),
+			Type: v.GetType(),
+		})
+	}
+
+	return results
+}
```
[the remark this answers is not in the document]
djeebus: This is by design; names are not needed by the orchestrator.

packages/api/internal/handlers/sandbox_create.go
```diff
@@ -219,6 +234,140 @@ func (a *APIStore) PostSandboxes(c *gin.Context) {
...
+	if !filepath.IsAbs(path) {
+		return "path must be absolute", false
+	}
+
+	if filepath.Clean(path) != path {
+		return "path must not contain any '.' or '..' components", false
+	}
+
+	return "", true
+}
```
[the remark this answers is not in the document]
djeebus: These are customer sandboxes, if they want to do this, there's probably a good reason for it.

packages/api/internal/handlers/volume_create.go
```diff
@@ -0,0 +1,161 @@
...
+		telemetry.ReportCriticalError(ctx, "error when getting team", apiErr.Err)
+
+		return
+	}
+
+	telemetry.SetAttributes(ctx,
+		telemetry.WithTeamID(team.ID.String()),
+	)
+
+	if !a.featureFlags.BoolFlag(ctx, feature_flags.PersistentVolumesFlag) {
```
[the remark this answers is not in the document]
djeebus: Done via middleware

PR #1806, opened 2026-01-29
feat: team-local template names

dobrac:
- adds also a response with names to the template update v2 endpoint

---

> [!NOTE]
> **Medium Risk**
> Touches template update response semantics and alias/namespace data migration/lookup, which could affect name resolution and backward compatibility if the namespace population/queries are wrong.
> 
> **Overview**
> **Adds team-local (namespaced) template naming and surfaces updated names via the v2 template PATCH.** The `PATCH /v2/templates/{templateID}` endpoint now returns a JSON body (`TemplateUpdateResponse`) containing the template’s resolved `names` after update, and client code is regenerated to parse the 200 response.
> 
> Alias handling is moved toward namespace-aware uniqueness: build registration checks alias existence via `CheckAliasExistsInNamespace`, and a new DB migration backfills `env_aliases.namespace` (setting team slug for private templates and duplicating public aliases with a namespaced row while preserving NULL-namespace for backward compatibility). Query aggregation for template names/aliases is adjusted to return empty arrays instead of NULL when no aliases exist, and related telemetry/error handling is refactored.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4d78b7bf17c61186d9197b4b43317f8cbea83fee. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api/internal/handlers/template_update.go
+++ b/packages/api/internal/handlers/template_update.go
@@ -24,8 +24,15 @@ import (
 func (a *APIStore) PatchTemplatesTemplateID(c *gin.Context, aliasOrTemplateID api.TemplateID) {
 	ctx := c.Request.Context()
 
-	_, _, ok := a.updateTemplate(ctx, c, aliasOrTemplateID, true)
-	if !ok {
+	_, _, apiErr := a.updateTemplate(ctx, c, aliasOrTemplateID, true)
+	if apiErr != nil {
+		if apiErr.Code < http.StatusInternalServerError {
+			telemetry.ReportError(ctx, apiErr.ClientMsg, apiErr.Err, telemetry.WithTemplateID(aliasOrTemplateID))
+		} else {
+			telemetry.ReportCriticalError(ctx, apiErr.ClientMsg, apiErr.Err, telemetry.WithTemplateID(aliasOrTemplateID))
+		}
+		a.sendAPIStoreError(c, apiErr.Code, apiErr.ClientMsg)
+
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


packages/api/internal/handlers/template_update.go
```diff
@@ -20,31 +20,47 @@ import (
...
+
+type templateUpdateResult struct {
+	Names []string
 }
 
-// updateTemplate contains the shared logic for updating a template
-func (a *APIStore) updateTemplate(c *gin.Context, aliasOrTemplateID api.TemplateID, createBackwardCompatAlias bool) {
+// updateTemplate contains the shared logic for updating a template.
+// Returns nil if an error occurred (error response already sent), or the result on success.
+func (a *APIStore) updateTemplate(c *gin.Context, aliasOrTemplateID api.TemplateID, createBackwardCompatAlias bool) *templateUpdateResult {
```
jakubno: can't we return api.APIError as on other places and return the response from handler? This behavior seems little bit weird to me

packages/api/internal/handlers/template_update.go
```diff
@@ -66,41 +82,45 @@ func (a *APIStore) updateTemplate(c *gin.Context, aliasOrTemplateID api.Template
...
-		})
-		if err != nil {
-			if dberrors.IsNotFoundError(err) {
-				a.sendAPIStoreError(c, http.StatusNotFound, fmt.Sprintf("Template '%s' not found or you don't have access to it", aliasOrTemplateID))
-				telemetry.ReportError(ctx, "template not found", err, telemetry.WithTemplateID(aliasInfo.TemplateID))
-
-				return
-			}
+	// Update template only if public field is provided
+	if body.Public == nil {
```
jakubno, line 111: This wil break when we add some other field, but it's probably fine

dobrac: yes, but it's ok for now I think

packages/api/internal/handlers/template_update.go
dobrac: yes, will redo

dobrac: done

PR #1819, opened 2026-01-30
Fix ENV value escaping in template builds

mishushakov:
we pass the input as a string to printf and then replace env vars in the string with envsubst

---

> [!NOTE]
> **Medium Risk**
> Touches shell command construction for template builds, so escaping mistakes could break env interpolation or cause build failures; changes are scoped and backed by added integration tests.
> 
> **Overview**
> Fixes template-build `ENV` handling by safely escaping values before piping them through `printf`, preventing malformed parsing and blocking command substitution (backticks/`$()`), while still allowing `$VAR` interpolation. Expands integration coverage with new `ENV` test cases for quotes, backslashes, PEM-style strings, multiline values, and injection attempts, and updates the expected base image package set to include `iptables`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1887c12370e2a68368653436e21a0d568bc71e8f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/integration/internal/tests/api/templates/build_template_test.go
+++ b/tests/integration/internal/tests/api/templates/build_template_test.go
@@ -216,6 +216,139 @@ func TestTemplateBuildENV(t *testing.T) {
 				}),
 			},
 		},
+		{
+			name:         "ENV with PEM-style dashes",
+			templateName: "test-ubuntu-env-pem-dashes",
+			buildConfig: api.TemplateBuildStartV2{
+				Force:     utils.ToPtr(ForceBaseBuild),
+				FromImage: utils.ToPtr("ubuntu:22.04"),
+				Steps: utils.ToPtr([]api.TemplateStep{
+					{
+						Type:  "ENV",
+						Force: utils.ToPtr(true),
+						Args:  utils.ToPtr([]string{"PEM_HEADER", "-----BEGIN DATA-----"}),
+					},
+					{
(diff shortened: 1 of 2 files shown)


tests/integration/internal/tests/api/templates/build_template_test.go
```diff
@@ -886,7 +1019,9 @@ func TestTemplateBuildInstalledPackagesAvailable(t *testing.T) {
 		"curl",
 		"ca-certificates",
 		"fuse3",
+		"iptables",
```
jakubno: how are iptables relevant to this?

packages/orchestrator/internal/template/build/phases/base/provision.sh
```diff
@@ -16,7 +16,7 @@ is_package_installed() {
 }
 
 # Install required packages if not already installed
-PACKAGES="systemd systemd-sysv openssh-server sudo chrony linuxptp socat curl ca-certificates fuse3 iptables git"
+PACKAGES="systemd systemd-sysv openssh-server sudo chrony linuxptp socat curl ca-certificates fuse3 iptables git gettext-base"
```
dobrac: I would be for not having the dependency here as otherwise you'd need to do a fallback solution if the dependency is not installed

PR #1822, opened 2026-01-31
feat(envd): validate access token via MMDS hash during resume

dobrac:
Add MMDS-based access token validation for /init requests. This allows the orchestrator to securely update (or reset) the access token during sandbox resume by passing the token hash through MMDS metadata.

Changes:
- Add AccessTokenHash field to MMDS metadata
- Add HashAccessToken function using SHA-512
- Validate /init requests against MMDS hash OR existing token
- Pass access token through to FC Resume for MMDS setup

---

> [!NOTE]
> **Medium Risk**
> Touches sandbox authentication/token lifecycle and Firecracker MMDS wiring; incorrect validation or MMDS availability could block init/resume flows or unintentionally allow/deny token resets.
> 
> **Overview**
> Envd now **excludes `POST /init` from the standard `X-Access-Token` middleware** and instead validates `/init` requests by requiring the provided token to match either the currently configured token or a hash published in Firecracker MMDS (with first-time setup allowed when no token/MMDS hash exists). This also enables an orchestrator-controlled *explicit token reset* by publishing `hash("")` in MMDS and sending a nil token.
> 
> The orchestrator’s Firecracker resume path now writes `accessTokenHash` into MMDS metadata (computed via a new shared `keys.HashAccessToken` SHA-512 helper), and envd gains an MMDS client abstraction plus host support for fetching just the access-token hash. Tests were expanded/updated to cover the new unauthorized cases and the changed `/init` failure status (now `401` instead of `409` for token-change attempts).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3f8adf4bc1983a7c37989093ce45149fb94ded0e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/envd/internal/api/init.go
+++ b/packages/envd/internal/api/init.go
@@ -112,9 +112,7 @@ func (a *API) PostInit(w http.ResponseWriter, r *http.Request) {
 			err = a.SetData(ctx, logger, initRequest)
 			if err != nil {
 				switch {
-				case errors.Is(err, ErrAccessTokenMismatch):
-					w.WriteHeader(http.StatusUnauthorized)
-				case errors.Is(err, ErrAccessTokenResetNotAuthorized):
+				case errors.Is(err, ErrAccessTokenMismatch), errors.Is(err, ErrAccessTokenResetNotAuthorized):
 					w.WriteHeader(http.StatusUnauthorized)
 				default:
 					logger.Error().Msgf("Failed to set data: %v", err)


packages/envd/internal/api/init.go
```diff
@@ -22,16 +24,68 @@ import (
...
+	return hex.EncodeToString(h[:])
+}
+
+// validateInitAccessToken validates the access token for /init requests.
+// Token is valid if it matches the existing token OR the MMDS hash.
+// If neither exists, first-time setup is allowed.
+func (a *API) validateInitAccessToken(ctx context.Context, logger zerolog.Logger, requestToken *string) error {
+	// Fast path: token matches existing
+	if a.accessToken != nil && requestToken != nil && *requestToken == *a.accessToken {
+		return nil
```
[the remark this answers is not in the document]
dobrac: I think thats ok. I don't want (need to) to use the X-Access-Token from the header in /init

packages/envd/internal/host/mmds.go
```diff
@@ -108,6 +111,25 @@ func getMMDSOpts(ctx context.Context, client *http.Client, token string) (*MMDSO
 	return &opts, nil
 }
 
+// GetAccessTokenHashFromMMDS reads the access token hash from MMDS.
+// This is used to validate that /init requests come from the orchestrator.
+func GetAccessTokenHashFromMMDS(ctx context.Context) (string, error) {
+	client := &http.Client{Timeout: mmdsAccessTokenTimeout}
```
ValentaTomas: Can we just have a global client that does not use keepalive, so we don't create in inside of the method?

packages/envd/internal/api/init.go
ValentaTomas, line 128: Right now, is the MMDS polling useful for anything? Can we just load the values from there on init, without pooling it in the background?

dobrac: I don't think so really. Let's rework it in another PR

packages/envd/internal/host/mmds.go
dobrac: moved

packages/envd/internal/api/auth.go
```diff
@@ -19,10 +19,12 @@ const (
 )
 
 // paths that are always allowed without general authentication
+// POST/init is secured via MMDS hash validation instead
 var allowedPaths = []string{
```
sitole:
```suggestion
var authExcludedPaths = []string{
```

packages/envd/internal/api/init.go
```diff
@@ -17,21 +17,87 @@ import (
...
 	maxTimeInFuture = 5 * time.Second
 )
 
+// validateInitAccessToken validates the access token for /init requests.
+// Token is valid if it matches the existing token OR the MMDS hash.
+// If neither exists, first-time setup is allowed.
+//
+// Returns:
+//   - ErrAccessTokenMismatch (401): invalid token provided
+//   - ErrAccessTokenConflict (409): trying to reset token without authorization
```
sitole: I think calling it "conflict" is incorrect.

dobrac: would you keep it all under just 401?

packages/envd/internal/api/init.go
```diff
@@ -17,21 +17,87 @@ import (
...
+		return nil // first-time setup
+	case requestToken == nil:
+		logger.Error().Msg("Access token change not authorized")
+
+		return ErrAccessTokenConflict
+	default:
+		logger.Error().Msg("Access token validation failed")
+
+		return ErrAccessTokenMismatch
+	}
```
sitole, line 57: Do we need to log when contexted error is returned

dobrac: probably not 👍

packages/envd/internal/host/mmds.go
```diff
@@ -19,12 +19,22 @@ const (
 
 	mmdsDefaultAddress  = "169.254.169.254"
 	mmdsTokenExpiration = 60 * time.Second
+
+	mmdsAccessTokenTimeout = 10 * time.Second
```
sitole:
```suggestion
	mmdsAccessTokenRequestClientTimeout = 10 * time.Second
```

`mmdsAccessTokenTimeout` fells more like expiration of token itself.

packages/orchestrator/internal/sandbox/fc/process.go
```diff
@@ -449,6 +451,12 @@ func (p *Process) Resume(
 		LogsCollectorAddress: fmt.Sprintf("http://%s/logs", slot.HyperloopIPString()),
 	}
 
+	if accessToken != nil && *accessToken != "" {
+		meta.AccessTokenHash = keys.HashAccessToken(*accessToken)
+	} else {
+		meta.AccessTokenHash = keys.HashAccessToken("")
```
sitole, line 457: Why not omit it?

dobrac: to distinguish between:
- reset (allow no token, mmds has hash of empty)
- no token provided (but mmds was not really set, e.g. old orchestrator node)

packages/envd/internal/api/init.go
sitole: One thing is how to name the error (we can name it explicitly); the second is the HTTP code we are returning. Not sure if we will find something more specific than 401.

dobrac: ok, will rename + change to 401

packages/envd/internal/api/init.go
```diff
@@ -51,11 +109,13 @@ func (a *API) PostInit(w http.ResponseWriter, r *http.Request) {
...
-			err = a.SetData(logger, initRequest)
+			err = a.SetData(ctx, logger, initRequest)
 			if err != nil {
 				switch {
-				case errors.Is(err, ErrAccessTokenAlreadySet):
-					w.WriteHeader(http.StatusConflict)
+				case errors.Is(err, ErrAccessTokenMismatch):
+					w.WriteHeader(http.StatusUnauthorized)
+				case errors.Is(err, ErrAccessTokenResetNotAuthorized):
+					w.WriteHeader(http.StatusUnauthorized)
```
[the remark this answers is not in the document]
sitole: @dobrac maybe do this?

PR #1823, opened 2026-01-31
feat(envd): memory secure token storage

dobrac:
> [!NOTE]
> **Medium Risk**
> Touches authentication/token validation and request parsing; mistakes could cause auth regressions or token-reset edge cases, though changes are scoped and covered by unit/integration tests.
> 
> **Overview**
> Introduces **memory-secure access token handling** in `envd` by replacing plain `*string` tokens with a new `SecureToken` backed by `memguard` (secure allocation, constant-time compare, explicit `Destroy`, and zeroing of input buffers).
> 
> Updates `/init` and auth/signing flows to avoid keeping token material in normal memory: the `/init` handler now reads the raw body to allow wiping it after JSON parsing, transfers tokens via `TakeFrom` and destroys unused tokens, and MMDS hash validation now hashes token bytes (`HashAccessTokenBytes`) with wiped copies. OpenAPI-generated clients are adjusted to use `SecureToken` types (aliased to `string` in orchestrator/integration clients), and tests are updated/added to cover secure token behavior and prevent deadlocks.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3c4ec8a91eda0a9486a148ac930405d12d073cce. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/envd/internal/api/secure_token.go
+++ b/packages/envd/internal/api/secure_token.go
@@ -99,20 +99,19 @@ func (s *SecureToken) TakeFrom(src *SecureToken) {
 		return
 	}
 
+	// Extract buffer from source
 	src.mu.Lock()
-	defer src.mu.Unlock()
+	buffer := src.buffer
+	src.buffer = nil
+	src.mu.Unlock()
 
+	// Install buffer in destination
 	s.mu.Lock()
-	defer s.mu.Unlock()
-
-	// Destroy current buffer
 	if s.buffer != nil {
 		s.buffer.Destroy()
 	}
-
-	// Transfer ownership
-	s.buffer = src.buffer
-	src.buffer = nil
+	s.buffer = buffer
+	s.mu.Unlock()
 }
 
 // Equals checks if token matches using constant-time comparison.
@@ -143,12 +142,12 @@ func (s *SecureToken) EqualsSecure(other *SecureToken) bool {
 		return s.IsSet()
 	}
 
-	other.mu.RLock()
-	defer other.mu.RUnlock()
-
-	if other.buffer == nil || !other.buffer.IsAlive() {
+	// Get a copy of other's bytes (avoids holding two locks simultaneously)
+	otherBytes, err := other.Bytes()
+	if err != nil {
 		return false
 	}
+	defer memguard.WipeBytes(otherBytes)
 
 	s.mu.RLock()
 	defer s.mu.RUnlock()
@@ -157,7 +156,7 @@ func (s *SecureToken) EqualsSecure(other *SecureToken) bool {
 		return false
 	}
 
-	return s.buffer.EqualTo(other.buffer.Bytes())
+	return s.buffer.EqualTo(otherBytes)
 }
 
 // IsSet returns true if a token is stored.
--- a/packages/envd/internal/api/secure_token_test.go
+++ b/packages/envd/internal/api/secure_token_test.go
@@ -365,6 +365,40 @@ func TestSecureTokenEqualsSecure(t *testing.T) {
 		assert.True(t, st2.EqualsSecure(st1))
 	})
 
+	t.Run("concurrent TakeFrom and EqualsSecure do not deadlock", func(t *testing.T) {
+		t.Parallel()
+		// This test verifies the fix for the lock ordering deadlock bug.
+
+		const iterations = 100
+
+		for range iterations {
+			a := &SecureToken{}
+			err := a.Set([]byte("token-a"))
+			require.NoError(t, err)
+
+			b := &SecureToken{}
+			err = b.Set([]byte("token-b"))
+			require.NoError(t, err)
+
+			var wg sync.WaitGroup
+			wg.Add(2)
+
+			// Goroutine 1: a.TakeFrom(b)
+			go func() {
+				defer wg.Done()
+				a.TakeFrom(b)
+			}()
+
+			// Goroutine 2: b.EqualsSecure(a)
+			go func() {
+				defer wg.Done()
+				b.EqualsSecure(a)
+			}()
+
+			wg.Wait()
+		}
+	})
+
 	t.Run("returns false for different tokens", func(t *testing.T) {
 		t.Parallel()
 		st1 := &SecureToken{}


packages/envd/internal/api/secure_token_test.go
```diff
@@ -0,0 +1,140 @@
...
+
+	var wg sync.WaitGroup
+	const numGoroutines = 100
+
+	// Concurrent reads
+	for range numGoroutines {
+		wg.Go(func() {
+			st.IsSet()
+			st.Equals("initial-token")
+		})
```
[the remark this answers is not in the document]
dobrac: outdated suggestion

packages/envd/internal/api/auth.go
```diff
@@ -50,17 +50,24 @@ func (a *API) WithAuthorization(handler http.Handler) http.Handler {
...
+			tokenBytes[i] = 0
+		}
+	}()
 
 	var signature string
 	hasher := keys.NewSHA256Hashing()
 
 	if signatureExpiration == nil {
-		signature = fmt.Sprintf("%s:%s:%s:%s", path, operation, username, *a.accessToken)
+		signature = fmt.Sprintf("%s:%s:%s:%s", path, operation, username, string(tokenBytes))
```
djeebus: Not your line originally, but might be more readable to do `strings.Join([]string{path, operation, username, string(tokenBytes)}, ":")`

packages/envd/internal/api/secure_token.go
```diff
@@ -0,0 +1,93 @@
...
+	defer s.mu.Unlock()
+
+	// Destroy old token first (zeros memory)
+	if s.buffer != nil {
+		s.buffer.Destroy()
+		s.buffer = nil
+	}
+
+	// Create new LockedBuffer from bytes (source is wiped automatically)
+	s.buffer = memguard.NewBufferFromBytes([]byte(token))
```
djeebus: What does "source is wiped automatically" mean? If we were to add an `if token != "" { panic("this did not work" }` after this line, would it not panic?

dobrac: very good point. The []byte() is cleaned, but the token remains. I'll refactor it to be already serialized from JSON to the memguarded memory

dobrac: reworked to parsing the JSON directly to the SecureToken and wiping the buffer after. Minimizing a bit more the trace

packages/envd/internal/api/auth.go
dobrac: updated

packages/envd/internal/api/secure_token.go
```diff
@@ -0,0 +1,186 @@
...
+// UnmarshalJSON implements json.Unmarshaler to securely parse a JSON string
+// directly into memguard, wiping the input bytes after copying.
+//
+// Access tokens are hex-encoded HMAC-SHA256 hashes (64 chars of [0-9a-f]),
+// so they never contain JSON escape sequences.
+func (s *SecureToken) UnmarshalJSON(data []byte) error {
+	// JSON strings are quoted, so minimum valid is `""` (2 bytes).
+	if len(data) < 2 || data[0] != '"' || data[len(data)-1] != '"' {
+		return errors.New("invalid secure token JSON string")
+	}
```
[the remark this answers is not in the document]
dobrac: its alright.

packages/envd/internal/api/init.go
```diff
@@ -36,9 +37,11 @@ const (
 // validateInitAccessToken validates the access token for /init requests.
 // Token is valid if it matches the existing token OR the MMDS hash.
 // If neither exists, first-time setup is allowed.
-func (a *API) validateInitAccessToken(ctx context.Context, requestToken *string) error {
+func (a *API) validateInitAccessToken(ctx context.Context, requestToken *SecureToken) error {
+	requestTokenSet := requestToken != nil && requestToken.IsSet()
+
 	// Fast path: token matches existing
-	if a.accessToken != nil && requestToken != nil && *requestToken == *a.accessToken {
+	if a.accessToken.IsSet() && requestTokenSet && a.accessToken.EqualsSecure(requestToken) {
```
ValentaTomas, line 44: Btw, for the equal, it looks like stdlib might have some of the functionality already https://pkg.go.dev/crypto/subtle

ValentaTomas, line 44: Both request and access tokens are hashes inside sandbox, right?

dobrac: the EqualsSecure is not my implementation of the crypto, but rather than concurrency safe usage of the `memguard.LockedBuffer` EqualTo method

dobrac: no, they're raw tokens unfortunately as of now

packages/envd/internal/api/secure_token.go
```diff
@@ -0,0 +1,186 @@
...
+	return s.buffer.EqualTo(other.buffer.Bytes())
+}
+
+// IsSet returns true if a token is stored.
+func (s *SecureToken) IsSet() bool {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
+	return s.buffer != nil && s.buffer.IsAlive()
+}
```
djeebus, line 173: Would it clean things up to make this struct nil-pointer-receiver-safe? 

```golang
func (s *SecureToken) IsSet() bool {
	if s == nil { return false }

	// implementation details
}
```

That way we can do things like this:

```golang
// old way
if tok != nil && tok.IsSet() { ... }

// new way
if tok.IsSet() { ... }
```

Not necessary, just a thought.

packages/envd/internal/api/secure_token.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	content := data[1 : len(data)-1]
+
+	// Access tokens are hex strings - reject if contains backslash
+	if bytes.ContainsRune(content, '\\') {
+		memguard.WipeBytes(data)
+
+		return errors.New("invalid secure token: unexpected escape sequence")
+	}
```
djeebus, line 68: Why do we need to prevent this situation?

packages/envd/internal/api/init.go
ValentaTomas: Will we be able to switch? That should be possible based on the envd version, right?

dobrac: We'll be absolutely able to switch to store only the hashed token. If only sending the hash, probably as you're saying based on the envd version, should be fine.

packages/envd/internal/api/secure_token.go
dobrac: let me try that, that sounds like a good idea

packages/envd/internal/api/secure_token.go
dobrac: JSON escape sequences require processing: JSON strings can contain escape sequences like \", \\, \n, \uXXXX, etc. Normally you'd use strconv.Unquote to decode these, but that allocates new memory. We want to copy the token directly into memguard-protected memory without intermediate allocations.

-> The backslash check lets us safely skip JSON unescaping logic while ensuring we reject any input that would require it. It's a security optimization specific to the hex-only token format.

packages/envd/internal/api/secure_token.go
dobrac: updated

PR #1829, opened 2026-02-02
Feature: allow auto-resume on traffic to paused sandboxes (feature flagged)

matthewlouisbrockman:
# Proxy Auto‑Resume (Paused Sandboxes)

  ## What’s implemented
  - **Auto‑resume policy** on sandbox create (`auto_resume = any | authed | null`).
  - **Proxy‑side auto‑resume flow**:
    - Proxy routes via routing catalog.
    - If upstream fails (e.g. 502), proxy checks paused state via API gRPC.
    - If auto‑resume is disabled, proxy returns **409 Paused**.
    - If auto‑resume is allowed, proxy triggers resume and the request may succeed on retry.
  - **API gRPC service** for proxy:
    - `GetPausedInfo` returns paused state + policy (from DB snapshot + best‑effort running check).
    - `ResumeSandbox` enforces policy and resumes.
  - **Concurrent resume lock** when Redis is configured (lock is skipped if Redis is disabled).
  - **Tests**: integration tests cover policy matrix + concurrent resume behavior.

  ## How to use (API)
  ### 1) Set policy on sandbox create
  ```json
  {
    "templateID": "base",
    "autoResume": "any",
    "timeout": 600,
    "autoPause": false
  }
```
  Supported values:

  - any   — resume for any request
  - authed — resume only when proxy request contains valid auth
  - null  — never auto‑resume (always 409)

  ### 2) Proxy behavior

  - If sandbox is paused and auto‑resume is disabled:
      - Proxy returns 409 with paused error page.
  - If auto‑resume is enabled:
      - Proxy triggers resume via API gRPC and the request may succeed on retry.

  ## Auth requirements for authed

  - Request must include:
      - Authorization: Bearer <access_token> or
      - X-API-Key: <team_api_key>

  ## Deployment prerequisites

  - Client‑proxy requires API_GRPC_ADDRESS (API gRPC service).
    Example: api.service.consul:5009

  ## Key files

  - packages/client-proxy/internal/proxy/proxy.go
  - packages/client-proxy/internal/proxy/paused_checker_grpc.go
  - packages/api/internal/handlers/proxy_grpc.go
  - packages/shared/pkg/grpc/proxy/auto_resume_policy.go
  - tests/integration/internal/tests/proxies/auto_resume_test.go

  ## Tests

  make -C tests/integration test/proxies/auto_resume_test.go

  ## Notes / Behavior

  - Proxy checks routing catalog first for performance.
  - If upstream fails (502), proxy checks paused state and returns 409 when auto‑resume isn’t allowed.
  - Resume attempts use auth metadata from the incoming request.

### CODE REVIEW

File: iac/provider-gcp/nomad-cluster/scripts/start-client.sh
Comment (jakubno) [inline iac/provider-gcp/nomad-cluster/scripts/start-client.sh:1]: is this related?

File: iac/provider-gcp/nomad/jobs/client-proxy.hcl [hunk 1]
```diff
@@ -101,6 +101,9 @@ job "client-proxy" {
         REDIS_CLUSTER_URL   = "${redis_cluster_url}"
         REDIS_TLS_CA_BASE64 = "${redis_tls_ca_base64}"
 
+        API_GRPC_ADDRESS = "api-grpc.service.consul:${api_grpc_port}"
+        AUTO_RESUME_ENABLED = "${auto_resume_enabled}"
```
jakubno: this won't be needed if we have feature flag

iac/provider-gcp/nomad/main.tf
```diff
@@ -162,7 +163,10 @@ resource "nomad_job" "client_proxy" {
       redis_cluster_url   = local.redis_cluster_url
       redis_tls_ca_base64 = trimspace(data.google_secret_manager_secret_version.redis_tls_ca_base64.secret_data)
 
-      image_name = data.google_artifact_registry_docker_image.client_proxy_image.self_link
+      image_name          = data.google_artifact_registry_docker_image.client_proxy_image.self_link
+      domain_name         = var.domain_name
```
jakubno: probably not needed

packages/api/internal/handlers/proxy_grpc.go
```diff
@@ -0,0 +1,236 @@
...
+
+	timeout := sandbox.SandboxTimeoutDefault
+	if req.GetTimeoutSeconds() > 0 {
+		timeout = time.Duration(req.GetTimeoutSeconds()) * time.Second
+	}
+	if timeout > time.Duration(team.Limits.MaxLengthHours)*time.Hour {
+		return nil, status.Error(codes.InvalidArgument, "timeout exceeds team limit")
+	}
+
+	running, runErr := s.api.orchestrator.GetSandbox(ctx, team.Team.ID, sandboxID)
```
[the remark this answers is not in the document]
matthewlouisbrockman: aight, fix deploying

packages/shared/pkg/grpc/proxy/proxy.proto
```diff
@@ -0,0 +1,37 @@
...
+  // Sandbox ID to resume.
+  string sandbox_id = 1;
+  // Optional timeout override in seconds (must be within team limits).
+  int32 timeout_seconds = 2;
+}
+
+// ProxySandboxService is used by the client-proxy to check paused state
+// and resume sandboxes. Authorization/X-API-Key headers are forwarded
+// via gRPC metadata for authed resumes.
+service ProxySandboxService {
```
jakubno: I would probably call it just SandboxService or APIService if we ever decide somebody else should speak with it as well

packages/shared/pkg/grpc/proxy/proxy.proto
```diff
@@ -0,0 +1,37 @@
...
+  // Sandbox ID (short or full), as used by the client-proxy.
+  string sandbox_id = 1;
+}
+
+// Response describing whether a paused snapshot exists and its auto-resume policy.
+message SandboxPausedInfoResponse {
+  // True when a paused snapshot exists for the sandbox.
+  bool paused = 1;
+  // Auto-resume policy: "any", "authed", or "null".
+  string auto_resume_policy = 2;
```
jakubno: can it be enum?

packages/api/internal/handlers/sandbox_metadata_update.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/gin-gonic/gin"
+
+	"github.com/e2b-dev/infra/packages/api/internal/api"
+	"github.com/e2b-dev/infra/packages/api/internal/auth"
+	types "github.com/e2b-dev/infra/packages/api/internal/db/types"
+	"github.com/e2b-dev/infra/packages/api/internal/utils"
+	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
+)
+
+func (a *APIStore) PatchSandboxesSandboxIDMetadata(c *gin.Context, id api.SandboxID) {
```
jakubno: can we do this in separate PR? It has some issues because of the data synchronization

packages/api/internal/orchestrator/update_lifecycle.go
```diff
@@ -0,0 +1,39 @@
...
+	"errors"
+	"net/http"
+
+	"github.com/google/uuid"
+
+	"github.com/e2b-dev/infra/packages/api/internal/api"
+	"github.com/e2b-dev/infra/packages/api/internal/sandbox"
+)
+
+func (o *Orchestrator) UpdateSandboxLifecycle(ctx context.Context, teamID uuid.UUID, sandboxID string, autoPause bool) *api.APIError {
```
jakubno: Is this crucial for the feature? This has some issues, we could add it very simply after we fully migrate the state to redis

packages/api/internal/orchestrator/pause_instance.go
```diff
@@ -97,9 +101,40 @@ func (o *Orchestrator) pauseSandbox(ctx context.Context, node *nodemanager.Node,
 		return fmt.Errorf("error pausing sandbox: %w", err)
 	}
 
+	o.storePausedSandbox(ctx, sbx)
+
 	return nil
 }
 
+func (o *Orchestrator) storePausedSandbox(ctx context.Context, sbx sandbox.Sandbox) {
```
jakubno: I think this could be also separate PR as it's only optimization

packages/client-proxy/internal/proxy/proxy.go
```diff
@@ -45,7 +99,46 @@ func catalogResolution(ctx context.Context, sandboxId string, c catalog.Sandboxe
...
+		s, err := c.GetSandbox(ctx, sandboxId)
+		if err == nil {
+			return s.OrchestratorIP, nil
+		}
+
+		if !errors.Is(err, catalog.ErrSandboxNotFound) {
+			return "", fmt.Errorf("failed to get sandbox from catalog during resume: %w", err)
+		}
+
+		time.Sleep(resumeWaitInterval)
```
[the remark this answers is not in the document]
matthewlouisbrockman: fix in

packages/api/internal/handlers/proxy_grpc.go
```diff
@@ -0,0 +1,236 @@
...
+		nil,
+	)
+	if apiErr != nil {
+		return nil, status.Errorf(codes.Internal, "resume failed: %s", apiErr.ClientMsg)
+	}
+
+	return &emptypb.Empty{}, nil
+}
+
+func (s *ProxySandboxService) resolveAuthTeam(ctx context.Context, snapshotTeamID uuid.UUID) (*teamtypes.Team, bool, error) {
```
jakubno: how do we know api key here?

packages/api/internal/handlers/sandbox_get.go
```diff
@@ -78,6 +78,8 @@ func (a *APIStore) GetSandboxesSandboxID(c *gin.Context, id string) {
 			EnvdAccessToken: sbx.EnvdAccessToken,
 			Domain:          sbxDomain,
 		}
+		autoPause := sbx.AutoPause
```
jakubno: Can you add a PR fixing this, so we get it to upstream quicker? 🙏🏻

packages/api/internal/handlers/proxy_grpc.go
```diff
@@ -0,0 +1,236 @@
...
+
+	return nil, false, nil
+}
+
+func autoResumePolicyFromMetadata(metadata map[string]string) string {
+	if metadata == nil {
+		return "null"
+	}
+
+	value := strings.TrimSpace(strings.ToLower(metadata["auto_resume"]))
```
jakubno: metadata are user defined, we should use a new field

packages/api/internal/handlers/proxy_grpc.go
```diff
@@ -0,0 +1,236 @@
...
+type ProxySandboxService struct {
+	proxygrpc.UnimplementedProxySandboxServiceServer
+	api *APIStore
+}
+
+func NewProxySandboxService(api *APIStore) *ProxySandboxService {
+	return &ProxySandboxService{api: api}
+}
+
+func (s *ProxySandboxService) GetPausedInfo(ctx context.Context, req *proxygrpc.SandboxPausedInfoRequest) (*proxygrpc.SandboxPausedInfoResponse, error) {
```
jakubno: what's this endpoint good for?

packages/api/internal/handlers/proxy_grpc.go
```diff
@@ -0,0 +1,236 @@
...
+	if authTeam != nil {
+		team = authTeam
+	} else {
+		team, err = dbapi.GetTeamByID(ctx, s.api.authDB, snap.Snapshot.TeamID)
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "failed to load team: %v", err)
+		}
+	}
+
+	timeout := sandbox.SandboxTimeoutDefault
```
jakubno, line 73: Relevant to this, which is even more low hanging fruit, we should add auto extend timeout on a request

packages/api/internal/handlers/proxy_grpc.go
```diff
@@ -0,0 +1,236 @@
...
+		nodeID,
+		snap.Snapshot.BaseEnvID,
+		autoPause,
+		envdAccessToken,
+		snap.Snapshot.AllowInternetAccess,
+		network,
+		nil,
+	)
+	if apiErr != nil {
+		return nil, status.Errorf(codes.Internal, "resume failed: %s", apiErr.ClientMsg)
```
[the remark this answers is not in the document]
matthewlouisbrockman: fixed

packages/api/internal/orchestrator/lifecycle.go
```diff
@@ -39,4 +39,10 @@ func (o *Orchestrator) addSandboxToRoutingTable(ctx context.Context, sandbox san
 	if err != nil {
 		logger.L().Error(ctx, "error adding routing record to catalog", zap.Error(err), logger.WithSandboxID(sandbox.SandboxID))
 	}
+
+	if o.pausedCatalog != nil {
+		if err := o.pausedCatalog.DeletePaused(ctx, sandbox.SandboxID); err != nil {
```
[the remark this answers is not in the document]
matthewlouisbrockman: fixed

packages/client-proxy/internal/proxy/paused_checker.go
```diff
@@ -0,0 +1,159 @@
...
+	autoResumeEnabled bool
+	client            *http.Client
+}
+
+type PausedInfo struct {
+	Paused           bool
+	AutoResumePolicy string
+}
+
+func NewApiPausedSandboxChecker(baseURL, adminToken, apiKey string, autoResumeEnabled bool) (PausedSandboxChecker, error) {
```
jakubno: I don't see this used

packages/client-proxy/internal/proxy/paused_checker_grpc.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+type grpcPausedSandboxChecker struct {
+	conn   *grpc.ClientConn
+	client proxygrpc.ProxySandboxServiceClient
+}
+
+func NewGrpcPausedSandboxChecker(address string) (PausedSandboxChecker, error) {
+	if strings.TrimSpace(address) == "" {
+		return nil, nil
```
jakubno: it should probably return an error

packages/client-proxy/internal/proxy/proxy.go
```diff
@@ -30,10 +32,64 @@ const (
...
+			var pausedFound bool
+
+			if pausedCatalog != nil {
+				info, pausedErr := pausedCatalog.GetPaused(ctx, sandboxId)
+				if pausedErr != nil {
+					if !errors.Is(pausedErr, catalog.ErrPausedSandboxNotFound) {
+						logger.L().Warn(ctx, "paused catalog lookup failed", zap.Error(pausedErr), logger.WithSandboxID(sandboxId))
+					}
+				} else {
+					pausedInfo = PausedInfo{
```
jakubno: why?

packages/client-proxy/internal/proxy/proxy.go
```diff
@@ -30,10 +32,64 @@ const (
...
+	resumeWaitInterval         = 200 * time.Millisecond
+	resumeWaitTimeout          = 30 * time.Second
+	resumeTimeoutSeconds int32 = 600
+)
+
+func catalogResolution(ctx context.Context, sandboxId string, c catalog.SandboxesCatalog, pausedCatalog catalog.PausedSandboxesCatalog, pausedChecker PausedSandboxChecker, autoResumeEnabled bool, requestHasAuth bool) (string, error) {
 	s, err := c.GetSandbox(ctx, sandboxId)
 	if err != nil {
 		if errors.Is(err, catalog.ErrSandboxNotFound) {
+			var pausedInfo PausedInfo
```
jakubno: can we maybe extract the logic to separate function and add a few comments 🙏🏻

packages/api/internal/handlers/sandbox_metadata_update.go
```diff
@@ -0,0 +1,46 @@
...
+
+	body, err := utils.ParseBody[api.PatchSandboxesSandboxIDMetadataJSONRequestBody](ctx, c)
+	if err != nil {
+		a.sendAPIStoreError(c, http.StatusBadRequest, fmt.Sprintf("Error when parsing request: %s", err))
+		telemetry.ReportCriticalError(ctx, "error when parsing request", err)
+
+		return
+	}
+
+	if body.Metadata == nil {
```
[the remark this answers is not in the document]
matthewlouisbrockman: PatchSandboxesSandboxIDMetadata calls validateAutoResumeMetadata, and it normalizes/
  validates auto_resume to any|authed|null (see packages/api/internal/handlers/metadata_validation.go and
  packages/api/internal/handlers/sandbox_metadata_update.go).

packages/client-proxy/internal/proxy/paused_checker.go
```diff
@@ -0,0 +1,159 @@
...
+		return nil, fmt.Errorf("invalid API base URL: %w", err)
+	}
+
+	return &apiPausedSandboxChecker{
+		baseURL:           strings.TrimRight(baseURL, "/"),
+		adminToken:        adminToken,
+		apiKey:            apiKey,
+		autoResumeEnabled: autoResumeEnabled,
+		client: &http.Client{
+			Timeout: 2 * time.Second,
```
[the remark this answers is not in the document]
matthewlouisbrockman: making 'em both longer

packages/api/internal/handlers/proxy_grpc.go
matthewlouisbrockman: kk

iac/provider-gcp/nomad-cluster/scripts/start-client.sh
matthewlouisbrockman: mmm, can pull that - i break deploying to dev when i don't have that

matthewlouisbrockman: setting disk size to 200g fixes it

iac/provider-gcp/nomad/jobs/client-proxy.hcl
matthewlouisbrockman: yep! moved over

iac/provider-gcp/nomad/main.tf
matthewlouisbrockman: yep, old code not needed anymore, shoulda removed

packages/shared/pkg/grpc/proxy/proxy.proto
matthewlouisbrockman: enum it is

packages/api/internal/handlers/sandbox_metadata_update.go
matthewlouisbrockman: moved out

packages/api/internal/orchestrator/update_lifecycle.go
matthewlouisbrockman: k, rm'd

packages/api/internal/handlers/proxy_grpc.go
matthewlouisbrockman: we pull the token from the auth headers, um, i guess need to double check that's there like i think it it

packages/api/internal/orchestrator/pause_instance.go
matthewlouisbrockman: rm

packages/client-proxy/internal/proxy/paused_checker.go
matthewlouisbrockman: oops

packages/client-proxy/internal/proxy/paused_checker_grpc.go
matthewlouisbrockman: removed him altogether

packages/client-proxy/internal/proxy/proxy.go
matthewlouisbrockman: for when pulling from redis, no longer doing that though

packages/client-proxy/internal/proxy/proxy.go
matthewlouisbrockman: done

packages/api/internal/handlers/proxy_grpc.go
matthewlouisbrockman: makes sense, added to linear https://linear.app/e2b/issue/ENG-3463/auto-extend-timeout-on-a-request

packages/api/internal/handlers/sandbox_get.go
matthewlouisbrockman: pull from snapshot

packages/api/internal/handlers/proxy_grpc.go
matthewlouisbrockman: k, this was the bulk of the changes, i think g2g

packages/client-proxy/internal/proxy/paused_checker_grpc.go
matthewlouisbrockman: err wait returning error

packages/api/internal/handlers/sandbox_get.go
matthewlouisbrockman: ohhhh it's cause pause != resume got it

packages/api/internal/handlers/proxy_grpc.go
matthewlouisbrockman: it's the gate for handling the logic when the sandbox call fails - if it's paused and if the user can resume

matthewlouisbrockman: (called by NewGrpcPausedSandboxChecker below)

spec/openapi.yml
```diff
@@ -520,6 +526,12 @@ components:
           description:
             Allow sandbox to access the internet. When set to false, it behaves the same as specifying denyOut
             to 0.0.0.0/0 in the network config.
+        auto_resume:
```
ValentaTomas: I know there are several places where this varies, but we should sync on snake vs camel case in the spec.

spec/openapi.yml
```diff
@@ -374,6 +374,9 @@ components:
           description: Identifier of the client
         envdVersion:
           $ref: "#/components/schemas/EnvdVersion"
+        autoPause:
```
ValentaTomas: We return this info, because it can be dynamically changed now?

ValentaTomas: One note—I think we only expose the autopause in the beta SDK for now.

matthewlouisbrockman: eep it could be changed in original version, can't be changed anymore and missed that one

matthewlouisbrockman: oh, wait, does it actually hurt to keep him in

spec/openapi.yml
matthewlouisbrockman: `allow_internet_access` above it is snake, can go with the camel case the rest of the openapi has

spec/openapi.yml
matthewlouisbrockman: removed

packages/client-proxy/internal/proxy/proxy.go
```diff
@@ -45,7 +61,127 @@ func catalogResolution(ctx context.Context, sandboxId string, c catalog.Sandboxe
...
+	pausedInfo PausedInfo,
+) (string, error) {
+	// Shared implementation so callers can reuse pre-fetched paused info.
+	logSleeping(ctx, sandboxId)
+
+	// Decide if we are allowed to auto-resume for this request.
+	canAutoResume := shouldAutoResume(pausedInfo.AutoResumePolicy, autoResumeEnabled, requestHasAuth) && pausedChecker != nil
+	if canAutoResume {
+		logger.L().Info(ctx, "auto-resuming sandbox", logger.WithSandboxID(sandboxId))
+		if err := pausedChecker.Resume(ctx, sandboxId, resumeTimeoutSeconds); err != nil {
```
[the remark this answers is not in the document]
matthewlouisbrockman: fixed

spec/openapi.yml
matthewlouisbrockman: switched to camelCase

.github/workflows/integration_tests.yml
```diff
@@ -26,11 +26,21 @@ jobs:
       - name: Start Services
         uses: ./.github/actions/start-services
 
+      - name: Debug Service Config
```
jakubno: Clean up

iac/provider-gcp/nomad/jobs/client-proxy.hcl
```diff
@@ -101,6 +101,8 @@ job "client-proxy" {
         REDIS_CLUSTER_URL   = "${redis_cluster_url}"
         REDIS_TLS_CA_BASE64 = "${redis_tls_ca_base64}"
 
+        API_GRPC_ADDRESS = "api-grpc.service.consul:${api_grpc_port}"
```
jakubno: can we pass it as a whole addresses, similarly as LOGS_COLLECTOR_ADDRESS?

tests/integration/internal/utils/team.go
```diff
@@ -86,3 +88,31 @@ func CreateAPIKey(t *testing.T, ctx context.Context, c *api.ClientWithResponses,
 
 	return apiKey.JSON201.Key
 }
+
+func CreateAPIKeyInDB(t *testing.T, db *setup.Database, userID uuid.UUID, teamID uuid.UUID) string {
```
jakubno: Why adding this instead of using already present `CreateAPIKey`?

tests/integration/internal/utils/request.go
```diff
@@ -21,9 +22,23 @@ func NewRequest(sbx *api.Sandbox, url *url.URL, port int, extraHeaders *http.Hea
 	if url.Hostname() == "localhost" {
 		host = fmt.Sprintf("%d-%s-%s.%s", port, sbx.SandboxID, sbx.ClientID, "localhost")
 	} else {
-		// Extract top level domain from EnvdProxy
-		eTLD, _ := publicsuffix.EffectiveTLDPlusOne(url.Hostname())
-		host = fmt.Sprintf("%d-%s-%s.%s:%s", port, sbx.SandboxID, sbx.ClientID, eTLD, url.Port())
+		baseDomain := url.Hostname()
```
jakubno, line 25: could you please explain what this should exactly do?

packages/shared/pkg/grpc/proxy/proxy.proto
```diff
@@ -0,0 +1,45 @@
...
+option go_package = "github.com/e2b-dev/infra/packages/shared/pkg/grpc/proxy;proxygrpc";
+
+// Request paused snapshot info for a sandbox.
+message SandboxPausedInfoRequest {
+  // Sandbox ID (short or full), as used by the client-proxy.
+  string sandbox_id = 1;
+}
+
+// AutoResumePolicy describes when a paused sandbox should auto-resume.
+enum AutoResumePolicy {
```
jakubno, line 10: what's the diff between AUTO_RESUME_POLICY_UNSPECIFIED and AUTO_RESUME_POLICY_NULL?

packages/shared/pkg/grpc/proxy/proxy.proto
```diff
@@ -0,0 +1,45 @@
...
+
+package proxy;
+
+import "google/protobuf/empty.proto";
+
+option go_package = "github.com/e2b-dev/infra/packages/shared/pkg/grpc/proxy;proxygrpc";
+
+// Request paused snapshot info for a sandbox.
+message SandboxPausedInfoRequest {
+  // Sandbox ID (short or full), as used by the client-proxy.
```
jakubno: probably not needed to mention (short or full)

packages/db/migrations/20260204180000_sandbox_resume_metadata.sql
```diff
@@ -0,0 +1,12 @@
+-- +goose Up
+-- +goose StatementBegin
+alter table snapshots
+    add column sandbox_resumes_on varchar(10) null;
```
jakubno: I thought we agreed to put it into the config field

packages/client-proxy/main.go
```diff
@@ -134,12 +135,32 @@ func run() int {
...
+	}
+
 	// Proxy sandbox http traffic to orchestrator nodes
-	trafficProxy, err := e2bproxy.NewClientProxy(
+	autoResumeEnabled := featureFlagsClient.BoolFlag(ctx, featureflags.SandboxAutoResumeFlag, featureflags.ServiceContext(serviceName))
+	if !config.AutoResumeEnabled {
+		autoResumeEnabled = false
+	}
+	launchDarklyEnabled := strings.TrimSpace(os.Getenv("LAUNCH_DARKLY_API_KEY")) != ""
+	l.Info(ctx, "auto-resume configuration",
```
jakubno: is this debug?

packages/api/internal/handlers/proxy_grpc.go
```diff
@@ -0,0 +1,335 @@
...
+
+	timeout := sandbox.SandboxTimeoutDefault
+	if req.GetTimeoutSeconds() > 0 {
+		timeout = time.Duration(req.GetTimeoutSeconds()) * time.Second
+	}
+	if timeout > time.Duration(team.Limits.MaxLengthHours)*time.Hour {
+		return nil, status.Error(codes.InvalidArgument, "timeout exceeds team limit")
+	}
+
+	lock, lockAcquired, lockErr := s.tryAcquireResumeLock(ctx, sandboxID)
```
jakubno, line 81: could you explain what are you trying to do here?

packages/api/internal/handlers/proxy_grpc.go
```diff
@@ -0,0 +1,335 @@
...
+	if lockErr != nil {
+		logger.L().Warn(ctx, "Failed to acquire proxy resume lock, proceeding without lock", zap.Error(lockErr), logger.WithSandboxID(sandboxID))
+	} else if !lockAcquired {
+		if waitErr := s.api.orchestrator.WaitForSandboxInRoutingCatalog(ctx, sandboxID, proxyResumeWaitTimeout); waitErr != nil {
```
jakubno: could you explain this? btw I remembered you mentioning some issue with the redis catalog missing the key, I've just checked and the key IS SET before the resume returns, it's done as a part of resume, so if you have issues with that, it may be caching?

packages/api/internal/handlers/proxy_grpc.go
```diff
@@ -0,0 +1,335 @@
...
+	proxygrpc.UnimplementedSandboxServiceServer
+
+	api *APIStore
+}
+
+func NewSandboxService(api *APIStore) *SandboxService {
+	return &SandboxService{api: api}
+}
+
+func (s *SandboxService) GetPausedInfo(ctx context.Context, req *proxygrpc.SandboxPausedInfoRequest) (*proxygrpc.SandboxPausedInfoResponse, error) {
```
jakubno: why can't we use only 1 endpoint? this way you'll end up doing the same DB queries twice for the sandbox which is supposed to be resumed

packages/api/internal/orchestrator/nodemanager/sandboxes.go
```diff
@@ -86,6 +86,7 @@ func (n *Node) GetSandboxes(ctx context.Context) ([]sandbox.Sandbox, error) {
 				teamID,
 				buildID,
 				config.GetMetadata(),
+				nil,
```
jakubno: why?

packages/api/internal/orchestrator/delete_instance.go
```diff
@@ -84,7 +85,7 @@ func (o *Orchestrator) removeSandboxFromNode(ctx context.Context, sbx sandbox.Sa
 
 	// Only remove from routing table if the node is managed by Nomad
 	// For remote cluster nodes we are using gPRC metadata for routing registration instead
-	if node.IsNomadManaged() {
+	if node.IsNomadManaged() || env.IsLocal() {
```
jakubno, line 88: could you do separate PR for this? 🙏🏻

packages/client-proxy/internal/cfg/model.go
```diff
@@ -9,6 +9,9 @@ type Config struct {
 	RedisURL         string `env:"REDIS_URL"`
 	RedisClusterURL  string `env:"REDIS_CLUSTER_URL"`
 	RedisTLSCABase64 string `env:"REDIS_TLS_CA_BASE64"`
+
+	ApiGrpcAddress    string `env:"API_GRPC_ADDRESS"`
+	AutoResumeEnabled bool   `env:"AUTO_RESUME_ENABLED" envDefault:"true"`
```
jakubno: is this even used? I thought we're using feature flag

packages/client-proxy/internal/proxy/paused_checker_grpc.go
```diff
@@ -0,0 +1,65 @@
...
+	"fmt"
+	"strings"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+
+	proxygrpc "github.com/e2b-dev/infra/packages/shared/pkg/grpc/proxy"
+)
+
+type grpcPausedSandboxChecker struct {
```
jakubno: the name is little bit off as it also resumes the sandbox

packages/api/internal/handlers/proxy_grpc.go
```diff
@@ -0,0 +1,335 @@
...
+		team = authTeam
+	} else {
+		team, err = dbapi.GetTeamByID(ctx, s.api.authDB, snap.Snapshot.TeamID)
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "failed to load team: %v", err)
+		}
+	}
+
+	timeout := sandbox.SandboxTimeoutDefault
+	if req.GetTimeoutSeconds() > 0 {
```
jakubno, line 66: I think this should be from the db as user should be allow to set it themself

packages/client-proxy/internal/proxy/proxy.go
```diff
@@ -45,7 +63,144 @@ func catalogResolution(ctx context.Context, sandboxId string, c catalog.Sandboxe
 	return s.OrchestratorIP, nil
 }
 
+func handlePausedSandbox(
```
jakubno, line 61: i would merge two calls, feels like an extra call with no added value

packages/client-proxy/internal/proxy/proxy.go
```diff
@@ -45,7 +63,144 @@ func catalogResolution(ctx context.Context, sandboxId string, c catalog.Sandboxe
...
+	ctx context.Context,
+	sandboxId string,
+	c catalog.SandboxesCatalog,
+	pausedChecker PausedSandboxChecker,
+	autoResumeEnabled bool,
+	requestHasAuth bool,
+	pausedInfo PausedInfo,
+) (string, error) {
+	// Shared implementation so callers can reuse pre-fetched paused info.
+	logSleeping(ctx, sandboxId)
```
jakubno: ?

packages/client-proxy/internal/proxy/proxy.go
```diff
@@ -45,7 +63,144 @@ func catalogResolution(ctx context.Context, sandboxId string, c catalog.Sandboxe
...
+	canAutoResume := shouldAutoResume(pausedInfo.AutoResumePolicy, autoResumeEnabled, requestHasAuth) && pausedChecker != nil
+	if canAutoResume {
+		logger.L().Info(ctx, "auto-resuming sandbox", logger.WithSandboxID(sandboxId))
+		if err := pausedChecker.Resume(ctx, sandboxId, resumeTimeoutSeconds); err != nil {
+			logger.L().Warn(ctx, "auto-resume failed", zap.Error(err), logger.WithSandboxID(sandboxId))
+			if isAuthResumeError(err) {
+				canAutoResume = false
+			}
+		} else {
+			nodeIP, waitErr := waitForCatalog(ctx, sandboxId, c)
```
jakubno: isn't this all caused by having 500ms ttl in redis local cache?

packages/client-proxy/internal/proxy/proxy.go
```diff
@@ -137,3 +299,70 @@ func NewClientProxy(meterProvider metric.MeterProvider, serviceName string, port
 
 	return proxy, nil
 }
+
+func pausedFallbackHandler(
```
jakubno, line 273: why do we have to have the logic on both places? catalog resolution and here?

packages/db/migrations/20260204180000_sandbox_resume_metadata.sql
matthewlouisbrockman: i think the `sandbox_resumes_on` was its own column, the configs around it were gonna be in config

packages/api/internal/orchestrator/delete_instance.go
matthewlouisbrockman: lemme see if it passes tests without it

tests/integration/internal/utils/request.go
matthewlouisbrockman: I believe this is mainly used in tests. Added tests to make sure the client proxy was routing correctly when I was resuming boxes and passing the auth to track where things were breaking, but was failing because too much was getting stripped from the url on the dev box which has the double subdomain and routed requests to the base test domain cause it would strip it (so like, blah.e2btest.dev becomes e2btest.dev)

I think old logic was causing all tests using this on dev boxes to fail actually; `closed_port_test.go` also fails without it - reverted and ran against my devbox and got 
`DONE 2 runs, 2 tests, 1 failure, 1 error in 7.460s`

added back in, 
`DONE 2 tests in 4.760s`

packages/shared/pkg/grpc/proxy/proxy.proto
matthewlouisbrockman: combined them, didn't clean up, finishing the combination

packages/client-proxy/main.go
```diff
@@ -134,12 +135,32 @@ func run() int {
 	info := &internal.ServiceInfo{}
 	info.SetStatus(ctx, internal.Healthy)
 
+	pausedChecker, err := e2bproxy.NewGrpcPausedSandboxChecker(config.ApiGrpcAddress)
```
sitole: Can we make the pause checker optional if the API gRPC address was not provided?

packages/shared/pkg/grpc/proxy/proxy.proto
```diff
@@ -0,0 +1,45 @@
...
+// Request to resume a paused sandbox.
+message SandboxResumeRequest {
+  // Sandbox ID to resume.
+  string sandbox_id = 1;
+  // Optional timeout override in seconds (must be within team limits).
+  int32 timeout_seconds = 2;
+}
+
+// SandboxService is used by the client-proxy to check paused state
+// and resume sandboxes. Authorization/X-API-Key headers are forwarded
```
sitole: Is this true? gRPC client in client proxy is created without auth.

packages/client-proxy/main.go
matthewlouisbrockman: rmd

packages/api/internal/handlers/proxy_grpc.go
matthewlouisbrockman: it was a github actions redis issue; checking how much of this i can yeet

matthewlouisbrockman: o, wait, this was 2 days ago checking

packages/api/internal/handlers/proxy_grpc.go
matthewlouisbrockman: removing, i did not sleep much 2 night ago, yeah, this is supposed to just rely on the api's resume logic now instead of its own

packages/api/internal/orchestrator/delete_instance.go
matthewlouisbrockman: ohhh separate PR and merge into this branch

packages/api/internal/handlers/proxy_grpc.go
matthewlouisbrockman: oh, I think we want that guy. Only want to do 1 resume call at a time against the API, so there's a lock on asking the API to resume.

packages/api/internal/handlers/proxy_grpc.go
matthewlouisbrockman: k, yeah, so basically if it doesn't get the lock, it assumes someone else is resuming and waits for it to resume and then returns

packages/api/internal/orchestrator/nodemanager/sandboxes.go
matthewlouisbrockman: was thinking orchestrator doesn't need to know although actually i guess it does sometimes for when the api pulls from node sync

matthewlouisbrockman: (tldr, when wait a long time and pause it and then resume it drops it so do need to pass it in here i think)

packages/client-proxy/internal/proxy/proxy.go
matthewlouisbrockman: comparing vs catalogResolution? it's checking first if it's in the catalog and trying to connect; however, since sometimes u can try to connect to a sandbox that's still pausing the pausedFallbackHandler is a backup that fires off a resume after failing to connect to someone that it should have been able to connect to

packages/client-proxy/internal/proxy/proxy.go
matthewlouisbrockman: just making sure the sandbox is up and ready

packages/client-proxy/internal/proxy/proxy.go
matthewlouisbrockman: renamed to logSandboxAppearsPaused

packages/api/internal/orchestrator/nodemanager/sandboxes.go
matthewlouisbrockman: passing it in

packages/api/internal/orchestrator/delete_instance.go
matthewlouisbrockman: https://github.com/e2b-dev/infra/pull/1854

packages/api/internal/handlers/proxy_grpc.go
matthewlouisbrockman: this ends up being complicated i think - it looks like we only have the concept of EOL for the sandbox, not a kill after inactivity? it's not looking like we haev the original... that something we should maybe put in a different PR later?

packages/client-proxy/internal/proxy/proxy.go
matthewlouisbrockman: merged

packages/shared/pkg/grpc/proxy/proxy.proto
matthewlouisbrockman: `ctx = withProxyAuthMetadata(ctx, r.Header)`

passes it in i believe, or were you asking about something else

packages/client-proxy/main.go
matthewlouisbrockman: yep!

packages/client-proxy/internal/proxy/proxy.go
matthewlouisbrockman: actually deleted it when just kept logic on api

packages/api/internal/handlers/proxy_grpc.go
matthewlouisbrockman: k, just moving all logic to API

packages/client-proxy/internal/cfg/model.go
matthewlouisbrockman: fair enough, yeeting

packages/api/internal/handlers/proxy_grpc.go
matthewlouisbrockman: gonna go figure out how to store on create

packages/api/internal/handlers/proxy_grpc.go
matthewlouisbrockman: o, will switch this to waiting on the lock to release

packages/db/migrations/20260204180000_sandbox_resume_metadata.sql
matthewlouisbrockman: moving it into config

packages/client-proxy/internal/proxy/proxy.go
matthewlouisbrockman: got rid of polling seems to work

packages/client-proxy/internal/proxy/paused_checker.go
```diff
@@ -0,0 +1,7 @@
+package proxy
+
+import "context"
+
+type PausedSandboxChecker interface {
```
jakubno:
```suggestion
type PausedSandboxResumer interface {
```

packages/client-proxy/main.go
```diff
@@ -134,12 +135,28 @@ func run() int {
...
+
+			return 1
+		}
+	} else {
+		l.Warn(ctx, "API gRPC address not set; paused sandbox checks disabled")
+	}
+
 	// Proxy sandbox http traffic to orchestrator nodes
-	trafficProxy, err := e2bproxy.NewClientProxy(
+	autoResumeEnabled := featureFlagsClient.BoolFlag(ctx, featureflags.SandboxAutoResumeFlag, featureflags.ServiceContext(serviceName))
```
jakubno, line 151: can we do it on the request?

packages/db/pkg/types/types.go
```diff
@@ -30,8 +30,10 @@ type SandboxNetworkConfig struct {
 }
 
 type PausedSandboxConfig struct {
-	Version string                `json:"version"`
-	Network *SandboxNetworkConfig `json:"network,omitempty"`
+	Version               string                `json:"version"`
+	Network               *SandboxNetworkConfig `json:"network,omitempty"`
+	SandboxResumesOn      *string               `json:"sandboxResumesOn,omitempty"`
+	SandboxTimeoutSeconds *int32                `json:"sandboxTimeoutSeconds,omitempty"`
```
jakubno, line 36: Can we also encapsulate this similarly as a network config?

packages/shared/pkg/grpc/proxy/auto_resume_policy.go
```diff
@@ -0,0 +1,44 @@
...
+	switch value {
+	case "any":
+		return AutoResumePolicy_AUTO_RESUME_POLICY_ANY
+	case "authed":
+		return AutoResumePolicy_AUTO_RESUME_POLICY_AUTHED
+	case "null", "":
+		return AutoResumePolicy_AUTO_RESUME_POLICY_NULL
+	default:
+		return AutoResumePolicy_AUTO_RESUME_POLICY_NULL
+	}
```
jakubno, line 15: either
```suggestion
	default:
		return AutoResumePolicy_AUTO_RESUME_POLICY_NULL
	}
```

or at least add a log for the unexpected value (can it be `""`)?
```
	case "null":
		return AutoResumePolicy_AUTO_RESUME_POLICY_NULL
	default:
		logger.L().Warn(...)
		return AutoResumePolicy_AUTO_RESUME_POLICY_NULL
	}
```

Similarly for the other places

packages/client-proxy/internal/proxy/proxy.go
```diff
@@ -71,13 +186,19 @@ func NewClientProxy(meterProvider metric.MeterProvider, serviceName string, port
...
+				if errors.Is(err, ErrNodeNotFound) {
+					return nil, reverseproxy.NewErrSandboxNotFound(sandboxId)
+				}
+
 				if !errors.Is(err, ErrNodeNotFound) {
 					l.Warn(ctx, "failed to resolve node ip with Redis resolution", zap.Error(err))
 				}
 
-				return nil, reverseproxy.NewErrSandboxNotFound(sandboxId)
+				return nil, err
```
jakubno, line 211: I think this change doesn't do what you may intended

packages/client-proxy/internal/proxy/proxy.go
```diff
@@ -137,3 +259,61 @@ func NewClientProxy(meterProvider metric.MeterProvider, serviceName string, port
 
 	return proxy, nil
 }
+
+func pausedFallbackHandler(
```
jakubno, line 273: I think we should handle this better, this is trying to minimize the issue surface. I'd for now remove and we can decide later how to solve this. Just create a linear task with the described problem (sandbox may be pausing)

scripts/smoke-proxy-auto-resume-concurrent.sh
jakubno, line 1: do we need write it bash? could we use integration tests for the stuff we want to test as otherwise there's no guarantee it's passing

packages/api/internal/orchestrator/keep_alive.go
```diff
@@ -17,7 +17,7 @@ import (
 
 var errMaxInstanceLengthExceeded = fmt.Errorf("max instance length exceeded")
 
-func (o *Orchestrator) KeepAliveFor(ctx context.Context, teamID uuid.UUID, sandboxID string, duration time.Duration, allowShorter bool) *api.APIError {
+func (o *Orchestrator) KeepAliveFor(ctx context.Context, teamID uuid.UUID, sandboxID string, duration time.Duration, allowShorter bool, persistTimeout bool) *api.APIError {
```
jakubno, line 20: lets revert this as discussed

spec/openapi.yml
```diff
@@ -520,6 +523,12 @@ components:
           description:
             Allow sandbox to access the internet. When set to false, it behaves the same as specifying denyOut
             to 0.0.0.0/0 in the network config.
+        autoResume:
+          type: string
+          enum:
+            - any
+            - authed
```
jakubno: What about and making it required?
```suggestion
          enum:
            - any
            - authed
            - off
```

packages/shared/pkg/grpc/proxy/proxy.proto
```diff
@@ -0,0 +1,29 @@
...
+
+import "google/protobuf/empty.proto";
+
+option go_package = "github.com/e2b-dev/infra/packages/shared/pkg/grpc/proxy;proxygrpc";
+
+// AutoResumePolicy describes when a paused sandbox should auto-resume.
+enum AutoResumePolicy {
+  AUTO_RESUME_POLICY_NULL = 0;
+  AUTO_RESUME_POLICY_ANY = 1;
+  AUTO_RESUME_POLICY_AUTHED = 2;
```
jakubno: let's remove authed for now to simplify

packages/orchestrator/orchestrator.proto
```diff
@@ -46,6 +46,10 @@ message SandboxConfig {
   optional bool allow_internet_access = 21;
 
   optional SandboxNetworkConfig network = 22;
+  // Auto-resume policy for paused sandboxes (stored in API snapshots).
+  optional string sandbox_resumes_on = 23;
+  // TTL in seconds as originally configured by the API.
+  optional int32 sandbox_timeout_seconds = 24;
```
jakubno, line 52: scope the resume similarly as network

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -109,6 +109,8 @@ type Metadata struct {
 
 	StartedAt time.Time
 	EndAt     time.Time
+	// Original TTL configured by the API.
+	TimeoutSeconds *int32
```
jakubno, line 113: just keep it in the api config

jakubno (changes requested):

packages/client-proxy/internal/proxy/paused_checker.go
matthewlouisbrockman: k

packages/shared/pkg/grpc/proxy/proxy.proto
matthewlouisbrockman: aight here goes

packages/client-proxy/main.go
matthewlouisbrockman: err, will we need to do this on the API then? or does the client have the ability to pull per user darkly stuff

PR #1832, opened 2026-02-03
chore: limit sandbox egress tcp connections

dobrac:
> [!NOTE]
> **Medium Risk**
> Touches the TCP firewall connection handling path and introduces connection rejection based on a runtime flag, which could cause unexpected egress failures if misconfigured or if limiter bookkeeping is wrong.
> 
> **Overview**
> Adds a per-sandbox connection limiter to the TCP egress firewall proxy, enforcing a configurable max concurrent connections per sandbox via the new `tcpfirewall-max-connections-per-sandbox` feature flag (negative = unlimited) and rejecting excess connections with a new `limit_exceeded` error type.
> 
> Wires feature flags into `tcpfirewall.New(...)` call sites (orchestrator, build tooling, benchmarks) and expands TCP firewall telemetry with a new per-sandbox connections histogram (`orchestrator.tcpfirewall.connections.per_sandbox`). Also updates an integration test to use the `envd.PostInitJSONRequestBody` type.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 20f95158a67b381a78f90088ac089ea98e882674. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/orchestrator/internal/tcpfirewall/limiter.go
+++ b/packages/orchestrator/internal/tcpfirewall/limiter.go
@@ -1,24 +1,31 @@
 package tcpfirewall
 
 import (
-	"sync"
 	"sync/atomic"
+
+	"github.com/e2b-dev/infra/packages/shared/pkg/smap"
 )
 
 // ConnectionLimiter tracks and limits per-sandbox TCP connections.
 type ConnectionLimiter struct {
-	connections sync.Map // map[string]*atomic.Int64
+	connections *smap.Map[*atomic.Int64]
 }
 
 // NewConnectionLimiter creates a new ConnectionLimiter.
 func NewConnectionLimiter() *ConnectionLimiter {
-	return &ConnectionLimiter{}
+	return &ConnectionLimiter{
+		connections: smap.New[*atomic.Int64](),
+	}
 }
 
 func (l *ConnectionLimiter) getCounter(sandboxID string) *atomic.Int64 {
-	counter, _ := l.connections.LoadOrStore(sandboxID, &atomic.Int64{})
+	return l.connections.Upsert(sandboxID, &atomic.Int64{}, func(exists bool, valueInMap, newValue *atomic.Int64) *atomic.Int64 {
+		if exists {
+			return valueInMap
+		}
 
-	return counter.(*atomic.Int64)
+		return newValue
+	})
 }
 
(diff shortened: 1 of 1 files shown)


packages/shared/pkg/feature-flags/flags.go
```diff
@@ -148,7 +148,7 @@ var (
 
 	// TCPFirewallMaxConnectionsPerSandbox is the maximum number of concurrent TCP firewall
 	// connections allowed per sandbox. 0 means no limit.
-	TCPFirewallMaxConnectionsPerSandbox = newIntFlag("tcpfirewall-max-connections-per-sandbox", 100)
+	TCPFirewallMaxConnectionsPerSandbox = newIntFlag("tcpfirewall-max-connections-per-sandbox", 0)
```
sitole: Istn -1 more explicit for infinite limits?

packages/orchestrator/internal/tcpfirewall/limiter.go
```diff
@@ -0,0 +1,71 @@
+package tcpfirewall
+
+import (
+	"sync"
+	"sync/atomic"
+)
+
+// ConnectionLimiter tracks and limits per-sandbox TCP connections.
+type ConnectionLimiter struct {
+	connections sync.Map // map[string]*atomic.Int64
```
jakubno: was there reason for using sync.Map instead of `smap.Map`?

packages/shared/pkg/feature-flags/flags.go
dobrac: I think 0 is fine, 0 is also the default integer value

dobrac: (in the condition it's effectively anything below 1:

```
if maxLimit > 0 && current >= int64(maxLimit) {
			return current, false
		}
```
)

dobrac: Will change to -1 and allow 0 connections to be defined

packages/orchestrator/internal/tcpfirewall/limiter.go
dobrac: no reason, will change it

dobrac: changed

PR #1842, opened 2026-02-03
Use v2 endpoint for service discovery

sitole:
- Uses the new V2 endpoint for remote cluster discovery.
- Use of a limited OpenAPI spec that shows only public endpoints.

---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it changes a core discovery code path and the response schema it depends on; failures would impact cross-cluster node discovery until all services are deployed with the new endpoint.
> 
> **Overview**
> Remote cluster discovery now calls `GET /v1/service-discovery` (via `V1ServiceDiscoveryWithResponse`) and reads orchestrators from the new `ClusterServiceDiscovery` response shape.
> 
> The Edge OpenAPI spec and generated client were updated accordingly: adds `ClusterServiceDiscovery`, introduces `V1ServiceDiscovery` in the client, marks the old orchestrators-only endpoint as deprecated, and removes several previously-generated/internal endpoints and types (e.g., node drain/kill, traffic health check, `ClusterNode`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2a3caf0d2be0aed4383ba833a3511007424becc1. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api/internal/clusters/discovery/remote.go
+++ b/packages/api/internal/clusters/discovery/remote.go
@@ -29,7 +29,7 @@ func (sd *RemoteServiceDiscovery) Query(ctx context.Context) ([]Item, error) {
 	ctx, span := tracer.Start(ctx, "query-remote-cluster-nodes", trace.WithAttributes(telemetry.WithClusterID(sd.clusterID)))
 	defer span.End()
 
-	res, err := sd.client.V2ServiceDiscoveryGetOrchestratorsWithResponse(ctx)
+	res, err := sd.client.V1ServiceDiscoveryWithResponse(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get cluster instances from service discovery: %w", err)
 	}
--- a/packages/shared/pkg/http/edge/generated.go
+++ b/packages/shared/pkg/http/edge/generated.go
@@ -85,11 +85,6 @@ type ClusterNodeInfo struct {
 // ClusterNodeStatus State of the cluster node
 type ClusterNodeStatus string
 
-// ClusterOrchestratorDiscovery defines model for ClusterOrchestratorDiscovery.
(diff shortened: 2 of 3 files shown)


sitole:
We can safely replace directly as all remote clusters are already updated.

packages/api/internal/clusters/discovery/remote.go
```diff
@@ -29,7 +29,7 @@ func (sd *RemoteServiceDiscovery) Query(ctx context.Context) ([]Item, error) {
 	ctx, span := tracer.Start(ctx, "query-remote-cluster-nodes", trace.WithAttributes(telemetry.WithClusterID(sd.clusterID)))
 	defer span.End()
 
-	res, err := sd.client.V1ServiceDiscoveryGetOrchestratorsWithResponse(ctx)
+	res, err := sd.client.V2ServiceDiscoveryGetOrchestratorsWithResponse(ctx)
```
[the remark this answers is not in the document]
sitole: https://github.com/e2b-dev/infra/pull/1842#issuecomment-3843479253

packages/api/internal/clusters/discovery/remote.go
```diff
@@ -42,7 +42,7 @@ func (sd *RemoteServiceDiscovery) Query(ctx context.Context) ([]Item, error) {
 		return nil, errors.New("request to get nodes returned nil response")
 	}
 
-	nodes := *res.JSON200
+	nodes := res.JSON200.Orchestrators
```
djeebus, line 45: Do we still need to test for `res.JSON200 == nil`?

sitole: Probably yes, at least looks to me more cleaner

PR #1845, opened 2026-02-03
fix(envd): add Content-Disposition header to file downloads

beran-t:
## Summary
- Adds `Content-Disposition` header to `/files` endpoint responses using Go's `mime.FormatMediaType`
- Browsers now save downloaded files with the correct filename and extension
- Uses `inline` disposition so browsers can display files directly when supported

## Implementation
Single line addition using the stdlib `mime` package:
```go
w.Header().Set("Content-Disposition", mime.FormatMediaType("inline", map[string]string{"filename": filepath.Base(resolvedPath)}))
```

`mime.FormatMediaType` handles all the RFC 6266/5987 edge cases automatically:
- Quotes filenames containing spaces or special characters
- Escapes backslashes and quotes
- Uses `filename*=utf-8''...` encoding for non-ASCII filenames (e.g. Chinese characters)
- Preserves dotfiles like `.env` and `.gitignore`

## Test plan
- [x] Unit tests covering: simple filenames, spaces, quotes, backslashes, unicode, dotfiles, nested paths
- [ ] CI passes

### DIFF
--- a/packages/envd/main.go
+++ b/packages/envd/main.go
@@ -47,7 +47,7 @@ const (
 )
 
 var (
-	Version = "0.5.0"
+	Version = "0.5.1"
 
 	commitSHA string
 


sitole (approved):

packages/envd/internal/api/download.go
```diff
@@ -111,5 +113,6 @@ func (a *API) GetFiles(w http.ResponseWriter, r *http.Request, params GetFilesPa
 	}
 	defer file.Close()
 
+	w.Header().Set("Content-Disposition", mime.FormatMediaType("attachment", map[string]string{"filename": filepath.Base(resolvedPath)}))
```
jakubno: Do we have to set `attachment`? It may be breaking for some users

also can you implement both `filename` and `filename*` and make sure all recommendation followed?
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Disposition#as_a_response_header_for_the_main_body

sitole: Agree that inline is maybe better for back compatibility here.

beran-t: 1. Changed `attachment` to `inline` to avoid breaking changes 
2. Implemented RFC 6266 compliant header with both filename and filename*:
`Content-Disposition: inline; filename="download.pdf"; filename*=utf-8''%E6%96%87%E6%A1%A3.pdf`

Following all MDN recommendations:

- filename: ASCII-only fallback for older browsers (non-ASCII chars stripped, falls back to “download” + extension if needed)
- filename*: UTF-8 percent-encoded per RFC 5987 for modern browsers
- Quotes and backslashes properly escaped in filename
- No percent escapes in filename (as recommended, since browsers handle them inconsistently)

packages/envd/internal/api/download.go
```diff
@@ -4,15 +4,67 @@ import (
...
+	var asciiOnly strings.Builder
+	for _, r := range filename {
+		if r <= 127 {
+			asciiOnly.WriteRune(r)
+		}
+	}
+	asciiFilename := asciiOnly.String()
+
+	// If base name is empty after stripping non-ASCII, use "download" + extension
+	ext := filepath.Ext(asciiFilename)
```
jakubno: this won't work e.g. for `.env`

jakubno:
and if you try with code from here? https://github.com/e2b-dev/infra/pull/1845/changes/f23cd84876a2a335b0c57ab3b5222c7f6f22b168?

beran-t:
## Docker Integration Test Results

Tested against compiled envd running in Docker (linux/amd64).
Uses `mime.FormatMediaType("inline", ...)` from Go stdlib.

	w.Header().Set("Content-Disposition", mime.FormatMediaType("inline", map[string]string{"filename": filepath.Base(resolvedPath)}))

| Input | Content-Disposition |
|-------|---------------------|
| `file.txt` | `inline; filename=file.txt` |
| `my document.pdf` | `inline; filename="my document.pdf"` |
| `.env` | `inline; filename=.env` |
| `.gitignore` | `inline; filename=.gitignore` |
| `.env.local` | `inline; filename=.env.local` |
| `文档.pdf` | `inline; filename*=utf-8''%E6%96%87%E6%A1%A3.pdf` |
| `café.txt` | `inline; filename*=utf-8''caf%C3%A9.txt` |
| ` .env` | `inline; filename=" .env"` |

jakubno:
> ## Docker Integration Test Results
> Tested against compiled envd running in Docker (linux/amd64). Uses `mime.FormatMediaType("inline", ...)` from Go stdlib.
> 
> ```
> w.Header().Set("Content-Disposition", mime.FormatMediaType("inline", map[string]string{"filename": filepath.Base(resolvedPath)}))
> ```
> 
> Input	Content-Disposition
> `file.txt`	`inline; filename=file.txt`
> `my document.pdf`	`inline; filename="my document.pdf"`
> `.env`	`inline; filename=.env`
> `.gitignore`	`inline; filename=.gitignore`
> `.env.local`	`inline; filename=.env.local`
> `文档.pdf`	`inline; filename*=utf-8''%E6%96%87%E6%A1%A3.pdf`
> `café.txt`	`inline; filename*=utf-8''caf%C3%A9.txt`
> ` .env`	`inline; filename=" .env"`

can we use this?

PR #1846, opened 2026-02-04
Create an nfs proxy, portmap

djeebus:
- adds port mapper to orchestrator. sandboxes uses this via iptables redirects to locate the nfs "server" (proxy) that hosts the persistent volumes
- adds nfs proxy that chroots users into their mounted volume

---

> [!NOTE]
> **High Risk**
> Introduces network-facing NFS/portmap services and volume mount/exec behavior in `envd`, which can impact isolation and host filesystem access if mount-prefix or request routing is wrong.
> 
> **Overview**
> Adds a persistent-volume NFS proxy in the orchestrator, including a minimal portmap RPC service so sandboxes can discover and mount volumes by name and be *jailed* to their team/volume prefix, with additional logging and panic-recovery wrappers and new unit/e2e tests. The API layer is updated to pass config through cluster initialization and expose volume create/delete via an orchestrator instance, plus local dev/envd wiring to mount the provided NFS targets (and local seeding/.env updated for volume mounts).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9ba1e895ec55b3ae170f097fff51a58d221de3e0. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/envd/internal/api/store.go
+++ b/packages/envd/internal/api/store.go
@@ -86,8 +86,8 @@ func (a *API) GetMetrics(w http.ResponseWriter, r *http.Request) {
 
 func (a *API) getLogger(err error) *zerolog.Event {
 	if err != nil {
-		return a.logger.Error().Err(err) // nolint:zerologlint // this is only prep
+		return a.logger.Error().Err(err) //nolint:zerologlint // this is only prep
 	}
 
-	return a.logger.Info() // nolint:zerologlint // this is only prep
+	return a.logger.Info() //nolint:zerologlint // this is only prep
 }


packages/orchestrator/internal/portmap/main.go
```diff
@@ -0,0 +1,106 @@
...
+		Vers: mapping.Vers,
+		Prot: mapping.Prot,
+	}]
+	if !ok {
+		logger.L().Warn(h.ctx, "[portmap handler] port not found")
+
+		return 0
+	}
+
+	logger.L().Debug(h.ctx, "[portmap handler] port found",
```
levb, line 70: 2/5 log should be outside the locks, 0/5 defer before Lock?

packages/orchestrator/internal/portmap/main.go
```diff
@@ -0,0 +1,106 @@
...
+	portmap "github.com/zeldovich/go-rpcgen/rfc1057"
+	"go.uber.org/zap"
+
+	"github.com/e2b-dev/infra/packages/shared/pkg/logger"
+)
+
+type handlers struct {
+	ctx  context.Context //nolint:containedctx // can't change the API, still need it
+	maps map[key]portmap.Uint32
+	lock sync.RWMutex
```
levb, line 16: Would sync.Map perform substantially worse here? 0/5 It _may look_ cleaner.

packages/orchestrator/internal/portmap/server.go
```diff
@@ -0,0 +1,99 @@
...
+	s *rfc1057.Server
+}
+
+func NewPortMap(ctx context.Context) *Server {
+	s := rfc1057.MakeServer()
+	h := newHandlers()
+
+	var handler rfc1057.PMAP_PROG_PMAP_VERS_handler
+	handler = h
+	handler = newRecovery(ctx, handler)
```
levb: nit, 0/5 naming: s/new/wrapWith - more intention revealing?

packages/orchestrator/internal/nfs/main.go
```diff
@@ -0,0 +1,136 @@
+package nfs
```
levb: 0/5 total nit, naming. Maybe it's just me, but main.go is for `main`. If a person needs to identify the main file in the package, it should be package_name.go; or is there another "main" connotation here that I am missing?

packages/orchestrator/internal/nfsproxy/proxy.go
```diff
@@ -0,0 +1,136 @@
...
+func getPrefixFromSandbox(sandboxes *sandbox.Map, config cfg.Config) jailed.GetPrefix {
+	filesystemsByType := buildFilesystems(config.PersistentVolumeMounts)
+
+	return func(_ context.Context, conn net.Conn, request nfs.MountRequest) (billy.Filesystem, string, error) {
+		sbx, err := sandboxes.GetByHostPort(conn.RemoteAddr().String())
+		if err != nil {
+			return nil, "", err
+		}
+
+		dirPath := string(request.Dirpath)
```
levb: 0/5, nit: may be more readable inlined, with comments?

packages/orchestrator/internal/nfs/main.go
levb: Also, package name would read better as `nfsproxy` when used, more intention revealing, see comment in orchestrator/main

packages/orchestrator/main.go
```diff
@@ -449,6 +452,39 @@ func run(config cfg.Config) (success bool) {
 		},
 	})
 
+	// nfs proxy server
```
levb, line 470: 0/5, nit, ha, and here we could've used some intention-revealing functions, since so many details... 
I'd argue for something like
```
if len(config.PersistentVolumeMounts) > 0 {
  err, closer(s) := nfsproxy.Start(...
  append closers
}
```

packages/orchestrator/internal/portmap/main.go
djeebus: It'd be cleaner (avoids explicit locking) but also dirtier (type assertions when pulling things out). Maybe balances out. If you feel strongly, happy to swap!

levb: 0/5, not strong.

packages/envd/internal/api/init.go
```diff
@@ -217,9 +218,30 @@ func (a *API) SetData(ctx context.Context, logger zerolog.Logger, data PostInitJ
 		a.defaults.Workdir = data.DefaultWorkdir
 	}
 
+	if data.VolumeMounts != nil {
+		for _, volume := range *data.VolumeMounts {
+			logger.Debug().Msgf("Mounting %s at %q", volume.NfsTarget, volume.Path)
+
+			go a.setupNfs(context.WithoutCancel(ctx), volume.NfsTarget, volume.Path)
```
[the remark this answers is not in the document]
djeebus: There's a chicken-and-egg problem here:
1. orchestrator creates sandbox
2. orchestrator calls envd at `GET /init` and waits for the result
2a. envd mounts nfs shares and waits for them to complete
2b. orchestrator looks up the nfs client by ip address in the sandboxes map
3. orchestrator adds sandbox to the map

The problem here is that 2b won't succeed until 3 is done. Right now we just run it twice. A better solution would be to use an auth token instead of checking ip address, but I wanted to get the v0 done first. We can improve this later.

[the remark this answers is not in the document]
levb: This also seem a documentable (if unfortunate) behavior

packages/orchestrator/internal/nfsproxy/proxy.go
```diff
@@ -0,0 +1,134 @@
...
+	for name, path := range config.PersistentVolumeMounts {
+		filesystemsByType[name] = osfs.New(path)
+	}
+
+	var handler nfs.Handler
+	handler = jailed.NewNFSHandler(
+		getPrefixFromSandbox(sandboxes, filesystemsByType),
+		getChangeFromFilesystem,
+	)
+	handler = helpers.NewCachingHandler(handler, cacheLimit)
```
levb, line 112: nit: 0/5 `wrapWithCaching`,..

packages/orchestrator/internal/sandbox/envd.go
```diff
@@ -105,22 +115,9 @@ func (s *Sandbox) initEnvd(ctx context.Context) (e error) {
...
-		s.Config.Envd.AccessToken,
-		s.internalConfig.EnvdInitRequestTimeout,
-		s.Config.Envd.Vars,
-		s.Runtime.SandboxID,
-		s.Config.Envd.Version,
-		hyperloopIP,
-		s.Config.Envd.DefaultUser,
-		s.Config.Envd.DefaultWorkdir,
-	)
+	response, count, err := s.doRequestWithInfiniteRetries(ctx, http.MethodPost, address)
```
levb, line 120: :+1:

packages/orchestrator/.env.local
```diff
@@ -2,7 +2,7 @@ ARTIFACTS_REGISTRY_PROVIDER=Local
 DEFAULT_PERSISTENT_VOLUME_TYPE=test-volume-type
 ENVIRONMENT=local
 FIRECRACKER_VERSIONS_DIR=../fc-versions/builds
-HOST_ENVD_PATH=../envd/bin/envd
+HOST_ENVD_PATH=../envd/bin/debug/envd
```
levb: is this intentional in the PR?

packages/envd/internal/api/init.go
```diff
@@ -213,9 +214,30 @@ func (a *API) SetData(ctx context.Context, logger zerolog.Logger, data PostInitJ
...
+
 	return nil
 }
 
+func (a *API) setupNfs(ctx context.Context, nfsTarget, path string) {
+	data, err := exec.CommandContext(ctx, "bash", "-c", fmt.Sprintf(`
+set -e
+mkdir -p %q
+mount -v -t nfs -o mountproto=tcp,mountport=2049,proto=tcp,port=2049,nfsvers=3,noacl %q %q
+`, path, nfsTarget, path)).CombinedOutput()
```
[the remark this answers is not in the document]
djeebus: This is true, will be fixed in a future release.

packages/orchestrator/internal/nfsproxy/proxy.go
djeebus: we don't own the name of this function =/

packages/envd/internal/api/init.go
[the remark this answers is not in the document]
djeebus: I'm going to do my best to fix it in a follow-up PR, but if I can't, I'll document it :+1:

PR #1849, opened 2026-02-04
Configure FUSE for non-root user access

mishushakov:
Configure FUSE to allow non-root users to access FUSE devices in templates.

Changes:
- Enable `user_allow_other` in `/etc/fuse.conf` during provisioning
- Set `/dev/fuse` permissions to 666 for non-root access
- Add integration test to verify FUSE configuration

---

> [!NOTE]
> **Medium Risk**
> Changes low-level sandbox OS provisioning and device permissions, which could impact isolation/security or break FUSE mounts if misconfigured.
> 
> **Overview**
> Updates base template provisioning to enable `user_allow_other` in `/etc/fuse.conf` and persist permissive `/dev/fuse` (0666) device permissions via `systemd-tmpfiles`, and adds integration tests that validate the config and that non-root users can read/write the FUSE device inside a sandbox.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 08f334bc75467cc6305e9a2c91dc2718a48bb97b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/orchestrator/internal/template/build/phases/base/provision.sh
```diff
@@ -34,6 +34,24 @@ else
...
+    elif ! grep -q '^user_allow_other' /etc/fuse.conf; then
+        echo "user_allow_other" >> /etc/fuse.conf
+    fi
+else
+    echo "user_allow_other" > /etc/fuse.conf
+fi
+
+# Set /dev/fuse permissions to 666 for non-root access
+if [ -e /dev/fuse ]; then
+    chmod 666 /dev/fuse
```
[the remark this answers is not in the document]
mishushakov: I think this is fine because we don't do reboot

jakubno (changes requested):
Could you share more context? Not sure if we should be setting different values then default.
- they probably won't be the recommended values
- it can be breaking for people already using fuse

packages/orchestrator/pkg/template/build/phases/base/provision.sh
```diff
@@ -38,6 +38,24 @@ else
...
+        echo "user_allow_other" >> /etc/fuse.conf
+    fi
+else
+    echo "user_allow_other" > /etc/fuse.conf
+fi
+
+# Set /dev/fuse permissions to 666 for non-root access
+# Use systemd-tmpfiles to set permissions at boot
+mkdir -p /etc/tmpfiles.d
+echo 'z /dev/fuse 0666 root root -' > /etc/tmpfiles.d/fuse.conf
```
arkamar, line 44: This is reasonable change in my opinion. I support it. It is common permission setup for fuse in most of the systems I have seen so far. Kernel creates most of the devices with 0600 but they are later changed with respect to udev rules in /dev. We don't use udev in sandboxes and tmpfiles should work reasonably well here.

packages/orchestrator/pkg/template/build/phases/base/provision.sh
```diff
@@ -38,6 +38,24 @@ else
...
+if [ -f /etc/fuse.conf ]; then
+    # Uncomment user_allow_other if it exists, or add it if it doesn't
+    if grep -q '^#user_allow_other' /etc/fuse.conf; then
+        sed -i 's/^#user_allow_other/user_allow_other/' /etc/fuse.conf
+    elif ! grep -q '^user_allow_other' /etc/fuse.conf; then
+        echo "user_allow_other" >> /etc/fuse.conf
+    fi
+else
+    echo "user_allow_other" > /etc/fuse.conf
+fi
```
arkamar: However, I am not sure if we really want to enable this? What are the other users in the sandbox they need to access the mounted fs? We have `user` only by default. I think it would make sense to drop this part, document it and let users decide if they want that as they can change this in template build.

jakubno (approved):

PR #1856, opened 2026-02-05
fix(envd): ensure proper closure of multipart parts on error in PostFiles h…

uucloud:
…andler

### DIFF
--- a/packages/envd/internal/api/upload.go
+++ b/packages/envd/internal/api/upload.go
@@ -148,6 +148,36 @@ func resolvePath(part *multipart.Part, paths *UploadSuccess, u *user.User, defau
 	return filePath, nil
 }
 
+func (a *API) handlePart(r *http.Request, part *multipart.Part, paths UploadSuccess, u *user.User, uid, gid int, operationID string, params PostFilesParams) (*EntryInfo, int, error) {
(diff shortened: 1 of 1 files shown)


uucloud:
> Could you extract the inner loop logic in a separate function, where we could just call `defer`?

Done.

packages/envd/internal/api/upload.go
```diff
@@ -148,6 +148,38 @@ func resolvePath(part *multipart.Part, paths *UploadSuccess, u *user.User, defau
...
+		Str(string(logs.OperationIDKey), operationID).
+		Str("event_type", "file_processing").
+		Logger()
+
+	status, err := processFile(r, filePath, part, uid, gid, logger)
+	if err != nil {
+		return status, err
+	}
+
+	*paths = append(*paths, EntryInfo{
```
jakubno: Last request 🙏🏻  I'm not a fan of mutating slices inside the functions, can we make it pure by returning `EntryInfo` and if there isn't an error, append it to the slice? 🙏🏻

uucloud: Done, handlePart now returns EntryInfo.

jakubno (approved):

PR #1857, opened 2026-02-05
docs: fix inaccuracies in self-host.md guide

jakubno:
Inspired by #1848
Made the whole flow little bit easier

---

> [!NOTE]
> **Medium Risk**
> Touches Terraform provisioning for Secret Manager resources and state moves, which can break applies or recreate secrets if misconfigured; operational impact is limited to deployment/setup workflows.
> 
> **Overview**
> Makes self-hosting setup more turnkey by reorganizing `.env.template` into required/optional blocks with safer defaults (including larger default boot disks), generalizing `Makefile` Terraform targets to respect `PROVIDER`, and updating `self-host.md` instructions and SDK snippets. In GCP Terraform, Secret Manager resources are moved from `api.tf`/`init/main.tf` into a dedicated `init/secrets.tf`, `main.tf` now consumes those secrets via `module.init` outputs, init adds a longer Secrets API wait, and `moved` blocks are introduced to preserve state for the relocated secrets.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 36d65f9980b44a11e93c3b2a02e32a84771ed761. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -1,5 +1,6 @@
-PROVIDER=gcp # aws or gcp
-PREFIX="e2b-" # prefix identifier for all resources
+# ------------------------------------------- Required block -----------------------------------------------------------
+# These variables are required and don't have default values
+# You must provide values for these to run the deployment successfully
 
 # your GCP project ID
 GCP_PROJECT_ID=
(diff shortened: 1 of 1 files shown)


.env.template
```diff
@@ -9,9 +9,7 @@ GCP_REGION=
 GCP_ZONE=
 
 # prod, staging, dev
-TERRAFORM_ENVIRONMENT=
-# GCS bucket name
-TERRAFORM_STATE_BUCKET=
+TERRAFORM_ENVIRONMENT=dev
```
arkamar, line 26: We newly set most of the variables to a default value in this part of the template, however, some are still left empty and it is mandatory to set them to meaningful values. I believe that the template would benefit from foreword where it will be explicitly stated that all unset variables in this part must to be set. It would be better for user experience, rather than guessing it from error messages.

self-host.md
```diff
@@ -87,7 +87,7 @@ When using SDK pass domain when creating new `Sandbox` in JS/TS SDK
 ```js
 import { Sandbox } from "@e2b/sdk";
```
arkamar:
```suggestion
import { Sandbox } from "e2b";
```
This seems to be outdated.

arkamar (approved):
LGTM

PR #1860, opened 2026-02-05
chore: make build status interchangeable

dobrac:
Make the status "uploaded" and "success" interchangeable. Same applies for "waiting" and "snapshotting". 

Introduce also a third united state like "in_progress", "ready".

Adds the source type to envs to allow filtering by "template" or "snapshot".

---

> [!NOTE]
> **High Risk**
> Includes non-transactional DB migrations with triggers/backfills and changes status semantics across core template/snapshot flows, so mistakes can affect build orchestration, concurrency limiting, and which envs are surfaced as templates vs snapshots.
> 
> **Overview**
> Normalizes build status handling by introducing `env_builds.status_group` (computed from existing `status` via trigger + backfill) and updating API/cache/query logic to compare/read grouped statuses (`pending`/`in_progress`/`ready`/`failed`) so legacy raw states like `waiting`/`snapshotting` and `uploaded`/`success` become interchangeable in read paths.
> 
> Adds `envs.source` (`template` default, `snapshot` for pause/resume) with a trigger/backfill and updates template-related queries/endpoints to explicitly filter to `source='template'` while snapshot creation inserts `source='snapshot'`; sqlc models/overrides and tests are updated accordingly, and template-manager now maps status groups back to a concrete `status` when persisting.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 26b1075f253f1f5b415fd3ff36c21eb908845cc3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/db/migrations/20260210120002_add_status_group_column.sql
+++ b/packages/db/migrations/20260210120002_add_status_group_column.sql
@@ -21,7 +21,7 @@ CREATE TRIGGER trg_compute_status_group
   FOR EACH ROW EXECUTE FUNCTION compute_status_group();
 
 -- +goose StatementBegin
-DO $$
+CREATE OR REPLACE PROCEDURE backfill_status_group() AS $$
 DECLARE
   batch_size INT := 50000;
   rows_updated INT;
@@ -40,11 +40,15 @@ BEGIN
       LIMIT batch_size
     );
     GET DIAGNOSTICS rows_updated = ROW_COUNT;
+    COMMIT;
     EXIT WHEN rows_updated = 0;
-    PERFORM pg_sleep(0.1);
   END LOOP;
-END $$;
+END;
+$$ LANGUAGE plpgsql;
 -- +goose StatementEnd
+
+CALL backfill_status_group();
+DROP PROCEDURE backfill_status_group();
 
 ALTER TABLE public.env_builds
   ADD CONSTRAINT chk_status_group_not_null


packages/db/migrations/20260210120001_add_env_and_build_source_columns.sql
```diff
@@ -0,0 +1,42 @@
...
+-- Add source column to envs table to track where the env came from
+-- 'template' = built from Dockerfile (default)
+-- 'snapshot' = created from pause/resume
+ALTER TABLE "public"."envs"
+    ADD COLUMN IF NOT EXISTS "source" text NOT NULL DEFAULT 'template';
+
+-- Create trigger first so any concurrent snapshot inserts are handled
+CREATE OR REPLACE FUNCTION sync_env_source_on_snapshot_insert() RETURNS TRIGGER AS $$
+BEGIN
+    UPDATE "public"."envs" SET source = 'snapshot' WHERE id = NEW.env_id;
```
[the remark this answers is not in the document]
dobrac: The trigger is intentional as a safety net — it catches any snapshot inserts that happen outside of the UpsertSnapshot query (e.g., manual inserts, other code paths, future queries). The redundancy is by design.

packages/db/queries/templates/create_template.sql
```diff
@@ -1,6 +1,6 @@
 -- name: CreateOrUpdateTemplate :exec
-INSERT INTO "public"."envs"(id, team_id, created_by, updated_at, public, cluster_id)
-VALUES (@template_id, @team_id, @created_by, NOW(),FALSE, @cluster_id)
+INSERT INTO "public"."envs"(id, team_id, created_by, updated_at, public, cluster_id, source)
+VALUES (@template_id, @team_id, @created_by, NOW(), FALSE, @cluster_id, 'template')
 ON CONFLICT (id) DO UPDATE
 SET updated_at  = NOW(),
     build_count = envs.build_count + 1;
```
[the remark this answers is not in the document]
dobrac: This can't happen — you can't build a template if the env already exists as a snapshot. The ON CONFLICT path is only hit when re-building an existing template (same ID), in which case source is already 'template'.

packages/db/queries/builds/get_template_build.sql
```diff
@@ -3,4 +3,5 @@ SELECT sqlc.embed(e), sqlc.embed(eb)
 FROM "public"."envs" e
 JOIN "public"."env_build_assignments" eba ON eba.env_id = e.id
 JOIN "public"."env_builds" eb ON eb.id = eba.build_id
-WHERE e.id = sqlc.arg(template_id) AND eb.id = sqlc.arg(build_id);
+WHERE e.id = sqlc.arg(template_id) AND eb.id = sqlc.arg(build_id)
+  AND e.source = 'template';
```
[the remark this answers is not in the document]
dobrac: This is the intended behavior — the filter ensures we only operate on template builds. Returning no results for a snapshot env_id is correct; it should never be called with one.

packages/api/internal/template/register_build.go
```diff
@@ -195,8 +195,10 @@ func RegisterBuild(
 	telemetry.ReportEvent(ctx, "marked previous builds as failed")
 
 	// Insert the new build
+	// TODO: Switch to dbtypes.BuildStatusPending once all consumers are migrated to use Is*() helpers.
```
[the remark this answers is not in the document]
dobrac: Tracked in [ENG-3469].

packages/db/queries/builds/finish_template_build.sql
```diff
@@ -3,7 +3,7 @@ UPDATE "public"."env_builds"
 SET
     finished_at = NOW(),
     total_disk_size_mb = @total_disk_size_mb,
-    status = 'uploaded',
+    status = @status,
```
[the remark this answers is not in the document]
dobrac: Not a SQL injection risk — sqlc generates parameterized queries ($1, $2), not string interpolation. The only caller is SetFinished which passes a Go constant (types.BuildStatusUploaded). The status value never comes from external input.

packages/db/queries/builds/get_concurrent_template_builds.sql
```diff
@@ -3,5 +3,5 @@ SELECT eb.* FROM env_build_assignments eba
 JOIN env_builds eb ON eb.id = eba.build_id
 WHERE
     eba.env_id = @template_id
-    AND eb.status in ('waiting', 'building')
+    AND eb.status IN ('waiting', 'building', 'snapshotting', 'pending', 'in_progress')
```
djeebus: It seems that things like this might be better put into a [generated column](https://www.postgresql.org/docs/current/ddl-generated-columns.html)?

packages/db/pkg/types/types.go
```diff
@@ -34,15 +34,44 @@ type PausedSandboxConfig struct {
...
+
+// IsReady returns true if the build completed successfully.
+func (s BuildStatus) IsReady() bool {
+	return s == BuildStatusReady || s == BuildStatusUploaded || s == BuildStatusSuccess
+}
+
+// IsFailed returns true if the build failed.
+func (s BuildStatus) IsFailed() bool {
+	return s == BuildStatusFailed
+}
```
djeebus: These feel like they should be stored as computed columns in the database, and sent out. Otherwise we're defining these in at least a few different places.

djeebus (approved):

packages/db/pkg/types/types.go
dobrac: Thats a good idea! Reworked

packages/db/queries/builds/get_concurrent_template_builds.sql
dobrac: (answered in the other comment)

packages/db/migrations/20260210120002_add_status_group_column.sql
```diff
@@ -0,0 +1,69 @@
...
+  VALIDATE CONSTRAINT chk_status_group_not_null;
+
+ALTER TABLE public.env_builds
+  ALTER COLUMN status_group SET NOT NULL;
+
+ALTER TABLE public.env_builds
+  DROP CONSTRAINT chk_status_group_not_null;
+
+CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_env_builds_status_group
+  ON public.env_builds(status_group);
```
djeebus, line 67: do these need to be wrapped in `StatementBegin` and `StatementEnd` as well?

dobrac: only if I want to run them as one statement, otherwise its okay

packages/db/migrations/20260210120002_add_status_group_column.sql
```diff
@@ -0,0 +1,69 @@
...
+  CHECK (status_group IS NOT NULL) NOT VALID;
+
+ALTER TABLE public.env_builds
+  VALIDATE CONSTRAINT chk_status_group_not_null;
+
+ALTER TABLE public.env_builds
+  ALTER COLUMN status_group SET NOT NULL;
+
+ALTER TABLE public.env_builds
+  DROP CONSTRAINT chk_status_group_not_null;
```
djeebus, line 64: Is it worth the effort to `ADD CONSTRAINT chk_status_group_in_list CHECK (status_group IN ('pending', 'in_progress', 'ready'))`?

dobrac: I was checking with Claude, the reason for is just to make the change to not null fast and not lock the table for a long time. I don't think it hurts

packages/api/internal/template-manager/template_status.go
```diff
@@ -271,33 +271,52 @@ func (tm *TemplateManager) SetStatus(ctx context.Context, buildID uuid.UUID, sta
...
+	case types.BuildStatusBuilding, types.BuildStatusSnapshotting:
+		return types.BuildStatusGroupInProgress
+	case types.BuildStatusUploaded, types.BuildStatusSuccess:
+		return types.BuildStatusGroupReady
+	case types.BuildStatusFailed:
+		return types.BuildStatusGroupFailed
+	default:
+		return types.BuildStatusGroupFailed
+	}
+}
```
djeebus: I bet you could remove this and pass both `status` and `statusGroup` into `SetStatus`

djeebus (approved):

packages/api/internal/template-manager/template_status.go
dobrac: changed a bit to passing the status group and then mapping it back for the database (for now)

PR #1862, opened 2026-02-06
Feature: Sandbox AutoResume

matthewlouisbrockman:
## Summary
  - Adds autoResume as a structured config end-to-end:
      - API/OpenAPI: autoResume: { policy: "any" | "off" }
      - Orchestrator gRPC: SandboxConfig.auto_resume is now SandboxAutoResumeConfig (with policy).
  - Client-proxy can auto-resume a paused sandbox on catalog miss by calling the API gRPC resume endpoint, gated
    by a feature flag.
      - Flag key: sandbox-auto-resume
      - Evaluated in sandbox context using the sandbox ID.

  ### Notes

  - Resume requests use a fixed 5 minute timeout (timeoutSeconds=300) in the proxy-to-API gRPC call.
  - Auto-resume behavior is also gated by sandbox config: API only resumes if the snapshot config has
    autoResume.policy == "any".

  ## Test Plan

  ### create sandbox with auto-resume enabled
  curl -sS -X POST "$API_BASE/sandboxes" \
    -H "X-API-Key: $API_KEY" \
    -H "Content-Type: application/json" \
    -d '{
      "templateID":"'"$TEMPLATE_ID"'",
      "timeout":30,
      "autoPause":true,
      "autoResume":{"policy":"any"}
    }'

  ### pause, then hit through proxy to trigger resume
  curl -sS -X POST "$API_BASE/sandboxes/$SANDBOX_ID/pause" -H "X-API-Key: $API_KEY"
  curl -sS -H "Host: 3210-$SANDBOX_ID.$PROXY_DOMAIN" "http://$PROXY_ADDR/"

  ## LaunchDarkly / Flag Note

  - sandbox-auto-resume is default false (so prod behavior is unchanged unless enabled in LaunchDarkly).


packages/api/internal/handlers/proxy_grpc.go
```diff
@@ -0,0 +1,269 @@
...
+		}
+
+		return nil, status.Errorf(codes.Internal, "failed to get snapshot: %v", err)
+	}
+
+	teamID := snap.Snapshot.TeamID
+
+	// Prototype: always resume for a short fixed TTL to validate the request flow.
+	// We'll revisit per-team limits/policies later.
+	timeout := 5 * time.Minute
```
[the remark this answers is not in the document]
matthewlouisbrockman: on auto-resume, we don't have a great way to say what the timeout should be; implemented a complex storage / updating process in original PR but the concept's not quite well defined enough to make that part of the initial implementation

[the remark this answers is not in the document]
matthewlouisbrockman: fixed wording

packages/api/internal/handlers/proxy_grpc.go
```diff
@@ -0,0 +1,269 @@
...
+		}
+
+		return nil, status.Errorf(codes.Internal, "failed to get snapshot: %v", err)
+	}
+
+	teamID := snap.Snapshot.TeamID
+
+	// Fot initial version, we resume for 5 minutes.
+	// We don't have a well definied flow for the lifecycle here yet so will iterate on this later.
+	timeout := 5 * time.Minute
```
jakubno: can we add it as a task to linear and add here a reference?

packages/api/internal/handlers/proxy_grpc.go
```diff
@@ -0,0 +1,269 @@
...
+
+	var autoResume *dbtypes.SandboxAutoResumePolicy
+	if snap.Snapshot.Config != nil {
+		autoResume = snap.Snapshot.Config.AutoResume
+	}
+	if !dbtypes.IsAutoResumeAny(autoResume) {
+		return nil, status.Error(codes.NotFound, "sandbox auto-resume disabled")
+	}
+
+	lock, lockAcquired, lockErr := s.tryAcquireResumeLock(ctx, sandboxID)
```
jakubno: you don't need to do this. It's handled downstream

packages/api/internal/handlers/proxy_grpc.go
```diff
@@ -0,0 +1,269 @@
...
+	}
+	if lock != nil {
+		defer func() {
+			if releaseErr := lock.Release(context.WithoutCancel(ctx)); releaseErr != nil {
+				logger.L().Warn(ctx, "Failed to release proxy resume lock", zap.Error(releaseErr), logger.WithSandboxID(sandboxID))
+			}
+		}()
+	}
+
+	running, runErr := s.api.orchestrator.GetSandbox(ctx, teamID, sandboxID)
```
jakubno: you already have this info right?

packages/api/internal/handlers/proxy_grpc.go
```diff
@@ -0,0 +1,269 @@
...
+	if snap.Snapshot.Config != nil {
+		network = snap.Snapshot.Config.Network
+	}
+
+	team, err := dbapi.GetTeamByID(ctx, s.api.authDB, teamID)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to get team: %v", err)
+	}
+
+	headers := http.Header{}
```
jakubno, line 88: if you move team up, I believe the same is done on multiple places ,maybe we could create a function and reuse it

packages/api/internal/handlers/proxy_grpc.go
```diff
@@ -0,0 +1,269 @@
...
+		snap.Snapshot.AllowInternetAccess,
+		network,
+		nil, // mcp
+	)
+	if apiErr != nil {
+		return nil, status.Errorf(grpcCodeFromHTTPStatus(apiErr.Code), "resume failed: %s", apiErr.ClientMsg)
+	}
+
+	// Return routing info so the proxy can send the request immediately (no catalog polling required).
+	running, runErr = s.api.orchestrator.GetSandbox(ctx, teamID, sandboxID)
```
jakubno: why? `startSandbox` is returning sandbox data

packages/db/pkg/types/types.go
```diff
@@ -29,9 +29,21 @@ type SandboxNetworkConfig struct {
 	Ingress *SandboxNetworkIngressConfig `json:"ingress,omitempty"`
 }
 
+type SandboxAutoResumePolicy string
```
jakubno, line 32: can this be a struct, there will be e.g. the timeout later and maybe more

packages/client-proxy/main.go
```diff
@@ -167,7 +182,11 @@ func run() int {
 	}
 
 	var closers []Closeable
-	closers = append(closers, featureFlagsClient, catalog)
+	closers = append(closers, catalog)
+	closers = append(closers, featureFlags)
```
jakubno: you could keep this

packages/client-proxy/internal/proxy/paused_sandbox_resumer_grpc.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func NewGrpcPausedSandboxResumer(address string) (PausedSandboxResumer, error) {
+	// Client-proxy uses this gRPC client to trigger ResumeSandbox when needed.
+	if strings.TrimSpace(address) == "" {
+		return nil, fmt.Errorf("api grpc address is required")
+	}
+
+	// Note: use Dial (not grpc.NewClient) to stay compatible with the grpc-go version used in this repo.
+	conn, err := grpc.Dial(address, grpc.WithTransportCredentials(insecure.NewCredentials()))
```
jakubno: We don't use `grpc.Dial` anywhere, also it's deprecated

packages/client-proxy/internal/proxy/proxy.go
```diff
@@ -45,7 +57,46 @@ func catalogResolution(ctx context.Context, sandboxId string, c catalog.Sandboxe
...
+	ctx context.Context,
+	sandboxId string,
+	pausedChecker PausedSandboxResumer,
+	featureFlags *featureflags.Client,
+) (string, error) {
+	if pausedChecker == nil {
+		return "", nil
+	}
+
+	if featureFlags != nil && !featureFlags.BoolFlag(ctx, featureflags.SandboxAutoResumeFlag, featureflags.SandboxContext(sandboxId)) {
```
jakubno: we don't have to be such defensive, `featureFlags` is always defined

packages/client-proxy/internal/proxy/proxy.go
```diff
@@ -45,7 +57,46 @@ func catalogResolution(ctx context.Context, sandboxId string, c catalog.Sandboxe
...
+	if err == nil {
+		return nodeIP, nil
+	}
+
+	if isNotPausedError(err) {
+		// API says it can't resume (no snapshot / not resumable).
+		return "", nil
+	}
+
+	return "", err
```
jakubno: feel free to ignore, but we usually would write something like that - handling the error case first 
```suggestion
	if err != nil {
		if isNotPausedError(err) {
			// API says it can't resume (no snapshot / not resumable).
			return "", nil
		}
		return "", err
	}

	return nodeIP, nil
```

packages/orchestrator/orchestrator.proto
```diff
@@ -46,6 +46,9 @@ message SandboxConfig {
   optional bool allow_internet_access = 21;
 
   optional SandboxNetworkConfig network = 22;
+
+  // Auto-resume policy for paused sandboxes.
+  optional string auto_resume = 23;
```
jakubno: sometging like
```suggestion
  optional SandboxAutoResumeConfig auto_resume = 23;
```

packages/shared/pkg/feature-flags/flags.go
```diff
@@ -90,6 +90,7 @@ var (
 	BestOfKTooManyStartingFlag          = newBoolFlag("best-of-k-too-many-starting", false)
 	EdgeProvidedSandboxMetricsFlag      = newBoolFlag("edge-provided-sandbox-metrics", false)
 	CreateStorageCacheSpansFlag         = newBoolFlag("create-storage-cache-spans", env.IsDevelopment())
+	SandboxAutoResumeFlag               = newBoolFlag("sandbox-auto-resume", true)
```
jakubno:
```suggestion
	SandboxAutoResumeFlag               = newBoolFlag("sandbox-auto-resume", false)
```

packages/api/internal/handlers/proxy_grpc.go
matthewlouisbrockman: https://linear.app/e2b/issue/ENG-3473/codify-what-timeout-should-be

packages/api/internal/handlers/proxy_grpc.go
matthewlouisbrockman: once remove lock yep

packages/api/internal/handlers/proxy_grpc.go
matthewlouisbrockman: Not sure I follow this one

packages/db/pkg/types/types.go
matthewlouisbrockman: moved to struct everywhere

packages/api/internal/handlers/proxy_grpc.go
matthewlouisbrockman: don't think it includes the routing info but can figure out something there

jakubno: We can probably return `sandbox.Sandbox` and convert to API type in the handler

matthewlouisbrockman: added a wrapper around startSandbox

packages/api/internal/handlers/proxy_grpc.go
matthewlouisbrockman: moving it up but the getTeamByID i think only used by me, had to make that function (it required user + team before, but the snapshot just has the team)

packages/db/pkg/types/types.go
matthewlouisbrockman: policy still a string but the config's a struct

tests/integration/internal/utils/sandbox.go
```diff
@@ -117,6 +125,13 @@ func SetupSandboxWithCleanup(t *testing.T, c *api.ClientWithResponses, options .
 			continue
 		}
 
+		if createSandboxResponse.StatusCode() != http.StatusCreated {
+			t.Logf("Sandbox creation failed status=%d body=%s", createSandboxResponse.StatusCode(), string(createSandboxResponse.Body))
+			if createSandboxResponse.JSON400 != nil {
+				t.Logf("Sandbox creation JSON400=%+v", *createSandboxResponse.JSON400)
+			}
```
jakubno: can we do this to generalize?
```suggestion
			t.Logf("Sandbox creation=%+v", *createSandboxResponse)
```

spec/openapi.yml
```diff
@@ -255,6 +255,21 @@ components:
...
+      default: "off"
+      enum:
+        - any
+        - off
+
+    SandboxAutoResumeConfig:
+      type: object
+      description: Auto-resume configuration for paused sandboxes. Default is off.
+      properties:
+        policy:
```
jakubno, line 278: I would make it probably required, the whole object should remain optional though

packages/shared/pkg/grpc/proxy/proxy.proto
```diff
@@ -0,0 +1,24 @@
...
+package proxy;
+
+option go_package = "github.com/e2b-dev/infra/packages/shared/pkg/grpc/proxy;proxygrpc";
+
+// Request to resume a paused sandbox.
+message SandboxResumeRequest {
+  // Sandbox ID to resume.
+  string sandbox_id = 1;
+  // Optional timeout override in seconds (must be within team limits).
+  int32 timeout_seconds = 2;
```
jakubno: I think client proxy won't ever know this, will it?

packages/orchestrator/orchestrator.proto
```diff
@@ -46,6 +46,15 @@ message SandboxConfig {
...
   optional SandboxNetworkConfig network = 22;
+
+  // Auto-resume policy for paused sandboxes.
+  optional SandboxAutoResumeConfig auto_resume = 23;
+}
+
+message SandboxAutoResumeConfig {
+  // Policy values are owned by the API layer today (e.g. "off", "any").
+  // Empty / unset means "off".
+  optional string policy = 1;
```
jakubno: think should probably be required again if the whole object is optional

packages/api/main.go
```diff
@@ -361,6 +365,20 @@ func run() int {
 	apiStore := handlers.NewAPIStore(ctx, tel, config)
 	cleanupFns = append(cleanupFns, apiStore.Close)
 
+	grpcAddr := fmt.Sprintf("0.0.0.0:%d", config.APIGrpcPort)
+	grpcListener, err := (&net.ListenConfig{}).Listen(ctx, "tcp", grpcAddr)
+	if err != nil {
+		l.Fatal(ctx, "failed to create proxy grpc listener", zap.Error(err))
+	}
+
+	grpcServer := grpc.NewServer(
```
jakubno: can. you check the grpc server in orchestrator and make it similar? Mainly keep alive and recovery

packages/api/main.go
```diff
@@ -407,6 +425,17 @@ func run() int {
 		}
 	})
 
+	wg.Go(func() {
+		defer cancel()
+
+		l.Info(ctx, "gRPC service starting", zap.Uint16("port", config.APIGrpcPort))
+		err := grpcServer.Serve(grpcListener)
+		if err != nil && !errors.Is(err, net.ErrClosed) && !errors.Is(err, grpc.ErrServerStopped) {
```
jakubno: // Serve will return a non-nil error unless Stop or GracefulStop is called.

I believe there shouldn't be any error then

packages/api/main.go
```diff
@@ -433,6 +462,12 @@ func run() int {
 			exitCode.Add(1)
 			l.Error(ctx, "Http service shutdown error", zap.Int("port", port), zap.Error(err))
 		}
+
+		grpcServer.GracefulStop()
+		if err := grpcListener.Close(); err != nil && !errors.Is(err, net.ErrClosed) {
```
jakubno: It shouldn't return `net.ErrClosed`

packages/api/internal/handlers/proxy_grpc.go
```diff
@@ -0,0 +1,188 @@
...
+
+	var autoResume *dbtypes.SandboxAutoResumeConfig
+	if snap.Snapshot.Config != nil {
+		autoResume = snap.Snapshot.Config.AutoResume
+	}
+	if autoResume == nil || autoResume.Policy == nil || *autoResume.Policy != dbtypes.SandboxAutoResumeAny {
+		return nil, status.Error(codes.NotFound, "sandbox auto-resume disabled")
+	}
+
+	running, runErr := s.api.orchestrator.GetSandbox(ctx, teamID, sandboxID)
```
jakubno: why do you get the sandbox here again?

packages/api/internal/handlers/proxy_grpc.go
```diff
@@ -0,0 +1,188 @@
...
+
+	node := s.api.orchestrator.GetNode(sbx.ClusterID, sbx.NodeID)
+	if node == nil || node.IPAddress == "" {
+		return nil, status.Error(codes.Internal, "sandbox resumed but routing info is not available yet")
+	}
+
+	return &proxygrpc.SandboxResumeResponse{OrchestratorIp: node.IPAddress}, nil
+}
+
+func grpcCodeFromHTTPStatus(statusCode int) codes.Code {
```
jakubno: can you move this to shared utils, it may come handy to some one else later

tests/integration/internal/tests/proxies/auto_resume_test.go
```diff
@@ -0,0 +1,72 @@
...
+
+	sbx := utils.SetupSandboxWithCleanup(t, c, utils.WithAutoPause(true), utils.WithAutoResume(api.Any))
+
+	pauseResp, err := c.PostSandboxesSandboxIDPauseWithResponse(ctx, sbx.SandboxID, setup.WithAPIKey())
+	require.NoError(t, err)
+	require.Equal(t, http.StatusNoContent, pauseResp.StatusCode())
+
+	require.Eventually(t, func() bool {
+		res, err := c.GetSandboxesSandboxIDWithResponse(ctx, sbx.SandboxID, setup.WithAPIKey())
+		if err != nil || res.JSON200 == nil {
```
jakubno: you should never see an error or other status than 200

tests/integration/internal/tests/proxies/auto_resume_test.go
```diff
@@ -0,0 +1,72 @@
...
+			return false
+		}
+
+		return res.JSON200.State == api.Paused
+	}, 30*time.Second, time.Second, "sandbox did not pause in time")
```
jakubno: you can porbably poll faster to make the test pass faster, like 100 ms

tests/integration/internal/tests/proxies/auto_resume_test.go
```diff
@@ -0,0 +1,72 @@
...
+	require.NoError(t, err)
+
+	client := &http.Client{Timeout: 10 * time.Second}
+	port := 3210
+
+	require.Eventually(t, func() bool {
+		req := utils.NewRequest(sbx, proxyURL, port, nil)
+		resp, err := client.Do(req)
+		if err != nil {
+			t.Logf("proxy request error: %v", err)
```
jakubno: you should fail here

tests/integration/internal/tests/proxies/auto_resume_test.go
```diff
@@ -0,0 +1,72 @@
...
+		return res.JSON200.State == api.Paused
+	}, 30*time.Second, time.Second, "sandbox did not pause in time")
+
+	proxyURL, err := url.Parse(setup.EnvdProxy)
+	require.NoError(t, err)
+
+	client := &http.Client{Timeout: 10 * time.Second}
+	port := 3210
+
+	require.Eventually(t, func() bool {
```
jakubno: why eventually?

tests/integration/internal/tests/proxies/auto_resume_test.go
```diff
@@ -0,0 +1,72 @@
...
+		req := utils.NewRequest(sbx, proxyURL, port, nil)
+		resp, err := client.Do(req)
+		if err != nil {
+			t.Logf("proxy request error: %v", err)
+
+			return false
+		}
+		defer resp.Body.Close()
+
+		if resp.StatusCode == http.StatusBadGateway {
```
jakubno: this isn't a good test, won't you get the same error if the sandbox isn't running? 

you should start a server inside the sandbox, pause it and you should test the server is accessible even after you pause the sandbox

jakubno (changes requested):

packages/shared/pkg/grpc/proxy/proxy.proto
matthewlouisbrockman: fair enough, yeeting the ability to pass timeout_seconds at all and just leaving the hardcoded 300 seconds.

packages/api/internal/handlers/proxy_grpc.go
matthewlouisbrockman: uhh yeah good point, should just be calling the api only

packages/api/main.go
matthewlouisbrockman: o, i should be able to use the same factory

matthewlouisbrockman: err maybe not, but can copy enough of it

matthewlouisbrockman: made a copy of it

tests/integration/internal/tests/proxies/auto_resume_test.go
matthewlouisbrockman: ah, yeah. testing both the server method + exec method now

tests/integration/internal/tests/proxies/auto_resume_test.go
matthewlouisbrockman: eh, yeah i guess it should be available immediately

matthewlouisbrockman: i may have needed it, slower to start

packages/shared/pkg/utils/grpc_codes.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"net/http"
+
+	"google.golang.org/grpc/codes"
+)
+
+// GRPCCodeFromHTTPStatus maps an HTTP status code to the closest matching gRPC
+// codes.Code.
+func GRPCCodeFromHTTPStatus(statusCode int) codes.Code {
```
jakubno, line 11: this handles only errors, what if i pass 200 here?

packages/api/internal/factories/grpc.go
```diff
@@ -0,0 +1,60 @@
...
+	"go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/keepalive"
+
+	e2bgrpc "github.com/e2b-dev/infra/packages/shared/pkg/grpc"
+	"github.com/e2b-dev/infra/packages/shared/pkg/logger"
+	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
+)
+
+func NewGRPCServer(tel *telemetry.Client) *grpc.Server {
```
jakubno, line 17: the same code is orchestrator package, can we move it to shared and reuse?

packages/api/internal/handlers/sandbox_create.go
```diff
@@ -148,8 +148,8 @@ func (a *APIStore) PostSandboxes(c *gin.Context) {
 	var autoResume *types.SandboxAutoResumeConfig
 	if body.AutoResume != nil {
 		autoResume = &types.SandboxAutoResumeConfig{}
-		if body.AutoResume.Policy != nil {
-			policy := types.SandboxAutoResumePolicy(*body.AutoResume.Policy)
+		if body.AutoResume.Policy != "" {
```
jakubno: this check is not unnecessary

packages/api/internal/orchestrator/create_instance.go
```diff
@@ -216,7 +216,7 @@ func (o *Orchestrator) CreateSandbox(
 	if autoResume != nil {
 		orchAutoResume = &orchestrator.SandboxAutoResumeConfig{}
 		if autoResume.Policy != nil {
```
jakubno: the db type isn't updated

packages/api/internal/factories/grpc.go
jakubno: the same code is orchestrator package, can we move it to shared and reuse?

packages/api/main.go
```diff
@@ -464,10 +457,6 @@ func run() int {
 		}
 
 		grpcServer.GracefulStop()
-		if err := grpcListener.Close(); err != nil && !errors.Is(err, net.ErrClosed) {
```
jakubno: i think you need to stop the listener doesn't you?

packages/client-proxy/internal/proxy/proxy.go
```diff
@@ -98,6 +106,7 @@ func isNotPausedError(err error) bool {
 
 func NewClientProxy(meterProvider metric.MeterProvider, serviceName string, port uint16, catalog catalog.SandboxesCatalog, pausedSandboxResumer PausedSandboxResumer, featureFlagsClient *featureflags.Client) (*reverseproxy.Proxy, error) {
 	getTargetFromRequest := reverseproxy.GetTargetFromRequest(env.IsLocal())
+	negResumeCache := newNegativeResumeCache()
```
jakubno: let's not complicate with cache right now 🙏🏻

packages/shared/pkg/utils/grpc_codes.go
jakubno, line 11: this handles only errors, what if i pass 200 here?

matthewlouisbrockman: it's only for mapping error codes (only called if apiError != nil) but implementing a 200 catch

packages/api/internal/factories/grpc.go
matthewlouisbrockman: hmm, im mostly worried about messing with the orchestrator if i don't have to, any reason to force reuse this early?

packages/api/internal/handlers/sandbox_create.go
matthewlouisbrockman: autoResume can be nil, not sure the policy can be anymore

matthewlouisbrockman: it's a string so it could be "" right?

matthewlouisbrockman: making a function to extract it + tests to handle different inputs

packages/client-proxy/internal/proxy/proxy.go
matthewlouisbrockman: kk

packages/api/internal/factories/grpc.go
matthewlouisbrockman: hmm, moving it to shared here goes

packages/shared/pkg/utils/grpc_codes.go
matthewlouisbrockman: duplicate from github issues earlier

packages/api/main.go
matthewlouisbrockman: grpc.Server.GracefulStop() should stop it already, it's passed the listener in 425 `err := grpcServer.Serve(grpcListener)` and whenever the server returns it closes the listerner

matthewlouisbrockman:
(rm'd some duplicate columns from github stability issues earlier)

packages/api/internal/handlers/sandbox_create.go
```diff
@@ -235,6 +226,36 @@ func (a *APIStore) PostSandboxes(c *gin.Context) {
...
+	}
+
+	policy := api.SandboxAutoResumePolicy(strings.TrimSpace(string(autoResume.Policy)))
+	if policy == "" {
+		return nil, &api.APIError{
+			Code:      http.StatusBadRequest,
+			ClientMsg: "autoResume.policy is required",
+			Err:       errors.New("autoResume.policy is required"),
+		}
+	}
```
jakubno: this is handled by the switch

packages/api/internal/orchestrator/nodemanager/sandboxes.go
```diff
@@ -78,10 +78,18 @@ func (n *Node) GetSandboxes(ctx context.Context) ([]sandbox.Sandbox, error) {
...
 		if autoResumeCfg := config.GetAutoResume(); autoResumeCfg != nil {
-			autoResume = &types.SandboxAutoResumeConfig{}
-			if p := strings.TrimSpace(autoResumeCfg.GetPolicy()); p != "" {
-				policy := types.SandboxAutoResumePolicy(p)
-				autoResume.Policy = &policy
+			p := strings.TrimSpace(autoResumeCfg.GetPolicy())
+			if p == "" {
+				p = string(types.SandboxAutoResumeOff)
+			}
+
```
jakubno: This is kinda unnecessary, this is coming from our data - no need to trim. I would just pass it, no need to do the switch either

packages/api/internal/orchestrator/create_instance.go
```diff
@@ -214,9 +214,15 @@ func (o *Orchestrator) CreateSandbox(
 
 	var orchAutoResume *orchestrator.SandboxAutoResumeConfig
 	if autoResume != nil {
-		orchAutoResume = &orchestrator.SandboxAutoResumeConfig{}
-		if autoResume.Policy != nil {
-			orchAutoResume.Policy = string(*autoResume.Policy)
+		policy := string(autoResume.Policy)
+		if policy == "" {
```
jakubno: if you want, feel free to validate and return 400 on invalid requests, but let's not introduce some unexpected behavior

packages/db/pkg/types/types.go
```diff
@@ -37,8 +37,8 @@ const (
 )
 
 type SandboxAutoResumeConfig struct {
-	// Policy is optional; unset means "off".
-	Policy *SandboxAutoResumePolicy `json:"policy,omitempty"`
+	// Policy defaults to "off". Empty/unknown values should be treated as "off".
```
jakubno: Empty/unknown value isn't valid

packages/client-proxy/internal/proxy/proxy.go
```diff
@@ -45,9 +54,46 @@ func catalogResolution(ctx context.Context, sandboxId string, c catalog.Sandboxe
...
+		}
 
+		return "", err
+	}
+
+	return nodeIP, nil
+}
+
+func isNotResumableError(err error) bool {
+	var grpcStatus interface{ GRPCStatus() *status.Status }
```
jakubno: can you use 		st, ok := status.FromError(err) to check if that's grpc error?
 we use it in other places of the codebase

packages/api/internal/handlers/sandbox_create.go
matthewlouisbrockman: k

matthewlouisbrockman: (it's a different error message but it should be prevented anyway)

packages/api/internal/orchestrator/nodemanager/sandboxes.go
matthewlouisbrockman: k, removed

packages/api/internal/orchestrator/create_instance.go
matthewlouisbrockman: ah, once it's inside it's trusted, kk

packages/client-proxy/internal/proxy/proxy.go
matthewlouisbrockman: done

packages/db/pkg/types/types.go
matthewlouisbrockman: rm'd

packages/api/internal/orchestrator/nodemanager/sandboxes.go
```diff
@@ -75,6 +75,17 @@ func (n *Node) GetSandboxes(ctx context.Context) ([]sandbox.Sandbox, error) {
...
 		}
 
+		var autoResume *types.SandboxAutoResumeConfig
+		if autoResumeCfg := config.GetAutoResume(); autoResumeCfg != nil {
+			p := autoResumeCfg.GetPolicy()
+			if p == "" {
+				p = string(types.SandboxAutoResumeOff)
+			}
+
+			policy := types.SandboxAutoResumePolicy(p)
```
jakubno, line 81: if the cfg isn't `nil` it shouldn't be ever empty 
```suggestion
			p := autoResumeCfg.GetPolicy()
			policy := types.SandboxAutoResumePolicy(p)
```

tests/integration/internal/tests/proxies/auto_resume_test.go
```diff
@@ -0,0 +1,171 @@
...
+	err := utils.ExecCommand(t, ctx, sbx, envdClient, "ls")
+	require.NoError(t, err)
+
+	// Pause the sandbox.
+	pauseResp, err := c.PostSandboxesSandboxIDPauseWithResponse(ctx, sbx.SandboxID, setup.WithAPIKey())
+	require.NoError(t, err)
+	require.Equal(t, http.StatusNoContent, pauseResp.StatusCode())
+
+	// Wait for sandbox to be paused.
+	deadline := time.Now().Add(30 * time.Second)
```
jakubno: the sandbox is paused right after pause returns

tests/integration/internal/tests/proxies/auto_resume_test.go
```diff
@@ -0,0 +1,171 @@
...
+		require.NotNil(t, res.JSON200, "expected 200 response, got status %d", res.StatusCode())
+
+		if res.JSON200.State == api.Paused {
+			break
+		}
+
+		require.True(t, time.Now().Before(deadline), "sandbox did not pause in time, state: %s", res.JSON200.State)
+		time.Sleep(100 * time.Millisecond)
+	}
+
```
jakubno:
```suggestion
	res, err := c.GetSandboxesSandboxIDWithResponse(ctx, sbx.SandboxID, setup.WithAPIKey())
	require.NoError(t, err)
	require.NotNil(t, res.JSON200, "expected 200 response, got status %d", res.StatusCode())
	require.Equal(res.JSON200.State, api.Paused)
```

tests/integration/internal/tests/proxies/auto_resume_test.go
```diff
@@ -0,0 +1,171 @@
...
+			break
+		}
+
+		if time.Now().After(deadline) {
+			require.NoError(t, err, "exec command did not succeed after auto-resume within timeout")
+		}
+
+		t.Logf("Exec failed (retrying): %v", err)
+		time.Sleep(100 * time.Millisecond)
+	}
```
jakubno:
```suggestion
	err = utils.ExecCommand(t, ctx, sbx, envdClient, "ls")
	require.NoError(t, err)
```

tests/integration/internal/tests/proxies/auto_resume_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+	// Verify the sandbox is running again.
+	res, err := c.GetSandboxesSandboxIDWithResponse(ctx, sbx.SandboxID, setup.WithAPIKey())
+	require.NoError(t, err)
+	require.NotNil(t, res.JSON200, "expected 200 response, got status %d", res.StatusCode())
+	require.Equal(t, api.Running, res.JSON200.State, "sandbox should be running after auto-resume")
+}
+
+func TestSandboxAutoResumeViaProxy(t *testing.T) {
```
jakubno, line 53: same here

jakubno (changes requested):
The code looks good! Just some improvements to tests and it's ready. Can you please also add a negative test (sandbox without auto resume won't resume)

matthewlouisbrockman:
github action failed but a lot on that one failed, rerunning to see if it's an issue with the specific runner

matthewlouisbrockman:
k, so it works with retries, so there's something on the original resume path (which has the api return the routing info directly) that breaks things

packages/client-proxy/internal/proxy/proxy.go
```diff
@@ -33,54 +33,65 @@ const (
 
 var ErrNodeNotFound = errors.New("node not found")
 
-func catalogResolution(ctx context.Context, sandboxId string, c catalog.SandboxesCatalog, pausedChecker PausedSandboxResumer, featureFlags *featureflags.Client) (string, error) {
+type autoResumeResult uint8
+
+const (
+	autoResumeNone autoResumeResult = iota
```
jakubno: autoResumeNone probably not needed

packages/client-proxy/internal/proxy/proxy.go
```diff
@@ -33,54 +33,65 @@ const (
...
 func handlePausedSandbox(
 	ctx context.Context,
 	sandboxId string,
 	pausedChecker PausedSandboxResumer,
 	featureFlags *featureflags.Client,
-) (string, error) {
+) (string, autoResumeResult, error) {
 	if pausedChecker == nil {
-		return "", nil
+		return "", autoResumeNone, nil
```
jakubno: can we simplify to 
```
		return "", autoResumeNotAllowed, nil
```

packages/client-proxy/internal/proxy/proxy.go
```diff
@@ -33,54 +33,65 @@ const (
...
+type autoResumeResult uint8
+
+const (
+	autoResumeNone autoResumeResult = iota
+	autoResumeSucceeded
+	autoResumeNotAllowed
+	autoResumeErrored
+)
+
+func catalogResolution(ctx context.Context, sandboxId string, c catalog.SandboxesCatalog, pausedChecker PausedSandboxResumer, featureFlags *featureflags.Client) (string, autoResumeResult, error) {
```
jakubno: why does this function return `autoResumeResult`?

packages/client-proxy/internal/proxy/proxy.go
matthewlouisbrockman: k, removed the catalog returning the autoResume result

packages/client-proxy/internal/proxy/proxy.go
matthewlouisbrockman: came from removing None

packages/client-proxy/internal/proxy/proxy.go
matthewlouisbrockman: k, removed it, used in tests.

PR #1863, opened 2026-02-06
fix: allow host parsing for proxy on test subdomains

matthewlouisbrockman:
## Summary
  - stop stripping subdomains in request host handling
  - allow `testing.e2bdev.dev` alongside the main domain
  - keep existing production behavior otherwise unchanged

  ## Why
  - the previous logic collapsed subdomains, which prevented using testing/staging domains

### DIFF
--- a/tests/integration/go.mod
+++ b/tests/integration/go.mod
@@ -23,7 +23,6 @@ require (
 	github.com/google/uuid v1.6.0
 	github.com/oapi-codegen/runtime v1.1.1
 	github.com/stretchr/testify v1.11.1
-	golang.org/x/net v0.47.0
 	golang.org/x/sync v0.18.0
 	google.golang.org/grpc v1.75.1
 	google.golang.org/protobuf v1.36.9
@@ -108,6 +107,7 @@ require (
 	go.uber.org/zap v1.27.0 // indirect
 	golang.org/x/crypto v0.45.0 // indirect
 	golang.org/x/mod v0.29.0 // indirect
+	golang.org/x/net v0.47.0 // indirect
 	golang.org/x/sys v0.38.0 // indirect
 	golang.org/x/term v0.37.0 // indirect
 	golang.org/x/text v0.31.0 // indirect


tests/integration/internal/utils/request.go
```diff
@@ -46,6 +50,22 @@ func NewRequest(sbx *api.Sandbox, url *url.URL, port int, extraHeaders *http.Hea
...
+func routingDomainFromProxyHost(proxyHost string) string {
+	labels := strings.Split(proxyHost, ".")
+	if len(labels) > 1 {
+		switch labels[0] {
+		case "client", "session":
+			return strings.Join(labels[1:], ".")
+		}
+	}
+
+	if eTLD, err := publicsuffix.EffectiveTLDPlusOne(proxyHost); err == nil && eTLD != "" {
```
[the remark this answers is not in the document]
matthewlouisbrockman: it's a test, not fixing

tests/integration/internal/utils/request.go
```diff
@@ -46,6 +50,22 @@ func NewRequest(sbx *api.Sandbox, url *url.URL, port int, extraHeaders *http.Hea
 	}
 }
 
+func routingDomainFromProxyHost(proxyHost string) string {
+	labels := strings.Split(proxyHost, ".")
+	if len(labels) > 1 {
+		switch labels[0] {
+		case "client", "session":
+			return strings.Join(labels[1:], ".")
```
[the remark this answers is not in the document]
matthewlouisbrockman: it's a test, not fixing

tests/integration/internal/utils/request.go
```diff
@@ -46,6 +50,22 @@ func NewRequest(sbx *api.Sandbox, url *url.URL, port int, extraHeaders *http.Hea
 	}
 }
 
+func routingDomainFromProxyHost(proxyHost string) string {
+	labels := strings.Split(proxyHost, ".")
+	if len(labels) > 1 {
+		switch labels[0] {
+		case "client", "session":
```
jakubno: can't we do `return strings.Join(labels[1:], ".")` for all?

because if somebody will name it e.g. `client-proxy` (which is mentioned in `.env.template`) it won't work

matthewlouisbrockman: o, yeah, that works

matthewlouisbrockman: err wait checking

matthewlouisbrockman: oh yeah that's yeah

PR #1868, opened 2026-02-09
improve: add StreamingChunker for progressive 4MB chunk fetching

levb:
### Summary

Thanks to @djeebus for the idea, and @claude for the help!

  - Add `StreamingChunker` that reads upstream data progressively and notifies callers as soon as their specific byte range arrives, instead of blocking until the entire 4MB chunk is fetched
  - Add `StreamingReader` interface (OpenRangeReader) to all storage backends (GCS, filesystem, AWS, NFS cache layer)
  - `StreamingChunker`'s use is controlled by `use-streaming-chunker` feature flag.

###  Motivation

The existing `Chunker` fetches an entire 4MB chunk before returning any data. A 4KB page fault triggers a 4MB download and the caller blocks until all 4MB arrive. Concurrent requests for different parts of the same chunk also block until the first fetch completes entirely.

With GCS latency this means a caller needing just the first 4KB waits up to 100ms for the full 4MB. The `StreamingChunker` reduces first-byte latency to the time it takes to stream just the requested blocks.

###  Design

  - Progressive fetch: a single goroutine per 4MB chunk streams data from upstream, marking blocks as cached in the mmap as they arrive
  - Sorted waiter list: callers register with their endByte and are notified in order as data arrives — O(satisfied) per progress tick instead of O(all)
  - Shared fetch goroutine: detached via context.WithoutCancel so it continues even if the first caller cancels; other waiters still benefit
  - Partial data on error: blocks written before an upstream failure stay cached and remain servable
  - Panic recovery: runFetch recovers panics and transitions waiters to error state instead of hanging forever
  - Timeout safety net: 60s fetch timeout at the chunker level, plus per-reader timeout at the GCS layer

###  Test plan

_NS: still reviewing the tests/bench_

  - TestStreamingChunker_BasicSlice — single slice returns correct data
  - TestStreamingChunker_CacheHit — second read served from cache, no upstream calls
  - TestStreamingChunker_ConcurrentSameChunk — 10 goroutines on same chunk, all get correct data
  - TestStreamingChunker_EarlyReturn — first block returns ~75x faster than last block (13ms vs 1041ms)
  - TestStreamingChunker_ErrorKeepsPartialData — upstream fails at 2MB, first page still cached
  - TestStreamingChunker_ContextCancellation — cancelled caller doesn't kill fetch for others
  - TestStreamingChunker_LastBlockPartial — unaligned file size handled correctly
  - TestStreamingChunker_MultiChunkSlice — request spanning two 4MB chunks
  - TestStreamingChunker_PanicRecovery — upstream panic returns error, doesn't hang or crash
  - BenchmarkRandomAccess — A/B comparison with simulated GCS latency
  - All tests pass with -race
  - make lint — 0 issues


packages/shared/pkg/storage/storage_google.go
```diff
@@ -224,6 +225,33 @@ func (o *gcpObject) Size(ctx context.Context) (int64, error) {
 	return attrs.Size, nil
 }
 
+func (o *gcpObject) OpenRangeReader(ctx context.Context, off, length int64) (io.ReadCloser, error) {
```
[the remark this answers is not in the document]
levb: `googleReadTimeout` is appropriate for the GCS operations, regardless of what the parent thinks; I think it's ok as is.

packages/orchestrator/internal/sandbox/block/streaming_chunk.go
```diff
@@ -0,0 +1,418 @@
...
+	})
+	s.waiters = slices.Insert(s.waiters, idx, w)
+
+	s.mu.Unlock()
+
+	select {
+	case err := <-w.ch:
+		return err
+	case <-ctx.Done():
+		return ctx.Err()
```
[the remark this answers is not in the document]
levb: A very small "leak" that will get cleaned up at timeout for the goroutine, 3/5 it's fine.

packages/shared/pkg/storage/storage_cache_seekable.go
```diff
@@ -122,6 +126,95 @@ func (c *cachedSeekable) ReadAt(ctx context.Context, buff []byte, offset int64)
 	return readCount, err
 }
 
+func (c *cachedSeekable) OpenRangeReader(ctx context.Context, off, length int64) (io.ReadCloser, error) {
+	// Try NFS cache file first
+	chunkPath := c.makeChunkFilename(off)
```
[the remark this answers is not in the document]
levb: Removed; dead code

packages/orchestrator/internal/sandbox/block/streaming_chunk.go
```diff
@@ -0,0 +1,418 @@
...
+	}
+
+	// Compute which 4MB chunks overlap with the requested range
+	firstChunkOff := (off / storage.MemoryChunkSize) * storage.MemoryChunkSize
+	lastChunkOff := ((off + length - 1) / storage.MemoryChunkSize) * storage.MemoryChunkSize
+
+	var eg errgroup.Group
+
+	for fetchOff := firstChunkOff; fetchOff <= lastChunkOff; fetchOff += storage.MemoryChunkSize {
+		eg.Go(func() error {
```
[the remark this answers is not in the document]
levb: It is per-iteration scoped in modern go

packages/orchestrator/internal/sandbox/build/storage_diff.go
```diff
@@ -94,15 +98,21 @@ func (b *StorageDiff) Init(ctx context.Context) error {
 		return errMsg
 	}
 
-	chunker, err := block.NewChunker(size, b.blockSize, obj, b.cachePath, b.metrics)
+	var c block.Chunker
+	if b.featureFlags != nil && b.featureFlags.BoolFlag(ctx, featureflags.UseStreamingChunkerFlag) {
```
djeebus: Can you check the flag usage in launch darkly to ensure that this context has some meaningful kinds? team, template, etc.

packages/orchestrator/internal/sandbox/block/streaming_chunk.go
```diff
@@ -0,0 +1,418 @@
...
+			clippedEnd := min(off+length, chunkEnd, c.size)
+			clippedLen := clippedEnd - clippedOff
+
+			if clippedLen <= 0 {
+				return nil
+			}
+
+			session := c.getOrCreateSession(ctx, fetchOff)
+
+			return session.registerAndWait(ctx, clippedOff, clippedLen)
```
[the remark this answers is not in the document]
levb: properly scoped in modern go

packages/orchestrator/internal/sandbox/build/storage_diff.go
levb: I believe so, https://github.com/e2b-dev/infra/blob/9dca4694fde3839f4db7db850faabfd65a810b3c/packages/orchestrator/internal/template/build/builder.go#L120 and https://github.com/e2b-dev/infra/blob/9dca4694fde3839f4db7db850faabfd65a810b3c/packages/orchestrator/internal/server/sandboxes.go#L62

@djeebus  Is there a way to extract these from `ctx` and log, to be sure?

packages/shared/pkg/storage/storage_aws.go
dobrac, line 250: can this part be implemented using `OpenRangeReader` (same for GCP)

packages/orchestrator/internal/sandbox/block/streaming_chunk.go
```diff
@@ -0,0 +1,418 @@
...
+	slice, err := c.Slice(ctx, off, int64(len(b)))
+	if err != nil {
+		return 0, fmt.Errorf("failed to slice cache at %d-%d: %w", off, off+int64(len(b)), err)
+	}
+
+	return copy(b, slice), nil
+}
+
+func (c *StreamingChunker) WriteTo(ctx context.Context, w io.Writer) (int64, error) {
+	chunk := make([]byte, storage.MemoryChunkSize)
```
dobrac, line 207: wondering if we could just extend the `Chunker` and avoid reimplementing some of the methods

packages/orchestrator/internal/sandbox/block/streaming_chunk.go
```diff
@@ -0,0 +1,418 @@
...
+		timer.Failure(ctx, length,
+			attribute.String(pullType, pullTypeLocal),
+			attribute.String(failureReason, failureTypeLocalRead))
+
+		return nil, fmt.Errorf("failed read from cache at offset %d: %w", off, err)
+	}
+
+	// Compute which 4MB chunks overlap with the requested range
+	firstChunkOff := (off / storage.MemoryChunkSize) * storage.MemoryChunkSize
+	lastChunkOff := ((off + length - 1) / storage.MemoryChunkSize) * storage.MemoryChunkSize
```
dobrac: can we reuse the existing `header.Block*` helpers here or add a new ones?

packages/orchestrator/internal/sandbox/block/streaming_chunk.go
```diff
@@ -0,0 +1,418 @@
...
+		state:    fetchStateRunning,
+	}
+
+	c.fetchMu.Lock()
+	if existing, ok := c.fetchMap[fetchOff]; ok {
+		c.fetchMu.Unlock()
+
+		return existing
+	}
+	c.fetchMap[fetchOff] = s
```
dobrac, line 304: we could use the smap instead (from the shared package)

packages/orchestrator/internal/sandbox/block/streaming_chunk.go
```diff
@@ -0,0 +1,418 @@
...
+			attribute.String(failureReason, failureTypeLocalRead))
+
+		return nil, fmt.Errorf("failed read from cache at offset %d: %w", off, err)
+	}
+
+	// Compute which 4MB chunks overlap with the requested range
+	firstChunkOff := (off / storage.MemoryChunkSize) * storage.MemoryChunkSize
+	lastChunkOff := ((off + length - 1) / storage.MemoryChunkSize) * storage.MemoryChunkSize
+
+	var eg errgroup.Group
```
dobrac, line 248: how do we handle if concurrent Slice calls are executed for overlapping ranges? Especially when the overlap is only in the 4MB chunk, not in the exactly offset + length range

packages/orchestrator/internal/sandbox/block/streaming_chunk.go
```diff
@@ -0,0 +1,418 @@
...
+			}
+		}
+		s.waiters = nil
+
+		return
+	}
+
+	// Progress: pop satisfied waiters from the sorted front.
+	i := 0
+	for i < len(s.waiters) && s.waiters[i].endByte <= s.bytesReady {
```
dobrac: have you considered using the `sync.Cond` instead?

packages/orchestrator/internal/sandbox/block/streaming_chunk.go
levb: Duplicating the original with a feature flag to choose (rather than modifying a common implementation) was a deliberate choice, to reduce risks. Once proven, StreamingChunker should be the only kind, and the legacy one deprecated and removed.

packages/shared/pkg/storage/storage_aws.go
levb: 2/5 I was trying to keep the existing code as little changed as possible; in the compression branch a lot of this will be refactored anyway

packages/orchestrator/internal/sandbox/block/streaming_chunk.go
levb: I haven't really used it before; according to Claud all waiters would be waking up on each broadcast; this way we notify just those who are ready. Also, with a channel it easy to plug the ctx->Done check.

packages/orchestrator/internal/sandbox/block/streaming_chunk.go
levb: fetch sessions are keyed off the MemoryChunkSize-aligned offsets, getOrCreateSession will create only 1 session per such chunk. I added a test.

packages/orchestrator/internal/sandbox/block/streaming_chunk.go
levb: But these are not in blocks! These are in MemoryChunkSize-s. The cache still uses the block-alignment to track isCached, and the helper functions

packages/orchestrator/internal/sandbox/block/streaming_chunk.go
levb: What would be the advantage? The string key is a turn off for this use case.

packages/orchestrator/internal/sandbox/block/streaming_chunk.go
```diff
@@ -0,0 +1,418 @@
...
+	blockSize := s.cache.BlockSize()
+	// endByte is the byte offset (relative to chunkOff) past which all blocks
+	// covering [off, off+length) are fully cached.
+	lastBlockIdx := (off + length - 1 - s.chunkOff) / blockSize
+	endByte := (lastBlockIdx + 1) * blockSize
+
+	s.mu.Lock()
+
+	// Fast path: already cached (handles pre-existing cache from prior sessions).
+	if s.cache.isCached(off, length) {
```
djeebus: if this will never flip from true to false, we could move the lock further down.

packages/orchestrator/internal/sandbox/block/streaming_chunk.go
```diff
@@ -0,0 +1,418 @@
...
+// Must be called with s.mu held.
+func (s *fetchSession) notifyWaiters(sendErr error) {
+	// Terminal: notify every remaining waiter.
+	if s.state != fetchStateRunning {
+		for _, w := range s.waiters {
+			if sendErr != nil && w.endByte > s.bytesReady {
+				w.ch <- sendErr
+			} else {
+				w.ch <- nil
+			}
```
djeebus: we should probably close the channel after sending, since nothing else will ever publish here, right?

packages/orchestrator/internal/sandbox/block/streaming_chunk.go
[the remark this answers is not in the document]
djeebus: This ain't your daddy's go! 🤣

packages/orchestrator/internal/sandbox/block/streaming_chunk.go
```diff
@@ -0,0 +1,418 @@
...
+	// overlapping the requested range. Always a multiple of blockSize.
+	endByte int64
+	ch      chan error // buffered cap 1
+}
+
+const (
+	fetchStateRunning = iota
+	fetchStateDone
+	fetchStateErrored
+)
```
djeebus: would be a nice devex improvement to add a `type fetchState int` and make these `const fetchStateRunning fetchState = iota`, etc.

packages/orchestrator/internal/sandbox/block/streaming_chunk.go
```diff
@@ -0,0 +1,418 @@
...
+		}
+	}()
+
+	mmapSlice, releaseLock, err := c.cache.addressBytes(s.chunkOff, s.chunkLen)
+	if err != nil {
+		s.mu.Lock()
+		s.state = fetchStateErrored
+		s.fetchErr = err
+		s.notifyWaiters(err)
+		s.mu.Unlock()
```
djeebus: Might be nice to add a function for this:

```golang
func (s *fetchSession) setResults(fetchState int, err error) {
	s.mu.Lock()
	s.state = fetchState
	s.fetchErr = err
	s.notifyWaiters(err)
	s.mu.Unlock()
}

File: packages/orchestrator/internal/sandbox/block/streaming_chunk.go (hunk 12 above)
Comment (levb) [reply packages/orchestrator/internal/sandbox/block/streaming_chunk.go]: I guess so, we can't change the same region's bits twice. I wonder if this second `isCached` check is even meaningful.

File: packages/orchestrator/internal/sandbox/build/storage_diff.go (hunk 5 above)
Comment (djeebus) [reply packages/orchestrator/internal/sandbox/build/storage_diff.go]: If you wanted to, you could expose `feature_flags.getContext` and call it. Seems like a lot of work though ...

Comment (djeebus) [reply packages/orchestrator/internal/sandbox/build/storage_diff.go]: Also, I don't believe b.featureFlags should ever be nil, right?

File: packages/shared/pkg/storage/storage_aws.go
Comment (dobrac) [reply packages/shared/pkg/storage/storage_aws.go:250]: ok

File: packages/orchestrator/internal/sandbox/block/streaming_chunk.go (hunk 8 above)
Comment (dobrac) [reply packages/orchestrator/internal/sandbox/block/streaming_chunk.go]: The main purpose is to unite these computations behind helper functions to help preventing accidental errors. Block, chunk, or any other name doesn't change the concept of the equation

File: packages/orchestrator/internal/sandbox/block/streaming_chunk.go (hunk 9 above)
Comment (dobrac) [reply packages/orchestrator/internal/sandbox/block/streaming_chunk.go:304]: just a simplification of the code (+sharding). We have it in our codebase already - keep the consistency

File: packages/orchestrator/internal/sandbox/block/streaming_chunk.go (hunk 10 above)
Comment (dobrac) [reply packages/orchestrator/internal/sandbox/block/streaming_chunk.go:248]: alright, thx

File: packages/orchestrator/internal/sandbox/block/streaming_chunk.go (hunk 15 above)
Comment (dobrac) [reply packages/orchestrator/internal/sandbox/block/streaming_chunk.go]: +1, would definitely be more readable

File: packages/orchestrator/internal/sandbox/block/streaming_chunk.go [hunk 16]
```diff
@@ -300,6 +297,16 @@ func (c *StreamingChunker) getOrCreateSession(ctx context.Context, fetchOff int6
...
 }
 
+func (s *fetchSession) setState(state fetchState, err error, onlyIfRunning bool) {
+	s.mu.Lock()
+	if !onlyIfRunning || s.state == fetchStateRunning {
+		s.state = state
+		s.fetchErr = err
+		s.notifyWaiters(err)
+	}
+	s.mu.Unlock()
```
dobrac: lets use defer

packages/orchestrator/internal/sandbox/block/fetchmap_bench_test.go
```diff
@@ -0,0 +1,166 @@
+package block
```
levb: This needs to be removed before merge - only here to demonstrate PR comments

packages/orchestrator/internal/sandbox/block/fetchmap_bench_test.md
```diff
@@ -0,0 +1,98 @@
+fetchmap benchmark: mutex+map vs concurrent-map
```
levb: This needs to be removed before merge - only here to demonstrate PR comments

packages/orchestrator/internal/sandbox/block/streaming_chunk.go
levb: I added a benchmark that shows that it'd be slower. I don't think the difference here is really material, and if you think it's worth it, I can easily replace. See (fetchmap_bench_test_*)

packages/shared/pkg/feature-flags/flags.go
```diff
@@ -92,6 +92,7 @@ var (
 	BestOfKTooManyStartingFlag          = newBoolFlag("best-of-k-too-many-starting", false)
 	EdgeProvidedSandboxMetricsFlag      = newBoolFlag("edge-provided-sandbox-metrics", false)
 	CreateStorageCacheSpansFlag         = newBoolFlag("create-storage-cache-spans", env.IsDevelopment())
+	UseStreamingChunkerFlag             = newBoolFlag("use-streaming-chunker", false)
```
levb: NS: set to true and rerun integ test

levb: Added https://app.launchdarkly.com/projects/default/flags/use-streaming-chunker

packages/orchestrator/internal/sandbox/build/storage_diff.go
```diff
@@ -94,15 +101,28 @@ func (b *StorageDiff) Init(ctx context.Context) error {
 		return errMsg
 	}
 
-	chunker, err := block.NewChunker(size, b.blockSize, obj, b.cachePath, b.metrics)
+	var c block.Chunker
+	useStreaming := b.featureFlags.BoolFlag(ctx, featureflags.UseStreamingChunkerFlag)
+	logger.L().Debug(ctx, "initializing chunker",
```
levb: Remove before merge - was added to see/demonstrate the behavior in CI

packages/orchestrator/internal/sandbox/block/streaming_chunk.go
levb: `setState`

packages/orchestrator/internal/sandbox/block/streaming_chunk.go
dobrac: thx or the thorough test, lets leave it then

packages/orchestrator/internal/sandbox/block/streaming_chunk.go
```diff
@@ -22,6 +22,15 @@ const (
...
 	defaultFetchTimeout = 60 * time.Second
+
+	// readBatchSize is the maximum number of bytes read from the upstream
+	// socket in a single Read call. Larger batches reduce syscall overhead
+	// (a 4KB read-cap on a 4MB chunk = 1024 syscalls). The notification
+	// granularity remains blockSize — after each Read we mark all newly
+	// completed blocks and wake their waiters. Read is allowed to return
+	// fewer bytes than requested, so we still wake promptly when the kernel
+	// buffer has less than a full batch.
+	readBatchSize = 256 * 1024 // 256 KB
```
dobrac: is this something we might want to adjust using FF?

dobrac (approved):

packages/orchestrator/internal/sandbox/block/streaming_chunk.go
levb: Done.

packages/orchestrator/internal/sandbox/block/streaming_chunk.go
```diff
@@ -373,7 +380,7 @@ func (c *StreamingChunker) progressiveRead(ctx context.Context, s *fetchSession,
 	defer reader.Close()
 
 	blockSize := c.cache.BlockSize()
-	readBatch := max(blockSize, minReadBatchSize)
+	readBatch := max(blockSize, c.minReadBatchSize)
```
dobrac: wouldn't it make sense to make this changeable on the fly rather than on the chunker creation?

packages/orchestrator/internal/sandbox/block/streaming_chunk.go
```diff
@@ -23,14 +24,9 @@ const (
...
-	// socket in a single Read call. Larger batches reduce syscall overhead
-	// (a 4KB read-cap on a 4MB chunk = 1024 syscalls). The notification
-	// granularity remains blockSize — after each Read we mark all newly
-	// completed blocks and wake their waiters. Read is allowed to return
-	// fewer bytes than requested, so we still wake promptly when the kernel
-	// buffer has less than a full batch.
-	readBatchSize = 256 * 1024 // 256 KB
+	// minReadBatchSize is the floor for the read batch size when blockSize
+	// is very small (e.g. 4KB rootfs). The actual batch is max(blockSize, minReadBatchSize).
+	minReadBatchSize = 16 * 1024 // 16 KB
```
dobrac: can you explain the thinking behind coming to 16KB (from 256KB) here please?

dobrac (commented):
> (chunkers are not used in CI integ test)

do we want to maybe use the chunker in the integration tests? Or have you at least checked they're passing with it?

packages/orchestrator/internal/sandbox/block/streaming_chunk.go
levb: It was from testing in my cluster*. The larger buffer was slowing down the callbacks on the 4K files, and there was no significant difference measured during the runs between doing the Read(16K) vs Read(256K) and a little slower than Read(4Mb). I have not had the time to review the Google clients (RPC vs HTTP) to see about their internal buffering. Claude blurted something out about gRPC client transmitting 16K at a time, I don't really think that's the case but the 16Kb did work better than 256k in my test.

The max(blockSize, ...) is so that in the 2Mb case we Read 2Mb at a time, no reason to split it smaller anyway.

*The test numbers were very noisy, I consider them "low confidence" and don't even want to paste them here since "next run may be different". With the flag ON there was _some_ improvement, most of the time.

Unfortunately, and that's why I spent a long time testing back and forth yesterday, there were runs when the new code performed slower overall (Sandbox resume), but it appears mostly due to GCP variability. Interestingly, both chunkers had occasions of taking O(10ms) to serve from MMap. Garbage Collection/MMap variability?...

PR #1874, opened 2026-02-09
stdin-close: prevent deadlock when large unconsumed buffer

matthewlouisbrockman:
handles an edge case with the `CloseStdin` where a full buffer can cause deadlock where can't close if unable to write

`WriteStdin`: now grabs a snapshot of p.stdin under stdinMu and releases the lock before calling Write.
`CloseStdin`: now snapshots p.stdin, sets `p.stdin = nil` under the lock (so future writes fail fast), then closes outside the lock.

### DIFF
--- a/packages/envd/internal/services/process/handler/handler.go
+++ b/packages/envd/internal/services/process/handler/handler.go
@@ -50,7 +50,7 @@ type Handler struct {
 	outCtx    context.Context //nolint:containedctx // todo: refactor so this can be removed
 	outCancel context.CancelFunc
 
-	stdinMu sync.Mutex
+	stdinMu sync.RWMutex
 	stdin   io.WriteCloser
 
 	DataEvent *MultiplexedChannel[rpc.ProcessEvent_Data]
@@ -329,9 +329,9 @@ func (p *Handler) WriteStdin(data []byte) error {
 		return fmt.Errorf("tty assigned to process — input should be written to the pty, not the stdin")
 	}
 
-	p.stdinMu.Lock()
+	p.stdinMu.RLock()
 	stdin := p.stdin
-	p.stdinMu.Unlock()
+	p.stdinMu.RUnlock()
 
 	if stdin == nil {
 		return fmt.Errorf("stdin not enabled or closed")
@@ -341,9 +341,9 @@ func (p *Handler) WriteStdin(data []byte) error {
 	if err != nil {
 		// CloseStdin can race with an in-flight write; in that case, the write may fail with a
 		// low-level "bad file descriptor"/"broken pipe" style error.
-		p.stdinMu.Lock()
+		p.stdinMu.RLock()
 		closed := p.stdin == nil
-		p.stdinMu.Unlock()
+		p.stdinMu.RUnlock()
 		if closed {
 			return fmt.Errorf("stdin closed or closing: %w", err)
 		}


packages/envd/internal/services/process/handler/handler.go
```diff
@@ -330,13 +330,14 @@ func (p *Handler) WriteStdin(data []byte) error {
 	}
 
 	p.stdinMu.Lock()
```
djeebus: Is it worth it to swap this to a RWMutex so that the slow lock only happens on close?

matthewlouisbrockman: will look into it. taking a look at the claude guys as well, I guess I should have run the claude /review locally first too

packages/envd/internal/services/process/handler/close_stdin_test.go
```diff
@@ -0,0 +1,98 @@
...
+		close(b.closeCalled)
+		close(b.unblockWrite)
+	}
+	return nil
+}
+
+func TestCloseStdinDoesNotDeadlockWithConcurrentWrite(t *testing.T) {
+	bw := newBlockingWriteCloser()
+
+	h := &Handler{
```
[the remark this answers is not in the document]
matthewlouisbrockman: ---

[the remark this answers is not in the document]
matthewlouisbrockman: oops that was for the above one, not gonna fix this one.

packages/envd/internal/services/process/handler/handler.go
```diff
@@ -330,13 +330,14 @@ func (p *Handler) WriteStdin(data []byte) error {
...
+	stdin := p.stdin
+	p.stdinMu.Unlock()
 
-	if p.stdin == nil {
+	if stdin == nil {
 		return fmt.Errorf("stdin not enabled or closed")
 	}
 
-	_, err := p.stdin.Write(data)
+	_, err := stdin.Write(data)
```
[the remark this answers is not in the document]
matthewlouisbrockman: fixed with envd: clarify WriteStdin error when stdin is closed concurrently, not sure it's that idk

packages/envd/internal/services/process/handler/handler.go
matthewlouisbrockman: Doesn't look like worth it - only holding stdinMu to snapshot/nil the pointer, the slower write/close ops are outside the lock. looks like RWMutex introduces a bit of extra overhead that can delay closestdin when lot of read-lock churn.

packages/envd/internal/services/process/handler/close_stdin_test.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+}
+
+func (b *blockingWriteCloser) Write(p []byte) (int, error) {
+	b.once.Do(func() { close(b.writeStarted) })
+	<-b.unblockWrite
+	return len(p), nil
+}
+
+func (b *blockingWriteCloser) Close() error {
```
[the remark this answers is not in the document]
matthewlouisbrockman: kk fixing

packages/envd/internal/services/process/handler/handler.go
matthewlouisbrockman: huh, wait, 

Goroutines | Mutex | RWMutex | Winner
-- | -- | -- | --
1 | 1.7 ns | 3.2 ns | Mutex (2x faster)
4 | 57 ns | 17 ns | RWMutex (3x faster)
8 | 93 ns | 66 ns | RWMutex (1.4x faster

with not doing any work, if we do some work,

Goroutines | Mutex | RWMutex | Diff
-- | -- | -- | --
1 | 237 ns | 233 ns | ~same
4 | 81 ns | 81 ns | ~same
8 | 130 ns | 90 ns | RWMutex ~30% faster

it is actually faster!

matthewlouisbrockman: eh, i guess it's like... ns tho so not really matter. gonna leave as is cause I dont know what RWMutex is but if it was doign a lot of concurrent locking I guess i see where it matters

djeebus: It *might* be worth it under high concurrency, but I'm not sure if that's reasonable. RWMutex is a mutex that allows lots of readers to share the lock, but if a writer wants to do something, the mutex blocks new readers, waits for existing readers to unlock, then gives an exclusive lock to the writer until its done. Then when the writer unlocks, it goes back to a free-for-all. Probably more accurate for this situation (since the "readers" are just accessing an instance concurrently, but the "writer" (the closer) is the only one modifying it. Not big one way or the other though, unless we get lots and lots of readers.

PR #1880, opened 2026-02-10
feat: collect cpu&mem sandbox stats from host

arkamar:
> [!NOTE]
> **Medium Risk**
> Introduces a new background sampler in the sandbox lifecycle and new ClickHouse writes/schema, which could affect sandbox shutdown timing and ClickHouse load if enabled.
> 
> **Overview**
> Adds a new ClickHouse-backed pipeline to *periodically sample Firecracker host CPU/memory usage per sandbox* (every 5s plus a final sample on shutdown) and store it in a new `sandbox_host_stats` table with a 7-day TTL. The orchestrator now optionally wires a host-stats delivery into `sandbox.Factory` and starts/stops the collector based on the `HostStatsEnabled` feature flag, while ClickHouse delivery lifecycle is updated to avoid closing shared connections from individual delivery targets.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 20196225ac81113921dda089019838f33aee98ec. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/orchestrator/internal/sandbox/sandbox.go
+++ b/packages/orchestrator/internal/sandbox/sandbox.go
@@ -370,61 +370,6 @@ func handleSpanError(span trace.Span, err *error) {
 	}
 }
 
-// initializeHostStatsCollector initializes the host stats collector for a sandbox if enabled.
-// This is a best-effort operation - errors are logged but do not fail the sandbox initialization.
-func (f *Factory) initializeHostStatsCollector(
-	ctx context.Context,
-	sbx *Sandbox,
-	fcHandle *fc.Process,
-	buildID string,
-	runtime RuntimeMetadata,
-	config Config,
-) {
-	hostStatsEnabled := f.featureFlags.BoolFlag(ctx, featureflags.HostStatsEnabled)
-	if !hostStatsEnabled || f.hostStatsDelivery == nil {
-		return
-	}
-
-	firecrackerPID, err := fcHandle.Pid()
-	if err != nil {
-		logger.L().Error(ctx, "failed to get firecracker PID for host stats",
-			zap.String("sandbox_id", runtime.SandboxID),
-			zap.Error(err))
-
-		return
-	}
-
-	teamID, err := uuid.Parse(runtime.TeamID)
-	if err != nil {
-		logger.L().Error(ctx, "error parsing team ID", logger.WithTeamID(runtime.TeamID), zap.Error(err))
-	}
-
-	collector, err := NewHostStatsCollector(
-		HostStatsMetadata{
-			SandboxID:   runtime.SandboxID,
-			ExecutionID: runtime.ExecutionID,
-			TemplateID:  runtime.TemplateID,
-			BuildID:     buildID,
-			TeamID:      teamID,
-			VCPUCount:   config.Vcpu,
-			MemoryMB:    config.RamMB,
-		},
-		int32(firecrackerPID),
-		f.hostStatsDelivery,
-	)
-	if err != nil {
-		logger.L().Error(ctx, "failed to create host stats collector",
-			zap.String("sandbox_id", runtime.SandboxID),
-			zap.Error(err))
-
-		return
-	}
-
-	sbx.hostStatsCollector = collector
-
-	go collector.Start(ctx)
-}
-
 // ResumeSandbox resumes the sandbox from already saved template or snapshot.
 // IMPORTANT: You must Close() the sandbox after you are done with it.
 func (f *Factory) ResumeSandbox(
@@ -718,7 +663,9 @@ func (f *Factory) ResumeSandbox(
 
 	telemetry.ReportEvent(execCtx, "envd initialized")
 
-	f.initializeHostStatsCollector(execCtx, sbx, fcHandle, meta.Template.BuildID, runtime, config)
+	if f.featureFlags.BoolFlag(execCtx, featureflags.HostStatsEnabled) {
+		initializeHostStatsCollector(execCtx, sbx, fcHandle, meta.Template.BuildID, runtime, config, f.hostStatsDelivery)
+	}
 
 	go sbx.Checks.Start(execCtx)
 
--- a/packages/orchestrator/internal/sandbox/hoststats.go
+++ b/packages/orchestrator/internal/sandbox/hoststats.go
@@ -1,0 +1,67 @@
+package sandbox
+
+import (
+	"context"
+
+	"github.com/google/uuid"
+	"go.uber.org/zap"
+
+	"github.com/e2b-dev/infra/packages/clickhouse/pkg/hoststats"
+	"github.com/e2b-dev/infra/packages/orchestrator/internal/sandbox/fc"
+	"github.com/e2b-dev/infra/packages/shared/pkg/logger"
+)
+
+// initializeHostStatsCollector initializes the host stats collector for a sandbox.
+// This is a best-effort operation - errors are logged but do not fail the sandbox initialization.
+func initializeHostStatsCollector(
+	ctx context.Context,
+	sbx *Sandbox,
+	fcHandle *fc.Process,
+	buildID string,
+	runtime RuntimeMetadata,
+	config Config,
+	hostStatsDelivery hoststats.Delivery,
+) {
+	if hostStatsDelivery == nil {
+		return
+	}
+
+	firecrackerPID, err := fcHandle.Pid()
+	if err != nil {
+		logger.L().Error(ctx, "failed to get firecracker PID for host stats",
+			zap.String("sandbox_id", runtime.SandboxID),
+			zap.Error(err))
+
+		return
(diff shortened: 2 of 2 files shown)


packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -657,6 +663,59 @@ func (f *Factory) ResumeSandbox(
...
+			// Get template metadata for BuildID
+			meta, metaErr := t.Metadata()
+			buildID := ""
+			if metaErr == nil {
+				buildID = meta.Template.BuildID
+			}
+
+			teamID, err := uuid.Parse(runtime.TeamID)
+			if err != nil {
+				logger.L().Error(ctx, "error parsing team ID", zap.String("team_id", runtime.TeamID), zap.Error(err))
```
```suggestion
				logger.L().Error(ctx, "error parsing team ID", logger.WithTeamID(runtime.TeamID), zap.Error(err))
```

packages/clickhouse/pkg/hoststats/hoststats.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"context"
+	"time"
+
+	"github.com/google/uuid"
+)
+
+// SandboxHostStat represents a single host-level statistics sample
+// for a Firecracker process running a sandbox
+type SandboxHostStat struct {
```
jakubno, line 12: Maybe it would be worth it to also save total memory + cpu count, it should be very well compressible, so it won't take much data and it may make our life easier

packages/orchestrator/internal/sandbox/sandbox.go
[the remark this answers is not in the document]
arkamar: The idea was to collect values event if TeamID is not for whatever reason available. Collected data should contain enough identifiers for reasonable processing even without TeamID.

packages/clickhouse/pkg/hoststats/delivery.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func NewDefaultClickhouseHostStatsDelivery(
+	ctx context.Context,
+	conn driver.Conn,
+	featureFlags *flags.Client,
+) (*ClickhouseDelivery, error) {
+	maxBatchSize := featureFlags.IntFlag(ctx, flags.ClickhouseBatcherMaxBatchSize)
+	maxDelay := time.Duration(featureFlags.IntFlag(ctx, flags.ClickhouseBatcherMaxDelay)) * time.Millisecond
+	batcherQueueSize := featureFlags.IntFlag(ctx, flags.ClickhouseBatcherQueueSize, flags.SandboxContext("clickhouse-host-stats-batcher"))
```
jakubno: It isn't `sandboxID`, no need to use it here anyway
```suggestion
	batcherQueueSize := featureFlags.IntFlag(ctx, flags.ClickhouseBatcherQueueSize)
```

packages/orchestrator/internal/sandbox/hoststats_collector.go
```diff
@@ -0,0 +1,136 @@
...
+	BuildID     string
+	TeamID      uuid.UUID
+}
+
+func NewHostStatsCollector(
+	metadata HostStatsMetadata,
+	firecrackerPID int32,
+	delivery hoststats.Delivery,
+) (*HostStatsCollector, error) {
+	proc, err := process.NewProcess(firecrackerPID)
```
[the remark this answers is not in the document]
arkamar: orchestrator will be notified when the firecracker dies unexpectedly which will stop the collector as well.

packages/orchestrator/main.go
```diff
@@ -324,6 +327,24 @@ func run(config cfg.Config) (success bool) {
 
 		sbxEventsDeliveryTargets = append(sbxEventsDeliveryTargets, sbxEventsDeliveryClickhouse)
 		closers = append(closers, closer{"sandbox events delivery for clickhouse", sbxEventsDeliveryClickhouse.Close})
+
+		// Clickhouse sandbox host stats delivery
+		clickhouseConnHostStats, err := clickhouse.NewDriver(config.ClickhouseConnectionString)
```
jakubno: maybe we don't need to create two drivers, but we can move the closing outside the batcher. imho it make more anyway, if we pass it as an argument to the constructor, the object shouldn't close it

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -657,6 +663,59 @@ func (f *Factory) ResumeSandbox(
 
 	telemetry.ReportEvent(execCtx, "envd initialized")
 
+	// Initialize host stats collector if enabled
+	hostStatsEnabled := f.featureFlags.BoolFlag(ctx, featureflags.HostStatsEnabled)
```
jakubno: can we move all this logic to some helper function. It's already huge and this logic isn't "core", so it would make sense to move it to a separate function handling the metrics logic.

I think it will also improve readability as we won't have to do so nested if statement

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -657,6 +663,59 @@ func (f *Factory) ResumeSandbox(
...
+	if hostStatsEnabled && f.hostStatsDelivery != nil {
+		firecrackerPID, err := fcHandle.Pid()
+		if err != nil {
+			logger.L().Error(ctx, "failed to get firecracker PID for host stats",
+				zap.String("sandbox_id", runtime.SandboxID),
+				zap.Error(err))
+		} else {
+			// Get template metadata for BuildID
+			meta, metaErr := t.Metadata()
```
jakubno: this is already in scope and with handled error

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -657,6 +663,59 @@ func (f *Factory) ResumeSandbox(
...
+				// Take first sample immediately after envd initialization
+				err = collector.CollectSample(ctx)
+				if err != nil {
+					logger.L().Error(ctx, "failed to collect initial host stats",
+						zap.String("sandbox_id", runtime.SandboxID),
+						zap.Error(err))
+				}
+
+				// Start periodic sampling in background
+				collector.Start(execCtx)
```
jakubno: I think it's more often to call `go` from outside and keep the function synchronous. It helps with testing and gives you more flexibility

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -657,6 +663,59 @@ func (f *Factory) ResumeSandbox(
...
+			)
+			if err != nil {
+				logger.L().Error(ctx, "failed to create host stats collector",
+					zap.String("sandbox_id", runtime.SandboxID),
+					zap.Error(err))
+			} else {
+				sbx.hostStatsCollector = collector
+
+				// Take first sample immediately after envd initialization
+				err = collector.CollectSample(ctx)
```
jakubno: this could be also part of collector for consistency as `Stop` is already doing the last collection

packages/clickhouse/pkg/hoststats/delivery.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func (c *ClickhouseDelivery) Close(context.Context) error {
+	defer c.conn.Close()
+
+	return c.batcher.Stop()
+}
+
+func (c *ClickhouseDelivery) batchInserter(ctx context.Context, stats []SandboxHostStat) error {
+	batch, err := c.conn.PrepareBatch(ctx, InsertSandboxHostStatQuery, driver.WithReleaseConnection())
```
jakubno, line 97: do you think we need to release the connection here?

jakubno (changes requested):

packages/clickhouse/pkg/hoststats/delivery.go
arkamar: no, we don't need to release the connection here. I reworked hoststats and events Close() methods, the clickhouse connection is released in main.go.

jakubno: I think this is little bit different object. The connection here is the actual connection, the `conn` in main is connection pool

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -364,6 +370,61 @@ func handleSpanError(span trace.Span, err *error) {
 	}
 }
 
+// initializeHostStatsCollector initializes the host stats collector for a sandbox if enabled.
+// This is a best-effort operation - errors are logged but do not fail the sandbox initialization.
+func (f *Factory) initializeHostStatsCollector(
```
jakubno: can we move it to another file and I would make it just pure function instead of being a `Factory` method. I don't see much reason why it should live on `Factory`, but open to discussion

jakubno (approved):

PR #1891, opened 2026-02-11
Instrument storage Size() calls with latency metrics

levb:
Add timer recording to GCS and NFS cache Size() methods to measure how much time is actually spent on Size() calls during sandbox creation. This data will help decide whether to move forward with https://github.com/e2b-dev/infra/pull/1887, which defers Size() calls but introduces a fair amount of complexity in the Size() handling.

Metrics produced:
- orchestrator.storage.gcs.read with operation=Size
- orchestrator.storage.slab.nfs.read with operation=Size (vs ReadAt)

### DIFF
--- a/packages/shared/pkg/storage/storage_cache_seekable.go
+++ b/packages/shared/pkg/storage/storage_cache_seekable.go
@@ -139,8 +139,8 @@ func (c *cachedSeekable) Size(ctx context.Context) (n int64, e error) {
 
 	size, err := c.readLocalSize(ctx)
 	if err == nil {
-		recordCacheRead(ctx, true, 8, cacheTypeSeekable, cacheOpSize)
-		readTimer.Success(ctx, 8)
+		recordCacheRead(ctx, true, 0, cacheTypeSeekable, cacheOpSize)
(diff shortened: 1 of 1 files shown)


packages/shared/pkg/storage/storage_cache_seekable.go
```diff
@@ -129,12 +135,16 @@ func (c *cachedSeekable) Size(ctx context.Context) (n int64, e error) {
 		span.End()
 	}()
 
+	readTimer := cacheSlabReadTimerFactory.Begin(attribute.String(nfsCacheOperationAttr, nfsCacheOperationAttrSize))
+
 	size, err := c.readLocalSize(ctx)
 	if err == nil {
 		recordCacheRead(ctx, true, 8, cacheTypeSeekable, cacheOpSize)
+		readTimer.Success(ctx, 8)
```
jakubno: could you add a comment for `8`, ideally make it constant 🙏🏻

levb: Let me change all these to 0, this was a mistake, would lead to inconsistencies.

PR #1893, opened 2026-02-11
Add files API

djeebus:
TODO:
- [x] combine the "token" endpoint functionality into "get volume"
- [x] add tests to `entryInfo` helper functions
- [x] retry orchestrator errors that indicate "volume type unknown"
- [ ] sdk integration testing
- [x] assign files "volume" user and "volume" group by default
- [x] centralize tryParseUUID, refactor to use it
- [x] add rotation support
- [x] asymmetric by default
- [x] if volume type unknown error occurs, only log once, not once per orchestrator

later:
- [ ] follow up later with jwks
- [ ] unify delete function

---

> [!NOTE]
> **High Risk**
> Introduces new token-signing configuration and changes how persistent volumes are provisioned/mounted and accessed (Filestore/NFS + proxy pathing), which can impact security and data availability if misconfigured. Also adds substantial new orchestrator I/O surface area that needs careful validation and rollout.
> 
> **Overview**
> Adds a first-pass “files API” for volumes by extending the orchestrator volume service with filesystem-style operations (stat/list/create/update/delete for files and dirs) and wiring the API to mint and return a signed per-volume content token on volume create/get responses. Deployment/IaC is updated to support persistent volume types on GCP (new Filestore-backed mounts, configurable NFS version, and propagated volume/token env vars), and the NFS proxy is hardened and reworked to mount by `teamID`/`volumeID` with improved jail/path validation and expanded e2e coverage.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6d92ec6cfd887855ca5a4e9f276a4653395f61bc. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api/internal/handlers/volume_util.go
+++ b/packages/api/internal/handlers/volume_util.go
@@ -116,6 +116,7 @@ func (a *APIStore) executeOnOrchestratorByClusterID(
 
 		if node.Status() != api.NodeStatusReady {
 			notReadyNodeCount++
+
 			continue
 		}
 
@@ -153,7 +154,7 @@ func (a *APIStore) executeOnOrchestratorByClusterID(
 		return err
 	}
 
-	if receivedUnknownVolumeTypeErrors == len(nodes)-notReadyNodeCount {
+	if receivedUnknownVolumeTypeErrors == len(nodes)-notReadyNodeCount && receivedUnknownVolumeTypeErrors > 0 {
 		return fmt.Errorf("%w: %s", ErrUnknownVolumeType, unknownVolumeType)
 	}
 


spec/openapi.yml
```diff
@@ -3122,3 +3160,131 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/VolumeDirectoryListing"
+        "404":
+          description: "path not found"
+        "500":
+          $ref: "#/components/responses/500"        
+    post:
+      description: List directory contents
```
mishushakov: incorrect

spec/openapi.yml
```diff
@@ -3122,3 +3160,131 @@ paths:
...
+      parameters:
+        - $ref: "#/components/parameters/volumeID"
+        - $ref: "#/components/parameters/path"
+      responses:
+        "204":
+          description: "Successfully deleted a directory"
+        "404":
+          description: "path not found"
+        "500":
+          $ref: "#/components/responses/500"
```
mishushakov: I would just want delete both using /volumes/{volumeID}/file

spec/openapi.yml
```diff
@@ -1696,6 +1702,93 @@ components:
       required:
         - name
 
+    VolumeDirectoryItem:
+      type: object
+      properties:
+        name:
+          type: string
+          description: Name of the file or directory
+        isDirectory:
```
mishushakov: we can have a type enum here: FILE, DIRECTORY, SYMLINK, FIFO, SOCKET

spec/openapi.yml
```diff
@@ -3122,3 +3215,153 @@ paths:
...
+          content:
+            application/octet-stream:
+              schema:
+                type: string
+                format: binary
+        "404":
+          description: "path not found"
+        "500":
+          $ref: "#/components/responses/500"
+    post:
```
mishushakov: this will also create directories recursively when writing to /this/path/does/not/exist

spec/openapi.yml
```diff
@@ -3122,3 +3215,153 @@ paths:
...
+        - $ref: "#/components/parameters/volumeID"
+        - $ref: "#/components/parameters/path"
+      responses:
+        "204":
+          description: "Successfully created a directory"
+        "404":
+          description: "path not found"
+        "500":
+          $ref: "#/components/responses/500"
+    delete:
```
mishushakov: this will no longer be needed as we can use file endpoint to delete files

spec/openapi.yml
```diff
@@ -3122,3 +3215,153 @@ paths:
...
+              type: string
+              format: binary
+      responses:
+        "200":
+          description: "Successfully created a file"
+        "404":
+          description: "path not found"
+        "500":
+          $ref: "#/components/responses/500"
+    delete:
```
mishushakov: this will also delete directories

spec/openapi.yml
```diff
@@ -1696,6 +1702,93 @@ components:
...
+        - ctime
+        - mtime
+
+    VolumeStat:
+      type: object
+      properties:
+        name:
+          type: string
+        type:
+          type: string
```
mishushakov: should be enum, string is too vague

spec/openapi.yml
```diff
@@ -1696,6 +1702,93 @@ components:
...
+        - path
+        - size
+        - mode
+        - permissions
+        - owner
+        - group
+        - mtime
+        - ctime
+
+    VolumeDirectoryListing:
```
mishushakov: any reason this is not just a list of VolumeDirectoryItem?

spec/openapi.yml
mishushakov: we will also have a force option to recursively delete all files in a directory (will throw if there are any files still left in a dir when deleting)

spec/openapi.yml
mishushakov: we can add a force option to forcefully override a file if it exists. i am not sure if it wouldn't be annoying though

spec/openapi.yml
```diff
@@ -3122,3 +3215,153 @@ paths:
...
+          description: "Successfully retrieved a directory listing"
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/VolumeDirectoryListing"
+        "404":
+          description: "path not found"
+        "500":
+          $ref: "#/components/responses/500"
+    post:
```
mishushakov: should also be able to create directories recursively: /this/should/work

spec/openapi.yml
```diff
@@ -3122,3 +3215,153 @@ paths:
...
+          description: Successfully retrieved path information
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/VolumeStat"
+        "404":
+          $ref: "#/components/responses/404"
+
+  /volumes/{volumeID}/dir:
+    get:
```
mishushakov: it should maybe throw an error when trying to list a file

spec/openapi.yml
```diff
@@ -1696,6 +1702,93 @@ components:
...
+          description: Last modification time of the file or directory
+      required:
+        - name
+        - isDirectory
+        - mode
+        - size
+        - ctime
+        - mtime
+
+    VolumeStat:
```
mishushakov: VolumeEntryStat?

spec/openapi.yml
```diff
@@ -1696,6 +1702,93 @@ components:
       required:
         - name
 
+    VolumeDirectoryItem:
```
mishushakov: VolumeEntry?

spec/openapi.yml
mishushakov: as discussed on slack there will be optional parameter to specify user + group, I propose this format: user:group

spec/openapi.yml
djeebus: Pretty sure you literally cannot have a fifo or socket over nfs. symlinks should be added though, thanks!

spec/openapi.yml
djeebus: why make the backend parse strings if we don't have to?

spec/openapi.yml
djeebus: This gives us a place to put other metadata if we need to in the future: pagination, sorting, timing, size, etc. In general it's best to avoid returning an array directly, as it prevents you from doing things like that. Not super opposed to it, it just might cause problems in the future. 

... also not sure what the rest of our APIs look like. probably best to make this look like those.

mishushakov: yeah I think Sandbox list is a good inspiration here

spec/openapi.yml
```diff
@@ -3168,3 +3225,206 @@ paths:
...
+          description: Successfully retrieved path information
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/VolumeEntryStat"
+        "404":
+          $ref: "#/components/responses/404"
+
+  /volumes/{volumeID}/dir:
+    get:
```
mishushakov: can we specify depth?

spec/openapi.yml
```diff
@@ -3168,3 +3225,206 @@ paths:
...
+          schema:
+            type: integer
+            format: uint32
+        - name: mode
+          in: query
+          description: Mode of the created directory
+          schema:
+            type: integer
+            format: uint32
+        - name: create_parents
```
mishushakov: here you use snake_case, elsewhere camelCase

spec/openapi.yml
```diff
@@ -3168,3 +3225,206 @@ paths:
...
+      responses:
+        "204":
+          description: "Successfully deleted a directory"
+        "404":
+          description: "path not found"
+        "500":
+          $ref: "#/components/responses/500"
+
+  /volumes/{volumeID}/file:
+    get:
```
mishushakov: we do not need userid here?

spec/openapi.yml
```diff
@@ -3168,3 +3225,206 @@ paths:
...
+            application/json:
+              schema:
+                $ref: "#/components/schemas/VolumeEntryStat"
+        "404":
+          $ref: "#/components/responses/404"
+
+  /volumes/{volumeID}/dir:
+    get:
+      description: List directory contents
+      tags: [volume_content]
```
mishushakov: please use volumes tag like in the other endpoint!!

spec/openapi.yml
```diff
@@ -3168,3 +3225,206 @@ paths:
...
+          $ref: "#/components/responses/500"
+    delete:
+      description: Delete a file
+      tags: [volume_content]
+      security:
+        - AccessTokenAuth: []
+        - ApiKeyAuth: []
+        - Supabase1TokenAuth: []
+          Supabase2TeamAuth: []
+      parameters:
```
mishushakov: no recursive option here?

spec/openapi.yml
```diff
@@ -3168,3 +3225,206 @@ paths:
...
+          content:
+            application/octet-stream:
+              schema:
+                type: string
+                format: binary
+        "404":
+          description: "path not found"
+        "500":
+          $ref: "#/components/responses/500"
+    post:
```
mishushakov: here we should return the entry info

spec/openapi.yml
```diff
@@ -3168,3 +3225,206 @@ paths:
...
+          description: Force overwrite of an existing file
+          schema:
+            type: boolean
+      requestBody:
+        required: true
+        content:
+          application/octet-stream:
+            schema:
+              type: string
+              format: binary
```
mishushakov: can this be FormData?

mishushakov: worried that the SDK client will serialize ArrayBuffer bodies.

spec/openapi.yml
```diff
@@ -3168,3 +3225,206 @@ paths:
...
+          in: query
+          description: Mode of the created file
+          schema:
+            type: integer
+            format: uint32
+        - name: force
+          in: query
+          description: Force overwrite of an existing file
+          schema:
+            type: boolean
```
mishushakov: since we're re-using this in dirs can we make it some other type, like VolumeWriteOpts?

spec/openapi.yml
djeebus: I don't think so; it's assumed that, if the API user has access to the volume, they can access the file.

spec/openapi.yml
djeebus: How do you recursively delete a file?

spec/openapi.yml
djeebus: We cannot, for a few reasons:
1. There's just no way to put multiple parameters into a single schema that can be referenced
2. Even if there was, they're slightly different - "Mode of the created file" vs "Mode of the created directory"

spec/openapi.yml
djeebus: I can support that too :+1:

djeebus: Actually, let's avoid the added complexity and overheard of multipart. Just use `file.stream()` and it won't buffer it.

spec/openapi.yml
mishushakov: i was thinking if we use the same endpoint for deleting folders

spec/openapi.yml
djeebus: We'll get to cleaning this up after the first release.

spec/openapi.yml
djeebus: We'll get to this after the first release.

spec/openapi.yml
djeebus: Not yet, but I'll add that soon. Maybe after this release though.

packages/api/internal/handlers/volumecontent_filecreate.go
```diff
@@ -0,0 +1,99 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to receive finish message: %w", err)
+		}
+
+		entry := toVolumeEntryStat(finish.GetEntry())
+		c.JSON(http.StatusCreated, entry)
+
+		return nil
+	})
+}
```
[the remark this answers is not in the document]
djeebus: We rely on the orchestrator to cleanse paths, the api doesn't need to.

spec/openapi.yml
```diff
@@ -3246,3 +3298,261 @@ paths:
...
+          schema:
+            type: integer
+            format: uint32
+        - name: mode
+          in: query
+          description: Mode of the created directory
+          schema:
+            type: integer
+            format: uint32
+        - name: createParents
```
mishushakov: can we name this also force for consistency?

iac/provider-gcp/nomad-cluster/main.tf
```diff
@@ -7,9 +7,9 @@ locals {
   nfs_mount_path   = "/orchestrator/shared-store"
   nfs_mount_subdir = "chunks-cache"
   nfs_mount_opts = join(",", [ // for more docs, see https://linux.die.net/man/5/nfs
-    format("nfsvers=%s", var.filestore_cache_enabled ? module.filestore[0].nfs_version == "NFS_V3" ? "3" : "4" : ""),
+    format("nfsvers=%s", var.filestore_cache_enabled ? module.filestore[0].nfs_version == "NFS_V3" ? "3" : "4.1" : ""),
```
levb: this is an unrelated update, right?

iac/provider-gcp/redis/variables.tf
```diff
@@ -16,8 +16,7 @@ variable "gcp_zone" {
 }
 
 variable "network_name" {
-  type    = string
-  default = "default"
```
levb, line 20: ?

packages/orchestrator/internal/volumes/dir_delete.go
```diff
@@ -0,0 +1,42 @@
...
+		fn = os.RemoveAll
+	} else {
+		fn = os.Remove
+	}
+
+	logger.L().Info(ctx, "removing directory",
+		zap.String("path", fullPath),
+	)
+
+	if err := fn(fullPath); err != nil { // todo: better error handling
```
levb: What better error handling?

levb (commented):
LGTM, but I am not sure I sufficiently understand all the mechanics, so comment only.

packages/orchestrator/internal/volumes/dir_delete.go
djeebus: Oops, already taken care of by adding lines 17-19, I'll remove this comment.

iac/provider-gcp/redis/variables.tf
djeebus: I only removed the buried default. This is an internal module, only called by the parent, and adding the default only allows us to forget to set it via the parent field. The default is now more appropriate up one level at the public interface level, in `iac/provider-gcp/variables.tf`.

iac/provider-gcp/nomad-cluster/main.tf
djeebus: Unrelated, but similar to another bug caught by code review in provider-gcp/main.tf, line 55. It's the same bug, so I fixed it here too.

iac/provider-gcp/variables.tf
```diff
@@ -598,3 +598,25 @@ variable "loki_use_v13_schema_from" {
     error_message = "must be YYYY-MM-DD"
   }
 }
+
+variable "persistent_volume_types" {
+  description = "Persistence layer for volumes"
+
+  type = map(object({
+    tier        = string
+    location    = optional(string)
```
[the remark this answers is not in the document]
djeebus: Not true, `location` is optional and defaults (I believe) to the zone of the provider.

spec/openapi.yml
```diff
@@ -3246,3 +3302,260 @@ paths:
           $ref: "#/components/responses/404"
         "500":
           $ref: "#/components/responses/500"
+
+  /volumes/{volumeID}/stat:
+    get:
+      description: Get path information
+      tags: [volumes]
+      security:
+        - AccessTokenAuth: []
```
djeebus: remove these, obsolete for new endpoints

packages/orchestrator/internal/volumes/service.go
```diff
@@ -13,68 +16,94 @@ import (
...
-		return nil, status.Errorf(codes.Internal, "failed to create volume: %v", err)
+	// already a grpc error?
+	if _, ok := status.FromError(err); ok {
+		return err
 	}
 
-	return &orchestrator.VolumeCreateResponse{}, nil
-}
+	if os.IsNotExist(err) || errors.Is(err, syscall.ENOENT) {
+		return status.Error(codes.NotFound, err.Error())
```
djeebus: Make the difference between user facing errors and internal grpc codes clearer. Something like:

```golang
		e := status.New(codes.NotFound, err.Error())
		return e.WithDetails(
			UserFacingError{Code: "file_not_found", Message: "File %s not found."},
		)
```

packages/orchestrator/internal/volumes/stat.go
```diff
@@ -0,0 +1,84 @@
...
+		fType = orchestrator.FileType_FILE_TYPE_SYMLINK
+	case info.IsDir():
+		fType = orchestrator.FileType_FILE_TYPE_DIRECTORY
+	default:
+		fType = orchestrator.FileType_FILE_TYPE_FILE
+	}
+
+	if !strings.HasPrefix(volumeRelPath, "/") {
+		volumeRelPath = "/" + volumeRelPath
+	}
```
djeebus: Use similar process as the code here: packages/envd/internal/services/filesystem/utils.go

djeebus: And explicitly: 
mode = `0o777` (permissions in octal)
permissions = `rwxrwxrwx` (permissions as a string)

packages/orchestrator/orchestrator.proto
```diff
@@ -149,23 +149,173 @@ service SandboxService {
...
+  optional uint32 mode = 3;
+  optional uint32 uid = 4;
+  optional uint32 gid = 5;
+}
+
+message VolumeFileUpdateResponse {
+  EntryInfo entry = 1;
+}
+
 service VolumeService {
```
djeebus: Move `VolumeService` into its own file.

packages/orchestrator/internal/volumes/service.go
```diff
@@ -13,68 +16,94 @@ import (
...
+func BuildVolumePathParts(teamID, volumeID string) []string {
+	return []string{
+		fmt.Sprintf("team-%s", teamID),
+		fmt.Sprintf("vol-%s", volumeID),
+	}
 }
 
-func (v *VolumeService) buildVolumePath(volumeType, teamID, volumeID string) (string, *status.Status) {
-	volPath, ok := v.config.PersistentVolumeMounts[volumeType]
+func (s *Service) buildVolumePath(volume *orchestrator.VolumeInfo, subPath string) (string, error) {
```
djeebus: document why this is secure, and how.

packages/orchestrator/internal/volumes/volume_create.go
```diff
@@ -0,0 +1,33 @@
...
+	}()
+
+	volumePath, err := s.buildVolumePath(request.GetVolume(), "")
+	if err != nil {
+		return nil, fmt.Errorf("failed to build volume path: %w", err)
+	}
+
+	logger.L().Info(ctx, "creating volume",
+		zap.String("path", volumePath),
+	)
```
djeebus: swap logs for trace attr/event

packages/api/internal/handlers/sandboxes_list.go
```diff
@@ -333,5 +333,5 @@ func convertFromDBMountsToAPIMounts(mounts []*dbtypes.SandboxVolumeMountConfig)
 		})
 	}
 
-	return results
+	return &results
```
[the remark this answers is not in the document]
djeebus: This is correct. We want to always return `volumeMounts: []`, but we want the client to know that it's okay if they don't receive them. This is done for backwards sdk compatibility.

iac/provider-gcp/variables.tf
```diff
@@ -598,3 +598,26 @@ variable "loki_use_v13_schema_from" {
...
+
+  type = map(object({
+    allow_deletion = optional(bool)
+    tier           = string
+    location       = optional(string)
+    capacity_gb    = number
+  }))
+
+  default = {}
+}
```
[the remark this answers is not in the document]
djeebus: This will get caught at runtime, I'm okay with that.

iac/provider-gcp/nomad-cluster/scripts/start-client.sh
```diff
@@ -100,6 +100,12 @@ mount "${NFS_MOUNT_PATH}"
 mkdir -p "${NFS_MOUNT_PATH}/${NFS_MOUNT_SUBDIR}" && chmod +w "${NFS_MOUNT_PATH}/${NFS_MOUNT_SUBDIR}"
 %{ endif }
 
+%{ for name, config in PERSISTENT_VOLUME_TYPES }
```
dobrac, line 103: how does this work with adding a new one while we don't have yet rolled all orchestrators?

iac/provider-gcp/nomad-cluster/main.tf
```diff
@@ -7,9 +7,9 @@ locals {
   nfs_mount_path   = "/orchestrator/shared-store"
   nfs_mount_subdir = "chunks-cache"
   nfs_mount_opts = join(",", [ // for more docs, see https://linux.die.net/man/5/nfs
-    format("nfsvers=%s", var.filestore_cache_enabled ? module.filestore[0].nfs_version == "NFS_V3" ? "3" : "4" : ""),
+    format("nfsvers=%s", var.filestore_cache_enabled ? module.filestore[0].nfs_version : ""),
```
dobrac, line 10: previsouly this was 3 or 4, now its 3 or 4.1, is that ok?

packages/api/internal/handlers/volume_create.go
```diff
@@ -106,6 +107,13 @@ func (a *APIStore) PostVolumes(c *gin.Context) {
 	}
 
 	if err := a.createVolume(ctx, clusterID, volume); err != nil {
+		if errors.Is(err, ErrClusterNotFound) {
+			a.sendAPIStoreError(c, http.StatusInternalServerError, "Cluster not found")
```
dobrac: is this an internal server error?

packages/envd/internal/services/filesystem/utils_test.go
```diff
@@ -60,250 +54,3 @@ func TestIsPathOnNetworkMount_FuseMount(t *testing.T) {
 	require.NoError(t, err)
 	assert.False(t, isNetworkSource, "source directory should not be detected as network filesystem")
 }
-
-func TestGetEntryType(t *testing.T) {
```
dobrac, line 64: should there be entryInfo tests instead?

packages/shared/pkg/logger/sandbox/logger.go
```diff
@@ -43,7 +43,7 @@ func NewLogger(ctx context.Context, loggerProvider log.LoggerProvider, config Sa
 		ServiceName:       config.ServiceName,
 		IsInternal:        config.IsInternal,
 		IsDebug:           true,
-		DisableStacktrace: !config.IsInternal,
+		DisableStacktrace: true,
```
dobrac: we don't want to disable them globally. But, probably keep them only for errors

packages/shared/pkg/telemetry/tracing.go
```diff
@@ -48,7 +48,6 @@ func ReportError(ctx context.Context, message string, err error, attrs ...attrib
 	logger.L().With(attributesToZapFields(attrs...)...).Warn(ctx, message, zap.Error(err))
 
 	span.RecordError(fmt.Errorf("%s: %w", message, err),
-		trace.WithStackTrace(true),
```
dobrac, line 51: why removing it here?

packages/orchestrator/internal/volumes/service.go
```diff
@@ -1,45 +1,84 @@
...
 	"github.com/e2b-dev/infra/packages/orchestrator/internal/cfg"
+	"github.com/e2b-dev/infra/packages/shared/pkg/filesystem"
 	"github.com/e2b-dev/infra/packages/shared/pkg/grpc/orchestrator"
 )
 
-type VolumeService struct {
+const (
+	defaultDirMode  uint32 = 0o777
+	defaultFileMode uint32 = 0o666
+	defaultOwnerID  uint32 = 1000
```
dobrac: should we do the custom (non 1000/1001) owner/group id by default?

packages/api/internal/handlers/volume_token.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/google/uuid"
+
+	"github.com/e2b-dev/infra/packages/api/internal/api"
+	"github.com/e2b-dev/infra/packages/shared/pkg/clusters"
+	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
+)
+
+const gracePeriod = time.Minute
+
+func (a *APIStore) PostVolumesVolumeIDToken(c *gin.Context, volumeID api.VolumeID) {
```
dobrac: is this or the token refresh?

iac/provider-gcp/variables.tf
```diff
@@ -603,3 +609,38 @@ variable "loki_use_v13_schema_from" {
...
+
+variable "default_persistent_volume_type" {
+  type    = string
+  default = ""
+}
+
+variable "network_name" {
```
dobrac, line 649: what is this used for now?

spec/openapi.yml
```diff
@@ -3287,3 +3289,27 @@ paths:
           $ref: "#/components/responses/404"
         "500":
           $ref: "#/components/responses/500"
+
+  /volumes/{volumeID}/token:
+    post:
+      description: Create a token for future requests
```
dobrac: this should be probably part of the get/create volume request

packages/orchestrator/internal/nfsproxy/proxy.go
```diff
@@ -80,7 +81,7 @@ func getPrefixFromSandbox(sandboxes *sandbox.Map, filesystemsByType map[string]b
 			return nil, "", fmt.Errorf("failed to mount %q (%s): %w", volumeName, volumeMount.Type, ErrVolumeTypeNotSupported)
 		}
 
-		prefixParts := []string{sbx.Metadata.Runtime.TeamID, volumeName}
+		prefixParts := volumes.BuildVolumePathParts(sbx.Metadata.Runtime.TeamID, volumeMount.ID)
```
dobrac: with this usage, this is not true anymore:

```
	// TeamID optional, used only for logging
	TeamID string
```

I think we might want to move the teamID out of the runtime metadata as they're meant to be kinda "non-required". Or, just making the teamID required and make sure its always present

packages/orchestrator/internal/nfsproxy/e2e_test.go
```diff
@@ -184,7 +198,7 @@ func TestIntegrationTest(t *testing.T) {
 
 	// run the actual test
 	code, out, err := testCtr.Exec(t.Context(),
-		[]string{"/e2e_test.sh"},
+		[]string{"/e2e_start.sh"},
```
dobrac, line 201: shouldnt this be e2e_test.sh?

packages/orchestrator/internal/volumes/dir_create.go
```diff
@@ -0,0 +1,86 @@
...
+
+	var fn makeDir
+	if request.GetCreateParents() {
+		fn = os.MkdirAll
+	} else {
+		fn = os.Mkdir
+	}
+	if err := fn(fullPath, os.FileMode(mode)); err != nil {
+		if os.IsNotExist(err) {
+			return nil, newAPIError(ctx, codes.NotFound, "path_not_found", "failed to mkdir: parent of %q not found.", fullPath)
```
dobrac: lets centralize the `devCode`s under constants in a separate file?

packages/orchestrator/internal/volumes/dir_create.go
```diff
@@ -0,0 +1,86 @@
...
+			return nil, newAPIError(ctx, codes.NotFound, "path_not_found", "failed to chmod: %q not found.", fullPath)
+		}
+
+		return nil, fmt.Errorf("failed to set directory mode: %w", err)
+	}
+
+	entry, err := toEntryFromPath(fullPath, request.GetPath())
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil, newAPIError(ctx, codes.NotFound, "path_not_found", "failed to stat: %q not found.", fullPath)
```
dobrac: isn't this internal server error?

packages/orchestrator/internal/volumes/dir_delete.go
```diff
@@ -0,0 +1,64 @@
...
+func (s *Service) DeleteDir(ctx context.Context, request *orchestrator.VolumeDirDeleteRequest) (r *orchestrator.VolumeDirDeleteResponse, err error) {
+	ctx, span := tracer.Start(ctx, "delete directory in volume")
+	defer func() {
+		setSpanStatus(span, err)
+		span.End()
+	}()
+
+	relPath := request.GetPath()
+	if relPath == "" || relPath == "/" || relPath == "." {
+		return nil, newAPIError(ctx, codes.InvalidArgument, "empty_path", "path cannot be empty")
```
dobrac: this is more of an invalid path, rather than empty

packages/orchestrator/internal/volumes/dir_delete.go
```diff
@@ -0,0 +1,64 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if filepath.Clean(rootPath) == filepath.Clean(fullPath) {
+		return nil, newAPIError(ctx, codes.InvalidArgument, "cannot_delete_root", "cannot delete root directory")
+	}
+
+	var fn removeFunc
+	if request.GetRecursive() {
```
dobrac: is there any actual usage of having to define the recursive explicitly?

packages/orchestrator/internal/volumes/dir_list.go
```diff
@@ -0,0 +1,58 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	span.AddEvent("listing directory", trace.WithAttributes(
+		attribute.String("path", fullPath),
+	))
+
+	items, err := os.ReadDir(fullPath)
+	if err != nil { // todo: better error handling
```
dobrac: todo?

packages/orchestrator/internal/volumes/dir_list.go
```diff
@@ -0,0 +1,58 @@
...
+	var results []*orchestrator.VolumeDirectoryItem
+	for _, item := range items {
+		info, err := item.Info()
+		if err != nil {
+			return nil, fmt.Errorf("failed to get info for item %q: %w", item.Name(), err)
+		}
+
+		absPath := filepath.Join(fullPath, item.Name())
+		relPath := filepath.Join(request.GetPath(), item.Name())
+		entry := toEntryFromOSInfo(absPath, relPath, info)
```
dobrac: double checking, can you escape here out of your nfs folder?

packages/orchestrator/internal/volumes/file_create.go
```diff
@@ -0,0 +1,131 @@
...
+var ErrUnexpectedStart = errors.New("unexpected start message")
+
+func (s *Service) CreateFile(server orchestrator.VolumeService_CreateFileServer) (err error) {
+	ctx, span := tracer.Start(server.Context(), "create file in volume")
+	defer func() {
+		setSpanStatus(span, err)
+		span.End()
+	}()
+
+	req, err := server.Recv()
```
dobrac, line 29: does the client streaming actually work for the Python SDK?

packages/orchestrator/internal/volumes/file_create.go
```diff
@@ -0,0 +1,131 @@
...
+
+	span.AddEvent("creating file", trace.WithAttributes(
+		attribute.String("path", fullPath),
+		attribute.Int64("uid", int64(uid)),
+		attribute.Int64("gid", int64(gid)),
+		attribute.Int64("mode", int64(mode)),
+		attribute.Bool("force", start.GetForce()),
+	))
+
+	if start.GetForce() {
```
dobrac, line 55: yes yes, get force :strong:

packages/orchestrator/internal/volumes/file_create.go
```diff
@@ -0,0 +1,131 @@
...
+			if deleteErr != nil {
+				logger.L().Error(ctx, "failed to delete file after error", zap.Error(deleteErr))
+			}
+		}
+	}()
+
+	for {
+		req, err := server.Recv()
+		if err != nil {
+			return fmt.Errorf("failed to receive chunk: %w", err)
```
dobrac, line 84: should this delete the file it the file already existed before?

packages/orchestrator/internal/volumes/service.go
```diff
@@ -1,45 +1,84 @@
...
+	volumeID := volume.GetVolumeId()
 	if !tryParseUUID(volumeID) {
-		return "", status.Newf(codes.InvalidArgument, "invalid volume ID %q", volumeID)
+		return "", status.Newf(codes.InvalidArgument, "invalid volume ID %q", volumeID).Err()
 	}
 
-	volumePath := filepath.Join(volPath, teamID, volumeID)
+	volumeParts := append([]string{volTypePath}, BuildVolumePathParts(teamID, volumeID)...)
+	volumePath := filepath.Join(volumeParts...)
+	if subPath != "" {
```
dobrac: small nit, might be worth doing a const for "ROOT" as ""

packages/orchestrator/internal/volumes/service.go
```diff
@@ -1,45 +1,84 @@
...
+		subPath = strings.TrimPrefix(subPath, "/")
+		subPath = filepath.Clean(subPath)
+		fullPath := filepath.Join(volumePath, subPath)
+
+		result, err := filepath.Rel(volumePath, fullPath)
+		if err != nil || strings.HasPrefix(result, "..") {
```
dobrac: how did you come up that this will happen and its enough `strings.HasPrefix(result, "..")`?

packages/orchestrator/internal/nfsproxy/jailed/fs.go
```diff
@@ -68,19 +68,43 @@ func (j jailedFS) Remove(filename string) error {
 }
 
 func (j jailedFS) Join(elem ...string) string {
```
dobrac: is this method actually properly used? Shouldn't the cleanup be used also for other methods? (like symlink, rename)

iac/provider-gcp/nomad-cluster/scripts/start-client.sh
djeebus: If only some of the orchestrators have the new volume type, but we've already updated launchdarkly / default persistent volume type to point to the new one, then the orchestrator will return an error. We'll catch that error and retry on a new orchestrator. Eventually we'll either find one that has the new volume type, or we'll give up completely.

In practice, when deploying a new volume type, it should be a multi step process:
1. add the volume type to the orchestrator via terraform
2. wait for orchestrator roll out
3. use launch darkly to enable the new volume type for one user/sandbox/template/etc to ensure that it works
4. change the default, via launch darkly, either via progressive rollout or all at once
5. change the env var to the new value

iac/provider-gcp/nomad-cluster/main.tf
djeebus: Yeah, turns out "4" was completely wrong. it should always be "4.1". GCP only supports 4.1, not 4.

packages/api/internal/handlers/volume_create.go
djeebus: The cluster ID isn't something the customer provides us with, that's something that we already know about a team. It's not a 4xx error (they didn't give us the cluster ID, so it can't be their fault), so it's our fault. It might be a 503 (the request cannot be handled)?

packages/envd/internal/services/filesystem/utils_test.go
djeebus: It doesn't really do anything anymore, although `getFileOwnership` still does some meaningful work. I'll add some tests around the helper functions.

packages/api/internal/handlers/volume_create.go
dobrac: 503 sounds better than just plain internal server error. Its a bit config error in a way

packages/shared/pkg/telemetry/tracing.go
djeebus: Because it makes local development impossibly noisy? I reverted it though.

packages/orchestrator/internal/volumes/service.go
djeebus: By default, the files should be readable/writable by the default user in the sandbox, right? I would assume that means they should line up with the user's uid/gid. Is that not the case?

packages/api/internal/handlers/volume_token.go
djeebus: This is the token. We don't have a refresh endpoint (you just get a new token with your original access token). We can make this flow more complicated though, if there's value in it.

iac/provider-gcp/variables.tf
djeebus: We have (currently) 8 places in our code where we define a `variable "network_name"`, but there was actually no way to unify those, no global way to define "don't use `default`, use this instead". This just ties those together to a single place.

packages/orchestrator/internal/nfsproxy/proxy.go
djeebus: I don't see any reason why TeamID *shouldn't* be required metadata here. Would we ever want a sandbox to be shared across teams? If not, we know the team id when we create the sandbox. I can make it consistently required, if I'm not wrong about the above.

packages/orchestrator/internal/nfsproxy/e2e_test.go
djeebus: `e2e_start.sh` calls `e2e_test.sh` twice - once against a local ext4 file system (iirc) and once against the nfs proxy, to ensure that the test is valid.

packages/orchestrator/internal/volumes/dir_create.go
djeebus: I added it to the proto contract so we get constants and type safety for free

packages/orchestrator/internal/volumes/dir_create.go
djeebus: Probably not (but possibly?):

- `mkdir /a/b/c`, but only `/a` exists: path_not_found b/c `b` is missing and we didn't say `create_parents=true`
- `mkdir /a`, but for some reason the volume root `/mnt/volume_types/default/$TEAM_ID/$VOL_ID` is missing: should be internal server error

I'll split those up.

packages/orchestrator/internal/volumes/dir_delete.go
djeebus: I was copying the linux usage, which separates the two uses:
- `rmdir $path` - only delete the directory, and only if it's empty. this is our `recursive=false`
- `rm $path -r` - delete the directory and everything under it. `recursive=true`
This gives users a safe option and a dangerous option. We can combine them if there's no value here.

packages/orchestrator/internal/volumes/dir_list.go
djeebus: I put that originally, but I think this implementation is just fine. removing the todo!

packages/orchestrator/internal/volumes/dir_list.go
djeebus: Both of these must be valid UUID, so no escape possible. I'll change the code to make that clear as well.

packages/orchestrator/internal/volumes/file_create.go
djeebus: This is only an internal API, the client still gets a simple http request:

```mermaid
graph LR
  client --http request--> api
  api --grpc stream--> orchestrator
```

packages/orchestrator/internal/volumes/file_create.go
djeebus: as discussed, going to stop removing the file on error

packages/orchestrator/internal/volumes/service.go
djeebus: I think that'd give a false idea; root could be `""`, `"/"`, or `"."`, or a few other things. This is just "if they pass a path"

packages/orchestrator/internal/volumes/service.go
djeebus: `filepath.Clean` will resolve all ".." args, leaving only those that cause it to go above the current path: the ones at the beginning:

```
func main() {
	path := "../../a/../../../b/c/d/e/../../"
	result := filepath.Clean(path)
	println(result)   // prints ../../../../b/c
}
```

```
func main() {
	path := "./a/b/../../../d/e/f/g/../../"
	result := filepath.Clean(path)
	println(result)   // prints ../d/e
}
```
Therefor, all we need to do is check to see if the path is strying to break out is to check if it starts with ".."

packages/orchestrator/internal/nfsproxy/jailed/fs.go
djeebus: The `willscott/go-nfs` library calls `fs.Join` on every nfs request before doing anything with the path.

... that said, we're relying on his library A LOT, so we might want to allocate some time to do a security review of each of the methods, make sure we're not missing anything.

... maybe pay someone else to do it.

packages/orchestrator/internal/volumes/service.go
dobrac: (we've discussed IRL) the "default" here is not really the "default" in a sandbox, which might be confusing

packages/api/internal/handlers/volume_token.go
dobrac: I'm more for returning the token in the getVolume/createVolume requests to save the query

packages/orchestrator/internal/nfsproxy/proxy.go
dobrac: Its just the whole nature of the runtime metadata meaning in this context: they're optional, used only for observability, metrics, logs

packages/orchestrator/internal/volumes/dir_delete.go
dobrac: I see, I've never understood why is the option there in the first place in the linux commands. I would combine them

packages/orchestrator/internal/volumes/file_create.go
dobrac: ok

packages/orchestrator/internal/volumes/service.go
dobrac: hmm, should it be nil then? Or some other token? I was thinking that because in a lot of places we pass "" and say: this is root

packages/orchestrator/internal/nfsproxy/jailed/fs.go
dobrac: hmm, if thats okay work, I would move this validation to our code, rather then relying on the go-nfs

packages/orchestrator/internal/volumes/service.go
djeebus: Will add a "volume" user/group and by default add files to that group

packages/orchestrator/internal/nfsproxy/proxy.go
djeebus: TODO: Remove team ID from path, store it in DB, pass subpath around via API

packages/orchestrator/internal/volumes/dir_delete.go
djeebus: TODO: remove recursive flag, remove everything recursively always

packages/orchestrator/internal/volumes/dir_list.go
djeebus: TODO: build relative path from result of buildVolumePath, not `request.GetPath()`, which should never be used past `buildVolumePath`

djeebus: TODO: maybe process path in middleware before it gets to the handlers? Think on it.

packages/orchestrator/internal/volumes/service.go
djeebus: TODO: add another method to build volume root, not subpath, to avoid the `""`

packages/orchestrator/internal/nfsproxy/jailed/fs.go
djeebus: Have to rely on go-nfs :cry:

packages/orchestrator/internal/volumes/service.go
djeebus: I've set it to 9090 (just an arbitrarily high number), and we can add the user to our base image at a later date.

packages/orchestrator/internal/volumes/service.go
```diff
@@ -2,79 +2,242 @@ package volumes
...
+
+func (v volumePaths) isRoot() bool {
+	return filepath.Clean(v.BasePath) == filepath.Clean(v.FullPath)
+}
+
+func (s *Service) buildPaths(request volumePathRequest) (volumePaths, error) {
+	volume := request.GetVolume()
+
+	volumeType := volume.GetVolumeType()
+	volTypePath, ok := s.config.PersistentVolumeMounts[volumeType]
```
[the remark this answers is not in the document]
djeebus: Not true, `GetVolumeType()` checks itself for nil and returns an "" instead of panicing.

packages/orchestrator/internal/volumes/dir_list.go
```diff
@@ -0,0 +1,60 @@
...
+	))
+
+	items, err := os.ReadDir(paths.HostFullPath)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil, newAPIError(ctx, codes.NotFound, http.StatusNotFound, orchestrator.UserErrorCode_PATH_NOT_FOUND, "failed to read: %q not found.", request.GetPath())
+		}
+
+		return nil, fmt.Errorf("failed to read directory %q: %w", paths.HostFullPath, err)
+	}
```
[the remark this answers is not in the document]
djeebus: This is fine, as we aren't displaying grpc errors to users, only the content of `UserError`. the rest is for us to use internally.

packages/orchestrator/internal/volumes/dir_list.go
```diff
@@ -0,0 +1,60 @@
...
+		attribute.String("path", paths.HostFullPath),
+	))
+
+	items, err := os.ReadDir(paths.HostFullPath)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil, newAPIError(ctx, codes.NotFound, http.StatusNotFound, orchestrator.UserErrorCode_PATH_NOT_FOUND, "failed to read: %q not found.", request.GetPath())
+		}
+
+		return nil, fmt.Errorf("failed to read directory %q: %w", paths.HostFullPath, err)
```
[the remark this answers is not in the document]
djeebus: This is fine, we keep this internal.

packages/api/internal/cfg/model.go
```diff
@@ -55,23 +63,27 @@ type Config struct {
...
+	SandboxStorageBackend string `env:"SANDBOX_STORAGE_BACKEND" envDefault:"memory"`
+
+	DomainName string `env:"DOMAIN_NAME" envDefault:""`
 }
 
 type VolumesTokenConfig struct {
 	Issuer        string            `env:"VOLUME_TOKEN_ISSUER,required"`
-	SigningMethod jwt.SigningMethod `env:"VOLUME_TOKEN_SIGNING_METHOD" envDefault:"HS256"`
+	SigningMethod jwt.SigningMethod `env:"VOLUME_TOKEN_SIGNING_METHOD"       envDefault:"HS256"`
 	SigningKey    []byte            `env:"VOLUME_TOKEN_SIGNING_KEY,required"`
```
dobrac: how would we handle some rotation?

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -214,19 +219,28 @@ func (s *Server) Create(ctx context.Context, req *orchestrator.SandboxCreateRequ
...
 }
 
-func createVolumeMountModelsFromAPI(volumeMounts []*orchestrator.SandboxVolumeMount) []sandbox.VolumeMountConfig {
+func createVolumeMountModelsFromAPI(volumeMounts []*orchestrator.SandboxVolumeMount) ([]sandbox.VolumeMountConfig, error) {
+	var errs []error
+
 	results := make([]sandbox.VolumeMountConfig, 0, len(volumeMounts))
 
 	for _, v := range volumeMounts {
+		volumeID, err := uuid.Parse(v.GetId())
```
dobrac, line 234: any general approach behind doing uuid.Parse vs `tryParseUUID`?

packages/orchestrator/internal/volumes/dir_list.go
```diff
@@ -20,34 +20,36 @@ func (s *Service) ListDir(ctx context.Context, request *orchestrator.VolumeDirLi
...
-	if err != nil { // todo: better error handling
+	items, err := os.ReadDir(paths.HostFullPath)
+	if err != nil {
 		if os.IsNotExist(err) {
-			return nil, newAPIError(ctx, codes.NotFound, "path_not_found", "failed to read: %q not found.", fullPath)
+			return nil, newAPIError(ctx, codes.NotFound, http.StatusNotFound, orchestrator.UserErrorCode_PATH_NOT_FOUND, "failed to read: %q not found.", request.GetPath())
 		}
 
-		return nil, fmt.Errorf("failed to read directory %q: %w", fullPath, err)
+		return nil, fmt.Errorf("failed to read directory %q: %w", paths.HostFullPath, err)
```
dobrac, line 42: where should we return the hostfullPath in errors and where request.GetPath()?

packages/orchestrator/internal/volumes/file_delete.go
```diff
@@ -18,23 +19,23 @@ func (s *Service) DeleteFile(ctx context.Context, request *orchestrator.VolumeFi
...
+		return nil, newAPIError(ctx, codes.InvalidArgument, http.StatusBadRequest, orchestrator.UserErrorCode_CANNOT_DELETE_ROOT, "cannot delete root directory")
+	}
+
 	span.AddEvent("deleting file", trace.WithAttributes(
-		attribute.String("path", fullPath),
+		attribute.String("path", paths.HostFullPath),
 	))
 
-	if err := os.Remove(fullPath); err != nil {
+	if err := os.Remove(paths.HostFullPath); err != nil {
```
dobrac, line 36: do we verify somewhere that `DeleteFile` references a `file`?

packages/orchestrator/internal/volumes/service.go
```diff
@@ -41,68 +44,163 @@ func New(config cfg.Config) *Service {
...
+}
+
+var _ volumePathRequest = (*volumeRequest)(nil)
+
+func (v volumeRequest) GetVolume() *orchestrator.VolumeInfo {
+	return v.request.GetVolume()
+}
+
+func (v volumeRequest) GetPath() string {
+	return ""
```
dobrac, line 71: always empty?

packages/orchestrator/internal/volumes/service.go
```diff
@@ -41,68 +44,163 @@ func New(config cfg.Config) *Service {
...
+		return false
 	}
 
-	return volumePath, nil
+	return true
 }
 
-func tryParseUUID(id string) bool {
-	_, err := uuid.Parse(id)
+func tryParseUUID(id string) (uuid.UUID, bool) {
```
dobrac: this function now exists in two places

packages/orchestrator/internal/volumes/stat.go
```diff
@@ -19,34 +20,25 @@ func (s *Service) Stat(ctx context.Context, request *orchestrator.StatRequest) (
...
-		attribute.String("path", fullPath),
+		attribute.String("path", paths.HostFullPath),
 	))
 
-	info, err := filesystem.GetEntryFromPath(fullPath)
+	info, err := filesystem.GetEntryFromPath(paths.HostFullPath)
 	if err != nil {
 		if os.IsNotExist(err) {
-			return nil, newAPIError(ctx, codes.NotFound, "path_not_found", "failed to stat: %q not found.", fullPath)
+			return nil, newAPIError(ctx, codes.NotFound, http.StatusBadRequest, orchestrator.UserErrorCode_PATH_NOT_FOUND, "failed to stat: %q not found.", request.GetPath())
```
dobrac, line 35: general QQ, why are we using the request.GetPath and not the ClientPath?

packages/orchestrator/internal/nfsproxy/jailed/fs.go
```diff
@@ -14,21 +14,47 @@ type jailedFS struct {
...
+			break
+		}
+		if after, ok := strings.CutPrefix(s, "../"); ok {
+			s = after
+
+			continue
+		}
+
+		break
+	}
```
dobrac: How about using this instead?

```
// Force the path to stay within the jail by cleaning it as an absolute path,
// then stripping the leading slash before joining with the prefix.
s := filepath.Clean("/" + p)
return filepath.Join(j.prefix, strings.TrimPrefix(s, "/"))
```

packages/api/internal/cfg/model.go
djeebus: Simplest way (which requires a little more work):
1. add some kind of version or timestamp to the signing key. For example: "2026-v1=somesup3rs3cur3k3y"
2. add the "2026-v1" to the header of the jwt
3. the validating side reads the metadata and uses the appropriate key
Should we add that now?

packages/orchestrator/internal/server/sandboxes.go
djeebus: Nothing intentional, just a matter of how the code looked at the time. I can centralize it somewhere.

packages/orchestrator/internal/volumes/dir_list.go
djeebus: - UserError messages should show the original request path, to avoid revealing how we're cleaning up the path before using it. this makes it more difficult for them to find path traversal weaknesses
- All other errors should not be returned to clients, and can have the host path to make it easier to diagnose issues

packages/orchestrator/internal/volumes/file_delete.go
djeebus: At some point in the near future (or maybe now?) we're planning on unifying the behavior (delete file / delete directory are functionally the same at this point) so the distinction does not matter.

packages/orchestrator/internal/volumes/service.go
djeebus: `volumeRequest` represents a request to build the path to a volume without a subpath. I unified the behavior to always expect a path on the request object (for example `path, err := buildPathsFromRequest(request)`), but for the few requests that don't have a path (`CreateVolume` and `DeleteVolume`) they needed a wrapper that returns a dummy string.

packages/orchestrator/internal/volumes/service.go
djeebus: added to //todo in the description to unify this.

packages/orchestrator/internal/volumes/stat.go
djeebus: Mentioned above, but for more details:

If I'm a hacker, and I send "/a/b/../../../etc/shadow", and we clean it up (with a bug) and return an error that says "/mnt/volumes/etc/shadow does not exist", now the attacker knows they just need to add two more ".." to the path. If we return the original path and say "/a/b/../../../etc/shadow does not exist", we've given them no extra information.

packages/orchestrator/internal/nfsproxy/jailed/fs.go
djeebus: The unit tests seem to pass, I'll merge this and we'll see what happens :crossed_fingers:  Definitely easier to read!

packages/orchestrator/internal/volumes/service.go
dobrac: ok, thx

packages/api/internal/cfg/model.go
djeebus: Adding now.

packages/api/.env.local
```diff
@@ -15,3 +15,5 @@ PERSISTENT_VOLUME_MOUNTS=test-volume-type:./.data/test-volume
 POSTGRES_CONNECTION_STRING=postgres://postgres:postgres@localhost:5432/postgres?sslmode=disable
 REDIS_URL=localhost:6379
 SANDBOX_ACCESS_TOKEN_HASH_SEED=--sandbox-access-token-hash-seed--
+VOLUME_TOKEN_ISSUER=local.e2b.dev
+VOLUME_TOKEN_SIGNING_KEY=5M6IOYJ/y4ukqjGJbWE/SEEjr38O28E+
```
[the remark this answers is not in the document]
djeebus: This is only for testing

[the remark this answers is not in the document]
djeebus: This is not a real token; only prepopulating for local dev convenience.

packages/api/.env.local
```diff
@@ -15,3 +15,7 @@ PERSISTENT_VOLUME_MOUNTS=test-volume-type:./.data/test-volume
 POSTGRES_CONNECTION_STRING=postgres://postgres:postgres@localhost:5432/postgres?sslmode=disable
 REDIS_URL=localhost:6379
 SANDBOX_ACCESS_TOKEN_HASH_SEED=--sandbox-access-token-hash-seed--
+VOLUME_TOKEN_ISSUER=local.e2b.dev
+VOLUME_TOKEN_SIGNING_METHOD=ES256
+VOLUME_TOKEN_SIGNING_KEY=ECDSA:LS0tLS1CRUdJTiBQUklWQVRFIEtFWS0tLS0tCk1JR0hBZ0VBTUJNR0J5cUdTTTQ5QWdFR0NDcUdTTTQ5QXdFSEJHMHdhd0lCQVFRZ2xwTHZaNnZmSGNYanBvMFYKUzVncW5JM21kdTNBSGJ0UnhmcU1CNWViT3I2aFJBTkNBQVQ3MHBiSE5NY015eFRxTVJiRjN0bXNwemZQTDIwQQpldk1lWGN5MFVDMkRiRXJ2bXBuT2hDamxFT2l4TXVDRzdUekJDdllLNUM0bGdxY0FjVnZVclBwZgotLS0tLUVORCBQUklWQVRFIEtFWS0tLS0tCg==
+VOLUME_TOKEN_SIGNING_KEY_NAME=local-dev-2026-03-03
```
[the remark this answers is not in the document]
djeebus: this is random garbage, only used for local dev work

packages/api/.env.local
```diff
@@ -15,3 +15,7 @@ PERSISTENT_VOLUME_MOUNTS=test-volume-type:./.data/test-volume
 POSTGRES_CONNECTION_STRING=postgres://postgres:postgres@localhost:5432/postgres?sslmode=disable
 REDIS_URL=localhost:6379
 SANDBOX_ACCESS_TOKEN_HASH_SEED=--sandbox-access-token-hash-seed--
+VOLUME_TOKEN_ISSUER=local.e2b.dev
+VOLUME_TOKEN_SIGNING_METHOD=ES256
+VOLUME_TOKEN_SIGNING_KEY=ECDSA:LS0tLS1CRUdJTiBQUklWQVRFIEtFWS0tLS0tCk1JR0hBZ0VBTUJNR0J5cUdTTTQ5QWdFR0NDcUdTTTQ5QXdFSEJHMHdhd0lCQVFRZ2xwTHZaNnZmSGNYanBvMFYKUzVncW5JM21kdTNBSGJ0UnhmcU1CNWViT3I2aFJBTkNBQVQ3MHBiSE5NY015eFRxTVJiRjN0bXNwemZQTDIwQQpldk1lWGN5MFVDMkRiRXJ2bXBuT2hDamxFT2l4TXVDRzdUekJDdllLNUM0bGdxY0FjVnZVclBwZgotLS0tLUVORCBQUklWQVRFIEtFWS0tLS0tCg==
```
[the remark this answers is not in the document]
djeebus: this is only for local dev

packages/api/Makefile
```diff
@@ -84,3 +86,25 @@ test:
...
+.PHONY: generate-key
+generate-local-key:
+	# Generate a new ECDSA key pair (P-256)
+	openssl ecparam -name prime256v1 -genkey -noout -out private.pem
+	openssl ec -in private.pem -pubout -out public.pem
+	# Update .env.local with the new private key and name stamped with today's date
+	KEY=$$(base64 -w0 private.pem); \
+	NAME=local-dev-$$(date +%F); \
+	if grep -q '^VOLUME_TOKEN_SIGNING_KEY=' .env.local; then \
+		sed -i "s|^VOLUME_TOKEN_SIGNING_KEY=.*|VOLUME_TOKEN_SIGNING_KEY=ECDSA:$${KEY}|" .env.local; \
```
dobrac, line 99: small nit, the .env.local might now be the .env file you're using rn

djeebus: This is *only* used for local dev, not generating keys for deployment. `.env.local` is in `/packages/api`, and isn't the configuration file for deployment.

packages/orchestrator/internal/volumes/service.go
```diff
@@ -2,79 +2,233 @@ package volumes
...
+		return volumePaths{}, status.Errorf(codes.InvalidArgument, "invalid path: traversal detected")
 	}
 
-	if !tryParseUUID(volumeID) {
-		return "", status.Newf(codes.InvalidArgument, "invalid volume ID %q", volumeID)
+	return volumePaths{
+		HostVolumePath: basePath,
+		ClientPath:     filepath.Join("/", relPath), // Ensures leading slash
+		HostFullPath:   fullPath,
+	}, nil
```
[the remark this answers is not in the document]
djeebus: This isn't really true, we use the new paths.

packages/api/internal/handlers/volume_create.go
```diff
@@ -125,9 +133,18 @@ func (a *APIStore) PostVolumes(c *gin.Context) {
...
 	}
 
-	result := api.Volume{
+	token, err := generateVolumeContentToken(a.config.VolumesToken, volume, team)
+	if err != nil {
+		a.sendAPIStoreError(c, http.StatusInternalServerError, "Failed to generate volume content token")
+		telemetry.ReportCriticalError(ctx, "Failed to generate volume content token", err)
+
+		return
+	}
```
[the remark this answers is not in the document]
djeebus: Unfortunate but true

packages/orchestrator/internal/volumes/file_create.go
```diff
@@ -0,0 +1,120 @@
...
+			}
+
+			return server.SendAndClose(&orchestrator.VolumeFileCreateResponse{
+				Entry: entry,
+			})
+
+		default:
+			return ErrUnexpectedStart
+		}
+	}
```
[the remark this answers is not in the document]
djeebus: This is acceptable, as it's how linux works in general.

packages/api/internal/handlers/volume_util.go
```diff
@@ -78,17 +94,90 @@ func (a *APIStore) executeOnOrchestrator(
...
+			logger.L().Warn(clientCtx, "failed to make orchestrator call, retrying ... ", zap.Error(err))
+
+			continue
+		}
+
+		return err
+	}
+
+	if receivedUnknownVolumeTypeErrors == len(nodes)-notReadyNodeCount && receivedUnknownVolumeTypeErrors > 0 {
+		return fmt.Errorf("%w: %s", ErrUnknownVolumeType, unknownVolumeType)
```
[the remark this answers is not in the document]
djeebus: a retryable failure implies an unhealthy orchestrator, I'm fine with this.

iac/provider-gcp/nomad/main.tf
```diff
@@ -412,6 +421,7 @@ module "orchestrator" {
   redis_url                    = local.redis_url
   redis_cluster_url            = local.redis_cluster_url
   redis_tls_ca_base64          = trimspace(data.google_secret_manager_secret_version.redis_tls_ca_base64.secret_data)
+  persistent_volume_mounts     = var.persistent_volume_mounts
```
jakubno, line 424: this

PR #1897, opened 2026-02-11
Fix missing default tag when building templates with explicit tags

beran-t:
## Reported issue
https://e2b-team.slack.com/archives/C0ACD647MRR/p1770807025214659

## Summary

- **Bug fix**: When building a template with explicit tags (e.g. `tags: ["v1", "preprod"]`), the `default` tag was not created. This caused `Sandbox.create(name)` to return 404 and `tpl ls` to show zeroed-out metadata, even though the build succeeded.
- **New endpoint**: Added `GET /templates/{templateID}/tags` to list all tags for a template, addressing the feature gap where tags could be assigned/removed but never read.

## Root Cause

In `register_build.go`, the default tag was only added when `len(tags) == 0`. When explicit tags were provided, no `default` tag was created. The DB trigger (`sync_env_build_assignment`) that was supposed to auto-create defaults doesn't fire in the v3 flow because `CreateTemplateBuild` doesn't set `env_id`.

**Fix**: Changed the condition to always include the `default` tag using `slices.Contains`, so every build is resolvable by bare name.

## Changes

| File | Change |
|------|--------|
| `packages/api/internal/template/register_build.go` | Always include `default` tag in build assignments |
| `packages/api/internal/handlers/template_tags.go` | Add `GetTemplatesTemplateIDTags` handler |
| `packages/db/queries/templates/get_template_tags.sql` | New SQL query for listing tags per template |
| `spec/openapi.yml` | Add `GET /templates/{templateID}/tags` endpoint + response schemas |

### DIFF
--- a/tests/integration/internal/api/generated.go
+++ b/tests/integration/internal/api/generated.go
@@ -1092,6 +1092,21 @@ type TemplateStep struct {
 	Type string `json:"type"`
 }
 
+// TemplateTagDetail defines model for TemplateTagDetail.
+type TemplateTagDetail struct {
+	// BuildID Identifier of the build associated with this tag
+	BuildID openapi_types.UUID `json:"buildID"`
+
+	// Tag Tag name
+	Tag string `json:"tag"`
+}
+
+// TemplateTagsResponse defines model for TemplateTagsResponse.
+type TemplateTagsResponse struct {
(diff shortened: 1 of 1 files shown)


dobrac (changes requested):
thats desired functionality

PR #1904, opened 2026-02-12
feat: add GET /templates/{templateID}/tags endpoint

beran-t:
## Summary
- Adds a new `GET /templates/{templateID}/tags` API endpoint that returns all tags assigned to a template, each with its associated build ID and creation timestamp
- Creates the underlying `ListTemplateTags` SQL query and updates the OpenAPI spec with a `TemplateTag` response schema
- Complements the existing `POST /templates/tags` (assign) and `DELETE /templates/tags` (delete) endpoints so the SDK can list tags

### DIFF
--- a/tests/integration/internal/api/generated.go
+++ b/tests/integration/internal/api/generated.go
@@ -6210,6 +6210,7 @@ type GetTemplatesTemplateIDTagsResponse struct {
 	HTTPResponse *http.Response
 	JSON200      *[]TemplateTag
 	JSON401      *N401
+	JSON403      *N403
 	JSON404      *N404
 	JSON500      *N500
 }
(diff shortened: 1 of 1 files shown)


packages/api/internal/handlers/template_tags.go
```diff
@@ -289,3 +289,58 @@ func (a *APIStore) DeleteTemplatesTags(c *gin.Context) {
 
 	c.Status(http.StatusNoContent)
 }
+
+// GetTemplatesTemplateIDTags lists all tags for a template
+func (a *APIStore) GetTemplatesTemplateIDTags(c *gin.Context, templateID api.TemplateID) {
```
dobrac, line 294: I know it's called templateID, but I think it should work also with the template name (namespace/name)

packages/db/queries/templates/list_template_tags.sql
```diff
@@ -0,0 +1,7 @@
+-- name: ListTemplateTags :many
+-- Lists all tag assignments for a given template.
+-- @template_id: the template ID to look up
+SELECT eba.tag, eba.build_id, eba.created_at
```
dobrac: you can have multiple entries for the same tag name for different build ids. The latest one (created_at DESC) is the one in use that should be returned

dobrac (approved):

PR #1909, opened 2026-02-13
Security secure sandbox urls

matthewlouisbrockman:
Check resuming sandbox has prior credentials over http when `allow_public_traffic`:`false` and envd when `private` are set.

Passes the relevant token via gRPC headers to the API which validates before allowing resuming.

### DIFF
--- a/packages/shared/pkg/proxy/proxy_test.go
+++ b/packages/shared/pkg/proxy/proxy_test.go
@@ -197,7 +197,9 @@ func TestProxyRoutesToTargetServer(t *testing.T) {
 
 	proxy, port, err := newTestProxy(t, getDestination)
 	require.NoError(t, err)
-	defer proxy.Close()
+	t.Cleanup(func() {
+		proxy.Close()
+	})
 
 	assert.Equal(t, uint64(0), proxy.TotalPoolConnections())
 	assert.Equal(t, uint64(0), backend.RequestCount())
@@ -223,14 +225,18 @@ func TestProxyResumePermissionDeniedErrorTemplate(t *testing.T) {
 
 	proxy, port, err := newTestProxy(t, getDestination)
 	require.NoError(t, err)
-	defer proxy.Close()
+	t.Cleanup(func() {
+		proxy.Close()
+	})
 
 	t.Run("json for non-browser", func(t *testing.T) {
 		t.Parallel()
 		proxyURL := fmt.Sprintf("http://127.0.0.1:%d/hello", port)
 		resp, err := httpGet(t, proxyURL)
 		require.NoError(t, err)
-		defer resp.Body.Close()
+		t.Cleanup(func() {
+			_ = resp.Body.Close()
+		})
 
 		require.Equal(t, http.StatusForbidden, resp.StatusCode)
 		require.Equal(t, "application/json; charset=utf-8", resp.Header.Get("Content-Type"))
@@ -256,7 +262,9 @@ func TestProxyResumePermissionDeniedErrorTemplate(t *testing.T) {
 
 		resp, err := httpGetWithHeaders(t, proxyURL, headers)
 		require.NoError(t, err)
-		defer resp.Body.Close()
+		t.Cleanup(func() {
+			_ = resp.Body.Close()
+		})
 
 		require.Equal(t, http.StatusForbidden, resp.StatusCode)
 		require.Equal(t, "text/html; charset=utf-8", resp.Header.Get("Content-Type"))


packages/api/internal/handlers/proxy_grpc.go
```diff
@@ -85,6 +86,25 @@ func (s *SandboxService) ResumeSandbox(ctx context.Context, req *proxygrpc.Sandb
 		network = snap.Snapshot.Config.Network
 	}
 
+	// Validate traffic access token for sandboxes with private ingress.
+	if network != nil && network.Ingress != nil && network.Ingress.AllowPublicAccess != nil && !*network.Ingress.AllowPublicAccess {
```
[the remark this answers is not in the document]
matthewlouisbrockman: oh. yeah this is like super important lmao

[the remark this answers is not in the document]
matthewlouisbrockman: fixed

packages/api/internal/handlers/proxy_grpc.go
```diff
@@ -85,6 +86,25 @@ func (s *SandboxService) ResumeSandbox(ctx context.Context, req *proxygrpc.Sandb
...
+		expectedToken, tokenErr := s.api.accessTokenGenerator.GenerateTrafficAccessToken(sandboxID)
+		if tokenErr != nil {
+			return nil, status.Errorf(codes.Internal, "failed to generate expected traffic token: %v", tokenErr)
```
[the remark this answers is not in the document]
matthewlouisbrockman: fixed

packages/api/internal/handlers/proxy_grpc.go
```diff
@@ -85,6 +86,25 @@ func (s *SandboxService) ResumeSandbox(ctx context.Context, req *proxygrpc.Sandb
...
+		}
+
+		var providedToken string
+		if md, ok := metadata.FromIncomingContext(ctx); ok {
+			if vals := md.Get("e2b-traffic-access-token"); len(vals) > 0 {
+				providedToken = vals[0]
+			}
+		}
+
+		if providedToken != expectedToken {
```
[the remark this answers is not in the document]
matthewlouisbrockman: fixed

packages/api/internal/handlers/proxy_grpc.go
```diff
@@ -85,6 +89,46 @@ func (s *SandboxService) ResumeSandbox(ctx context.Context, req *proxygrpc.Sandb
...
+				ctx,
+				"invalid sandbox request port metadata for resume",
+				zap.Error(parseErr),
+				zap.String("request_port", vals[0]),
+				logger.WithSandboxID(sandboxID),
+			)
+		} else {
+			isNonEnvdTraffic = requestPort != uint64(consts.DefaultEnvdServerPort)
+		}
+	}
```
jakubno: can this be handled in some helper function?

packages/api/internal/handlers/proxy_grpc.go
```diff
@@ -85,6 +89,46 @@ func (s *SandboxService) ResumeSandbox(ctx context.Context, req *proxygrpc.Sandb
...
+
+	// Validate traffic access token for sandboxes with private ingress.
+	if network != nil && network.Ingress != nil && network.Ingress.AllowPublicAccess != nil && !*network.Ingress.AllowPublicAccess && isNonEnvdTraffic {
```
jakubno: shouldn't you do basically the same for envdTraffic?

tests/integration/internal/tests/proxies/traffic_access_token_test.go
```diff
@@ -193,3 +197,97 @@ func TestEnvdPortIsNotAffectedByTrafficAccessToken(t *testing.T) {
...
+	serverStream, err := envdClient.ProcessClient.Start(serverCtx, serverReq)
+	require.NoError(t, err)
+	defer func() {
+		serverCancel()
+		if streamErr := serverStream.Close(); streamErr != nil {
+			t.Logf("Error closing server stream: %v", streamErr)
+		}
+	}()
+
+	time.Sleep(time.Second)
```
jakubno: no need to sleep if we use 
```
	resp = utils.WaitForStatus(t, client, sbx, proxyURL, port, nil, http.StatusOK)
```

tests/integration/internal/tests/proxies/traffic_access_token_test.go
```diff
@@ -193,3 +197,97 @@ func TestEnvdPortIsNotAffectedByTrafficAccessToken(t *testing.T) {
...
+
+	client := &http.Client{Timeout: 10 * time.Second}
+
+	sbxWithoutToken := *sbx
+	sbxWithoutToken.TrafficAccessToken = nil
+
+	// While running, missing traffic token must be rejected.
+	resp := utils.WaitForStatus(t, client, &sbxWithoutToken, proxyURL, port, nil, http.StatusForbidden)
+	require.NotNil(t, resp)
+	require.NoError(t, resp.Body.Close())
```
jakubno: no need to do this, this is tested elsewhere

tests/integration/internal/tests/proxies/traffic_access_token_test.go
```diff
@@ -193,3 +197,97 @@ func TestEnvdPortIsNotAffectedByTrafficAccessToken(t *testing.T) {
...
+	require.NoError(t, err)
+	require.Equal(t, http.StatusNoContent, pauseResp.StatusCode())
+
+	res, err := c.GetSandboxesSandboxIDWithResponse(ctx, sbx.SandboxID, setup.WithAPIKey())
+	require.NoError(t, err)
+	require.NotNil(t, res.JSON200, "expected 200 response, got status %d", res.StatusCode())
+	require.Equal(t, api.Paused, res.JSON200.State)
+
+	// While paused, missing token must not auto-resume and request should fail.
+	resp = utils.WaitForStatus(t, client, &sbxWithoutToken, proxyURL, port, nil, http.StatusBadGateway)
```
jakubno: no need to use `WaitForStatus`, do just one request and the status code should be the one expected

tests/integration/internal/tests/proxies/traffic_access_token_test.go
```diff
@@ -193,3 +197,97 @@ func TestEnvdPortIsNotAffectedByTrafficAccessToken(t *testing.T) {
...
+	require.NotNil(t, resp)
+	require.NoError(t, resp.Body.Close())
+
+	res, err = c.GetSandboxesSandboxIDWithResponse(ctx, sbx.SandboxID, setup.WithAPIKey())
+	require.NoError(t, err)
+	require.NotNil(t, res.JSON200, "expected 200 response, got status %d", res.StatusCode())
+	require.Equal(t, api.Paused, res.JSON200.State)
+
+	// Valid token request should auto-resume and succeed.
+	resp = utils.WaitForStatus(t, client, sbx, proxyURL, port, nil, http.StatusOK)
```
jakubno: same here

packages/api/internal/handlers/proxy_grpc.go
matthewlouisbrockman: yes - was considering it out of scope (we weren't doing auth) but makes sense to include it! will get that in

matthewlouisbrockman:
moving to draft while looking at the auth

packages/api/internal/handlers/proxy_grpc.go
```diff
@@ -29,6 +33,57 @@ func NewSandboxService(api *APIStore) *SandboxService {
...
+	}
+
+	return requestPort != uint64(consts.DefaultEnvdServerPort)
+}
+
+func isPrivateIngressTraffic(network *dbtypes.SandboxNetworkConfig) bool {
+	return network != nil && network.Ingress != nil && network.Ingress.AllowPublicAccess != nil && !*network.Ingress.AllowPublicAccess
+}
+
+func tokensMatch(providedToken string, expectedToken string) bool {
```
jakubno, line 79: could you explain why we can't just compare the strings?

packages/client-proxy/internal/proxy/proxy.go
```diff
@@ -41,11 +41,11 @@ const (
 	autoResumeErrored
 )
 
-func catalogResolution(ctx context.Context, sandboxId string, c catalog.SandboxesCatalog, pausedChecker PausedSandboxResumer, featureFlags *featureflags.Client) (string, error) {
+func findActiveNode(ctx context.Context, sandboxId string, sandboxPort uint64, trafficAccessToken string, envdAccessToken string, c catalog.SandboxesCatalog, pausedChecker PausedSandboxResumer, featureFlags *featureflags.Client) (string, error) {
```
jakubno: why the rename? it has the same problem as the previous name taking into account the resume logic

packages/client-proxy/internal/proxy/proxy.go
```diff
@@ -81,9 +84,13 @@ func handlePausedSandbox(
 	}
 
 	logger.L().Info(ctx, "catalog miss, attempting resume via api", logger.WithSandboxID(sandboxId))
-	nodeIP, err := pausedChecker.Resume(ctx, sandboxId)
+	nodeIP, err := pausedChecker.Resume(ctx, sandboxId, sandboxPort, trafficAccessToken, envdAccessToken)
 	if err != nil {
-		if isNotResumableError(err) {
+		if code, ok := getNotResumableCode(err); ok {
+			if code == codes.PermissionDenied {
+				return "", autoResumeErrored, reverseproxy.NewErrSandboxResumePermissionDenied(sandboxId)
```
jakubno: it didn't error

packages/client-proxy/internal/proxy/proxy.go
```diff
@@ -93,13 +100,18 @@ func handlePausedSandbox(
 	return nodeIP, autoResumeSucceeded, nil
 }
 
-func isNotResumableError(err error) bool {
+func getNotResumableCode(err error) (codes.Code, bool) {
```
jakubno: I think you're missing the function here, you just want to extract the grpc error

packages/client-proxy/internal/proxy/proxy.go
```diff
@@ -127,8 +139,17 @@ func NewClientProxy(meterProvider metric.MeterProvider, serviceName string, port
 				zap.Int64("content_length", r.ContentLength),
 			)
 
-			nodeIP, err := catalogResolution(ctx, sandboxId, catalog, pausedSandboxResumer, featureFlagsClient)
+			trafficAccessToken := r.Header.Get("e2b-traffic-access-token")
```
jakubno: isn't there a constant for that?

packages/client-proxy/internal/proxy/proxy.go
```diff
@@ -127,8 +139,17 @@ func NewClientProxy(meterProvider metric.MeterProvider, serviceName string, port
...
+			envdAccessToken := r.Header.Get("X-Access-Token")
```
jakubno: this is definitely a constant somewhere already

packages/client-proxy/internal/proxy/proxy_test.go
```diff
@@ -67,69 +87,164 @@ func TestCatalogResolution_CatalogHit_EmptyIPReturnsEmpty(t *testing.T) {
...
 	c := catalog.NewMemorySandboxesCatalog()
 	ff := newFF(t, true)
 
-	_, err := catalogResolution(context.Background(), "sbx", c, nil, ff)
+	_, err := findActiveNode(context.Background(), "sbx", 8000, "", "", c, nil, ff)
 	require.ErrorIs(t, err, ErrNodeNotFound)
 }
 
-func TestCatalogResolution_CatalogMiss_AutoResumeNotAllowed_FlagDisabled(t *testing.T) {
+func TestHandlePausedSandbox_NoResumer(t *testing.T) {
```
jakubno: _MissingTrafficAccessToken?

packages/shared/pkg/proxy/errors.go
```diff
@@ -33,6 +33,20 @@ func (e SandboxNotFoundError) Error() string {
 	return "sandbox not found"
 }
 
+type SandboxResumePermissionDeniedError struct {
+	SandboxId string
+}
+
+func NewErrSandboxResumePermissionDenied(sandboxId string) *SandboxResumePermissionDeniedError {
```
jakubno, line 40: there are corresponding html pages for errors, so you can return nicer error for people in browser

tests/integration/internal/tests/proxies/traffic_access_token_test.go
```diff
@@ -193,3 +197,156 @@ func TestEnvdPortIsNotAffectedByTrafficAccessToken(t *testing.T) {
...
+
+	proxyURL, err := url.Parse(setup.EnvdProxy)
+	require.NoError(t, err)
+
+	client := &http.Client{Timeout: 10 * time.Second}
+	envdPort := int(consts.DefaultEnvdServerPort)
+
+	// Verify envd is reachable with valid access token while running.
+	headers := &http.Header{"X-Access-Token": []string{*sbx.EnvdAccessToken}}
+	resp := utils.WaitForStatus(t, client, sbx, proxyURL, envdPort, headers, http.StatusNotFound)
```
jakubno: No need to wait for status and why should it return 404?

tests/integration/internal/tests/proxies/traffic_access_token_test.go
```diff
@@ -193,3 +197,156 @@ func TestEnvdPortIsNotAffectedByTrafficAccessToken(t *testing.T) {
...
+	res, err = c.GetSandboxesSandboxIDWithResponse(ctx, sbx.SandboxID, setup.WithAPIKey())
+	require.NoError(t, err)
+	require.NotNil(t, res.JSON200, "expected 200 response, got status %d", res.StatusCode())
+	require.Equal(t, api.Paused, res.JSON200.State)
+
+	// Valid envd access token should auto-resume.
+	req = utils.NewRequest(sbx, proxyURL, envdPort, headers)
+	resp, err = client.Do(req)
+	require.NoError(t, err)
+	require.Equal(t, http.StatusNotFound, resp.StatusCode) // envd has no GET / handler
```
jakubno: can you use /health?

packages/api/internal/handlers/proxy_grpc.go
matthewlouisbrockman: was using subtle.ConstantTimeCompare to avoid doing direct string comparisong - https://github.com/e2b-dev/infra/pull/1909#discussion_r2801837198 was complaining about it

matthewlouisbrockman: *am using. if there's not enough variance that it's not a real concern can go back to the original string comparison

packages/client-proxy/internal/proxy/proxy.go
matthewlouisbrockman: Renamed because it's no longer only hitting the catalog, can switch name back

matthewlouisbrockman: (it now hits the API as well directly if it isn't in the catalog)

packages/client-proxy/internal/proxy/proxy.go
matthewlouisbrockman: k, switching to 403

packages/client-proxy/internal/proxy/proxy.go
matthewlouisbrockman: moving it back though

packages/client-proxy/internal/proxy/proxy.go
matthewlouisbrockman: k

packages/client-proxy/internal/proxy/proxy.go
matthewlouisbrockman: using the one in grpc.proxy.metadata

packages/client-proxy/internal/proxy/proxy.go
matthewlouisbrockman: added to grpc.proxy.metadata

packages/client-proxy/internal/proxy/proxy_test.go
matthewlouisbrockman: TestHandlePausedSandbox_PermissionDenied has missing access token; can add to that and do wrong access token as well as missing access token?

packages/shared/pkg/proxy/errors.go
matthewlouisbrockman: made a resume html page

matthewlouisbrockman: can just have it re-use another one i guess if don't want to expose it exists or something

tests/integration/internal/tests/proxies/traffic_access_token_test.go
matthewlouisbrockman: fair point. changed to `req := utils.NewRequest(sbx, proxyURL, envdPort, headers)`

matthewlouisbrockman: removing the existing `waitforstatus` on that file too

packages/api/internal/handlers/proxy_grpc.go
jakubno: Good point, maybe we should do the same in other places, could you create a new task for that 🙏🏻

PR #1916, opened 2026-02-15
feat(api): add redis cache

jakubno:
> [!NOTE]
> **Medium Risk**
> Touches shared caching primitives and switches the template build cache behavior to Redis-backed reads/writes, which can impact performance and correctness under load if TTL/refresh semantics differ.
> 
> **Overview**
> This PR introduces a new generic `RedisCache` with TTL, singleflight miss deduplication, and optional background refresh, and refactors the template build cache to use it directly (removing the prior custom Redis utilities and L1 in-memory layer). It also renames the in-memory cache type to `MemoryCache` and updates existing auth/template caches and tests accordingly, including a Lua-based status update that resets Redis TTL on write.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3b1cf56817a7c6484a54eb8c60c7ee6663158304. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/shared/pkg/cache/redis.go
+++ b/packages/shared/pkg/cache/redis.go
@@ -15,6 +15,10 @@ import (
 
 const (
 	RedisRefreshIntervalOff = 0
+
+	// redisNoPTTL is the value returned by go-redis PTTL when a key has no expiration.
+	// Redis returns the integer -1; go-redis converts it to -1 * time.Nanosecond.
+	redisNoPTTL = -1 * time.Nanosecond
 )
 
 // RedisConfig holds the configuration for a RedisCache.
@@ -213,7 +217,7 @@ func (rc *RedisCache[V]) deleteFromRedis(ctx context.Context, key string) {
 // handleRefresh checks if the key should be refreshed in the background and refreshes it if necessary
 func (rc *RedisCache[V]) handleRefresh(ctx context.Context, key string, remainingTTL time.Duration, dataCallback DataCallback[V]) {
 	// check if key isn't persistent, this shouldn't really happen as it should be set with an expiration time
-	if remainingTTL == -1 {
+	if remainingTTL == redisNoPTTL {
 		logger.L().Warn(ctx, "redis key unexpectedly persistent", zap.String("key", key))
 
 		return
--- a/packages/shared/pkg/cache/redis_test.go
+++ b/packages/shared/pkg/cache/redis_test.go
@@ -31,6 +31,18 @@ func newTestRedisCache(t *testing.T, redisClient redis.UniversalClient) *RedisCa
 	})
 }
 
+func TestRedisNoPTTL_MatchesRedisBehavior(t *testing.T) {
+	t.Parallel()
+	redisClient := redis_utils.SetupInstance(t)
+
+	// Set a key with no expiration
+	err := redisClient.Set(t.Context(), "test:no-ttl", "value", 0).Err()
+	require.NoError(t, err)
+
+	pttl := redisClient.PTTL(t.Context(), "test:no-ttl").Val()
+	assert.Equal(t, redisNoPTTL, pttl, "redisNoPTTL constant should match PTTL result for a key with no expiration")
+}
+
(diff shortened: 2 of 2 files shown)


packages/shared/pkg/cache/layered.go
```diff
@@ -0,0 +1,238 @@
...
+		}
+
+		// Backfill into Redis
+		lc.setInRedis(ctx, key, value)
+
+		return value, nil
+	}
+}
+
+// getFromRedis fetches both the value and remaining TTL from Redis using a pipeline.
```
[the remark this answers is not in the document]
jakubno: That shouldn't happen, these items always have timeotu

packages/shared/pkg/cache/layered.go
```diff
@@ -0,0 +1,238 @@
...
+	redisKey := lc.RedisKey(key)
+
+	pipe := lc.config.RedisClient.Pipeline()
+	getCmd := pipe.Get(ctx, redisKey)
+	ttlCmd := pipe.TTL(ctx, redisKey)
+
+	_, err := pipe.Exec(ctx)
+	if err != nil {
+		return zero, 0, err
+	}
```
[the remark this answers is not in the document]
jakubno: That's why we check the error

packages/shared/pkg/cache/redis.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	r, _, _ := rc.fetchGroup.Do(key, func() (any, error) {
+		ctx := context.WithoutCancel(ctx)
+		// Double-check Redis (another goroutine may have populated it)
+		if v, _, redisErr := rc.getFromRedis(ctx, key); redisErr == nil {
+			return result{value: v}, nil
+		}
+
+		// Call the data callback with a cancel-free context
```
arkamar: This comment seems to be a bit off because the dataCallback() is called with context that can be canceled by timeout.

packages/api/internal/cache/templates/template_build_test.go
```diff
@@ -157,3 +150,46 @@ func TestRedisTemplatesBuildCache_SetStatus_UpdatesAndInvalidatesL1(t *testing.T
...
+	shortTTL := 30 * time.Second
+	err = redisClient.Set(ctx, buildKey, buildJSON, shortTTL).Err()
+	require.NoError(t, err)
+
+	// Verify the initial TTL is short
+	ttlBefore, err := redisClient.TTL(ctx, buildKey).Result()
+	require.NoError(t, err)
+	assert.LessOrEqual(t, ttlBefore, shortTTL)
+
+	// Update status — this should reset TTL to redisBuildCacheTTL (5 minutes)
```
arkamar: nitpick: the `redisBuildCacheTTL` is no longer used, it was renamed to `buildCacheTTL`. There is multiple places like this one in this file.

packages/shared/pkg/cache/redis.go
```diff
@@ -0,0 +1,213 @@
...
+
+// GetOrSet retrieves a value from Redis, falling back to dataCallback on miss.
+// On callback hit, the value is backfilled into Redis.
+// Singleflight deduplicates concurrent misses for the same key.
+func (rc *RedisCache[V]) GetOrSet(ctx context.Context, key string, dataCallback DataCallback[V]) (V, error) {
+	// Fast path: try Redis outside singleflight
+	value, remainingTTL, err := rc.getFromRedis(ctx, key)
+	if err == nil {
+		age := rc.config.TTL - remainingTTL
+		if rc.config.RefreshInterval != RedisRefreshIntervalOff && age > rc.config.RefreshInterval {
```
arkamar: It seems that it might happen that `remainingTTL` is set to `-1` if the key exists, but it has no associated expiration (it's persistent). `-1` value is represented in nanoseconds, so, no big deal, the code should probably stay like this, but it may be worth commenting, that we don't want to handle this unlikely situation.

PR #1919, opened 2026-02-16
include estimated usage in sandbox end of life webhook

matthewlouisbrockman:
- Improves bookkeeping by adding `execution` based on the orchestrator's sandbox data in end of life paused/killed webhooks into the event_data field (currently contains metadata).
- has `startedAt` in the metadata use the same mutex setter/getter pattern as `endAt`

```
func (s *Server) getSandboxUsageData(sbx *sandbox.Sandbox) map[string]any {
	return map[string]any{
		"started_at":  sbx.StartedAt.UTC().Format(time.RFC3339Nano),
		"vcpu_count":      sbx.Config.Vcpu,
		"memory_mb":       sbx.Config.RamMB,
		"duration": time.Since(sbx.StartedAt).Seconds(),
	}
}
```

Helps with simplifying bookkeeping on estimating where resources are allocated.

### DIFF
--- a/packages/api/internal/cache/templates/alias_cache.go
+++ b/packages/api/internal/cache/templates/alias_cache.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"time"
 
 	"github.com/google/uuid"
 	"go.opentelemetry.io/otel/attribute"
@@ -14,6 +15,13 @@ import (
 	"github.com/e2b-dev/infra/packages/db/queries"
 	"github.com/e2b-dev/infra/packages/shared/pkg/cache"
 	"github.com/e2b-dev/infra/packages/shared/pkg/id"
+)
+
+const (
+	templateInfoExpiration = 5 * time.Minute
+	refreshInterval        = 1 * time.Minute
+	refreshTimeout         = 30 * time.Second
+	callbackTimeout        = 30 * time.Second
 )
 
 // AliasInfo holds resolved alias information
--- a/packages/api/internal/cache/templates/cache.go
+++ b/packages/api/internal/cache/templates/cache.go
@@ -4,10 +4,10 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"strings"
 	"time"
 
 	"github.com/google/uuid"
+	"github.com/redis/go-redis/v9"
 	"go.opentelemetry.io/otel"
 
 	"github.com/e2b-dev/infra/packages/api/internal/api"
@@ -21,48 +21,54 @@ import (
 )
 
 const (
-	templateInfoExpiration = 5 * time.Minute
-	refreshInterval        = 1 * time.Minute
-	refreshTimeout         = 30 * time.Second
-	callbackTimeout        = 30 * time.Second
+	templateCacheTTL             = 5 * time.Minute
+	templateCacheRefreshInterval = 1 * time.Minute
+	templateCacheTimeout         = 2 * time.Second
+
+	templateCacheKeyPrefix = "template:info"
 )
 
 var tracer = otel.Tracer("github.com/e2b-dev/infra/packages/api/internal/cache/templates")
 
 func buildCacheKey(templateID, tag string) string {
-	return templateID + id.TagSeparator + tag
+	// Wrap templateID in {} so it becomes a Redis hash tag — all keys for the
+	// same template land on the same hash slot in Redis Cluster, enabling
+	// pipelined prefix deletion in InvalidateAllTags.
+	return fmt.Sprintf("{%s}:%s", templateID, tag)
 }
 
 // TemplateInfo holds cached template with build information
 type TemplateInfo struct {
-	Template  *api.Template
-	TeamID    uuid.UUID
-	ClusterID uuid.UUID
-	Build     *queries.EnvBuild
-	Tag       string
+	Template  *api.Template     `json:"template"`
+	TeamID    uuid.UUID         `json:"team_id"`
+	ClusterID uuid.UUID         `json:"cluster_id"`
+	Build     *queries.EnvBuild `json:"build"`
+	Tag       string            `json:"tag"`
 }
 
 // TemplateCache caches template+build by templateID:tag.
 // This is a simple lookup layer - resolution happens in AliasCache.
 type TemplateCache struct {
-	cache      *cache.MemoryCache[*TemplateInfo]
+	cache      *cache.RedisCache[*TemplateInfo]
 	db         *sqlcdb.Client
 	aliasCache *AliasCache
 }
 
-func NewTemplateCache(db *sqlcdb.Client) *TemplateCache {
-	config := cache.Config[*TemplateInfo]{
-		TTL:             templateInfoExpiration,
-		RefreshInterval: refreshInterval,
-		RefreshTimeout:  refreshTimeout,
-		CallbackTimeout: callbackTimeout,
+func NewTemplateCache(db *sqlcdb.Client, redisClient redis.UniversalClient) *TemplateCache {
+	redisCache := cache.NewRedisCache[*TemplateInfo](cache.RedisConfig[*TemplateInfo]{
+		TTL:             templateCacheTTL,
+		RefreshInterval: templateCacheRefreshInterval,
+		RedisTimeout:    templateCacheTimeout,
+		RedisClient:     redisClient,
+		RedisPrefix:     templateCacheKeyPrefix,
+
 		ExtractKeyFunc: func(value *TemplateInfo) string {
 			return buildCacheKey(value.Template.TemplateID, value.Tag)
 		},
-	}
+	})
 
 	return &TemplateCache{
-		cache:      cache.NewMemoryCache(config),
+		cache:      redisCache,
 		db:         db,
 		aliasCache: NewAliasCache(db),
 	}
@@ -114,9 +120,7 @@ func (c *TemplateCache) getByID(ctx context.Context, templateID string, tag *str
 	}
 	cacheKey := buildCacheKey(templateID, tagValue)
 
-	info, err := c.cache.GetOrSet(ctx, cacheKey, func(ctx context.Context, _ string) (*TemplateInfo, error) {
-		return c.fetchTemplateWithBuild(ctx, templateID, tag)
-	})
+	info, err := c.cache.GetOrSet(ctx, cacheKey, c.fetchTemplateWithBuild(templateID, tag))
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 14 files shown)


packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -520,3 +522,12 @@ func (s *Server) prepareSandboxEventData(ctx context.Context, sbx *sandbox.Sandb
 
 	return teamID, buildId, eventData
 }
+
+func (s *Server) getSandboxUsageData(sbx *sandbox.Sandbox) map[string]any {
+	return map[string]any{
+		"started_at_utc":  sbx.StartedAt.UTC().Format(time.RFC3339Nano),
+		"vcpu_count":      sbx.Config.Vcpu,
+		"memory_mb":       sbx.Config.RamMB,
+		"runtime_seconds": time.Since(sbx.StartedAt).Seconds(),
```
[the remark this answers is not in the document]
matthewlouisbrockman: i don't think need to worry about this with the estimated_usage, purely a low impact helper to make it easy to know what resources are getting used, doesn't include a lot of edge cases yet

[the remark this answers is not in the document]
sitole: @matthewlouisbrockman i think data race is actually possible here. Imho, we should do a mutex for getter/setter or similar.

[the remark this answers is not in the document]
matthewlouisbrockman: aight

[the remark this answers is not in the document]
matthewlouisbrockman: updated it! now `startedAt` uses the exact same pattern as `endAt`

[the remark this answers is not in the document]
matthewlouisbrockman: @claude does this address the issue?

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -1116,7 +1135,7 @@ func (s *Sandbox) WaitForEnvd(
 			attribute.Int64("timeout_ms", s.internalConfig.EnvdInitRequestTimeout.Milliseconds()),
 		))
 		// Update the sandbox as started now
-		s.Metadata.StartedAt = time.Now()
+		s.SetStartedAt(time.Now())
```
djeebus, line 1147: why reset it here?

matthewlouisbrockman: existing behavior, it just sets the sandbox started at once envd starts so; i think so it tracks started at from `sandbox is ready` vs `sandbox requested` (edit: github apparently treated that as html tags)

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -364,6 +365,7 @@ func (s *Server) Delete(ctxConn context.Context, in *orchestrator.SandboxDeleteR
 	}()
 
 	teamID, buildId, eventData := s.prepareSandboxEventData(ctx, sbx)
+	eventData["estimated_usage"] = s.getSandboxUsageData(sbx)
```
sitole: Please use constant for map key.

Why its called `estimated_usage`, values inside are exact, not estimated.

matthewlouisbrockman: the `cpu` and `RAM` are kind of like... baseline requested, not the exact usage of the sandbox, wasn't sure a better way to reflect that

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -520,3 +523,14 @@ func (s *Server) prepareSandboxEventData(ctx context.Context, sbx *sandbox.Sandb
...
 }
+
+func (s *Server) getSandboxUsageData(sbx *sandbox.Sandbox) map[string]any {
+	startedAt := sbx.GetStartedAt()
+
+	return map[string]any{
+		"started_at_utc":  startedAt.UTC().Format(time.RFC3339Nano),
+		"vcpu_count":      sbx.Config.Vcpu,
+		"memory_mb":       sbx.Config.RamMB,
+		"runtime_seconds": time.Since(startedAt).Seconds(),
```
sitole: What about calling it execution/execution_time?

packages/orchestrator/internal/server/sandboxes.go
sitole: So lets call it execution. It will give you execution time and resources you request. Later we can add fields explicitly showing used resources.

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -520,3 +523,14 @@ func (s *Server) prepareSandboxEventData(ctx context.Context, sbx *sandbox.Sandb
 
 	return teamID, buildId, eventData
 }
+
+func (s *Server) getSandboxUsageData(sbx *sandbox.Sandbox) map[string]any {
+	startedAt := sbx.GetStartedAt()
+
+	return map[string]any{
+		"started_at_utc":  startedAt.UTC().Format(time.RFC3339Nano),
```
sitole: Drop `utc`. Do we want `RFC3339Nano` or just `RFC3339`? Is there any other usage of nano somewhere?

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -520,3 +523,14 @@ func (s *Server) prepareSandboxEventData(ctx context.Context, sbx *sandbox.Sandb
 
 	return teamID, buildId, eventData
 }
+
+func (s *Server) getSandboxUsageData(sbx *sandbox.Sandbox) map[string]any {
+	startedAt := sbx.GetStartedAt()
+
+	return map[string]any{
```
sitole: Let's put it behind a feature flag so we can enable it once the cluster is fully rolled out.

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -520,3 +523,14 @@ func (s *Server) prepareSandboxEventData(ctx context.Context, sbx *sandbox.Sandb
 
 	return teamID, buildId, eventData
 }
+
+func (s *Server) getSandboxUsageData(sbx *sandbox.Sandbox) map[string]any {
```
sitole: Please add a test to ensure we do not break the contract with feature updates here.

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -348,6 +364,8 @@ func (f *Factory) CreateSandbox(
 		exit: exit,
 	}
 
+	sbx.SetStartedAt(time.Now())
```
sitole: Why are you calling it like that and not setting it directly on the struct? This pattern is used on two places

packages/orchestrator/internal/server/sandboxes.go
matthewlouisbrockman: k

packages/orchestrator/internal/server/sandboxes.go
matthewlouisbrockman: Screenshot 2026-02-16 at 11 22 50 AM
it's like 50/50 i guess

matthewlouisbrockman: RFC3339 should be fine I guess

packages/orchestrator/internal/server/sandboxes.go
matthewlouisbrockman: aight

packages/orchestrator/internal/server/sandboxes.go
sitole: @matthewlouisbrockman Looks like nano is used only where it's required by the services contract (open telemetry, cursors for more granular cursors).

packages/orchestrator/internal/server/sandboxes.go
matthewlouisbrockman: k

packages/orchestrator/internal/sandbox/sandbox.go
matthewlouisbrockman: i think need to because of the mutex, lemme look into this

matthewlouisbrockman: k, didn't need to do that for the mutex, can still inline it

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -520,6 +546,17 @@ func (s *Server) prepareSandboxEventData(ctx context.Context, sbx *sandbox.Sandb
 	return teamID, buildId, eventData
 }
 
+func (s *Server) getSandboxUsageData(sbx *sandbox.Sandbox) map[string]any {
```
sitole:
```suggestion
func (s *Server) getSandboxExecutionData(sbx *sandbox.Sandbox) map[string]any {
```

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -392,7 +396,29 @@ func (s *Server) Pause(ctx context.Context, in *orchestrator.SandboxPauseRequest
 		return nil, status.Errorf(codes.Internal, "error snapshotting sandbox '%s': %s", in.GetSandboxId(), err)
 	}
 
-	s.publishSandboxEvent(ctx, sbx, events.SandboxPausedEventPair.Type)
+	teamID, buildId, eventData := s.prepareSandboxEventData(ctx, sbx)
+	if s.featureFlags.BoolFlag(ctx, featureflags.ExecutionMetricsOnWebhooksFlag) {
+		eventData["execution"] = s.getSandboxUsageData(sbx)
```
sitole: Let's add a constant or test to ensure we don't accidentally rename it later.

packages/shared/pkg/feature-flags/flags.go
```diff
@@ -94,6 +94,7 @@ var (
 	CreateStorageCacheSpansFlag         = newBoolFlag("create-storage-cache-spans", env.IsDevelopment())
 	SandboxAutoResumeFlag               = newBoolFlag("sandbox-auto-resume", env.IsDevelopment())
 	PersistentVolumesFlag               = newBoolFlag("can-use-persistent-volumes", env.IsDevelopment())
+	ExecutionMetricsOnWebhooksFlag      = newBoolFlag("execution-metrics-on-webhooks", false)
```
sitole: Lets add link to Linear task to remove it later

sitole (approved):
lgtm, just nits. After this is merged lets please open PR with updated docs - https://github.com/e2b-dev/docs.

packages/orchestrator/internal/server/sandboxes.go
matthewlouisbrockman: constant makes sense

packages/shared/pkg/feature-flags/flags.go
matthewlouisbrockman: added https://linear.app/e2b/issue/ENG-3519/remove-feature-flag-after-stable

PR #1938, opened 2026-02-18
chore: denormalize builds ownership

dobrac:
In order to scan through builds faster, denormalize the team_id by including it for builds too. The case where one build could be owned by multiple teams is not supported/advised and only first assigned team will be set for the build.

---

> [!NOTE]
> **Medium Risk**
> Adds a large online backfill plus new trigger and index changes on `env_builds`, which could impact production performance and query plans during/after migration, but doesn’t change auth or external security boundaries.
> 
> **Overview**
> This PR denormalizes build ownership by adding nullable `team_id` to `env_builds`, backfilling it in batches from `env_build_assignments` → `envs`, and keeping it automatically populated via an `AFTER INSERT` trigger on `env_build_assignments`. It updates sqlc models/queries to select/scan `eb.team_id` and replaces the old `created_at`-only index with new composite indexes keyed by `(team_id, created_at, id)` to speed team-scoped build pagination.
> 
> Potential issues: the trigger assumes each build maps to a single team (multi-team assignments will silently keep the first), and the migration drops `idx_env_builds_created_at` which may regress any non-team-scoped queries that relied on it. The backfill procedure commits in a loop and does large batch updates, so it can still create sustained write/IO load during migration despite being cursor-based.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 766cba964e924b12e6d318094ff93f4c20edc667. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/db/migrations/20260218120000_add_team_id_to_env_builds.sql
+++ b/packages/db/migrations/20260218120000_add_team_id_to_env_builds.sql
@@ -87,9 +87,12 @@ CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_env_builds_team_status_pagination
 CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_env_builds_team_env_created_id
   ON public.env_builds (team_id, env_id, created_at DESC, id DESC);
 
+DROP INDEX CONCURRENTLY IF EXISTS idx_env_builds_created_at;
+
 -- +goose Down
 -- +goose NO TRANSACTION
 
+CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_env_builds_created_at ON env_builds (created_at DESC);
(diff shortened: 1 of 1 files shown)


packages/db/migrations/20260218120000_add_team_id_to_env_builds.sql
```diff
@@ -0,0 +1,67 @@
+-- +goose Up
+-- +goose NO TRANSACTION
+
+ALTER TABLE public.env_builds ADD COLUMN IF NOT EXISTS team_id uuid;
```
jakubno, line 4: do i understand this is temp and we'll be setting this in the code and then we can make it non nullable?

dobrac: eventually yes, not right now as it requires two step migration

PR #1940, opened 2026-02-18
feat: add cgroups v2 resource accounting for sandboxes

arkamar:
Add host-level CPU and memory resource accounting for Firecracker sandboxes using Linux cgroups v2. Each sandbox gets its own cgroup under `/sys/fs/cgroup/e2b/sbx-{sandboxID}`, enabling per-sandbox resource usage tracking that flows into ClickHouse via the existing hoststats collector.

Sandboxes are placed into their cgroup at clone time using `CLONE_INTO_CGROUP` (`SysProcAttr.UseCgroupFD`), eliminating the race condition where Firecracker could exit before being assigned to its cgroup via post-creation PID write. The `memory.peak` file is reset after each sample using per-FD kernel semantics (kernel 5.7+), enabling interval-based peak tracking instead of lifetime peak.

---

> [!NOTE]
> **Medium Risk**
> Touches sandbox process startup and host stats ingestion, so misconfiguration or kernel/cgroup differences could break sandbox launches or reduce metric reliability. Also adds system-level filesystem writes under `/sys/fs/cgroup`, which may fail in restricted environments and needs careful rollout/monitoring.
> 
> **Overview**
> Adds cgroup v2-based per-sandbox CPU and memory accounting and persists the new metrics to ClickHouse host stats. The orchestrator now creates a dedicated cgroup for each sandbox, starts Firecracker with `CLONE_INTO_CGROUP` via a cgroup directory FD, samples `cpu.stat`/`memory.current` (and interval `memory.peak` when available), and cleans up the cgroup on shutdown; ClickHouse schema/insert paths are extended with new cgroup metric columns.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4c14e8227383744b2a93f3a27ef9c03bf2f76e00. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/shared/pkg/feature-flags/flags.go
+++ b/packages/shared/pkg/feature-flags/flags.go
@@ -94,7 +94,6 @@ var (
 	CreateStorageCacheSpansFlag         = newBoolFlag("create-storage-cache-spans", env.IsDevelopment())
 	SandboxAutoResumeFlag               = newBoolFlag("sandbox-auto-resume", env.IsDevelopment())
 	PersistentVolumesFlag               = newBoolFlag("can-use-persistent-volumes", env.IsDevelopment())
-	EnableCgroupAccounting              = newBoolFlag("enable-cgroup-accounting", env.IsDevelopment())
 )
 
 type IntFlag struct {
--- a/packages/orchestrator/main.go
+++ b/packages/orchestrator/main.go
@@ -347,22 +347,16 @@ func run(config cfg.Config) (success bool) {
 	}
 
 	// cgroup manager for resource accounting
-	var cgroupManager cgroup.Manager
-	if featureFlags.BoolFlag(ctx, featureflags.EnableCgroupAccounting) {
-		var err error
-		cgroupManager, err = cgroup.NewManager()
-		if err != nil {
-			logger.L().Fatal(ctx, "failed to initialize cgroup manager", zap.Error(err))
-		}
-
-		if err := cgroupManager.Initialize(ctx); err != nil {
-			logger.L().Fatal(ctx, "failed to initialize root cgroup", zap.Error(err))
-		}
-
-		logger.L().Info(ctx, "cgroup accounting enabled", zap.String("root", cgroup.RootCgroupPath))
-	} else {
-		logger.L().Info(ctx, "cgroup accounting disabled")
+	cgroupManager, err := cgroup.NewManager()
+	if err != nil {
+		logger.L().Fatal(ctx, "failed to initialize cgroup manager", zap.Error(err))
 	}
+
+	if err := cgroupManager.Initialize(ctx); err != nil {
+		logger.L().Fatal(ctx, "failed to initialize root cgroup", zap.Error(err))
+	}
+
+	logger.L().Info(ctx, "cgroup accounting enabled", zap.String("root", cgroup.RootCgroupPath))
 
 	// redis
 	redisClient, err := sharedFactories.NewRedisClient(ctx, sharedFactories.RedisConfig{


packages/shared/pkg/feature-flags/flags.go
```diff
@@ -94,6 +94,7 @@ var (
 	CreateStorageCacheSpansFlag         = newBoolFlag("create-storage-cache-spans", env.IsDevelopment())
 	SandboxAutoResumeFlag               = newBoolFlag("sandbox-auto-resume", env.IsDevelopment())
 	PersistentVolumesFlag               = newBoolFlag("can-use-persistent-volumes", env.IsDevelopment())
+	EnableCgroupAccounting              = newBoolFlag("enable-cgroup-accounting", env.IsDevelopment())
```
jakubno: I think the FF here doesn't make much sense and we can just start it with it as it can't be changed during runtime

packages/orchestrator/internal/sandbox/cgroup/manager.go
```diff
@@ -0,0 +1,312 @@
...
+	logger.L().Debug(ctx, "created cgroup for sandbox",
+		logger.WithSandboxID(sandboxID),
+		zap.String("path", cgroupPath),
+		zap.Int("fd", handle.GetFD()),
+		zap.Bool("peak_reset_available", memoryPeakFile != nil))
+
+	return handle, nil
+}
+
+func (m *managerImpl) getStatsForPath(ctx context.Context, cgroupPath string, memoryPeakFile *os.File) (*Stats, error) {
```
jakubno, line 237: isn't there a library for the parsing?

packages/orchestrator/internal/sandbox/cgroup/manager.go
```diff
@@ -0,0 +1,312 @@
...
+	// Reset per-FD peak for next interval
+	if _, err := memoryPeakFile.WriteString("0"); err != nil {
+		logger.L().Debug(ctx, "failed to reset memory.peak", zap.Error(err))
+	}
+
+	return peakBytes, nil
+}
+
+// sandboxCgroupPath returns the filesystem path for a sandbox's cgroup
+func (m *managerImpl) sandboxCgroupPath(sandboxID string) string {
```
jakubno, line 313: How sure we are about clean up and would it cause any issue if the file already existed (pause followed with resume or snapshot)

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -1023,6 +1052,33 @@ func pauseProcessRootfs(
 	return rootfsDiff, rootfsHeader, nil
 }
 
+// createCgroup creates a cgroup for sandbox resource accounting if cgroup
+// accounting is enabled (cgroupManager is non-nil). It registers cleanup with
+// the provided Cleanup so the cgroup is removed on error paths.
+//
+// Returns the CgroupHandle and the cgroup directory FD to pass to the
+// Firecracker process. If cgroup accounting is disabled, returns (nil, cgroup.NoCgroupFD).
+func (f *Factory) createCgroup(ctx context.Context, sandboxID string, cleanup *Cleanup) (*cgroup.CgroupHandle, int) {
```
jakubno: can you add span for tracing for the function? 🙏🏻

jakubno: and maybe it doesn't have to live on `Factory`

packages/orchestrator/internal/sandbox/cgroup/manager.go
arkamar: I originally found https://pkg.go.dev/github.com/containerd/cgroups/v3, which looked promising, but I stuck with custom parser because of `memory.peak` file, where we want the reset for the next interval. Then, it looked weird to me to have mixed library parser with custom parser, so, I stick with custom one.

packages/orchestrator/internal/sandbox/cgroup/manager.go
arkamar: The clean up seems to be pretty solid. I tested it in few scenarios, e.g.: regular `e2b sbx cr ...`, pause followed with resume, and sending `kill` from the host directly to firecracker. The sandbox cgroup was cleaned in all of them. The only scenario where the sandbox cgroup was not cleaned properly was when I killed orchestrator with running sandbox (with TERM signal).

The only issue it can cause is inaccurate statistics from previous runs, e.g. cpu.stats will not start from 0, memory.current and memory.peak should be fine.

Actually, when I am thinking about it more, should the sandbox cgroup be preserved in pause followed with resume scenario, otherwise you would have to deal with it somehow in database.

jakubno: We do handle it, you can distinguish the parts based on the executionID

PR #1942, opened 2026-02-19
Fix: Template status temporary logs order parity per direction with Loki

ben-fornefeld:
> [!NOTE]
> **Low Risk**
> Behavioral change is limited to log entry ordering in the template status response, with added tests; no auth, data mutation, or storage logic changes.
> 
> **Overview**
> Fixes temporary template build log ordering so it matches Loki/persistent logs for both forward and backward directions by sorting entries by timestamp (stable for equal timestamps) instead of reversing slices. Adds regression tests covering direction ordering and parity between temporary vs persistent log sources, plus a Loki `ResponseMapper` ordering test.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 778215c0822baff3b59ce085e117919ed16da763. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/orchestrator/internal/template/server/template_status.go
+++ b/packages/orchestrator/internal/template/server/template_status.go
@@ -45,10 +45,15 @@ func (s *ServerStore) TemplateBuildStatus(ctx context.Context, in *template_mana
 	}
 
 	logLines := buildInfo.GetLogs()
-	// If the direction is backward, we need to reverse the log entries to have them in the descending order
-	if direction == template_manager.LogsDirection_Backward {
-		slices.Reverse(logLines)
-	}
+
+	// Keep response ordering aligned with persistent log mapping.
+	slices.SortStableFunc(logLines, func(a, b *template_manager.TemplateBuildLogEntry) int {
+		if direction == template_manager.LogsDirection_Backward {
+			return b.GetTimestamp().AsTime().Compare(a.GetTimestamp().AsTime())
+		}
+
+		return a.GetTimestamp().AsTime().Compare(b.GetTimestamp().AsTime())
+	})
 
 	logEntries := make([]*template_manager.TemplateBuildLogEntry, 0)
 	logsCrawled := int32(0)
@@ -77,15 +82,6 @@ func (s *ServerStore) TemplateBuildStatus(ctx context.Context, in *template_mana
 
 		logEntries = append(logEntries, entry)
 	}
-
-	// Keep response ordering aligned with persistent log mapping.
-	slices.SortStableFunc(logEntries, func(a, b *template_manager.TemplateBuildLogEntry) int {
-		if direction == template_manager.LogsDirection_Backward {
-			return b.GetTimestamp().AsTime().Compare(a.GetTimestamp().AsTime())
-		}
-
-		return a.GetTimestamp().AsTime().Compare(b.GetTimestamp().AsTime())
-	})
 
 	result := buildInfo.GetResult()
 	if result == nil {
--- a/packages/orchestrator/internal/template/server/template_status_test.go
+++ b/packages/orchestrator/internal/template/server/template_status_test.go
@@ -1,7 +1,6 @@
 package server
 
 import (
-	"context"
 	"encoding/json"
 	"strings"
 	"testing"
@@ -49,9 +48,9 @@ func TestTemplateBuildStatus_DirectionOrdering(t *testing.T) {
 			expected:  []string{"first", "same-a", "same-b", "last"},
 		},
 		{
-			name:      "backward_sorts_descending_and_keeps_reversed_equal_timestamps_stable",
+			name:      "backward_sorts_descending_and_keeps_equal_timestamps_stable",
 			direction: template_manager.LogsDirection_Backward,
-			expected:  []string{"last", "same-b", "same-a", "first"},
+			expected:  []string{"last", "same-a", "same-b", "first"},
 		},
 	}
 
--- a/packages/shared/pkg/logs/loki/loki_test.go
+++ b/packages/shared/pkg/logs/loki/loki_test.go
@@ -1,7 +1,6 @@
 package loki
 
 import (
-	"context"
 	"encoding/json"
 	"testing"
 	"time"


packages/orchestrator/internal/template/server/template_status_test.go
jakubno, line 1: do we have to use loki package here?

ben-fornefeld: not necessarily, but if i replace it with local expectations the test loses it's meaning. this one should make sure we have order parity, no matter the `source` we request template build status from.

ben-fornefeld: since loki is disabled in integration tests, alternative i'm seeing would be two separate tests for loki and template manager, that define the order expectation both in their test files respectively. downsides here is that the contract definition is duplicated.

packages/orchestrator/internal/template/server/template_status_test.go
```diff
@@ -0,0 +1,171 @@
...
+				Start:      timestamppb.New(now.Add(-1 * time.Hour)),
+				End:        timestamppb.New(now.Add(1 * time.Hour)),
+				Direction:  &tc.direction,
+			})
+			require.NoError(t, err)
+
+			actualEntries := response.GetLogEntries()
+			require.Len(t, actualEntries, len(lines))
+
+			expectedMessages, err := persistentMapperOrder(lines, tc.direction)
```
sitole: I would prefer to build it manually; this way, you are just layering logic. There should not be any connection between template builder and Loki, imho.

packages/orchestrator/internal/template/server/template_status.go
```diff
@@ -78,10 +78,14 @@ func (s *ServerStore) TemplateBuildStatus(ctx context.Context, in *template_mana
 		logEntries = append(logEntries, entry)
 	}
 
-	// If the direction is backward, we need to reverse the log entries again to have them back in the ascending order
-	if direction == template_manager.LogsDirection_Backward {
-		slices.Reverse(logEntries)
-	}
+	// Keep response ordering aligned with persistent log mapping.
+	slices.SortStableFunc(logEntries, func(a, b *template_manager.TemplateBuildLogEntry) int {
```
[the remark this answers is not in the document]
sitole: @ben-fornefeld 
If logs insert is not somehow buffered, we should be able to move `slices.SortStableFunc` and replace `slices.Reverse(logLines)`.

packages/orchestrator/internal/template/server/template_status_test.go
```diff
@@ -0,0 +1,123 @@
...
+			direction: template_manager.LogsDirection_Backward,
+			expected:  []string{"last", "same-b", "same-a", "first"},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			response, err := serverStore.TemplateBuildStatus(context.Background(), &template_manager.TemplateStatusRequest{
```
dobrac:
```suggestion
			response, err := serverStore.TemplateBuildStatus(t.Context(), &template_manager.TemplateStatusRequest{
```

packages/shared/pkg/logs/loki/loki_test.go
```diff
@@ -0,0 +1,79 @@
...
+			direction: logproto.BACKWARD,
+			expected:  []string{"third", "second", "first"},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			entries, err := ResponseMapper(context.Background(), res, 0, nil, tc.direction)
```
dobrac:
```suggestion
			entries, err := ResponseMapper(t.Context(), res, 0, nil, tc.direction)
```

packages/orchestrator/internal/template/server/template_status_test.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+}
+
+func newTestServerStore(t *testing.T, logLines []testLogLine) (*ServerStore, string) {
+	t.Helper()
+
+	buildLogs := buildlogger.NewLogEntryLogger()
+	writeTestBuildLogs(t, buildLogs, logLines)
+
+	buildCache := templatecache.NewBuildCache(context.Background(), noop.NewMeterProvider())
```
dobrac:
```suggestion
	buildCache := templatecache.NewBuildCache(t.Context(), noop.NewMeterProvider())
```

packages/api/internal/clusters/resources_test.go
```diff
@@ -0,0 +1,165 @@
...
+				persistentCalls++
+				entries := make([]logs.LogEntry, len(tc.expected))
+				copy(entries, tc.expected)
+
+				return entries, nil
+			}
+
+			temporarySource := api.LogsSourceTemporary
+			fromTemporary, temporaryErr := getBuildLogsWithSources(
+				context.Background(),
```
dobrac:
```suggestion
				t.Context(),
```

packages/api/internal/clusters/resources_test.go
```diff
@@ -0,0 +1,165 @@
...
+			assert.Equal(t, tc.expected, fromTemporary)
+			assert.Equal(t, 1, templateClient.callCount)
+			assert.Equal(t, 0, persistentCalls)
+
+			templateClient.callCount = 0
+			persistentCalls = 0
+
+			persistentSource := api.LogsSourcePersistent
+			fromPersistent, persistentErr := getBuildLogsWithSources(
+				context.Background(),
```
dobrac:
```suggestion
				t.Context(),
```

ben-fornefeld:
> The fix is not correct when you apply pagination. If logs are not in correct order, you'll miss/get different data because the filtering happens before sorting.

goal here was to unify it with the loki provider but i see the concern. we can sort it before instead of using reverse to fix the underlying inconsistency then?

dobrac (approved):

PR #1949, opened 2026-02-19
feat(client-proxy): add gRPC client metrics for auto-resume

matthewlouisbrockman:
Adds telemetry to the gRPC calls for the auto-resume gRPC calls

### DIFF
--- a/iac/modules/job-otel-collector/configs/otel-collector.yaml
+++ b/iac/modules/job-otel-collector/configs/otel-collector.yaml
@@ -214,6 +214,7 @@ processors:
           - "db.sql.connection.*"
           - "vault.*"
           - "client_proxy.*"
+          - "rpc\\.client\\..*"
           - "Click*"
           - "otelcol.exporter.queue.capacity"
           - "otelcol.exporter.queue.size"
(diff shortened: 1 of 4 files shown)


djeebus (approved):

iac/modules/job-otel-collector/configs/otel-collector.yaml
```diff
@@ -214,6 +214,7 @@ processors:
           - "db.sql.connection.*"
           - "vault.*"
           - "client_proxy.*"
+          - "rpc\\.client\\..*"
```
jakubno, line 217: can we do it the other way (server) to already take advantage of the setup?

jakubno: if not, can you please check what were the issues for server, we had to drop some labels because the cardinality was huge

matthewlouisbrockman: checking

matthewlouisbrockman: yeah im fine just pulling the server metrics i think, double checking can see

matthewlouisbrockman: yeah, not _too_ worried about the client->api hop at this point, seeing the history there fine starting with just measuring the existing telemetry from the server. closing this PR

matthewlouisbrockman: o, actually yeah taht's already there so this was just duplicating

PR #1956, opened 2026-02-20
chore(api): move auth to shared package

jakubno:
I didn't refactor the code and kept using the local APIError etc so there isn't that much changes, I'll follow up with clean up later

---

> [!NOTE]
> **Medium Risk**
> Refactors request authentication and token validation paths into a new module; behavior should be equivalent but mistakes could cause auth bypass/lockouts or incorrect HTTP status handling.
> 
> **Overview**
> Moves API authentication out of `packages/api` into a new reusable `packages/auth` module (auth middleware, token/key validation, team/user lookups, and caching), and switches the API server to use the new `AuthService` + `CreateAuthenticationFunc` while updating all call sites to use shared `auth` constants/types and centralized `APIError`/`SendAPIStoreError` helpers.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e1cde3c44f673a23cf3669c84e23b2e5f5aafd34. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/api/internal/db/auth_store.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"context"
+
+	"github.com/google/uuid"
+
+	"github.com/e2b-dev/infra/packages/api/internal/db/types"
+	authdb "github.com/e2b-dev/infra/packages/db/pkg/auth"
+)
+
+// AuthStoreImpl implements sharedauth.AuthStore[*types.Team] using the auth DB client.
```
dobrac: can you add the var _ = ...  for the implemented interface?

packages/api/internal/handlers/store.go
```diff
@@ -142,7 +142,9 @@ func NewAPIStore(ctx context.Context, tel *telemetry.Client, config cfg.Config)
 		logger.L().Fatal(ctx, "Initializing Orchestrator client", zap.Error(err))
 	}
 
-	authCache := authcache.NewTeamAuthCache()
+	authCache := sharedauth.NewAuthCache[*types.Team]()
+	authStore := dbapi.NewAuthStore(authDB)
```
dobrac: it would be better to abstract it in a way that we're requesting/validating the auth against a remote server. That way we could just swap the implementation easily

dobrac (approved):

packages/api/internal/handlers/store.go
jakubno: I'll do it in next PR

PR #1958, opened 2026-02-21
chore(api): refactor template caches

jakubno:
- Use alias cache to only resolve templateID to simplify the invalidation
- Split template cache to 2 separate caches based on the necessary information (template cache requires the template to have a ready template)  
- Split the tests

---

> [!NOTE]
> **Medium Risk**
> Changes cache key invalidation and public/cluster metadata retrieval paths, so stale-cache or access-control regressions are possible if alias keys or metadata caching diverge from DB state.
> 
> **Overview**
> Refactors template caching by separating *immutable* alias resolution (now returns only `templateID`/`teamID`) from *mutable* template metadata (new Redis-backed `TemplateMetadataCache` for `public` and `cluster_id`), and adds a `ResolveAliasWithMetadata` helper used by handlers and template creation flows.
> 
> Simplifies alias invalidation by removing the Redis reverse-index approach and instead returning alias cache keys from DB delete operations (`DeleteTemplate`, `DeleteOtherTemplateAliases`) so callers can explicitly invalidate affected alias keys, plus adds targeted invalidation to prevent stale negative-cache tombstones after builds/alias changes; tests are split/updated accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 649998c541ec0a44bb711426acdee6678493609f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/db/queries/delete_old_aliases.sql.go
+++ b/packages/db/queries/delete_old_aliases.sql.go
@@ -27,11 +27,11 @@ func (q *Queries) DeleteOtherTemplateAliases(ctx context.Context, envID string) 
 	defer rows.Close()
 	var items []string
 	for rows.Next() {
-		var aliasKey string
-		if err := rows.Scan(&aliasKey); err != nil {
+		var alias_key string
+		if err := rows.Scan(&alias_key); err != nil {
 			return nil, err
 		}
-		items = append(items, aliasKey)
+		items = append(items, alias_key)
 	}
 	if err := rows.Err(); err != nil {
 		return nil, err


packages/api/internal/cache/templates/alias_cache.go
```diff
@@ -10,30 +10,26 @@ import (
...
-	aliasReverseIndexPrefix = "template:alias-reverse"
+	aliasCacheKeyPrefix = "template:alias"
 )
 
-// AliasInfo holds resolved alias information
+// AliasInfo holds resolved alias information (immutable mapping data only).
+// Mutable metadata like Public is in TemplateMetadata.
 type AliasInfo struct {
 	TemplateID string    `json:"template_id"`
 	TeamID     uuid.UUID `json:"team_id"`
```
dobrac, line 32: the teamID is now duplicit between aliasInfo and metadata. Do we need it in both places?

packages/api/internal/cache/templates/template_metadata_cache.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	return metadata, nil
+}
+
+func (c *TemplateMetadataCache) fetchFromDB(ctx context.Context, templateID string) (*TemplateMetadata, error) {
+	ctx, span := tracer.Start(ctx, "fetch template metadata from DB")
+	defer span.End()
+
+	result, err := c.db.GetTemplateById(ctx, templateID)
```
dobrac, line 69: when starting a sandbox for the first time from the template, we're adding one more DB request, correct? 

So basically the GetTemplateByID will by called twice

packages/api/internal/cache/templates/alias_cache.go
jakubno: we could have it only in metadata, but in some places this prevents doing extra query and teamID is immutable

packages/api/internal/cache/templates/template_metadata_cache.go
jakubno: i don't think so, the template metadata isn't used there

packages/api/internal/cache/templates/alias_cache.go
dobrac: ok

packages/api/internal/cache/templates/template_metadata_cache.go
dobrac: Is it not? You need to have the information if you can access the template -> the Public field

jakubno: There's the whole tempalte with build object used

jakubno: It's basically the same as before now it's just separated to two calls onside the app (alias -> templateID -> template with build) before it was hidden

dobrac: ok then, thx for the explanation 🙏

PR #1964, opened 2026-02-23
chore(cache): add locking when updating the cache to prevent stale writes

jakubno:
Also changed build cache to call invalidate instead of directly updating the state to simplify the logic

---

> [!NOTE]
> **Medium Risk**
> Touches shared caching behavior and adds distributed locking, which can affect cache freshness/latency and behavior under contention or Redis issues across multiple services.
> 
> **Overview**
> Adds a distributed per-key lock to `RedisCache` writes (miss backfills, background refreshes, and deletes) to reduce cross-instance thundering-herd and stale overwrites, while degrading gracefully when the lock can’t be obtained. Template build status caching is simplified by removing the Redis-side `SetStatus` mutation and instead invalidating the cached build entry after DB status updates so the next read refetches fresh data.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 16a5614c491b28e1c641577ec60d51189d54509f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/shared/pkg/cache/redis.go
```diff
@@ -85,6 +116,16 @@ func (rc *RedisCache[V]) GetOrSet(ctx context.Context, key string, dataCallback
...
+
+		// Acquire distributed lock if enabled
+		lock, lockErr := rc.acquireLock(ctx, key, redislock.LinearBackoff(rc.config.LockRetryInterval))
+		if lockErr != nil {
+			logger.L().Warn(ctx, "RedisCache: failed to acquire lock, proceeding without lock",
+				zap.String("key", key),
+				zap.Error(lockErr))
+		}
+		defer rc.releaseLock(context.WithoutCancel(ctx), lock, key)
+
```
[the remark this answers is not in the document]
jakubno: This is on purpose to prevent stale data, we could use CAS with versions, but that sounds more complicated

packages/shared/pkg/cache/redis.go
```diff
@@ -103,6 +141,15 @@ func (rc *RedisCache[V]) GetOrSet(ctx context.Context, key string, dataCallback
...
 			storeKey = rc.config.ExtractKeyFunc(v)
 		}
+
+		if lockErr != nil {
+			logger.L().Warn(ctx, "RedisCache: failed to acquire refresh lock, skipping update in redis",
+				zap.String("key", key),
+				zap.Error(lockErr))
+
+			return result{value: v}, nil
+		}
```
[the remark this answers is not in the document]
jakubno: Shouldn't really happen, prevents stale reads

packages/shared/pkg/cache/redis.go
```diff
@@ -40,13 +46,21 @@ type RedisConfig[V any] struct {
...
+	// Defaults to 50ms.
+	LockRetryInterval time.Duration
 }
 
 // RedisCache is a generic two-tier cache: Redis + user callback.
 type RedisCache[V any] struct {
 	config       RedisConfig[V]
 	fetchGroup   singleflight.Group
 	redisRefresh singleflight.Group
+	lockClient   *redislock.Client // nil when locking disabled
```
djeebus: might be cleaner and less dangerous to implement via noop struct so we can avoid the nil checks

packages/shared/pkg/cache/redis.go
```diff
@@ -85,6 +113,16 @@ func (rc *RedisCache[V]) GetOrSet(ctx context.Context, key string, dataCallback
...
 	r, _, _ := rc.fetchGroup.Do(key, func() (any, error) {
 		ctx := context.WithoutCancel(ctx)
+
+		// Acquire distributed lock if enabled
+		lock, lockErr := rc.acquireLock(ctx, key, redislock.LinearBackoff(rc.config.LockRetryInterval))
+		if lockErr != nil {
+			logger.L().Warn(ctx, "RedisCache: failed to acquire lock, proceeding without lock",
+				zap.String("key", key),
+				zap.Error(lockErr))
+		}
```
djeebus: is this safe? why bother locking if you can move on when it's not been acquired?

jakubno: otherwise the request will fail, on lockErr we don't set the value
(and also all others waiting in the singleflight)

packages/shared/pkg/cache/redis.go
```diff
@@ -103,6 +137,15 @@ func (rc *RedisCache[V]) GetOrSet(ctx context.Context, key string, dataCallback
 		if rc.config.ExtractKeyFunc != nil {
 			storeKey = rc.config.ExtractKeyFunc(v)
 		}
+
+		if lockErr != nil {
```
sitole, line 142: Why is the lock err not checked immediately after acquire?

jakubno: Added comment
>		// We want to get the results even without the lock to prevent failing all the waiting requests

packages/shared/pkg/cache/redis.go
```diff
@@ -120,7 +163,21 @@ func (rc *RedisCache[V]) Set(ctx context.Context, key string, value V) {
...
+
+	ctx, cancel := context.WithTimeout(ctx, rc.config.RedisTimeout)
+	defer cancel()
+
+	redisKey := rc.RedisKey(key)
+	if err := rc.config.RedisClient.Del(ctx, redisKey).Err(); err != nil {
+		logger.L().Warn(ctx, "RedisCache: Redis DEL error",
+			zap.String("key", redisKey),
+			zap.Error(err))
+	}
```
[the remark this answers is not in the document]
jakubno: Added comment

PR #1989, opened 2026-02-25
fix: define UFFD_FEATURE_WP_ASYNC when missing

bchalios:
Debian bookworm ships linux header 6.1. UFFD_FEATURE_WP_ASYNC was introduced with 6.7. Help CGO by defining it when its missing. Value taken from here:
https://github.com/torvalds/linux/blob/master/include/uapi/linux/userfaultfd.h#L249C1-L249C40


sitole:
Looks like the unit test failed because some 3rd part url request failed. Re-round should be okay

PR #1990, opened 2026-02-25
Silence hostname missing red herring warning

bchalios:
We're trying `hostname` and then fall-back to `hostnamectl`. If the former fails it prints out a warning even though the latter succeeds. The failure logs something in stderr which sometimes is confusing. Redirect it stderr to /dev/null, and check the HOSTNAME variable to see if it's set and if not, fail.

### DIFF
--- a/packages/api/Makefile
+++ b/packages/api/Makefile
@@ -1,7 +1,8 @@
 ENV := $(shell cat ../../.last_used_env || echo "not-set")
 -include ../../.env.${ENV}
 PREFIX := $(strip $(subst ",,$(PREFIX)))
-HOSTNAME := $(shell hostname || hostnamectl hostname)
+HOSTNAME := $(shell hostname 2> /dev/null || hostnamectl hostname 2> /dev/null)
+$(if $(HOSTNAME),,$(error Failed to determine hostname: both 'hostname' and 'hostnamectl' failed))
 
 expectedMigration := $(shell ./../../scripts/get-latest-migration.sh)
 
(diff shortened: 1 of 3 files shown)


packages/api/Makefile
```diff
@@ -1,7 +1,8 @@
 ENV := $(shell cat ../../.last_used_env || echo "not-set")
 -include ../../.env.${ENV}
 PREFIX := $(strip $(subst ",,$(PREFIX)))
-HOSTNAME := $(shell hostname || hostnamectl hostname)
+HOSTNAME := $(shell hostname 2> /dev/null || hostnamectl hostname 2> /dev/null)
```
arkamar, line 4:
```suggestion
HOSTNAME ?= $(shell hostname 2> /dev/null || hostnamectl hostname 2> /dev/null)
```
Now, when you are on it, how about changing `:=` to `?=`, this would allow us having exported `HOSTNAME` in some situations and override the `hostname` call. Well, I am not sure if it would be useful, maybe in situations where none of them are for what ever reason available?

bchalios: Do you mean in case we don't have access to neither `hostname` and `hostnamectl` available in the system, or do you have another use case for this?

arkamar: Well, I am unable to come up with anything else than not having access to neither `hostname` nor `hostnamectl` right now. I don't have another use case for this. I guess `?=` is not worth it until anyone find need for it :)

PR #1999, opened 2026-02-26
fix(api): race conditions in connect and sandbox changing state

jakubno:
Simplifies the logic in `KeepAliveFor` and prevents some race conditions and overall simplifies the code  

---

> [!NOTE]
> **Medium Risk**
> Touches core sandbox lifecycle endpoints and `KeepAliveFor`, changing return types and HTTP error semantics (notably 409 conflicts during transitions), which could affect client behavior and reconnect flows under load. Access-control behavior is tightened in in-memory storage by enforcing `teamID` on `Get`/`Update`, reducing cross-team leakage risk but potentially surfacing new 404s where callers passed empty IDs.
> 
> **Overview**
> Improves sandbox connect/keepalive handling during transitions by making `KeepAliveFor` return the updated sandbox, distinguishing *not found* vs *not running* via new `NotRunningError`, and mapping transitional states to clearer 409 responses/messages; `connect` now retries a few times on state changes before falling back to snapshot resume, while `refresh`/`timeout` drop pre-`GetSandbox` checks and rely on `KeepAliveFor` for consistent ownership/errors, alongside centralized user-facing messages and stricter `teamID` enforcement in memory storage plus updated tests for the new timing/status behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4b402d0005662f87def8e2139b754f5ab23ad3f3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api/internal/handlers/sandbox_connect.go
+++ b/packages/api/internal/handlers/sandbox_connect.go
@@ -84,7 +84,7 @@ func (a *APIStore) PostSandboxesSandboxIDConnect(c *gin.Context, sandboxID api.S
 		}
 
 		if notRunningErr.State == sandbox.StateKilling {
-			a.sendAPIStoreError(c, http.StatusNotFound, utils.SandboxNotFoundMsg(sandboxID))
+			a.sendAPIStoreError(c, http.StatusConflict, utils.SandboxChangingStateMsg(sandboxID, notRunningErr.State))
 
 			return
 		}
--- a/packages/api/internal/orchestrator/keep_alive.go
+++ b/packages/api/internal/orchestrator/keep_alive.go
@@ -51,7 +51,7 @@ func (o *Orchestrator) KeepAliveFor(ctx context.Context, teamID uuid.UUID, sandb
 	if err != nil {
 		switch {
 		case errors.As(err, &sbxNotRunningErr):
-			return nil, &api.APIError{Code: http.StatusConflict, ClientMsg: utils.SandboxChangingStateMsg(sandboxID, string(sbxNotRunningErr.State)), Err: err}
+			return nil, &api.APIError{Code: http.StatusConflict, ClientMsg: utils.SandboxChangingStateMsg(sandboxID, sbxNotRunningErr.State), Err: err}
 		case errors.As(err, &sbxNotFoundErr):
 			return nil, &api.APIError{Code: http.StatusNotFound, ClientMsg: utils.SandboxNotFoundMsg(sandboxID), Err: err}
 		case errors.Is(err, errMaxInstanceLengthExceeded):
--- a/packages/api/internal/utils/messages.go
+++ b/packages/api/internal/utils/messages.go
@@ -1,11 +1,24 @@
 package utils
 
-import "fmt"
+import (
+	"fmt"
+
+	"github.com/e2b-dev/infra/packages/api/internal/sandbox"
+)
 
 func SandboxNotFoundMsg(sandboxID string) string {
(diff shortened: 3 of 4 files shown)


packages/api/internal/handlers/sandbox_connect.go
```diff
@@ -58,54 +58,52 @@ func (a *APIStore) PostSandboxesSandboxIDConnect(c *gin.Context, sandboxID api.S
...
-				zap.String("node_id", sandboxData.NodeID),
-			)
+		}
 
-			apiErr := a.orchestrator.KeepAliveFor(ctx, teamID, sandboxID, timeout, false)
-			if apiErr != nil {
-				logger.L().Error(ctx, "Error when resuming sandbox", zap.Error(apiErr.Err))
-				a.sendAPIStoreError(c, apiErr.Code, apiErr.ClientMsg)
+		if notRunningErr.State == sandbox.StateKilling {
+			a.sendAPIStoreError(c, http.StatusNotFound, sandboxNotFoundMsg(sandboxID))
```
djeebus: We should return a more details error message than just "sandbox not found". Maybe "sandbox is being paused; try again later"?

djeebus: Alternately, if Killing is temporary, maybe we should wait until it's done and then report "sandbox not found" or "sandbox paused", so that the customer knows they need to start the sandbox again.

packages/api/internal/orchestrator/keep_alive.go
```diff
@@ -44,30 +45,30 @@ func (o *Orchestrator) KeepAliveFor(ctx context.Context, teamID uuid.UUID, sandb
 	}
 
 	var sbxNotFoundErr *sandbox.NotFoundError
+	var sbxNotRunningErr *sandbox.NotRunningError
 	sbx, err := o.sandboxStore.Update(ctx, teamID, sandboxID, updateFunc)
 	if err != nil {
 		switch {
+		case errors.As(err, &sbxNotRunningErr):
+			return nil, &api.APIError{Code: http.StatusNotFound, ClientMsg: "Sandbox not found", Err: err}
```
djeebus: We should return a more detailed error message here. "Sandbox state is %s, not Running" or something like that.

packages/api/internal/handlers/sandbox_connect.go
jakubno: Killing is forever

packages/api/internal/orchestrator/keep_alive.go
jakubno: I suggest to handle this better in another PR, we can do something similar as in `conenct`

jakubno: Changed the message to explain what's happening + changed status code to 409

packages/api/internal/orchestrator/keep_alive.go
```diff
@@ -34,7 +35,8 @@ func (o *Orchestrator) KeepAliveFor(ctx context.Context, teamID uuid.UUID, sandb
 		}
 
 		if !allowShorter && endTime.Before(sbx.EndTime) {
-			return sbx, sandbox.ErrCannotShortenTTL
+			// If shorter than the current end time, we don't extend, so we can return
+			return sbx, nil
```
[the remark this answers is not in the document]
jakubno: we'll remove this soon, not such issue anyway

PR #2003, opened 2026-02-26
feat(envd): set higher CPU priority for envd and reset it for child processes

arkamar:
Set Nice=-20 in envd.service to improve responsiveness under load. Reset nice to the default level (0) for user processes, PTY processes, and socat port-forwarding processes so only envd itself benefits from the elevated priority.

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes process scheduling priority at the systemd level and actively adjusts child process niceness, which could affect responsiveness and behavior under load if `setpriority` fails or is applied inconsistently.
> 
> **Overview**
> Runs `envd` with elevated CPU priority (`Nice=-20`) while explicitly resetting spawned process nice levels back to normal (0) after start, so only the daemon retains the higher priority. Also bumps the reported `envd` version to `0.5.4`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit aed26be1ea0c604f175f623a870d2bd483101746. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/envd/main.go
+++ b/packages/envd/main.go
@@ -47,7 +47,7 @@ const (
 )
 
 var (
-	Version = "0.5.3"
+	Version = "0.5.4"
 
 	commitSHA string
 


packages/envd/internal/port/forward.go
```diff
@@ -167,6 +167,11 @@ func (f *Forwarder) startPortForwarding(ctx context.Context, p *PortToForward) {
 		return
 	}
 
+	// Reset nice to default (0) so socat doesn't inherit envd's high priority (Nice=-20).
+	if niceErr := syscall.Setpriority(syscall.PRIO_PROCESS, cmd.Process.Pid, 0); niceErr != nil {
```
jakubno: I think `socat` could need lower `nice` value as well, without `socat` you'll lose the connection and it may be again hard to debug the issue

packages/envd/internal/services/process/handler/handler.go
```diff
@@ -170,6 +170,10 @@ func New(
 			return nil, connect.NewError(connect.CodeInvalidArgument, fmt.Errorf("error starting pty with command '%s' in dir '%s' with '%d' cols and '%d' rows: %w", cmd, cmd.Dir, req.GetPty().GetSize().GetCols(), req.GetPty().GetSize().GetRows(), err))
 		}
 
+		if niceErr := resetNice(cmd.Process.Pid); niceErr != nil {
```
jakubno: is this needed? isn't it enough in `Start()`, how is tty different?

arkamar: It is not needed. I just didn't realize that `pty.StartWithSize()` actually calls `cmd.Start()`. Thanks.

packages/envd/internal/port/forward.go
arkamar: Yes, you are right. Maybe it is not necessary to change it at all. Now, thinking more about it, we can probably keep the same nice as it is an infrastructure critical process.

PR #2007, opened 2026-02-27
feature flag removing the local cache on redis call

matthewlouisbrockman:
adds a feature flag around the local 500ms local cache

- Currently, if we delete/pause a sandbox, we remove it from the catalog. However, the client proxies don't have a way to easily invalidate the cache if they've talked to it in the last 500ms. This leads to a stale condition for whether or not a sandbox is active or not.
- Current load balancer does a round robin on incoming requests, so the cache is only effective for when the load balancer happens to send the request to a client proxy that accessed that sandbox in the last 500ms. leads to more randomness than it appears to help, but this lets us check if things blow up 

---

> [!NOTE]
> **Low Risk**
> Behavior remains the same by default (flag defaults to enabled), but flipping the flag will increase Redis read traffic/latency and any missed wiring of the feature-flags client would now cause a nil dereference in `GetSandbox`.
> 
> **Overview**
> Adds a new `sandbox-catalog-local-cache` feature flag and plumbs the feature-flags client into the Redis-backed sandbox catalog so `GetSandbox` can optionally bypass the 500ms in-process TTL cache and always hit Redis when the flag is disabled.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6197b73d3ca2acb3e6716efe3217d336d7e072fc. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api/internal/orchestrator/orchestrator.go
+++ b/packages/api/internal/orchestrator/orchestrator.go
@@ -84,7 +84,7 @@ func New(
 
 	var routingCatalog e2bcatalog.SandboxesCatalog
 	if redisClient != nil {
-		routingCatalog = e2bcatalog.NewRedisSandboxesCatalog(redisClient, nil)
+		routingCatalog = e2bcatalog.NewRedisSandboxesCatalog(redisClient, featureFlags)
 	} else {
 		routingCatalog = e2bcatalog.NewMemorySandboxesCatalog()
 	}
--- a/packages/shared/pkg/sandbox-catalog/catalog_redis.go
+++ b/packages/shared/pkg/sandbox-catalog/catalog_redis.go
@@ -48,7 +48,7 @@ func (c *RedisSandboxCatalog) GetSandbox(ctx context.Context, sandboxID string) 
 	spanCtx, span := tracer.Start(ctx, "sandbox-catalog-get")
 	defer span.End()
 
-	useLocalCache := c.featureFlags == nil || c.featureFlags.BoolFlag(spanCtx, featureflags.SandboxCatalogLocalCacheFlag)
+	useLocalCache := c.featureFlags.BoolFlag(spanCtx, featureflags.SandboxCatalogLocalCacheFlag)
 
 	if useLocalCache {
 		sandboxInfo := c.cache.Get(sandboxID)


packages/api/internal/orchestrator/orchestrator.go
```diff
@@ -84,7 +84,7 @@ func New(
 
 	var routingCatalog e2bcatalog.SandboxesCatalog
 	if redisClient != nil {
-		routingCatalog = e2bcatalog.NewRedisSandboxesCatalog(redisClient)
```
dobrac, line 87: note, we're not using this one yet afaik

packages/shared/pkg/sandbox-catalog/catalog_redis.go
```diff
@@ -45,9 +48,13 @@ func (c *RedisSandboxCatalog) GetSandbox(ctx context.Context, sandboxID string)
 	spanCtx, span := tracer.Start(ctx, "sandbox-catalog-get")
 	defer span.End()
 
-	sandboxInfo := c.cache.Get(sandboxID)
-	if sandboxInfo != nil {
-		return sandboxInfo.Value(), nil
+	useLocalCache := c.featureFlags == nil || c.featureFlags.BoolFlag(spanCtx, featureflags.SandboxCatalogLocalCacheFlag)
```
dobrac: I would do the featureFlags nil check everywhere. The AI is overly cautios about these

packages/api/internal/orchestrator/orchestrator.go
```diff
@@ -84,7 +84,7 @@ func New(
...
+		routingCatalog = e2bcatalog.NewRedisSandboxesCatalog(redisClient, nil)
```
dobrac: lets pass the featureFlagsClient

packages/shared/pkg/sandbox-catalog/catalog_redis.go
```diff
@@ -45,9 +48,13 @@ func (c *RedisSandboxCatalog) GetSandbox(ctx context.Context, sandboxID string)
 	spanCtx, span := tracer.Start(ctx, "sandbox-catalog-get")
 	defer span.End()
 
-	sandboxInfo := c.cache.Get(sandboxID)
-	if sandboxInfo != nil {
-		return sandboxInfo.Value(), nil
+	useLocalCache := c.featureFlags.BoolFlag(spanCtx, featureflags.SandboxCatalogLocalCacheFlag)
```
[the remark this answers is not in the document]
matthewlouisbrockman: we hcanged it, not nil

matthewlouisbrockman:
@cursor review

packages/api/internal/orchestrator/orchestrator.go
matthewlouisbrockman: I think we are

dobrac (approved):

PR #2028, opened 2026-02-28
fix(envd): eliminate OOM score and nice priority race in child processes

arkamar:
## Summary

- Wraps user commands in a shell that sets `oom_score_adj` and resets nice priority
  **before** `exec`, eliminating the race where grandchildren could inherit envd's
  protected OOM score (-1000) or elevated CPU priority (nice -20)
- Sets `Nice=-20` on the envd systemd service for improved daemon responsiveness under load
- Logs the user's original command instead of the internal wrapper in all error messages
  and structured logs

## Motivation

Previously, `adjustOomScore` and `resetNice` were called **after** `cmd.Start()`,
leaving a race window where the child process (and any grandchildren it spawns quickly)
could run with envd's inherited OOM score of -1000 and nice of -20. Under specific
scenarios — e.g. shells that immediately fork subprocesses — grandchildren would be
born before the parent's values were corrected, making them effectively immune to the
OOM killer and running at elevated CPU priority.

The fix uses an inline `/bin/sh` wrapper:
```sh
echo 100 > /proc/$$/oom_score_adj && exec nice -n 0 "$@"
```

This sets the OOM score and nice value atomically before `exec`-ing the user's command,
so all descendants inherit the correct values from the very first instruction.

## Note on the logging commit

The commit that strips the wrapper from logged commands (`cf306862e`) can be excluded
from this PR if desired. It is included because without it, users would see confusing
log output like `/bin/sh -c echo 100 > /proc/$$/oom_score_adj && exec nice -n 0 ...`
instead of their actual command. The tradeoff is that if the command fails inside the
wrapper (e.g. `/bin/sh` itself errors), the error message won't show the full wrapped
command. I believe the cleaner user experience is worth it in practice.

## Obsoletes

Obsoletes #2003 — this PR subsumes the nice priority feature from that PR and fixes
the race condition in its `resetNice` approach.

Closes: #2003

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit 93360127dc83a8a81f3b8b7d20eefd406d3d96ea. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/envd/internal/services/process/handler/handler.go
+++ b/packages/envd/internal/services/process/handler/handler.go
@@ -71,6 +71,17 @@ func (p *Handler) userCommand() string {
 	return strings.Join(append([]string{p.Config.GetCmd()}, p.Config.GetArgs()...), " ")
 }
 
+// currentNice returns the nice value of the current process.
+func currentNice() int {
+	prio, err := syscall.Getpriority(syscall.PRIO_PROCESS, 0)
+	if err != nil {
+		return 0
+	}
+
+	// Getpriority returns 20 - nice on Linux.
+	return 20 - prio
+}
+
 func New(
 	ctx context.Context,
 	user *user.User,
@@ -86,7 +97,9 @@ func New(
 	// Wrap the command in a shell that sets the OOM score and nice value before exec-ing the actual command.
 	// This eliminates the race window where grandchildren could inherit the parent's protected OOM score (-1000)
 	// or high CPU priority (nice -20) before the post-start calls had a chance to correct them.
-	oomWrapperScript := fmt.Sprintf(`echo %d > /proc/$$/oom_score_adj && exec nice -n %d "${@}"`, defaultOomScore, defaultNice)
+	// nice(1) applies a relative adjustment, so we compute the delta from the current (inherited) nice to the target.
+	niceDelta := defaultNice - currentNice()
+	oomWrapperScript := fmt.Sprintf(`echo %d > /proc/$$/oom_score_adj && exec /usr/bin/nice -n %d "${@}"`, defaultOomScore, niceDelta)
 	wrapperArgs := append([]string{"-c", oomWrapperScript, "--", req.GetProcess().GetCmd()}, req.GetProcess().GetArgs()...)
 	cmd := exec.CommandContext(ctx, "/bin/sh", wrapperArgs...)
 


packages/envd/internal/services/process/handler/handler.go
```diff
@@ -72,7 +91,17 @@ func New(
...
-	cmd := exec.CommandContext(ctx, req.GetProcess().GetCmd(), req.GetProcess().GetArgs()...)
+	// User command string for logging (without the internal wrapper details).
+	userCmd := strings.Join(append([]string{req.GetProcess().GetCmd()}, req.GetProcess().GetArgs()...), " ")
+
+	// Wrap the command in a shell that sets the OOM score and nice value before exec-ing the actual command.
+	// This eliminates the race window where grandchildren could inherit the parent's protected OOM score (-1000)
+	// or high CPU priority (nice -20) before the post-start calls had a chance to correct them.
+	// nice(1) applies a relative adjustment, so we compute the delta from the current (inherited) nice to the target.
+	niceDelta := defaultNice - currentNice()
+	oomWrapperScript := fmt.Sprintf(`echo %d > /proc/$$/oom_score_adj && exec /usr/bin/nice -n %d "${@}"`, defaultOomScore, niceDelta)
```
dobrac, line 102: Is the user under which the command runs always able to execute it and change it? We're not running all commands as root.

dobrac (changes requested):

packages/envd/internal/services/process/handler/handler.go
arkamar: Non-root users can increase both values, the OOM score and nice. I have checked and tested it. Btw, this should be covered by our tests as well, right?

So, there is an edge case, when `root` will change `envd` OOM score value to higher value that our `defaultOomScore`, then the echo will fail, because non-privileged user is not allowed to decrease the value lower than the value set by privileged user.

> The value of /proc/<pid>/oom_score_adj may be reduced no lower than the last value set by a CAP_SYS_RESOURCE process. To reduce the value any lower requires CAP_SYS_RESOURCE.

The nice behaves similarly. However, this can happen only if the sandbox user will temper with `envd` setting. But maybe we can write the oneliner to suppress errors and continue.

The only correct way how to do it properly is to do a fork, set OOM score and nice as a privileged user, drop the privileges and exec to user command, but there is no way how to do it with `os/exec.Cmd` in Go.

dobrac: > Non-root users can increase both values, the OOM score and nice. I have checked and tested it. Btw, this should be covered by our tests as well, right?

You're right, we should have tests for both `root` and `user`.

> defaultOomScore, then the echo will fail, because non-privileged user is not allowed to decrease the value lower than the value set by privileged user.

Thx. I think this is fine then. If you tamper with envd, I think you should expect errors.


> The only correct way how to do it properly is to do a fork, set OOM score and nice as a privileged user, drop the privileges and exec to user command, but there is no way how to do it with os/exec.Cmd in Go.

If this would be an issue, we could use CGo if it's not possible using C in the future. No need to do this now.

PR #2032, opened 2026-03-02
feat(orchestrator): implement direct peer-to-peer chunk transfer

dobrac:
Add peer-to-peer chunks streaming between orchestrators while pause/resume and snapshots are uploading to GCS. Redis is used to keep the routing status if the build is on the orchestrator (and where), or not -> fallback to GCS (NFS) - the regular flow.

This allows to upload the pause/resume and snapshots asynchronously while not requiring to resume on the same node. And they can be available right away.

```mermaid
flowchart TB
    subgraph OrcA["Orchestrator A (Snapshot Creator)"]
        pause["Pause / Checkpoint sandbox"]
        cache["Cache snapshot locally<br/>(memfile, rootfs, snapfile, metadata, headers)"]
        register["Register in Redis<br/>SET peer:&lt;buildID&gt; &lt;addrA&gt;<br/>TTL = uploadTimeout + 2min"]
        asyncUpload["Start async GCS upload<br/>(goroutine)"]
        chunkService["ChunkService gRPC server<br/>GetBuildSeekable (streaming)<br/>GetBuildBlob (streaming)<br/>GetBuildFileSize / Exists"]
        uploadDone["Upload complete"]
        markUploaded["uploadedBuilds.Store(buildID)<br/>→ ChunkService returns use_storage=true"]
        unregister["Redis DEL peer:&lt;buildID&gt;<br/>+ auto-expires after 22min as safety net"]
    end

    subgraph Redis["Redis"]
        redisKey["peer:&lt;buildID&gt; → orchestrator-A:5000"]
    end

    subgraph GCS["GCS (Cloud Storage)"]
        bucket["Template bucket<br/>(memfile, rootfs, snapfile,<br/>metadata, headers)"]
    end

    subgraph OrcB["Orchestrator B (Resume Consumer)"]
        resume["Resume sandbox<br/>→ GetTemplate(buildID)"]
        routing["routingProvider.OpenSeekable / OpenBlob<br/>extract buildID from storage path"]
        resolve["Resolver.resolve(buildID)"]
        redisLookup["Redis GET peer:&lt;buildID&gt;"]
        decision{Peer found<br/>in Redis?}
        dial["Dial Orchestrator A gRPC"]
        peerRead["peerSeekable.ReadAt / peerBlob.Stream<br/>→ gRPC stream from Orchestrator A"]
        checkAvail{Peer response?}
        gotData["Use streamed data ✓"]
        fallbackGCS["Fallback → NFS Cache → GCS"]
        flipFlag["Set uploaded flag<br/>(skip peer for all future reads)"]
    end

    pause --> cache --> register
    register --> redisKey
    cache --> asyncUpload
    cache --> chunkService

    asyncUpload -.->|"uploading..."| bucket
    asyncUpload --> uploadDone
    uploadDone --> markUploaded --> unregister
    unregister -.->|"DEL"| redisKey

    resume --> routing --> resolve --> redisLookup
    redisLookup -.->|"GET"| redisKey
    redisLookup --> decision

    decision -->|"No (key expired<br/>or upload done)"| fallbackGCS
    decision -->|"Yes → addrA"| dial
    dial --> peerRead
    peerRead -.->|"gRPC stream"| chunkService
    peerRead --> checkAvail

    checkAvail -->|"data available"| gotData
    checkAvail -->|"not_available"| fallbackGCS
    checkAvail -->|"use_storage<br/>(upload finished)"| flipFlag
    flipFlag --> fallbackGCS

    fallbackGCS -.->|"read"| bucket
```

This mechanics should also handle correctly nested layers not uploaded yet to a persistent storage. Here is an example of that flow.

```mermaid
sequenceDiagram
    participant A as Orchestrator A
    participant Redis
    participant GCS
    participant B as Orchestrator B

    Note over A: Checkpoint 1 → build_1
    A->>Redis: SET peer:build_1 → addrA
    A->>GCS: Upload build_1 (slow)

    Note over A: Checkpoint 2 → build_2 (child of build_1)
    A->>Redis: SET peer:build_2 → addrA
    A->>GCS: Upload build_2 (slow)

    Note over A: Checkpoint 3 → build_3 (child of build_2)
    A->>Redis: SET peer:build_3 → addrA
    A->>GCS: Upload build_3 (fast!)

    GCS-->>A: build_3 upload done
    A->>Redis: DEL peer:build_3

    Note over B: Resume build_3
    B->>GCS: Fetch build_3 ✓ (in GCS)
    B->>Redis: GET peer:build_2
    Redis-->>B: addrA
    B->>A: P2P fetch build_2 ✓
    B->>Redis: GET peer:build_1
    Redis-->>B: addrA
    B->>A: P2P fetch build_1 ✓
    Note over B: Full chain resolved — resume works ✓
```

---

> [!NOTE]
> **High Risk**
> High risk because it changes sandbox `Pause`/`Checkpoint` snapshot upload lifecycle (including optional async completion) and introduces new cross-node gRPC streaming + Redis coordination that could affect availability, data consistency, and security if misconfigured.
> 
> **Overview**
> This PR enables **peer-to-peer streaming of snapshot/template files between orchestrators during the upload window**, by adding a new gRPC `ChunkService` that serves cached diffs/blobs from the local template cache and a Redis-backed `peerclient` resolver/routing `StorageProvider` that transparently falls back to persistent storage once the upload completes. It updates `Pause` to always upload in the background, adds a feature-flagged async `Checkpoint` mode, and adjusts checkpoint metadata handling to write prefetch mappings into the locally cached `metadata.json` so it is uploaded in the same pass. It also wires in `NODE_IP`/node address configuration for peer discovery, tracks upload completion to flip peers to `use_storage`, and extends snapshot-template persistence to store `origin_node_id` (and `build_id`) in the database.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ccc10821c1f31c3ca1883bfc18867a62f36a0132. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/orchestrator/internal/server/chunks.go
+++ b/packages/orchestrator/internal/server/chunks.go
@@ -50,7 +50,7 @@ func toGRPCError(err error) error {
 func (s *Server) GetBuildFileSize(ctx context.Context, req *orchestrator.GetBuildFileSizeRequest) (*orchestrator.GetBuildFileSizeResponse, error) {
 	telemetry.SetAttributes(ctx, telemetry.WithBuildID(req.GetBuildId()), attribute.String("file_name", req.GetFileName()))
 
-	if _, uploaded := s.uploadedBuilds.Load(req.GetBuildId()); uploaded {
+	if s.uploadedBuilds.Get(req.GetBuildId()) != nil {
 		telemetry.SetAttributes(ctx, attribute.Bool("uploaded", true))
 
 		return &orchestrator.GetBuildFileSizeResponse{Availability: peerUseStorage}, nil
@@ -78,7 +78,7 @@ func (s *Server) GetBuildFileSize(ctx context.Context, req *orchestrator.GetBuil
 func (s *Server) GetBuildFileExists(ctx context.Context, req *orchestrator.GetBuildFileExistsRequest) (*orchestrator.GetBuildFileExistsResponse, error) {
 	telemetry.SetAttributes(ctx, telemetry.WithBuildID(req.GetBuildId()), attribute.String("file_name", req.GetFileName()))
 
-	if _, uploaded := s.uploadedBuilds.Load(req.GetBuildId()); uploaded {
+	if s.uploadedBuilds.Get(req.GetBuildId()) != nil {
 		telemetry.SetAttributes(ctx, attribute.Bool("uploaded", true))
 
 		return &orchestrator.GetBuildFileExistsResponse{Availability: peerUseStorage}, nil
@@ -116,7 +116,7 @@ func (s *Server) ReadAtBuildSeekable(req *orchestrator.ReadAtBuildSeekableReques
 		attribute.Int64("length", req.GetLength()),
 	)
 
-	if _, uploaded := s.uploadedBuilds.Load(req.GetBuildId()); uploaded {
+	if s.uploadedBuilds.Get(req.GetBuildId()) != nil {
 		telemetry.SetAttributes(ctx, attribute.Bool("uploaded", true))
 
 		return stream.Send(&orchestrator.ReadAtBuildSeekableResponse{Availability: peerUseStorage})
@@ -151,7 +151,7 @@ func (s *Server) GetBuildBlob(req *orchestrator.GetBuildBlobRequest, stream orch
 		attribute.String("file_name", req.GetFileName()),
 	)
 
-	if _, uploaded := s.uploadedBuilds.Load(req.GetBuildId()); uploaded {
+	if s.uploadedBuilds.Get(req.GetBuildId()) != nil {
 		telemetry.SetAttributes(ctx, attribute.Bool("uploaded", true))
 
 		return stream.Send(&orchestrator.GetBuildBlobResponse{Availability: peerUseStorage})
--- a/packages/orchestrator/internal/server/main.go
+++ b/packages/orchestrator/internal/server/main.go
@@ -3,7 +3,9 @@ package server
 import (
 	"context"
 	"sync"
+	"time"
 
+	"github.com/jellydator/ttlcache/v3"
 	"go.opentelemetry.io/otel/metric"
 	"go.uber.org/zap"
 	"golang.org/x/sync/semaphore"
@@ -24,6 +26,10 @@ import (
 	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
 )
 
+// Matches the template cache TTL so entries live as long as the
+// templates they refer to and are cleaned up automatically.
+const uploadedBuildsTTL = 1 * time.Hour
+
 type Server struct {
 	orchestrator.UnimplementedSandboxServiceServer
 	orchestrator.UnimplementedChunkServiceServer
@@ -42,7 +48,7 @@ type Server struct {
 	sbxEventsService  *events.EventsService
 	startingSandboxes *semaphore.Weighted
 	peerRegistry      peerclient.Registry
-	uploadedBuilds    sync.Map // buildID → struct{}: builds whose GCS upload is done
+	uploadedBuilds    *ttlcache.Cache[string, struct{}]
 }
 
 type ServiceConfig struct {
@@ -62,6 +68,11 @@ type ServiceConfig struct {
 }
 
 func New(ctx context.Context, cfg ServiceConfig) *Server {
+	uploadedBuilds := ttlcache.New[string, struct{}](
+		ttlcache.WithTTL[string, struct{}](uploadedBuildsTTL),
+	)
+	go uploadedBuilds.Start()
+
 	server := &Server{
 		config:            cfg.Config,
 		sandboxFactory:    cfg.SandboxFactory,
@@ -76,6 +87,7 @@ func New(ctx context.Context, cfg ServiceConfig) *Server {
 		sbxEventsService:  cfg.SbxEventsService,
 		startingSandboxes: semaphore.NewWeighted(maxStartingInstancesPerNode),
 		peerRegistry:      cfg.PeerRegistry,
+		uploadedBuilds:    uploadedBuilds,
 	}
 
 	meter := cfg.Tel.MeterProvider.Meter("orchestrator.sandbox")
--- a/packages/orchestrator/internal/server/sandboxes.go
+++ b/packages/orchestrator/internal/server/sandboxes.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/google/uuid"
+	"github.com/jellydator/ttlcache/v3"
 	"github.com/launchdarkly/go-sdk-common/v3/ldcontext"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
@@ -642,8 +643,8 @@ func (s *Server) snapshotAndCacheSandbox(
 		}
 
 		completeUpload := func(ctx context.Context) {
-			// Mark in-memory so ChunkService tells peers to switch to GCS.
-			s.uploadedBuilds.Store(meta.Template.BuildID, struct{}{})
+			// Signal in-flight peer streams to switch to GCS.
+			s.uploadedBuilds.Set(meta.Template.BuildID, struct{}{}, ttlcache.DefaultTTL)
 
 			// Remove from Redis so new nodes go directly to GCS.
 			if err := s.peerRegistry.Unregister(ctx, meta.Template.BuildID); err != nil {
--- a/packages/orchestrator/internal/sandbox/template/peerclient/seekable.go
+++ b/packages/orchestrator/internal/sandbox/template/peerclient/seekable.go
@@ -33,7 +33,9 @@ func (s *peerSeekable) Size(ctx context.Context) (int64, error) {
 				return peerAttempt[int64]{value: resp.GetTotalSize(), hit: true}, nil
 			}
 
-			logger.L().Warn(ctx, "failed to get build file size from peer", logger.WithBuildID(s.buildID), zap.Error(err))
+			if err != nil {
+				logger.L().Warn(ctx, "failed to get build file size from peer", logger.WithBuildID(s.buildID), zap.Error(err))
+			}
 
 			return peerAttempt[int64]{}, nil
 		},


packages/orchestrator/internal/sandbox/build/cache.go
```diff
@@ -130,6 +130,17 @@ func (s *DiffStore) Has(d Diff) bool {
 	return s.cache.Has(d.CacheKey())
 }
 
+// Lookup returns the cached Diff for the given key without initialising a new one.
```
levb, line 133: At the risk of making the PR even bigger, 0/5 rename `Get` to `GetOrInit`(?), and `Lookup` to `Get`

packages/orchestrator/internal/sandbox/template/peerserver/file.go
```diff
@@ -0,0 +1,109 @@
...
+	"fmt"
+	"io"
+	"os"
+
+	tmpl "github.com/e2b-dev/infra/packages/orchestrator/internal/sandbox/template"
+	"github.com/e2b-dev/infra/packages/shared/pkg/storage"
+)
+
+// chunkSendSize is the maximum number of bytes per Send call.
+const chunkSendSize = storage.MemoryChunkSize
```
levb, line 14: Since we are moving to MMAP caching in 2MB chunks rather than 4, does it make sense to move 4MB at a time in this PR? Can we move to 2MB for the peer transfer? (The motivation was to reduce the MMAP "pressure", not the remote i/o optimization.

Also, are we streaming just the dirty blocks? If so, maybe a much smaller max(256Kb, blockSize) would serve better?

packages/orchestrator/internal/sandbox/template/peerserver/header.go
```diff
@@ -0,0 +1,46 @@
...
+	defer span.End()
+
+	device, err := f.getDevice(ctx)
+	if err != nil {
+		span.RecordError(err)
+
+		return fmt.Errorf("get device: %w", err)
+	}
+
+	h := device.Header()
```
levb, line 33: NS: Why does the header come from the ReadOnly device? because that's what we have in the local cache/go memory space that already has the header?  But with compression we will not store the headers until the files are uploaded, maybe a race there coming?

packages/orchestrator/internal/sandbox/template/peerprovider/helpers_test.go
```diff
@@ -0,0 +1,28 @@
+package peerprovider
```
levb: 0/5 Does this need to be a separate file or can these functions go to one of the primary test files?

packages/shared/pkg/storage/storage_cache.go
```diff
@@ -22,6 +22,23 @@ const (
 	cacheDirPermissions  = 0o700
 )
 
+// skipCacheWritebackKeyType is the context key type for skipping NFS cache writeback.
+type skipCacheWritebackKeyType struct{}
+
+// WithSkipCacheWriteback returns a context that signals the NFS cache layer to
```
levb, line 28: NS: why is it bad to NFS-cache prefetched reads? One would think they'd be reusable, too.

packages/orchestrator/chunks.proto
```diff
@@ -0,0 +1,74 @@
...
+  string build_id = 1;
+  // file_name is one of: "snapfile", "metadata.json"
+  string file_name = 2;
+}
+
+message GetBuildFileExistsResponse {
+  PeerAvailability availability = 1;
+}
+
+message GetBuildSeekableRequest {
```
levb: The name should reflect `ReadAt` since that is what it proxies, right?

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -505,22 +511,39 @@ func (s *Server) Checkpoint(ctx context.Context, in *orchestrator.SandboxCheckpo
 	// Setup lifecycle for the resumed sandbox
 	s.setupSandboxLifecycle(ctx, resumedSbx)
 
-	// Upload prefetch mapping in background
+	// Embed prefetch data into the metadata so it's uploaded with the snapshot files in a single pass.
```
levb, line 515: NS - need an explanation

packages/orchestrator/internal/sandbox/template/peerprovider/peerprovider.go
```diff
@@ -0,0 +1,43 @@
...
+var tracer = otel.Tracer("github.com/e2b-dev/infra/packages/orchestrator/internal/sandbox/template/peerprovider")
+
+var (
+	// ErrNotSupported is returned when a source type does not implement an operation.
+	ErrNotSupported = errors.New("operation not supported")
+	// ErrNotAvailable is returned when the requested build is not in the local peer cache.
+	ErrNotAvailable = errors.New("not available in local peer cache")
+)
+
+// Sender receives chunks of file data for streaming to a caller.
```
levb: 0/5 make the comment more intention-revealing, re: "_Sender receives_ chunks..."

packages/orchestrator/internal/sandbox/template/peerserver/peerserver.go
```diff
@@ -0,0 +1,43 @@
...
+	GetCachedTemplate(buildID string) (tmpl.Template, bool)
+}
+
+// BlobSource serves whole-file reads and existence checks (snapfile, metadata, headers).
+type BlobSource interface {
+	Stream(ctx context.Context, sender Sender) error
+	Exists(ctx context.Context) (bool, error)
+}
+
+// SeekableSource serves random-access reads with offset/length and size queries (memfile, rootfs).
```
levb, line 39: Please note if the requests are block-aligned (I think they must be)

packages/orchestrator/internal/sandbox/template/cache.go
```diff
@@ -60,12 +63,15 @@ func NewCache(
...
 	persistence storage.StorageProvider,
 	metrics blockmetrics.Metrics,
+	peers peerstorage.Resolver,
 ) (*Cache, error) {
 	cache := ttlcache.New(
 		ttlcache.WithTTL[string, Template](templateExpiration),
 	)
 
 	cache.OnEviction(func(ctx context.Context, _ ttlcache.EvictionReason, item *ttlcache.Item[string, Template]) {
+		peers.PurgeUploaded(item.Key())
```
levb: if I understand this correctly, 0/5 rename `PurgeUploaded` with `Purge`, since in this case we are purging because of the cache eviction, not because the file was uploaded? Or is there another `Purge()` already?

packages/orchestrator/internal/sandbox/template/peerclient/resolver.go
```diff
@@ -0,0 +1,162 @@
...
+	addr, found, err := r.readPeerAddress(ctx, buildID)
+	if err != nil {
+		return attrResolveRedisError, resolveResult{}
+	}
+
+	if !found {
+		return attrResolveNoPeer, resolveResult{}
+	}
+
+	if r.isSelfAddress(addr) {
```
levb, line 140: How can this happen?

packages/orchestrator/internal/server/chunks.go
```diff
@@ -0,0 +1,178 @@
...
+
+	size, err := src.Size(ctx)
+	if err != nil {
+		return nil, toGRPCError(err)
+	}
+
+	return &orchestrator.GetBuildFileSizeResponse{TotalSize: size}, nil
+}
+
+func (s *Server) GetBuildFileExists(ctx context.Context, req *orchestrator.GetBuildFileExistsRequest) (*orchestrator.GetBuildFileExistsResponse, error) {
```
levb, line 78: What necessitated the entire `GetBuildFileExists`? I think our exists methods just wrap size() > 0?

packages/orchestrator/internal/sandbox/template/peerprovider/file.go
```diff
@@ -0,0 +1,109 @@
+package peerprovider
```
levb: 1/5 `peerserver` and the other `peerclient`

levb (changes requested):
Minor comments/questions

packages/orchestrator/internal/sandbox/template/peerserver/file.go
dobrac: in this case, the chunking is used only to split the whole files headers. I can put there 2MB, no issue with that - this just reused the constant.

For the diffs, it uses the blockSize to chunk the answer + there is a minimum set

packages/orchestrator/internal/sandbox/template/peerserver/header.go
dobrac: > Why does the header come from the ReadOnly device? because that's what we have in the local cache/go memory space that already has the header?

yes

> But with compression we will not store the headers until the files are uploaded, maybe a race there coming?

we'll provide "early" headers without the compression mapping -> update them to the ones with compression

packages/orchestrator/internal/sandbox/template/peerprovider/helpers_test.go
dobrac: moved the writeTempFile, keeping the `collectSender` as it's reused across most of the test files

packages/shared/pkg/storage/storage_cache.go
dobrac: maybe it's ok. In my testing though, the pressure on the files was causing the sandbox resume to take too long. We can circle back on this and enable it later

packages/orchestrator/internal/server/sandboxes.go
dobrac: do you mean an explanation for how prefetching works or just this place?

packages/orchestrator/chunks.proto
dobrac: renamed to `ReadAtBuildSeekableRequest`

packages/orchestrator/internal/sandbox/template/peerclient/resolver.go
dobrac: when the sandbox is resuming on the same orchestrator node. But practically it should not happen as it should be already cached

packages/orchestrator/internal/server/chunks.go
dobrac: Claude. I don't think it's actually used, but I wanted to have it implemented in case someone uses it in the future

packages/orchestrator/internal/sandbox/template/cache.go
dobrac: updated

packages/orchestrator/internal/sandbox/template/peerserver/peerserver.go
dobrac: added comment

packages/orchestrator/internal/server/sandboxes.go
levb: just the metadata embedding, I did not understand it

packages/orchestrator/internal/sandbox/build/cache.go
dobrac: I can do a follow up PR

packages/orchestrator/internal/sandbox/template/peerprovider/file.go
dobrac: renamed

packages/orchestrator/internal/server/sandboxes.go
dobrac: when the paused metadata are created, they lack prefetch info. The prefetch info is optional (optimization), but in this case, I can get it before any upload happens. -> In that case, I can just update the metadata struct with the prefetch data right before the upload, so they're available

packages/orchestrator/internal/server/main.go
```diff
@@ -39,6 +41,8 @@ type Server struct {
 	featureFlags      *featureflags.Client
 	sbxEventsService  *events.EventsService
 	startingSandboxes *semaphore.Weighted
+	peerRegistry      peerclient.Registry
+	uploadedBuilds    sync.Map // buildID → struct{}: builds whose GCS upload is done
```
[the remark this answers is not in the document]
dobrac: addressed

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -596,33 +628,64 @@ func (s *Server) snapshotAndCacheSandbox(
...
+
+	// Register in Redis so other orchestrators can find us for peer routing.
+	if s.featureFlags.BoolFlag(ctx, featureflags.PeerToPeerChunkTransferFlag) {
+		if err := s.peerRegistry.Register(ctx, meta.Template.BuildID, redisPeerKeyTTL); err != nil {
+			logger.L().Warn(ctx, "failed to register peer address for routing", zap.String("build_id", meta.Template.BuildID), zap.Error(err))
+		}
+
+		completeUpload := func(ctx context.Context) {
+			// Mark in-memory so ChunkService tells peers to switch to GCS.
+			s.uploadedBuilds.Store(meta.Template.BuildID, struct{}{})
```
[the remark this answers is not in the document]
dobrac: addressed

PR #2037, opened 2026-03-02
Feature: Sandbox logs level & text search filters

ben-fornefeld:
> [!NOTE]
> **Medium Risk**
> Extends public log-fetching APIs and Loki query construction; mis-handling of filtering or query sanitization could impact log visibility and introduce performance regressions on large log streams. Remote path also depends on an Edge feature header, so mismatched deployments may yield unexpected unfiltered results while only logging a warning.
> 
> **Overview**
> Adds optional `level` and `search` query parameters to sandbox logs retrieval (API v2 and Edge), plumbing them through handlers and `ClusterResource` implementations for both local (Loki) and remote (Edge) providers. Loki querying now builds a LogQL pipeline that JSON-parses and regex-matches `message` with escaping to reduce injection risk, and the remote provider emits a warning when filters are requested but the Edge response lacks the feature-flag header; OpenAPI specs and generated clients/mocks are updated accordingly, with a small unit test covering query escaping.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0bfaf82b62f1010f8c6b6f900759623f9672220b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/shared/pkg/logs/loki/provider_test.go
+++ b/packages/shared/pkg/logs/loki/provider_test.go
@@ -22,7 +22,6 @@ func TestSanitizeLokiLabel(t *testing.T) {
 	}
 
 	for _, tt := range tests {
-		tt := tt
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
 			assert.Equal(t, tt.expected, sanitizeLokiLabel(tt.input))
@@ -45,7 +44,6 @@ func TestSanitizeLogMessageRegexFilter(t *testing.T) {
 	}
 
 	for _, tt := range tests {
-		tt := tt
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
 			assert.Equal(t, tt.expected, sanitizeLogMessageRegexFilter(tt.input))


packages/api/internal/clusters/resources_remote.go
```diff
@@ -116,22 +128,34 @@ func (r *ClusterResourceProviderImpl) GetSandboxLogs(ctx context.Context, teamID
...
-		}
+	filtersRequested := level != nil || (search != nil && *search != "")
+	if filtersRequested && (res.HTTPResponse == nil || res.HTTPResponse.Header.Get(consts.EdgeFeatureSandboxLogsLevelTextFilteringEnabledHeader) == "") {
+		logger.L().Warn(
+			ctx,
+			"edge response missing logs level+text filtering enabled header",
+			logger.WithSandboxID(sandboxID),
+			zap.Bool("level_filter", level != nil),
+			zap.Bool("search_filter", search != nil && *search != ""),
+		)
```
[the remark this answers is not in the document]
ben-fornefeld: @dobrac we have to decide how we want to handle this. i think logging for now is fine, but we probably want better systems to either fallback or fail transparently

dobrac: can you move this to a helper function

packages/shared/pkg/logs/loki/provider.go
```diff
@@ -105,10 +102,52 @@ func (l *LokiQueryProvider) QuerySandboxLogs(
 	return lm, nil
 }
 
-// sanitizeLokiLabel sanitizes input strings for use in Loki LogQL queries
```
dobrac, line 108: why removing the comments here?

packages/shared/pkg/logs/loki/provider.go
```diff
@@ -105,10 +102,52 @@ func (l *LokiQueryProvider) QuerySandboxLogs(
...
+	query := fmt.Sprintf("{service=\"template-manager\", buildID=`%s`, envID=`%s`}", buildID, templateID)
+	if level == nil {
+		return query
+	}
+
+	return query + fmt.Sprintf(" | json | level =~ `%s`", minLevelRegexFilter(*level))
+}
+
+func buildSandboxLogsQuery(teamID string, sandboxID string, level *logs.LogLevel, search *string) string {
+	sandboxIDSanitized := sanitizeLokiLabel(sandboxID)
```
dobrac, line 147: the docs link is missing:

https://grafana.com/blog/2021/01/05/how-to-escape-special-characters-with-lokis-logql/

packages/shared/pkg/logs/loki/provider.go
```diff
@@ -105,10 +102,52 @@ func (l *LokiQueryProvider) QuerySandboxLogs(
...
+	teamIDSanitized := sanitizeLokiLabel(teamID)
+	query := fmt.Sprintf("{teamID=`%s`, sandboxID=`%s`, category!=\"metrics\"}", teamIDSanitized, sandboxIDSanitized)
+	if level == nil && (search == nil || *search == "") {
```
dobrac: instead of this you can use the utils.DerefOrDefault

packages/shared/pkg/logs/loki/provider.go
```diff
@@ -105,10 +102,52 @@ func (l *LokiQueryProvider) QuerySandboxLogs(
...
+}
+
+func buildBuildLogsQuery(templateID string, buildID string, level *logs.LogLevel) string {
+	// todo: service name is different here (because new merged orchestrator)
+	query := fmt.Sprintf("{service=\"template-manager\", buildID=`%s`, envID=`%s`}", buildID, templateID)
+	if level == nil {
+		return query
+	}
+
+	return query + fmt.Sprintf(" | json | level =~ `%s`", minLevelRegexFilter(*level))
```
dobrac, line 142: is the level 100% correctly escaped?

packages/shared/pkg/logs/loki/provider.go
```diff
@@ -105,10 +102,52 @@ func (l *LokiQueryProvider) QuerySandboxLogs(
...
 
-// sanitizeLokiLabel sanitizes input strings for use in Loki LogQL queries
-// by removing special characters that could enable query injection
 func sanitizeLokiLabel(input string) string {
-	// Remove characters that have special meaning in LogQL
-	// Backticks used to delimit label values
 	return strings.ReplaceAll(input, "`", "")
 }
+
+func sanitizeLogMessageRegexFilter(input string) string {
```
dobrac, line 114: can you add tests for this? Also, can you reason that this is enough?

packages/shared/pkg/logs/loki/provider.go
```diff
@@ -105,10 +102,52 @@ func (l *LokiQueryProvider) QuerySandboxLogs(
...
+	return fmt.Sprintf(".*%s.*", regexp.QuoteMeta(strings.ReplaceAll(input, "`", "")))
```
dobrac: could you also reuse here the: `sanitizeLokiLabel`

packages/shared/pkg/logs/loki/provider.go
ben-fornefeld: linked resources, checked again and added few tests

packages/api/internal/clusters/resources_remote.go
```diff
@@ -257,3 +248,20 @@ func handleEdgeErrorResponse(statusCode int, json400, json401, json500 *edgeapi.
...
+	filtersApplied := response != nil && response.Header.Get(consts.EdgeFeatureSandboxLogsLevelTextFilteringEnabledHeader) != ""
+	if !filtersRequested || filtersApplied {
+		return
+	}
+
+	edgeapi.WarnMissingFeatureHeader(
+		ctx,
+		consts.EdgeFeatureSandboxLogsLevelTextFilteringEnabledHeader,
+		"sandbox logs level+text filtering not supported",
+		logger.WithSandboxID(sandboxID),
```
dobrac, line 267: lets add also the cluster ID so we know which cluster doesn't support it

packages/shared/pkg/http/edge/incompatibility_log.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"context"
+
+	"go.uber.org/zap"
+
+	"github.com/e2b-dev/infra/packages/shared/pkg/logger"
+)
+
+func WarnMissingFeatureHeader(ctx context.Context, featureHeader string, msg string, fields ...zap.Field) {
```
dobrac, line 11: do we need this split to shared?

dobrac (changes requested):
also the tests please for the inputs validators

packages/shared/pkg/http/edge/incompatibility_log.go
ben-fornefeld: not really but i would like to keep it under "edge", since it's indirectly related to the headers in consts.go

dobrac: ok

PR #2039, opened 2026-03-02
Add API debug

ValentaTomas:
> [!NOTE]
> **Medium Risk**
> Introduces a new `pprof` HTTP server and enables runtime metric callbacks, which can increase CPU/memory overhead and expose profiling data if networking or port-forwarding accidentally makes `127.0.0.1:6060` reachable. Broad OTEL version bumps may also subtly change emitted metric/trace schemas and volumes.
> 
> **Overview**
> Adds a dedicated, localhost-bound `pprof` server (with a custom mux to avoid `http.DefaultServeMux`) and starts OTEL Go runtime metric instrumentation so services emit standard `go.*` runtime metrics; the local profiling workflow is updated to target port `6060`, the OTEL collector filter now allows `go.*` metrics through, and multiple modules bump OpenTelemetry (and `x/sys`) dependencies to the newer versions needed for runtime instrumentation.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 463c005c4075941b60f2155f2f70768f05386103. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/api/main.go
```diff
@@ -434,6 +448,19 @@ func run() int {
 		}
 	})
 
+	pprofServer := &http.Server{
```
levb, line 443: 👍

packages/api/main.go
```diff
@@ -285,6 +287,18 @@ func run() int {
 	defer sbxLoggerInternal.Sync()
 	sbxlogger.SetSandboxLoggerInternal(sbxLoggerInternal)
 
+	// Start Go runtime metrics collection (goroutines, heap, GC, etc.)
+	stopRuntime, err := tel.StartRuntimeInstrumentation()
+	if err != nil {
+		log.Printf("warning: failed to start runtime instrumentation: %v", err)
```
dobrac: lets use the logger instead

levb (approved):
LGTM, once the Claude comments are addressed.

👍 to pprof

packages/api/main.go
```diff
@@ -461,6 +488,10 @@ func run() int {
 			l.Error(ctx, "Http service shutdown error", zap.Int("port", port), zap.Error(err))
 		}
 
+		if err := pprofServer.Shutdown(ctx); err != nil {
```
dobrac: lets use cleanupFns instead?

ValentaTomas: We created it in the same wg as other servers, so it might make sense to handle the shutdown with the same wg. Open to alternatives.

PR #2067, opened 2026-03-06
Support raw body file uploads in envd

mishushakov:
## Summary

Adds support for uploading files via raw request body (application/octet-stream) in the envd file upload endpoint, in addition to the existing multipart/form-data support. This enables simpler file upload clients that don't need multipart support.

- When Content-Type is `application/octet-stream`, the entire request body is treated as file content
- The `path` query parameter is required for raw body uploads
- Multipart uploads remain the default for backwards compatibility

## Implementation

- Extracted `handleMultipartUpload()` for existing multipart form parsing
- Added `handleRawUpload()` for application/octet-stream content type  
- Updated OpenAPI spec to document both content types

---

> [!NOTE]
> **Medium Risk**
> Changes request parsing for the file upload endpoint by branching on `Content-Type`, which could affect existing clients or error handling paths. File write behavior is still the same, but new code paths (including gzip + raw body) add some integration risk.
> 
> **Overview**
> Adds support for uploading a single file via a raw `application/octet-stream` request body on the `/files` endpoint (requiring the `path` query parameter), while keeping multipart uploads as the default for backwards compatibility. The handler now rejects unsupported content types explicitly, documents the new request body option in the OpenAPI spec, bumps the envd version, and adds coverage for raw uploads (including gzip, overwrite, and directory creation cases).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e5d7d33403441be70cfac7b0a8a9e062a72ed87f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/envd/main.go
```diff
@@ -47,7 +47,7 @@ const (
 )
 
 var (
-	Version = "0.5.4"
+	Version = "0.5.5"
```
jakubno:
```suggestion
	Version = "0.5.7"
```

PR #2072, opened 2026-03-07
Increase timeouts for the LB backend

ValentaTomas:
Previously they were lower than the api write timeouts, causing 504 where it should have been errors from the API.


iac/provider-gcp/nomad-cluster/network/ingress.tf
```diff
@@ -43,7 +43,7 @@ resource "google_compute_backend_service" "ingress" {
   session_affinity = null
   health_checks    = [google_compute_health_check.ingress.id]
 
-  timeout_sec = 65
+  timeout_sec = 80
```
dobrac, line 46: this should be changed also for the API, not only ingress

dobrac: my bad, missed the second file

PR #2074, opened 2026-03-08
feat: dynamic egress network rule updates for running sandboxes

levb:
## Summary

  - Add `PUT /sandboxes/{sandboxID}/network` API endpoint to dynamically update egress firewall rules (allow/deny CIDRs and domains) on running sandboxes without restart
  - Full end-to-end flow: API handler → sandbox store update → gRPC `UpdateNetwork` RPC → nftables rule replacement in sandbox network namespace
  - Domain-based filtering supported in `allowOut` via TLS SNI matching in the TCP proxy
  - Rules persist across pause/resume cycles
  - Thread-safe in-memory egress config using a single `RWMutex` protecting `startedAt`, `endAt`, and `networkEgress`
  - `GetNetwork()` guaranteed non-nil, eliminating nil checks in proxy and TCP firewall code paths
  - CIDRs validated before firewall reset to prevent partial rule application

## Test plan

  - [x] Integration test: dynamically allow a specific IP while denying all other traffic
  - [x] Integration test: clear all rules and verify traffic flows freely
  - [x] Integration test: replace rules (PUT semantics, not additive)
  - [x] Integration test: 404 for nonexistent sandbox
  - [x] Integration test: 401 without API key
  - [x] Integration test: rules survive pause/resume cycle
  - [x] Integration test: multiple sequential updates each change actual connectivity
  - [x] Integration test: domain allow/remove via TLS SNI


packages/orchestrator/internal/sandbox/network/slot.go
```diff
@@ -297,6 +297,62 @@ func (s *Slot) ConfigureInternet(ctx context.Context, network *orchestrator.Sand
...
+	}
+
+	n, err := ns.GetNS(filepath.Join(netNamespacesDir, s.NamespaceID()))
+	if err != nil {
+		return fmt.Errorf("failed to get slot network namespace '%s': %w", s.NamespaceID(), err)
+	}
+	defer n.Close()
+
+	err = n.Do(func(_ ns.NetNS) error {
+		if err := s.Firewall.Reset(); err != nil {
```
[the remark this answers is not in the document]
levb: [942d1e1](https://github.com/e2b-dev/infra/pull/2074/commits/942d1e158eae37e59537245ec05f84cfbc98bcab) - a little bigger refactoring, but seemingly for the best.

packages/orchestrator/orchestrator.proto
```diff
@@ -147,4 +155,6 @@ service SandboxService {
   rpc Checkpoint(SandboxCheckpointRequest) returns (SandboxCheckpointResponse);
 
   rpc ListCachedBuilds(google.protobuf.Empty) returns (SandboxListCachedBuildsResponse);
+
+  rpc UpdateNetwork(SandboxUpdateNetworkRequest) returns (google.protobuf.Empty);
```
dobrac: should we use the SandboxUpdateRequest instead?

packages/orchestrator/orchestrator.proto
```diff
@@ -99,6 +99,14 @@ message SandboxUpdateRequest {
   google.protobuf.Timestamp end_time = 2;
 }
 
+message SandboxUpdateNetworkRequest {
+  string sandbox_id = 1;
+
+  repeated string allowed_cidrs = 2;
```
dobrac: reusing this message sounds to me like a better idea: SandboxNetworkEgressConfig

packages/orchestrator/internal/tcpfirewall/handlers_test.go
```diff
@@ -389,6 +389,8 @@ func TestIsEgressAllowed(t *testing.T) {
 					},
 				},
 			}
+			// Mirror the constructor: initialize networkEgress from config.
```
dobrac: why this change?

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -709,8 +749,9 @@ func (f *Factory) ResumeSandbox(
 		Config:  config,
 		Runtime: runtime,
 
-		startedAt: startedAt,
-		endAt:     endAt,
+		startedAt:     startedAt,
+		endAt:         endAt,
+		networkEgress: config.Network.GetEgress(),
```
dobrac: this is already saved on sandbox in config

tests/integration/internal/tests/api/sandboxes/sandbox_network_update_test.go
```diff
@@ -0,0 +1,338 @@
...
+				DenyOut:  s.denyOut,
+			})
+			require.Equal(t, http.StatusNoContent, resp.StatusCode())
+			verifyConnectivity(t, ctx, sbx, envdClient, s.checks)
+		})
+	}
+
+	// ── Pause/resume (must be last — changes sandbox lifecycle) ───────────
+
+	t.Run("pause_resume_preserves_rules", func(t *testing.T) { //nolint:paralleltest // sequential
```
dobrac, line 307: why the no paralleltest, sequential? Can we solve it differnetly then?

packages/api/internal/handlers/sandbox_network_update.go
```diff
@@ -0,0 +1,81 @@
...
+		allowedEntries = *body.AllowOut
+	}
+
+	var deniedEntries []string
+	if body.DenyOut != nil {
+		deniedEntries = *body.DenyOut
+	}
+
+	// Validate denyOut entries are valid IPs/CIDRs (not domains).
+	for _, entry := range deniedEntries {
```
dobrac: the validation function should be reused from the sandbox create

packages/api/internal/handlers/sandbox_network_update.go
```diff
@@ -0,0 +1,81 @@
...
+
+		if len(allowedDomains) > 0 && !hasBlockAll {
+			a.sendAPIStoreError(c, http.StatusBadRequest, ErrMsgDomainsRequireBlockAll)
+
+			return
+		}
+	}
+
+	if apiErr := a.orchestrator.UpdateSandboxNetworkConfig(ctx, team.ID, sandboxID, allowedEntries, deniedEntries); apiErr != nil {
+		telemetry.ReportError(ctx, "error updating sandbox network config", apiErr.Err)
```
dobrac: this ReportError/ReportCriticalError should be dependent on the apiErr.Code. We've already few places like this

packages/api/internal/orchestrator/update_network.go
```diff
@@ -0,0 +1,138 @@
...
+	allowedEntries []string,
+	deniedEntries []string,
+) *api.APIError {
+	// Split allowed entries into CIDRs/IPs and domains, matching the creation
+	// path in buildNetworkConfig.
+	allowedAddresses, allowedDomains := sandbox_network.ParseAddressesAndDomains(allowedEntries)
+
+	// If allowed domains are provided, add the default nameserver so the sandbox
+	// can resolve domain names — same as the creation path.
+	if len(allowedDomains) > 0 {
```
dobrac: should be again united with the sandbox creation

packages/api/internal/orchestrator/update_network.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/e2b-dev/infra/packages/api/internal/api"
+	"github.com/e2b-dev/infra/packages/api/internal/sandbox"
+	"github.com/e2b-dev/infra/packages/api/internal/utils"
+	"github.com/e2b-dev/infra/packages/db/pkg/types"
+	orchestratorgrpc "github.com/e2b-dev/infra/packages/shared/pkg/grpc/orchestrator"
+	sandbox_network "github.com/e2b-dev/infra/packages/shared/pkg/sandbox-network"
+	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
+)
+
+func (o *Orchestrator) UpdateSandboxNetworkConfig(
```
dobrac, line 23: this should have the same paradigm as the `KeepAliveFor`

packages/api/internal/orchestrator/update_network.go
```diff
@@ -0,0 +1,138 @@
...
+	if err != nil {
+		var sbxNotFoundErr *sandbox.NotFoundError
+		if errors.As(err, &sbxNotFoundErr) {
+			return &api.APIError{Code: http.StatusNotFound, ClientMsg: utils.SandboxNotFoundMsg(sandboxID), Err: err}
+		}
+
+		return &api.APIError{Code: http.StatusInternalServerError, ClientMsg: "Error reading sandbox", Err: err}
+	}
+
+	if sbx.State != sandbox.StateRunning {
```
dobrac: this needs to be checked in the updateFunc

packages/orchestrator/internal/proxy/proxy.go
```diff
@@ -72,10 +72,8 @@ func NewSandboxProxy(meterProvider metric.MeterProvider, port uint16, sandboxes
 				return nil, reverseproxy.NewErrSandboxNotFound(sandboxId)
 			}
 
-			var accessToken *string = nil
-			if net := sbx.Config.Network; net != nil && net.GetIngress() != nil {
-				accessToken = net.GetIngress().TrafficAccessToken
-			}
+			ingress := sbx.GetNetwork().GetIngress()
+			accessToken := ingress.TrafficAccessToken
```
dobrac: the ingress can be nil here

packages/orchestrator/internal/sandbox/network/slot.go
```diff
@@ -274,26 +274,38 @@ func (s *Slot) ConfigureInternet(ctx context.Context, network *orchestrator.Sand
...
+	if err != nil {
+		return fmt.Errorf("failed execution in network namespace '%s': %w", s.NamespaceID(), err)
+	}
 
-		return nil
+	return nil
+}
+
+// UpdateInternet replaces all user firewall rules atomically in a single nftables flush.
+func (s *Slot) UpdateInternet(ctx context.Context, allowedCIDRs, deniedCIDRs []string) error {
```
dobrac: why not passing the `network *orchestrator.SandboxNetworkConfig`?

packages/orchestrator/internal/sandbox/network/slot.go
```diff
@@ -274,26 +274,38 @@ func (s *Slot) ConfigureInternet(ctx context.Context, network *orchestrator.Sand
...
+	defer n.Close()
+
+	err = n.Do(func(_ ns.NetNS) error {
+		return s.Firewall.ReplaceUserRules(allowedCIDRs, deniedCIDRs)
 	})
 	if err != nil {
 		return fmt.Errorf("failed execution in network namespace '%s': %w", s.NamespaceID(), err)
 	}
 
+	hasRules := len(allowedCIDRs) > 0 || len(deniedCIDRs) > 0
```
dobrac: this boolean decision is quite finicky. I would not reset the `firewallCustomRules` back to false. Or, I would return the status from `ReplaceUserRules`

tests/integration/internal/tests/api/sandboxes/sandbox_network_update_test.go
levb: this was to re-use 1 sandbox; parallel tests would be starting a sandbox each, not needed really; 0/5 this is marginally faster

packages/orchestrator/orchestrator.proto
levb: Are you asking to consolidate the RPCs, the data structs, or something else?

dobrac: I think we could merge this to:

`rpc Update(SandboxUpdateRequest) returns (google.protobuf.Empty);`

It's effectively an update

levb: ACK. Not sure it's a good idea since the use cases and error handling for the different updates might be different; how do you propose we'd handle errors if both expiration and network are to be updated, and one fails? @dobrac

packages/orchestrator/internal/sandbox/network/slot.go
levb: Considering that the only use-case for the flag is deduplicate `ResetInternet`, I'll set it to true unconditionally.

packages/api/internal/handlers/sandbox_network_update.go
levb: done

packages/orchestrator/internal/proxy/proxy.go
levb: No, `GetNetwork()` returns a nil-free config

packages/orchestrator/orchestrator.proto
dobrac: can we fail both and revert to the original state?

packages/orchestrator/internal/proxy/proxy.go
dobrac: This is the `GetIngress()` implementation. How do you know it's nil-free?

```
func (x *SandboxNetworkConfig) GetIngress() *SandboxNetworkIngressConfig {
	if x != nil {
		return x.Ingress
	}
	return nil
}
```

levb: GetNetwork() returns non-nil "everything" specifically so that we don't need to check for `nil` everywhere; there is no semantic difference in network config AFAICT between `nil` and empty values.

```
...
	ingress := net.GetIngress()
	if ingress == nil {
		ingress = &orchestrator.SandboxNetworkIngressConfig{}
	}

	return &orchestrator.SandboxNetworkConfig{
		Egress:  egress,
		Ingress: ingress,
	}
```

packages/orchestrator/orchestrator.proto
levb: Updated to a single gRPC handler.

packages/orchestrator/internal/sandbox/sandbox.go
levb: Right, but `networkEgress ` is a mutable value, protected by `rwmu` and this line is the nil-safe initialization of the value (from the config, which is not mutable)

packages/orchestrator/internal/tcpfirewall/handlers_test.go
levb: Was a claude "accident" - was added purely for consistency with other sandbox construction, removed.

packages/orchestrator/internal/sandbox/sandbox.go
levb: As explained in [9524334](https://github.com/e2b-dev/infra/pull/2074/commits/95243344ab6af11930021e3a9bbc8a35a68440d7), this was necessary.

packages/api/internal/handlers/sandbox_network_update.go
```diff
@@ -0,0 +1,63 @@
...
+	if body.AllowOut != nil {
+		allowedEntries = *body.AllowOut
+	}
+
+	var deniedEntries []string
+	if body.DenyOut != nil {
+		deniedEntries = *body.DenyOut
+	}
+
+	if apiErr := validateEgressRules(allowedEntries, deniedEntries); apiErr != nil {
```
dobrac, line 49: the domains should be validated too, not only the IPs (you should be able to update everything in the network config)

packages/api/internal/orchestrator/update_network.go
```diff
@@ -0,0 +1,128 @@
...
+func (o *Orchestrator) UpdateSandboxNetworkConfig(
+	ctx context.Context,
+	teamID uuid.UUID,
+	sandboxID string,
+	allowedEntries []string,
+	deniedEntries []string,
+) *api.APIError {
+	// Split allowed entries into CIDRs/IPs and domains, matching the creation
+	// path in buildNetworkConfig.
+	allowedAddresses, allowedDomains := sandbox_network.ParseAddressesAndDomains(allowedEntries)
```
dobrac: this logic is done also in one more place, right? Can we unite them?

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -267,29 +267,91 @@ func (s *Server) Update(ctx context.Context, req *orchestrator.SandboxUpdateRequ
...
-	)
+			return rollback, nil
+		})
+	}
+
+	if req.GetEgress() != nil {
+		updates = append(updates, func(ctx context.Context, sbx *sandbox.Sandbox) (func(), error) {
+			oldEgress := sbx.GetNetworkEgress()
+			rollback := func() {
+				_ = sbx.Slot.UpdateInternet(ctx, oldEgress)
```
dobrac: note, the context might be canceled at this point. I would pass context to the rollback and make it without cancel

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -267,29 +267,91 @@ func (s *Server) Update(ctx context.Context, req *orchestrator.SandboxUpdateRequ
 		return nil, status.Error(codes.NotFound, "sandbox not found")
 	}
 
-	sbx.SetEndAt(req.GetEndTime().AsTime())
+	var updates []func(ctx context.Context, sbx *sandbox.Sandbox) (rollback func(), err error)
```
dobrac: can we do a small file/struct wrapper for the rollback functionality please?

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -140,29 +140,68 @@ type Metadata struct {
...
-	m.endAtMu.Lock()
-	defer m.endAtMu.Unlock()
+	m.rwmu.Lock()
+	defer m.rwmu.Unlock()
 
 	m.endAt = t
 }
 
+// GetNetworkEgress returns the sandbox network egress config in a thread-safe manner.
+func (m *Metadata) GetNetworkEgress() *orchestrator.SandboxNetworkEgressConfig {
```
dobrac: lets move the getter and setter on the Config itself, including new lock for that field

packages/api/internal/handlers/sandbox_network_update.go
levb: @dobrac I added, but I question whether it's of value - we just validate some ASCII formatting... Unless you meant to Lookup at this time - not sure if it's worth it.

packages/api/internal/orchestrator/update_network.go
levb: done

packages/api/internal/handlers/sandbox_network_update.go
dobrac: its a user input validation, I think it should be the same as for the sandbox creation

packages/orchestrator/internal/server/sandboxes.go
levb: See `ApplyAllOrRollback`

packages/api/internal/handlers/sandbox_create.go
```diff
@@ -523,6 +524,26 @@ func validateEgressRules(allowOut, denyOut []string) *api.APIError {
...
+			if validateDomain != "*" {
+				if _, err := idna.Lookup.ToASCII(validateDomain); err != nil {
+					return &api.APIError{
+						Code:      http.StatusBadRequest,
+						Err:       fmt.Errorf("invalid allowed domain %q: %w", domain, err),
+						ClientMsg: fmt.Sprintf("invalid allowed domain: %s", domain),
+					}
+				}
+			}
+		}
```
dobrac: I'm sorry I got confused. I don't think we need this at all

packages/api/internal/orchestrator/create_instance.go
```diff
@@ -64,7 +69,6 @@ func buildNetworkConfig(network *types.SandboxNetworkConfig, allowInternetAccess
 	// Handle the case where internet access is explicitly disabled
 	// This should be applied after copying the network config to preserve allowed addresses
 	if allowInternetAccess != nil && !*allowInternetAccess {
-		// Block all internet access - this overrides any other blocked addresses
```
dobrac, line 66: why removing the comment?

packages/api/internal/orchestrator/create_instance.go
```diff
@@ -64,7 +69,6 @@ func buildNetworkConfig(network *types.SandboxNetworkConfig, allowInternetAccess
 	// Handle the case where internet access is explicitly disabled
 	// This should be applied after copying the network config to preserve allowed addresses
 	if allowInternetAccess != nil && !*allowInternetAccess {
```
dobrac, line 70: I think we should support the allowInternetAccess for the network update too as a separate option. Then, this should be also part of the common helper util

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -71,13 +71,64 @@ type Config struct {
...
+
+	ingress := c.network.GetIngress()
+	if ingress == nil {
+		ingress = &orchestrator.SandboxNetworkIngressConfig{}
+	}
+
+	return &orchestrator.SandboxNetworkConfig{
+		Egress:  egress,
+		Ingress: ingress,
+	}
```
dobrac: is there a better way to do this? looks error prone

packages/orchestrator/internal/sandbox/network/firewall.go
```diff
@@ -240,120 +241,72 @@ func (fw *Firewall) installRules() error {
...
-func addCIDRToSet(conn *nftables.Conn, ipset set.Set, cidr string) error {
-	current, err := ipset.Elements(conn)
-	if err != nil {
-		return err
-	}
+// clearAndReplaceCIDRs flushes a set and repopulates it with the given CIDRs.
+// Handles the special 0.0.0.0/0 case which the firewall_toolkit validation
+// rejects (0.0.0.0 is "unspecified") by directly creating nftables elements.
+func clearAndReplaceCIDRs(conn *nftables.Conn, s set.Set, cidrs []string) error {
+	conn.FlushSet(s.Set())
```
dobrac: doesn't this also flush the changes?

packages/orchestrator/internal/tcpfirewall/handlers_test.go
```diff
@@ -382,11 +382,11 @@ func TestIsEgressAllowed(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
+			cfg := &sandbox.Config{}
+			cfg.SetNetwork(tt.network)
```
dobrac: should we do a constructor instead?

packages/shared/pkg/utils/apply_all_or_none.go
```diff
@@ -0,0 +1,33 @@
+package utils
+
+import "context"
+
+// UpdateFunc performs an update and returns a rollback function to revert it.
+// The rollback receives a context that is guaranteed not to be canceled.
+type UpdateFunc = func(ctx context.Context) (rollback func(ctx context.Context), err error)
```
dobrac, line 7: I was thinking about having stateful state keeping struct. But this is fine

packages/orchestrator/internal/sandbox/network/firewall.go
levb: `FlushSets` just reset `cc.messages` to a base in conn, does not flush, we do that atomically for both sets (deny. allow) in line #272 with `conn.Flush()`; and we apply the changes to each set individually, so there's no interplay between the calls to `clearAndReplaceCIDRs`, but I'll clarify the code a little.

levb: Also, what was the concern in terms of the functional (mis-)behavior? 1/5 I can add maybe a test for the concern.

packages/shared/pkg/utils/apply_all_or_none.go
levb: I thought this was simpler, fewer notes. All the state can be in the closures.

packages/api/internal/orchestrator/create_instance.go
levb: accident, restored.

packages/orchestrator/internal/sandbox/sandbox.go
levb: Thanks for flagging this, simplified - the constructor ensures a non-nil network and mutex.

packages/api/internal/orchestrator/create_instance.go
levb: Changing the overall egress/ingress access mode ON/OFF is a bit more involved, 1/5 maybe a separate PR for both after we have the dynamic settings worked out?

dobrac: lets do in another PR on top of this one

packages/orchestrator/internal/sandbox/network/firewall.go
dobrac: Okay, thx. The concern was behavior one where we could end up with incorrect firewall settings for a small amount of time

PR #2077, opened 2026-03-09
Increase start-client DNS wait timeouts from 10s to 60s

sitole:
Consul DNS and systemd-resolved can take longer than 10 seconds to become ready, causing the user-data script to exit before Nomad starts.

---

> [!NOTE]
> **Low Risk**
> Small change to boot-time retry timing only; primary risk is slightly slower node startup when DNS is genuinely broken, but no security-sensitive logic is modified.
> 
> **Overview**
> In both AWS and GCP client `start-client.sh` boot scripts, increases the readiness wait loops for Consul DNS (port 8600) and subsequent `systemd-resolved` DNS resolution checks from 10 to 60 attempts/seconds, updating log messages and failure thresholds so the scripts don’t exit prematurely before starting Nomad.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8f0642ac36380729a5f4c7ac87193f342c3a7813. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/iac/provider-aws/modules/nodepool-client/scripts/start-client.sh
+++ b/iac/provider-aws/modules/nodepool-client/scripts/start-client.sh
@@ -201,13 +201,13 @@ echo $overcommitment_hugepages >/proc/sys/vm/nr_overcommit_hugepages
 
 # Give Consul a moment to start its DNS server on port 8600
 echo "- Waiting for Consul DNS to start on port 8600..."
-for i in {1..10}; do
+for i in {1..60}; do
   if nc -z 127.0.0.1 8600 2>/dev/null; then
-    echo "- Consul DNS is ready (attempt $i/10)"
+    echo "- Consul DNS is ready (attempt $i/60)"
     break
   fi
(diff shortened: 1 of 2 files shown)


iac/provider-aws/modules/nodepool-client/scripts/start-client.sh
```diff
@@ -201,13 +201,13 @@ echo $overcommitment_hugepages >/proc/sys/vm/nr_overcommit_hugepages
 
 # Give Consul a moment to start its DNS server on port 8600
 echo "- Waiting for Consul DNS to start on port 8600..."
-for i in {1..10}; do
+for i in {1..30}; do
```
bchalios: Are 30 seconds enough? Does it make sense to increase it even more to avoid intermittent failures?

sitole: Increasing to 60s, makes sense

PR #2082, opened 2026-03-09
feat: reserve disk space for guest OS during template build

dobrac:
## Summary
Add configurable reserved disk space (MB) for the guest OS via feature flag `BuildReservedDiskSpaceMB`. Reserved blocks are set in two places to cover fresh and cached builds:

1. **Base builder**: Sets reserved blocks on host via `tune2fs -r` after final filesystem resize
2. **Finalize layer**: Sets reserved blocks inside sandbox via `tune2fs -r /dev/vda` before snapshot

This ensures the guest OS (running as root) always has available space, protecting systemd and other critical services from disk-full conditions even when non-root processes fill the disk.

Default value is 0 (no reserved space), preserving current behavior.

---

> [!NOTE]
> **Medium Risk**
> Touches template build/finalize filesystem provisioning and relies on running `tune2fs` against the root device, so misconfiguration or missing tooling could cause build failures or unexpected disk accounting; behavior is gated by a default-off feature flag.
> 
> **Overview**
> Adds a new feature flag (`BuildReservedDiskSpaceMB`) to reserve an amount of ext4 space for root during template builds, applying `tune2fs -r` both on the host-built rootfs after resizing and again inside the finalize sandbox before snapshotting; includes unit tests for parsing ext4 metadata output and an integration test that validates the reserved block count via `tune2fs` in a sandbox created from the built template.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7a8a4b3c86555c0fcd2c6123d0c718e4a7d3ff84. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/orchestrator/internal/template/build/core/filesystem/ext4.go
+++ b/packages/orchestrator/internal/template/build/core/filesystem/ext4.go
@@ -42,8 +42,13 @@ func Make(ctx context.Context, rootfsPath string, sizeMb int64, blockSize int64)
 	cmd := exec.CommandContext(ctx,
 		"mkfs.ext4",
 		// Matches the final ext4 features used by tar2ext4 tool
-		// But enables resize_inode, sparse_super (default, required for resize_inode), has_journal (default), metadata_csum (default)
-		"-O", `^dir_index,^64bit,^dir_nlink,ext_attr,sparse_super2,filetype,extent,flex_bg,large_file,huge_file,extra_isize`,
+		// But enables resize_inode, sparse_super (default, required for resize_inode), has_journal (default), metadata_csum (default).
+		// orphan_file is disabled (added as default in e2fsprogs >= 1.47.0) to ensure guest e2fsprogs tools
+		// (tune2fs, resize2fs, e2fsck) from older images (e.g. Ubuntu 22.04, Debian 11) can write to
+		// the filesystem. Without this, any write operation from the guest fails with "unsupported
+		// read-only feature(s)" when the host e2fsprogs is newer than the guest's.
+		// See https://e2fsprogs.sourceforge.net/e2fsprogs-release.html#1.47.0
+		"-O", `^dir_index,^64bit,^dir_nlink,^orphan_file,ext_attr,sparse_super2,filetype,extent,flex_bg,large_file,huge_file,extra_isize`,
 		"-b", strconv.FormatInt(blockSize, 10),
 		"-m", strconv.FormatInt(reservedBlocksPercentage, 10),
 		"-i", strconv.FormatInt(inodesRatio, 10),


packages/orchestrator/internal/template/build/core/filesystem/ext4.go
```diff
@@ -42,8 +42,13 @@ func Make(ctx context.Context, rootfsPath string, sizeMb int64, blockSize int64)
...
 		// Matches the final ext4 features used by tar2ext4 tool
-		// But enables resize_inode, sparse_super (default, required for resize_inode), has_journal (default), metadata_csum (default)
-		"-O", `^dir_index,^64bit,^dir_nlink,ext_attr,sparse_super2,filetype,extent,flex_bg,large_file,huge_file,extra_isize`,
+		// But enables resize_inode, sparse_super (default, required for resize_inode), has_journal (default), metadata_csum (default).
+		// orphan_file is disabled (added as default in e2fsprogs >= 1.47.0) to ensure guest e2fsprogs tools
+		// (tune2fs, resize2fs, e2fsck) from older images (e.g. Ubuntu 22.04, Debian 11) can write to
+		// the filesystem. Without this, any write operation from the guest fails with "unsupported
+		// read-only feature(s)" when the host e2fsprogs is newer than the guest's.
+		// See https://e2fsprogs.sourceforge.net/e2fsprogs-release.html#1.47.0
+		"-O", `^dir_index,^64bit,^dir_nlink,^orphan_file,ext_attr,sparse_super2,filetype,extent,flex_bg,large_file,huge_file,extra_isize`,
```
djeebus, line 51: Might be nice to split this up and join after, which would give us a place to comment on each flag individually:

```golang
"-O", strings.Join([]string{
    "^dir_index", // do not use hashed b-trees for lookups
	"^64bit", // no need, as our file systems will not be greater than ~2.2 TB (512 byte blocks * 2^32)
    ... // etc
}, ","),
```

packages/orchestrator/internal/template/build/core/filesystem/ext4_test.go
```diff
@@ -0,0 +1,90 @@
+package filesystem
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestParseFreeBlocks(t *testing.T) {
```
djeebus, line 10: I know it's not the point of this PR, but to get free blocks, instead of `debugfs` and string parsing, we could use `stat -fs "%f" /`, which just outputs a number.

packages/orchestrator/internal/template/build/phases/finalize/builder.go
```diff
@@ -219,6 +219,16 @@ func (ppb *PostProcessingBuilder) postProcessingFn(userLogger logger.Logger) lay
 				return
 			}
 
+			// Set reserved disk space for the guest OS before syncing
+			if reservedDiskSpaceMB := int64(ppb.featureFlags.IntFlag(ctx, featureflags.BuildReservedDiskSpaceMB)); reservedDiskSpaceMB > 0 {
```
dobrac, line 223: it would be better to move this to the place where the envd is updated -> that way old templates would have it already for the build, not only for the sandbox

it's just a different place where to put this code

packages/orchestrator/internal/template/build/core/filesystem/ext4_test.go
dobrac: does stat -fs work on not mounted filesystem in a file?

packages/orchestrator/internal/template/build/phases/finalize/builder.go
matthewlouisbrockman: so in here in `layer_executor`?
```
	// Update envd binary to the latest version
	if cmd.UpdateEnvd {
	  ...
	}
```

dobrac: (it can be done in a follow up)

matthewlouisbrockman: merging as is then making PR for this

packages/orchestrator/internal/template/build/core/filesystem/ext4_test.go
djeebus: It works on the underlying filesystem, regardless of where it's mounted:

```
$ stat -fc "%f" /tmp/claude   # sub folder of the mounted file system
3922238
$ stat -fc "%f" /tmp          # root of the mount, same file system as /tmp/claude
3922238
$ stat -fc "%f" /             # different mounted file system
44120469
```

packages/orchestrator/internal/template/build/core/filesystem/ext4.go
arkamar: addressed in https://github.com/e2b-dev/infra/pull/2183

packages/orchestrator/internal/template/build/core/filesystem/ext4_test.go
dobrac: Right, but the filesystem is not mounted anywhere (at all), would that work too?

dobrac: The debugfs allows to execute certain action on not mounted filesystem. If you check the usage, we're passing just a file containing the ext4 fs there

PR #2087, opened 2026-03-10
fix(orch): error finding sandbox for source addr

jakubno:
> [!NOTE]
> **Medium Risk**
> Touches sandbox lifecycle tracking and routing visibility (map insertion/removal and running-state filtering), which could affect proxy/firewall lookups and max-sandbox enforcement if the started/MarkRunning transitions are wrong. Changes are localized but in a critical control path for sandbox networking and cleanup.
> 
> **Overview**
> This PR changes sandbox registration so sandboxes are inserted into the shared sandbox map earlier (before Firecracker create/resume) and only become visible to most consumers once marked running, preventing “error finding sandbox for source addr” during startup/resume while keeping `List`/`Get`/count logic focused on fully-running sandboxes. It wires the sandbox map into `sandbox.Factory`, moves lifecycle-guarded removals into factory-managed cleanup, removes ad-hoc map insert/remove calls from build/resume utilities, and standardizes logging by extracting the remote IP and emitting it via a new `logger.WithSandboxIP` field.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 16b7189939913a2d730375c0c48e3cad816f2020. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/orchestrator/internal/sandbox/map.go
+++ b/packages/orchestrator/internal/sandbox/map.go
@@ -123,10 +123,6 @@ func (m *Map) Remove(ctx context.Context, sandboxID string) {
 }
 
 func (m *Map) RemoveByLifecycleID(ctx context.Context, sandboxID, lifecycleID string) {
-	logger.L().Info(ctx, "removing sandbox from map by lifecycle ID",
-		logger.WithSandboxID(sandboxID),
-	)
-
 	removed := m.sandboxes.RemoveCb(sandboxID, func(_ string, v *Sandbox, exists bool) bool {
 		if !exists {
 			return false
@@ -140,6 +136,10 @@ func (m *Map) RemoveByLifecycleID(ctx context.Context, sandboxID, lifecycleID st
 	})
 
 	if removed {
+		logger.L().Info(ctx, "removing sandbox from map by lifecycle ID",
+			logger.WithSandboxID(sandboxID),
+		)
+
 		go m.trigger(func(s MapSubscriber) {
 			s.OnRemove(sandboxID)
 		})


packages/orchestrator/internal/sandbox/map.go
```diff
@@ -54,10 +77,24 @@ func (m *Map) GetByHostPort(hostPort string) (*Sandbox, error) {
...
 			return sbx, nil
 		}
+
+		// Keep the reference, if we don't find live sandbox use the dead one
+		deadMatch = sbx
+	}
+
+	if deadMatch != nil {
+		return deadMatch, nil
 	}
```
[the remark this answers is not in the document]
jakubno: yea, that's actually the purpose of this PR you know

packages/orchestrator/internal/sandbox/map.go
```diff
@@ -4,10 +4,13 @@ import (
 	"fmt"
 	"net"
 	"sync"
+	"time"
 
 	"github.com/e2b-dev/infra/packages/shared/pkg/smap"
 )
 
+const deadSandboxGracePeriod = 30 * time.Second
```
dobrac: this is not really a correct fix. The correct fix should not remove the sandbox from the map before it is stopped

You can see that we're incorrectly using the sandboxes map as status during pause

jakubno: I was able to simulate the problem and the issue (or at least part of it) comes from start, not stop
Also added better logging to easier debug

dobrac: fair, its most probably both. The fix looks much much better now 🙏

packages/orchestrator/internal/sandbox/map.go
```diff
@@ -37,17 +39,39 @@ func (m *Map) trigger(fn func(MapSubscriber)) {
...
+	}
+
+	return count
 }
 
 func (m *Map) Get(sandboxID string) (*Sandbox, bool) {
-	return m.sandboxes.Get(sandboxID)
+	sbx, ok := m.sandboxes.Get(sandboxID)
+	if !ok || !sbx.IsRunning() {
+		return nil, false
```
dobrac, line 67: will this work correctly if you can't get the non running sandbox at all?

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -728,8 +723,6 @@ func (s *Server) setupSandboxLifecycle(ctx context.Context, sbx *sandbox.Sandbox
 			sbxlogger.I(sbx).Error(ctx, "failed to cleanup sandbox, will remove from cache", zap.Error(cleanupErr))
 		}
 
-		s.sandboxes.RemoveByLifecycleID(sbx.Runtime.SandboxID, sbx.LifecycleID)
-
 		closeErr := s.proxy.RemoveFromPool(sbx.LifecycleID)
```
dobrac, line 726: this could be also moved now to the sandboxFactory. Can be a follow up PR

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -745,14 +738,14 @@ func (s *Server) acquireSandboxForSnapshot(ctx context.Context, sandboxID string
...
-	sbx, ok := s.sandboxes.Get(sandboxID)
+	sbx, ok := s.sandboxFactory.Sandboxes.Get(sandboxID)
 	if !ok {
 		telemetry.ReportCriticalError(ctx, "sandbox not found", nil)
 
 		return nil, status.Error(codes.NotFound, "sandbox not found")
 	}
 
-	s.sandboxes.Remove(sandboxID)
+	s.sandboxFactory.Sandboxes.Remove(ctx, sandboxID)
```
dobrac, line 748: won't this break it the same way again? If the sandbox is not in the map, there will be errors during pause

jakubno: Not sure, but we'll have much better observability. I didn't want to overcomplicate it unnecessarily

packages/orchestrator/internal/sandbox/map.go
jakubno: you can get it in `GetByHostPort`, otherwise the functionality should be the same as it was before

packages/orchestrator/internal/server/sandboxes.go
dobrac: I'm pretty sure there is networking happening during pause. We can proceed as it is though if you want to fix it in follow up

dobrac (approved):

packages/orchestrator/internal/server/sandboxes.go
jakubno: I am not that sure about that, I think there shouldn't be any requests at the point where the FC is not running anymore

jakubno: okay, you were right :D

PR #2088, opened 2026-03-10
fix(orch): wait for fault pages to finish

jakubno:
> [!NOTE]
> **Medium Risk**
> Touches concurrency and shutdown behavior in page-fault handling; if `faultPage()` stalls on I/O or ioctl, `Close()` may now block longer and faults arriving during shutdown are dropped.
> 
> **Overview**
> Ensures `Userfaultfd.Close()` waits for any in-flight `faultPage()` operations to finish by marking the instance as `closed` and taking an exclusive `settleRequests` lock, and makes `faultPage()` short-circuit once closed to avoid running ioctls during teardown.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 005ca2acd4f45d8255e10d46a473367df28d8275. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/orchestrator/internal/sandbox/uffd/userfaultfd/userfaultfd.go
+++ b/packages/orchestrator/internal/sandbox/uffd/userfaultfd/userfaultfd.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"sync"
+	"sync/atomic"
 	"syscall"
 	"unsafe"
 
@@ -49,6 +50,7 @@ type Userfaultfd struct {
 	wg errgroup.Group
 
 	logger logger.Logger
+	closed atomic.Bool
 }
 
 // NewUserfaultfdFromFd creates a new userfaultfd instance with optional configuration.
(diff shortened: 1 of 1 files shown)


packages/orchestrator/internal/sandbox/uffd/userfaultfd/userfaultfd.go
```diff
@@ -79,6 +81,14 @@ func NewUserfaultfdFromFd(fd uintptr, src block.Slicer, m *memory.Mapping, logge
 }
 
 func (u *Userfaultfd) Close() error {
```
bchalios, line 83: At what point do we call `Userfaultfd.Close()`?

In any case, I'm scared that what you're trying to achieve here won't work. Firecracker sends the actual file descriptor. It doesn't call `dup` and send a copy (not even sure whether `dup` would work here). So, if your intention is to avoid issuing `ioctl`s on a closed file descriptor this won't work in case firecracker stopped and closed the file descriptor while doing so.

What would (partially) solve the issue is stopping the `uffd` process before killing the Firecracker process. If, though, the Firecracker process dies by itself, then I think nothing can work.

PR #2091, opened 2026-03-10
use chroot/umount/pivotroot instead of jailing

djeebus:
Tasks remaining:
- [x] Resolve race condition (sandbox can't init before adding to map, can't add to map until it's init'd)

Follow up tasks:
- [ ] Enable tracing/logging per connection via feature flag
- [ ] Use LifecycleID to track sandbox volume mounts rather than SandboxID

---

> [!NOTE]
> **Medium Risk**
> Introduces new mount-namespace + `pivot_root` chroot implementation and changes NFS mount behavior, which can break volume isolation/mount reliability or require elevated privileges in production. Also expands IaC/env configurability, increasing risk of misconfiguration at deploy time.
> 
> **Overview**
> Switches the orchestrator’s volume filesystem isolation to a new `internal/chrooted` implementation that uses a dedicated mount namespace plus `mount`/`pivot_root`/`umount`, and adds a small builder to derive per-team/per-volume paths from `PERSISTENT_VOLUME_MOUNTS`. The orchestrator config now includes NFS proxy logging/tracing/level env vars (with custom parsing for `nfs.LogLevel`), `envd` now waits for all NFS mounts to complete and uses more resilient mount options, and the Terraform/Nomad job wiring is extended to pass arbitrary extra env vars and to make GCP persistent volume types support custom NFS mount options and optional performance (IOPS) config; multiple Go modules also get dependency bumps and mockery generation is moved into the orchestrator package.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit efdf97bb4eb9f397f377de39f3a225a9cd4dd085. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/iac/provider-gcp/persistent-volumes.tf
+++ b/iac/provider-gcp/persistent-volumes.tf
@@ -9,7 +9,6 @@ module "persistent-volume-types" {
   location           = each.value.location
   network_name       = var.network_name
   nfs_version        = each.value.nfs_version
-  nfs_mount_options  = each.value.mount_options
   prefix             = var.prefix
   tier               = each.value.tier
   performance_config = each.value.performance_config


packages/orchestrator/internal/chrooted/change.go
```diff
@@ -0,0 +1,44 @@
...
+	"time"
+
+	"github.com/go-git/go-billy/v5"
+)
+
+var _ billy.Change = (*IsolatedFS)(nil)
+
+func fullPath(fs billy.Filesystem, name string) string {
+	path := fs.Join(fs.Root(), name)
+	if !strings.HasPrefix(path, "/") {
```
dobrac: what does this prevent?

packages/orchestrator/internal/chrooted/chroot.go
```diff
@@ -0,0 +1,132 @@
...
+				return fmt.Errorf("failed to chdir: %w", err)
+			}
+
+			if err = syscall.Unmount("/"+randomDirName, syscall.MNT_DETACH); err != nil {
+				return fmt.Errorf("failed to unmount: %w", err)
+			}
+
+			if err = os.Remove("/" + randomDirName); err != nil {
+				return fmt.Errorf("failed to remove %q: %w", oldRootPath, err)
+			}
```
dobrac, line 110: why the "/" prefixing?

packages/orchestrator/internal/nfsproxy/chroot/nfs.go
```diff
@@ -0,0 +1,78 @@
...
+		filesystem = unwrappable.Unwrap()
+	}
+}
+
+func (h *NFSHandler) FSStat(_ context.Context, _ billy.Filesystem, _ *nfs.FSStat) error {
+	return nil
+}
+
+func (h *NFSHandler) ToHandle(_ billy.Filesystem, _ []string) []byte {
+	panic("this should be intercepted by the caching handler")
```
dobrac, line 169: are the panics here _safe_ ?

dobrac (changes requested):
Have you considered using library for at least some parts of the functionality?

(I suppose Kubernetes ecosystem could have pretty good ones)

packages/orchestrator/internal/chrooted/change.go
djeebus: It doesn't prevent anything, it just cleans up the path.

The fs being passed in represents the volume; it also happens to be the path to the file in the mount namespace, since we mounted the volume as the root of the mount namespace. In other words, if they're looking for "/a/b/c.txt", that path is also "/a/b/c.txt" inside of the mount namespace as well. 

This code just makes sure that the path being passed in is an absolute path.

packages/orchestrator/internal/chrooted/chroot.go
djeebus: It's not strictly necessary, it just keeps it obvious where we're trying to manipulate files. after `syscall.Chdir`, we could drop the "/". It's redundant, but makes obvious what we're trying to do.

packages/orchestrator/internal/nfsproxy/chroot/nfs.go
djeebus: Depends on your definition of "safe":

- They would be caught by the "recovery" middleware, so in that sense they're safe
- These functions will never be called, as we _do_ have the caching middleware
- There's no way to return an error from this function, so panicing seems like the only option. That said, I think it does still do wonky things further up the stack.

These scenarios would only come into play if we removed the caching handler though, so these panics are to make clear to future developers that these functions absolutely won't be called at runtime.

dobrac: Thank you. As long as it doesn't crash the whole orchestrator and the crash is contained. Its fine

packages/orchestrator/internal/chrooted/mountns.go
```diff
@@ -0,0 +1,268 @@
...
+		}()
+
+		return toRun()
+	}
+
+	var wg sync.WaitGroup
+	var innerError error
+
+	wg.Go(func() {
+		runtime.LockOSThread()
```
levb: Should we have a limit on how many threads we use for this, and fail if we hit it?

packages/orchestrator/internal/chrooted/tracker.go
```diff
@@ -0,0 +1,225 @@
...
+	var (
+		sbxCount    int
+		mountCount  int
+		fsCount     int
+		deleteCount int
+	)
+
+	usedVolumeIDs := make(map[string]struct{})
+
+	for _, sbx := range c.sandboxes.Items() {
```
dobrac: this array can change between this GC function finishes its runtime

packages/orchestrator/internal/chrooted/tracker.go
```diff
@@ -0,0 +1,225 @@
...
+
+	"github.com/e2b-dev/infra/packages/orchestrator/internal"
+	"github.com/e2b-dev/infra/packages/orchestrator/internal/cfg"
+	"github.com/e2b-dev/infra/packages/orchestrator/internal/nfsproxy/chroot"
+	"github.com/e2b-dev/infra/packages/orchestrator/internal/sandbox"
+	"github.com/e2b-dev/infra/packages/orchestrator/internal/volumes"
+	"github.com/e2b-dev/infra/packages/shared/pkg/logger"
+)
+
+type FilesystemsCache struct {
```
levb: 1/5 `FilesystemsCache` is confusing - what is really "cached"? It seems more of a tracker, maybe just `Filesystems` (chroot.Filesystems)

packages/orchestrator/internal/chrooted/chroot.go
```diff
@@ -0,0 +1,132 @@
...
+		if err = syscall.Mount("", "/", "", syscall.MS_SLAVE|syscall.MS_REC, ""); err != nil {
+			return fmt.Errorf("failed to bind mount /: %w", err)
+		}
+
+		if err = syscall.Mount(path, path, "", syscall.MS_BIND|syscall.MS_REC, ""); err != nil {
+			return fmt.Errorf("failed to bind mount %q: %w", path, err)
+		}
+
+		for {
+			randomDirName := fmt.Sprintf(".old-root.%d", rand.Intn(1000000))
```
dobrac, line 80: why not use uuid? What happens if there is a collision?

packages/orchestrator/internal/nfsproxy/chroot/isolated.go
```diff
@@ -0,0 +1,132 @@
...
+
+			if err = syscall.Unmount("/"+randomDirName, syscall.MNT_DETACH); err != nil {
+				return fmt.Errorf("failed to unmount: %w", err)
+			}
+
+			if err = os.Remove("/" + randomDirName); err != nil {
+				return fmt.Errorf("failed to remove %q: %w", oldRootPath, err)
+			}
+
+			break
```
dobrac: this could be an infinite loop. We should error out after some amount of attempts

packages/orchestrator/internal/chrooted/mountns.go
djeebus: I'm genuinely not sure how many volumes is reasonable here. we'll see a max of `max sandboxes * max mounts per sandbox`. 10? 100? 1000? 10000? This may prove to be an issue.

packages/orchestrator/internal/chrooted/tracker.go
djeebus: That's not a huge issue, I think. This doesn't have to get all of them, and we'll get the ones we missed in a minute. It's just to keep the ceiling low.

packages/orchestrator/internal/chrooted/chroot.go
djeebus: random int just seemed cheaper, but we could use uuid instead. Whatever we use here will probably be fine 99.999% of the time. 

If there's a collision, `Mkdir` fails, and we loop and try again. No big deal.

If that wasn't implemented, then the consequences depend on the scenario:
- If it was a real directory in that volume, then that directory would become temporarily unusable (as we'll mount over it) and then we'll try (and fail) to delete it. Not a huge deal
- If it was a file, then we would fail to mount, as you can't mount on top of a file. The whole `chroot` function would fail, and they would be prevented from mounting it.
- If we were attempting to mount it twice simultaneously, we'd possibly end up with a mount on top of a mount, which is technically acceptable, but would be very confusing for the end user. I think it might also contain the host's root if we only unmounted it once but not twice. Definitely bad.

packages/orchestrator/internal/chrooted/tracker.go
dobrac: I meant that it can undercount - new sandbox will be spawned during the cleanup previous mounts

packages/orchestrator/internal/chrooted/chroot.go
dobrac: Ok, sounds alright

packages/orchestrator/internal/chrooted/mountns.go
dobrac: its ok to not limit it right now. I'm not sure what would the limit behavior be like

packages/orchestrator/internal/chrooted/tracker.go
djeebus: ooh. damn, that's a good point. okay, I'm going to have to fork the nfs server, I think. we need a hook for unmounting a filesystem, which we don't currently have. Might be a problem, as documented [here](https://github.com/willscott/go-nfs/pull/28)

packages/orchestrator/internal/volumes/dir_create.go
```diff
@@ -32,37 +32,21 @@ func (s *Service) CreateDir(ctx context.Context, request *orchestrator.VolumeDir
...
 
 	if request.GetCreateParents() {
-		// Create only parent directories with defaultDirMode and fix permissions against umask.
-		parent := filepath.Dir(paths.HostFullPath)
-		if err := ensureParentDirs(paths.HostVolumePath, parent, os.FileMode(defaultDirMode)); err != nil {
+		if err := fs.MkdirAll(path, os.FileMode(mode)); err != nil {
 			return nil, fmt.Errorf("failed to prepare parent directories: %w", err)
 		}
+	} else if err := fs.Mkdir(filepath.Dir(path), os.FileMode(mode)); err != nil {
+		return nil, fmt.Errorf("failed to prepare parent directories: %w", err)
```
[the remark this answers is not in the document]
djeebus: :sob: we're missing some tests here, going to add them now to make sure we don't rely on AI to point this out.

packages/orchestrator/internal/volumes/dir_list.go
```diff
@@ -45,56 +46,48 @@ func (s *Service) ListDir(ctx context.Context, request *orchestrator.VolumeDirLi
...
+		itemPath := filepath.Join(path, item.Name())
+		symlinkDest := ""
+		if item.Mode()&os.ModeSymlink != 0 {
+			// try, but if it fails to resolve, no big deal
+			symlinkDest, err = fs.EvalSymlinks(itemPath)
+			if err != nil {
+				logger.L().Warn(ctx, "failed to resolve symlink", zap.String("path", itemPath), zap.Error(err))
 			}
-
-			return nil, fmt.Errorf("failed to get info for item %s/%q: %w", paths.HostFullPath, item.Name(), err)
```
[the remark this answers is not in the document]
djeebus: That's quite alright, partial is better than none.

packages/orchestrator/internal/nfsproxy/proxy.go
```diff
@@ -31,40 +27,62 @@ type Proxy struct {
...
+			return nil
+		}
+
+		return err
+	}
+
+	return nil
+}
+
+var mountPath = regexp.MustCompile(`^/[^/]+$`)
```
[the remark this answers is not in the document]
djeebus: True, but not a real problem. We don't expose that as a feature.

packages/orchestrator/internal/volumes/service.go
```diff
@@ -2,224 +2,187 @@ package volumes
...
+func (s *Service) getFilesystem(ctx context.Context, request volumeOnly) (*chrooted.Chrooted, error) {
 	volume := request.GetVolume()
+	volumeType := volume.GetVolumeType()
 
-	// 1. Resolve Volume Type Path
-	volTypePath, ok := s.config.PersistentVolumeMounts[volume.GetVolumeType()]
-	if !ok {
-		st, _ := status.Newf(codes.NotFound, "volume type %q not found", volume.GetVolumeType()).
-			WithDetails(&orchestrator.UnknownVolumeTypeError{VolumeType: volume.GetVolumeType()})
-
```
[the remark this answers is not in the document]
djeebus: Yup, we expected this. Currently this is worth the trade off, but performance testing will come soon.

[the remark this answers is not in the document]
djeebus: True, but good enough for now. We'll make this more efficient later, if necessary

packages/orchestrator/internal/chrooted/tracker.go
djeebus: We came up with a better idea - all mounts are specific to sandboxes, which means that when a sandbox is paused or killed, we can kill the chroots

packages/orchestrator/internal/volumes/service.go
```diff
@@ -2,224 +2,195 @@ package volumes
...
-	// 3. Construct Base Path
-	basePath := filepath.Join(append([]string{volTypePath}, BuildVolumePathParts(teamID, volumeID)...)...)
-
-	// 4. Sanitize Subpath & Build Full Path
-	subPath := filepath.Clean("/" + request.GetPath()) // Forces absolute-style cleaning
-	fullPath := filepath.Join(basePath, subPath)
-
-	// 5. Security Check (Directory Traversal)
-	relPath, err := filepath.Rel(basePath, fullPath)
-	if err != nil || strings.HasPrefix(relPath, "..") {
```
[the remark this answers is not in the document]
djeebus: known issue, currently worth it for the security it brings, but we'll revisit when we start getting performance numbers.

iac/provider-gcp/persistent-volume-types/main.tf
```diff
@@ -21,4 +21,13 @@ resource "google_filestore_instance" "persistent-volumes" {
     modes   = ["MODE_IPV4"]
     network = var.network_name
   }
+
+  dynamic "performance_config" {
+    for_each = var.performance_config == null ? [] : [var.performance_config]
+    content {
+      fixed_iops {
+        max_iops = performance_config.value.max_iops
```
dobrac, line 29: do we need/want this?

iac/provider-gcp/main.tf
```diff
@@ -67,18 +67,21 @@ locals {
...
-        "nconnect=7",       // use multiple connections
-        "noac",             // disable attribute cache
-        "noacl",            // do not use an acl
-        "cto",              // enable "close-to-open" attribute checks
-        "nolock",           // do not use locking
-        "noresvport",       // use a non-privileged source port
-        "retrans=2",        // retry two times before performing recovery actions
-        "sec=sys",          // use AUTH_SYS for all requests
-        "timeo=600",        // wait 60 seconds (measured in deci-seconds) before retrying a failed request
+        "actimeo=10",           // cache directory and regular file attributes for exactly 10 seconds
```
dobrac: do we want to have any caches? Will these params work correctly when we have two sandboxes on two different orchestrators accessing the same files?

packages/orchestrator/internal/cfg/model.go
```diff
@@ -76,6 +76,8 @@ type Config struct {
 	NodeIP                     string            `env:"NODE_IP"                      envDefault:"localhost"`
 	NodeLabels                 []string          `env:"NODE_LABELS"                  envSeparator:","`
 	OrchestratorLockPath       string            `env:"ORCHESTRATOR_LOCK_PATH"       envDefault:"/orchestrator.lock"`
+	NFSProxyLogging            bool              `env:"NFS_PROXY_LOGGING"            envDefault:"false"`
+	NFSProxyTracing            bool              `env:"NFS_PROXY_TRACING"            envDefault:"true"`
```
dobrac, line 83: why having these optional?

packages/orchestrator/internal/chrooted/fs.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+func (fs *Chrooted) Stat(filename string) (info os.FileInfo, finalPath string, err error) {
+	err = fs.act(func() error {
+		info, err = os.Lstat(filename)
+		if err != nil {
+			return err
+		}
+
+		if info.Mode()&os.ModeSymlink == os.ModeSymlink {
```
dobrac: are we doing the handling manually because we want to change how the stat works or because we cannot call `os.Stat`?

packages/orchestrator/internal/nfsproxy/tracing/change.go
```diff
@@ -0,0 +1,60 @@
...
+var _ billy.Change = (*tracingChange)(nil)
+
+func newChange(ctx context.Context, change billy.Change) billy.Change {
+	return &tracingChange{ctx: ctx, inner: change}
+}
+
+func (l *tracingChange) Chmod(name string, mode os.FileMode) (err error) {
+	_, finish := startSpan(l.ctx, "Change.Chmod",
+		attribute.String("nfs.name", name),
+		attribute.String("nfs.mode", mode.String()))
```
dobrac, line 26: won't this slow down the actions significantly?

packages/orchestrator/internal/nfsproxy/tracing/conn.go
```diff
@@ -0,0 +1,20 @@
...
+)
+
+func OnConnect(ctx context.Context, _ net.Conn) context.Context {
+	ctx, _ = tracer.Start(ctx, "start nfs proxy server connection") //nolint:spancheck // called by OnDisconnect
+
+	return ctx
+}
+
+func OnDisconnect(ctx context.Context, _ net.Conn) {
+	span := trace.SpanFromContext(ctx)
```
dobrac: can we guarantee the span is from `tracer.Start(ctx, "start nfs proxy server connection")`?

packages/orchestrator/internal/nfsproxy/tracing/fs.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+func (l *tracingFS) Create(filename string) (f billy.File, err error) {
+	ctx, finish := startSpan(l.ctx, "FS.Create", attribute.String("nfs.filename", filename))
+
+	f, err = l.inner.Create(filename)
+	if f != nil {
+		f = wrapFile(ctx, f, finish)
+	} else {
+		finish(err)
```
dobrac, line 31: why not handling the error like:

```
if err != nil {
		finish(err)

		return
}

f = wrapFile(ctx, f, finish)
```

packages/orchestrator/internal/nfsproxy/tracing/fs.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+func (l *tracingFS) Join(elem ...string) string {
+	return l.inner.Join(elem...)
+}
+
+func (l *tracingFS) TempFile(dir, prefix string) (f billy.File, err error) {
+	ctx, finish := startSpan(l.ctx, "FS.TempFile",
+		attribute.String("nfs.dir", dir),
+		attribute.String("nfs.prefix", prefix))
```
dobrac, line 109: these can be slow too

packages/orchestrator/internal/volumes/service_test.go
```diff
@@ -29,122 +29,80 @@ func TestBuildVolumePath(t *testing.T) {
...
 	t.Run("success", func(t *testing.T) {
 		t.Parallel()
 
-		testCases := map[string]struct {
-			input string
-
-			expectedFullPath   string
-			expectedJailedPath string
-			expectedBasePath   string
-		}{
```
dobrac, line 50: do we still test these somewhere?

packages/orchestrator/internal/volumes/service.go
```diff
@@ -2,224 +2,195 @@ package volumes
...
-			zap.String("volume_id", volume.GetVolumeId()),
-			zap.String("team_id", volume.GetTeamId()),
-		)
-
-		return false
+	path := request.GetPath()
+	if !filepath.IsAbs(path) {
+		path = "/" + path
 	}
+	path = filepath.Clean(path)
```
dobrac, line 125: do we need to do this?

packages/orchestrator/internal/volumes/service.go
```diff
@@ -2,224 +2,195 @@ package volumes
...
-		AccessedTime:  toTimestamp(fileInfo.AccessedTime),
-		Type:          toType(fileInfo.Type),
+	if base := getBase(fileInfo.Sys()); base != nil {
+		entry.AccessedTime = toTimestampFromSpec(base.Atim)
+		entry.CreatedTime = toTimestampFromSpec(base.Ctim)
+		entry.ModifiedTime = toTimestampFromSpec(base.Mtim)
+		entry.Uid = base.Uid
+		entry.Gid = base.Gid
+	} else if !fileInfo.ModTime().IsZero() {
+		entry.ModifiedTime = toTimestampFromTime(fileInfo.ModTime())
```
dobrac: is this logic distinct from envd handling?

packages/orchestrator/internal/volumes/service.go
```diff
@@ -2,224 +2,195 @@ package volumes
...
 	}
 
 	return entry
 }
 
-func toType(fileType filesystem.FileType) orchestrator.FileType {
-	switch fileType {
-	case filesystem.DirectoryFileType:
+func getBase(sys any) *syscall.Stat_t {
```
dobrac: can you add comment explaining this function?

packages/orchestrator/internal/volumes/service.go
```diff
@@ -232,3 +203,52 @@ func setSpanStatus(span trace.Span, err error) {
 
 	span.SetStatus(otelcodes.Ok, "")
 }
+
+func ensureDirs(fs *chrooted.Chrooted, dirPath string, uid, gid uint32) error {
```
dobrac, line 188: should we move this to a different go file + add tests? Isn't there already same/similar functionality in envd?

packages/orchestrator/internal/volumes/internal_test.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/e2b-dev/infra/packages/orchestrator/internal/cfg"
+	"github.com/e2b-dev/infra/packages/orchestrator/internal/chrooted"
+	"github.com/e2b-dev/infra/packages/shared/pkg/grpc/orchestrator"
+)
+
+const (
+	volumeType = "test-vt"
+)
+
+func setupTestService(t *testing.T) (*Service, string, *orchestrator.VolumeInfo) {
```
dobrac, line 19: we should have a bit different pattern/location already for common test helper functions

packages/orchestrator/internal/volumes/fs_helpers.go
```diff
@@ -1,62 +0,0 @@
...
-import (
-	"fmt"
-	"os"
-	"path/filepath"
-)
-
-// ensureParentDirs creates all missing parent directories for dirPath (not including dirPath itself)
-// with the provided mode, then applies chmod with the same mode to only those directories that were
-// created during this call (to counteract process umask). Traversal is bounded by volRoot.
-func ensureParentDirs(volRoot, dirPath string, mode os.FileMode) error {
```
dobrac, line 12: oh, it was in a separate file already, why was it moved?

packages/orchestrator/internal/volumes/dir_list.go
```diff
@@ -22,79 +23,97 @@ const (
...
-			if errors.Is(err, os.ErrNotExist) {
-				logger.L().Warn(ctx, "item deleted during traversal", zap.String("path", filepath.Join(paths.HostFullPath, item.Name())))
-
-				continue
+		itemPath := filepath.Join(path, item.Name())
+		symlinkDest := ""
+		if item.Mode()&os.ModeSymlink != 0 {
+			// try, but if it fails to resolve, no big deal
+			var errSL error
+			symlinkDest, errSL = fs.EvalSymlinks(itemPath)
```
dobrac: what will be in the symlinkDest if it fails to resolve? We should know the destination even without resolve, right?

packages/orchestrator/internal/volumes/dir_delete.go
```diff
@@ -20,20 +22,48 @@ func (s *Service) DeleteDir(ctx context.Context, request *orchestrator.VolumeDir
...
+			return nil, newAPIError(ctx,
+				codes.NotFound,
+				http.StatusNotFound,
+				orchestrator.UserErrorCode_PATH_NOT_FOUND,
+				"failed to delete: %q not found.",
+				request.GetPath(),
+			).Err()
+		}
+
+		logger.L().Error(ctx, "failed to stat directory", zap.String("path", path), zap.Error(err))
```
dobrac: remove the log line, already returning error

packages/orchestrator/internal/volumes/dir_delete.go
```diff
@@ -20,20 +22,48 @@ func (s *Service) DeleteDir(ctx context.Context, request *orchestrator.VolumeDir
...
+	if _, _, err := fs.Stat(path); err != nil {
+		if os.IsNotExist(err) {
+			return nil, newAPIError(ctx,
+				codes.NotFound,
+				http.StatusNotFound,
+				orchestrator.UserErrorCode_PATH_NOT_FOUND,
+				"failed to delete: %q not found.",
+				request.GetPath(),
+			).Err()
+		}
```
dobrac, line 51: this if (not exists) else error pattern is in many places. Could we deduplicate it?

packages/orchestrator/internal/volumes/dir_create.go
```diff
@@ -16,77 +17,86 @@ import (
...
+	entry := toEntry(path, symlink, stat)
 
-		return nil, fmt.Errorf("failed to set directory mode: %w", err)
+	return &orchestrator.VolumeDirCreateResponse{Entry: entry}, nil
+}
+
+func processError(ctx context.Context, s string, err error) error {
+	if err == nil {
+		return nil
 	}
```
dobrac: when is this if used for example?

iac/provider-gcp/persistent-volume-types/main.tf
djeebus: Not sure how we're going to configure our persistent volumes, but if we do end up configuring this, we'll need this block to keep our terraform state in sync with production

iac/provider-gcp/main.tf
djeebus: These caches improve performance by ~4x, but only (potentially) put us in a bad state for up to 10 seconds. I think it's a trade off that we'll want to adjust in the future, but the performance hit for disabling these was real.

packages/orchestrator/internal/cfg/model.go
djeebus: I think it's not necessary to have both, and probably wasteful. Logging is more useful when coding locally, tracing is more valuable in production ... I think. Happy to change those values, but they're going to be SUPER chatty in production.

iac/provider-gcp/persistent-volume-types/main.tf
dobrac: doesn't it scale automatically to the maximum value with the provisioned space in filestore? Or is this limit for user per one mount?

packages/orchestrator/internal/chrooted/fs.go
djeebus: It's specifically because we want to return two things: metadata of the symlink, and path of the destination file. We can't do both with either lstat or stat:
- stat: returns the metadata of the destination file
- lstat: returns the metadata of the symlink file, and the location of the *next hop*

The "next hop" is important, as it's *almost* what we want, but not quite. For instance, in this scenario:

- `/a/b/c`: symlink to `/a/midlink`
- `/a/midlink`: symlink to `/finaldest`
- `/finaldest`: actual file.

Running `stat("/a/b/c")` provides metadata on `/finaldest`. Running `lstat("/a/b/c")` provides metadata on `/a/b/c`, and returns "/a/midlink" as the next hop, which isn't what we're looking for.

This is all to provide compatibility with existing envd APIs.

iac/provider-gcp/main.tf
dobrac: I understand the performance improvements. If you have two sandboxes that rely on having the same data, but the data is wrong, it's an issue. I don't think any speed improvement over correctness is in this scenario great.

packages/orchestrator/internal/nfsproxy/chroot/nfs.go
```diff
@@ -0,0 +1,180 @@
...
+	chroots := h.chrootsBySandboxID[sandboxID]
+	delete(h.chrootsBySandboxID, sandboxID)
+	h.mu.Unlock()
+
+	for _, chroot := range chroots {
+		err := chroot.Close()
+		if err != nil {
+			logger.L().Warn(context.Background(), "failed to close chroot",
+				zap.String("path", chroot.Root()),
+				zap.String("sandboxID", sandboxID),
```
dobrac: withSandboxID

packages/orchestrator/internal/nfsproxy/tracing/conn.go
djeebus: We cannot guarantee it, but it currently is due to the changes we made in the upstream library (the ctx returned from OnConnect is the same one passed into OnDisconnect). We can make this more explicitly by wrapping and returning the connection though. I'll do that today

packages/orchestrator/internal/nfsproxy/tracing/fs.go
djeebus: AI slop that was good enough =P I'll have it fix those to be less ... silly.

packages/orchestrator/internal/nfsproxy/chroot/nfs.go
```diff
@@ -0,0 +1,180 @@
...
+	if volumeMount.ID == uuid.Nil {
+		return nil, ErrVolumeID
+	}
+
+	fs, err := h.builder.Chroot(ctx, volumeMount.Type, teamID, volumeMount.ID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to mount %q: %w", volumeName, err)
+	}
+
+	sandboxID := sbx.Metadata.Runtime.SandboxID
```
dobrac, line 138: should we use the execution ID instead?

packages/orchestrator/internal/volumes/service_test.go
djeebus: No reason to test these in particular, because there's no longer any path traversal to escape from. If they want to get "../../../../../etc/passwd", they can do that, and it works fine (resolving to the file from within the jail).

packages/orchestrator/internal/nfsproxy/chroot/nfs.go
```diff
@@ -0,0 +1,180 @@
...
+	h.mu.Lock()
+	h.chrootsBySandboxID[sandboxID] = append(h.chrootsBySandboxID[sandboxID], fs)
+	h.mu.Unlock()
+
+	return fs, nil
+}
+
+func (h *NFSHandler) Change(_ context.Context, filesystem billy.Filesystem) billy.Change {
+	for {
```
dobrac, line 147: could this be an infinite loop? Should we do some limit to the depth?

packages/orchestrator/internal/volumes/service.go
djeebus: We need the `"/" + path`, but there's no longer much value in `filepath.Clean`. It probably doesn't hurt to leave it in there, and it might help on some future edge case ("defense in depth").

packages/orchestrator/internal/nfsproxy/chroot/nfs.go
```diff
@@ -0,0 +1,180 @@
...
+		if !ok {
+			panic(fmt.Sprintf("no idea how to find an *Chrooted from this filesystem: %T", filesystem))
+		}
+
+		filesystem = unwrappable.Unwrap()
+	}
+}
+
+func (h *NFSHandler) FSStat(_ context.Context, _ billy.Filesystem, _ *nfs.FSStat) error {
+	return nil
```
dobrac, line 165: is it ok this is nil and not an error?

packages/orchestrator/internal/chrooted/chroot.go
```diff
@@ -0,0 +1,132 @@
...
+
+	return fs, nil
+}
+
+const maxMountAttempts = 10
+
+var ErrFailedToMount = errors.New("retries exhausted")
+
+func (fs *Chrooted) chroot(path string) error {
+	return fs.ns.Do(func() error {
```
dobrac: is the fs.ns.Do used here instead of fs.act here deliberately for some reason?

packages/orchestrator/internal/chrooted/mountns.go
```diff
@@ -0,0 +1,247 @@
...
+	return nil
+}
+
+func (ns *mountNS) Path() string {
+	return ns.file.Name()
+}
+
+func (ns *mountNS) Fd() uintptr {
+	return ns.file.Fd()
+}
```
dobrac, line 106: should there be a lock around these? If not, does there need to be a lock on line 71?

packages/orchestrator/internal/chrooted/mountns.go
```diff
@@ -0,0 +1,247 @@
...
+			resultCh <- result{err: fmt.Errorf("failed to unshare namespace: %w", err)}
+			runtime.UnlockOSThread()
+
+			return
+		}
+
+		// get the mount namespace that we just created
+		tempNS, err := getCurrentNSNoLock()
+		if err != nil {
+			_ = threadNS.Set()
```
dobrac: why ignoring the error?

packages/orchestrator/internal/volumes/service.go
djeebus: I think so? It was quite similar, but at one point it was different in ways that were useful in the moment. I can try to unify this logic again. 

I think the biggest difference was UID/GID vs username / group name. We can't get the names reliably without a sandbox.

packages/orchestrator/internal/volumes/fs_helpers.go
djeebus: iirc it was the last function there, and it felt silly to have a whole file for one function. I can move it back if you feel strongly.

packages/orchestrator/internal/volumes/dir_list.go
djeebus: If it fails to resolve, we'll have the last final leg. 

For example:
- `/a` symlinks to `/b`
- `/b` symlinks to `/c`
- `/c` does not exist

We'll return `/b`.

packages/orchestrator/internal/chrooted/fs.go
dobrac: Ok, I was asking because its the `Stat` action, so I would assume it would behave same as the os.Stat function

packages/orchestrator/internal/volumes/service_test.go
dobrac: I think we might want to still test for these cases, even if they should work fine as of now

packages/orchestrator/internal/volumes/dir_delete.go
djeebus: Not in a way that makes our code appropriately flexible in the future. At one point these errors were translated in a central way, but that made it really non-obvious when errors *should* have returned a custom error but weren't, or needed to return a slightly different error - couldn't do it b/c of the centralization. This may look verbose, but they're occasionally slightly different. I can centralize the creation of that error message, but then again, that makes it non obvious when the wrong thing was done.

Right now, every error that returns a user-friendly message is returned via `newAPIError` in the rpc handler. Anything that's *not* should receive the grpc equivalent of "internal server error". We can move some of that code to a `newPathNotFoudnAPIError` function, but I don't think we'd want to move the logic. API error handling should (imho) very explicit and clear. It's easy to return the wrong error message when you hide the error creation.

packages/orchestrator/internal/volumes/dir_create.go
djeebus: Oops, relic from old usage. Removed.

packages/orchestrator/internal/volumes/service.go
dobrac: okay, would be nice to unify so they won't drift as the logic is not entirely non-trivial

packages/orchestrator/internal/volumes/fs_helpers.go
dobrac: not strong opinion, the tests are more mportant

packages/orchestrator/internal/nfsproxy/chroot/nfs.go
djeebus: That feels like it'd be more correct, but for one problem: the sandbox map notifies subscribers of removals by SandboxID. It supports removing by LifecycleID (is that different than ExecutionID?) but still only notifies by SandboxID. We could refactor it to get access to the whole sandbox being removed though, I suspect.

packages/orchestrator/internal/nfsproxy/chroot/nfs.go
djeebus: A wrapped filesystem would have to wrap itself somehow, which none of ours support. This is (almost) entirely our invention, at most we have:

- recovery
- logged
- tracing
- caching
- chroot

packages/orchestrator/internal/volumes/dir_list.go
dobrac: double checking, it's not gonna rewrite the symlinkDest with `""`, correct?

packages/orchestrator/internal/nfsproxy/chroot/nfs.go
djeebus: The client uses this for identifying things like total size, free size, total/free files, etc. Not necessary to implement, and by not modifying it, we end up with `1 << 62` for all values, which is ... probably fine. We're telling customers they have infinite space, this just confirms it. 

Returning an error here would tell the client that the request has failed, which would lead to bad behavior.

packages/orchestrator/internal/chrooted/chroot.go
djeebus: It only avoids a lock, which wouldn't be useful here.

packages/orchestrator/internal/chrooted/mountns.go
djeebus: Both of those functions are concurrency-safe, no need for a lock.

packages/orchestrator/internal/chrooted/mountns.go
djeebus: This is AI code, but I assume to avoid err collisions. Plus we're already in a bad state (get current ns failed) so it's likely that setting the threadns will fail as well. I'll log it out though, could reveal an issue.

packages/orchestrator/internal/volumes/service.go
djeebus: envd's implementation has a hardcoded directory permission that we don't necessary want, and skips a chmod to fix umask as well. I don't know the context well enough to know whether that was intentional or not, but didn't want differences introduced during refactor to delay roll out. We can follow up with a refactor and focused testing in the future, but I don't think there's a reason to add "regression test envd directory creation" to the test plan for this PR.

packages/orchestrator/internal/volumes/dir_delete.go
dobrac: ok

packages/orchestrator/internal/nfsproxy/chroot/nfs.go
dobrac: Sorry, I've meant the LifecycleID.

Yes, it makes sense to add the lifecycle ID to the hook too, but lets keep it with the sandboxID for now. I don't think it has large issues then

packages/orchestrator/internal/nfsproxy/chroot/nfs.go
dobrac: Right, understood, makes sense. Can you add small explanation comment there please?

packages/orchestrator/internal/chrooted/chroot.go
dobrac: -> just trying if we can't remove the `ns` from the struct

packages/orchestrator/internal/chrooted/fs.go
djeebus: Same! That's how the first version worked, but @mishushakov pointed out that we wanted it to function the same as envd, which has this extra functionality.

packages/orchestrator/internal/nfsproxy/tracing/change.go
djeebus: Yeah, it looks like, even after disabling stat and handle spans, tracing slows things down by maybe 5-10%. I think what I'd like to do is find a way to turn on tracing at runtime, but only for certain requests. Maybe by team id, sandbox id, something like that. It's definitely too slow to leave on for everyone all the time, but I suspect it'll be a useful runtime diagnostic tool.

packages/orchestrator/internal/volumes/fs_helpers.go
djeebus: It does have tests in service_test.go

packages/orchestrator/internal/nfsproxy/tracing/change.go
djeebus: Refactoring this is going to take a bit of time. We'll make this a follow up task.

iac/provider-gcp/main.tf
djeebus: Definitely agree there. To be clear, this is caching attributes, not data (size, permissions, mod times, things like that). So the only data that might be stale would be "how big is this file?" or "is this file executable?" things like that. Still worth disabling?

Similarly, the `lookupcache=positive` caches whether or not files exist. None of this caches data, only file existence.

packages/orchestrator/internal/volumes/dir_list.go
djeebus: Seems like I didn't understand correctly what EvalSymlinks did. Refactoring to reuse shared functions.

packages/orchestrator/.claude/settings.local.json
```diff
@@ -0,0 +1,8 @@
+{
+  "permissions": {
+    "allow": [
+      "Bash(sudo go test:*)",
+      "Bash(go build:*)"
+    ]
+  }
```
dobrac: is this ok to commit?

packages/orchestrator/internal/chrooted/chroot.go
```diff
@@ -61,19 +44,30 @@ func Chroot(ctx context.Context, source string, opts ...Option) (*Chrooted, erro
...
 	for _, opt := range opts {
 		opt(fs)
 	}
 
 	return fs, nil
 }
 
+func (fs *Chrooted) act(fn func() error) error {
+	return fs.ns.Do(func() error {
+		return fn()
```
dobrac: do we not need the lock anymore?

packages/orchestrator/internal/chrooted/mountns.go
```diff
@@ -6,11 +6,25 @@ import (
...
+	"go.opentelemetry.io/otel/metric"
 	"go.uber.org/zap"
 	"golang.org/x/sys/unix"
 
 	"github.com/e2b-dev/infra/packages/shared/pkg/logger"
+	"github.com/e2b-dev/infra/packages/shared/pkg/utils"
+)
+
+var (
+	meter = otel.GetMeterProvider().Meter("shared.pkg.storage")
```
dobrac: lets name it so we know its not related to the sandbox "storage", but volumes

packages/orchestrator/internal/chrooted/mountns.go
```diff
@@ -207,7 +221,10 @@ func tempMountNS(ctx context.Context) (*mountNS, error) {
 		// get the mount namespace that we just created
 		tempNS, err := getCurrentNSNoLock()
 		if err != nil {
-			_ = threadNS.Set()
+			if err2 := threadNS.Set(); err2 != nil {
+				logger.L().Error(ctx, "failed to restore original namespace", zap.Error(err2))
```
dobrac: nit: do we have more metadata for the log message?

packages/orchestrator/internal/chrooted/mountns.go
```diff
@@ -223,9 +240,15 @@ func tempMountNS(ctx context.Context) (*mountNS, error) {
...
 			select {
 			case req := <-tempNS.reqCh:
+				delay := time.Since(req.start)
+				requestLatency.Record(ctx, delay.Microseconds())
+
 				req.done <- req.fn()
 			case <-tempNS.stopCh:
-				_ = threadNS.Set()
+				if err2 := threadNS.Set(); err2 != nil {
+					logger.L().Error(ctx, "failed to restore original namespace", zap.Error(err2))
```
dobrac: same here with the metadata

packages/orchestrator/internal/nfsproxy/spanconn.go
```diff
@@ -0,0 +1,36 @@
...
+
+	conn = wrapConn(conn, span)
+
+	return ctx, conn //nolint:spancheck // called by OnDisconnect
+}
+
+func onDisconnect(_ context.Context, conn net.Conn) {
+	cws, ok := conn.(*connWithSpan)
+	if ok {
+		cws.span.End()
```
dobrac: we should add some log when this is not ok, because that means we're leaking memory

packages/orchestrator/.claude/settings.local.json
djeebus: Oops, sorry, nope, not okay. This got by me before I added it to my global ignore =/

packages/orchestrator/internal/chrooted/chroot.go
djeebus: Nope! The mountNS already has a lock, so this was redundant.

PR #2093, opened 2026-03-11
feat: dynamic ingress control for running sandboxes

levb:
## Summary

  Extends `PUT /sandboxes/{sandboxID}/network` with ingress filtering and host masking. All fields (egress + ingress) are replaced on every PUT — clients must send the full desired network config.

### Ingress filtering (`allowIn` / `denyIn`)
  - Allow/deny rules by client IP/CIDR V4 with optional port or port-range (e.g. `10.0.0.0/8:80-443`)
  - `allowIn` requires `denyIn` to include `0.0.0.0/0` (explicit deny-all-first)
  - Allow takes precedence over deny; unmatched traffic is allowed by default
  - Domains not supported
  - envd port is exempt from ingress restrictions

  ### Host masking (`maskRequestHost`)
  - Replaces the Host header for all proxied sandbox requests
  - Supports `{port}` placeholder for dynamic port substitution
  - Set to `null` to clear

  ### Client IP flow
  - Client proxy extracts real client IP from `X-Forwarded-For` (second-to-last entry, prevents spoofing)
  - Sets internal `X-E2B-Client-IP` header, stripped by the orchestrator proxy before reaching the sandbox

  ### Pre-parsed ACLs
  - Egress and ingress rules parsed once at config time into `Egress`/`Ingress` structs (`shared/pkg/sandbox-network/`)
  - Egress ACL used by TCP firewall; ingress ACL evaluated per-request in orchestrator HTTP proxy

  ### Other
  - `AllowPublicAccess` is preserved across updates (not changeable via PUT)
  - Network update applied transactionally on orchestrator (egress + ingress); rolled back on failure
  - Browser-friendly HTML error page for denied ingress requests


packages/orchestrator/internal/proxy/proxy.go
```diff
@@ -88,6 +89,32 @@ func NewSandboxProxy(meterProvider metric.MeterProvider, port uint16, sandboxes
 				}
 			}
 
+			if isNonEnvdTraffic {
+				// Handle port allowlist/denylist for non-envd traffic.
+				// Priority: allow wins → deny → default allow.
+				if !containsPort(ingress.GetAllowedPorts(), port) {
+					if containsPort(ingress.GetDeniedPorts(), port) {
+						return nil, reverseproxy.NewErrPortNotAllowed(sandboxId, port)
+					}
```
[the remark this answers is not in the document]
levb: Understood. This was an attempt to mimic the egress logic in base. It's probably much more intuitive for the user to treat allowPorts list as exclusive if denyPorts is empty; or maybe a * to express all ports?

spec/openapi.yml
```diff
@@ -282,6 +282,30 @@ components:
...
+            minimum: 1
+            maximum: 65535
+        denyPorts:
+          type: array
+          description: List of ports blocked from outside access. Ignored if allowPorts is set.
+          items:
+            type: integer
+            minimum: 1
+            maximum: 65535
+        allowIn:
```
dobrac, line 285: I think we could merge the allowPorts with allow in so you could do:

allowIn: ["1.1.1.1", ":3003", "8.8.4.4:80"]

spec/openapi.yml
```diff
@@ -2357,6 +2381,34 @@ paths:
                   description: List of denied CIDR blocks or IP addresses for egress traffic. Domain names are not supported for deny rules.
                   items:
                     type: string
+                allowPorts:
```
dobrac: same as for sbx create

tests/integration/internal/tests/api/sandboxes/sandbox_network_update_test.go
```diff
@@ -336,3 +343,404 @@ func TestUpdateNetworkConfig(t *testing.T) { //nolint:tparallel // subtests are
...
+	r.body.AllowIn = &cidrs
+
+	return r
+}
+
+func (r *ingressRules) maskHost(h string) *ingressRules {
+	r.body.MaskRequestHost = &h
+
+	return r
+}
```
dobrac: do we need to construct it like this and not use the struct directly instead?

packages/orchestrator/internal/proxy/proxy.go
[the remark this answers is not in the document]
levb: 2/5 we make this simple - allowPorts and denyPorts are mutually exclusive. `allowPorts` assumes everything else is denied, and denyPorts assumes everything else is allowed.

dobrac (changes requested):
Discussed IRL. We will go for parity with the egress firewall and combine ports with cidrs.

packages/shared/pkg/proxy/template/browser_ingress_denied.html
```diff
@@ -0,0 +1,22 @@
...
+    <img src="https://hebbkx1anhila5yf.public.blob.vercel-storage.com/Symbol%20Gradient-Kr5pnWlK3ZhzBcRGf6Am4cNbJvY1Ge.svg" alt="Logo" class="logo">
+    <div class="error-header">
+        <h1 class="error-title">Access Denied</h1>
+        <p class="error-message">{{.Message}}</p>
+    </div>
+    <div class="error-details">
+        <span class="sandbox-url">{{.Host}}</span>
+        <div class="error-code">Blocked by the sandbox network policy for sandbox <span class="highlight">{{.SandboxId}}</span></div>
+    </div>
+    <p class="help-text">The sandbox owner has restricted access to this sandbox. Contact them if you believe this is an error.</p>
```
dobrac:
```suggestion
    <p class="help-text">The sandbox has restricted access. It can not be reached {{ .IP }} on port {{.Port}}</p>
```

packages/api/internal/handlers/sandbox_create.go
```diff
@@ -504,46 +507,134 @@ func validateNetworkConfig(network *api.SandboxNetworkConfig) *api.APIError {
...
+	allowIn := sharedUtils.DerefOrDefault(network.AllowIn, nil)
+
+	return validateIngressRules(allowIn, denyIn)
 }
 
 // validateEgressRules validates egress allow/deny rules:
-// - denyOut entries must be valid IPs or CIDRs (not domains)
-// - allowOut entries must be valid IPs, CIDRs, or domain names
+// - denyOut entries must be specified IPs or CIDRs (not domains, no ports)
+// - allowOut entries can be specified IPs, CIDRs, or domain names (no ports)
```
dobrac:
```suggestion
// - denyOut entries must be specified IPs or CIDRs (not domains)
// - allowOut entries can be specified IPs, CIDRs, or domain names
```

packages/api/internal/handlers/sandbox_create.go
```diff
@@ -504,46 +507,134 @@ func validateNetworkConfig(network *api.SandboxNetworkConfig) *api.APIError {
...
+				Code:      http.StatusBadRequest,
+				Err:       fmt.Errorf("invalid allow out entry %q: port-specific rules are not supported for egress", entry),
+				ClientMsg: fmt.Sprintf("invalid allow out entry %q: port-specific rules are not supported for egress", entry),
+			}
+		}
 
-		for _, addr := range allowedAddresses {
-			if !sandbox_network.IsSpecifiedIPOrCIDR(addr) {
+		if sandbox_network.IsIPOrCIDR(host) {
+			if !sandbox_network.IsSpecifiedIPOrCIDR(host) {
```
dobrac, line 581: why not use the `ParseAddressesAndDomains`?

packages/api/internal/handlers/sandbox_create.go
```diff
@@ -504,46 +507,134 @@ func validateNetworkConfig(network *api.SandboxNetworkConfig) *api.APIError {
...
+	if err != nil {
+		return err
+	}
+
+	if !sandbox_network.IsIPOrCIDR(host) {
+		return fmt.Errorf("domains are not supported for ingress rules")
+	}
+
+	// IsSpecifiedIPOrCIDR allows 0.0.0.0/0 but not ::/0; ingress needs both.
+	if !sandbox_network.IsSpecifiedIPOrCIDR(host) && host != "::/0" {
```
dobrac: we don't support ipv6, lets remove it from the checks

packages/api/internal/handlers/sandbox_create.go
```diff
@@ -504,46 +507,134 @@ func validateNetworkConfig(network *api.SandboxNetworkConfig) *api.APIError {
...
+// validateIngressEntry validates a single ingress rule entry (CIDR[:port]).
+func validateIngressEntry(entry string) error {
+	host, portStr, err := sandbox_network.SplitHostPort(entry)
+	if err != nil {
+		return err
+	}
+
+	if !sandbox_network.IsIPOrCIDR(host) {
+		return fmt.Errorf("domains are not supported for ingress rules")
+	}
```
dobrac, line 630: I would prefer to use the `ParseAddressesAndDomains` to split them explicitly first (we can rename it to `SplitAddressesAndDomains`)

packages/api/internal/handlers/sandbox_create_test.go
```diff
@@ -240,6 +259,91 @@ func TestValidateNetworkConfig(t *testing.T) {
 			},
 			wantErr: false,
 		},
+		// Ingress port validation tests (unified CIDR:port format)
+		{
+			name: "valid allowIn with port",
+			network: &api.SandboxNetworkConfig{
+				AllowIn: &[]string{"0.0.0.0/0:80", "0.0.0.0/0:443"},
```
dobrac, line 241: lets make the `:80` also supported input

packages/api/internal/handlers/sandbox_create_test.go
```diff
@@ -240,6 +259,91 @@ func TestValidateNetworkConfig(t *testing.T) {
...
+			network: &api.SandboxNetworkConfig{
+				AllowIn: &[]string{"10.0.0.0/8"},
+				DenyIn:  &[]string{"0.0.0.0/0"},
+			},
+			wantErr: false,
+		},
+		{
+			name: "valid allowIn CIDR from IP with deny-all",
+			network: &api.SandboxNetworkConfig{
+				AllowIn: &[]string{"1.2.3.4/32"},
```
dobrac, line 295: `1.2.3.4` should be checked as well as valid input

dobrac: that means also `1.2.3.4:80`

packages/api/internal/handlers/sandbox_create_test.go
```diff
@@ -259,6 +363,43 @@ func TestValidateNetworkConfig(t *testing.T) {
 			},
 			wantErr: false,
 		},
+		// Port-specific egress rules are rejected
+		{
+			name: "allow_out with port is rejected",
+			network: &api.SandboxNetworkConfig{
+				AllowOut: &[]string{"8.8.8.8:80"},
```
dobrac: this test is duplicit

packages/api/internal/handlers/sandbox_create_test.go
```diff
@@ -259,6 +363,43 @@ func TestValidateNetworkConfig(t *testing.T) {
...
+			},
+			wantErr:    true,
+			wantCode:   http.StatusBadRequest,
+			wantErrMsg: `invalid allow out entry "8.8.8.8:80": port-specific rules are not supported for egress`,
+		},
+		{
+			name: "deny_out with port is rejected",
+			network: &api.SandboxNetworkConfig{
+				DenyOut: &[]string{"10.0.0.0/8:22"},
```
dobrac: this test is duplicit too

packages/api/internal/orchestrator/create_instance.go
```diff
@@ -31,23 +31,70 @@ import (
...
+	var allowedAddresses []string
+	var allowedDomains []string
+
+	for _, entry := range egressUpdate.AllowedAddresses {
+		if sandbox_network.IsIPOrCIDR(entry) {
+			allowedAddresses = append(allowedAddresses, sandbox_network.AddressStringToCIDR(entry))
+		} else {
+			allowedDomains = append(allowedDomains, entry)
+		}
+	}
```
dobrac: why moving this outside of the parsing function?

packages/api/internal/orchestrator/create_instance.go
```diff
@@ -31,23 +31,70 @@ import (
...
-// allow/deny entry lists. It splits allowed entries into CIDRs and domains,
-// and adds the default nameserver when domains are present so the sandbox can
-// resolve them.
-func buildEgressConfig(allowedEntries, deniedEntries []string) *orchestrator.SandboxNetworkEgressConfig {
-	allowedAddresses, allowedDomains := sandbox_network.ParseAddressesAndDomains(allowedEntries)
+// allow/deny entry lists. The API has already validated these entries.
+// CIDRs and domains are split into their respective proto fields.
+// When any allowed entry is a domain, the default nameserver is injected so
+// the sandbox can resolve domain names.
+func buildEgressConfig(egressUpdate *types.SandboxNetworkEgressConfig) *orchestrator.SandboxNetworkEgressConfig {
```
dobrac, line 38: as the function declaration has changed, maybe we should rename it as it now converts the config to orchestrator config

packages/api/internal/orchestrator/create_instance.go
```diff
@@ -31,23 +31,70 @@ import (
...
 	}
 }
 
+// parseIngressRules converts pre-validated ingress rule strings (CIDR[:port[-port]])
+// into structured IngressRule protos. The API has already validated these entries
+// as valid CIDRs with optional port ranges, so parsing errors are silently skipped.
+func parseIngressRules(entries []string) []*orchestrator.IngressRule {
+	if len(entries) == 0 {
+		return nil
+	}
```
dobrac: this if is redundant

packages/api/internal/orchestrator/create_instance.go
```diff
@@ -31,23 +31,70 @@ import (
...
+
+	rules := make([]*orchestrator.IngressRule, 0, len(entries))
+	for _, entry := range entries {
+		host, portStr, err := sandbox_network.SplitHostPort(entry)
+		if err != nil {
+			continue // pre-validated by API handler
```
dobrac: lets keep the error, even if its validated by the API handler

packages/api/internal/orchestrator/create_instance.go
```diff
@@ -31,23 +31,70 @@ import (
...
+		}
+
+		cidr := sandbox_network.AddressStringToCIDR(host)
+
+		var portLow, portHigh uint32
+		if portStr != "" {
+			lo, hi, err := sandbox_network.ParsePortRange(portStr)
+			if err != nil {
+				continue // pre-validated by API handler
```
dobrac: same here

packages/api/internal/orchestrator/update_network.go
```diff
@@ -38,11 +44,20 @@ func (o *Orchestrator) UpdateSandboxNetworkConfig(
...
+
+		// Preserve AllowPublicAccess from creation — it can't be changed via
+		// PUT because the traffic access token isn't returned to clients.
+		var previousAllowPublic *bool
+		if sbx.Network.Ingress != nil {
+			previousAllowPublic = sbx.Network.Ingress.AllowPublicAccess
 		}
 
+		sbx.Network.Ingress = ingressUpdate
+		sbx.Network.Ingress.AllowPublicAccess = previousAllowPublic
```
dobrac: lets do this update on the `ingressUpdate` and switch lines 56, 57

packages/orchestrator/internal/proxy/proxy.go
```diff
@@ -87,8 +88,25 @@ func NewSandboxProxy(meterProvider metric.MeterProvider, port uint16, sandboxes
...
+				}
+
+				if !ingressACL.IsAllowed(ip, uint16(port)) {
+					return nil, reverseproxy.NewErrIngressDenied(sandboxId, clientIP, uint16(port))
+				}
+			}
+
+			// Strip the internal client IP header so it never reaches the sandbox.
+			// Must happen after extractClientIP reads it, and before Rewrite clones r into r.Out.
+			r.Header.Del(reverseproxy.ClientIPHeader)
```
dobrac: this might be a breaking change, correct?

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -78,10 +80,14 @@ type Config struct {
...
-	// mu protects mutable sub-fields of Network (Egress, Ingress).
-	// The Network pointer itself is set once at construction and never replaced.
+	// mu protects mutable sub-fields: Network (Egress, Ingress) and ACL pointers.
+	// ACL values are immutable once created — only the pointer is swapped.
 	mu      *sync.RWMutex
 	Network *orchestrator.SandboxNetworkConfig
+
+	// Pre-parsed ACLs — each ACL is immutable; pointer swapped under mu on update.
+	egressACL  *sandboxnetwork.ACL
+	ingressACL *sandboxnetwork.ACL
```
dobrac: lets do a new struct for the SandboxNetworkConfig that has only the info that is not present in the ACLs

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -120,6 +139,99 @@ func (c *Config) GetNetworkIngress() *orchestrator.SandboxNetworkIngressConfig {
...
+
+	if len(allowedRules) == 0 && len(deniedRules) == 0 {
+		return nil
+	}
+
+	return &sandboxnetwork.ACL{
+		Allowed: allowedRules,
+		Denied:  deniedRules,
+	}
+}
```
dobrac: these two can be united to one function

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -120,6 +139,99 @@ func (c *Config) GetNetworkIngress() *orchestrator.SandboxNetworkIngressConfig {
...
+	return out
+}
+
+// parseIngressRules converts proto IngressRules directly to parsed Rules.
+func parseIngressRules(rules []*orchestrator.IngressRule) []sandboxnetwork.Rule {
+	out := make([]sandboxnetwork.Rule, 0, len(rules))
+	for _, r := range rules {
+		_, ipNet, err := net.ParseCIDR(r.GetCidr())
+		if err != nil {
+			continue // pre-validated by API
```
dobrac: keep the error

packages/orchestrator/internal/sandbox/sandbox.go
```diff
@@ -120,6 +139,99 @@ func (c *Config) GetNetworkIngress() *orchestrator.SandboxNetworkIngressConfig {
...
+		}
+
+		out = append(out, sandboxnetwork.Rule{Host: cidr, IPNet: ipNet})
+	}
+
+	return out
+}
+
+// parseIngressRules converts proto IngressRules directly to parsed Rules.
+func parseIngressRules(rules []*orchestrator.IngressRule) []sandboxnetwork.Rule {
```
dobrac: lets move all the transformations to a separate file (at least)

packages/orchestrator/internal/server/sandboxes.go
```diff
@@ -311,13 +313,36 @@ func (s *Server) Update(ctx context.Context, req *orchestrator.SandboxUpdateRequ
...
+	if req.GetIngress() != nil {
+		updates = append(updates, func(_ context.Context) (func(context.Context), error) {
+			oldIngress := sbx.Config.GetNetworkIngress()
+
+			sbx.Config.SetNetworkIngress(req.GetIngress())
+
+			ingress := req.GetIngress()
+			eventData["network_ingress"] = map[string]any{
+				"allowed":           ingress.GetAllowed(),
+				"denied":            ingress.GetDenied(),
```
dobrac: what data will this return?

packages/orchestrator/pkg/tcpfirewall/handlers.go
```diff
@@ -165,48 +149,40 @@ func proxyWithIPVerification(ctx context.Context, conn net.Conn, upstreamAddr st
 //  1. Allow domain / Allow CIDR (if either matches → allow)
 //  2. Deny domain / Deny CIDR (if either matches → deny)
 //  3. Default: allow
-func isEgressAllowed(sbx *sandbox.Sandbox, hostname string, ip net.IP) (bool, MatchType, error) {
+func isEgressAllowed(sbx *sandbox.Sandbox, hostname string, ip net.IP) (bool, MatchType) {
 	egress := sbx.Config.GetNetworkEgress()
 	if egress == nil {
-		// No egress configuration, allow all traffic.
```
dobrac, line 171: keep the comment

packages/orchestrator/pkg/tcpfirewall/handlers.go
```diff
@@ -165,48 +149,40 @@ func proxyWithIPVerification(ctx context.Context, conn net.Conn, upstreamAddr st
...
-		return true, MatchTypeNone, nil
+		return true, MatchTypeNone
```
dobrac, line 154: lets move the default `None` to a const

packages/orchestrator/internal/tcpfirewall/handlers.go
```diff
@@ -165,48 +149,40 @@ func proxyWithIPVerification(ctx context.Context, conn net.Conn, upstreamAddr st
...
+		return true, MatchTypeNone
 	}
 
-	// Priority 2: Check denied CIDRs
-	for _, cidr := range egress.GetDeniedCidrs() {
-		_, ipNet, err := net.ParseCIDR(cidr)
-		if err != nil {
-			return false, MatchTypeNone, fmt.Errorf("invalid denied CIDR %q: %w", cidr, err)
+	for i := range acl.Allowed {
+		if acl.Allowed[i].ContainsIP(ip) {
```
dobrac: why not use the iterated value and access the value by the index?

packages/orchestrator/internal/tcpfirewall/handlers.go
```diff
@@ -165,48 +149,40 @@ func proxyWithIPVerification(ctx context.Context, conn net.Conn, upstreamAddr st
...
+			return true, MatchTypeCIDR
 		}
+	}
 
-		if ipNet.Contains(ip) {
-			return false, MatchTypeCIDR, nil // Blocked by CIDR
+	for i := range acl.Denied {
+		if acl.Denied[i].ContainsIP(ip) {
```
dobrac: same here

packages/shared/pkg/proxy/template/ingress_error.go
```diff
@@ -0,0 +1,39 @@
...
+	Message   string `json:"message"`
+	Code      int    `json:"code"`
+	Host      string `json:"-"`
+}
+
+func (e ingressDeniedData) StatusCode() int {
+	return e.Code
+}
+
+func NewIngressDeniedError(sandboxId string, host string, clientIP string, port uint16) *TemplatedError[ingressDeniedData] {
```
dobrac:
```suggestion
func NewIngressDeniedError(sandboxID string, host string, clientIP string, port uint16) *TemplatedError[ingressDeniedData] {
```

packages/shared/pkg/proxy/errors.go
```diff
@@ -82,6 +82,24 @@ func NewErrInvalidTrafficAccessToken(sandboxId string, header string) *InvalidTr
 	}
 }
 
+type IngressDeniedError struct {
+	SandboxId string
+	ClientIP  string
+	Port      uint16
+}
+
+func NewErrIngressDenied(sandboxId string, clientIP string, port uint16) *IngressDeniedError {
```
dobrac, line 91:
```suggestion
func NewErrIngressDenied(sandboxID string, clientIP string, port uint16) *IngressDeniedError {
```

spec/openapi.yml
```diff
@@ -271,17 +271,40 @@ components:
           description: Specify if the sandbox URLs should be accessible only with authentication.
         allowOut:
           type: array
-          description: List of allowed destinations for egress traffic. Each entry can be a CIDR block (e.g. "8.8.8.8/32"), a bare IP address (e.g. "8.8.8.8"), or a domain name (e.g. "example.com", "*.example.com"). Allowed entries always take precedence over denied entries.
```
dobrac, line 274: why changing the comments?

spec/openapi.yml
```diff
@@ -271,17 +271,40 @@ components:
...
+          description: >
+            Allowed egress destinations. Format: IP, CIDR, domain, or wildcard
+            (e.g. "8.8.8.8", "10.0.0.0/8", "example.com", "*.example.com").
+            Allow takes precedence over deny. Domains require denyOut to include "0.0.0.0/0".
+            Unmatched traffic is allowed by default.
           items:
             type: string
         denyOut:
           type: array
-          description: List of denied CIDR blocks or IP addresses for egress traffic. Domain names are not supported for deny rules.
```
dobrac, line 279: same here

spec/openapi.yml
```diff
@@ -271,17 +271,40 @@ components:
...
+          description: >
+            Denied egress destinations. Format: IP or CIDR
+            (e.g. "8.8.8.8", "10.0.0.0/8"). Domains not supported. Use "0.0.0.0/0" to block all.
+            Unmatched traffic is allowed by default.
           items:
             type: string
         maskRequestHost:
           type: string
-          description: Specify host mask which will be used for all sandbox requests
```
dobrac, line 284: unnecessary change too

spec/openapi.yml
```diff
@@ -2349,14 +2372,28 @@ paths:
               properties:
                 allowOut:
                   type: array
-                  description: List of allowed destinations for egress traffic. Each entry can be a CIDR block (e.g. "8.8.8.8/32"), a bare IP address (e.g. "8.8.8.8"), or a domain name (e.g. "example.com", "*.example.com"). Allowed entries always take precedence over denied entries.
```
dobrac, line 2385: I think lets keep it duplicit

packages/shared/pkg/sandbox-network/rule.go
```diff
@@ -0,0 +1,133 @@
...
+// IsAllowed checks if an IP + port combination is allowed by the ACL.
+// Priority: allow wins → deny → default allow.
+// Returns true when the ACL is nil (no rules).
+func (a *ACL) IsAllowed(ip net.IP, port uint16) bool {
+	if a == nil {
+		return true
+	}
+
+	for i := range a.Allowed {
+		if a.Allowed[i].ContainsIP(ip) && a.Allowed[i].PortInRange(port) {
```
dobrac: the same thing with range, why not to use the value from the range directly?

packages/shared/pkg/sandbox-network/rule.go
```diff
@@ -0,0 +1,133 @@
...
+			return true
+		}
+	}
+
+	for i := range a.Denied {
+		if a.Denied[i].ContainsIP(ip) && a.Denied[i].PortInRange(port) {
```
dobrac: same here

packages/shared/pkg/sandbox-network/rule.go
```diff
@@ -0,0 +1,133 @@
...
+	return a != nil && (len(a.Allowed) > 0 || len(a.Denied) > 0)
+}
+
+// SplitHostPort splits a network rule string into host and port parts.
+// Uses net.SplitHostPort for bracket/IPv6 handling, with fallback for bare hosts.
+// Returns empty port string when no port is specified.
+func SplitHostPort(s string) (host, port string, err error) {
+	h, p, splitErr := net.SplitHostPort(s)
+	if splitErr != nil {
+		// Strip brackets for bare "[::1]" (no port).
```
dobrac: why supporting [ ] brackets?

packages/shared/pkg/sandbox-network/rule.go
```diff
@@ -0,0 +1,133 @@
...
+	return true
+}
+
+// HasRules returns true if any rules are configured.
+func (a *ACL) HasRules() bool {
+	return a != nil && (len(a.Allowed) > 0 || len(a.Denied) > 0)
+}
+
+// SplitHostPort splits a network rule string into host and port parts.
+// Uses net.SplitHostPort for bracket/IPv6 handling, with fallback for bare hosts.
```
dobrac: ipv6 support is not present and can be skipped

packages/shared/pkg/sandbox-network/access_control.go
```diff
@@ -0,0 +1,133 @@
...
+		// Strip brackets for bare "[::1]" (no port).
+		if strings.HasPrefix(s, "[") && strings.HasSuffix(s, "]") {
+			return s[1 : len(s)-1], "", nil
+		}
+
+		return s, "", nil
+	}
+
+	if h == "" {
+		h = "0.0.0.0/0" // ":443" means all IPs, port 443
```
dobrac: lets use the const variable ALL_TRAFFIC

tests/integration/internal/tests/api/sandboxes/sandbox_network_update_test.go
```diff
@@ -302,37 +436,267 @@ func TestUpdateNetworkConfig(t *testing.T) { //nolint:tparallel // subtests are
...
+		require.NotEqual(t, http.StatusForbidden, request(testPort+1, ""))
+	})
+
+	// =====================================================================
+	// Pause/resume preserves egress, ingress, and mask (must be last)
+	// =====================================================================
+
+	t.Run("pause_resume_preserves_all", func(t *testing.T) { //nolint:paralleltest // sequential
+		denyInPortV4 := fmt.Sprintf("0.0.0.0/0:%d", testPort)
+		denyInPortV6 := fmt.Sprintf("[::/0]:%d", testPort)
```
dobrac: no need to test for ipv6, applies to all added tests

packages/shared/pkg/sandbox-network/rule_test.go
```diff
@@ -0,0 +1,157 @@
...
+		{"IP with port range", "8.8.8.8:80-443", "8.8.8.8", "80-443"},
+		{"trailing colon", "8.8.8.8:", "8.8.8.8", ""},
+		{"port only", ":443", "0.0.0.0/0", "443"},
+		{"bare colon", ":", "0.0.0.0/0", ""},
+		{"bare IPv6 CIDR", "::/0", "::/0", ""},
+		{"IPv6 loopback", "::1", "::1", ""},
+		{"bracketed IPv6 with port", "[::1]:80", "::1", "80"},
+		{"bracketed IPv6 no port", "[::1]", "::1", ""},
+		{"bracketed IPv6 CIDR with port", "[::/0]:443", "::/0", "443"},
+		{"bracketed IPv6 with range", "[::1]:80-443", "::1", "80-443"},
```
dobrac: we can skip

packages/shared/pkg/proxy/proxy.go
```diff
@@ -107,3 +108,37 @@ func (p *Proxy) ListenAndServe(ctx context.Context) error {
 func (p *Proxy) Serve(l net.Listener) error {
 	return p.Server.Serve(tracking.NewListener(l, &p.currentServerConnsCounter))
 }
+
+// ClientIPHeader is an internal header set by client-proxy to forward
+// the real client IP to the orchestrator proxy. It is always stripped
+// before the request is forwarded to the sandbox.
+const ClientIPHeader = "X-E2B-Client-IP"
```
dobrac:
```suggestion
const E2BClientIPHeader = "X-E2B-Client-IP"
```

packages/shared/pkg/proxy/proxy.go
```diff
@@ -107,3 +108,37 @@ func (p *Proxy) ListenAndServe(ctx context.Context) error {
...
+//
+//	X-Forwarded-For: <client-supplied>, <real-client-IP>, <GCP-LB-IP>
+//
+// We take the second-to-last entry, which is the real client IP added by
+// the load balancer. Taking the first entry would be spoofable.
+
+func ExtractClientIP(r *http.Request) string {
+	if clientIP := r.Header.Get(ClientIPHeader); clientIP != "" {
+		return clientIP
+	}
```
dobrac: should we move this to a separate function to see improper configurations early? (e.g. when this header is missing, we should not maybe evaluate the firewall)

dobrac (changes requested):

packages/api/internal/handlers/sandbox_create.go
levb: We do for ingress filtering, we have to I think - it's coming in the header from GCP LB.

packages/api/internal/handlers/sandbox_create.go
levb: too many nuances - ports, domains, and why double-scan? I simplified the error handling so these 2 read better

packages/api/internal/handlers/sandbox_create.go
levb: But why, if we have to scan both allow and deny anyway, again?

packages/api/internal/handlers/sandbox_create_test.go
levb: it got dropped from the tests, should be supported

packages/orchestrator/internal/proxy/proxy.go
levb: No, this is our header that no one else is setting; we just use it to get the ClientIP to the sidecar proxy, since XFF was getting stripped out

packages/api/internal/orchestrator/create_instance.go
levb: 1/5 keep it; `egressConfigDbToOrchestrator` is worse IMO. Please propose a better name if you have one in mind.

packages/api/internal/orchestrator/create_instance.go
levb: `parseIngressRules` is part of the `buildNetworkConfig` (and `buildEgressConfig`) family of helpers; none return errors - all rely on prevalidation. I am not sure what is different here?

I think there's a historical separation (handlers/orch interface) in place here, I'd argue parsing _is_ validation, and we could drop the `validateIngress/Egress` functions altogether and use error-returning versions of `build...Config` - but that is a bigger change.

dobrac: the idea is:<br><br>API: validates the input from the user, returns nice user messages and exits early<br>API - orchestrator: This one is a bit more complicated. I still think the conversion should exit with an error. All of the code paths work correctly now, but bugs can be introduced easily and having an error there will surface them faster rather then silent ignore.<br><br>Orchestrator: parses the input however it needs for it to work and exits with an error if its invalid. This is handled as orchestrator 500 error. This should not happen - but if it does, we should still handle it and not hide it which could lead into undefined behavior for the user. I think we can do this especially as we're converting the network config to the ACLs

packages/orchestrator/internal/sandbox/sandbox.go
levb: `SandboxNetworkConfig` is a protobuf used to communicate API->orchestrator, `in/egressACL` are compiled values in the orchestrator. Are you suggesting that we compute the ACLs in the API, serialize into protobuf as such, and then deserialize in the orchestrator? 

```
type ACL struct {
	Allowed []Rule
	Denied  []Rule
}

type Rule struct {
	Host      string
	IPNet     *net.IPNet
	PortStart uint16 // 0 means all ports
	PortEnd   uint16 // 0 means all ports
	IsDomain  bool
}
```

ipNet is just a bunch of []byte

packages/api/internal/handlers/sandbox_create.go
dobrac: I would prefer a specific split of domains and addresses over using `IsIPOrCIDR` to do this. The logic can look different if you find better way.<br>There is no need to optimize for the double-scan as of now

packages/orchestrator/internal/sandbox/sandbox.go
dobrac: I mean just instead of saving the `SandboxNetworkConfig` on the config struct, we can do a new struct that will reflect better the remaining values missing from ACLs. There is no need to keep the raw data in the config right?

packages/api/internal/handlers/sandbox_create.go
levb: But that (using `IsIPOrCIDR`) is all that `ParseAddressesAndDomains` does? And to work around ports in ingress, it'd need to actually do the string splitting, and then check. 

The logic lives in 3 different places now - validate... the API inputs, build... them into orch protos, and in orchestrator - cache compiled values (ACLs). But it's quite simple.

Ideally - we have 1 shared ACL parser, takes in the strings, returns ACLs. Validation is (parse, throw away the result). I started that way, but it led to too many flags due to the differences between the use cases, so I inlined instead.

```
func ParseAddressesAndDomains(entries []string) (addresses []string, domains []string) {
	for _, entry := range entries {
		**if IsIPOrCIDR(entry)** {
			addresses = append(addresses, entry)
		} else {
			domains = append(domains, entry)
		}
	}

	return addresses, domains
}
```

packages/orchestrator/internal/sandbox/sandbox.go
levb: transformations: API unJSON-ed -> (db/types) -> proto ??

packages/orchestrator/internal/server/sandboxes.go
levb: [] of 
```
type IngressRule struct {
	state         protoimpl.MessageState
	sizeCache     protoimpl.SizeCache
	unknownFields protoimpl.UnknownFields

	Cidr     string `protobuf:"bytes,1,opt,name=cidr,proto3" json:"cidr,omitempty"`
	PortLow  uint32 `protobuf:"varint,2,opt,name=port_low,json=portLow,proto3" json:"port_low,omitempty"`    // 0 = all ports
	PortHigh uint32 `protobuf:"varint,3,opt,name=port_high,json=portHigh,proto3" json:"port_high,omitempty"` // 0 = all ports
}

```

levb: 2 related issues here:
1) is it ok to send this to clickhouse - I don't know. We use strings elsewhere, but the map is coded as map[string]any, and the contents gets JSON-fied, right?
2) I'd 2/5 prefer that we do not do `IngressRule` proto at all, and use a (pre-validated) `string` of `host:lo-hi` but you asked to reduce the parsing/validation; strings could have been used here directly; otherwise we'd need an `ACL.String()` of sorts if we are to include the ACLs as strings

packages/orchestrator/internal/sandbox/sandbox.go
dobrac: > transformations: API unJSON-ed -> (db/types) -> proto ??

Could you please elaborate what do you mean?

I meant just organization wise moving these parsers like `parseIngressRules` outside of the sandbox.go. The sandbox.go is already really large and has too many responsibilities. I think its fair to not grow it more

packages/orchestrator/internal/server/sandboxes.go
dobrac: I think we can skip (remove) it from the event altogether

packages/api/internal/handlers/sandbox_create.go
dobrac: > But that (using IsIPOrCIDR) is all that ParseAddressesAndDomains does? And to work around ports in ingress, it'd need to actually do the string splitting, and then check.

Yes, that is true, but I think its ok.

Anyway, lets keep it as it is. I think we could be able to merge the validation + parsing logic on the API together once we stop saving the API state on the orchestrators

packages/orchestrator/internal/sandbox/sandbox.go
levb: Done. Nice cleanup. Thx for the suggestion.

packages/orchestrator/internal/server/sandboxes.go
levb: I now set it same as egress (strings), or do you want to remove both updates' eventData?

packages/orchestrator/pkg/tcpfirewall/handlers.go
levb: It now reads like this, no comment should be needed:
```
	if egress.NoFirewallRules() && egress.NoHTTPHostDomainRules() {
		return true, MatchTypeNone
	}
```

packages/api/internal/orchestrator/create_instance.go
levb: In the new simplified version these methods are trivial, inclined to skip error checking here (see network.go)

packages/orchestrator/internal/sandbox/sandbox.go
levb: (duplicate of https://github.com/e2b-dev/infra/pull/2093#discussion_r2960573934, can discuss there)

packages/api/internal/handlers/sandbox_create_test.go
```diff
@@ -105,50 +105,42 @@ func TestValidateNetworkConfig(t *testing.T) {
 			},
 			wantErr:    true,
 			wantCode:   http.StatusBadRequest,
-			wantErrMsg: "invalid denied CIDR not-a-cidr",
+			wantErrMsg: `invalid denied CIDR not-a-cidr`,
 		},
-		// Domain validation tests
```
levb, line 111: Removed because these are already covered by the wildcard tests, redundant

PR #2096, opened 2026-03-11
fix(envd): handle race condition in EnsureDirs for concurrent file uploads

mishushakov:
## Summary
- Fixed race condition in `EnsureDirs` by always calling `Chown` even when `os.IsExist` is returned from `os.Mkdir`
- `Chown` is idempotent, so calling it ensures proper ownership even if another concurrent request created the directory
- Added integration test `TestConcurrentFileUpload` to verify concurrent file uploads to the same nested directory work correctly

## Changes
The `EnsureDirs` function now handles the race condition where multiple concurrent requests try to create the same directory path. When `os.Mkdir` returns `os.IsExist` (indicating another process created the directory between the `os.Stat` check and the `os.Mkdir` call), we still call `os.Chown` to ensure proper ownership. This is safe because `Chown` is idempotent and ensures the directory has the correct ownership regardless of which request created it.

## Test plan
- ✅ Test concurrent file uploads to verify no race conditions occur
- ✅ Verify proper ownership (chown) is applied to all created directories
- ✅ Integration test `TestConcurrentFileUpload` exercises the race condition fix

<div><a href="https://cursor.com/agents/bc-5728fb7d-4424-4695-a638-9ea4798b044d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-5728fb7d-4424-4695-a638-9ea4798b044d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/packages/envd/internal/permissions/path.go
+++ b/packages/envd/internal/permissions/path.go
@@ -75,15 +75,12 @@ func EnsureDirs(path string, uid, gid int) error {
 
 		if err != nil && os.IsNotExist(err) {
 			err = os.Mkdir(subpath, 0o755)
-			if os.IsExist(err) {
-				// Another process created the directory between our Stat and Mkdir calls.
-				continue
-			}
-
-			if err != nil {
+			if err != nil && !os.IsExist(err) {
 				return fmt.Errorf("failed to create directory: %w", err)
 			}
 
+			// Chown even if another process created the directory (os.IsExist),
+			// since it may not have called Chown yet. Chown is idempotent.
 			err = os.Chown(subpath, uid, gid)
 			if err != nil {
 				return fmt.Errorf("failed to chown directory: %w", err)


tests/integration/internal/tests/envd/filesystem_test.go
```diff
@@ -370,3 +373,68 @@ func TestRelativePath(t *testing.T) {
...
+	sbx := utils.SetupSandboxWithCleanup(t, c)
+	envdClient := setup.GetEnvdClient(t, ctx)
+
+	baseDir := "/home/user/concurrent-test/nested/dir"
+
+	// Upload multiple files concurrently to the same nested directory using force mode.
+	// Without force, this would race on directory creation.
+	// Note: we avoid calling require/assert from goroutines since FailNow
+	// calls runtime.Goexit which cannot be caught by recover, causing deadlocks.
+	errs := make(chan error, 4)
```
jakubno: you can use errgroup

packages/envd/internal/permissions/path.go
```diff
@@ -94,3 +94,35 @@ func EnsureDirs(path string, uid, gid int) error {
 
 	return nil
 }
+
+func EnsureDirsForce(path string, uid, gid int) error {
```
jakubno: is the force paramater needed? Imho it's race condition we should be handling anyway
```go
func EnsureDirs(path string, uid, gid int) error {
	subpaths := getSubpaths(path)
	for _, subpath := range subpaths {
		info, err := os.Stat(subpath)
		if err != nil && !os.IsNotExist(err) {
			return fmt.Errorf("failed to stat directory: %w", err)
		}

		if err != nil && os.IsNotExist(err) {
			err = os.Mkdir(subpath, 0o755)
			if os.IsExist(err) {
				// This can happen due to race condition with another process which creates the directory after we check for existence
				// but before we try to create it.In that case, we can just ignore the error and move on.
				continue
			}
			if err != nil {
				return fmt.Errorf("failed to create directory: %w", err)
			}

			err = os.Chown(subpath, uid, gid)
			if err != nil {
				return fmt.Errorf("failed to chown directory: %w", err)
			}

			continue
		}

		if !info.IsDir() {
			return fmt.Errorf("path is a file: %s", subpath)
		}
	}

	return nil
}

```

PR #2097, opened 2026-03-11
feat(orchestrator): add local filesystem upload endpoint for build cache storage

arkamar:
Adds a local HTTP upload endpoint to the orchestrator so that template builds
using `STORAGE_PROVIDER=Local` can upload artifacts via HMAC-signed URLs,
mirroring the signed-URL flow used with GCS/S3.

## Changes

- **Local upload handler** (`packages/orchestrator/internal/localupload/`) — new
  `PUT /upload` endpoint gated by HMAC-signed tokens. 
- **HMAC signing** (`packages/shared/pkg/storage/storage_fs.go`) —
  `ComputeUploadHMAC` and `ValidateUploadToken` generate and verify per-request
  tokens. The HMAC key is generated with `crypto/rand` at process startup.
  `UploadSignedURL` produces complete signed URLs for clients.
- **Storage refactor** — added `GetProviderType()` and `IsLocal()` helpers to
  `StorageProvider`. The upload handler is only registered when the provider is
  local. `buildPersistence` is now injected into `tmplserver.New` instead of
  being created internally.
- **Integration test** — new COPY step integration test for template builds
  (`tests/api/templates/build_template_test.go`).
- **Unit tests** — comprehensive test coverage for the handler (method
  validation, missing params, invalid/expired tokens, path traversal, overwrites,
  nested dirs, empty bodies, wrong-path tokens) and HMAC functions (determinism,
  round-trip through URL encoding, special characters).

---

> [!NOTE]
> **Medium Risk**
> Adds a new HTTP `PUT /upload` surface and changes how local filesystem storage generates upload URLs; mistakes could allow unintended file writes or break template build COPY uploads. Scope is limited to `STORAGE_PROVIDER=Local` and includes token + path traversal checks and tests.
> 
> **Overview**
> Adds support for *signed uploads when using local filesystem storage* by generating HMAC-backed upload URLs in the local storage provider and exposing a `PUT /upload` handler in the orchestrator (registered only when `STORAGE_PROVIDER=Local`). This refactors template-manager initialization to receive `buildPersistence` from the orchestrator, wires in `LOCAL_UPLOAD_BASE_URL` configuration, and extends test coverage with unit tests for token validation/path safety plus an integration test that exercises a template build `COPY` step via the signed upload flow.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 831fff51588c68965888af7d8f3359bff6e4d6cf. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/orchestrator/internal/localupload/handler.go
+++ b/packages/orchestrator/internal/localupload/handler.go
@@ -86,12 +86,10 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	tmpPath := tmpFile.Name()
-
-	// Clean up the temp file on any failure path.
+	stored := false
 	defer func() {
-		if tmpPath != "" {
-			os.Remove(tmpPath)
+		if !stored {
+			os.Remove(tmpFile.Name())
 		}
 	}()
 
@@ -108,20 +106,19 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if err := os.Chmod(tmpPath, 0o644); err != nil {
+	if err := os.Chmod(tmpFile.Name(), 0o644); err != nil {
 		http.Error(w, "failed to set file permissions", http.StatusInternalServerError)
 
 		return
 	}
 
-	if err := os.Rename(tmpPath, fullPath); err != nil {
+	if err := os.Rename(tmpFile.Name(), fullPath); err != nil {
 		http.Error(w, "failed to store file", http.StatusInternalServerError)
 
 		return
 	}
 
-	// Rename succeeded — prevent deferred cleanup from removing the final file.
-	tmpPath = ""
+	stored = true
 
 	w.WriteHeader(http.StatusOK)
 }


packages/orchestrator/main.go
```diff
@@ -745,6 +762,39 @@ func (e serviceDoneError) Error() string {
...
+// for COPY instructions. The returned handler is nil for non-local providers.
+func setupBuildStorage(ctx context.Context, limiter *limit.Limiter, grpcPort uint16) (storage.StorageProvider, *localupload.Handler, error) {
+	cfg := storage.BuildCacheStorageConfig.WithLimiter(limiter)
+
+	var uploadHandler *localupload.Handler
+
+	if storage.IsLocal() {
+		hmacKey := make([]byte, 32)
+		if _, err := rand.Read(hmacKey); err != nil {
+			return nil, nil, fmt.Errorf("generate HMAC key: %w", err)
```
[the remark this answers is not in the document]
arkamar: I think it does not matter that much, this feature is mainly for testing purposes.

packages/orchestrator/internal/localupload/handler.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	if err := os.Rename(tmpPath, fullPath); err != nil {
+		http.Error(w, "failed to store file", http.StatusInternalServerError)
+
+		return
+	}
+
+	// Rename succeeded — prevent deferred cleanup from removing the final file.
+	tmpPath = ""
```
jakubno: isn't the file already at `fullPath`?

packages/orchestrator/main.go
```diff
@@ -745,6 +762,39 @@ func (e serviceDoneError) Error() string {
 	return fmt.Sprintf("service %s finished", e.name)
 }
 
+// setupBuildStorage creates the build cache storage provider and, when using
+// local filesystem storage, the HTTP upload handler that accepts file uploads
+// for COPY instructions. The returned handler is nil for non-local providers.
+func setupBuildStorage(ctx context.Context, limiter *limit.Limiter, grpcPort uint16) (storage.StorageProvider, *localupload.Handler, error) {
```
jakubno: would it make sense to have this in storage package rather than `main.go`?

packages/orchestrator/main.go
```diff
@@ -745,6 +762,39 @@ func (e serviceDoneError) Error() string {
...
+
+	var uploadHandler *localupload.Handler
+
+	if storage.IsLocal() {
+		hmacKey := make([]byte, 32)
+		if _, err := rand.Read(hmacKey); err != nil {
+			return nil, nil, fmt.Errorf("generate HMAC key: %w", err)
+		}
+
+		uploadBaseURL := env.GetEnv("LOCAL_UPLOAD_BASE_URL", fmt.Sprintf("http://localhost:%d", grpcPort))
```
jakubno: can we move this to be part of the config?

packages/shared/pkg/storage/storage.go
```diff
@@ -131,10 +155,10 @@ var BuildCacheStorageConfig = StorageConfig{
 }
 
 func GetStorageProvider(ctx context.Context, cfg StorageConfig) (StorageProvider, error) {
-	provider := Provider(env.GetEnv(storageProviderEnv, string(DefaultStorageProvider)))
+	provider := GetProviderType()
 
 	if provider == LocalStorageProvider {
-		return newFileSystemStorage(cfg.GetLocalBasePath()), nil
+		return newFileSystemStorage(cfg.GetLocalBasePath(), cfg.UploadBaseURL, cfg.HMACKey), nil
```
jakubno: what about passing whole config here?

packages/shared/pkg/storage/storage.go
```diff
@@ -103,6 +115,8 @@ type StorageConfig struct {
 	GetLocalBasePath func() string
 	GetBucketName    func() string
 	limiter          *limit.Limiter
+	UploadBaseURL    string
+	HMACKey          []byte
```
jakubno: these can be probably private

packages/orchestrator/internal/localupload/handler.go
arkamar: Yes, it is. This prevents deferred `os.Remove()` call on non-existent file. However, the pattern wasn't well readable, therefore, I changed it in commit 831fff515. `stored` should be more explanatory.

packages/orchestrator/main.go
arkamar: I don't think so. `storage` package is from `shared/` but this function also uses `localupload.Handler` from `orchestrator/` and it mixes all of that together. It is a local helper. I think we would end up with small 30-line function split into two smaller functions across two packages for no real gain if we will try to move it.

Its code was originally directly in `run()` but I moved it to local helper later. I think it is good as it is now.

PR #2099, opened 2026-03-11
feat(api): use Redis PubSub for state transition

jakubno:
Improves the speed for transitions, prevents spamming redis with repeated polls, reduced the risk of blocking all connections by polling state change


packages/api/internal/sandbox/storage/redis/utils.go
```diff
@@ -8,11 +8,19 @@ import (
 
 const (
 	sandboxKeyPrefix    = "sandbox:storage"
-	transitionKeyPrefix = "transition:"
+	transitionKeyPrefix = "transition"
```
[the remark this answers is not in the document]
jakubno: It isn't used in prod yet, so it's fine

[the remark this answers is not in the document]
jakubno: Not used in production yet

packages/api/internal/sandbox/storage/redis/subscription_manager.go
```diff
@@ -0,0 +1,97 @@
...
+// fans out transition-complete signals to registered in-process waiters.
+type subscriptionManager struct {
+	mu      sync.RWMutex
+	waiters map[string]map[chan struct{}]struct{} // routingKey → registered waiters
+
+	ps     *redis.PubSub
+	cancel context.CancelFunc
+}
+
+func newSubscriptionManager(ctx context.Context, redisClient redis.UniversalClient) *subscriptionManager {
```
djeebus: Instead of passing in a `ctx` and automatically subscribing, it's probably more comfortable to expose a `Start` and `Stop` functions. That makes closing these structs easier and safer.

packages/api/internal/sandbox/storage/redis/subscription_manager_test.go
```diff
@@ -0,0 +1,268 @@
...
+	return storage.subManager
+}
+
+func TestSubscriptionManager_SubscribeAndDispatch(t *testing.T) {
+	t.Parallel()
+
+	m := setupTestManager(t)
+
+	ch, cleanup := m.subscribe("key1")
+	defer cleanup()
```
djeebus: Use `t.Cleanup(cleanup)` instead, to make sure it gets called in case of panics. Similar in the rest of the file.

packages/api/internal/sandbox/storage/redis/subscription_manager.go
```diff
@@ -0,0 +1,97 @@
...
+	ch := m.ps.Channel()
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case msg, ok := <-ch:
+			if !ok {
+				return
+			}
+			m.dispatch(msg.Payload)
```
djeebus: maybe make this a goroutine?

packages/api/internal/sandbox/storage/redis/utils.go
```diff
@@ -8,11 +8,19 @@ import (
...
 	sandboxesKey        = "sandboxes"
 	indexKey            = "index"
 )
 
+// getGlobalTransitionNotifyChannel is the single Redis PubSub channel used by
+// all sandboxes across all teams. The per-sandbox routing key is embedded in
+// the message payload so one connection per API pod is sufficient.
+func getGlobalTransitionNotifyChannel() string {
+	return redis_utils.CreateKey(sandboxKeyPrefix, transitionKeyPrefix, notifySuffix)
+}
```
djeebus: Seems overly complicated and memory hungry for something that could just be:

```golang
const globalTransitionNotifyChannel = "sandbox:storage:transition:notify"
```

If we wanted it configurable, we could pass it in from `redis.NewStorage` and store it as a field. Either way, we're not encouraging go to allocate memory for a string every time that a message gets published.

djeebus (approved):

packages/api/internal/sandbox/storage/redis/subscription_manager.go
jakubno: - dispatch is extremely cheap — it takes an RLock, iterates a small map, and does non-blocking channel sends.
  - Goroutine-per-message adds allocation overhead that likely exceeds the cost of dispatch itself
  - The channel buffer is size 1, so dispatch is never blocked by a slow consumer

PR #2104, opened 2026-03-11
chore(orch): simplify sandbox map cleanup flow

jakubno:
I just realized it's should be enough to remove it after the sandbox is dead, no need to do any magic there

---

> [!NOTE]
> **Low Risk**
> Small refactor to cleanup ordering and logging around sandbox map eviction; behavior should be equivalent but could affect edge-case removal timing if cleanup runs before insertion.
> 
> **Overview**
> This changes sandbox lifecycle-map cleanup to register `RemoveByLifecycleID` only after a sandbox is inserted (using the sandbox’s own newly-generated `LifecycleID`) for both create and resume flows, removing the earlier precomputed ID/early cleanup hook. It also enriches `RemoveByLifecycleID` removal logs with the sandbox’s host IP to make eviction events easier to trace.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 24d6db84de92029b7b550c2ee927deab2cc0585e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/orchestrator/internal/sandbox/map.go
+++ b/packages/orchestrator/internal/sandbox/map.go
@@ -135,7 +135,7 @@ func (m *Map) RemoveByLifecycleID(ctx context.Context, sandboxID, lifecycleID st
 
 		sbx = v
 
-				return v.LifecycleID == lifecycleID
+		return v.LifecycleID == lifecycleID
 	})
 
 	if removed {


packages/orchestrator/internal/sandbox/map.go
```diff
@@ -132,12 +133,14 @@ func (m *Map) RemoveByLifecycleID(ctx context.Context, sandboxID, lifecycleID st
 			return false
 		}
 
+		sbx = v
 		return v.LifecycleID == lifecycleID
```
jakubno:
```suggestion

				return v.LifecycleID == lifecycleID
```

packages/orchestrator/internal/sandbox/map.go
```diff
@@ -132,12 +133,15 @@ func (m *Map) RemoveByLifecycleID(ctx context.Context, sandboxID, lifecycleID st
 			return false
 		}
 
-		return v.LifecycleID == lifecycleID
+		sbx = v
+
+				return v.LifecycleID == lifecycleID
```
jakubno:
```suggestion
		return v.LifecycleID == lifecycleID
```

PR #2105, opened 2026-03-12
syscall is deprecated, use golang.org/x/sys instead

djeebus:
While trying to figure out "syscall" vs "golang.org/x/sys/unix", I came across [this deprecation text](https://pkg.go.dev/syscall). Seems like, where at all possible, we should be using the x/sys package.

---

> [!NOTE]
> **Medium Risk**
> Mostly a mechanical swap from `syscall` to `golang.org/x/sys/unix` across low-level process/filesystem/signal code; risk is moderate because subtle constant/type differences could affect shutdown signaling, error classification, or platform-specific behavior at runtime and in tests.
> 
> **Overview**
> Replaces most direct uses of the deprecated `syscall` package with `golang.org/x/sys/unix` across the API, proxy, dashboard API, envd, orchestrator, DB retry logic, and shared utilities (signals, process attributes, mmap/munmap, socket options, file stats, and errno comparisons), and updates module dependencies accordingly. The shared filesystem entry metadata extraction is adjusted to handle both `*syscall.Stat_t` and `*unix.Stat_t` via a new helper, with tests updated/added to validate UID/GID and nil/typed-nil handling.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7638025789ea73b2279204a4a9d4a6e0f32b217c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -188,7 +188,7 @@ fmt:
 
 .PHONY: lint
 lint:
-	go work edit -json | jq -r '.Use[].DiskPath' | xargs -I{} golangci-lint run {}/... --fix
+	go work edit -json | jq -r '.Use[].DiskPath' | xargs -P 4 -I{} golangci-lint run {}/... --fix
 
 .PHONY: generate-mocks
 generate-mocks:


Makefile
```diff
@@ -188,7 +188,7 @@ fmt:
 
 .PHONY: lint
 lint:
-	go work edit -json | jq -r '.Use[].DiskPath' | xargs -P 4 -I{} golangci-lint run {}/... --fix
```
dobrac, line 191: is this change desired?

djeebus: `golangci-lint` will already happily use 100% of your CPU, so I'm not sure what purpose the parallelization serves, other than to overload a computer (which it did a few times for me)?

running `go clean -cache && golangci-lint cache clean && make lint`
- without `-P`: 88 seconds cold, 9 seconds warm
- with `-P 2`: 81 seconds, 5 seconds warm
- with `-P 3`: 87 seconds, 4 seconds warm
- with `-P 4`: 90 seconds, 3.5 seconds warm

djeebus: It's not the point of this PR though, I'll revert it

PR #2110, opened 2026-03-12
fix: AWS self-hosting setup issues

ya-luotao:
## Summary

- **Fix PREFIX in env templates**: Remove quotes and inline comments from `PREFIX` in `.env.aws.template` and `.env.gcp.template`. Make includes env files literally, so `PREFIX="e2b-" # comment` results in the value `"e2b-" ` (with literal quotes and trailing space), which breaks `BUCKET_PREFIX` concatenation — the shell sees `528893196824-` as a command instead of part of the variable value.
- **Fix S3 bucket creation for us-east-1**: AWS rejects `--create-bucket-configuration LocationConstraint=us-east-1` since us-east-1 is the default region. The Makefile now conditionally omits this flag for us-east-1.
- **Fix Packer AMI build path in self-host.md**: The directory is `iac/provider-aws/nomad-cluster-disk-image`, not the non-existent `iac/provider-aws/packer`. Also clarified that it builds a single shared AMI (not per-node-type AMIs) and added the actual `make init` + `make build` commands.

## Test plan

- [ ] Verify `make init` works in us-east-1 with fresh S3 bucket creation
- [ ] Verify `make init` still works in non-us-east-1 regions
- [ ] Verify Packer AMI build works from the documented path

### DIFF
--- a/self-host.md
+++ b/self-host.md
@@ -147,28 +147,7 @@ Now, you should see the right quota options in `All Quotas` and be able to reque
     make build  # build the AMI (~5 min, launches a t3.large)
     ```
 7. Run `make build-and-upload` to build and push container images and binaries
-8. Copy Firecracker kernels and rootfs to your S3 buckets. You have two options:
-
-    **Option A** — Using `make` (requires [`gsutil`](https://cloud.google.com/storage/docs/gsutil_install)):
-    ```sh
-    make copy-public-builds
-    ```
-
-    **Option B** — Without `gsutil` (uses `aws` CLI with GCS S3-compatible endpoint):
-    ```sh
-    # Set your bucket prefix (PREFIX + AWS_ACCOUNT_ID + "-")
-    BUCKET_PREFIX="e2b-YOUR_ACCOUNT_ID-"
-
-    # Download from public GCS bucket via S3-compatible API
-    mkdir -p ./.kernels ./.firecrackers
-    aws s3 cp s3://e2b-prod-public-builds/kernels/ ./.kernels/ --recursive --no-sign-request --endpoint-url https://storage.googleapis.com
-    aws s3 cp s3://e2b-prod-public-builds/firecrackers/ ./.firecrackers/ --recursive --no-sign-request --endpoint-url https://storage.googleapis.com
-
-    # Upload to your S3 buckets
-    aws s3 cp ./.kernels/ s3://${BUCKET_PREFIX}fc-kernels/ --recursive --profile ${AWS_PROFILE}
-    aws s3 cp ./.firecrackers/ s3://${BUCKET_PREFIX}fc-versions/ --recursive --profile ${AWS_PROFILE}
-    rm -rf ./.kernels ./.firecrackers
-    ```
+8. Run `make copy-public-builds` to copy Firecracker kernels and rootfs to your S3 buckets
 9. Run `make plan-without-jobs` and then `make apply` to provision the cluster infrastructure
 10. Run `make plan` and then `make apply` to deploy all Nomad jobs (this also runs database migrations automatically via the API's db-migrator task)
 11. Setup data in the cluster by running `make prep-cluster` in `packages/shared` to create an initial user, team, and build a base template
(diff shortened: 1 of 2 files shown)


self-host.md
```diff
@@ -140,17 +145,49 @@ Now, you should see the right quota options in `All Quotas` and be able to reque
...
+    rm -rf .kernels .firecrackers
+    ```
 9. Run `make plan-without-jobs` and then `make apply` to provision the cluster infrastructure
 10. Run `make plan` and then `make apply` to deploy all Nomad jobs
-11. Setup data in the cluster by running `make prep-cluster` in `packages/shared` to create an initial user, team, and build a base template
+11. Run database migrations: `cd packages/db && make migrate`
+    > If using Supabase, the first migration (`20000101000000_auth.sql`) will fail because Supabase already provides the `auth` schema, `authenticated` role, and `auth.users` table. To skip it, manually mark it as applied before running `make migrate`:
+    > ```sh
+    > psql "$POSTGRES_CONNECTION_STRING" -c "INSERT INTO _migrations (version_id, is_applied) VALUES (20000101000000, true);"
+    > ```
```
sitole: This is not needed, api deployed as part of `make plan && make apply` will execute migrations.

self-host.md
```diff
@@ -140,17 +145,49 @@ Now, you should see the right quota options in `All Quotas` and be able to reque
...
+    for name in $(curl -s "https://storage.googleapis.com/storage/v1/b/e2b-prod-public-builds/o?prefix=firecrackers/" | python3 -c "import sys,json; [print(i['name']) for i in json.load(sys.stdin).get('items',[]) if int(i.get('size',0))>0]"); do
+      mkdir -p ".firecrackers/$(dirname "${name#firecrackers/}")"
+      curl -sL "https://storage.googleapis.com/e2b-prod-public-builds/$name" -o ".firecrackers/${name#firecrackers/}"
+    done
+
+    # Upload to S3
+    aws s3 cp .kernels/ "s3://${BUCKET_PREFIX}fc-kernels/" --recursive --profile default
+    aws s3 cp .firecrackers/ "s3://${BUCKET_PREFIX}fc-versions/" --recursive --profile default
+    rm -rf .kernels .firecrackers
+    ```
```
sitole: We can maybe replace `copy-public-builds` aws implementation to something like this instead:

```
mkdir -p ./.kernels
mkdir -p ./.firecrackers
aws s3 cp s3://e2b-prod-public-builds/kernels/ ./.kernels/ --recursive --no-sign-request --endpoint-url https://storage.googleapis.com
aws s3 cp s3://e2b-prod-public-builds/firecrackers/ ./.firecrackers/ --recursive --no-sign-request --endpoint-url https://storage.googleapis.com
aws s3 cp ./.kernels/ s3://${AWS_BUCKET_PREFIX}fc-kernels/ --recursive --profile ${AWS_PROFILE}
aws s3 cp ./.firecrackers/ s3://${AWS_BUCKET_PREFIX}fc-versions/ --recursive --profile ${AWS_PROFILE}
rm -rf ./.kernels
rm -rf ./.firecrackers
```

self-host.md
```diff
@@ -159,35 +154,24 @@ Now, you should see the right quota options in `All Quotas` and be able to reque
...
-    rm -rf .kernels .firecrackers
+    # Download from public GCS bucket via S3-compatible API
+    mkdir -p ./.kernels ./.firecrackers
+    aws s3 cp s3://e2b-prod-public-builds/kernels/ ./.kernels/ --recursive --no-sign-request --endpoint-url https://storage.googleapis.com
+    aws s3 cp s3://e2b-prod-public-builds/firecrackers/ ./.firecrackers/ --recursive --no-sign-request --endpoint-url https://storage.googleapis.com
+
+    # Upload to your S3 buckets
+    aws s3 cp ./.kernels/ s3://${BUCKET_PREFIX}fc-kernels/ --recursive --profile ${AWS_PROFILE}
+    aws s3 cp ./.firecrackers/ s3://${BUCKET_PREFIX}fc-versions/ --recursive --profile ${AWS_PROFILE}
+    rm -rf ./.kernels ./.firecrackers
```
sitole: Please remove all this and just replace `make copy-public-builds` command for aws to only use aws cli.

PR #2115, opened 2026-03-12
fix(aws): derive AMI name prefix from var.prefix

ya-luotao:
## Summary

- **Bug:** Packer names AMIs as `${prefix}orch-<timestamp>`, but all Terraform node pool modules hardcode `image_family_prefix = "e2b-orch-"`. Any custom PREFIX (e.g. `e2bdev-`) causes `terraform plan` to fail with "Your query returned no results" when looking up AMIs.
- **Fix:** Introduce `local.ami_family_prefix = "${var.prefix}orch-"` and use it for all 5 node pools (control_server, api, client, clickhouse, build). Remove the 4 now-unused `*_image_family_prefix` variables from `variables.tf`.
- **Bonus fix:** `build_image_family_prefix` was never passed from `main.tf` to the cluster module — it silently fell back to the submodule's hardcoded default. Now explicitly wired.

## Test plan

- [ ] With PREFIX=e2b- (default): `terraform plan` still finds AMIs named `e2b-orch-*` — no behavior change
- [ ] With PREFIX=e2bdev- (custom): `terraform plan` now correctly looks for `e2bdev-orch-*` AMIs

### DIFF
--- a/iac/provider-aws/main.tf
+++ b/iac/provider-aws/main.tf
@@ -108,7 +108,7 @@ module "cluster" {
   consul_gossip_encryption_key    = module.init.cluster.consul_gossip_encryption_key
 
   control_server_cluster_size        = var.control_server_cluster_size
-  control_server_image_family_prefix = local.ami_family_prefix
+  control_server_image_family_prefix = var.control_server_image_family_prefix != "" ? var.control_server_image_family_prefix : local.ami_family_prefix
   control_server_machine_type        = var.control_server_machine_type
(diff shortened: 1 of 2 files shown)


sitole (commented):
Please make it back compatible by changing value TF vars to empty string and then locally evaluate  `clickhouse_image_family_prefix = var.clickhouse_image_family_prefix == "" ?  locals.ami_family_prefix : var.clickhouse_image_family_prefix` we still need to be able to override ami for each node pool separately.

sitole (approved):

PR #2116, opened 2026-03-12
fix(orch): keep the sandbox in the map even after it starts shutting down to handle ip reverse lookups

jakubno:
Fix an issue with `error finding sandbox for source addr` when e.g. logs comes when the sandbox is shutting down

### DIFF
--- a/packages/orchestrator/pkg/nfsproxy/chroot/nfs.go
+++ b/packages/orchestrator/pkg/nfsproxy/chroot/nfs.go
@@ -137,10 +137,6 @@ func (h *NFSHandler) getChroot(ctx context.Context, remoteAddr net.Addr, request
 		return nil, fmt.Errorf("%w: %w", ErrUnknownSandbox, err)
 	}
 
-	if !sbx.IsRunning() {
-		return nil, fmt.Errorf("%w: sandbox is not running", ErrUnknownSandbox)
-	}
-
 	// normalize the mount path
 	requestedPath := string(request.Dirpath)
 	regexpMatch := mountPath.MatchString(requestedPath)


packages/orchestrator/pkg/sandbox/map.go
```diff
@@ -95,19 +101,74 @@ func (m *Map) Insert(ctx context.Context, sbx *Sandbox) {
...
 	go m.trigger(func(s MapSubscriber) {
 		s.OnInsert(sbx)
 	})
 }
 
+// MarkDead transitions a sandbox to the dead state. OnRemove subscribers are
+// notified immediately (so the proxy / firewall limiter can clean up), but the
+// entry stays in the map for deadEvictionGracePeriod so that IP-based lookups
+// still resolve while the Firecracker process finishes shutting down.
+func (m *Map) MarkDead(ctx context.Context, sandboxID string) {
```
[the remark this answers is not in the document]
jakubno: already fixed

packages/orchestrator/pkg/server/sandboxes.go
```diff
@@ -645,7 +644,7 @@ func (s *Server) Checkpoint(ctx context.Context, in *orchestrator.SandboxCheckpo
 		if err := res.snapshot.Upload(uploadCtx, s.persistence, res.templateFiles); err != nil {
 			telemetry.ReportCriticalError(ctx, "error uploading snapshot for checkpoint", err, telemetry.WithSandboxID(in.GetSandboxId()))
 
-			s.sandboxFactory.Sandboxes.Remove(ctx, resumedSbx.Runtime.SandboxID)
+			s.sandboxFactory.Sandboxes.MarkDead(ctx, resumedSbx.Runtime.SandboxID)
```
dobrac: do we need this call here at all? I would assume the `stopSandboxAsync` does the same

packages/orchestrator/pkg/sandbox/sandbox.go
```diff
@@ -135,6 +135,21 @@ type EnvdMetadata struct {
...
+const (
+	// StatusStarting is the initial state: the sandbox is being created but
+	// is not yet ready to serve traffic.
+	StatusStarting SandboxStatus = 0
+	// StatusRunning means the sandbox is fully initialised and healthy.
+	StatusRunning SandboxStatus = 1
+	// StatusDead means the sandbox has been killed or paused. It remains in
+	// the map briefly so that IP-based lookups (logs, NFS, TCP firewall)
+	// still resolve, but it is excluded from all "live" queries.
+	StatusDead SandboxStatus = 2
```
dobrac: More accurate would be probably `stopping`, `removing`, or `shutting_down` as the sandbox is technically still running

packages/orchestrator/pkg/sandbox/map.go
```diff
@@ -150,9 +204,13 @@ func (m *Map) RemoveByLifecycleID(ctx context.Context, sandboxID, lifecycleID st
...
 
-		go m.trigger(ctx, func(ctx context.Context, s MapSubscriber) {
-			s.OnRemove(ctx, sbx)
-		})
+		// Only fire OnRemove if it wasn't already notified subscribers.
+		// MarkDead sets StatusDead and fires OnRemove
+		if !alreadyDead {
+			go m.trigger(ctx, func(ctx context.Context, s MapSubscriber) {
+				s.OnRemove(ctx, sbx)
+			})
```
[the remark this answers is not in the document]
dobrac: this is valid

packages/orchestrator/pkg/server/sandboxes.go
jakubno: it looks like it doesn't, it calls `Stop` and `Stop` doesn't call clean up, only `Shutdown` or `Close` does :hidehepain:

PR #2120, opened 2026-03-12
chore(api): add auth attributes

jakubno:
image

---

> [!NOTE]
> **Medium Risk**
> Touches authentication validation call paths and logging/telemetry context propagation; mistakes could break auth middleware wiring or leak/misattribute request metadata.
> 
> **Overview**
> This PR propagates the `gin.Context` into auth validation calls so the service can attach **masked API keys/access tokens**, `user.id`, and `team.id` as OpenTelemetry attributes on the parent request span. It also standardizes proxy request logging by introducing `logger.ProxyRequestFields` (including client IP extraction) and adds helpers for masking tokens (`keys.MaskToken`) plus corresponding logger/telemetry field helpers, with a small generic constraint update to require teams expose `TeamID()`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 57e6e27c575f9d228534ef5791ba4dedb8e769ba. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/auth/pkg/auth/service.go
+++ b/packages/auth/pkg/auth/service.go
@@ -86,7 +86,7 @@ func (s *AuthService[T]) ValidateAPIKey(ctx context.Context, ginCtx *gin.Context
 
 	//nolint:contextcheck // We use the gin request context to set attributes on the parent span.
 	telemetry.SetAttributes(ginCtx.Request.Context(),
-		telemetry.WithMaskedAPIKey(apiKey),
+		telemetry.WithMaskedAPIKey(keys.MaskToken(keys.ApiKeyPrefix, apiKey)),
 		telemetry.WithTeamID(result.TeamID()),
 	)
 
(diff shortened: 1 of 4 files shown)


packages/auth/pkg/auth/service.go
```diff
@@ -79,11 +84,17 @@ func (s *AuthService[T]) ValidateAPIKey(ctx context.Context, apiKey string) (T,
 		}
 	}
 
+	//nolint:contextcheck // We use the gin request context to set attributes on the parent span.
+	telemetry.SetAttributes(ginCtx.Request.Context(),
+		telemetry.WithAPIKey(apiKey),
```
sitole: Let's directly call `telemetry.WithMaskedAPIKey` to make it extra safe.

sitole: Is there a case where you can then easily search for the API key? Does it make sense to also log api key ID?

sitole (changes requested):

packages/auth/pkg/auth/service.go
```diff
@@ -79,11 +84,17 @@ func (s *AuthService[T]) ValidateAPIKey(ctx context.Context, apiKey string) (T,
 		}
 	}
 
+	//nolint:contextcheck // We use the gin request context to set attributes on the parent span.
+	telemetry.SetAttributes(ginCtx.Request.Context(),
+		telemetry.WithMaskedAPIKey(apiKey),
```
sitole: My point with change from `telemetry.WithAPIKey` to `telemetry.WithMaskedAPIKey` to pass already masked value and rename function in way that its ovious from name already. You probably don't want to send a raw api key/auth token to a logging library; it's a different scope that should not ever receive any sensitive values in the first place.

PR #2122, opened 2026-03-12
chore(api): add redis sync with orch

jakubno:
In case of some some random error (DB, node time out), the sandbox is removed from sandbox store, but it stays running on the node forever, this will make sure it gets properly cleaned up.

### DIFF
--- a/packages/api/internal/sandbox/store.go
+++ b/packages/api/internal/sandbox/store.go
@@ -174,9 +174,9 @@ func (s *Store) Reconcile(ctx context.Context, sandboxes []Sandbox, nodeID strin
 		wg := sync.WaitGroup{}
 		for _, sbx := range sbxsToBeSynced {
 			wg.Go(func() {
-				ctx, cancel := context.WithTimeout(ctx, sbxRemoveTimeout)
+				ctx, cancel := context.WithTimeout(context.WithoutCancel(ctx), sbxRemoveTimeout)
 				defer cancel()
-				s.callbacks.RemoveSandboxFromNode(context.WithoutCancel(ctx), sbx)
+				s.callbacks.RemoveSandboxFromNode(ctx, sbx)
 			})
 		}
 


packages/api/internal/sandbox/storage/redis/main.go
```diff
@@ -54,6 +61,84 @@ func (s *Storage) Close() {
 }
 
 // Sync is here only for legacy reasons, redis backend doesn't need any sync
```
sitole: Outdated comment

packages/api/internal/sandbox/storage/redis/main.go
```diff
@@ -54,6 +61,84 @@ func (s *Storage) Close() {
...
+		keys := make([]string, len(candidates))
+		for i, c := range candidates {
+			keys[i] = c.key
+		}
+
+		cmd := pipe.MGet(ctx, keys...)
+		batches = append(batches, batchInfo{cmd: cmd, candidates: candidates})
+	}
+
+	_, err := pipe.Exec(ctx)
```
sitole, line 113: Will the pipeline work correctly if you are pushing different team shards into a single exec?

packages/api/internal/sandbox/storage/redis/main.go
```diff
@@ -54,6 +61,84 @@ func (s *Storage) Close() {
 }
 
 // Sync is here only for legacy reasons, redis backend doesn't need any sync
-func (s *Storage) Sync(_ []sandbox.Sandbox, _ string) []sandbox.Sandbox {
-	return nil
+func (s *Storage) Sync(ctx context.Context, sbxs []sandbox.Sandbox, nodeID string) []sandbox.Sandbox {
```
sitole: If the method returns a list of orphans per node, shouldn't we rename the function? From `sync`, it's not clear what it should return. If not, at least add a comment and move the logic for orphan filtering to a separate private method.

packages/api/internal/sandbox/store.go
```diff
@@ -156,11 +162,22 @@ func (s *Store) WaitForStateChange(ctx context.Context, teamID uuid.UUID, sandbo
...
-		err := s.Add(ctx, sbx, false)
-		if err != nil {
-			logger.L().Error(ctx, "Failed to re-add sandbox during sync", zap.Error(err), logger.WithSandboxID(sbx.SandboxID))
+	sbxsToBeSynced := s.storage.Sync(ctx, sandboxes, nodeID)
+
+	if s.storage.Name() == StorageNameRedis {
+		// Redis is the source of truth — divergent sandboxes are orphans running
+		// on the node but not present in the store. Kill them.
+		for _, sbx := range sbxsToBeSynced {
+			s.callbacks.RemoveSandboxFromNode(ctx, sbx)
```
sitole: Maybe do multiple in parallel?

packages/api/internal/sandbox/store.go
```diff
@@ -156,11 +162,22 @@ func (s *Store) WaitForStateChange(ctx context.Context, teamID uuid.UUID, sandbo
...
 
 func (s *Store) Sync(ctx context.Context, sandboxes []Sandbox, nodeID string) {
-	sbxs := s.storage.Sync(sandboxes, nodeID)
-	for _, sbx := range sbxs {
-		err := s.Add(ctx, sbx, false)
-		if err != nil {
-			logger.L().Error(ctx, "Failed to re-add sandbox during sync", zap.Error(err), logger.WithSandboxID(sbx.SandboxID))
+	sbxsToBeSynced := s.storage.Sync(ctx, sandboxes, nodeID)
+
+	if s.storage.Name() == StorageNameRedis {
```
sitole, line 171: Don't like IFs like this, but hopefully we will be able to remove quickly.

jakubno: Once we migrate to redis completely, it'll be simplified

packages/api/internal/sandbox/storage/redis/main.go
jakubno: Should be fine:
```
yes, the pipeline works correctly across different team shards.

  Here's why:

  1. Each MGET is slot-safe. Keys use hash tags via SameSlot(teamID) → {teamID}, so all keys within a single MGET call hash to the same Redis Cluster slot. See utils.go:39 — GetTeamPrefix wraps the team ID in curly braces.
  2. go-redis v9 cluster pipelines handle cross-slot commands transparently. When pipe.Exec() runs against a cluster client, go-redis internally groups the queued commands by their target node (based on hash slot) and dispatches them in parallel to the correct nodes. Each
   MGET lands on the node that owns that team's slot.

  So the two-level design is correct:
  - Within a team: MGET is safe because hash tags force all keys to one slot
  - Across teams: The pipeline splits commands by node automatically — no manual sharding needed
  ```

File: packages/api/internal/sandbox/store.go [hunk 6]
```diff
@@ -156,11 +162,22 @@ func (s *Store) WaitForStateChange(ctx context.Context, teamID uuid.UUID, sandbo
...
+		// Redis is the source of truth — divergent sandboxes are orphans running
+		// on the node but not present in the store. Kill them.
+		for _, sbx := range sbxsToBeSynced {
+			go s.callbacks.RemoveSandboxFromNode(ctx, sbx)
```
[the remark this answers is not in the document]
sitole: Context is one thing, but it also escapes from sync contexts - it's possible it will be removed twice if it's slow. Lets do wait group instead

PR #2144, opened 2026-03-16
fix(storage): prevent caching truncated data in NFS cache

dobrac:
Stop writing short reads not ending with EOF to the NFS cache to prevent persisting corrupt/incomplete chunks. Fix also the cacheWriteThroughReader close with the same approach.

---

> [!NOTE]
> **Medium Risk**
> Changes caching eligibility and EOF handling for both NFS cache reads and GCS `ReadAt`, which can alter I/O behavior and cache hit rates; mistakes here could cause missed caching or continued corruption under edge-case short reads.
> 
> **Overview**
> Prevents corrupt/truncated data from being persisted in the NFS seekable cache by only treating reads as cacheable when they are complete (full buffer, or a non-empty short read explicitly explained by `io.EOF`). Cached `ReadAt` results now detect and reject truncated cache files (falling back to the remote read), remote `ReadAt` writeback is gated by the same completeness check, and `OpenRangeReader`’s write-through caching now writes on `Close` only when the total streamed bytes match the expected range length. GCS `ReadAt` is tightened to use `io.ReadFull` and normalize `io.ErrUnexpectedEOF` to `io.EOF`, and the test suite is expanded to cover truncated-cache fallback, EOF preservation, skip-writeback behavior, and range-reader truncation cases.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1d86f41686a16f973d24a710c52fd77d547b7d6e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/shared/pkg/storage/storage_cache_seekable.go
+++ b/packages/shared/pkg/storage/storage_cache_seekable.go
@@ -92,14 +92,10 @@ func (c *cachedSeekable) ReadAt(ctx context.Context, buff []byte, offset int64) 
 	readTimer := cacheSlabReadTimerFactory.Begin(attribute.String(nfsCacheOperationAttr, nfsCacheOperationAttrReadAt))
 	count, err := c.readAtFromCache(ctx, chunkPath, buff)
 	if ignoreEOF(err) == nil {
-		if isCompleteRead(count, len(buff), err) {
-			recordCacheRead(ctx, true, int64(count), cacheTypeSeekable, cacheOpReadAt)
-			readTimer.Success(ctx, int64(count))
+		recordCacheRead(ctx, true, int64(count), cacheTypeSeekable, cacheOpReadAt)
+		readTimer.Success(ctx, int64(count))
 
-			return count, err // return `err` in case it's io.EOF
-		}
-
-		err = fmt.Errorf("cached chunk is truncated: got %d bytes, expected %d", count, len(buff))
+		return count, err // return `err` in case it's io.EOF
 	}
 	readTimer.Failure(ctx, int64(count))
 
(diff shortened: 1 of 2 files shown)


packages/shared/pkg/storage/storage_google.go
```diff
@@ -298,26 +298,20 @@ func (o *gcpObject) ReadAt(ctx context.Context, buff []byte, off int64) (n int,
...
-		n += nr
-
-		if err == nil {
-			continue
-		}
-
-		if errors.Is(err, io.EOF) {
-			break
-		}
+	n, err = io.ReadFull(reader, buff)
```
dobrac, line 301: this should now match the AWS implementation

packages/shared/pkg/storage/storage_cache_seekable.go
```diff
@@ -199,7 +202,11 @@ func (r *cacheWriteThroughReader) Read(p []byte) (int, error) {
 func (r *cacheWriteThroughReader) Close() error {
 	closeErr := r.inner.Close()
 
-	if r.buf.Len() > 0 {
+	// Only cache when the total bytes read match the expected length.
+	// Unlike ReadAt where io.EOF can justify a short read (last chunk),
+	// a streaming reader always ends with EOF regardless of whether the
+	// data was truncated, so the byte count is the only reliable check.
+	if r.buf.Len() > 0 && int64(r.buf.Len()) == r.expectedLen {
```
jakubno, line 209: can expectedLen be 0?

dobrac: I don't think so with the way we're working with it right now (Always a full chunk or shorter chunk)

PR #2164, opened 2026-03-18
fix: allow make generate to work on mac

matthewlouisbrockman:
mockery loads packages/orchestrator/internal/sandbox/block on the host platform. On macOS, that package was failing to compile because cache.go referenced Linux-only x/sys/unix APIs (RemoteIovec,ProcessVMReadv) unconditionally.

### changed:
- moves copyProcessMemory into a Linux-only file
- adds a non-Linux stub so the package still loads on macOS
- marks the process-memory tests as Linux-only

---

> [!NOTE]
> **Medium Risk**
> Moves Linux-only process-memory copying behind build tags and adjusts `Cache.FileSize()` arithmetic, touching low-level mmap/process-memory code paths that could affect cache sizing and memory copy behavior on Linux.
> 
> **Overview**
> Makes the `sandbox/block` package compile on macOS and other non-Linux hosts by moving the `ProcessVMReadv`-based `copyProcessMemory` implementation into a Linux-only file and providing a non-Linux stub, while also marking the related tests as Linux-only and fixing `Cache.FileSize()` to compute allocated size with a compatible `st_blocks * bsize` calculation.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 116f99f4c8b330806feac1c619eed8b39a0278f5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/orchestrator/internal/sandbox/block/cache.go
+++ b/packages/orchestrator/internal/sandbox/block/cache.go
@@ -307,7 +307,13 @@ func (c *Cache) FileSize() (int64, error) {
 		return 0, fmt.Errorf("failed to get file stats: %w", err)
 	}
 
-	return stat.Blocks * 512, nil
+	var fsStat syscall.Statfs_t
+	err = syscall.Statfs(c.filePath, &fsStat)
+	if err != nil {
+		return 0, fmt.Errorf("failed to get disk stats for path %s: %w", c.filePath, err)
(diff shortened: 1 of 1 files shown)


dobrac:
lets run it in Docker instead

PR #2166, opened 2026-03-18
feat: add egress/ingress Off switches for traffic blocking

levb:
## Summary

  - Adds `egressOff` and `ingressOff` boolean fields to the sandbox network config (OpenAPI, proto, DB types)
  - When set, all traffic is denied immediately without iterating allow/deny rules
  - Off is mutually exclusive with allow/deny rules — the API rejects requests that set both
  - The legacy `allowInternetAccess=false` now sets `egressOff=true` internally instead of injecting `0.0.0.0/0` into the deny list
  - Egress Off is enforced at three layers: nftables (non-TCP), TCP firewall proxy (TCP with hostname), and pre-parsed ACL checks
  - Ingress Off is enforced at the reverse proxy layer, blocking all non-envd inbound traffic
  - Both switches work for create and dynamic update (`PUT /sandboxes/{id}/network`)


dobrac (commented):
Lets not do the `Off` right now and implement only the "AllowInternetAccess" switch for the egress network setup in the update_network API call.

dobrac (changes requested):

levb:
@dobrac I am not sure I see the value of adding this shortcut: it would accomplish nothing different from setting egress to Deny All, which can already be accomplished with the Update API. `AllowInternetAccess` is not used for anything else. I recommend we do not do it and close the PR. WDYT?

PR #2168, opened 2026-03-19
Refactor: Split dashboard-api queries & migrations

ben-fornefeld:


packages/dashboard-api/internal/handlers/build.go
```diff
@@ -22,7 +22,7 @@ func (s *APIStore) GetBuildsBuildId(c *gin.Context, buildId api.BuildId) {
 	teamID := auth.MustGetTeamInfo(c).Team.ID
 	telemetry.SetAttributes(ctx, telemetry.WithTeamID(teamID.String()), telemetry.WithBuildID(buildId.String()))
 
-	row, err := s.db.GetBuildInfoByTeamAndBuildID(ctx, queries.GetBuildInfoByTeamAndBuildIDParams{
+	row, err := s.db.GetBuildInfoByTeamAndBuildID(ctx, dashboardqueries.GetBuildInfoByTeamAndBuildIDParams{
```
dobrac, line 25: I know these queries are called only in the dashboard-api, but lets keep them colocated to the database (source), rather than usage. -> they should stay in core

packages/dashboard-api/main.go
```diff
@@ -93,22 +95,33 @@ func run() int {
...
 
-	err = sqlcdb.CheckMigrationVersion(ctx, config.PostgresConnectionString, expectedMigration)
+	expectedDashboardMigration, err := strconv.ParseInt(expectedDashboardMigrationTimestamp, 10, 64)
 	if err != nil {
-		l.Fatal(ctx, "failed to check migration version", zap.Error(err))
+		l.Warn(ctx, "Failed to parse expected dashboard migration timestamp", zap.Error(err))
+		expectedDashboardMigration = 0
+	}
+
+	err = sqlcdb.CheckMigrationVersionWithTable(ctx, config.PostgresConnectionString, "_migrations", expectedCoreMigration)
```
dobrac, line 110: the migration table should not be necessary to provide here, correct? Or can we at least reuse the const for the infra?

packages/db/pkg/dashboard/client.go
```diff
@@ -0,0 +1,46 @@
...
+
+const poolName = "dashboard"
+
+type Client struct {
+	*dashboardqueries.Queries
+
+	conn *pgxpool.Pool
+}
+
+func NewClient(ctx context.Context, databaseURL string, options ...pool.Option) (*Client, error) {
```
dobrac, line 22: why do we need this copied here?

packages/dashboard-api/main.go
```diff
@@ -93,22 +95,33 @@ func run() int {
...
+	if err != nil {
+		l.Fatal(ctx, "failed to check core migration version", zap.Error(err))
+	}
+
+	err = sqlcdb.CheckMigrationVersionWithTable(ctx, config.PostgresConnectionString, "_migrations_dashboard", expectedDashboardMigration)
```
dobrac, line 115: lets move `_migrations_dashboard` to const

dobrac (changes requested):

packages/dashboard-api/internal/handlers/build.go
ben-fornefeld: good point

packages/dashboard-api/main.go
ben-fornefeld: why not necessary?

dobrac: because its the default one

PR #2179, opened 2026-03-19
Use lifecycle id, add metrics

djeebus:
Also adds some observability and makes meter names consistent (similar to traces)

---

> [!NOTE]
> **Medium Risk**
> Medium risk due to interface changes in `sandbox.MapSubscriber` (now passes `context.Context` and full `*Sandbox` on callbacks) and new NFS proxy instrumentation that wraps filesystem operations, which could impact cleanup behavior and request latency if misconfigured.
> 
> **Overview**
> This PR updates sandbox lifecycle event subscriptions to pass `context.Context` and the full sandbox object (enabling consumers to key cleanup by `LifecycleID`), adds optional OpenTelemetry metrics for the NFS proxy (including per-operation call counts/latencies plus chroot mount/unmount/gauge metrics), and standardizes meter names across services to use package paths; CI is updated to auto-run new scripts that normalize tracer/meter declarations.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3d2e41221fbeb799ae4a3c714a2ce0fdf6f8d0a7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api/Makefile
+++ b/packages/api/Makefile
@@ -1,5 +1,7 @@
 ENV := $(shell cat ../../.last_used_env || echo "not-set")
 -include ../../.env.${ENV}
+
+LOCAL_BUILD_TARGET ?= build-debug
 PREFIX := $(strip $(subst ",,$(PREFIX)))
 HOSTNAME := $(shell hostname 2> /dev/null || hostnamectl hostname 2> /dev/null)
 $(if $(HOSTNAME),,$(error Failed to determine hostname: both 'hostname' and 'hostnamectl' failed))
@@ -54,8 +56,7 @@ define setup_local_env
 endef
 
 .PHONY: run-local
-run-local:
-	make build-debug
+run-local: $(LOCAL_BUILD_TARGET)
 	$(call setup_local_env)
 	NODE_ID=$(HOSTNAME) ./bin/api --port 3000
 
--- a/packages/client-proxy/Makefile
+++ b/packages/client-proxy/Makefile
@@ -1,5 +1,7 @@
 ENV := $(shell cat ../../.last_used_env || echo "not-set")
 -include ../../.env.${ENV}
+
+LOCAL_BUILD_TARGET ?= build-debug
 PREFIX := $(strip $(subst ",,$(PREFIX)))
 HOSTNAME := $(shell hostname 2> /dev/null || hostnamectl hostname 2> /dev/null)
 $(if $(HOSTNAME),,$(error Failed to determine hostname: both 'hostname' and 'hostnamectl' failed))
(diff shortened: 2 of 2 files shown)


packages/orchestrator/pkg/proxy/proxy.go
```diff
@@ -197,6 +197,6 @@ func (p *SandboxProxy) OnInsert(_ *sandbox.Sandbox) {}
 
 // OnRemove is called when a sandbox is removed from the map.
 // It cleans up the connection limiter entry for the sandbox.
-func (p *SandboxProxy) OnRemove(sandboxID string) {
-	p.limiter.Remove(sandboxID)
+func (p *SandboxProxy) OnRemove(sbx *sandbox.Sandbox) {
```
levb: why this change if you still just need the ID?

packages/shared/pkg/logger/fields.go
```diff
@@ -13,6 +13,10 @@ func WithSandboxID(sandboxID string) zap.Field {
 	return zap.String("sandbox.id", sandboxID)
 }
 
+func WithLifecycleID(lifecycleID string) zap.Field {
+	return zap.String("lifecycle.id", lifecycleID)
```
dobrac: `sandbox.lifecycle.id`

packages/orchestrator/pkg/proxy/proxy.go
djeebus: My understanding is that, if you pause/resume the sandbox, the sandbox ID stays the same, but the lifecycle ID is different. You still cannot (currently) resume the same sandbox more than once, so the sandbox map is still keyed by sandbox ID, but this helps to ensure that, in the case of a quick stop/start, we still remove the correct mounts at the right time.

packages/orchestrator/pkg/cfg/model_test.go
```diff
@@ -64,17 +64,17 @@ func TestParse(t *testing.T) {
...
-		t.Setenv("NFS_PROXY_TRACING", "false")
+	t.Run("nfs proxy metrics can be disabled", func(t *testing.T) {
+		t.Setenv("NFS_PROXY_METRICS", "false")
 
 		config, err := Parse()
 		require.NoError(t, err)
-		assert.False(t, config.NFSProxyTracing)
+		assert.False(t, config.NFSProxyMetrics)
 	})
 }
```
dobrac, line 80: should this be for both?

djeebus: This is really just to verify that caarlos0/env does what it's supposed to do (support defaulting  a boolean variable to true, then allowing overrides to false). The defaults changed, so the test had to change as well.

PR #2191, opened 2026-03-20
fix(api): race conditions with multiple APIs and fresh orchestrators

jakubno:
Fixes:
//  1. A new orchestrator node is running 
//  2. Nomad service discovery knows about it 
//  3. API 1 creates a sandbox on the node
//  4. API 2 receives a request to manipulate sandbox on the node, which it doesn't know about yet

---

> [!NOTE]
> **Medium Risk**
> Changes node lookup/connection and synchronization concurrency behavior, which can affect sandbox operations under load if the new on-demand discovery or singleflight keys/timeouts are wrong. Risk is moderated by added unit/integration-style tests covering cache-miss and concurrency scenarios.
> 
> **Overview**
> Fixes a multi-API race where sandbox operations can target an orchestrator that joined recently and isn’t yet in the local node cache by adding an on-demand discovery+connect path on cache misses, plus `singleflight`-based deduplication for concurrent connect/discovery attempts. It also enables immediate cluster instance resync (`SyncInstances`) and makes `Synchronize.Sync` context-cancellable via a semaphore to avoid concurrent sync rounds blocking indefinitely, with new tests validating the new cache-miss and concurrency behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ad5d1ff96963d3775d4816d12504323ff3142987. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/api/internal/orchestrator/client.go
+++ b/packages/api/internal/orchestrator/client.go
@@ -45,6 +45,9 @@ func (o *Orchestrator) connectToNode(ctx context.Context, discovered nodemanager
 }
 
 func (o *Orchestrator) connectToClusterNode(ctx context.Context, cluster *clusters.Cluster, i *clusters.Instance) {
+	ctx, span := tracer.Start(ctx, "connect-to-cluster-node")
+	defer span.End()
+
 	// connectGroup is keyed by scopedNodeID so that concurrent callers targeting
 	// the same cluster instance share a single dial attempt.
 	scopedKey := o.scopedNodeID(cluster.ID, i.NodeID)
@@ -139,6 +142,9 @@ func (o *Orchestrator) GetNode(clusterID uuid.UUID, nodeID string) *nodemanager.
 // discoveryGroup ensures that concurrent requests targeting the same missing
 // node share a single discovery attempt rather than fanning out.
 func (o *Orchestrator) getOrConnectNode(ctx context.Context, clusterID uuid.UUID, nodeID string) *nodemanager.Node {
+	ctx, span := tracer.Start(ctx, "get-or-connect-node")
+	defer span.End()
+
 	if node := o.GetNode(clusterID, nodeID); node != nil {
 		return node
 	}
@@ -160,7 +166,7 @@ func (o *Orchestrator) getOrConnectNode(ctx context.Context, clusterID uuid.UUID
 		defer cancel()
 
 		if clusterID == consts.LocalClusterID {
-			o.discoverNomadNode(connectCtx)
+			o.discoverNomadNodes(connectCtx)
 		} else {
 			o.discoverClusterNode(connectCtx, clusterID)
 		}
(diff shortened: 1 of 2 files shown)


packages/api/internal/orchestrator/client.go
```diff
@@ -94,6 +124,105 @@ func (o *Orchestrator) GetNode(clusterID uuid.UUID, nodeID string) *nodemanager.
...
+// performs a targeted on-demand discovery and connection attempt, handling the race
+// condition where a new orchestrator joined the cluster after this API instance's last
+// sync cycle but another API instance already routed a sandbox there.
+//
+// There are two distinct gaps that must be covered:
+//   - Gap 1 (0–5 s for clusters, 0–20 s for Nomad): the node exists in the upstream
+//     source (Nomad / remote service discovery) but has not yet been pulled into the
+//     local instance map by the background sync loop.
+//   - Gap 2 (0–20 s): the node is in the local instance map but has not yet been
+//     promoted into o.nodes by keepInSync.
```
sitole, line 140: We can simplify by removing two layers os nodes sync. This should remove duplicated logic, but it should be done separately as it will introduce lot of changes.

packages/api/internal/orchestrator/client.go
```diff
@@ -94,6 +124,105 @@ func (o *Orchestrator) GetNode(clusterID uuid.UUID, nodeID string) *nodemanager.
...
+//
+// discoveryGroup ensures that concurrent requests targeting the same missing
+// node share a single discovery attempt rather than fanning out.
+func (o *Orchestrator) getOrConnectNode(ctx context.Context, clusterID uuid.UUID, nodeID string) *nodemanager.Node {
```
sitole, line 144: Let's add some tracing here so we can see why some sandbox requests will be slower.

packages/api/internal/orchestrator/client.go
```diff
@@ -94,6 +124,105 @@ func (o *Orchestrator) GetNode(clusterID uuid.UUID, nodeID string) *nodemanager.
...
+
+		return nil, nil
+	})
+
+	return o.GetNode(clusterID, nodeID)
+}
+
+// discoverNomadNode lists all ready Nomad nodes and connects any that are not yet in the pool.
+// Once a new node is connected its orchestrator ID becomes the map key, making subsequent GetNode calls succeed.
+func (o *Orchestrator) discoverNomadNode(ctx context.Context) {
```
sitole:
```suggestion
func (o *Orchestrator) discoverNomadNodes(ctx context.Context) {
```

packages/api/internal/orchestrator/client.go
```diff
@@ -94,6 +124,105 @@ func (o *Orchestrator) GetNode(clusterID uuid.UUID, nodeID string) *nodemanager.
...
+		return
+	}
+
+	var wg sync.WaitGroup
+	defer wg.Wait()
+
+	for _, n := range nomadNodes {
+		if o.GetNodeByNomadShortID(n.NomadNodeShortID) == nil {
+			wg.Go(func() {
+				if err := o.connectToNode(ctx, n); err != nil {
```
sitole, line 199: I think we can hit the race of manual connect and natural periodic one?

sitole: Same for discoverClusterNode

jakubno: that's why there's `connectGroup`in `connectToNode` handling exactly this situation 
same for `discoverClusterNode`, there's `connectGroup` in `connectToClusterNode`

PR #2199, opened 2026-03-22
feat(orch): move reserved blocks setting from in-guest to host-side pre-boot hook

arkamar:
Narrows the issue introduced in #2082 where setting reserved disk space required running `tune2fs` inside the guest VM. This approach failed when the guest image didn't have `e2fsprogs` installed (e.g. `from_template` builds inheriting from older parents) or when the guest `e2fsprogs` version was incompatible with the host's ext4 features.

Obsoletes #2194, which tried to fix this by adding `e2fsprogs` to `provision.sh`. This would force cache invalidation for all existing templates but still didn't cover the `from_template` path where provisioning is skipped entirely.

Instead, a `PreBootFn` hook is added to `Factory.CreateSandbox()` that runs after the rootfs device is ready but before Firecracker boots. The finalize phase uses this hook to call `SetReservedBlocksOnHost` using the host's `tune2fs`, which is always available and compatible. The in-guest `SetReservedBlocksInGuest` is removed.

Close: #2194 

---

> [!NOTE]
> **Medium Risk**
> Touches sandbox creation by adding a pre-boot callback that can mutate the rootfs before Firecracker starts, which could affect boot behavior or disk state if misused. Risk is limited to template build sandboxes and is gated by a feature flag for reserved disk space.
> 
> **Overview**
> Adds an optional `PreBootFn` hook to sandbox creation that runs after the rootfs device is prepared but before Firecracker boots, enabling host-side filesystem adjustments. Template builds now use this hook (behind `BuildReservedDiskSpaceMB`) to set ext4 reserved blocks via host `tune2fs` instead of running `tune2fs` inside the guest, and the in-guest reserved-blocks command path is removed while feature-flag plumbing is threaded through the build step/user builders.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 21cd01a8aeed553e636f867d0b458e4e47787590. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/orchestrator/pkg/template/build/phases/finalize/builder.go
+++ b/packages/orchestrator/pkg/template/build/phases/finalize/builder.go
@@ -178,7 +178,9 @@ func (ppb *PostProcessingBuilder) Build(
 	sandboxOptions := []layer.CreateSandboxOption{
 		layer.WithIoEngine(ioEngine),
 	}
-	sandboxOptions = append(sandboxOptions, layer.ReservedBlocksOptions(ctx, ppb.featureFlags, ppb.Config.RootfsBlockSize())...)
+	if sourceLayer.Cached {
+		sandboxOptions = append(sandboxOptions, layer.ReservedBlocksOptions(ctx, ppb.featureFlags, ppb.Config.RootfsBlockSize())...)
+	}
 
 	// Always restart the sandbox for the final layer to properly wire the rootfs path for the final template
 	sandboxCreator := layer.NewCreateSandbox(


packages/orchestrator/pkg/template/build/phases/finalize/builder.go
```diff
@@ -174,12 +175,22 @@ func (ppb *PostProcessingBuilder) Build(
...
+		}
+	}
+
 	// Always restart the sandbox for the final layer to properly wire the rootfs path for the final template
 	sandboxCreator := layer.NewCreateSandbox(
 		sbxConfig,
 		ppb.sandboxFactory,
 		finalizeTimeout,
 		layer.WithIoEngine(ioEngine),
+		layer.WithPreBootFn(preBootFn),
```
djeebus: If this pattern continues, it might be easier to collect options up front, then pass them all in:

```golang
var options []CreateSandboxOption

if reservedDiskSpaceMB > 0 {
    options = append(options, layer.WithPreBootFn(myFancyFunc))
}

sandboxCreator : layer.NewCreateSandbox(
    ppb.sandboxFactory,
    finalizeTimeout,
    options...,
)
```

Just a thought!

arkamar: Addressed in the latest push. Thanks for suggestion!

packages/orchestrator/pkg/template/build/phases/finalize/builder.go
```diff
@@ -174,12 +174,18 @@ func (ppb *PostProcessingBuilder) Build(
 	span.SetAttributes(attribute.String("io_engine", ioEngine))
 	ppb.logger.Debug(ctx, "using io engine", zap.String("io_engine", ioEngine))
 
+	// Collect sandbox creation options
+	sandboxOptions := []layer.CreateSandboxOption{
+		layer.WithIoEngine(ioEngine),
+	}
+	sandboxOptions = append(sandboxOptions, layer.ReservedBlocksOptions(ctx, ppb.featureFlags, ppb.Config.RootfsBlockSize())...)
```
dobrac: this should be guarded by the `sourceLayer.Cached` too

PR #2204, opened 2026-03-23
Use base e2b SDK in tests that don't need code-interpreter

beran-t:
## Summary
- `internet-works.ts`, `time-is-synchronized/index.ts`, and `utils.ts` only use `sandbox.commands.run()` and sandbox lifecycle methods — no `runCode()` needed
- Changed these files to import from base `e2b` instead of `@e2b/code-interpreter`
- Added `e2b` as explicit dependency in package.json
- `run-code.ts` and `snapshot-and-resume.ts` still correctly use `@e2b/code-interpreter` (they call `runCode()`)

## Test plan
- [ ] Verify periodic tests still pass with the updated imports

### DIFF
--- a/tests/periodic-test/package.json
+++ b/tests/periodic-test/package.json
@@ -5,8 +5,8 @@
   "type": "module",
   "dependencies": {
     "@e2b/cli": "^2.7.3",
-    "@e2b/code-interpreter": "^2.3.3",
-    "e2b": "^2.3.0"
+    "@e2b/code-interpreter": "^2.4.0",
+    "e2b": "^2.15.0"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


tests/periodic-test/package.json
```diff
@@ -5,7 +5,8 @@
   "type": "module",
   "dependencies": {
     "@e2b/cli": "^2.7.3",
-    "@e2b/code-interpreter": "^2.3.3"
+    "@e2b/code-interpreter": "^2.3.3",
+    "e2b": "^2.3.0"
```
dobrac: is this the code-interpreter 2.3.3 version? Would it make sense to update to latest?

PR #2209, opened 2026-03-24
Rm client proxy cache flag

matthewlouisbrockman:
Resolves the feature flag for disabling local cache by letting the caller pass it in. API retains it, client proxy loses it.

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes runtime caching behavior for sandbox routing lookups (potential latency increase on client proxies and different cache consistency characteristics), and removes a feature-flag fallback path.
> 
> **Overview**
> This PR removes the `SandboxCatalogLocalCacheFlag` and instead makes Redis sandbox catalog local caching an explicit `useLocalCache` constructor parameter, with `client-proxy` now constructing the catalog with caching disabled while the API orchestrator keeps caching enabled. The Redis catalog implementation is updated to treat the cache as optional (`nil`) and the unit test is rewritten to assert behavior with caching enabled vs disabled without relying on feature-flag targeting.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f40d8ed6ce549a8f7317251c962684bc8799cbd6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/shared/pkg/sandbox-catalog/catalog_redis_test.go
+++ b/packages/shared/pkg/sandbox-catalog/catalog_redis_test.go
@@ -14,6 +14,8 @@ import (
 )
 
 func seedRedisCatalogEntry(t *testing.T) (context.Context, redis.UniversalClient, string, *SandboxInfo) {
+	t.Helper()
+
 	ctx := t.Context()
 	redisClient := redis_utils.SetupInstance(t)
 


packages/shared/pkg/sandbox-catalog/catalog_redis.go
```diff
@@ -24,21 +23,25 @@ const (
...
-	cache := ttlcache.New(ttlcache.WithTTL[string, *SandboxInfo](catalogRedisLocalCacheTtl), ttlcache.WithDisableTouchOnHit[string, *SandboxInfo]())
-	go cache.Start()
+func NewRedisSandboxesCatalog(redisClient redis.UniversalClient, useLocalCache bool) *RedisSandboxCatalog {
+	var cache *ttlcache.Cache[string, *SandboxInfo]
+	if useLocalCache {
+		cache = ttlcache.New(
+			ttlcache.WithTTL[string, *SandboxInfo](catalogRedisLocalCacheTtl),
+			ttlcache.WithDisableTouchOnHit[string, *SandboxInfo](),
+		)
+		go cache.Start()
```
dobrac: lets pass this as parameter instead of boolean

matthewlouisbrockman: k

matthewlouisbrockman: oh you meant an enum didn't you

dobrac: I've meant passing the instance directly. That should remove most ifs in the code path

packages/shared/pkg/sandbox-catalog/catalog_redis_test.go
```diff
@@ -40,52 +31,29 @@ func TestRedisCatalog_LocalCacheFlagServiceContext(t *testing.T) {
...
-			assert.NoError(t, catalog.Close(context.Background()))
-		})
-
-		got, err := catalog.GetSandbox(ctx, sbxID)
-		require.NoError(t, err)
-		require.Equal(t, expected.OrchestratorID, got.OrchestratorID)
-		require.Equal(t, expected.ExecutionID, got.ExecutionID)
-
-		assert.Nil(t, catalog.cache.Get(sbxID), "local cache should remain empty when flag is disabled for this service")
+	// Without local cache — reads from Redis, no cache created.
```
dobrac: this should be a test case rather than comment

packages/shared/pkg/sandbox-catalog/catalog_redis_test.go
```diff
@@ -40,52 +31,29 @@ func TestRedisCatalog_LocalCacheFlagServiceContext(t *testing.T) {
...
-		t.Cleanup(func() {
-			assert.NoError(t, catalog.Close(context.Background()))
-		})
+	got, err := noCacheCatalog.GetSandbox(ctx, sbxID)
+	require.NoError(t, err)
+	require.Equal(t, expected.OrchestratorID, got.OrchestratorID)
+	require.Equal(t, expected.ExecutionID, got.ExecutionID)
+	assert.Nil(t, noCacheCatalog.cache, "local cache should not be created when disabled")
+
+	// With local cache — reads from Redis, populates cache.
```
dobrac: this should be a test case rather than comment

packages/shared/pkg/sandbox-catalog/catalog_redis.go
```diff
@@ -24,33 +23,48 @@ const (
...
+
+		return cache
+	default:
+		return nil
 	}
 }
 
-var _ SandboxesCatalog = (*RedisSandboxCatalog)(nil)
```
dobrac, line 45: is this no longer relevant?

matthewlouisbrockman: was just moved up 10 lines

packages/shared/pkg/sandbox-catalog/catalog_redis_test.go
matthewlouisbrockman: splitting tests

packages/shared/pkg/sandbox-catalog/catalog_redis_test.go
matthewlouisbrockman: splitting tests

packages/shared/pkg/sandbox-catalog/catalog_redis.go
matthewlouisbrockman: ah, kk. modeling it after repository_noop

PR #2216, opened 2026-03-24
chore(envd): add request timeout field in logs

jakubno:
Add request timeout field to process start log to improve debugging. It should help to distinguish between expected cancels and other issues

### DIFF
--- a/packages/envd/internal/services/process/handler/handler.go
+++ b/packages/envd/internal/services/process/handler/handler.go
@@ -43,8 +43,7 @@ type ProcessExit struct {
 type Handler struct {
 	Config *rpc.ProcessConfig
 
-	logger         *zerolog.Logger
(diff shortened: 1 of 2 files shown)


packages/envd/internal/services/process/handler/handler.go
```diff
@@ -433,6 +437,7 @@ func (p *Handler) Start() (uint32, error) {
 		Str("event_type", "process_start").
 		Int("pid", p.cmd.Process.Pid).
 		Str("command", p.userCommand()).
+		Dur("request_timeout_ms", p.requestTimeout).
```
sitole: Why not put it directly as a parameter of the `Start()` method?

jakubno: that'll be probably cleaner

PR #2219, opened 2026-03-24
refactor(orchestrator): extract run() into pkg/factories for reuse

sitole:
Move the orchestrator lifecycle logic into an exported factories.Run() function with an Options struct that accepts Version, CommitSHA, and an EgressFactory callback. This allows orchestrator-ee to reuse the full init/shutdown flow while plugging in its own egress proxy implementation.

### DIFF
--- a/packages/orchestrator/main.go
+++ b/packages/orchestrator/main.go
@@ -10,6 +10,7 @@ import (
 const version = "0.1.0"
 
 var commitSHA string
+
 func main() {
 	factories.Run(factories.Options{
 		Version:       version,


packages/orchestrator/main.go
```diff
@@ -2,824 +2,35 @@ package main
...
-	"github.com/e2b-dev/infra/packages/shared/pkg/utils"
 )
 
-type closer struct {
-	name  string
-	close func(ctx context.Context) error
-}
+var commitSHA string
 
 const version = "0.1.0"
```
arkamar:
```suggestion
const version = "0.1.0"

var commitSHA string
```
Very niche nitpick, maybe OCD type, but can we swap `commitSHA` with `version` as it was originally? I know it is not mandatory, but `go doc` presents constants first followed by variables, if both are present in the package and we actually prefer this order, see the stat for this branch:
```
const first, then var: 100
var first, then const: 36
only const: 209
only var: 148
```

packages/orchestrator/main.go
```diff
@@ -2,824 +2,34 @@ package main
...
 
-type closer struct {
-	name  string
-	close func(ctx context.Context) error
-}
-
 const version = "0.1.0"
 
 var commitSHA string
-
```
arkamar, line 80: Ugh, my previous suggestion removed the extra space here :S

arkamar: ah, linter found it.

PR #2222, opened 2026-03-25
Increase granularity on ErrNotEvictable

matthewlouisbrockman:
Issue: When we skip evicting a sandbox, ErrNotEvictable is used to mean either an eviction is in progress or there's a state mismatch on the eviction timeout. This makes it difficult to tell why evictions weren't attempted.

Changes: Splits ErrNotEvictable into named errors (ErrEvictionInProgress, ErrEvictionNotNeeded) that specify why the eviction didn't happen.

### DIFF
--- a/packages/api/internal/orchestrator/evictor/evict.go
+++ b/packages/api/internal/orchestrator/evictor/evict.go
@@ -64,14 +64,14 @@ func (e *Evictor) Start(ctx context.Context) {
 					if err := e.removeSandbox(context.WithoutCancel(ctx), item.TeamID, item.SandboxID, sandbox.RemoveOpts{Action: action, Eviction: true}); err != nil {
 						if action == sandbox.StateActionPause {
 							switch {
-							case errors.Is(err, sandbox.ErrEvictionInProgress), errors.Is(err, sandbox.ErrEvictionNotNeeded):
+							case isNotEvictableError(err):
 								pause.LogSkipped(ctx, item.SandboxID, item.TeamID.String(), pause.ReasonTimeout, pause.SkipReasonNotEvictable)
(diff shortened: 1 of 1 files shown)


packages/api/internal/orchestrator/evictor/evict.go
```diff
@@ -64,14 +64,14 @@ func (e *Evictor) Start(ctx context.Context) {
...
-							case errors.Is(err, sandbox.ErrNotEvictable):
+							case errors.Is(err, sandbox.ErrEvictionInProgress), errors.Is(err, sandbox.ErrEvictionNotNeeded):
 								pause.LogSkipped(ctx, item.SandboxID, item.TeamID.String(), pause.ReasonTimeout, pause.SkipReasonNotEvictable)
 							case errors.Is(err, sandbox.ErrNotFound):
 								pause.LogSkipped(ctx, item.SandboxID, item.TeamID.String(), pause.ReasonTimeout, pause.SkipReasonNotFound)
 							default:
 								pause.LogFailure(ctx, item.SandboxID, item.TeamID.String(), pause.ReasonTimeout, err)
 							}
-						} else if !errors.Is(err, sandbox.ErrNotEvictable) && !errors.Is(err, sandbox.ErrNotFound) {
+						} else if !errors.Is(err, sandbox.ErrEvictionInProgress) && !errors.Is(err, sandbox.ErrEvictionNotNeeded) && !errors.Is(err, sandbox.ErrNotFound) {
```
jakubno: can you move this to some util function?

matthewlouisbrockman: done

PR #1875, opened 2026-02-10
feat: add initial ARM64 (aarch64) architecture support

tomassrnka:
## Summary

Adds ARM64/aarch64 architecture support to the E2B infrastructure, enabling builds and sandbox execution on Apple Silicon and other ARM64 hosts (via Lima VM + nested KVM).

**Changes by commit:**

1. **Makefiles** — Replace hardcoded `GOARCH=amd64` and `--platform linux/amd64` with `$(shell go env GOARCH)` across all 4 service Makefiles
2. **Go runtime detection** — Disable SMT on ARM64 (not supported), use `runtime.GOARCH` for OCI image platform, add ARM64 fallback for CPU detection (gopsutil doesn't populate Family/Model on ARM)
3. **Provision script** — Make `chattr` calls non-fatal (`|| true`) for busybox versions that lack it
4. **create-build** — Arch-aware Firecracker and kernel download URLs (tries `arm64/` subdirectory first, falls back to generic), `E2B_BASE_IMAGE` env var for base image override
5. **fetch-busybox** — Makefile target to swap the embedded x86 busybox binary with the system's ARM64 busybox-static before compilation

**Related PRs:**
- [fc-kernels: ARM64 kernel build support](https://github.com/e2b-dev/fc-kernels/pull/13)

## Test plan

- [ ] Build orchestrator, envd, API, and client-proxy natively on ARM64 Linux
- [ ] Run `make fetch-busybox` on ARM64 host to swap busybox binary
- [ ] Template build succeeds with `create-build` on ARM64
- [ ] Sandbox create/exec/delete works on ARM64 (Lima VM + KVM)
- [ ] Verify `uname -m` in sandbox returns `aarch64`
- [ ] Confirm no regression on x86_64 builds (all changes are backwards compatible)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **High Risk**
> High risk because it changes core sandbox/Firecracker startup and snapshot handling (including disabling seccomp on ARM64) and alters kernel/Firecracker path resolution and OCI platform selection, which can impact both security posture and runtime stability across architectures.
> 
> **Overview**
[...]

PR #2227, opened 2026-03-25
fix(orchestrator): harden provision.sh with chattr fallback and APT time fix

tomassrnka:
Cherrypicked fix from #1875 as separate PR

### DIFF
--- a/packages/orchestrator/pkg/template/build/phases/base/provision.sh
+++ b/packages/orchestrator/pkg/template/build/phases/base/provision.sh
@@ -7,7 +7,14 @@ RESULT_PATH="{{ .ResultPath }}"
 echo "Starting provisioning script"
 
 echo "Making configuration immutable"
-$BUSYBOX chattr +i /etc/resolv.conf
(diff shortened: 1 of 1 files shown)


packages/orchestrator/pkg/template/build/phases/base/provision.sh
```diff
@@ -7,7 +7,14 @@ RESULT_PATH="{{ .ResultPath }}"
 echo "Starting provisioning script"
 
 echo "Making configuration immutable"
-$BUSYBOX chattr +i /etc/resolv.conf
+if $BUSYBOX chattr +i /etc/resolv.conf 2>/dev/null; then
+    CHATTR_AVAILABLE=1
+elif command -v chattr >/dev/null 2>&1 && chattr +i /etc/resolv.conf 2>/dev/null; then
+    CHATTR_AVAILABLE=1
```
arkamar, line 13: Is this fallback really necessary? What I mean is that `busybox` should be always available, right? Therefore if `$BUSYBOX chattr +i` fail the `chattr +i` will fail as well. Or am I missing something?

PR #2034, opened 2026-03-02
feature: Compression for memfile/rootfs assets

levb:
### Summary

  Compression for data files (memfile, rootfs). Files are broken into independently decompressible frames (2 MiB, zstd), stored in GCS alongside V4 headers with per-mapping frame tables. Fully backward-compatible: the read path auto-detects V3/V4 headers and routes compressed vs uncompressed reads per-mapping. Gated by compress-config LaunchDarkly flag (per-team/cluster/template targeting).

### What changed

  - `FramedFile` interface replaces `Seekable` — unified `GetFrame(ctx, offset, frameTable, decompress, buf, readSize, onRead)` handles both compressed and uncompressed data
  - V4 header with `FrameTable` per mapping + `BuildFileInfo` (uncompressed size, SHA-256 checksum) per build; LZ4-block-compressed header blob
  - NFS cache extended for compressed frames (.frm files keyed by compressed offset+size); progressive streaming decompression on cache miss; write-through on upload
  - P2P resume integration — peers read uncompressed from origin during upload, then atomically swap to V4 header (CAS) when origin signals use_storage with serialized headers
  - compress-build CLI for background compression of existing uncompressed builds (supports --recursive for dependency chains)
  - New `Chunker` with mmap cache, and fetch sessions dedupe replacing streaming_chunk.go

### Read path
```
  NBD/UFFD/Prefetch
    → header.GetShiftedMapping(offset) → BuildMap + FrameTable
    → DiffStore.Get(ctx, diff)         → cached Chunker
    → Chunker.GetBlock(offset, len, ft)
        → mmap hit? return reference
        → miss: fetchSession (dedup) → GetFrame
            → NFS hit? decompress from disk → mmap
            → NFS miss? GCS range read → decompress → mmap + NFS write-back
```
### P2P header switchover
```
  Origin (pause):
    snapshot → register buildID in Redis → serve mmap cache via gRPC
    background: upload compressed data + V4 headers to GCS
    on completion: uploadedBuilds.Set(buildID, serialized V4 headers)
                → peerRegistry.Unregister(buildID)

  Peer (resume, upload in progress):
    GetFrame(ft=nil) → gRPC stream → origin serves from mmap (uncompressed)

  Peer (origin signals use_storage):
    checkPeerAvailability() → transitionHeaders.Store({memH, rootH})
                            → uploaded.Store(true)
    next GetFrame(ft=nil): ft==nil + transitionHeaders != nil
      → return PeerTransitionedError{headers}
[...]

PR #2235, opened 2026-03-26
perf(block): sync.Map -> atomic bitmap for cache dirty tracking

levb:
Pre-cursor to PR #2034, separated to reduce the scope there.

Replace the `sync.Map`-based dirty block tracking in `block.Cache` with a pre-allocated `[]atomic.Uint64` bitset. Each bit represents one block; atomic OR (`atomic.Uint64.Or`) is used for concurrent-safe marking.

## Problem

The block cache tracks which blocks have been written (are "dirty") so that `Slice()` can decide whether to serve from mmap or return `BytesNotAvailable`. The previous implementation used `sync.Map` with one entry per dirty block offset. For a 64 MiB file at 4K block size, that's up to 16,384 map entries — each requiring a heap-allocated key and value, plus the internal hash map overhead of sync.Map.

Every call to `setIsCached` allocated a `[]int64` via `header.BlocksOffsets()` and then called `sync.Map.Store()` in a loop. Every call to `isCached` did the same allocation and called `sync.Map.Load()` in a loop. These are the hottest paths in the block device — called on every NBD read and every chunk fetch.

## Solution

- `dirty` field changed from `sync.Map` to `[]atomic.Uint64`, sized at `ceil(numBlocks / 64)` and allocated once in `NewCache`.
- `setIsCached` computes a bitmask per 64-bit word and applies it with `atomic.Uint64.Or` — one atomic op per word instead of one map store per block.
- `isCached` / `isBlockCached` read bits with `atomic.Uint64.Load` — no allocation, no map lookup.
- `dirtySortedKeys` iterates words and uses `bits.TrailingZeros64` to extract set bits — produces a naturally sorted result without `slices.Sort`.

## Benchmarks

64 MiB cache, 4K blocks, 4 MiB chunks (1024 blocks per chunk):

```
                   │  sync.Map      │            bitmap                  │
                   │     sec/op     │   sec/op     vs base               │
MarkRangeCached-16   63857.50n ± 3%   25.04n ± 6%  -99.96% (p=0.002 n=6)
IsCached_Hit-16       19406.0n ± 1%   908.6n ± 0%  -95.32% (p=0.002 n=6)
IsCached_Miss-16     1168.500n ± 1%   2.813n ± 1%  -99.76% (p=0.002 n=6)
Slice_Hit-16          19296.0n ± 1%   913.7n ± 1%  -95.27% (p=0.002 n=6)
Slice_Miss-16        1161.000n ± 3%   3.765n ± 0%  -99.68% (p=0.002 n=6)

                   │ sync.Map     │              bitmap                     │
                   │     B/op     │     B/op      vs base                   │
MarkRangeCached-16   64.05Ki ± 0%   0.00Ki ± 0%  -100.00% (p=0.002 n=6)
IsCached_Hit-16      8.000Ki ± 0%   0.000Ki ± 0%  -100.00% (p=0.002 n=6)
IsCached_Miss-16     8.000Ki ± 0%   0.000Ki ± 0%  -100.00% (p=0.002 n=6)
Slice_Hit-16         8.000Ki ± 0%   0.000Ki ± 0%  -100.00% (p=0.002 n=6)
Slice_Miss-16        8.000Ki ± 0%   0.000Ki ± 0%  -100.00% (p=0.002 n=6)

                   │ sync.Map     │              bitmap                     │
                   │  allocs/op   │  allocs/op   vs base                    │
MarkRangeCached-16    2.049k ± 0%   0.000k ± 0%  -100.00% (p=0.002 n=6)
IsCached_Hit-16        1.000 ± 0%    0.000 ± 0%  -100.00% (p=0.002 n=6)
IsCached_Miss-16       1.000 ± 0%    0.000 ± 0%  -100.00% (p=0.002 n=6)
Slice_Hit-16           1.000 ± 0%    0.000 ± 0%  -100.00% (p=0.002 n=6)
Slice_Miss-16          1.000 ± 0%    0.000 ± 0%  -100.00% (p=0.002 n=6)
```

Zero allocations across all operations. `setIsCached` goes from 64µs to 25ns (2550x), `Slice` hit from 19µs to 914ns (21x).

### DIFF
--- a/packages/orchestrator/go.mod
+++ b/packages/orchestrator/go.mod
@@ -13,8 +13,6 @@ tool github.com/oapi-codegen/oapi-codegen/v2/cmd/oapi-codegen
 
 tool github.com/vektra/mockery/v3
 
-replace github.com/RoaringBitmap/roaring/v2 => github.com/e2b-dev/roaring/v2 v2.16.1-0.20260410073213-00ee8aef09f9
-
 require (
 	cloud.google.com/go/storage v1.59.2
 	connectrpc.com/connect v1.18.1
@@ -99,7 +97,7 @@ require (
 	github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/metric v0.54.0 // indirect
 	github.com/GoogleCloudPlatform/opentelemetry-operations-go/internal/resourcemapping v0.54.0 // indirect
 	github.com/Microsoft/go-winio v0.6.2 // indirect
-	github.com/RoaringBitmap/roaring/v2 v2.16.0 // indirect
+	github.com/RoaringBitmap/roaring/v2 v2.16.1 // indirect
 	github.com/andybalholm/brotli v1.2.0 // indirect
 	github.com/armon/go-metrics v0.4.1 // indirect
 	github.com/aws/aws-sdk-go-v2 v1.41.0 // indirect
--- a/packages/orchestrator/go.sum
+++ b/packages/orchestrator/go.sum
@@ -112,6 +112,8 @@ github.com/NYTimes/gziphandler v0.0.0-20170623195520-56545f4a5d46/go.mod h1:3wb0
 github.com/OneOfOne/xxhash v1.2.2/go.mod h1:HSdplMjZKSmBqAxg5vPj2TmRDmfkzw+cTzAElWljhcU=
 github.com/PuerkitoBio/purell v1.1.1/go.mod h1:c11w/QuzBsJSee3cPx9rAFu61PvFxuPbtSwDGJws/X0=
 github.com/PuerkitoBio/urlesc v0.0.0-20170810143723-de5bf2ad4578/go.mod h1:uGdkoq3SwY9Y+13GIhn11/XLaGBb4BfwItxLd5jeuXE=
+github.com/RoaringBitmap/roaring/v2 v2.16.1 h1:HmKj5kwyUJVvCuaM8BdLSHVlYn3O9hzgSpaog6bSxmY=
+github.com/RoaringBitmap/roaring/v2 v2.16.1/go.mod h1:eq4wdNXxtJIS/oikeCzdX1rBzek7ANzbth041hrU8Q4=
 github.com/Shopify/logrus-bugsnag v0.0.0-20171204204709-577dee27f20d/go.mod h1:HI8ITrYtUY+O+ZhtlqUnD8+KwNPOyugEhfP9fdUIaEQ=
 github.com/alecthomas/template v0.0.0-20160405071501-a0175ee3bccc/go.mod h1:LOuyumcjzFXgccqObfd/Ljyb9UuFJ6TxHnclSeseNhc=
 github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751/go.mod h1:LOuyumcjzFXgccqObfd/Ljyb9UuFJ6TxHnclSeseNhc=
@@ -415,8 +417,6 @@ github.com/dustin/go-humanize v1.0.1/go.mod h1:Mu1zIs6XwVuF/gI1OepvI0qD18qycQx+m
 github.com/dvyukov/go-fuzz v0.0.0-20210914135545-4980593459a1/go.mod h1:11Gm+ccJnvAhCNLlf5+cS9KjtbaD5I5zaZpFMsTHWTw=
 github.com/e2b-dev/go-nfs v0.0.0-20260318224420-f59b77ca8555 h1:cUImAB4xt5kewJV1cd0gJMspWgmuMgbpj3Q8qIOUWrY=
 github.com/e2b-dev/go-nfs v0.0.0-20260318224420-f59b77ca8555/go.mod h1:VhNccO67Oug787VNXcyx9JDI3ZoSpqoKMT/lWMhUIDg=
-github.com/e2b-dev/roaring/v2 v2.16.1-0.20260410073213-00ee8aef09f9 h1:RftDqjb6RbXf5suqa7SrPYkiFPMHoutEY7mi1Yu31o0=
-github.com/e2b-dev/roaring/v2 v2.16.1-0.20260410073213-00ee8aef09f9/go.mod h1:eq4wdNXxtJIS/oikeCzdX1rBzek7ANzbth041hrU8Q4=
 github.com/ebitengine/purego v0.9.0 h1:mh0zpKBIXDceC63hpvPuGLiJ8ZAa3DfrFTudmfi8A4k=
 github.com/ebitengine/purego v0.9.0/go.mod h1:iIjxzd6CiRiOG0UyXP+V1+jWqUXVjPKLAI0mRfJZTmQ=
 github.com/edsrzf/mmap-go v1.2.0 h1:hXLYlkbaPzt1SaQk+anYwKSRNhufIDCchSPkUD6dD84=
--- a/packages/shared/go.mod
+++ b/packages/shared/go.mod
@@ -4,13 +4,11 @@ go 1.25.4
 
 tool github.com/go-swagger/go-swagger/cmd/swagger
 
-replace github.com/RoaringBitmap/roaring/v2 => github.com/e2b-dev/roaring/v2 v2.16.1-0.20260410073213-00ee8aef09f9
-
 require (
 	cloud.google.com/go/artifactregistry v1.17.1
 	cloud.google.com/go/storage v1.59.2
 	connectrpc.com/connect v1.18.1
-	github.com/RoaringBitmap/roaring/v2 v2.16.0
+	github.com/RoaringBitmap/roaring/v2 v2.16.1
 	github.com/aws/aws-sdk-go-v2 v1.41.0
 	github.com/aws/aws-sdk-go-v2/config v1.32.6
 	github.com/aws/aws-sdk-go-v2/feature/s3/manager v1.17.74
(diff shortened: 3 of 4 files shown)


packages/orchestrator/pkg/sandbox/block/cache.go
```diff
@@ -291,16 +327,19 @@ func (c *Cache) WriteAtWithoutLock(b []byte, off int64) (int, error) {
 	return n, nil
 }
 
-// dirtySortedKeys returns a sorted list of dirty keys.
-// Key represents a block offset.
```
dobrac: lets keep it

packages/orchestrator/pkg/sandbox/block/cache.go
```diff
@@ -245,30 +248,63 @@ func (c *Cache) Slice(off, length int64) ([]byte, error) {
...
+func (c *Cache) isBlockCached(i int64) bool {
+	if i < 0 || i >= int64(len(c.dirty))*64 {
+		return false
+	}
+
+	return c.dirty[i/64].Load()&(1<<uint(i%64)) != 0
+}
+
 func (c *Cache) isCached(off, length int64) bool {
-	// Make sure the offset is within the cache size
```
dobrac, line 344: please keep

packages/orchestrator/pkg/sandbox/block/cache.go
```diff
@@ -245,30 +248,63 @@ func (c *Cache) Slice(off, length int64) ([]byte, error) {
...
 	if off >= c.size {
 		return false
 	}
 
-	// Cap if the length goes beyond the cache size, so we don't check for blocks that are out of bounds.
```
dobrac, line 349: keep

packages/orchestrator/pkg/sandbox/block/cache.go
```diff
@@ -245,30 +248,63 @@ func (c *Cache) Slice(off, length int64) ([]byte, error) {
 	return nil, BytesNotAvailableError{}
 }
 
+func (c *Cache) isBlockCached(i int64) bool {
```
dobrac: what does the i represent here? Can you rename the argument variable?

packages/orchestrator/pkg/sandbox/block/cache.go
```diff
@@ -49,7 +49,7 @@ type Cache struct {
 	blockSize int64
 	mmap      *mmap.MMap
 	mu        sync.RWMutex
-	dirty     sync.Map
+	dirty     []atomic.Uint64 // bitset indexed by off/blockSize — bit is set when block is present
```
dobrac: Have you considered using the bits-and-blooms/bitset instead?

levb: bits-and-blooms/bitset is not atomic, would require a mutex which then defeats the purpose of this PR. There is an atomic fork, but does not appear to be very used. I think the code here is custom-fit for the problem, well-localized, and clear in the context of the methods. If/once we do have other uses that are fit for it, then we migrate it to its own utility?

dobrac: Ok. Lets then move it to its own utility right away, there is almost none added cost doing it already

packages/shared/pkg/atomicbitset/bitset.go
```diff
@@ -0,0 +1,115 @@
...
+//
+// A Bitset must not be copied after first use (copies share the
+// underlying array).
+type Bitset struct {
+	words []atomic.Uint64
+	n     uint
+}
+
+// New returns a Bitset with capacity for n bits, all initially zero.
+func New(n uint) Bitset {
```
levb: I am at most 1/5 on `uint` here (and for indexing in the set in all methods). It's clean for the interface. The callers use `int64`,  which would be awkward for the bitset index IMO. 2/5 we should do a separate cleanup PR where we assume int == int64 and use `int`s everywhere except `uint`s where absolutely needed.

packages/orchestrator/pkg/sandbox/block/cache.go
```diff
@@ -245,31 +248,30 @@ func (c *Cache) Slice(off, length int64) ([]byte, error) {
 	return nil, BytesNotAvailableError{}
 }
 
+func (c *Cache) startBlock(off int64) uint {
+	return uint(off / c.blockSize)
+}
```
dobrac: lets use the `BlockIdx` in the implementation of startBlock

packages/orchestrator/pkg/sandbox/block/cache.go
```diff
@@ -87,12 +87,15 @@ func NewCache(size, blockSize int64, filePath string, dirtyFile bool) (*Cache, e
 		return nil, fmt.Errorf("error mapping file: %w", err)
 	}
 
+	numBlocks := (size + blockSize - 1) / blockSize
```
dobrac: `TotalBlocks` can be used instead

packages/orchestrator/pkg/sandbox/block/cache.go
dobrac: this comment is still missing

packages/orchestrator/pkg/sandbox/block/cache.go
dobrac: this comment is still missing

packages/orchestrator/pkg/sandbox/block/cache.go
```diff
@@ -49,7 +49,7 @@ type Cache struct {
 	blockSize int64
 	mmap      *mmap.MMap
 	mu        sync.RWMutex
-	dirty     sync.Map
+	dirty     atomicbitset.Bitset
```
dobrac: if we have a 100GB disk and we set a dirty block at the end, won't we use like 400KB of memory just for storing that one bit?

packages/shared/pkg/atomicbitset/bitset.go
dobrac: For this case, I think we should be explicit with uint64. Every other value than uint64 will be incorrectly handled in this code part as the `64` length is directly used in the operations.

packages/shared/pkg/atomicbitset/bitset.go
```diff
@@ -0,0 +1,115 @@
...
+		words: make([]atomic.Uint64, (n+63)/64),
+		n:     n,
+	}
+}
+
+// Len returns the capacity in bits.
+func (b *Bitset) Len() uint { return b.n }
+
+// Has reports whether bit i is set. Out-of-range returns false.
+func (b *Bitset) Has(i uint) bool {
```
dobrac: can we make all the functions implementations more readable so its clear what we're doing?

Something like for this one for example:

```
func (b *Bitset) Has(i uint) bool {
	if i >= b.n {
		return false
	}

	wordIndex := i / 64
	bitIndex  := i % 64

	// selector mask (a 1 at the specific bit position)
    mask := 1 << bitIndex

	wordValue := b.words[wordIndex].Load()
	return (wordValue & mask) != 0
}
```

packages/orchestrator/pkg/sandbox/block/cache.go
ValentaTomas: Won't this be 3.125MiB?

```
((100*1024^3)/4096/8)/1024/1024
```

ValentaTomas: To be fair here though, this is also a problem with the bitset we use now. Only the roaring bitmaps solve this, but they don't support the iteration/lookup we would ideally need, so for pause proccessing I would still like to use the current bitset (or implement similarly effective iterator over the bitmaps).

One other node—with the current NBD the cache map might not be under much contention (but might be relevant if we use that for the other caches).

dobrac: > Won't this be 3.125MiB?

divided by 64 as you have 64 bits (chunks) for one array slot

ValentaTomas: 👌 My bad (not bad in the end).

packages/shared/pkg/atomicbitset/bitset.go
levb: We do not _need_ 64 bit values for how many blocks there are, do we? The normal roaring bitset supports only 32-bit indices for addressing the bits. There is a 64 bit version, but less mature. And using a uint64 would still cause an ugly cast at the caller?

packages/orchestrator/pkg/sandbox/block/cache.go
levb: So I added 2 implementations now - roaring (by far the most compact, and plenty fast on its own, single threaded), and a sharded atomic uint64 array which is still quite compact and much (100x) faster when concurrency is involved.

dobrac: The roaring implementation looks great. Unless we really need the the speed optimization, I'm for opting for now for the roaring approach only, leaving the sharded atomic optimization outside of this PR

packages/shared/pkg/atomicbitset/bitset.go
dobrac: I thing it's okay to have the ~8TiB limit for now (for 4KiB blocks), I don't expect we would approach it any time soon

ValentaTomas:
For now, using roaring fork with a fix. After merging https://github.com/RoaringBitmap/roaring/pull/521 we can switch to the original lib.

levb:
👍

ValentaTomas (approved):

PR #2236, opened 2026-03-26
perf(block): precompute OTEL for chunker hot paths

levb:
Replace per-call `attribute.String(...)` allocation in every `Slice`/`ReadAt`/`runFetch` with precomputed `metric.MeasurementOption` values built once at package init.

- `Stopwatch.end`: build one `attribute.NewSet` and reuse across all three instrument calls (histogram, sum, count) instead of three separate `metric.WithAttributes(kv...)` allocations.
- New `PrecomputeAttrs(kv...)`: builds a reusable `MeasurementOption`.
- New `Stopwatch.Record(ctx, total, precomputedAttrs)`: zero per-call attribute allocation alternative to `Success`/`Failure`.
- Exported `Success`/`Failure` attribute vars for use with `PrecomputeAttrs`.

- Add `precomputedAttrs` struct with all Slice/fetch attribute combos.
- Package-level `chunkerAttrs` var built once at init.
- All `timer.Success(ctx, n, attribute.String(...))` calls replaced with `timer.Record(ctx, n, a.successFromCache)` etc.

FullFetchChunker.Slice on cache hit (64 MiB cache, 4K blocks, 4 MiB chunks):

```
                         │    baseline     │         precomputed attrs          │
                         │     sec/op      │   sec/op     vs base              │
ChunkerSlice_CacheHit-16     20.57µ ± 2%   19.85µ ± 3%  -3.49% (p=0.009 n=6)

                         │    baseline     │         precomputed attrs              │
                         │      B/op       │     B/op      vs base                  │
ChunkerSlice_CacheHit-16    9.211Ki ± 0%   8.047Ki ± 0%  -12.64% (p=0.002 n=6)

                         │    baseline     │        precomputed attrs             │
                         │   allocs/op     │ allocs/op   vs base                  │
ChunkerSlice_CacheHit-16     16.000 ± 0%    4.000 ± 0%  -75.00% (p=0.002 n=6)
```

75% fewer allocations (16 → 4) and 13% less memory per Slice call on the NBD page fault hot path.

### DIFF
--- a/packages/orchestrator/pkg/sandbox/block/chunk.go
+++ b/packages/orchestrator/pkg/sandbox/block/chunk.go
@@ -182,36 +182,35 @@ func (c *FullFetchChunker) WriteTo(ctx context.Context, w io.Writer) (int64, err
 
 func (c *FullFetchChunker) Slice(ctx context.Context, off, length int64) ([]byte, error) {
 	timer := c.metrics.SlicesTimerFactory.Begin()
-	a := chunkerAttrs
 
 	b, err := c.cache.Slice(off, length)
 	if err == nil {
-		timer.Record(ctx, length, a.successFromCache)
+		timer.RecordRaw(ctx, length, chunkerAttrs.successFromCache)
 
 		return b, nil
 	}
 
 	if !errors.As(err, &BytesNotAvailableError{}) {
-		timer.Record(ctx, length, a.failCacheRead)
+		timer.RecordRaw(ctx, length, chunkerAttrs.failCacheRead)
 
 		return nil, fmt.Errorf("failed read from cache at offset %d: %w", off, err)
 	}
 
 	chunkErr := c.fetchToCache(ctx, off, length)
 	if chunkErr != nil {
-		timer.Record(ctx, length, a.failRemoteFetch)
+		timer.RecordRaw(ctx, length, chunkerAttrs.failRemoteFetch)
 
 		return nil, fmt.Errorf("failed to ensure data at %d-%d: %w", off, off+length, chunkErr)
 	}
 
 	b, cacheErr := c.cache.Slice(off, length)
 	if cacheErr != nil {
-		timer.Record(ctx, length, a.failLocalReadAgain)
+		timer.RecordRaw(ctx, length, chunkerAttrs.failLocalReadAgain)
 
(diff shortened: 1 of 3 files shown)


packages/orchestrator/pkg/sandbox/block/chunk.go
```diff
@@ -122,43 +182,36 @@ func (c *FullFetchChunker) WriteTo(ctx context.Context, w io.Writer) (int64, err
 
 func (c *FullFetchChunker) Slice(ctx context.Context, off, length int64) ([]byte, error) {
 	timer := c.metrics.SlicesTimerFactory.Begin()
+	a := chunkerAttrs
```
dobrac: do we need to shadow the variable? (applies to all occurences)

packages/shared/pkg/telemetry/meters.go
```diff
@@ -424,7 +430,24 @@ func (t Stopwatch) end(ctx context.Context, result string, total int64, kv ...at
...
+// key-values. The option must include all attributes (including "result").
+// Use with Stopwatch.Record to avoid per-call attribute allocation.
+func PrecomputeAttrs(kv ...attribute.KeyValue) metric.MeasurementOption {
+	return metric.WithAttributeSet(attribute.NewSet(kv...))
+}
+
+// Record records an operation using a precomputed attribute option.
+// Zero-allocation alternative to Success/Failure for hot paths.
+func (t Stopwatch) Record(ctx context.Context, total int64, precomputedAttrs metric.MeasurementOption) {
+	amount := time.Since(t.start).Milliseconds()
```
dobrac, line 446: can we use here the .end directly?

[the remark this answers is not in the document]
dobrac: Can you please explain why is this not valid?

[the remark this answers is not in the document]
levb: The function does exactly what it is designed to do, record the precompiled full set, avoiding allocations. Renamed to `RecordRaw` to avoid confusion.

packages/orchestrator/pkg/sandbox/block/chunk.go
levb: Was just an alias for readability, removed.

packages/shared/pkg/telemetry/meters.go
levb: the other way around, `end()` can use `RecordRaw` once it has the opt, done.

PR #2246, opened 2026-03-27
feat(storage): add compression primitives: FrameTable, V4 header

levb:
## Summary

  Foundational types and utilities for compressed template storage, extracted as an independent PR ahead of the full compression read/write path.

  ### Plugged in (affects runtime now)
  - `MergeMappings` now returns `error` (propagates `FrameTable.Subset` errors through `toDiffMapping` → `ToDiffHeader`)
  - `NewMultipartUploaderWithRetryConfig` accepts optional `metadata` param (existing caller passes `nil`)
  - `BuildMap.FrameTable` field added (nil for all current V3 headers — no behavior change)
  - `header.Metadata` gains `MetadataVersionCompressed` constant
  - `show-build-diff` CLI updated for new `MergeMappings` signature

  ### Preparation only (no callers yet)
  - **FrameTable**: tracks per-frame compressed/uncompressed sizes, supports `Subset`, `Range`, `FrameFor`, `GetFetchRange` for offset translation
  - **Codec pools**: LZ4/zstd compressor pools with `CompressLZ4`/`DecompressLZ4` helpers
  - **Compressed upload pipeline**: `compressStream` (read → parallel compress → ordered upload), `partUploader` interface, `CompressBytes`
  - **CompressConfig**: env-var and LaunchDarkly-based configuration with `ResolveCompressConfig`
  - **V4 header format**: `SerializeHeader`/`DeserializeBytes` with FrameTable serialization, `BuildFileInfo`, `ValidateHeader`
  - **Template path helpers**: `CompressedDataPath`, `CompressedDataName`, `BaseFileName`
  - **Feature flags**: `CompressConfigFlag`, `CompressFileTypeKind`/`CompressUseCaseKind` context kinds
  - **Multipart upload**: `uploadPartSlices` (zero-copy multi-slice upload), `Start`/`UploadPart`/`Complete` methods on `MultipartUploader`


packages/shared/pkg/storage/header/serialization.go
```diff
@@ -47,24 +71,272 @@ func (m *Metadata) NextGeneration(buildID uuid.UUID) *Metadata {
...
+
+	if h.Metadata.Version <= 3 {
+		return raw, nil
+	}
+
+	// V4: keep Metadata prefix raw, then [uint32 uncompressed size] + [LZ4 block].
+	block := raw[metadataSize:]
+	compressed, err := storage.CompressLZ4(block)
+	if err != nil {
+		return nil, fmt.Errorf("failed to LZ4-compress v4 header mappings: %w", err)
```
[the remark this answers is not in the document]
levb: restored the streaming compression/decompression that deals with it

packages/shared/pkg/storage/compress_upload.go
```diff
@@ -0,0 +1,280 @@
...
+			part.addFrame(compressCtx, buf[:n], borrow, release)
+		}
+
+		if err != nil {
+			break
+		}
+
+		if part.compressedSize.Load() >= targetPartSize {
+			part.submit(ctx, q)
+			part, compressCtx = newPart(part.index+1, ctx, cfg.FrameEncodeWorkers)
```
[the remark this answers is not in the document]
levb: Only in-flight compressing frames can be added, other frames will go to the next part, this is by design

packages/shared/pkg/storage/compress_pool.go
```diff
@@ -0,0 +1,169 @@
...
+
+func (l *lz4FrameCompressor) Compress(src []byte) ([]byte, error) {
+	dst := make([]byte, lz4.CompressBlockBound(len(src)))
+
+	n, err := lz4.CompressBlock(src, dst, nil)
+	if err != nil {
+		return nil, fmt.Errorf("lz4 block compress: %w", err)
+	}
+
+	if n == 0 {
```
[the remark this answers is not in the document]
levb: Reverted to using the LZ4 stream mode that deals with this

packages/shared/pkg/storage/compress_pool.go
```diff
@@ -0,0 +1,169 @@
...
+		return nil, fmt.Errorf("lz4 block compress: incompressible data (%d bytes)", len(src))
+	}
+
+	return dst[:n], nil
+}
```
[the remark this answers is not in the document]
levb: Reverted to the streaming mode

packages/shared/pkg/storage/gcp_multipart.go
```diff
@@ -232,6 +289,60 @@ func (m *MultipartUploader) uploadPart(ctx context.Context, uploadID string, par
 	return etag, nil
 }
 
+// uploadPartSlices uploads a part from multiple byte slices without concatenating them.
+// It computes MD5 by hashing each slice and uses a ReaderFunc for retryable reads.
+func (m *MultipartUploader) uploadPartSlices(ctx context.Context, uploadID string, partNumber int, slices [][]byte) (string, error) {
```
dobrac, line 294: can we add tests for this?

packages/shared/pkg/storage/compress_upload.go
```diff
@@ -0,0 +1,280 @@
...
+	"sync"
+	"sync/atomic"
+
+	"golang.org/x/sync/errgroup"
+)
+
+// MaxCompressedHeaderSize is the maximum allowed decompressed header size (64 MiB).
+// Headers are typically a few hundred KiB (e.g., 100 layers × 256 frames × 32 bytes/frame ≈ 800 KB).
+// This is a safety bound to prevent unbounded allocation from corrupt data.
+const MaxCompressedHeaderSize = 64 << 20
```
dobrac, line 20: this should be enforced on upload, not download

packages/shared/pkg/storage/compress_config.go
```diff
@@ -0,0 +1,131 @@
...
+// via ResolveCompressConfig.
+type CompressConfig struct {
+	Enabled            bool   `env:"COMPRESS_ENABLED"              envDefault:"false"`
+	Type               string `env:"COMPRESS_TYPE"                 envDefault:"zstd"`
+	Level              int    `env:"COMPRESS_LEVEL"                envDefault:"2"`
+	FrameSizeKB        int    `env:"COMPRESS_FRAME_SIZE_KB"        envDefault:"2048"`
+	TargetPartSizeMB   int    `env:"COMPRESS_TARGET_PART_SIZE_MB"  envDefault:"50"`
+	FrameEncodeWorkers int    `env:"COMPRESS_FRAME_ENCODE_WORKERS" envDefault:"4"`
+	EncoderConcurrency int    `env:"COMPRESS_ENCODER_CONCURRENCY"  envDefault:"1"`
+}
```
dobrac, line 21: isn't this defined by the feature flags already?

packages/shared/pkg/storage/header/serialization.go
```diff
@@ -25,6 +33,25 @@ type Metadata struct {
...
+	BuildId            uuid.UUID
+	BuildStorageOffset uint64
+}
+
+type v4SerializableBuildMap struct {
+	Offset                   uint64
+	Length                   uint64
+	BuildId                  uuid.UUID
+	BuildStorageOffset       uint64
+	CompressionTypeNumFrames uint64 // CompressionType is stored as uint8 in the high byte, the low 24 bits are NumFrames
```
dobrac, line 48: why are we doing this in one int?

packages/shared/pkg/storage/header/metadata.go
```diff
@@ -93,6 +103,9 @@ func (d *DiffMetadata) ToDiffHeader(
 		return nil, fmt.Errorf("failed to create header: %w", err)
 	}
 
+	// Inherit upstream build file info (sizes + checksums).
+	header.BuildFiles = maps.Clone(originalHeader.BuildFiles)
```
dobrac, line 107: why not passing the value directly to the constructor?

packages/shared/pkg/storage/header/mapping.go
```diff
@@ -25,9 +28,40 @@ func (mapping *BuildMap) Copy() *BuildMap {
...
+	}
+
+	mappedRange := storage.Range{
+		Start:  int64(mapping.BuildStorageOffset),
+		Length: int(mapping.Length),
+	}
+
+	subset, err := frameTable.Subset(mappedRange)
+	if err != nil {
+		return fmt.Errorf("mapping at virtual offset %#x (storage offset %#x, length %#x): %w",
```
dobrac, line 56: why are we using hex for the offset and length?

packages/shared/pkg/storage/header/mapping.go
```diff
@@ -25,9 +28,40 @@ func (mapping *BuildMap) Copy() *BuildMap {
...
+// AddFrames associates compression frame information with this mapping.
+//
+// When a file is uploaded with compression, the compressor produces a FrameTable
+// that describes how the compressed data is organized into frames. This method
+// computes which compressed frames cover this mapping's data within the build's
+// storage file based on BuildStorageOffset and Length.
+//
+// Returns nil if frameTable is nil. Returns an error if the mapping's range
+// cannot be found in the frame table.
+func (mapping *BuildMap) AddFrames(frameTable *storage.FrameTable) error {
```
dobrac, line 44: this does replace/set rather than add, correct?

packages/shared/pkg/storage/header/mapping.go
```diff
@@ -160,6 +195,10 @@ func MergeMappings(
 					// the build storage offset is the same as the base mapping
 					BuildStorageOffset: base.BuildStorageOffset,
 				}
+				leftBase.FrameTable, err = base.FrameTable.Subset(storage.Range{Start: int64(leftBase.BuildStorageOffset), Length: int(leftBase.Length)})
```
dobrac, line 198: don't we have already the AddFrames function for this? (and all other occurrences)

packages/shared/pkg/storage/header/mapping.go
```diff
@@ -17,6 +19,7 @@ type BuildMap struct {
 	Length             uint64
 	BuildId            uuid.UUID
 	BuildStorageOffset uint64
+	FrameTable         *storage.FrameTable
```
dobrac, line 22: why do we have the frametable as pointer?

packages/shared/pkg/storage/header/header.go
```diff
@@ -1,26 +1,62 @@
...
 	startMap    map[int64]*BuildMap
 
 	Mapping []*BuildMap
 }
 
+// CloneForUpload returns a clone with copied Mapping and BuildFiles, safe to
+// mutate for serialization without racing with concurrent readers of the
+// original. Only serialization-relevant fields are populated (Metadata,
+// Mapping, BuildFiles); lookup indices (blockStarts, startMap) are left nil.
+func (t *Header) CloneForUpload() *Header {
```
dobrac, line 40: why is it "CloneForUpload" and not full "Clone"?

dobrac: also, this function is unused so its hard to evaluate its purpose

packages/shared/pkg/storage/header/serialization.go
```diff
@@ -47,24 +74,283 @@ func (m *Metadata) NextGeneration(buildID uuid.UUID) *Metadata {
...
+
+func serialize(metadata *Metadata, buildFiles map[uuid.UUID]BuildFileInfo, mappings []*BuildMap) ([]byte, error) {
 	var buf bytes.Buffer
 
 	err := binary.Write(&buf, binary.LittleEndian, metadata)
 	if err != nil {
 		return nil, fmt.Errorf("failed to write metadata: %w", err)
 	}
 
+	if metadata.Version >= 4 {
```
dobrac, line 92: can we do the 3/4 version logic in a way that its easier to understand the differences between them from the code?

There are currently lot of if/else statements scattered all over the methods

packages/shared/pkg/storage/template.go
```diff
@@ -53,10 +53,52 @@ func (t TemplateFiles) StorageMetadataPath() string {
...
+// Example: "{buildId}/memfile.zstd"
+func (t TemplateFiles) CompressedDataPath(fileName string, ct CompressionType) string {
+	return fmt.Sprintf("%s/%s", t.StorageDir(), CompressedDataName(fileName, ct))
+}
+
+// CompressedPath transforms a base object path (e.g. "buildId/memfile") into
+// the compressed data path (e.g. "buildId/memfile.zstd").
+func CompressedPath(basePath string, ct CompressionType) string {
+	return basePath + ct.Suffix()
+}
```
dobrac, line 81: I'm not a big fan even of the original methods. I would like to not extend them if its not necessary.

The `DataPath` looks like a backdoor for not having all the file names defined in constants.

packages/shared/pkg/storage/header/serialization.go
levb: old habit - didn't want to shift the alignment and didn't want to use more than I needed, will break into a separate byte + uint64.

packages/shared/pkg/storage/compress_config.go
levb: we need to override the FF settings for the integration tests, and also to configure the service in the future without FF

packages/shared/pkg/storage/compress_upload.go
levb: 0/5 just take it out? there is no constraint limit in the system other than OOM?

packages/shared/pkg/storage/header/serialization.go
levb: 👍

packages/shared/pkg/storage/header/mapping.go
levb: 1/5 Because it is so much easier to read? (rounded numbers, usually). Personally, 4/5 I'd prefer to have sizes consistently printed in hex everywhere. I had it at some point but reverted the changes as unrelated to save on # of changed lines.

packages/shared/pkg/storage/header/mapping.go
levb: This is called in `PendingBuildInfo.applyToHeader` on a mapping that knows its position but doesn't have any mappings yet, so it needs to extract them from the full FT (returned from an upload). A better name would have been perhaps, `InitFrames` or `ExtractFrames`?

A bigger issue though - `PendingBuildInfo.applyToHeader` is O(N2) ATM, doesn't need to be since both mappings and the master FT are sorted.

packages/shared/pkg/storage/header/mapping.go
levb: Because it is not always present (not in V3), one 64 bit word instead of 4.

packages/shared/pkg/storage/header/header.go
levb: It is used (in the next PR) for mutating the headers; as the comment says it's not a full `Clone` because it skips irrelevant fields.

packages/shared/pkg/storage/template.go
levb: I 100% agree - not a fan; but I've been trying to reduce the ## lines changed, even kept (brought back) the **unused** diff type param to save a couple dozen changed lines. Re-working the file type/path management to something elegant (as well as in general, extracting more commonalities between memfile and rootfs without overly abstracting them). 

2/5 Not in this PR?

PR #2247, opened 2026-03-28
feat(dashboard-api): supabase auth users sync background runner

ben-fornefeld:
**Motivation**

We're migrating our Database and need to de-couple current database triggers between auth and core schema. This PR allows having a sync between these two schema, that supports separate db configurations respectively.

### DIFF
--- a/packages/dashboard-api/main.go
+++ b/packages/dashboard-api/main.go
@@ -198,7 +198,7 @@ func run() int {
 
 		riverClient, err = backgroundworker.StartAuthUserSyncWorker(
 			ctx,
-			ctx,
+			signalCtx,
 			supabaseDB,
 			db,
 			l,


packages/db/pkg/auth/sql_queries/supabase_auth_user_sync/get_auth_user.sql
```diff
@@ -0,0 +1,4 @@
+-- name: GetAuthUserByID :one
+SELECT id, email
+FROM auth.users
+WHERE id = sqlc.arg(user_id)::uuid;
```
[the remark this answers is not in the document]
ben-fornefeld: fine, we require emails for auth

packages/db/pkg/auth/migrations/20260328000001_dashboard_supabase_auth_user_sync_queue.sql
```diff
@@ -0,0 +1,156 @@
+-- +goose Up
+-- +goose StatementBegin
+CREATE TABLE public.user_sync_queue (
```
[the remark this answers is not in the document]
ben-fornefeld: we want to execute this manually + it should not be tied to main db migrations since authDB will be separate going forward

dobrac: lets put it in the auth schema if we can. If not, then lets create a custom schema for this

ben-fornefeld: custom schema then

packages/dashboard-api/internal/backgroundworker/auth_user_sync.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+func NewAuthUserSyncWorker(mainDB *sqlcdb.Client, l logger.Logger) *AuthUserSyncWorker {
+	jobsCounter, err := otel.Meter("github.com/e2b-dev/infra/packages/dashboard-api/internal/backgroundworker")
+		"dashboard_api.auth_user_sync.jobs_total",
+		metric.WithDescription("Total auth user sync jobs by operation and result."),
+		metric.WithUnit("{job}"),
+	)
+	if err != nil {
+		l.Warn(context.Background(), "failed to initialize auth user sync metric", zap.Error(err))
```
dobrac: lets pass the context as a parameter

packages/db/pkg/auth/migrations/20260401000001_river_auth_custom_schema.sql
```diff
@@ -0,0 +1,13 @@
...
+-- +goose StatementBegin
+
+CREATE SCHEMA IF NOT EXISTS auth_custom;
+
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
+
+/* We don't want to drop the schema, as it is used by other services. */
```
dobrac: why not drop the schema here if the dashboard-api is the source of truth?

packages/db/pkg/auth/queries/ack.sql.go
dobrac, line 1: what are these files generated from?

packages/dashboard-api/Makefile
```diff
@@ -33,12 +37,14 @@ build-and-upload:
...
-	./bin/dashboard-api
+	@EXTRA_ENV=$(DASHBOARD_API_EXTRA_ENV); \
+	eval "env $$EXTRA_ENV ./bin/dashboard-api"
 
 .PHONY: run-local
 run-local:
 	make build
-	NODE_ID=$(HOSTNAME) ./bin/dashboard-api
+	@EXTRA_ENV=$(DASHBOARD_API_EXTRA_ENV); \
+	eval "env NODE_ID=$(HOSTNAME) $$EXTRA_ENV ./bin/dashboard-api"
```
dobrac: is the usage of eval safe (secure) here? Can we avoid it?

packages/dashboard-api/Makefile
```diff
@@ -13,6 +13,10 @@ endif
 HOSTNAME := $(shell hostname 2> /dev/null || hostnamectl hostname 2> /dev/null)
 $(if $(HOSTNAME),,$(error Failed to determine hostname: both 'hostname' and 'hostnamectl' failed))
 
+define DASHBOARD_API_EXTRA_ENV
```
dobrac: I think it might have been simpler to expose the env variable directly rather than having it nested under extra env

packages/dashboard-api/main.go
```diff
@@ -229,13 +232,46 @@ func run() int {
 
 	wg := sync.WaitGroup{}
 
+	var riverClient *river.Client[pgx.Tx]
+
+	if config.SupabaseAuthUserSyncEnabled {
+		workerLogger := l.With(zap.String("worker", "supabase_auth_user_sync"))
+
+		authPool := authDB.WritePool()
```
dobrac: the authDB is not Supabase DB, but a preparation for a future split of public.users, public.teams, etc to its own database

packages/db/pkg/auth/migrations/20260401000001_river_auth_custom_schema.sql
ben-fornefeld: belt has a migration that creates this schema. we don't want to delete that accidentally but i guess we can add the down migration here and just execute the migration manually

packages/dashboard-api/Makefile
ben-fornefeld: i had a conversation with @djeebus on this but, i think this way it's much simpler adding environment variables to our dev setups, without the need to put it inside secrets or hook up a new makefile -> terraform -> nomad code path

packages/dashboard-api/main.go
ben-fornefeld: is this about the naming of the config?

packages/db/pkg/auth/queries/ack.sql.go
ben-fornefeld: good catch. these were artifacts that the generator did not clean up manually

packages/dashboard-api/Makefile
ben-fornefeld: i think it was fine but i cleaned it up to not rely on eval

packages/db/pkg/supabase/migrations/20260401000003_river_auth_user_sync_triggers.sql
```diff
@@ -0,0 +1,99 @@
...
+AS $$
+BEGIN
+    INSERT INTO auth_custom.river_job (args, kind, max_attempts, queue, state)
+    VALUES (
+        jsonb_build_object('user_id', NEW.id, 'operation', 'upsert', 'email', NEW.email),
+        'auth_user_sync',
+        20,
+        'auth_sync',
+        'available'
+    );
```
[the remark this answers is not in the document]
ben-fornefeld: the river migrator creates the necessary tables inside dashboard-api. we have to:
- ensure the auth_custom schema exists
- deploy the dashboard-api
- run the migration

.env.gcp.template
```diff
@@ -77,6 +77,10 @@ CLICKHOUSE_CLUSTER_SIZE=1
 
 # Dashboard API instance count (default: 0)
 DASHBOARD_API_COUNT=
+# Additional non-reserved dashboard-api env vars passed directly to the Nomad job (default: {})
+# Reserved keys managed by the module cannot be overridden here.
+# Example: '{"SUPABASE_AUTH_USER_SYNC_ENABLED":"true"}'
```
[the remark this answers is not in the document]
ben-fornefeld: @cursoragent fix please

iac/modules/job-dashboard-api/jobs/dashboard-api.hcl
```diff
@@ -71,20 +71,9 @@ job "dashboard-api" {
...
-        CLICKHOUSE_CONNECTION_STRING           = "${clickhouse_connection_string}"
-        SUPABASE_JWT_SECRETS                   = "${supabase_jwt_secrets}"
-        REDIS_URL                              = "${redis_url}"
-        REDIS_CLUSTER_URL                      = "${redis_cluster_url}"
-        REDIS_TLS_CA_BASE64                    = "${redis_tls_ca_base64}"
-        OTEL_COLLECTOR_GRPC_ENDPOINT           = "${otel_collector_grpc_endpoint}"
-        LOGS_COLLECTOR_ADDRESS                 = "${logs_collector_address}"
+        %{ for key in sort(keys(env)) ~}
+        ${key} = "${env[key]}"
+        %{ endfor ~}
```
djeebus: You don't need to sort it manually; terraform does it for you:


```suggestion
        %{ for key, value in env ~}
        ${key} = "${value}"
        %{ endfor ~}
```

iac/modules/job-dashboard-api/main.tf
```diff
@@ -1,27 +1,54 @@
...
+
+  extra_env = {
+    for key, value in var.extra_env : key => value
+    if value != null && trimspace(value) != ""
+  }
+
+  conflicting_extra_env_keys = sort(tolist(setintersection(
+    toset(keys(local.base_env)),
+    toset(keys(local.extra_env)),
+  )))
```
djeebus: This is kind of an anti-pattern. If someone sets the values above, it's b/c they wanted to override it. We shouldn't forbid it.

iac/modules/job-dashboard-api/main.tf
```diff
@@ -1,27 +1,54 @@
...
+    REDIS_CLUSTER_URL                      = var.redis_cluster_url
+    REDIS_TLS_CA_BASE64                    = var.redis_tls_ca_base64
+    OTEL_COLLECTOR_GRPC_ENDPOINT           = "localhost:${var.otel_collector_grpc_port}"
+    LOGS_COLLECTOR_ADDRESS                 = "http://localhost:${var.logs_proxy_port.port}"
+  }
+
+  extra_env = {
+    for key, value in var.extra_env : key => value
+    if value != null && trimspace(value) != ""
+  }
```
djeebus: Why? Empty strings are valid env vars, and could be distinct from populated values.

packages/dashboard-api/internal/backgroundworker/config.go
```diff
@@ -0,0 +1,14 @@
...
+	// must match the schema used in packages/db/pkg/auth/migrations for River tables and triggers
+	AuthCustomSchema = "auth_custom"
+
+	// must match the queue value in packages/db/pkg/auth/migrations trigger SQL inserts
+	AuthUserProjectionQueue = "auth_user_projection"
+
+	// must match the kind value in packages/db/pkg/auth/migrations trigger SQL inserts
+	AuthUserProjectionKind = "auth_user_projection"
+
+	AuthUserProjectionMaxWorkers = 10
```
djeebus: Let's unexport these. The constants don't add value to the one log message they're attached to. If we really want to log them out, add something like "client configured with <values>" in the NewRiverClient function. Keeping them private gives us more flexibility wrt refactors in the future.

packages/dashboard-api/internal/backgroundworker/auth_user_sync_test.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+func TestAuthUserSync_EndToEnd(t *testing.T) {
+	t.Parallel()
+
+	db := testutils.SetupDatabase(t)
+
+	authMigrationsDir := "packages/db/pkg/auth/migrations"
+
+	db.ApplyMigrationsUpTo(t, 20260401000001, authMigrationsDir)
```
djeebus: To ensure that future tests function correctly, I'd migrate to latest every time.

packages/dashboard-api/internal/backgroundworker/auth_user_sync_test.go
```diff
@@ -0,0 +1,256 @@
...
+
+	authPool := db.AuthDb.WritePool()
+	require.NoError(t, RunRiverMigrations(t.Context(), authPool))
+
+	db.ApplyMigrations(t, authMigrationsDir)
+
+	runUpsertProjection(t, db)
+	runDeleteProjection(t, db)
+	runBurstBacklog(t, db)
```
djeebus: This test scares me in general. I'd much rather have a 200 line test that makes clear setup steps, action steps, and verification steps (even if they're interleaved). It's really unclear what this test does, and why it does it.

packages/dashboard-api/internal/cfg/model.go
```diff
@@ -18,6 +18,8 @@ type Config struct {
 	RedisURL         string `env:"REDIS_URL"`
 	RedisClusterURL  string `env:"REDIS_CLUSTER_URL"`
 	RedisTLSCABase64 string `env:"REDIS_TLS_CA_BASE64"`
+
+	AuthUserSyncBackgroundWorkerEnabled bool `env:"AUTH_USER_SYNC_BACKGROUND_WORKER_ENABLED" envDefault:"false"`
```
djeebus: We should probably avoid the yoda-speak:


```suggestion
	EnableAuthUserSyncBackgroundWorker bool `env:"ENABLE_AUTH_USER_SYNC_BACKGROUND_WORKER" envDefault:"false"`
```

packages/dashboard-api/internal/backgroundworker/auth_user_sync.go
```diff
@@ -0,0 +1,136 @@
...
+		attribute.String("job.operation", job.Args.Operation),
+		attribute.Int64("job.id", job.ID),
+		telemetry.WithUserID(job.Args.UserID),
+	}
+	telemetry.ReportEvent(ctx, "auth_user_sync.job.started", attrs...)
+
+	userID, err := uuid.Parse(job.Args.UserID)
+	if err != nil {
+		telemetry.ReportError(ctx, "auth user sync parse user_id", err, attrs...)
+		w.observeJob(ctx, job.Args.Operation, "cancelled")
```
djeebus: maybe "validation-failed" or "invalid arguments" instead?

packages/dashboard-api/internal/backgroundworker/auth_user_sync.go
```diff
@@ -0,0 +1,136 @@
...
+			w.observeJob(ctx, job.Args.Operation, "error")
+
+			return fmt.Errorf("delete public.users %s: %w", userID, err)
+		}
+
+	case "upsert":
+		if job.Args.Email == "" {
+			err := fmt.Errorf("missing email in job args for user %s", userID)
+			telemetry.ReportError(ctx, "auth user sync missing email", err, attrs...)
+			w.observeJob(ctx, job.Args.Operation, "cancelled")
```
djeebus: "invalid argument"?

packages/dashboard-api/internal/backgroundworker/auth_user_sync.go
```diff
@@ -0,0 +1,136 @@
...
+			telemetry.ReportError(ctx, "auth user sync upsert public user", err, attrs...)
+			w.observeJob(ctx, job.Args.Operation, "error")
+
+			return fmt.Errorf("upsert public.users %s: %w", userID, err)
+		}
+
+	default:
+		err := fmt.Errorf("unknown operation %q for user %s", job.Args.Operation, userID)
+		telemetry.ReportError(ctx, "auth user sync unknown operation", err, attrs...)
+		w.observeJob(ctx, job.Args.Operation, "cancelled")
```
djeebus: "invalid argument"?

packages/dashboard-api/main.go
```diff
@@ -244,13 +247,47 @@ func run() int {
...
+		riverClient, err = backgroundworker.NewRiverClient(authPool, workers)
+		if err != nil {
+			l.Fatal(ctx, "failed to create River client", zap.Error(err))
+		}
+
+		if err := riverClient.Start(signalCtx); err != nil {
+			l.Fatal(ctx, "failed to start River client", zap.Error(err))
+		}
+
+		l.Info(ctx, "background worker started", zap.String("queue", backgroundworker.AuthUserProjectionQueue), zap.String("schema", backgroundworker.AuthCustomSchema))
```
djeebus: maybe cleaner to put this in it's own function: `setupRiverWorker` or something

packages/dashboard-api/Makefile
```diff
@@ -13,32 +13,35 @@ endif
 HOSTNAME := $(shell hostname 2> /dev/null || hostnamectl hostname 2> /dev/null)
 $(if $(HOSTNAME),,$(error Failed to determine hostname: both 'hostname' and 'hostnamectl' failed))
 
+define export_extra_env
+export $$(printf '%s' "$$DASHBOARD_API_ENV_VARS" | jq -r '(if .=="" then empty elif type=="string" then (fromjson? // empty) else . end) | to_entries? // [] | map("\(.key)=\(.value|tostring)") | .[]' 2>/dev/null) 2>/dev/null;
```
djeebus: Where are you defining these? You could just run `ENV=a ENV=bb make run` and it'll receive both those env vars

packages/db/pkg/testutils/db.go
```diff
@@ -104,17 +111,27 @@ func SetupDatabase(t *testing.T) *Database {
...
 
-// runDatabaseMigrations executes all required database migrations
-func runDatabaseMigrations(t *testing.T, connStr string) {
+	db.applyGooseMigrations(t, 0, migrationDirs...)
+}
+
+func (db *Database) ApplyMigrationsUpTo(t *testing.T, version int64, migrationDirs ...string) {
+	t.Helper()
+
+	db.applyGooseMigrations(t, version, migrationDirs...)
```
djeebus: I don't think we should ever be running tests on an old version of the database. These tests will be stale as soon as a new migration is created, then we'll be testing that new code works against old dbs, which servers no purpose.

iac/modules/job-dashboard-api/main.tf
```diff
@@ -1,27 +1,37 @@
...
+    CLICKHOUSE_CONNECTION_STRING           = var.clickhouse_connection_string
+    SUPABASE_JWT_SECRETS                   = var.supabase_jwt_secrets
+    REDIS_URL                              = var.redis_url
+    REDIS_CLUSTER_URL                      = var.redis_cluster_url
+    REDIS_TLS_CA_BASE64                    = var.redis_tls_ca_base64
+    OTEL_COLLECTOR_GRPC_ENDPOINT           = "localhost:${var.otel_collector_grpc_port}"
+    LOGS_COLLECTOR_ADDRESS                 = "http://localhost:${var.logs_proxy_port.port}"
+  }
+
+  env = merge(local.base_env, var.extra_env)
```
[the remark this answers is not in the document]
ben-fornefeld: expected

iac/provider-gcp/Makefile
```diff
@@ -75,6 +75,7 @@ tf_vars := \
 	$(call tfvar, LOKI_BOOT_DISK_TYPE) \
 	$(call tfvar, LOKI_USE_V13_SCHEMA_FROM) \
 	$(call tfvar, DASHBOARD_API_COUNT) \
+	$(call tfvar, DASHBOARD_API_ENV_VARS) \
```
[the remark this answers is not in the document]
ben-fornefeld: tested this locally, i don't think it's a real issue. the JSON survives intact.

packages/dashboard-api/internal/backgroundworker/auth_user_sync.go
```diff
@@ -0,0 +1,136 @@
...
+
+func (AuthUserSyncArgs) Kind() string { return authUserProjectionKind }
+
+type AuthUserSyncWorker struct {
+	river.WorkerDefaults[AuthUserSyncArgs]
+
+	mainDB      *sqlcdb.Client
+	l           logger.Logger
+	jobsCounter metric.Int64Counter
+}
```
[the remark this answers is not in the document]
djeebus: This would be a useful addition; make it clear what interface this struct is *supposed* to be implementing

packages/dashboard-api/internal/backgroundworker/auth_user_sync.go
```diff
@@ -0,0 +1,136 @@
...
+
+func NewAuthUserSyncWorker(ctx context.Context, mainDB *sqlcdb.Client, meter metric.Meter, l logger.Logger) *AuthUserSyncWorker {
+	jobsCounter, err := meter.Int64Counter(
+		"jobs_total",
+		metric.WithDescription("Total auth user sync jobs by operation and result."),
+		metric.WithUnit("{job}"),
+	)
```
djeebus: No need to pass in the meter, create a meter locally (globally) and then use it:


```suggestion

var meter = otel.Meter("github.com/e2b-dev/infra/packages/...")  // scripts/fix-meters.sh will fix this for you

func NewAuthUserSyncWorker(ctx context.Context, mainDB *sqlcdb.Client, l logger.Logger) *AuthUserSyncWorker {
	jobsCounter, err := meter.Int64Counter(
		"jobs_total",
		metric.WithDescription("Total auth user sync jobs by operation and result."),
		metric.WithUnit("{job}"),
	)
```

djeebus: In fact, the counter itself can be global as well:


```suggestion
var meter = otel.Meter("github.com/e2b-dev/infra/packages/...")  // scripts/fix-meters.sh will fix this for you

var jobsCounter = utils.Must(meter.Int64Counter(
		"jobs_total",
		metric.WithDescription("Total auth user sync jobs by operation and result."),
		metric.WithUnit("{job}"),
	))
func NewAuthUserSyncWorker(ctx context.Context, mainDB *sqlcdb.Client, l logger.Logger) *AuthUserSyncWorker {
```

packages/dashboard-api/internal/backgroundworker/auth_user_sync.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+}
+
+func (w *AuthUserSyncWorker) Work(ctx context.Context, job *river.Job[AuthUserSyncArgs]) error {
+	attrs := []attribute.KeyValue{
+		attribute.String("job.kind", authUserProjectionKind),
+		attribute.String("job.operation", job.Args.Operation),
+		attribute.Int64("job.id", job.ID),
+		telemetry.WithUserID(job.Args.UserID),
+	}
```
djeebus: Might be worth it to create an attribute set here with `attribute.NewSet`, not sure though.

djeebus: Actually, makes more sense to add it to a span


```suggestion
    ctx, span := tracer.Start(ctx, "begin auth user sync", trace.WithAttributes(
		attribute.String("job.kind", authUserProjectionKind),
		attribute.String("job.operation", job.Args.Operation),
		attribute.Int64("job.id", job.ID),
		telemetry.WithUserID(job.Args.UserID),
	))
	defer span.End()
```

Then stop sending them into other function calls, as that would now duplicate them.

djeebus (approved):

packages/db/pkg/supabase/migrations/20260401000001_river_auth_custom_schema.sql
```diff
@@ -0,0 +1,31 @@
...
+END;
+$$;
+
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
+
+DROP ROLE IF EXISTS trigger_user;
+DROP SCHEMA IF EXISTS auth_custom;
```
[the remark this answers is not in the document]
ben-fornefeld: desired, we have other services potentially depend on the schema and don't want to cascade

packages/dashboard-api/main.go
```diff
@@ -243,48 +314,76 @@ func run() int {
...
+		g.Go(func() error {
+			if err := riverClient.Stop(ctx); err != nil {
+				return fmt.Errorf("shutdown River client: %w", err)
+			}
+
+			return nil
+		})
+	}
+
+	return g.Wait()
```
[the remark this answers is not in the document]
ben-fornefeld: invalid, the channel returns nil either way

iac/modules/job-dashboard-api/main.tf
```diff
@@ -1,27 +1,38 @@
+locals {
+  base_env = {
+    GIN_MODE                               = "release"
+    ENVIRONMENT                            = var.environment
+    NODE_ID                                = "$${node.unique.id}"
+    PORT                                   = "$${NOMAD_PORT_api}"
```
dobrac: this should stay in the dashboard-api.hcl config

.env.gcp.template
```diff
@@ -77,6 +77,12 @@ CLICKHOUSE_CLUSTER_SIZE=1
 
 # Dashboard API instance count (default: 0)
 DASHBOARD_API_COUNT=
+# Dashboard API Supabase DB connection string (default: POSTGRES_CONNECTION_STRING)
+SUPABASE_DB_CONNECTION_STRING=
+# Additional dashboard-api env vars passed directly to the Nomad job (default: {})
+# Values here are merged into the job env and can override module defaults.
+# Example: '{"ENABLE_AUTH_USER_SYNC_BACKGROUND_WORKER":"true"}'
+DASHBOARD_API_ENV_VARS=
```
dobrac: can we remove this altogether and just use the newly added TF_vars env variables?

iac/provider-gcp/Makefile
```diff
@@ -76,6 +76,8 @@ tf_vars := \
...
```
dobrac: we should be able to not add this anymore now

packages/db/pkg/supabase/migrations/20260401000003_river_auth_user_sync_triggers.sql
```diff
@@ -0,0 +1,100 @@
...
+    INSERT INTO auth_custom.river_job (args, kind, max_attempts, queue, state)
+    VALUES (
+        jsonb_build_object('user_id', NEW.id, 'operation', 'upsert'),
+        'auth_user_projection',
+        20,
+        'auth_user_projection',
+        'available'
+    );
+
+    PERFORM pg_notify('auth_custom.river_insert', '{"queue":"auth_user_projection"}');
```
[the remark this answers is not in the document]
ben-fornefeld: not true.

packages/dashboard-api/Makefile
```diff
@@ -13,32 +13,35 @@ endif
...
+endef
```
dobrac: not needed anymore

PR #2248, opened 2026-03-28
[chore] Add debug logs sandbox catalog updates

matthewlouisbrockman:
Adds debug logging to the two ways we can update the sandbox catalog:
- the sandbox sync
-  normal sandbox lifecycle routes

This helps us understand what causes 404s when it looks like the sandbox should be present in the catalog but isn't.


packages/api/internal/sandbox/storage/memory/sync.go
```diff
@@ -37,6 +39,13 @@ func (s *Storage) Sync(sandboxes []sandbox.Sandbox, nodeID string) []sandbox.San
 
 		_, found := sandboxMap[data.SandboxID]
 		if !found {
```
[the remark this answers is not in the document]
matthewlouisbrockman: i just want the logs, not looking at changing behavior

packages/shared/pkg/sandbox-catalog/catalog_memory.go
```diff
@@ -49,6 +51,7 @@ func (c *MemorySandboxCatalog) StoreSandbox(ctx context.Context, sandboxID strin
 	defer c.mtx.Unlock()
```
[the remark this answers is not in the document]
matthewlouisbrockman: changed to debug, but we are still logging debug. easier to filter, but we have many more logs and need to know why sometimes a sandbox isn't in the catalog right now when it's supposed to be

packages/shared/pkg/sandbox-catalog/catalog_memory.go
```diff
@@ -49,6 +51,7 @@ func (c *MemorySandboxCatalog) StoreSandbox(ctx context.Context, sandboxID strin
...
 
 	c.cache.Set(sandboxID, sandboxInfo, expiration)
```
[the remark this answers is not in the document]
matthewlouisbrockman: k

packages/api/internal/sandbox/storage/memory/sync.go
```diff
@@ -37,6 +39,13 @@ func (s *Storage) Sync(sandboxes []sandbox.Sandbox, nodeID string) []sandbox.San
...
+			logger.L().Debug(
+				context.Background(),
```
djeebus: If we can, it's probably worth the effort to add `ctx context.Context` to the function signature

packages/api/internal/sandbox/storage/memory/sync.go
```diff
@@ -48,6 +57,13 @@ func (s *Storage) Sync(sandboxes []sandbox.Sandbox, nodeID string) []sandbox.San
 			continue
 		}
 
+		logger.L().Debug(
+			context.Background(),
```
djeebus: Same as above, add `ctx` to function args

djeebus (approved):

packages/api/internal/sandbox/storage/memory/sync.go
matthewlouisbrockman: kk got it for ya

packages/api/internal/sandbox/storage/memory/sync.go
matthewlouisbrockman: see above

PR #2250, opened 2026-03-28
feat: support per-route timeout overrides for ingress path rules

djeebus:
Changes the additional_api_paths_handled_by_ingress variable from `list(string)` to also accepting `list(object({ paths, timeout_sec? }))`. Each entry now creates its own path_rule, and when timeout_sec is set, it adds a route_action.timeout that overrides the ingress backend default (80s). Fully backwards compatible.

Also move traefik configuration to a config file, and support extra config files.

This was done in order to increase timeouts for specific paths, and enhance our ability to configure traefik per environment without exposing new variables.

### DIFF
--- a/iac/modules/job-ingress/jobs/traefik.toml
+++ b/iac/modules/job-ingress/jobs/traefik.toml
@@ -1,6 +1,7 @@
 [accessLog]
 
 [api]
+  dashboard = true
   insecure = false
 
 [entryPoints]
@@ -19,6 +20,11 @@
   entryPoint = "web"
 
 [providers]
+  [providers.consulCatalog]
+    exposedByDefault = false
+    [providers.consulCatalog.endpoint]
+      address = "${consul_endpoint}"
+      token = "${consul_token}"
   [providers.file]
     directory = "/local/config"
     watch = false
@@ -27,8 +33,3 @@
     [providers.nomad.endpoint]
       address = "${nomad_endpoint}"
       token = "${nomad_token}"
-  [providers.consulCatalog]
-    exposedByDefault = false
-    [providers.consulCatalog.endpoint]
-      address = "${consul_endpoint}"
-      token = "${consul_token}"


iac/provider-gcp/nomad-cluster/network/main.tf
```diff
@@ -270,11 +270,21 @@ resource "google_compute_url_map" "orch_map" {
...
-      for_each = length(var.additional_api_paths_handled_by_ingress) > 0 ? [{}] : []
+      for_each = var.additional_api_paths_handled_by_ingress
 
       content {
-        paths   = var.additional_api_paths_handled_by_ingress
+        paths   = path_rule.value.paths
         service = google_compute_backend_service.ingress.self_link
+
+        dynamic "route_action" {
+          for_each = path_rule.value.timeout_sec != null ? [path_rule.value.timeout_sec] : []
```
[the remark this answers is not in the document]
djeebus: technically true, but not meaningfully true. gcp will happily provide an error message.

iac/provider-gcp/nomad-cluster/network/main.tf
```diff
@@ -270,11 +270,21 @@ resource "google_compute_url_map" "orch_map" {
...
+
+          content {
+            timeout {
```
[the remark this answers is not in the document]
djeebus: technically true, but not meaningfully true. gcp will happily provide an error message.

iac/provider-gcp/nomad-cluster/network/variables.tf
```diff
@@ -104,5 +104,8 @@ variable "labels" {
 }
 
 variable "additional_api_paths_handled_by_ingress" {
-  type = list(string)
+  type = list(object({
+    paths       = list(string)
+    timeout_sec = optional(number)
+  }))
 }
```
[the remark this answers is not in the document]
djeebus: technically true, but not meaningfully true. gcp will happily provide an error message.

iac/provider-aws/nomad/variables.tf
```diff
@@ -309,9 +309,8 @@ variable "launch_darkly_api_key" {
   sensitive = true
 }
 
-variable "additional_traefik_arguments" {
-  type    = list(string)
-  default = []
+variable "traefik_config_files" {
```
sitole, line 312: Please provide default

iac/modules/job-ingress/jobs/ingress.hcl
```diff
@@ -81,41 +81,31 @@ job "ingress" {
...
+
+      template {
+        data = ""
+        destination = "local/config/.keep"
+      }
+
+%{ for filename, content in config_files }
+      template {
+        data        = <<EOF
+${content}
```
sitole, line 103: Do we already have examples of what we want to reconfigure? I did not find any relevant PR using this.

iac/provider-gcp/nomad-cluster/network/variables.tf
```diff
@@ -104,5 +104,12 @@ variable "labels" {
...
 
 variable "additional_api_paths_handled_by_ingress" {
-  type = list(string)
+  type = list(object({
+    paths       = list(string)
+    timeout_sec = optional(number)
+  }))
+}
+
+variable "ingress_timeout_seconds" {
```
sitole, line 113: Can we set a default here?

iac/provider-gcp/nomad-cluster/network/variables.tf
```diff
@@ -104,5 +104,12 @@ variable "labels" {
 }
 
 variable "additional_api_paths_handled_by_ingress" {
-  type = list(string)
```
sitole, line 107: Empty default here would be nice

iac/provider-gcp/variables.tf
```diff
@@ -132,16 +132,14 @@ variable "ingress_count" {
 }
 
 variable "additional_api_paths_handled_by_ingress" {
```
sitole, line 134: I think there is a missing change in the makefile to remove old and add new vars.

iac/provider-gcp/nomad-cluster/network/variables.tf
sitole: The network module is used in our closed-source code, so you don't need to redefine it imho.

iac/provider-aws/nomad/variables.tf
djeebus: Are we planning on using these outside of provider-aws / provider-gcp? If we're not, I'm passing the values in from both modules. so defaults are never used. Adding defaults only hides the fact that we forgot to implement the configuration in aws or gcp.

If we're planning on publishing these modules to terraform (for example) or having people use them on their own, then the default is still useful.

iac/modules/job-ingress/jobs/ingress.hcl
djeebus: Using this feature we (and anyone else) can configure traefik any way they want. New files could override logging, improve tracing, add routers, etc. We'll be using it to increase timeouts for specific paths, but that'll be done directly in runtime configuration.

iac/provider-gcp/nomad-cluster/network/variables.tf
djeebus: This default would never be used; we pass a value in to network/variables.tf explicitly. We define a default in iac/provider-gcp/variables.tf though, which is used and passed up the chain to here.

iac/provider-gcp/nomad-cluster/network/variables.tf
djeebus: The default is defined in iac/provider-gcp/variables.tf and propagated here.

iac/provider-gcp/variables.tf
djeebus: No need, we support both the old and new versions (as of the latest commit)

iac/provider-aws/nomad/variables.tf
```diff
@@ -309,9 +309,8 @@ variable "launch_darkly_api_key" {
   sensitive = true
 }
 
-variable "additional_traefik_arguments" {
-  type    = list(string)
-  default = []
```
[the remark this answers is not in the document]
djeebus: The only aws nomad module consumer is `iac/provider-aws`, and it has `{}` defined as a default for `traefik_config_files`

sitole (approved):

iac/modules/job-ingress/jobs/ingress.hcl
```diff
@@ -81,41 +81,31 @@ job "ingress" {
...
+        data        = <<EOF
+${traefik_config}
+EOF
+        destination = "local/traefik.toml"
+      }
+
+      template {
+        data = ""
+        destination = "local/config/.keep"
+      }
```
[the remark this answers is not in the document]
djeebus: Not true, traefik ignores files with unexpected extensions: https://github.com/traefik/traefik/blob/master/pkg/provider/file/file.go#L409-L414

PR #2256, opened 2026-03-29
fix: data races in tests and NFS cache scanner

tomassrnka:
## Summary

- **Serialize goose migrations with mutex** in test utilities to prevent races on package globals
- **Use RunAndReturn in envd conversion test** to fix race on `connect.Response` headers
- **Replace `*os.File` fd with path string** in NFS cache scanner to fix race between `Close`/`Fd`
- **Capture loop vars in NBD `path_direct` goroutine closure** to prevent stale variable access
- **Use `sync.Map` and mutexes in multipart upload tests** to eliminate concurrent map writes
- **Use `atomic.Bool` and separate context vars** in errorcollector test to fix shared state races

All of these races were exposed by ARM64 testing but are architecture-independent bugs.

Part of #1875 (ARM64 support split).

## Test plan

- [ ] Verify CI passes (existing unit tests cover all changed code)
- [ ] Run tests with `-race` flag to confirm data races are resolved
- [ ] Confirm no regressions in NFS cache cleaner behavior

### DIFF
--- a/packages/orchestrator/pkg/sandbox/nbd/path_direct.go
+++ b/packages/orchestrator/pkg/sandbox/nbd/path_direct.go
@@ -78,7 +78,7 @@ func (d *DirectPathMount) Open(ctx context.Context) (retDeviceIndex uint32, err 
 
 	telemetry.ReportEvent(ctx, "got backend size")
 
-	var deviceIndex uint32
+	deviceIndex := uint32(math.MaxUint32)
 
 	for {
 		deviceIndex, err = d.devicePool.GetDevice(ctx)


packages/orchestrator/cmd/clean-nfs-cache/cleaner/stat_linux.go
```diff
@@ -30,11 +30,11 @@ func (c *Cleaner) stat(fullPath string) (*Candidate, error) {
...
-func (c *Cleaner) statInDir(df *os.File, filename string) (*File, error) {
+func (c *Cleaner) statInDir(dirPath string, filename string) (*File, error) {
 	c.StatxC.Add(1)
 	c.StatxInDirC.Add(1)
 	var statx unix.Statx_t
-	err := unix.Statx(int(df.Fd()), filename,
+	err := unix.Statx(unix.AT_FDCWD, filepath.Join(dirPath, filename),
 		unix.AT_STATX_DONT_SYNC|unix.AT_SYMLINK_NOFOLLOW|unix.AT_NO_AUTOMOUNT,
 		unix.STATX_ATIME|unix.STATX_SIZE,
 		&statx,
```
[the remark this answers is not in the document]
tomassrnka: Fixed — statInDir now stores the full path in a variable and uses it in both the Statx call and the error message.

packages/orchestrator/pkg/sandbox/nbd/path_direct.go
```diff
@@ -119,14 +119,19 @@ func (d *DirectPathMount) Open(ctx context.Context) (retDeviceIndex uint32, err
 			server.Close()
 
 			dispatch := NewDispatch(serverc, d.Backend)
+			// Capture loop variables for the goroutine closure to avoid a data
+			// race: deviceIndex is reassigned on each retry iteration of the
+			// outer for-loop while the goroutine may still read it.
+			devIdx := deviceIndex
+			sockIdx := i
```
jakubno: this isn't needed, it was an issue in previous go versions 
```suggestion
```

packages/orchestrator/pkg/sandbox/nbd/path_direct.go
```diff
@@ -78,7 +78,7 @@ func (d *DirectPathMount) Open(ctx context.Context) (retDeviceIndex uint32, err
 
 	telemetry.ReportEvent(ctx, "got backend size")
 
-	deviceIndex := uint32(math.MaxUint32)
```
jakubno, line 81: why was this changed?

tomassrnka: Reverted — the `math.MaxUint32` sentinel was intentional. The clean branch has no net diff on this file.

packages/orchestrator/pkg/sandbox/nbd/path_direct.go
tomassrnka: Removed — Go 1.22+ gives each iteration its own variable. The clean branch has no net diff on path_direct.go.

PR #2260, opened 2026-03-29
feat: arch-aware downloads in create-build and fetch-busybox target

tomassrnka:
## Summary
- Update `create-build` (`setupKernel`/`setupFC`) to use arch-prefixed download URLs (e.g. `vmlinux-<version>-arm64`) with automatic legacy fallback for backward compatibility
- Add `errNotFound` sentinel error for 404 handling in the download helper, and use atomic temp-file writes to avoid partial downloads on failure
- Add `fetch-busybox` Makefile target that swaps the embedded amd64 busybox binary for an arm64 one when building on ARM64 hosts (used as a dependency for `build-local` and `build-debug`)

**Depends on:** #2258 (target arch path resolution)
**Part of:** #1875 (ARM64 support)

## Test plan
- [ ] Verify `make build-local` on amd64 host still works (fetch-busybox should no-op with "Using bundled amd64 busybox")
- [ ] Verify `make build-local` on arm64 host triggers busybox swap
- [ ] Verify `create-build` downloads arch-prefixed kernel/FC binaries when available
- [ ] Verify `create-build` falls back to legacy (non-prefixed) URLs when arch-prefixed files return 404


packages/orchestrator/Makefile
```diff
@@ -125,6 +125,34 @@ build-template:
 	-kernel $(KERNEL_VERSION) \
 	-firecracker $(FIRECRACKER_VERSION)
 
+.PHONY: fetch-busybox
+fetch-busybox:
+	@ARCH=$$(go env GOARCH); \
+	BUSYBOX_TARGET=./pkg/template/build/core/systeminit/busybox_1.36.1-2; \
```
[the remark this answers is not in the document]
tomassrnka: Resolved by PR #2257 — once the Makefiles PR merges, build-local uses $(BUILD_ARCH) which defaults to `go env GOARCH`, matching fetch-busybox. On the current branch, build-local still hardcodes amd64, but the two PRs are designed to merge together.

packages/orchestrator/Makefile
```diff
@@ -125,6 +125,34 @@ build-template:
...
+	elif file "$$BUSYBOX_TARGET" 2>/dev/null | grep -q 'aarch64\|ARM aarch64'; then \
+		echo "✓ Busybox is already arm64"; \
+	elif command -v busybox >/dev/null 2>&1 && file "$$(command -v busybox)" 2>/dev/null | grep -q 'aarch64\|ARM aarch64' && file "$$(command -v busybox)" 2>/dev/null | grep -q 'statically linked'; then \
+		cp "$$(command -v busybox)" "$$BUSYBOX_TARGET" && \
+		echo "✓ Copied host busybox (arm64, static) to embedded path"; \
+	elif command -v apt-get >/dev/null 2>&1 && command -v dpkg-deb >/dev/null 2>&1; then \
+		echo "Fetching arm64 busybox via apt..."; \
+		TMPDIR=$$(mktemp -d); \
+		apt-get download busybox-static 2>/dev/null && \
+			dpkg-deb -x busybox-static_*.deb "$$TMPDIR" && \
```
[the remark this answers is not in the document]
tomassrnka: Edge case — the rm cleanup at the end of the target handles stale .deb files. In practice, apt-get download produces a deterministic filename so glob collision is extremely unlikely.

packages/orchestrator/Makefile
```diff
@@ -21,14 +21,14 @@ build:
 	@docker build --platform linux/amd64 --output=bin --build-arg COMMIT_SHA="$(COMMIT_SHA)" -f ./Dockerfile ..
 
 .PHONY: build-local
-build-local:
+build-local: fetch-busybox
 	# Allow for passing commit sha directly for docker builds
 	$(eval COMMIT_SHA ?= $(shell git rev-parse --short HEAD))
 	CGO_ENABLED=1 GOOS=linux GOARCH=amd64 go build -o bin/orchestrator -ldflags "-X=main.commitSHA=$(COMMIT_SHA)" .
```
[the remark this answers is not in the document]
tomassrnka: Resolved by PR #2257 — same as above, BUILD_ARCH will match go env GOARCH once Makefiles PR merges.

packages/orchestrator/cmd/create-build/main.go
```diff
@@ -412,63 +412,141 @@ func printLocalFileSizes(basePath, buildID string) {
...
+	arch := utils.TargetArch()
+	dstPath := filepath.Join(dir, version, arch, "vmlinux.bin")
+
 	if err := os.MkdirAll(filepath.Dir(dstPath), 0o755); err != nil {
 		return fmt.Errorf("mkdir kernel dir: %w", err)
 	}
 
 	if _, err := os.Stat(dstPath); err == nil {
-		fmt.Printf("✓ Kernel %s exists\n", version)
+		fmt.Printf("✓ Kernel %s (%s) exists\n", version, arch)
```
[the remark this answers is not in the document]
tomassrnka: Minor optimization — the legacy local cache at a different path layout doesn't help for the new arch-prefixed structure. Re-downloading is the correct behavior.

packages/orchestrator/cmd/create-build/main.go
```diff
@@ -412,63 +412,141 @@ func printLocalFileSizes(basePath, buildID string) {
...
+
+func download(ctx context.Context, rawURL, path string, perm os.FileMode) error {
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, rawURL, nil)
+	if err != nil {
+		return fmt.Errorf("invalid download URL %s: %w", rawURL, err)
+	}
+
 	resp, err := (&http.Client{Timeout: 5 * time.Minute}).Do(req)
 	if err != nil {
 		return err
```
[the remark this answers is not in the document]
tomassrnka: Intentional — GCS (storage.googleapis.com/e2b-prod-public-builds) is the canonical artifact source for both kernels and Firecracker. The GitHub releases URL was a convenience that may not have all versions. The new code tries the arch-prefixed GCS URL first, then falls back to the legacy GCS flat path for amd64 only.

packages/orchestrator/Makefile
```diff
@@ -21,14 +21,14 @@ build:
...
 	CGO_ENABLED=1 GOOS=linux GOARCH=amd64 go build -o bin/clean-nfs-cache -ldflags "-X=main.commitSHA=$(COMMIT_SHA)" ./cmd/clean-nfs-cache
 
 .PHONY: build-debug
-build-debug:
+build-debug: fetch-busybox
 	CGO_ENABLED=1 GOOS=linux GOARCH=amd64 go build -race -gcflags=all="-N -l" -o bin/orchestrator .
 
 .PHONY: run-debug
```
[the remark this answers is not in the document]
tomassrnka: Resolved by PR #2257 — build-local will use $(BUILD_ARCH) instead of hardcoded amd64 once the Makefiles PR merges.

packages/orchestrator/cmd/create-build/main.go
```diff
@@ -414,71 +414,141 @@ func printLocalFileSizes(basePath, buildID string) {
...
+
+	fmt.Printf("⬇ Downloading kernel %s (%s)...\n", version, arch)
+
+	if err := download(ctx, archURL, dstPath, 0o644); err == nil {
+		return nil
+	} else if !errors.Is(err, errNotFound) {
+		return fmt.Errorf("failed to download kernel: %w", err)
+	}
+
+	// Legacy URLs are x86_64-only; only fall back for amd64.
```
jakubno, line 444: as mentioned in another PR, not really needed

packages/orchestrator/cmd/create-build/main.go
```diff
@@ -414,71 +414,141 @@ func printLocalFileSizes(basePath, buildID string) {
...
+	} else if !errors.Is(err, errNotFound) {
+		return fmt.Errorf("failed to download Firecracker: %w", err)
+	}
+
+	// Legacy URLs are x86_64-only; only fall back for amd64.
+	if arch != "amd64" {
+		return fmt.Errorf("firecracker %s not found for %s (no legacy fallback for non-amd64)", version, arch)
+	}
+
+	legacyURL, err := url.JoinPath("https://storage.googleapis.com/e2b-prod-public-builds/fc-versions/", version, "firecracker")
```
jakubno, line 492: this is needed as for some reason we're missing v1.10 for amd64

packages/orchestrator/Makefile
```diff
@@ -132,6 +132,34 @@ build-template:
 	-kernel $(KERNEL_VERSION) \
 	-firecracker $(FIRECRACKER_VERSION)
 
+.PHONY: fetch-busybox
```
jakubno: mentioned earlier, can we build it ourself and make it consistent for both arch?

packages/orchestrator/pkg/template/build/core/systeminit/busybox_1.36.1-2_amd64
jakubno, line 1: can we keep the version in the name? could you also add source to the comment? 🙏🏻

tomassrnka: Done — renamed to `busybox_1.36.1-2_amd64` / `busybox_1.36.1-2_arm64` with source comments. The amd64 binary is a custom musl build (origin unknown, added in #1002). The arm64 binary is from Debian `busybox-static 1:1.36.1-9`. Added TODO to rebuild both from the same config.

packages/orchestrator/cmd/create-build/main.go
tomassrnka: Kept — confirmed needed for v1.10.

packages/orchestrator/cmd/create-build/main.go
tomassrnka: Confirmed — legacy fallback is kept for this reason.

packages/orchestrator/Makefile
tomassrnka: Replaced with committed multi-arch binaries using Go build tags. TODO in the arm64.go file to rebuild both from the same minimal config in the future.

packages/orchestrator/Makefile
```diff
@@ -132,6 +132,34 @@ build-template:
...
+fetch-busybox:
+	@ARCH=$$(go env GOARCH); \
```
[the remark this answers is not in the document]
tomassrnka: Resolved — `fetch-busybox` was removed entirely. Both busybox binaries are now committed to the repo and selected via Go build tags at compile time.

PR #2262, opened 2026-03-30
feat: add more observability to grpc

jakubno:
Add observability for gRPC connections, track count and state. For clients created by our code also track the durations

### DIFF
--- a/packages/shared/pkg/grpc/channelz.go
+++ b/packages/shared/pkg/grpc/channelz.go
@@ -8,7 +8,6 @@ import (
 	"sync"
 	"time"
 
-	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/metric"
 	"go.uber.org/zap"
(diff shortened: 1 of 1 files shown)


iac/modules/job-otel-collector/configs/otel-collector.yaml
```diff
@@ -230,6 +230,8 @@ processors:
           - "pgxpool.*"
           - "go\\..*"
           - "traefik.*"
+          - "grpc\\..*"
+          - "rpc.client.*"
```
jakubno:
```suggestion
          - "rpc\\.client.*"
```

packages/shared/pkg/grpc/channelz.go
```diff
@@ -0,0 +1,492 @@
...
+	if clientType == "" {
+		return
+	}
+
+	target := normalizeChannelzTarget(conn.Target())
+	if target == "" {
+		return
+	}
+
+	channelzState.registerTargetClientType(target, clientType)
```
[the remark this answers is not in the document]
jakubno: Added comment with this limitations, we can't track it precisely with such setup and we don't have such case rn

packages/shared/pkg/grpc/channelz.go
```diff
@@ -0,0 +1,506 @@
...
+}
+
+func StartChannelzSampler(ctx context.Context) {
+	channelzInitOnce.Do(func() {
+		meter := otel.Meter("github.com/e2b-dev/infra/packages/shared/pkg/grpc")
```
arkamar: there is the same `meter` created at package level in `connobserver.go`. Would it make sense to use the package-level one here as well?

PR #2263, opened 2026-03-30
 fix(envd): detect cgroup v2 filesystem before initializing Cgroup2Manager

DL1231:
## Summary

  On cgroup v1 hosts, `Cgroup2Manager` silently "succeeds" initialization because `/sys/fs/cgroup` is a tmpfs where `mkdir` and `writeFile` both work without error. The resulting file descriptors point to regular directories, not cgroup v2 entries.
  When Go's `clone3` syscall uses these fds with `CLONE_INTO_CGROUP`, the kernel returns `EBADF`, causing **all** `fork/exec` operations to fail — including process start and socat port forwarding.

  This PR adds a `statfs` check for `CGROUP2_SUPER_MAGIC` (`0x63677270`) before creating cgroup directories, so the manager correctly returns an error and falls back to `NoopManager` on cgroup v1 systems.

  ## Root Cause

  1. `/sys/fs/cgroup` on cgroup v1 is a **tmpfs**, not cgroup2fs
  2. `os.MkdirAll("/sys/fs/cgroup/ptys")` succeeds (creates a regular directory)
  3. `os.WriteFile("cpu.weight", "200")` succeeds (creates a regular file)
  4. `unix.Open()` returns a valid fd — but to a regular directory
  5. `UseCgroupFD: true` is set, Go uses `clone3(CLONE_INTO_CGROUP)`
  6. Kernel validates the fd is not a cgroup v2 fd → **EBADF**

  ## Symptoms

  fork/exec /bin/bash: bad file descriptor
  fork/exec /usr/bin/socat: bad file descriptor

  All child process creation fails; filesystem RPCs (which don't fork) work fine.

### DIFF
--- a/packages/envd/pkg/version.go
+++ b/packages/envd/pkg/version.go
@@ -1,3 +1,3 @@
 package pkg
 
-const Version = "0.5.9"
+const Version = "0.5.10"


packages/envd/internal/services/cgroups/cgroup2.go
```diff
@@ -51,6 +51,19 @@ func NewCgroup2Manager(opts ...Cgroup2ManagerOption) (*Cgroup2Manager, error) {
 		opt(&config)
 	}
 
+	// Verify cgroup v2 is available by checking the filesystem type.
+	// On cgroup v1, /sys/fs/cgroup is a tmpfs and directories/files can be
+	// created freely, causing Cgroup2Manager to "succeed" with invalid fds
+	// that the kernel rejects with EBADF on clone3(CLONE_INTO_CGROUP).
+	var st unix.Statfs_t
+	if err := unix.Statfs(config.rootPath, &st); err != nil {
+		return nil, fmt.Errorf("failed to statfs cgroup root %s: %w", config.rootPath, err)
```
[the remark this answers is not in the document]
DL1231: Good catch! When `rootPath` points to a not-yet-created subdirectory (e.g. `/sys/fs/cgroup/envd`), `Statfs` returns `ENOENT` and the manager silently falls back to `NoopManager`, even though `createCgroups` would have created the directory via `os.MkdirAll`.

Fixed by walking up to the nearest existing ancestor directory before calling `Statfs`. This way a custom `--cgroup-root` like `/sys/fs/cgroup/envd` correctly checks the cgroup2 filesystem type on `/sys/fs/cgroup` (which always exists), while still rejecting non-cgroup2 mounts.

packages/envd/internal/services/cgroups/cgroup2.go
```diff
@@ -51,6 +51,34 @@ func NewCgroup2Manager(opts ...Cgroup2ManagerOption) (*Cgroup2Manager, error) {
...
+			break
+		}
+		checkPath = parent
+	}
+	var st unix.Statfs_t
+	if err := unix.Statfs(checkPath, &st); err != nil {
+		return nil, fmt.Errorf("failed to statfs cgroup root %s: %w", checkPath, err)
+	}
+	const cgroup2SuperMagic = 0x63677270
+	if st.Type != cgroup2SuperMagic {
```
arkamar:
```suggestion
	if st.Type != unix.CGROUP2_SUPER_MAGIC {
```
The local constant is unnecessary, `unix` package provides the same value and it is already imported.

packages/envd/internal/services/cgroups/cgroup2.go
```diff
@@ -51,6 +51,34 @@ func NewCgroup2Manager(opts ...Cgroup2ManagerOption) (*Cgroup2Manager, error) {
...
+	for {
+		if _, err := os.Stat(checkPath); err == nil {
+			break
+		}
+		parent := filepath.Dir(checkPath)
+		if parent == checkPath {
+			break
+		}
+		checkPath = parent
+	}
```
arkamar: If the cgroup root path doesn't exist, this walks up to a parent that could be a different filesystem entirely (sysfs, rootfs), making the statfs check meaningless. The cgroup root should exist, and `NewCgroup2Manager()` should just error if it doesn't.

DL1231: Thanks for the review, fixed in [162bbda49](https://github.com/e2b-dev/infra/pull/2263/changes/162bbda49742d1c2a3932fb8a00ce127831151c7)

PR #2288, opened 2026-04-02
fix(dashboard-api): invalidate auth cache on team membership changes

ben-fornefeld:


ben-fornefeld:
this pr is only half of the story. it does not invalidate infra api's in memory auth cache

PR #2281, opened 2026-04-01
feat: download busybox from fc-busybox release instead of embedding

tomassrnka:
## Summary

- Download busybox binary from [e2b-dev/fc-busybox](https://github.com/e2b-dev/fc-busybox) GitHub release at build time instead of committing binaries to git
- Dockerfile: `ADD` from GitHub release, uses `TARGETARCH` for correct arch
- Makefile: `fetch-busybox` target for local dev builds (`build-local`, `build-debug`)
- Removes ~3.2MB of committed binaries from git

## Test plan

- [ ] Docker build produces working orchestrator for amd64
- [ ] Docker build produces working orchestrator for arm64
- [ ] `make build-local` downloads busybox and builds
- [ ] Integration tests pass (sandbox creation exercises busybox in rootfs)

PR #2292, opened 2026-04-02
fix: update test.Dockerfile after internal/ → pkg/ migration

tomassrnka:
## Summary

- Replace `COPY internal internal` with `COPY pkg pkg`, `COPY cmd cmd`, `COPY scripts scripts`
- The orchestrator code moved from `internal/` to `pkg/` in #2184 but `test.Dockerfile` was not updated
- Also adds `scripts/` (needed for `fetch-busybox` once #2281 lands) and `cmd/` for full coverage

## Test plan

- [ ] `make test-docker` completes successfully

### DIFF
--- a/packages/orchestrator/test.Dockerfile
+++ b/packages/orchestrator/test.Dockerfile
@@ -16,7 +16,7 @@ COPY .shared/pkg pkg
 
 WORKDIR /build/clickhouse
 
-# Copy clickhouse package dependencies  
+# Copy clickhouse package dependencies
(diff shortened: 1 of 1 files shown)


packages/orchestrator/test.Dockerfile
```diff
@@ -30,7 +30,9 @@ RUN go mod download
 
 # Copy source code
 COPY main.go Makefile ./
-COPY internal internal
+COPY scripts scripts
+COPY pkg pkg
+COPY cmd cmd
```
dobrac, line 35: this might not be necessary, but ok

PR #2298, opened 2026-04-02
refactor(storage): rename TemplateFiles → Paths, simplify path API

levb:
- Rename TemplateFiles → Paths (it maps a build ID to storage paths)
- Shorten method names: StorageMemfilePath → Memfile, StorageRootfsPath → Rootfs, etc.
- Rename ParseStoragePath → SplitPath
- Replace manual string concatenation in template/storage.go with Paths methods
- Rename template.go → paths.go to match the type name

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/packages/orchestrator/pkg/sandbox/template/storage.go
+++ b/packages/orchestrator/pkg/sandbox/template/storage.go
@@ -57,17 +57,19 @@ func NewStorage(
 	paths := storage.Paths{BuildID: buildId}
 
 	if h == nil {
-		var hdrPath string
 		headerObjectType, ok := storageHeaderObjectType(fileType)
 		if !ok {
 			return nil, build.UnknownDiffTypeError{DiffType: fileType}
 		}
 
+		var hdrPath string
 		switch fileType {
 		case build.Memfile:
 			hdrPath = paths.MemfileHeader()
 		case build.Rootfs:
 			hdrPath = paths.RootfsHeader()
+		default:
+			return nil, build.UnknownDiffTypeError{DiffType: fileType}
 		}
 
 		headerObject, err := persistence.OpenBlob(ctx, hdrPath, headerObjectType)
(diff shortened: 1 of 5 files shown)


packages/orchestrator/pkg/sandbox/template/storage.go
```diff
@@ -54,14 +54,23 @@ func NewStorage(
...
 		if !ok {
 			return nil, build.UnknownDiffTypeError{DiffType: fileType}
 		}
 
-		headerObject, err := persistence.OpenBlob(ctx, headerObjectPath, headerObjectType)
+		switch fileType {
+		case build.Memfile:
+			hdrPath = paths.MemfileHeader()
+		case build.Rootfs:
+			hdrPath = paths.RootfsHeader()
```
dobrac, line 70: lets add do default to return an error rather than using an empty string

packages/orchestrator/pkg/sandbox/template/storage.go
```diff
@@ -80,12 +89,20 @@ func NewStorage(
...
 		if !ok {
 			return nil, build.UnknownDiffTypeError{DiffType: fileType}
 		}
-		object, err := persistence.OpenSeekable(ctx, objectPath, objectType)
+
+		switch fileType {
+		case build.Memfile:
+			dataPath = paths.Memfile()
+		case build.Rootfs:
+			dataPath = paths.Rootfs()
```
dobrac, line 104: same here with the default

packages/orchestrator/pkg/sandbox/template/storage.go
```diff
@@ -54,14 +54,23 @@ func NewStorage(
 	persistence storage.StorageProvider,
 	metrics blockmetrics.Metrics,
 ) (*Storage, error) {
+	paths := storage.Paths{BuildID: buildId}
+
 	if h == nil {
-		headerObjectPath := buildId + "/" + string(fileType) + storage.HeaderSuffix
+		var hdrPath string
```
dobrac: small nit: better to colocate the variable with the assignments (applies to other places with similar logic)

packages/shared/pkg/storage/paths.go
```diff
@@ -0,0 +1,62 @@
...
+	return fmt.Sprintf("%s/%s", p.BuildID, SnapfileName)
+}
+
+func (p Paths) Metadata() string {
+	return fmt.Sprintf("%s/%s", p.BuildID, MetadataName)
+}
+
+// SplitPath splits a storage path of the form "{buildID}/{fileName}"
+// back into its components. This is the inverse of the path methods.
+func SplitPath(path string) (buildID, fileName string) {
```
dobrac, line 58: the `ParseStoragePath` made more sense to me, especially when we give meaning to the returned values

levb: In the compression branch, this is used to parse the P2P request paths, which are explicitly uncompressed; `ParseUncompressedStoragePath` (to avoid confusion with parsing compressed paths) - seemed too much for a 1-line function.

PR #2337, opened 2026-04-08
refactor(api,dashboard-api): Allow traceparent propagation into API

ben-fornefeld:
## Summary
- stop stripping incoming traceparent in the API tracing middleware
- extract remote trace context before starting the server span so caller trace IDs propagate through API spans
- allow traceparent, tracestate, and baggage through API and dashboard-api CORS headers


packages/shared/pkg/telemetry/traces.go
```diff
@@ -12,6 +12,8 @@ import (
 	"google.golang.org/grpc/encoding/gzip"
 )
 
+var contextPropagationHeaders = NewTextPropagator().Fields()
```
djeebus, line 15: We won't necessarily be using this one:


```suggestion
var contextPropagationHeaders []string

func init() {
	contextPropagationHeaders = otel.GetTextMapPropagator().Fields()
}
```

ValentaTomas:

PR #2343, opened 2026-04-09
Use consistent workflow dependencies

djeebus:
Deploys are executed against a ref ("main") and a sha ("HEAD-3", for example. the workflow is executed from "main", which checks out "HEAD-3", which can cause inconsistencies in inputs and side effects. this PR checks out the actions before they're read


djeebus:
> LGTM — straightforward CI workflow fix.

anything but straight forward ...

dobrac (approved):

PR #2344, opened 2026-04-09
refactor(storage): FrameTable - per-build instead of per-mapping

levb:
**Problem**: Numerous per-mapping allocations (FrameTable subsets) fragment the heap and pressure GC at scale. Mapping count grows with snapshots; build count stays bounded by layer depth.

**Solution**: Per-build FrameTable. Allocation count scales with N(builds) < N(mappings).

**Trade-off**: 24 bytes/entry vs 8 - absolute offsets are required to store non-continuous referenced frames, and somewhat balanced by deduplication (frames shared across mappings are now stored once).

 ## Summary

  - Move `FrameTable` from per-mapping (`BuildMap.FrameTable`) to per-build (`Header.Builds[id].FrameData`), looked up by build ID at read time
  - Eliminate `SetFrames`/`Subset`/`mergeFrameTables` machinery from mapping split/merge/normalize — net deletion of ~260 lines of complex frame-subsetting logic
  - `frameEntry` uses absolute `[StartU, EndU, StartC, EndC)` offsets with binary search, replacing relative `FrameOffset` + `[]FrameSize` with linear scan
  - `TrimToRanges` for sparse serialization: only frames referenced by mappings are written to the V4 header
  - `DeserializeFrameTable` validates OOM cap, U/C-space monotonicity, and positive frame sizes

### DIFF
--- a/packages/shared/pkg/storage/compress_frame_table.go
+++ b/packages/shared/pkg/storage/compress_frame_table.go
@@ -35,13 +35,14 @@ func (r Range) String() string {
 	return fmt.Sprintf("%d/%d", r.Offset, r.Length)
 }
 
-// frameEntry stores one frame as absolute [Start, End) byte offsets in both
-// address spaces.
+// frameEntry stores one frame as an absolute start offset plus size in both
+// address spaces. Fields must be exported for encoding/binary (Read/Write use reflection).
+// Field order chosen for optimal alignment: two int64 then two uint32 = 24 bytes, no padding.
 type frameEntry struct {
 	StartU int64
-	EndU   int64
 	StartC int64
-	EndC   int64
+	SizeU  int32
+	SizeC  int32
 }
 
 // FrameTable is the decompression index for a compressed diff file.
@@ -65,9 +66,9 @@ func NewFrameTable(ct CompressionType, sizes []FrameSize) *FrameTable {
 	for i, s := range sizes {
 		entries[i] = frameEntry{
 			StartU: u,
-			EndU:   u + int64(s.U),
 			StartC: c,
-			EndC:   c + int64(s.C),
+			SizeU:  s.U,
+			SizeC:  s.C,
 		}
 		u += int64(s.U)
 		c += int64(s.C)
@@ -85,10 +86,9 @@ func (ft *FrameTable) CompressionType() CompressionType {
 	return ft.compressionType
 }
 
-// IsCompressed reports whether ft is non-nil, has a compression type set, and
-// contains at least one frame.
+// IsCompressed reports whether ft is non-nil and has a compression type set.
 func (ft *FrameTable) IsCompressed() bool {
-	return ft != nil && ft.compressionType != CompressionNone && len(ft.entries) > 0
+	return ft != nil && ft.compressionType != CompressionNone
 }
 
 func (ft *FrameTable) NumFrames() int {
@@ -99,17 +99,20 @@ func (ft *FrameTable) NumFrames() int {
 	return len(ft.entries)
 }
 
+func (e frameEntry) endU() int64 { return e.StartU + int64(e.SizeU) }
+func (e frameEntry) endC() int64 { return e.StartC + int64(e.SizeC) }
+
 func (ft *FrameTable) FrameAt(i int) (startU, endU, startC, endC int64) {
 	e := ft.entries[i]
 
-	return e.StartU, e.EndU, e.StartC, e.EndC
+	return e.StartU, e.endU(), e.StartC, e.endC()
 }
 
 // UncompressedSize returns the total uncompressed size across all frames.
 func (ft *FrameTable) UncompressedSize() int64 {
 	var total int64
 	for _, e := range ft.entries {
-		total += e.EndU - e.StartU
+		total += int64(e.SizeU)
 	}
 
 	return total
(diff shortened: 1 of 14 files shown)


packages/shared/pkg/storage/header/serialization_v4.go
```diff
@@ -19,59 +19,74 @@ type v4SerializableBuildMap struct {
...
+	FileSize int64
 	Checksum [32]byte
 }
 
-// serializeV4 writes [Metadata] [uint32 uncompressedSize] [LZ4( BuildFiles + counted mappings + FrameTables )].
-func serializeV4(metadata *Metadata, buildFiles map[uuid.UUID]BuildFileInfo, mappings []*BuildMap) ([]byte, error) {
-	// --- raw metadata prefix (not compressed) ---
+// serializeV4 writes [Metadata] [uint32 uncompressedSize] [LZ4( Builds[] + mappings[] )].
+// Frame tables are sparse-trimmed to only frames referenced by mappings.
+func serializeV4(metadata *Metadata, builds map[uuid.UUID]BuildData, mappings []*BuildMap) ([]byte, error) {
```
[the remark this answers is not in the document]
levb: V4 has not been deployed yet, no compatibility issue

packages/shared/pkg/storage/header/serialization_v4.go
```diff
@@ -138,28 +131,37 @@ func deserializeV4(metadata *Metadata, blockData []byte) (*Header, error) {
...
-				Size:     entry.Size,
+
+			bd := BuildData{
+				Size:     entry.FileSize,
 				Checksum: entry.Checksum,
 			}
+
+			ft, err := storage.DeserializeFrameTable(reader)
+			if err != nil {
+				return nil, fmt.Errorf("failed to read frame table for build %s: %w", entry.BuildId, err)
```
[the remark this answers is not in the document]
levb: V4 has not been deployed yet, not relevant

packages/shared/pkg/storage/compress_frame_table.go
```diff
@@ -114,113 +91,200 @@ func (ft *FrameTable) IsCompressed() bool {
...
+
+	trimmed := make([]frameEntry, 0, kept)
+	for i, e := range ft.entries {
+		if keep[i] {
+			trimmed = append(trimmed, e)
+		}
 	}
 
-	return Range{Offset: start.U, Length: int(size.U)}, nil
+	return &FrameTable{compressionType: ft.compressionType, entries: trimmed}
```
dobrac: use the constructor instead

packages/shared/pkg/storage/compress_frame_table.go
```diff
@@ -12,60 +15,6 @@ const (
 	CompressionLZ4
 )
 
-// # Compression Layout
```
dobrac, line 15: is this no longer relevant?

packages/shared/pkg/storage/compress_frame_table.go
```diff
@@ -114,113 +91,200 @@ func (ft *FrameTable) IsCompressed() bool {
...
+
+	if ct == 0 && n > 0 {
+		return nil, fmt.Errorf("compression type is 0 but frame count is %d: corrupted header", n)
+	}
+	if ct == 0 || n == 0 {
+		return nil, nil
+	}
+
+	// Cap to prevent OOM from corrupted headers. 1<<20 frames × 24 bytes = 24 MiB.
+	const maxFrames = 1 << 20
```
dobrac: move to a file level constant, use the bytes utils

packages/shared/pkg/storage/compress_frame_table.go
```diff
@@ -114,113 +91,200 @@ func (ft *FrameTable) IsCompressed() bool {
...
-// its start position (in both address spaces) and full size.
-func (ft *FrameTable) locate(offset int64) (frameOffset FrameOffset, frameSize FrameSize, err error) {
-	if ft == nil {
-		return FrameOffset{}, FrameSize{}, fmt.Errorf("locate called with nil frame table - data is not compressed")
+// Serialize writes the frame table to w in binary little-endian format.
+// Nil-safe: writes zeros for type and count.
+func (ft *FrameTable) Serialize(w io.Writer) error {
+	if !ft.IsCompressed() {
+		z := [8]byte{} // two uint32 zeros: CompressionType=0, NumFrames=0
+		_, err := w.Write(z[:])
```
dobrac: Why do we need to write anything? If so, why does it need to be a special case?

packages/shared/pkg/storage/compress_frame_table.go
```diff
@@ -114,113 +91,200 @@ func (ft *FrameTable) IsCompressed() bool {
...
+	}
+
+	keep := make([]bool, len(ft.entries))
+	kept := 0
+
+	for _, r := range ranges {
+		startU, endU := r[0], r[1]
+
+		// Binary search: first frame whose EndU > startU.
+		lo := sort.Search(len(ft.entries), func(i int) bool {
```
dobrac, line 324: the binary search is in multiple places, can we unite the logic to a helper function?

dobrac (changes requested):

packages/shared/pkg/storage/compress_frame_table.go
levb: Thx! Refactored to make it more obvious; these 8 bytes are the "minimal viable frame table" that we serialize for uncompressed builds. 0 compression type, 0 frames. Was Claude's naïve implementation.

packages/shared/pkg/storage/compress_frame_table.go
levb: Those are on different types, would be just wrapping sort.Search, not sure the benefit?

dobrac: Alright, I thought they're doing the same thing. lgtm then

PR #2345, opened 2026-04-09
feat(resume-build): add optional telemetry/tracing support

arkamar:
Wire telemetry.New() into resume-build so traces and metrics are exported when `OTEL_COLLECTOR_GRPC_ENDPOINT` is set. When unset, behavior is unchanged (noop client, zero overhead).

---

This is useful for sandbox debugging.

### DIFF
--- a/packages/orchestrator/cmd/resume-build/main.go
+++ b/packages/orchestrator/cmd/resume-build/main.go
@@ -1049,7 +1049,10 @@ func run(ctx context.Context, buildID string, iterations int, coldStart, noPrefe
 	if verbose {
 		fmt.Println("🔧 Creating block metrics...")
 	}
-	blockMetrics, _ := blockmetrics.NewMetrics(tel.MeterProvider)
+	blockMetrics, err := blockmetrics.NewMetrics(tel.MeterProvider)
+	if err != nil {
+		return fmt.Errorf("block metrics: %w", err)
+	}
 
 	if verbose {
 		fmt.Println("🔧 Creating template cache...")
--- a/packages/orchestrator/cmd/resume-build/main.go
+++ b/packages/orchestrator/cmd/resume-build/main.go
@@ -18,7 +18,6 @@ import (
 	"connectrpc.com/connect"
 	"github.com/google/uuid"
 	"github.com/launchdarkly/go-sdk-common/v3/ldlog"
-	"go.opentelemetry.io/otel/metric/noop"
 	"golang.org/x/sys/unix"
 
 	"github.com/e2b-dev/infra/packages/clickhouse/pkg/hoststats"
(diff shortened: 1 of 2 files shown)


packages/orchestrator/cmd/resume-build/main.go
```diff
@@ -965,6 +965,18 @@ func run(ctx context.Context, buildID string, iterations int, coldStart, noPrefe
...
 
+	// Initialize telemetry (traces, metrics) if OTEL_COLLECTOR_GRPC_ENDPOINT is set.
+	// When unset, telemetry.New() returns a noop client with zero overhead.
+	tel, err := telemetry.New(ctx, "resume-build", "resume-build", "", "dev", uuid.NewString())
+	if err != nil {
+		return fmt.Errorf("telemetry: %w", err)
+	}
+	defer func() {
+		if err := tel.Shutdown(context.WithoutCancel(ctx)); err != nil {
+			log.Printf("error shutting down telemetry: %v", err)
```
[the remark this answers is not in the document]
arkamar: Not specific to this PR. Out of scope here.

packages/orchestrator/cmd/resume-build/main.go
```diff
@@ -965,6 +965,18 @@ func run(ctx context.Context, buildID string, iterations int, coldStart, noPrefe
 	}
 	sbxlogger.SetSandboxLoggerInternal(logger.NewNopLogger())
 
+	// Initialize telemetry (traces, metrics) if OTEL_COLLECTOR_GRPC_ENDPOINT is set.
+	// When unset, telemetry.New() returns a noop client with zero overhead.
+	tel, err := telemetry.New(ctx, "resume-build", "resume-build", "", "dev", uuid.NewString())
```
sitole: Can we have some help with this, which will be used for all telemetry.New across codebase?

sitole (changes requested):

packages/orchestrator/cmd/resume-build/main.go
arkamar: I am not sure what exactly do you have in mind? There are 4 calls in real services, all have build-time injected values and use `env.GetNodeID()`, so there isn't much shared boilerplate to extract. `resume-build` is the only outlier since it's a dev tool without ldflags. Are you concerned about the hardcoded dummy values, or something else?

sitole: The first thing is a comment noting the presence of `OTEL_COLLECTOR_GRPC_ENDPOINT`. Let's move it to `telemetry.New` fn definition instead.

Additionally, propagating dummy values this way isn't ideal. I understand they are not relevant in this context, so lets intead create something like `telemetry.NewAnonymous` that will not require dummy values. I am okay with `telemetry.NewAnonymous` internally using `telemetry.New` with some constant value.

arkamar: I see, thanks. I moved the comment, it makes sense and create `NewAnonymous()` which accepts `serviceName` as it is good for filtering.

sitole (approved):

PR #2347, opened 2026-04-09
fix(dashboard-api): team name validation in dashboard API

ben-fornefeld:
## Summary
- enforce the same team name validation in dashboard-api as the dashboard uses for team name updates
- trim names server-side before persisting so backend behavior matches the dashboard zod schema
- tighten the dashboard OpenAPI spec for this route and regenerate the embedded spec

## Details
The dashboard validates team names with these rules:
- trim surrounding whitespace
- require 1-32 characters after trimming
- allow only alphanumeric chunks separated by spaces, underscores, hyphens, or dots

OpenAPI can enforce structural constraints like `minLength`, `maxLength`, `pattern`, and `additionalProperties: false`, and this service already runs the oapi request validator middleware before handlers. But OpenAPI does not normalize the input, so the handler still needs validation to preserve parity with the dashboard's `.trim()` behavior before writing to the database.

## Testing
- go test ./packages/dashboard-api/internal/handlers/...


dobrac (changes requested):
please resolve the bot comments first :pray:

ValentaTomas:

PR #2348, opened 2026-04-09
Redis is a required part of our infra

djeebus:


packages/orchestrator/pkg/factories/run.go
```diff
@@ -383,17 +383,17 @@ func run(config cfg.Config, opts Options) (success bool) {
 		RedisTLSCABase64: config.RedisTLSCABase64,
 		PoolSize:         config.RedisPoolSize,
 	})
-	if err != nil && !errors.Is(err, sharedFactories.ErrRedisDisabled) {
+	if err != nil {
```
levb, line 386: 1/5 `client-proxy/main.go` still has a similar check, should it be eliminated also?

levb (changes requested):
1/5 (basically what Claude said ^^) Right now the error message a developer would see is "redis is disabled" wrapped inside "Could not connect to Redis", which is a bit misleading - they didn't fail to connect, they failed to configure. A dedicated config validation at startup ("REDIS_URL or REDIS_CLUSTER_URL must be set") would be clearer.

Are local runs affected by this change?

levb (approved):
LGTM, contingent on tests passing (still running)

PR #2349, opened 2026-04-10
feat: add rootfs usage rate limits and observability

dobrac:
Adds possible rate limiting and observability to Sandbox rootfs disk.

### DIFF
--- a/packages/orchestrator/pkg/sandbox/fc/fc_metrics.go
+++ b/packages/orchestrator/pkg/sandbox/fc/fc_metrics.go
@@ -35,6 +35,8 @@ var (
 	directionKey = attribute.Key("direction")
 	attrTX       = metric.WithAttributes(directionKey.String("tx"))
 	attrRX       = metric.WithAttributes(directionKey.String("rx"))
+	attrRead     = metric.WithAttributes(directionKey.String("read"))
+	attrWrite    = metric.WithAttributes(directionKey.String("write"))
 
 	// Counters — global totals, no sandbox_id to avoid high cardinality.
 	fcNetFails         = utils.Must(telemetry.GetCounter(fcMeter, telemetry.SandboxFCNetFails))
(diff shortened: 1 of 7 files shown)


packages/orchestrator/pkg/sandbox/fc/fc_metrics.go
```diff
@@ -204,6 +236,31 @@ func (p *Process) startMetricsReader(ctx context.Context) {
 			if n.TapReadFails > 0 {
 				fcNetTapIOFails.Add(ctx, int64(n.TapReadFails), attrRX)
 			}
+
+			// Block histograms — values are already per-flush deltas from Firecracker.
+			b := &m.Block
+
+			fcBlockBytes.Record(ctx, int64(b.ReadBytes), attrRead)
```
levb, line 242: 1/5 perf: there is no reason to record 0-counts here. If I understand correctly, these are periodic flushes and if the sandbox was idle we get 0 deltas, which are not valuable to report.

packages/orchestrator/pkg/sandbox/fc/fc_metrics.go
```diff
@@ -76,13 +90,31 @@ type firecrackerNetMetrics struct {
 	TapReadFails           uint64 `json:"tap_read_fails"`
 }
 
+// firecrackerBlockMetrics holds the Firecracker aggregate block device metrics.
+// Fields match Firecracker's BlockDeviceMetrics (per-flush deltas).
+// The aggregate "block" key sums over all drives; we only have one (rootfs).
+type firecrackerBlockMetrics struct {
```
levb, line 95: 0/5 style: mention in the comment that this is a deliberate subset of all FC metrics? Link to the full set?

levb (changes requested):
A couple perf/style nits.

Also, 2/5 as a separate refactoring PR: `TokenBucketConfig` is identical in both `featureflags` and `fc` packages. Consolidate?

packages/orchestrator/pkg/sandbox/fc/fc_metrics.go
dobrac: added

levb (approved):
LGTM

packages/orchestrator/pkg/sandbox/fc/fc_metrics.go
dobrac: We care about the size of the interval (the zeroes), so if we have 100 sandboxes running in 1 second interval, we know that e.g. 97 of them have 0 traffic and 3 of them used 5 bytes of traffic. Making the p50 still 0.

dobrac:
> TokenBucketConfig is identical in both featureflags and fc packages. Consolidate?

Agree here, lets do it separately

PR #2354, opened 2026-04-10
fix(orch): race condition in acquire while snapshotting

jakubno:
Make sure the sandbox is marked dead only if it's being really removed

Remove mutex and moves the logic to the atomic swap

### DIFF
--- a/packages/orchestrator/pkg/server/sandboxes.go
+++ b/packages/orchestrator/pkg/server/sandboxes.go
@@ -549,16 +549,16 @@ func (s *Server) Checkpoint(ctx context.Context, in *orchestrator.SandboxCheckpo
 		}
 	}
 
-	sbx, err := s.acquireSandboxForSnapshot(ctx, in.GetSandboxId())
-	if err != nil {
-		return nil, err
-	}
-
 	// Acquire the starting semaphore before resuming, same as Create/Pause.
 	if err := s.waitForAcquire(ctx); err != nil {
 		return nil, err
 	}
 	defer s.startingSandboxes.Release(1)
+
+	sbx, err := s.acquireSandboxForSnapshot(ctx, in.GetSandboxId())
+	if err != nil {
+		return nil, err
+	}
 
 	// Always stop the old sandbox when done — on success the resumed sandbox
 	// takes over, on failure this prevents a leaked sandbox that is running
--- a/packages/orchestrator/pkg/server/sandboxes.go
+++ b/packages/orchestrator/pkg/server/sandboxes.go
@@ -490,6 +490,7 @@ func (s *Server) Pause(ctx context.Context, in *orchestrator.SandboxPauseRequest
 	}
 
 	sbxlogger.E(sbx).Info(ctx, "Pausing sandbox")
+	s.sandboxFactory.Sandboxes.MarkStopping(ctx, sbx.Runtime.SandboxID, sbx.LifecycleID)
 
 	// Stop the old sandbox in background after we're done
 	defer s.stopSandboxAsync(context.WithoutCancel(ctx), sbx)
@@ -549,16 +550,16 @@ func (s *Server) Checkpoint(ctx context.Context, in *orchestrator.SandboxCheckpo
 		}
 	}
 
+	sbx, err := s.acquireSandboxForSnapshot(ctx, in.GetSandboxId())
+	if err != nil {
+		return nil, err
+	}
+
 	// Acquire the starting semaphore before resuming, same as Create/Pause.
 	if err := s.waitForAcquire(ctx); err != nil {
 		return nil, err
 	}
 	defer s.startingSandboxes.Release(1)
-
-	sbx, err := s.acquireSandboxForSnapshot(ctx, in.GetSandboxId())
-	if err != nil {
-		return nil, err
-	}
 
 	// Always stop the old sandbox when done — on success the resumed sandbox
 	// takes over, on failure this prevents a leaked sandbox that is running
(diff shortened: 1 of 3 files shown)


packages/orchestrator/pkg/server/sandboxes.go
```diff
@@ -549,17 +549,17 @@ func (s *Server) Checkpoint(ctx context.Context, in *orchestrator.SandboxCheckpo
...
 	// Acquire the starting semaphore before resuming, same as Create/Pause.
 	if err := s.waitForAcquire(ctx); err != nil {
 		return nil, err
 	}
 	defer s.startingSandboxes.Release(1)
 
+	sbx, err := s.acquireSandboxForSnapshot(ctx, in.GetSandboxId())
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
jakubno: this shouldn't happen

dobrac (approved):

packages/orchestrator/pkg/server/sandboxes.go
```diff
@@ -542,16 +556,16 @@ func (s *Server) Checkpoint(ctx context.Context, in *orchestrator.SandboxCheckpo
 			Build(),
 	)
 
-	// Check envd version before acquiring (which removes the sandbox from the map).
-	if preSbx, ok := s.sandboxFactory.Sandboxes.Get(in.GetSandboxId()); ok {
-		if err := utils.CheckEnvdVersionForSnapshot(preSbx.Config.Envd.Version); err != nil {
-			return nil, status.Errorf(codes.FailedPrecondition, "%s", err.Error())
-		}
+	sbx, ok := s.sandboxFactory.Sandboxes.Get(in.GetSandboxId())
```
jakubno, line 559: simplifed here as we don't need to get the same sandbox twice

packages/orchestrator/pkg/server/sandboxes.go
```diff
@@ -560,6 +574,13 @@ func (s *Server) Checkpoint(ctx context.Context, in *orchestrator.SandboxCheckpo
 	}
 	defer s.startingSandboxes.Release(1)
 
+	marked := s.sandboxFactory.Sandboxes.MarkStopping(ctx, sbx.Runtime.SandboxID, sbx.LifecycleID)
```
jakubno, line 577: instead of doing it in mutex, CAS to make sure the sandbox isn't being removed already

packages/orchestrator/pkg/server/sandboxes.go
```diff
@@ -484,9 +489,18 @@ func (s *Server) Pause(ctx context.Context, in *orchestrator.SandboxPauseRequest
...
-	if err != nil {
-		return nil, err
+	sbx, ok := s.sandboxFactory.Sandboxes.Get(in.GetSandboxId())
+	if !ok {
+		telemetry.ReportCriticalError(ctx, "sandbox not found", nil)
+
+		return nil, status.Error(codes.NotFound, "sandbox not found")
+	}
+
+	marked := s.sandboxFactory.Sandboxes.MarkStopping(ctx, sbx.Runtime.SandboxID, sbx.LifecycleID)
```
jakubno, line 499: instead of doing it in mutex, CAS to make sure the sandbox isn't being removed already

packages/orchestrator/pkg/server/sandboxes.go
```diff
@@ -425,13 +425,18 @@ func (s *Server) Delete(ctxConn context.Context, in *orchestrator.SandboxDeleteR
...
 	}
 
-	sbxlogger.E(sbx).Info(ctx, "Killing sandbox")
-
 	// Mark the sandbox as stopping so it is excluded from live queries (Get, Items,
 	// Count) but remains findable by IP (GetByHostPort) while the Firecracker
 	// process finishes shutting down.
 	// This prevents the sandbox to be synced to API again
-	s.sandboxFactory.Sandboxes.MarkStopping(ctx, sbx.Runtime.SandboxID, sbx.LifecycleID)
+	marked := s.sandboxFactory.Sandboxes.MarkStopping(ctx, sbx.Runtime.SandboxID, sbx.LifecycleID)
```
jakubno, line 432: the check wasn't here before, but the same holds + Get() returns NotFound if the sandbox is already marked

packages/orchestrator/pkg/server/sandboxes.go
```diff
@@ -425,13 +425,18 @@ func (s *Server) Delete(ctxConn context.Context, in *orchestrator.SandboxDeleteR
...
+	if !marked {
+		telemetry.ReportCriticalError(ctx, "failed to mark sandbox as stopping", nil, telemetry.WithSandboxID(in.GetSandboxId()))
+
+		return nil, status.Errorf(codes.Internal, "failed to delete sandbox '%s'", in.GetSandboxId())
```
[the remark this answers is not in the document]
jakubno: This shouldn't happen as this should be guarded by API

packages/orchestrator/pkg/server/sandboxes.go
```diff
@@ -542,16 +556,16 @@ func (s *Server) Checkpoint(ctx context.Context, in *orchestrator.SandboxCheckpo
...
+	if !ok {
+		telemetry.ReportCriticalError(ctx, "sandbox not found", nil)
+
+		return nil, status.Error(codes.NotFound, "sandbox not found")
 	}
 
-	sbx, err := s.acquireSandboxForSnapshot(ctx, in.GetSandboxId())
-	if err != nil {
-		return nil, err
+	// Check envd version before acquiring (which removes the sandbox from the map).
```
dobrac: this comment is probably stale

dobrac (approved):

packages/orchestrator/pkg/server/sandboxes.go
```diff
@@ -560,6 +574,13 @@ func (s *Server) Checkpoint(ctx context.Context, in *orchestrator.SandboxCheckpo
 	}
 	defer s.startingSandboxes.Release(1)
 
+	marked := s.sandboxFactory.Sandboxes.MarkStopping(ctx, sbx.Runtime.SandboxID, sbx.LifecycleID)
+	if !marked {
+		telemetry.ReportCriticalError(ctx, "failed to mark sandbox as stopping", nil)
+
+		return nil, status.Errorf(codes.Internal, "failed to checkpoint sandbox '%s'", in.GetSandboxId())
+	}
```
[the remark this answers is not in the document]
jakubno: it shouldn't happen

PR #2368, opened 2026-04-13
feat(api): make Nomad node pool configurable

chencjcj:
## Summary

The Nomad node pool name is currently hardcoded as `"default"` in the API's orchestrator client . This PR makes it configurable via the `NOMAD_NODE_POOL` environment variable, enabling deployments that use custom Nomad node pools.

## Changes

- Add `NomadNodePool` config field (`NOMAD_NODE_POOL` env var, defaults to `"default"`)
- Replace hardcoded `"default"` pool filter in `listNomadNodes` with the configurable value
- Pass `NOMAD_NODE_POOL` from Terraform (`var.orchestrator_node_pool`) to the API job in both GCP and AWS providers
- Update tests to include the new field

### DIFF
--- a/iac/provider-aws/nomad/main.tf
+++ b/iac/provider-aws/nomad/main.tf
@@ -141,6 +141,7 @@ module "api" {
   auth_db_min_idle_connections   = var.auth_db_min_idle_connections
 
   job_env_vars = {
+    NOMAD_NODE_POOL               = var.orchestrator_node_pool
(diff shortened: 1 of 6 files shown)


chencjcj:
@dobrac  friendly ping — would appreciate your review when you have time. Thanks!

jakubno:
@chencjcj , thanks for you contribution, but this is currently being reworked and we're moving to different kind of nomad discovery. That should fix this as well (#3176)

PR #2371, opened 2026-04-13
chore(clickhouse): add ttl only drop parts to tables

jakubno:
This should decrease load on clickhouse as the data will be dropped once the whole partition is expired instead of per row


sitole (approved):

sitole:
for context: https://clickhouse.com/docs/use-cases/observability/clickstack/ttl

PR #2372, opened 2026-04-13
chore(orch): log the process state

jakubno:
Log the process state even if in another state than `D`

### DIFF
--- a/packages/orchestrator/pkg/sandbox/fc/process.go
+++ b/packages/orchestrator/pkg/sandbox/fc/process.go
@@ -8,11 +8,11 @@ import (
 	"io"
 	"os"
 	"os/exec"
-	"strings"
 	"sync/atomic"
 	"syscall"
 	"time"
 
+	"github.com/shirou/gopsutil/v4/process"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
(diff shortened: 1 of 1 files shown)


packages/orchestrator/pkg/sandbox/fc/process.go
```diff
@@ -670,20 +670,20 @@ func (p *Process) Stop(ctx context.Context) error {
...
 			}
 
 			state, err := getProcessState(ctx, p.cmd.Process.Pid)
 			if err != nil {
 				logger.L().Warn(ctx, "failed to get fc process state after sending SIGKILL", zap.Error(err), logger.WithSandboxID(p.files.SandboxID))
-			} else if state == "D" {
-				logger.L().Info(ctx, "fc process is in the D state after we call SIGKILL", logger.WithSandboxID(p.files.SandboxID))
 			}
 
+			logger.L().Info(ctx, "sent SIGKILL to fc process", logger.WithSandboxID(p.files.SandboxID), zap.String("state", state))
```
arkamar: I would guess that `getProcessState()` will most commonly return error because process was killed and reaped and it no longer exists (kill syscall returns when signal was delivered). Therefore I would expect `state` equal to `""` in most cases. I would suggest to change it to something like this:
```go
state, err := getProcessState(ctx, p.cmd.Process.Pid)
if err != nil {
   // undecided what exactly to do here, because `err` is mostly a good sign here original `Warn()` is a bit misleading.
} else {
   logger.L().Info(ctx, "sent SIGKILL to fc process", logger.WithSandboxID(p.files.SandboxID), zap.String("state", state))
}

**Review by arkamar — CHANGES_REQUESTED** [review]

File: packages/orchestrator/pkg/sandbox/fc/process.go (hunk 1 above)
Comment (arkamar) [reply packages/orchestrator/pkg/sandbox/fc/process.go]: I don't like `getProcessState()` function. Calling extra process for one letter from `/proc/<pid>/stat` file is overkill. `gopsutils` has `StatusWithContext()` but it maps Linux letters to strings https://github.com/shirou/gopsutil/blob/c2a1624b9f3ed0b38ad67134b93397142ed67a23/process/process.go#L609

File: packages/orchestrator/pkg/sandbox/fc/process.go [hunk 2]
```diff
@@ -670,18 +662,22 @@ func (p *Process) Stop(ctx context.Context) error {
...
-			err := p.cmd.Process.Kill()
-			if err != nil {
-				logger.L().Warn(ctx, "failed to send SIGKILL to fc process", zap.Error(err), logger.WithSandboxID(p.files.SandboxID))
-			} else {
-				logger.L().Info(ctx, "sent SIGKILL to fc process because it was not responding to SIGTERM for 10 seconds", logger.WithSandboxID(p.files.SandboxID))
+			// Check process status right before Kill — the pre-SIGTERM status
+			// captured above is 10s stale and no longer useful here.
+			status, stateErr := getProcessStatus(p.cmd.Process.Pid)
+			if stateErr != nil && !errors.Is(stateErr, process.ErrorProcessNotRunning) {
+				logger.L().Warn(ctx, "failed to get fc process status before SIGKILL", zap.Error(stateErr), logger.WithSandboxID(p.files.SandboxID))
```
arkamar, line 672: Should we `return` here? The `kill` will fail as well, because the process is not running anymore.

arkamar (approved):
It is probably ok to leave it as it is, but I dropped the comment for the record. The chance, that the process will exit literally right after the timer event should be close to zero.

PR #2377, opened 2026-04-13
fix(api): cancel builds on template delete

jakubno:
Cancel active template builds when deleting template

### DIFF
--- a/packages/db/queries/templates/delete_template.sql
+++ b/packages/db/queries/templates/delete_template.sql
@@ -1,18 +1,28 @@
 -- name: DeleteTemplate :many
--- Deletes a template and returns its alias cache keys for cache invalidation.
--- Alias keys are captured via CTE before the cascade delete removes them.
+-- Deletes a template and returns alias cache keys and active builds.
+-- Both are captured via CTEs before the cascade delete removes them.
 WITH alias_keys AS (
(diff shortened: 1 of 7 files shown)


packages/db/queries/templates/delete_template.sql
```diff
@@ -1,18 +1,28 @@
...
+  AND envs_del.team_id = @team_id
+  RETURNING envs_del.id
 )
-SELECT alias_key FROM alias_keys
-WHERE EXISTS (SELECT 1 FROM deleted);
\ No newline at end of file
+SELECT alias_key, NULL::uuid AS build_id, NULL::uuid AS cluster_id, NULL::text AS cluster_node_id
+FROM alias_keys WHERE EXISTS (SELECT 1 FROM deleted)
+UNION ALL
+SELECT ''::text AS alias_key, build_id, cluster_id, cluster_node_id
```
dobrac, line 27: is the alias_key here not implemented?

dobrac (changes requested):

packages/db/queries/templates/delete_template.sql
jakubno: it's hack-y solution as we have two distinct lists returned from 1 query

PR #2378, opened 2026-04-13
perf(storage): pool multipart upload chunk buffers

ValentaTomas:
## Summary
- Add `sync.Pool` for the 50 MiB chunk buffers used during GCS multipart uploads
- Each upload part previously called `make([]byte, 50MiB)` — with concurrent parts via errgroup, this was the primary source of heap growth (~1.6 GB on prod, +136 MB in a single staging capture window)
- Pooled buffers are returned immediately after `uploadPart` completes; peak usage stays bounded at `maxConcurrency × 50 MiB` but subsequent uploads reuse buffers instead of pressuring the GC

## Test plan
- [x] Existing `TestMultipartUploader_*` tests pass (including stress, chaos, concurrency, boundary tests)
- [x] Full `packages/shared/pkg/storage/...` test suite passes
- [x] Full orchestrator and API builds pass

### DIFF
--- a/packages/shared/pkg/storage/gcp_multipart.go
+++ b/packages/shared/pkg/storage/gcp_multipart.go
@@ -29,6 +29,13 @@ const (
 	gcpMultipartUploadChunkSize = 50 * 1024 * 1024 // 50MB chunks
 )
 
+var chunkPool = sync.Pool{
+	New: func() any {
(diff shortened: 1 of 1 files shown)


ValentaTomas:
Superseded by #2381 — full rewrite with zero-copy slice-based upload, HTTP/2 transport, no MD5, no buffers.

PR #2381, opened 2026-04-13
perf(storage): rewrite multipart upload — zero-copy, HTTP/2, no MD5

ValentaTomas:
## Summary
- Rewrite GCS multipart upload as zero-copy slice-based package (`gcpmultipart`)
- Remove MD5 per-part, remove buffer allocations — `Upload(data []byte)` slices directly into caller's mmap'd data
- HTTP/2 transport with connection reuse (single TLS handshake for all parts)
- Add `StoreData(ctx, []byte)` to skip `os.ReadFile` for mmap'd cache files
- Memfile/rootfs uploads now go straight from `block.Cache` mmap → HTTP body

Supersedes #2378


jakubno (changes requested):
Can we separate this into multiple PRs? 🙏🏻

ValentaTomas:
Superseded by smaller split PRs: removing MD5/mutex first, then zero-copy upload path as follow-up.

PR #2386, opened 2026-04-14
Replace bits-and-blooms/bitset with roaring bitmaps

ValentaTomas:
Replace all direct usage of `bits-and-blooms/bitset` with `roaring/v2` (v2.17.0). Uses the new `Ranges()` iterator ([#522](https://github.com/RoaringBitmap/roaring/pull/522)) for range iteration in `CreateMapping` and `BitmapRanges`.

### DIFF
--- a/packages/orchestrator/go.mod
+++ b/packages/orchestrator/go.mod
@@ -17,6 +17,7 @@ require (
 	cloud.google.com/go/storage v1.59.2
 	connectrpc.com/connect v1.18.1
 	github.com/Merovius/nbd v0.0.0-20240812113926-fd65a54c9949
+	github.com/RoaringBitmap/roaring/v2 v2.18.0
 	github.com/aws/aws-sdk-go-v2/config v1.32.6
 	github.com/aws/aws-sdk-go-v2/credentials v1.19.6
 	github.com/aws/aws-sdk-go-v2/service/ecr v1.44.0
@@ -97,7 +98,6 @@ require (
 	github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/metric v0.54.0 // indirect
 	github.com/GoogleCloudPlatform/opentelemetry-operations-go/internal/resourcemapping v0.54.0 // indirect
 	github.com/Microsoft/go-winio v0.6.2 // indirect
-	github.com/RoaringBitmap/roaring/v2 v2.16.1 // indirect
 	github.com/andybalholm/brotli v1.2.0 // indirect
 	github.com/armon/go-metrics v0.4.1 // indirect
 	github.com/aws/aws-sdk-go-v2 v1.41.0 // indirect
--- a/packages/orchestrator/pkg/sandbox/block/cache.go
+++ b/packages/orchestrator/pkg/sandbox/block/cache.go
@@ -13,16 +13,16 @@ import (
 	"syscall"
 	"time"
 
-	"github.com/bits-and-blooms/bitset"
+	"github.com/RoaringBitmap/roaring/v2"
 	"github.com/edsrzf/mmap-go"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.uber.org/zap"
 	"golang.org/x/sys/unix"
 
-	"github.com/e2b-dev/infra/packages/shared/pkg/atomicbitset"
 	"github.com/e2b-dev/infra/packages/shared/pkg/logger"
 	"github.com/e2b-dev/infra/packages/shared/pkg/storage/header"
+	"github.com/e2b-dev/infra/packages/shared/pkg/syncroaring"
 	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
 )
 
@@ -53,7 +53,7 @@ type Cache struct {
 	blockSize int64
 	mmap      *mmap.MMap
 	mu        sync.RWMutex
-	dirty     *atomicbitset.Bitset
+	dirty     *syncroaring.Bitset
 	dirtyFile bool
 	closed    atomic.Bool
 }
@@ -72,7 +72,7 @@ func NewCache(size, blockSize int64, filePath string, dirtyFile bool) (*Cache, e
 			size:      size,
 			blockSize: blockSize,
 			dirtyFile: dirtyFile,
-			dirty:     atomicbitset.New(),
+			dirty:     syncroaring.New(),
 		}, nil
 	}
 
@@ -97,7 +97,7 @@ func NewCache(size, blockSize int64, filePath string, dirtyFile bool) (*Cache, e
 		size:      size,
 		blockSize: blockSize,
 		dirtyFile: dirtyFile,
-		dirty:     atomicbitset.New(),
+		dirty:     syncroaring.New(),
 	}, nil
 }
 
@@ -117,7 +117,7 @@ func (c *Cache) ExportToDiff(ctx context.Context, out *os.File) (*header.DiffMet
 	}
 
 	if c.mmap == nil {
-		return header.NewDiffMetadata(c.blockSize, bitset.New(0)), nil
+		return header.NewDiffMetadata(c.blockSize, roaring.New()), nil
 	}
 
 	f, err := os.Open(c.filePath)
@@ -136,7 +136,7 @@ func (c *Cache) ExportToDiff(ctx context.Context, out *os.File) (*header.DiffMet
 		logger.L().Warn(ctx, "error syncing file", zap.Error(err))
 	}
 
-	diffMetadata := header.NewDiffMetadata(c.blockSize, c.dirty.BitSet())
+	diffMetadata := header.NewDiffMetadata(c.blockSize, c.dirty.Clone())
 
 	dst := int(out.Fd())
 	var writeOffset int64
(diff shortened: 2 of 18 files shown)


packages/orchestrator/go.mod
```diff
@@ -13,10 +13,13 @@ tool github.com/oapi-codegen/oapi-codegen/v2/cmd/oapi-codegen
 
 tool github.com/vektra/mockery/v3
 
+replace github.com/RoaringBitmap/roaring/v2 => github.com/e2b-dev/roaring/v2 v2.16.1-0.20260414062913-60ea6a744802
```
jakubno: is this needed?

packages/shared/go.mod
```diff
@@ -4,17 +4,18 @@ go 1.25.4
 
 tool github.com/go-swagger/go-swagger/cmd/swagger
 
+replace github.com/RoaringBitmap/roaring/v2 => github.com/e2b-dev/roaring/v2 v2.16.1-0.20260414062913-60ea6a744802
```
jakubno: is this needed?

packages/orchestrator/go.mod
ValentaTomas: Yes, waiting for them to merge my bugfix upstream.

ValentaTomas: https://github.com/RoaringBitmap/roaring/pull/524

packages/shared/go.mod
ValentaTomas: Same as https://github.com/e2b-dev/infra/pull/2386#discussion_r3077764660

packages/orchestrator/pkg/sandbox/block/tracker.go
```diff
@@ -70,11 +62,11 @@ func (t *Tracker) Offsets() iter.Seq[int64] {
 	t.mu.RLock()
 	defer t.mu.RUnlock()
 
-	return bitsetOffsets(t.b.Clone(), t.BlockSize())
-}
+	snapshot := t.b.Clone()
```
jakubno, line 65: why do we need to clone here?

packages/envd/internal/services/cgroups/cgroup2_test.go
```diff
@@ -39,7 +39,7 @@ func TestCgroupRoundTrip(t *testing.T) {
 		return
 	}
 
-	maxTimeout := time.Second * 5
+	maxTimeout := time.Second * 15
```
levb: unrelated to the change, right?

packages/orchestrator/pkg/sandbox/fc/client.go
```diff
@@ -451,8 +451,8 @@ func (c *apiClient) memoryInfo(ctx context.Context, blockSize int64) (*header.Di
 	}
 
 	return &header.DiffMetadata{
-		Dirty:     bitset.From(res.Payload.Resident),
-		Empty:     bitset.From(res.Payload.Empty),
+		Dirty:     roaring.FromDense(res.Payload.Resident, false),
```
jakubno, line 454: could you explain why with `doCopy=false` and why you don't call `RunOptimize`

packages/orchestrator/pkg/sandbox/uffd/noop.go
```diff
@@ -38,14 +38,14 @@ func (m *NoopMemory) DiffMetadata(ctx context.Context, f *fc.Process) (*header.D
...
 
 	numberOfPages := header.TotalBlocks(m.size, m.blockSize)
 
-	empty := bitset.New(uint(numberOfPages))
-	empty.FlipRange(0, uint(numberOfPages))
+	allPages := roaring.New()
+	allPages.AddRange(0, uint64(numberOfPages))
 
-	empty = empty.Difference(dirty)
+	empty := roaring.AndNot(allPages, dirty)
```
jakubno: wouldn't this work better?
```suggestion
	empty := roaring.FlipInt(dirty, 0, uint64(numberOfPages))
```

packages/orchestrator/pkg/sandbox/block/cache.go
```diff
@@ -136,7 +136,7 @@ func (c *Cache) ExportToDiff(ctx context.Context, out *os.File) (*header.DiffMet
 		logger.L().Warn(ctx, "error syncing file", zap.Error(err))
 	}
 
-	diffMetadata := header.NewDiffMetadata(c.blockSize, c.dirty.BitSet())
+	diffMetadata := header.NewDiffMetadata(c.blockSize, c.dirty.Bitmap())
```
levb: 0/5 nit: Do we need to clone under the RLock here, or (since we are holding `c.mu` - maybe we can operate on `dirty`'s data directly here?

packages/orchestrator/pkg/sandbox/uffd/noop.go
```diff
@@ -38,14 +38,14 @@ func (m *NoopMemory) DiffMetadata(ctx context.Context, f *fc.Process) (*header.D
 		return nil, err
 	}
 
-	dirty := diffInfo.Dirty.Difference(diffInfo.Empty)
+	dirty := roaring.AndNot(diffInfo.Dirty, diffInfo.Empty)
```
jakubno: use
```suggestion
diffInfo.Dirty.AndNot(diffInfo.Empty)
```
to prevent allocating an extra bitmap

then in 
```
header.DiffMetadata{
		Dirty:     diffInfo.Dirty,
```

jakubno (changes requested):

packages/orchestrator/pkg/sandbox/uffd/userfaultfd/helpers_test.go
```diff
@@ -81,15 +80,22 @@ func (h *testHandler) executeWrite(ctx context.Context, op operation) error {
...
 
 	for _, operation := range ops {
 		if operation.mode&m != 0 {
-			b.Set(uint(operation.offset))
+			b.Add(uint32(operation.offset))
 		}
 	}
 
-	return slices.Collect(b.EachSet())
+	result := make([]uint, 0, b.GetCardinality())
```
levb, line 92: 0/5 nit: This function is called only in tests, should it be moved there? I flagged the allocation, but it's fine for the use case.

packages/shared/pkg/syncroaring/bitset.go
```diff
@@ -3,8 +3,7 @@ package atomicbitset
 import (
 	"sync"
 
-	roaring "github.com/RoaringBitmap/roaring/v2"
-	"github.com/bits-and-blooms/bitset"
+	"github.com/RoaringBitmap/roaring/v2"
```
levb, line 6: 0/5 now that it is always roaring and we intermix it with explicit roaring sets in the code, rename the package to `syncroaring`? the generic `atomicbitset` is no longer intention-revealing?

packages/shared/pkg/syncroaring/bitset.go
```diff
@@ -32,9 +31,9 @@ func (b *Bitset) SetRange(start, end uint64) {
 	b.bm.AddRange(start, end)
 }
 
-func (b *Bitset) BitSet() *bitset.BitSet {
+func (b *Bitset) Bitmap() *roaring.Bitmap {
```
levb: See my comment in `ExportToDiff` - I question the need to do this.

levb (changes requested):
a few nits, otherwise LGTM

packages/orchestrator/pkg/sandbox/fc/client.go
ValentaTomas: The optimizations needs to run in O(n) + these maps are short lived—they are here only until we finalize the snapshot, so not copying the unrelying representation and not optimizing skips the ops on a temporary object.

packages/orchestrator/pkg/sandbox/uffd/noop.go
ValentaTomas: Nice.

packages/orchestrator/pkg/sandbox/uffd/noop.go
ValentaTomas: Nice.

packages/orchestrator/pkg/sandbox/block/tracker.go
ValentaTomas: We return an iterator so to prevent conflicts we copy. This is used only in tests now.

packages/orchestrator/pkg/sandbox/fc/client.go
jakubno: the payload isn't used anywhere else right?

Would it make sense to parse directly to Bitmap?

jakubno (approved):
Nice! 🚀  1 suggestion left in comments

packages/orchestrator/pkg/sandbox/fc/client.go
ValentaTomas: The `FromDense`+`doCopy=false` should parse as directly to the bitset as possible.
Or do you mean just use the old bitmap we used before?

jakubno: I meant making the http client to parse directly to Bitmap, not sure if possible though

packages/orchestrator/pkg/sandbox/uffd/noop.go
```diff
@@ -38,18 +39,16 @@ func (m *NoopMemory) DiffMetadata(ctx context.Context, f *fc.Process) (*header.D
...
+	dirty := diffInfo.Dirty.Raw()
+	dirty.AndNot(diffInfo.Empty.Raw())
 
 	numberOfPages := header.TotalBlocks(m.size, m.blockSize)
 
-	empty := bitset.New(uint(numberOfPages))
-	empty.FlipRange(0, uint(numberOfPages))
-
-	empty = empty.Difference(dirty)
+	empty := roaring.Flip(dirty, 0, uint64(numberOfPages))
```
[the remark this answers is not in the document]
ValentaTomas: > So Firecracker returns exactly ceil(total_pages / 64) uint64 words. The last word has total_pages % 64 meaningful bits, with the remaining bits being padding. The Rust side uses div_ceil(64) which means those padding bits are uninitialized/zero in normal Firecracker code (Rust's vec![0u64; n] zero-initializes).

> So in practice, padding bits are zero and the RemoveRange is defensive. But it's the right thing to do — Firecracker's spec doesn't guarantee the padding bits are zero, and FromDense faithfully converts all 64 bits per word.

> The current code is correct. No further changes needed.

levb (approved):
LGTM

PR #2393, opened 2026-04-14
perf(storage): remove unnecessary mutex from multipart upload

ValentaTomas:
Remove mutex on parts array — goroutines write to distinct indices. Supersedes #2381.

### DIFF
--- a/packages/shared/pkg/storage/gcp_multipart.go
+++ b/packages/shared/pkg/storage/gcp_multipart.go
@@ -3,8 +3,6 @@ package storage
 import (
 	"bytes"
 	"context"
-	"crypto/md5"
-	"encoding/base64"
 	"encoding/xml"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


packages/shared/pkg/storage/gcp_multipart.go
```diff
@@ -314,8 +305,6 @@ func (m *MultipartUploader) uploadParts(ctx context.Context, maxConcurrency int,
 	g, ctx := errgroup.WithContext(ctx) // Context ONLY for waitgroup goroutines; canceled after errgroup finishes
 	g.SetLimit(maxConcurrency)          // Limit concurrent goroutines
 
-	// Thread-safe map to collect parts
-	var partsMu sync.Mutex
```
levb, line 318: You're removing the mutex because each slot in the array is modified exactly once, and never read until it is returned, right?

ValentaTomas: Yes.

ValentaTomas:
We will reopen this when we merge compression as that will provide us a way to check integrity that makes this obsolete.

PR #2394, opened 2026-04-14
perf(storage): zero-copy upload path from mmap to GCS

ValentaTomas:
Upload memfile/rootfs directly from mmap to GCS, skipping CachePath→StoreFile→ReadAt→alloc-per-chunk. Stacked on #2393.


packages/orchestrator/pkg/sandbox/template_build.go
```diff
@@ -58,18 +59,46 @@ func (t *TemplateBuild) uploadMemfileHeader(ctx context.Context, h *headers.Head
...
+	}
+	if ds, ok := object.(dataStorer); ok {
+		err := ds.StoreData(ctx, data)
+		release()
+
+		return err
+	}
+
+	// Fallback: release the mmap lock and upload from the file on disk.
+	release()
```
jakubno: could we wrap the pre fallback code to separate function, you could just defer release with risk to deadlock

packages/orchestrator/pkg/sandbox/template_build.go
```diff
@@ -58,18 +59,46 @@ func (t *TemplateBuild) uploadMemfileHeader(ctx context.Context, h *headers.Head
 	return nil
 }
 
-func (t *TemplateBuild) uploadMemfile(ctx context.Context, memfilePath string) error {
-	object, err := t.persistence.OpenSeekable(ctx, t.paths.Memfile(), storage.MemfileObjectType)
+func (t *TemplateBuild) uploadDiff(ctx context.Context, diff build.Diff, path string, objectType storage.SeekableObjectType) error {
+	data, release, err := diff.Data()
```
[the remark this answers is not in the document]
jakubno: I think it would be much cleaner to call `defer` release on anything and require to pass function in any case (which you do anyway)

packages/orchestrator/pkg/sandbox/template_build.go
```diff
@@ -58,18 +59,46 @@ func (t *TemplateBuild) uploadMemfileHeader(ctx context.Context, h *headers.Head
...
 	if err != nil {
 		return err
 	}
 
-	err = object.StoreFile(ctx, memfilePath)
+	if data == nil {
+		release()
+
+		return nil
```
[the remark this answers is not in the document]
jakubno: I think this would cause object not found error

packages/shared/pkg/storage/storage_google.go
```diff
@@ -479,6 +479,74 @@ func (o *gcpObject) StoreFile(ctx context.Context, path string) (e error) {
...
+			return fmt.Errorf("failed to write data (%d bytes): %w", len(data), err)
+		}
+
+		timer.Success(ctx, int64(len(data)))
+
+		return nil
+	}
+
+	timer := googleWriteTimerFactory.Begin(
+		attribute.String(gcsOperationAttr, gcsOperationAttrWriteFromFileSystem),
```
jakubno: `gcsOperationAttrWriteFromFileSystem` is it from filesystem?

packages/orchestrator/pkg/sandbox/template_build.go
```diff
@@ -58,18 +59,46 @@ func (t *TemplateBuild) uploadMemfileHeader(ctx context.Context, h *headers.Head
...
-		return fmt.Errorf("error when uploading memfile: %w", err)
+		release()
+
+		return err
 	}
 
-	return nil
+	// Zero-copy upload from the mmap'd diff data. The wrappers (cachedSeekable,
+	// peerSeekable) forward StoreData to the inner gcpObject.
+	type dataStorer interface {
```
jakubno: this interface is defined on other places as well, can we unite it?

packages/orchestrator/pkg/sandbox/template_build.go
```diff
@@ -58,18 +59,46 @@ func (t *TemplateBuild) uploadMemfileHeader(ctx context.Context, h *headers.Head
...
+		StoreData(ctx context.Context, data []byte) error
+	}
+	if ds, ok := object.(dataStorer); ok {
+		err := ds.StoreData(ctx, data)
+		release()
+
+		return err
+	}
+
+	// Fallback: release the mmap lock and upload from the file on disk.
```
jakubno: can this even happen?

packages/orchestrator/pkg/sandbox/template/peerclient/seekable.go
```diff
@@ -130,6 +130,24 @@ func (s *peerSeekable) StoreFile(ctx context.Context, path string) error {
 	return fallback.StoreFile(ctx, path)
 }
 
+func (s *peerSeekable) StoreData(ctx context.Context, data []byte) error {
+	fallback, err := s.getOrOpenBase(ctx)
+	if err != nil {
+		return err
+	}
+
+	type dataStorer interface {
```
jakubno: again

packages/orchestrator/pkg/sandbox/block/cache.go
```diff
@@ -410,6 +410,26 @@ func (c *Cache) Path() string {
...
+		return nil, func() {}, NewErrCacheClosed(c.filePath)
+	}
+
+	if c.mmap == nil {
+		c.mu.RUnlock()
+
+		return nil, func() {}, nil
+	}
+
+	return []byte(*c.mmap), func() { c.mu.RUnlock() }, nil
```
jakubno: nit, you can do
```suggestion
	return *c.mmap, func() { c.mu.RUnlock() }, nil
```

packages/shared/pkg/storage/storage_cache_seekable.go
```diff
@@ -266,6 +266,20 @@ func (c *cachedSeekable) Size(ctx context.Context) (n int64, e error) {
 	return size, nil
 }
 
+// StoreData forwards to the inner storage provider for zero-copy upload from mmap'd data.
+func (c *cachedSeekable) StoreData(ctx context.Context, data []byte) error {
```
jakubno: if we ever want to start saving data to outer layer (NFS cache) this path is missing that

packages/shared/pkg/storage/storage_cache_seekable.go
```diff
@@ -266,6 +266,20 @@ func (c *cachedSeekable) Size(ctx context.Context) (n int64, e error) {
...
+	type dataStorer interface {
+		StoreData(ctx context.Context, data []byte) error
+	}
+
+	ds, ok := c.inner.(dataStorer)
```
jakubno: can this happen?

packages/shared/pkg/storage/storage_cache_seekable.go
```diff
@@ -266,6 +266,20 @@ func (c *cachedSeekable) Size(ctx context.Context) (n int64, e error) {
...
+	if !ok {
+		return fmt.Errorf("inner storage does not support StoreData")
```
jakubno: if yes, isn't there also possibility to fallback?

packages/orchestrator/pkg/sandbox/template/peerclient/seekable.go
```diff
@@ -130,6 +130,24 @@ func (s *peerSeekable) StoreFile(ctx context.Context, path string) error {
...
+		StoreData(ctx context.Context, data []byte) error
+	}
+
+	ds, ok := fallback.(dataStorer)
+	if !ok {
+		return fmt.Errorf("base storage does not support StoreData")
```
jakubno: same here, can it happen, is there possible fallback?

jakubno (changes requested):

packages/shared/pkg/storage/storage_cache_seekable.go
ValentaTomas: I think the experiment with this didn't much work out as the NFS cache can't scale in these situations, so we will probably solve that differently.

ValentaTomas:
https://github.com/e2b-dev/infra/pull/2034 might change the interface, so we are temporarily closing this for now.

ValentaTomas:
Closed for now, will be revisited later.

PR #2395, opened 2026-04-14
perf(nbd): zero-alloc, zero-copy synchronous dispatch

ValentaTomas:
## Summary

- Replace async goroutine-per-request NBD dispatch with a synchronous loop using mmap slices directly
- **Reads**: `Overlay.ReadSlices` returns per-block mmap slices (overlay cache or template cache), written to the NBD socket via `writev` (`net.Buffers`) — zero user-space copies, zero allocations after warmup
- **Writes**: `Cache.WriteSlice` returns a writable mmap reference; socket data is read directly into the mmap page — zero user-space copies, zero allocations
- Removes per-request: goroutines, channels, `sync.Pool`, write lock, WaitGroup, and shutting-down machinery

### Profile context

From profiling on n2-highmem-80 nodes, NBD dispatch was the dominant allocator after the multipart upload fix (#2394):

| Path | Cumulative allocs | What it was |
|---|---|---|
| `Dispatch.Handle` (write buffers) | 57 TB (47%) | `make([]byte, request.Length)` per write |
| `cmdRead.func1` (read buffers) | 20 TB (16%) | `make([]byte, length)` per read |

This PR eliminates both — zero allocations after the first request at each size.

### How it works

**Reads**: Instead of allocating a buffer, copying mmap data into it, then writing to socket — we collect per-block mmap slice references and send them via `writev` in a single syscall.

**Writes**: Instead of allocating a buffer, reading socket data into it, then copying into mmap — we get a writable reference to the overlay cache mmap and read socket data directly into it.

Parallelism is preserved via `FlagCanMulticonn` (multiple NBD socket connections per device), rather than per-request goroutines within a connection.

## Test plan

- [x] `go build ./...` passes
- [x] `go vet` passes
- [ ] Existing NBD tests pass (require root / nbd kernel module)
- [ ] Profile on staging to verify allocation elimination


ValentaTomas:
Closing for now as this is a more invasive change.

PR #2406, opened 2026-04-15
feat(orch): support rootfs resizing when building from template

tvi:
Previously, building a template from another template (FromTemplate) always reused the source template's build artifacts as-is, with no ability to change the disk size. The rootfs size was baked into the header at base layer creation time and copied identically through every subsequent diff generation. Since resize2fs requires a local ext4 file, there was no way to resize the diff-chain-based virtual block device.

This commit modifies the FromTemplate path in the base build phase to support disk resizing by materializing the source template's rootfs into a new ext4 file:

1. The source template's rootfs is loaded from storage via NBD (same mechanism as cmd/mount-build-rootfs) and mounted read-only.
2. A fresh ext4 filesystem is created at the target size.
3. All files are copied from source to destination via rsync.
4. Fresh provisioning files (envd, busybox, provision script, system configs) are written to overwrite any stale versions from the source template.
5. The ext4 is shrunk, resized to the target DiskSizeMB, and integrity-checked.
6. The sandbox is re-provisioned (systemd install via busybox init).
7. A new snapshot layer is created (systemd boot, pause, upload) -- same flow as buildLayerFromOCI.

The FromTemplate Layer() method now checks the hash-based layer cache (which already includes DiskSizeMB in its hash key) instead of always returning Cached: true. Repeated builds with the same from-template + disk-size combination remain cached.

Structural changes:

- Extract NBD utilities (Cleaner, BuildDevice, TemplateRootfs, GetNBDDevice, MountNBDDevice) from pkg/sandbox/nbd/testutils into a new production package pkg/sandbox/nbd/nbdutil. The testutils package now re-exports from nbdutil for backward compatibility. Test-only types (ZeroDevice, LoggerOverlay) remain in testutils.

- Export ProvisioningFiles() from pkg/template/build/core/rootfs to allow both the OCI path (as OCI tar layers) and the new template path (as direct disk writes) to share the same file set. The existing additionalOCILayers() now delegates to ProvisioningFiles() internally.

- Add template_rootfs.go in the base build phase containing: buildLayerFromTemplate(), materializeTemplateRootfs(), copyFilesRsync(), and writeProvisioningFiles().


ValentaTomas:

PR #2407, opened 2026-04-15
Remount volumes on resume

djeebus:
Without the remount, NFS handles will all be stale (pointing to released mountnses) or invalid (if we resume on a different orchestrator)

### DIFF
--- a/packages/envd/internal/api/init.go
+++ b/packages/envd/internal/api/init.go
@@ -224,7 +224,9 @@ func (a *API) SetData(ctx context.Context, logger zerolog.Logger, data PostInitJ
 	}
 
 	if data.VolumeMounts != nil {
-		a.setupNFS(ctx, logger, data.LifecycleID, *data.VolumeMounts)
+		if err := a.setupNFS(ctx, logger, data.LifecycleID, *data.VolumeMounts); err != nil {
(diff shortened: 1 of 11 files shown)


packages/envd/internal/api/init.go
```diff
@@ -264,7 +261,22 @@ func (a *API) setupNFS(ctx context.Context, logger zerolog.Logger, mounts []Volu
 	}
 	defer a.isMountingNFS.Store(false)
 
-	logger.Debug().Msg("Mounting NFS volumes")
+	// Check if lifecycleID changed (indicates resume to potentially different server)
+	prevLifecycleID, _ := a.mountedLifecycleID.Load().(string)
+	lifecycleChanged := prevLifecycleID != "" && prevLifecycleID != lifecycleID
+
+	if lifecycleChanged {
```
dobrac: should we also invalidate the isMountedNFS?

dobrac (approved):

packages/envd/internal/api/init.go
djeebus: We only need to change that if we unmount the volumes, which we do on line 318 ... although it's only doing it for a single path. Don't think that's right. I'll clean that up.

dobrac (approved):

PR #2411, opened 2026-04-16
fix(orchestrator): add 5s buffer before network slot return and sandbox map release

jakubno:
Add a delay before returning the network slot to the pool during sandbox cleanup to allow in-flight requests to complete. Since the cleanup chain runs in LIFO order, this also delays the sandbox map removal.


dobrac (approved):
This will prevent the slot return, but it won't fix so that the sandbox map still has the mapping there if I'm not mistaken

PR #2414, opened 2026-04-16
fix(dashboard-api): bootstrap users with admin token only

ben-fornefeld:
## Summary
- change dashboard bootstrap to `POST /admin/users/{userId}/bootstrap` so the target user comes from the URL path
- remove `Supabase1TokenAuth` from that endpoint so `X-Admin-Token` alone is a valid auth method
- regenerate the dashboard OpenAPI bindings and add a request-path test proving admin-token-only auth reaches the handler

## Validation
- `go test ./packages/dashboard-api/internal/api ./packages/dashboard-api/internal/handlers ./packages/auth/pkg/auth`


packages/dashboard-api/internal/handlers/team_provisioning.go
```diff
@@ -40,12 +40,11 @@ type provisionedTeam struct {
...
-func (s *APIStore) PostAdminUsersBootstrap(c *gin.Context) {
+func (s *APIStore) PostAdminUsersUserIdBootstrap(c *gin.Context, userId api.UserId) {
 	ctx := c.Request.Context()
 	telemetry.ReportEvent(ctx, "bootstrap user")
 
-	userID := auth.MustGetUserID(c)
-	team, err := s.bootstrapUser(ctx, userID)
+	team, err := s.bootstrapUser(ctx, userId)
 	if err != nil {
 		s.handleProvisioningError(ctx, c, "bootstrap user", err)
```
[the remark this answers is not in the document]
ben-fornefeld: Fixed in `67bc33bd8`. `bootstrapUser` now maps a missing `auth.users` row to a `ProvisionError` with HTTP 404, and I added a handler regression test covering `POST /admin/users/{userId}/bootstrap` for a non-existent UUID.

packages/dashboard-api/internal/handlers/team_provisioning.go
```diff
@@ -96,6 +95,12 @@ func (s *APIStore) PostTeams(c *gin.Context) {
 
 func (s *APIStore) bootstrapUser(ctx context.Context, userID uuid.UUID) (provisionedTeam, error) {
 	authUser, err := s.supabaseDB.Write.GetAuthUserByID(ctx, userID)
+	if dberrors.IsNotFoundError(err) {
+		return provisionedTeam{}, &internalteamprovision.ProvisionError{
+			StatusCode: http.StatusNotFound,
+			Message:    "User not found",
+		}
+	}
```
dobrac, line 103: this is 500, right?

dobrac (approved):
small nit

packages/dashboard-api/internal/handlers/team_provisioning.go
ben-fornefeld: disagree. 500 would be in broader context acceptable. localized 404 is the right status code to return, how we react to it in other systems is more important.

PR #2423, opened 2026-04-17
perf(sandbox): reduce memory dirtying from journald and envd logging

ValentaTomas:
## Summary
- Add journald `Storage=none` drop-in inside guest VMs to prevent journal writes from dirtying memory pages during pause-resume cycles
- Remove stdout writer from envd in FC mode — debug logs still ship via HTTP exporter to the orchestrator, but never touch journald/disk inside the VM
- Add `measure-memory-dirtying.sh` benchmark script for measuring diff sizes across pause-resume cycles

## Motivation
During frequent pause-resume cycles, journald and envd stdout were constantly writing to memory, dirtying pages that then had to be captured in every snapshot diff layer. This increased layer sizes and slowed snapshot restore.

## Changes
- **`journald.conf.tpl`** (new): `Storage=none`, `MaxLevelConsole=warning`, `MaxLevelKMsg=warning` — system errors still reach the kernel console, but nothing is stored in journal files
- **`logger.go`**: In FC mode, envd writes only to the HTTP exporter (debug level preserved). In non-FC mode, stdout + debug unchanged
- **`envd version`**: Bumped to 0.5.15
- **`rootfs_test.go`**: Updated file count and added assertions for journald conf
- **`measure-memory-dirtying.sh`**: Benchmark script that chains build → pause-resume cycles and prints memfile/rootfs diff sizes

## Test plan
- [ ] Unit test passes (`TestAdditionalOCILayers`)
- [ ] Run benchmark on staging machine with GCS access to measure before/after diff sizes
- [ ] Verify envd debug logs still arrive at orchestrator via HTTP exporter

### DIFF
--- a/packages/envd/internal/logs/logger.go
+++ b/packages/envd/internal/logs/logger.go
@@ -18,18 +18,21 @@ func NewLogger(ctx context.Context, isNotFC bool, mmdsChan <-chan *host.MMDSOpts
 
 	exporters := []io.Writer{}
 
+	var level zerolog.Level
 	if isNotFC {
 		exporters = append(exporters, os.Stdout)
 	} else {
-		exporters = append(exporters, exporter.NewHTTPLogsExporter(ctx, isNotFC, mmdsChan), os.Stdout)
+		// HTTP exporter only — stdout goes to journald which dirties guest memory pages.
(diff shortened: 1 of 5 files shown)


ValentaTomas:
After running some benchmarks, the gains from this are less that idea, so closing for now.

PR #2445, opened 2026-04-18
refactor(uffd): drain pagefault events into a batch per poll cycle

ValentaTomas:
Extracted from #1896. Pure structural refactor of the UFFD event loop, no behaviour change.

`Userfaultfd.Serve` previously read exactly one event per poll/read cycle. This drains all queued events into a slice (until EAGAIN) and processes them as a batch in the same iteration.

Why: the upcoming `UFFD_EVENT_REMOVE` handling needs REMOVE and PAGEFAULT events from the same poll cycle delivered together, so REMOVEs can be applied to the page tracker before the matching faults run.

Same EINTR/EAGAIN handling, same WRITE/read branching, same `wg.Go` dispatch, same `faultPage` signature, same eagain/noData counters. The only structural addition is the outer `for _, pagefault := range pagefaults` loop, which accounts for most of the diff (indentation churn).

Note: the `arm64-tests / packages/shared` failure is the unrelated `TestChangeResponseHeader` flake fixed by #2443; this PR doesn't touch `packages/shared`.


packages/orchestrator/pkg/sandbox/uffd/userfaultfd/userfaultfd.go
```diff
@@ -201,71 +198,71 @@ outerLoop:
...
-			u.logger.Error(ctx, "uffd: read error", zap.Error(err))
+			msg := *(*UffdMsg)(unsafe.Pointer(&buf[0]))
 
-			return fmt.Errorf("failed to read: %w", err)
-		}
+			if msgEvent := getMsgEvent(&msg); msgEvent != UFFD_EVENT_PAGEFAULT {
+				u.logger.Error(ctx, "UFFD serve unexpected event type", zap.Any("event_type", msgEvent))
 
-		msg := *(*UffdMsg)(unsafe.Pointer(&buf[0]))
+				return ErrUnexpectedEventType
```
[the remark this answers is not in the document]
ValentaTomas: If we encounter a non-page-fault error here, we should just discard the FC. The previous behavior could potentially serve a few pages more, before closing the whole uffd anyway, so next errors would probably forever hang too, so there should not be an actual correctness change.

packages/orchestrator/pkg/sandbox/uffd/userfaultfd/userfaultfd.go
```diff
@@ -107,9 +107,6 @@ func (u *Userfaultfd) Serve(
 		{Fd: fdExit.Reader(), Events: unix.POLLIN},
 	}
 
-	eagainCounter := newCounterReporter(u.logger, "uffd: eagain during fd read (accumulated)")
-	defer eagainCounter.Close(ctx)
```
dobrac, line 114: do we not care about the counts here anymore?

dobrac (approved):

packages/orchestrator/pkg/sandbox/uffd/userfaultfd/userfaultfd.go
ValentaTomas: The EAGAIN now actually stops the FD reading and continues to serve the pagefaults that accumulated, as the EAGAIN is the error that the fd in this case should return when it reads all the pagefaults.
If we let it there it would essentially degenerate to printing "1" in all cases.

dobrac (approved):

PR #2452, opened 2026-04-20
fix(aws): require IMDSv2 for node metadata access

chux0519:
## Summary

This PR hardens the AWS provider node bootstrap flow by making EC2 metadata access work with IMDSv2 and by requiring metadata tokens on AWS node launch templates.

It also includes a small ClickHouse bootstrap compatibility fix for `nvme list` output variants.

## Motivation

The AWS provider bootstrap scripts read EC2 instance metadata during user data execution to discover the instance ID, region, availability zone, private IP, and EC2 tags. These values are used to configure
Nomad and Consul.

The previous implementation used unauthenticated metadata requests:

```bash
curl http://169.254.169.254/latest/meta-data/...
```

That works only when IMDSv1 is available. In AWS environments where IMDSv2 is required through launch template settings, AMI defaults, account defaults, or security baseline policies, these calls fail during
node bootstrap.

Instead of relying on environment-specific metadata defaults, the bootstrap scripts now request an IMDSv2 token and use it for metadata reads. With that in place, the launch templates can explicitly require
metadata tokens.

This makes AWS node startup less dependent on account-level or AMI-level defaults and aligns the provider with the stricter EC2 metadata posture recommended by AWS.

Separately, the ClickHouse bootstrap script maps an attached EBS volume to a local NVMe device by parsing nvme list. Some nvme-cli output variants expose the normalized EBS volume ID in the serial column
rather than the model column. Matching only one column can cause ClickHouse startup to wait until timeout even when the volume is attached.

## Changes

- request an IMDSv2 token from /latest/api/token
- pass X-aws-ec2-metadata-token to metadata and dynamic identity document requests
- set AWS node launch templates and the ClickHouse instance metadata options to http_tokens = "required"
- keep token acquisition tolerant so the scripts can still run in environments where IMDSv2 is not enforced
- match the normalized ClickHouse EBS volume ID against both observed nvme list columns

## Validation

Static checks:

```
bash -n \
  iac/provider-aws/nomad-cluster/scripts/run-nomad.sh \
  iac/provider-aws/nomad-cluster/scripts/run-consul.sh \
  iac/provider-aws/modules/nodepool-clickhouse/scripts/start-clickhouse.sh
```

```
terraform fmt -check \
  iac/provider-aws/modules/nodepool-api/main.tf \
  iac/provider-aws/modules/nodepool-clickhouse/main.tf \
  iac/provider-aws/modules/nodepool-client/main.tf \
  iac/provider-aws/modules/nodepool-control-server/main.tf
```

Runtime validation on AWS:

- applied the AWS provider Terraform changes in a dev environment
- confirmed the updated setup scripts were uploaded with new hashes
- rolled one API autoscaling group instance onto the updated launch template
- confirmed the replacement API instance launched with HttpTokens=required
- confirmed the API Nomad allocation started successfully
- confirmed the load balancer target became healthy
- confirmed the public API health endpoint returned success

Runtime result:

API node metadata option: HttpTokens=required
Instance refresh: Successful
API allocation: running
Load balancer target: healthy
API health: successful

### DIFF
--- a/iac/provider-aws/modules/nodepool-api/main.tf
+++ b/iac/provider-aws/modules/nodepool-api/main.tf
@@ -73,6 +73,10 @@ resource "aws_launch_template" "api" {
 
   vpc_security_group_ids = var.security_group_ids
 
+  metadata_options {
+    http_tokens = "required"
+  }
+
   iam_instance_profile {
     name = aws_iam_instance_profile.api.name
   }
--- a/iac/provider-aws/modules/nodepool-clickhouse/main.tf
+++ b/iac/provider-aws/modules/nodepool-clickhouse/main.tf
@@ -74,6 +74,10 @@ resource "aws_launch_template" "clickhouse" {
 
   vpc_security_group_ids = var.security_group_ids
 
+  metadata_options {
+    http_tokens = "required"
+  }
+
   iam_instance_profile {
     name = aws_iam_instance_profile.clickhouse.name
   }
@@ -110,6 +114,10 @@ resource "aws_instance" "clickhouse" {
 
   availability_zone = var.clickhouse_az
   subnet_id         = var.clickhouse_subnet_id
+
+  metadata_options {
+    http_tokens = "required"
+  }
 
   user_data = base64encode(templatefile("${local.scripts_path}/start-clickhouse.sh", {
     NODE_POOL                    = var.node_pool_name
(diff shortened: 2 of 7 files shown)


sitole (approved):

sitole:
@chux0519 thank you for your first E2B contribution.

sitole (approved):

PR #2454, opened 2026-04-20
perf: optimize error creation by using errors.New instead of fmt.Errorf

AdaAibaby:
Summary
This PR performs a large-scale refactoring across 80+ files to replace instances of fmt.Errorf containing static strings with errors.New.
Motivation
● Performance: Eliminates unnecessary format string parsing overhead for error messages that do not contain dynamic variables. While the impact per call is small, it reduces CPU cycles in hot paths across our infrastructure services (Orchestrator, API, Sandbox).
● Code Clarity: Using errors.New explicitly signals that the error message is constant. It improves readability by removing visual noise associated with formatting verbs where none are needed.
● Idiomatic Go: Aligns with Go best practices which suggest using errors.New for simple error creation.
Changes
● Replaced fmt.Errorf("static message") with errors.New("static message").
● Removed unused fmt imports where applicable.
● No functional logic changes; this is a pure refactor.
Testing
● All existing unit tests pass.
● Verified via go build that no regressions were introduced.


jakubno:
Thanks for suggestion @AdaAibaby, I enabled the linter in #2455 to enforce this + fixed some issues.

PR #2458, opened 2026-04-20
Users is now in the public schema, not auth

djeebus:


ben-fornefeld:
can we remove from the core migrations completely `e2b-dev/infra/packages/db/migrations/20000101000000_auth.sql` ?

levb (commented):
Looks simple, what's the motivation for this change?

djeebus:
> Looks simple, what's the motivation for this change?

Migrating the auth away from supabase in the OS repo means that a) users are no longer in the "auth" schema, and b) queries can no longer use the existing "auth"."user" table

packages/db/pkg/auth/queries/get_user.sql.go
```diff
@@ -12,12 +12,17 @@ import (
 )
 
 const getUser = `-- name: GetUser :one
-SELECT id, email, created_at FROM "auth"."users" where id = $1
+SELECT created_at, updated_at, id, email FROM "public"."users" where id = $1
```
levb: Was adding `updated_at` intentional?

levb (approved):

packages/db/pkg/auth/queries/get_user.sql.go
djeebus: It is ... expected, but not explicitly desired :sweat_smile: Let me see if I can slim down this query

PR #2459, opened 2026-04-20
Add vars to template build explicitly

djeebus:


.github/workflows/deploy-base-template.yml
```diff
@@ -49,5 +49,7 @@ jobs:
           bun run ./build.prod.ts
         env:
           E2B_API_KEY: ${{ secrets.E2B_API_KEY }}
+          E2B_DEBUG: "false"
```
levb, line 52: 0/5 What's wrong with more logs here?

levb (approved):

.github/workflows/deploy-base-template.yml
djeebus: It doesn't affect logs, it affects the url; E2B_DEBUG=true ignores other parameters and communicates with `http://localhost:3000`

PR #2464, opened 2026-04-21
feat(api): add PATCH /sandboxes/{id}/metadata

levb:
`PATCH /sandboxes/{id}/metadata` with merge semantics: string values upsert, `null` or `""` remove, absent keys are left alone. Works only on running sandboxes.

`RunningSandbox` gets a metadata field so List ships the live value instead of the create-time `APIStoredConfig` snapshot (modeled after startAt/endAt). `SandboxUpdateRequest` gains an optional SandboxMetadataUpdate wrapper to distinguish "don't touch" from an explicit patch.


ValentaTomas:
Still thinking about the PUT vs PATCH as it is possible that people will need to fetch the previous values to overwrite part of it now, right?

levb:
`PATCH` has its own issues. We could do both, but seems like client complexity. I am 0/5 since I don't know the use cases. `PUT` has the advantage of clarity.

ValentaTomas:
As we already don't allow nested, maybe targeted PUT or PATCH might work here?

So it is, for example something like:
```
{
  "a": ...,
  "b": ...,
}

/PATCH /sandboxes/.../metadata { "metadata": null }

->

{}

```

```
{
  "a": 1,
  "b": 2,
}

/PATCH /sandboxes/.../metadata { "metadata": { "c": 3 } }

->

{
  "a": 1,
  "b": 2,
  "c": 3,
}

```

```
{
  "a": 1,
  "b": 2,
}

/PATCH /sandboxes/.../metadata { "metadata": { "a": null } }

->

{
  "b": 2,
}

```

ValentaTomas:
I worry here, because if users already have things like user_id on the sandboxes, there *will* be cases where these ids will get wiped out and in case they stored more info there it is possible that they cannot get this info from other sources easily, apart from calling get metadata on the sandbox.

levb:
ok, will convert to PATCH, NP 👍

packages/api/internal/handlers/sandbox_metadata_patch.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+func (a *APIStore) PatchSandboxesSandboxIDMetadata(
+	c *gin.Context,
+	sandboxID string,
+) {
+	ctx := c.Request.Context()
+
+	var err error
+	sandboxID, err = utils.ShortID(sandboxID)
```
dobrac, line 23:
```suggestion
	sandboxID, err := utils.ShortID(sandboxID)
```

packages/orchestrator/orchestrator.proto
```diff
@@ -127,6 +132,10 @@ message RunningSandbox {
 
   google.protobuf.Timestamp start_time = 3;
   google.protobuf.Timestamp end_time = 4;
+
+  // Live user-facing metadata tags. Authoritative over config.metadata, which
+  // reflects only the create-time snapshot.
+  map<string, string> metadata = 5;
```
dobrac, line 138: why od we need this change?

packages/api/internal/orchestrator/patch_metadata.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func (o *Orchestrator) patchSandboxMetadataOnNode(
+	ctx context.Context,
+	sbx sandbox.Sandbox,
+	metadata map[string]string,
+) *api.APIError {
+	ctx, span := tracer.Start(ctx, "patch-sandbox-metadata-on-node",
+		trace.WithAttributes(
+			attribute.String("instance.id", sbx.SandboxID),
```
dobrac, line 83: telemetry.WithSandboxID

dobrac (changes requested):
Why don't we do the implementation similar as for the network update?

PR #2470, opened 2026-04-21
Feat/enable autoresume for byoc

matthewlouisbrockman:
Allows edge clients to check the API for autoresume using public gRPC over TLS and the oauth jwt

This got a bit larger but does two things:
- adds a route from the edge proxies to call the API to resume sandboxes when autoresume is on
- implements the oidc auth for the API so we can use the JWTs from edge <> our control plane
- new scope: `sandboxes:lifecycle`

New Variables Needed:

API:
- CLIENT_PROXY_OIDC_ISSUER_URL

BYOC:
- CLIENT_PROXY_API_EDGE_GRPC_ADDRESS=grpc-api.e2b.dev:443

When a request comes into the edge proxy, if the sandbox is not present in the edge catalog, the proxy calls home to the main API to check whether it can autoresume. If it can, the API resumes the sandbox through the BYOC orchestrator and returns the node route back to the edge proxy.

```
                           E2B CONTROL PLANE
                     +--------------------------+
                     | API                      |
                     |                          |
                     | ResumeSandbox gRPC       |
                     | api-grpc.<domain>        |
                     |                          |
                     | Auth checks:             |
                     | - Bearer OAuth token     |
                     | - iss/aud verification   |
                     | - token org_id matches   |
                     |   cluster.auth_org_id    |
                     | - forwarded request auth |
                     |   when required          |
                     +------------+-------------+
                                  ^      |
                                  |      |
                                  |      | 3. API asks BYOC orchestrator
                                  |      |    to resume/start sandbox
                                  |      v
BYOC / EDGE DATA PLANE            |  +---------------------------+
+----------------------------------|--| edge orchestrator node   |-----------+
|                                  |  | starts/resumes sandbox   |           |
|                                  |  | updates sandbox catalog  |           |
|                                  |  +------------+--------------+           |
|                                  |               |                          |
|                                  |               | resumed sandbox           |
|                                  |               v                          |
|                                  |  +---------------------------+           |
|                                  |  | sandbox app / envd        |           |
|                                  |  +---------------------------+           |
|                                  |                                          |
|  user traffic                    |                                          |
|      |                           |                                          |
|      | optional request auth:    |                                          |
|      | - traffic access token    |                                          |
|      | - envd access token       |                                          |
|      v                           |                                          |
|  +-------------------+           |                                          |
|  | BYOC LB / DNS     |           |                                          |
|  +---------+---------+           |                                          |
|            |                     |                                          |
|            v                     |                                          |
|  +-------------------+           |                                          |
|  | edge proxy        |           |                                          |
|  | client-proxy bin  |           |                                          |
|  |                   |           |                                          |
|  | API auth:         |           |                                          |
|  | OAuth2 client     |           |                                          |
|  | credentials token |           |                                          |
|  +---------+---------+           |                                          |
|            |                     |                                          |
|            | 1. lookup sandbox   |                                          |
|            v                     |                                          |
|  +-------------------+           |                                          |
|  | edge catalog      |           |                                          |
|  | sandbox -> node   |           |                                          |
|  +----+---------+----+           |                                          |
|       |         |                |                                          |
|       | hit     | miss           |                                          |
|       |         |                |                                          |
|       |         v                |                                          |
|       |   +-------------------------------+                                 |
|       |   | 2. auto-resume path           |                                 |
|       |   | edge proxy calls API          |                                 |
|       |   | ResumeSandbox(sandbox, port)  |---------------------------------+
|       |   | over gRPC + TLS               |
|       |   |                               |
|       |   | API auth metadata:            |
|       |   | authorization: Bearer <OAuth> |
|       |   |                               |
|       |   | forwarded request metadata:   |
|       |   | x-e2b-traffic-access-token    |
|       |   | x-e2b-envd-access-token       |
|       |   +---------------+---------------+
|       |                   |
|       |                   | 4. API validates OAuth org_id
|       |                   |    against cluster.auth_org_id,
|       |                   |    validates forwarded request
|       |                   |    token if required, then
|       |                   |    returns node route
|       |                   v
|       |          +-----------------------------+
|       +--------->| edge orchestrator node      |
|                  | orchestrator proxy          |
|                  +-------------+---------------+
|                                |
|                                | 5. original request continues
|                                |    to requested sandbox port
|                                v
|                  +-----------------------------+
|                  | sandbox app / envd          |
|                  +-----------------------------+
|                                                                       |
+-----------------------------------------------------------------------+
```


iac/provider-gcp/nomad-cluster/network/main.tf
```diff
@@ -66,6 +66,21 @@ locals {
...
+        protocol           = "HTTP"
+        request_path       = var.api_port.health_path
+        port               = var.api_port.port
+        timeout_sec        = 3
+        check_interval_sec = 3
+      }
+      groups = [{ group = var.api_instance_group }]
+    }
     docker-reverse-proxy = {
       protocol                        = "HTTP"
```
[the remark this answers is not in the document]
matthewlouisbrockman: k, added rate limiting on the endpoint; secret issue false positive but generally adding cloud armor rules seems good

iac/provider-aws/nomad/main.tf
```diff
@@ -98,6 +98,7 @@ module "client_proxy" {
   redis_cluster_url   = var.redis_cluster_url
   redis_tls_ca_base64 = var.redis_tls_ca_base64
   image               = data.aws_ecr_image.client_proxy.image_uri
+  api_secret          = var.api_secret
 
   otel_collector_grpc_endpoint = "localhost:${var.otel_collector_grpc_port}"
   logs_collector_address       = "http://localhost:${var.logs_proxy_port}"
```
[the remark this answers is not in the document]
matthewlouisbrockman: oh, right, aws. added

packages/client-proxy/internal/proxy/proxy.go
```diff
@@ -53,7 +53,24 @@ func catalogResolution(ctx context.Context, sandboxId string, sandboxPort uint64
...
+					}
+
+					return "", fmt.Errorf("failed to get resumed sandbox from catalog: %w", catalogErr)
+				}
+				if resumedSandbox.OrchestratorIP == "" {
+					return "", ErrNodeNotFound
+				}
+
+				return resumedSandbox.OrchestratorIP, nil
 			}
```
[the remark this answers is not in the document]
matthewlouisbrockman: it's sync, not async

sitole:
@matthewlouisbrockman, can you please share a diagram explaining wiring things together?

iac/modules/job-api/jobs/api.hcl
```diff
@@ -23,6 +23,10 @@ job "api" {
         static = "${api_grpc_port}"
       }
 
+      port "grpc-public" {
+        static = "${api_public_grpc_port}"
+      }
+
```
dobrac: do we need this if we're using traefik? If so, can the port be dynamic?

iac/modules/job-client-proxy/jobs/client-proxy.hcl
```diff
@@ -114,8 +114,15 @@ job "client-proxy" {
         REDIS_URL                = "${redis_url}"
 
         %{ if api_grpc_address != "" }
-        # used only when client-proxy is deployed directly in the cluster next to the API
-        API_GRPC_ADDRESS = "${api_grpc_address}"
+        # used by client-proxy to call API ResumeSandbox over gRPC
+        API_GRPC_ADDRESS             = "${api_grpc_address}"
+        # internal Consul addresses use plaintext; external addresses use TLS
+        API_GRPC_TLS                 = "${api_grpc_tls}"
```
dobrac: why do we need to define both? Doesn't client proxy always use the internal consul one?

iac/provider-gcp/nomad-cluster/network/main.tf
```diff
@@ -66,6 +66,21 @@ locals {
...
+      connection_draining_timeout_sec = 1
+      http_health_check = {
+        protocol           = "HTTP"
+        request_path       = var.api_port.health_path
+        port               = var.api_port.port
+        timeout_sec        = 3
+        check_interval_sec = 3
+      }
+      groups = [{ group = var.api_instance_group }]
+    }
```
dobrac: this should not be necessary as we're using ingress, right? Also, it should not route to the private api_port, but to the api_public_grpc_port

packages/api/internal/cfg/model.go
```diff
@@ -53,7 +53,12 @@ type Config struct {
 	RedisTLSCABase64 string `env:"REDIS_TLS_CA_BASE64"`
 	RedisPoolSize    int    `env:"REDIS_POOL_SIZE"     envDefault:"160"`
 
-	APIGrpcPort uint16 `env:"API_GRPC_PORT" envDefault:"5009"`
+	APIGrpcPort       uint16 `env:"API_GRPC_PORT"        envDefault:"5009"`
+	APIPublicGrpcPort uint16 `env:"API_PUBLIC_GRPC_PORT" envDefault:"5109"`
+	APISecret         string `env:"API_SECRET"`
```
dobrac: what is this for?

packages/api/internal/clusters/discovery/discovery.go
```diff
@@ -16,6 +16,10 @@ type Item struct {
 	// Instance ID that changes on each restart, available only for edge-backend service discovery.
 	InstanceID string
 
+	// IPAddress is the node IP/host, without service port, returned to edge/client proxy for sandbox routing.
+	// Remote cluster control-plane calls still go through the cluster gRPC proxy.
+	IPAddress string
```
dobrac: could you please explain the role of this?

packages/api/internal/handlers/client_proxy_oauth.go
```diff
@@ -0,0 +1,63 @@
...
+type oidcClientProxyOAuthVerifier struct {
+	verifier *oidc.IDTokenVerifier
+}
+
+func NewClientProxyOAuthVerifier(ctx context.Context, issuerURL string, audience string) (ClientProxyOAuthVerifier, error) {
+	issuerURL = strings.TrimSpace(issuerURL)
+	audience = strings.TrimSpace(audience)
+
+	if issuerURL == "" && audience == "" {
+		return nil, nil
```
dobrac: lets do noop client instead

packages/api/internal/handlers/client_proxy_oauth.go
```diff
@@ -0,0 +1,63 @@
+package handlers
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/coreos/go-oidc/v3/oidc"
```
dobrac: is the golang library `https://golang.org/x/oauth2` not enough? If not, why not using the OpenID certified one? https://github.com/zitadel/oidc

packages/api/internal/orchestrator/autoresume.go
```diff
@@ -30,13 +31,6 @@ func (o *Orchestrator) HandleExistingSandboxAutoResume(
 
 	attempts := 0
 
-	// Existing sandbox auto-resume state handling:
-	// - running: return the current node IP immediately
-	// - pausing/snapshotting: wait for the transition, refresh state, and retry
-	// - killing: treat as not found
-	// - anything else: return internal error
-	// - internal transition wait timeout: treat as "still transitioning"
-	// - caller cancellation/deadline: propagate the context error
```
dobrac, line 39: is this not valid anymore?

packages/api/internal/orchestrator/autoresume.go
```diff
@@ -100,20 +94,24 @@ func (o *Orchestrator) HandleExistingSandboxAutoResume(
...
-			if node == nil {
+			nodeIP := ""
+			if node != nil {
+				nodeIP = routeNodeIPAddress(node, env.IsLocal())
+			}
+			if nodeIP == "" {
 				logger.L().Error(
 					ctx,
-					"Sandbox is running but routing info is not available during auto-resume",
+					"Sandbox is running but orchestrator IP is not available during auto-resume",
```
dobrac: lets keep routing info

packages/api/internal/orchestrator/autoresume.go
```diff
@@ -100,20 +94,24 @@ func (o *Orchestrator) HandleExistingSandboxAutoResume(
...
 					logger.WithSandboxID(sandboxID),
 					logger.WithTeamID(teamID.String()),
 					logger.WithNodeID(sbx.NodeID),
 					zap.Stringer("cluster_id", sbx.ClusterID),
 				)
 
-				return "", false, errors.New("sandbox is running but routing info is not available yet")
+				return "", false, errors.New("sandbox is running but orchestrator IP is not available yet")
```
dobrac: lets keep routing info

packages/api/internal/orchestrator/autoresume.go
```diff
@@ -100,20 +94,24 @@ func (o *Orchestrator) HandleExistingSandboxAutoResume(
 			return "", false, apisandbox.ErrNotFound
 		case apisandbox.StateRunning:
 			node := o.getOrConnectNode(ctx, sbx.ClusterID, sbx.NodeID)
-			if node == nil {
+			nodeIP := ""
+			if node != nil {
+				nodeIP = routeNodeIPAddress(node, env.IsLocal())
```
dobrac, line 107: why is this custom compared to the normal sandbox/orchestrator communication flow?

packages/client-proxy/internal/proxy/paused_sandbox_resumer_grpc.go
```diff
@@ -17,28 +21,71 @@ import (
...
+
+func (c GrpcOAuthConfig) Enabled() bool {
+	return strings.TrimSpace(c.ClientID) != "" ||
+		strings.TrimSpace(c.ClientSecret) != "" ||
+		strings.TrimSpace(c.TokenURL) != ""
+}
+
+func (c GrpcOAuthConfig) tokenSource(ctx context.Context) (oauth2.TokenSource, error) {
+	if !c.Enabled() {
+		return nil, nil
```
dobrac: lets use noop instead of nil

packages/db/migrations/20260423170000_cluster_auth_org_id.sql
```diff
@@ -0,0 +1,17 @@
+-- +goose Up
+-- +goose StatementBegin
+ALTER TABLE clusters
+    ADD COLUMN auth_org_id TEXT;
```
dobrac, line 4: do we need it nullable?

packages/db/pkg/tests/snapshots/upsert_snapshot_test.go
dobrac, line 1: this change looks unrelated

packages/db/pkg/testutils/queries.go
dobrac, line 1: unrelated

packages/db/queries/snapshots/create_new_snapshot.sql
```diff
@@ -44,6 +44,7 @@ snapshot as (
         metadata = excluded.metadata,
         sandbox_started_at = excluded.sandbox_started_at,
         origin_node_id = excluded.origin_node_id,
+        allow_internet_access = excluded.allow_internet_access,
```
dobrac: unrelated

dobrac (changes requested):

iac/modules/job-api/jobs/api.hcl
matthewlouisbrockman: moved into traefik [Route public API gRPC through ingress](https://github.com/e2b-dev/infra/pull/2470/commits/104a8bbc20f0eed0a0d05de7474dfb340112d383)

iac/modules/job-client-proxy/jobs/client-proxy.hcl
matthewlouisbrockman: it's used in byoc, but infer it without needing to pass it better i think now

iac/provider-gcp/nomad-cluster/network/ingress.tf
```diff
@@ -55,6 +55,27 @@ resource "google_compute_backend_service" "ingress" {
   }
 }
 
+resource "google_compute_backend_service" "ingress_grpc" {
+  name = "${var.prefix}ingress-grpc"
+
+  protocol  = "H2C"
+  port_name = var.ingress_port.name
```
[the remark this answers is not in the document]
matthewlouisbrockman: i mean, it's public ini that it comes from traefik, intentional

iac/modules/job-client-proxy/jobs/client-proxy.hcl
matthewlouisbrockman: https://github.com/e2b-dev/infra/pull/2470/commits/60df4dfa9970b8c464e6fe98db1fc64d01ffe485

packages/api/internal/cfg/model.go
matthewlouisbrockman: was from before switched to oauth, missed that one

packages/api/internal/clusters/discovery/discovery.go
matthewlouisbrockman: it's the ip on the byoc side for routing after autoresume

packages/api/internal/handlers/client_proxy_oauth.go
matthewlouisbrockman: zitadel's for issuing the tokens, don't need to do that, we're using workos for it

matthewlouisbrockman: (https://github.com/coreos/go-oidc has more stars + for this sorta token checking)

packages/api/internal/orchestrator/autoresume.go
matthewlouisbrockman: it's valid still, putting it back in

packages/api/internal/orchestrator/autoresume.go
matthewlouisbrockman: k

packages/api/internal/orchestrator/autoresume.go
matthewlouisbrockman: k

packages/api/internal/orchestrator/autoresume.go
matthewlouisbrockman: it's just for cicd for the localhost being empty - edge has the ip address now when it didn't before

packages/client-proxy/internal/proxy/paused_sandbox_resumer_grpc.go
matthewlouisbrockman: k

packages/db/migrations/20260423170000_cluster_auth_org_id.sql
matthewlouisbrockman: at least not until existing entries are set I think

packages/db/pkg/tests/snapshots/upsert_snapshot_test.go
matthewlouisbrockman: moved to chore/snapshot-allow-internet-upsert

packages/db/pkg/testutils/queries.go
matthewlouisbrockman: moved to chore/snapshot-allow-internet-upsert

packages/db/queries/snapshots/create_new_snapshot.sql
matthewlouisbrockman: moved to chore/snapshot-allow-internet-upsert

packages/api/internal/handlers/client_proxy_oauth.go
matthewlouisbrockman: k https://github.com/e2b-dev/infra/pull/2470/commits/761e74e6fff81f89534809204b099a11f423ac1b

packages/api/internal/cfg/model.go
```diff
@@ -53,7 +53,11 @@ type Config struct {
 	RedisTLSCABase64 string `env:"REDIS_TLS_CA_BASE64"`
 	RedisPoolSize    int    `env:"REDIS_POOL_SIZE"     envDefault:"160"`
 
-	APIGrpcPort uint16 `env:"API_GRPC_PORT" envDefault:"5009"`
+	APIGrpcPort       uint16 `env:"API_GRPC_PORT"        envDefault:"5009"`
+	APIPublicGrpcPort uint16 `env:"API_PUBLIC_GRPC_PORT" envDefault:"5109"`
```
sitole: Naming port public on a service that is all exposed public doesn't add sense to me. Let's remove public from the name.

packages/api/internal/clusters/discovery/remote.go
```diff
@@ -49,6 +65,7 @@ func (sd *RemoteServiceDiscovery) Query(ctx context.Context) ([]Item, error) {
 			UniqueIdentifier: n.ServiceInstanceID,
 			NodeID:           n.NodeID,
 			InstanceID:       n.ServiceInstanceID,
+			IPAddress:        ipAddressFromServiceHost(n.ServiceHost),
```
sitole: Lets match the structure we have for local discovery and call it `LocalIPAddress` and in addition add `LocalInstanceApiPort`

sitole: Local in name is okay, as it suggests it's local in the cluster, so it should be named local even for a remote cluster.

packages/api/internal/clusters/discovery/discovery.go
sitole: Lets re-use already existing `LocalIPAddress` and `LocalInstanceApiPort` for this. 
Followup to comment https://github.com/e2b-dev/infra/pull/2470/changes#r3146018209

packages/api/internal/clusters/instance.go
```diff
@@ -34,6 +34,7 @@ type Instance struct {
 
 	ClusterID uuid.UUID
 	NodeID    string
+	IPAddress string
```
sitole: Call it `LocalIPAddress` to stay uniform

iac/provider-gcp/nomad-cluster/network/main.tf
```diff
@@ -250,6 +250,11 @@ resource "google_compute_url_map" "orch_map" {
     path_matcher = "api-paths"
   }
 
+  host_rule {
```
sitole: This should not be needed as an ingress load balancer is used, right? 
Intsted you need to add it into list of subdomains we are generating DNS harnessing to ingress load balancer - https://github.com/e2b-dev/infra/blob/9c3b7c5dbd181ba819084276b8228f70936e911e/iac/provider-gcp/nomad-cluster/network/ingress.tf#L2

iac/provider-gcp/nomad-cluster/network/ingress.tf
```diff
@@ -55,6 +55,27 @@ resource "google_compute_backend_service" "ingress" {
   }
 }
 
+resource "google_compute_backend_service" "ingress_grpc" {
```
sitole: This backend should be used in the Ingress load balancer, right? I think H2C should be able to handle both HTTP 1.1 and HTTP 2 without issues, so the ingress load balancer only needs the backend. There is probably some migration needed to roll it safely.

iac/modules/job-api/jobs/api.hcl
```diff
@@ -23,6 +23,8 @@ job "api" {
         static = "${api_grpc_port}"
       }
 
+      port "grpc_public" {}
+
```
sitole: Why are there two gRPC endpoints exposed? Cannot we just merge them into one called `api_grpc`?

sitole: To confirm, we have one internal and one public-facing port, correct? Can we use a single port and include an auth mechanism, even for in-cluster use? I believe these in-cluster calls should be authorized too.

packages/shared/pkg/grpc/proxy/metadata.go
```diff
@@ -7,6 +7,10 @@ const (
 	MetadataSandboxRequestPort = "e2b-sandbox-request-port"
 	// MetadataEnvdAccessToken is forwarded by client-proxy for envd traffic on secure sandboxes.
 	MetadataEnvdAccessToken = "e2b-envd-access-token"
+	// MetadataClientProxyAuthToken authenticates client-proxy resume requests to the API.
+	MetadataClientProxyAuthToken = "x-e2b-client-proxy-auth"
```
sitole: Can we remove `x-` prefix? Using this prefix for HTTP headers is considered deprecated.

packages/api/internal/handlers/client_proxy_oauth.go
```diff
@@ -0,0 +1,69 @@
+package handlers
```
sitole: If there's no blocker, can we move this and the existing handler to `internal/handlers-grpc` or similar to avoid mixing it with the existing rest api handlers?

packages/api/internal/handlers/client_proxy_oauth.go
```diff
@@ -0,0 +1,69 @@
...
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/coreos/go-oidc/v3/oidc"
+)
+
+type ClientProxyOAuthVerifier interface {
```
sitole: This logic shouldn't be part of the handlers package. Consider moving it to pkg under api/internal or shared library. Some unit tests would be nice to have.

packages/api/internal/handlers/client_proxy_oauth.go
```diff
@@ -0,0 +1,69 @@
...
+	return ClientProxyOAuthClaims{}, errors.New("client proxy OIDC verifier is not configured")
+}
+
+func (v *oidcClientProxyOAuthVerifier) VerifyClaims(ctx context.Context, rawToken string) (ClientProxyOAuthClaims, error) {
+	idToken, err := v.verifier.Verify(ctx, rawToken)
+	if err != nil {
+		return ClientProxyOAuthClaims{}, fmt.Errorf("verify client proxy OIDC token: %w", err)
+	}
+
+	claims := struct {
```
sitole: Lets use global structure instead of anoynmous

packages/db/migrations/20260423170000_cluster_auth_org_id.sql
sitole, line 4: Mark it unique

packages/orchestrator/pkg/proxy/proxy.go
```diff
@@ -79,11 +79,11 @@ func NewSandboxProxy(meterProvider metric.MeterProvider, port uint16, sandboxes
 			// Handle traffic access token validation.
 			// We are skipping envd port as it has its own access validation mechanism.
 			if accessToken != "" && isNonEnvdTraffic {
-				accessTokenRaw := r.Header.Get(trafficAccessTokenHeader)
+				accessTokenRaw := r.Header.Get(proxygrpc.MetadataTrafficAccessToken)
```
sitole: It's not correct to have it in the grpc package when it's only used there and doesn't belong to the grpc impl in any way.

packages/client-proxy/internal/proxy/proxy.go
```diff
@@ -107,6 +120,11 @@ func handlePausedSandbox(
 		return "", autoResumeErrored, err
 	}
 
+	nodeIP = strings.TrimSpace(nodeIP)
```
sitole: Unify logic here and line 53, both of which check the same condition with the unknown orchestrator IP.

packages/client-proxy/internal/proxy/paused_sandbox_resumer_grpc.go
```diff
@@ -17,28 +22,92 @@ import (
...
+
+var noopOAuthTokenSource = oauth2.StaticTokenSource(&oauth2.Token{})
+
+func (c GrpcOAuthConfig) Enabled() bool {
+	return strings.TrimSpace(c.ClientID) != "" ||
+		strings.TrimSpace(c.ClientSecret) != "" ||
+		strings.TrimSpace(c.TokenURL) != ""
+}
+
+func (c GrpcOAuthConfig) tokenSource(ctx context.Context) (oauth2.TokenSource, error) {
```
sitole: Let's use two different implementations that will be put in place during the GRPC OAuth client creation. This way you are doing a lot of unnecessary requests, always checking empty strings and so on. 

Making an implementation decision during startup will also make it easier for testing.

packages/client-proxy/internal/proxy/paused_sandbox_resumer_grpc.go
```diff
@@ -17,28 +22,92 @@ import (
 )
 
 type grpcPausedSandboxResumer struct {
```
sitole, line 21: Let's clean up this file and separate the resume logic from the OAuth client so they aren't mixed together.

packages/client-proxy/internal/proxy/paused_sandbox_resumer_grpc.go
```diff
@@ -61,6 +130,14 @@ func (c *grpcPausedSandboxResumer) Resume(ctx context.Context, sandboxId string,
 		ctx = metadata.AppendToOutgoingContext(ctx, proxygrpc.MetadataEnvdAccessToken, envdAccessToken)
 	}
 
+	token, tokenErr := c.oauthTokenSource.Token()
```
sitole: Is there a reason why envd access token and traffic access token are sent as metadata instead of adding them as a field directly into the typed gRPC message?

Looks a little more complicated to me on both sides, as without the gRPC type, you need to not forget.

iac/modules/job-api/jobs/api.hcl
sitole: For the local cluster, maybe use the existing `ADMIN_TOKEN` and in the client-proxy unify implementation with just "static token" handling?

packages/api/internal/handlers/proxy_grpc.go
```diff
@@ -123,6 +202,27 @@ func (s *SandboxService) ResumeSandbox(ctx context.Context, req *proxygrpc.Sandb
...
+		if team.ClusterID != nil {
+			cluster, found := s.api.clusters.GetClusterById(*team.ClusterID)
+			if !found {
+				return nil, status.Errorf(codes.Internal, "cluster with ID '%s' not found", *team.ClusterID)
+			}
+
+			authOrgID = strings.TrimSpace(cluster.AuthOrgID)
+		}
+
+		if err := validateClientProxyAuth(ctx, incomingMetadata, s.clientProxyOAuth, authOrgID); err != nil {
```
sitole: This branch means for each request, we need to do a SQL query? This is a publicly exposed endpoint. To prevent DDoSing our database, we should first check the OAuth server for a valid auth token, and then access the database.

sitole: Please double-check if the database query is really needed. Each API should have loaded all used BYOC clusters in memory that are periodically synced with the database. From this cached list, you should be able to access the auth org ID even without a database query.

sitole (changes requested):
Feel free to sync on next steps before doing any implementation changes.

packages/api/internal/clusters/discovery/discovery.go
matthewlouisbrockman: k that works

iac/modules/job-api/jobs/api.hcl
matthewlouisbrockman: internal does not have the second auth wrapper around it, it wasn't publicly exposed

packages/api/internal/cfg/model.go
matthewlouisbrockman: changing it to edge for now work? not sure about combining them right now

packages/api/internal/clusters/discovery/remote.go
matthewlouisbrockman: k

packages/api/internal/clusters/instance.go
matthewlouisbrockman: k

iac/provider-gcp/nomad-cluster/network/main.tf
matthewlouisbrockman: moved to the ingress

iac/provider-gcp/nomad-cluster/network/ingress.tf
matthewlouisbrockman: that's definitely the question - lemme see if anyone knows why we aren't *h2c to begin with

matthewlouisbrockman: actually, maybe doing a separate PR to upgrade it first makes sense

packages/db/migrations/20260423170000_cluster_auth_org_id.sql
matthewlouisbrockman: can orgs only have one cluster each?

iac/provider-gcp/nomad-cluster/network/ingress.tf
matthewlouisbrockman: Experimented with it in https://github.com/e2b-dev/infra/pull/2506. Got too many 502s if directly changed to H2C but no issues if created new backend and then moved the rules over to the new backend. Creating new backend first in that PR (keeping it separate from this), then if want to combine them can just move the existing rules to the new backend safely and delete the old one.

We're likely going to need to do this again though for to use HTTP/2 instead of H2C at some point but leaving that outside this PR.

matthewlouisbrockman: (will merge that into this one once approved)

packages/shared/pkg/grpc/proxy/metadata.go
matthewlouisbrockman: can remove it from the test altogether, switched over to the JWT and that was from when i was using the cluster token

packages/api/internal/handlers/client_proxy_oauth.go
matthewlouisbrockman: hmmm, can move the client proxy oauth out but the existing grpc handler uses the colocated api store a lot... is handlers supposed to be *rest only?

packages/orchestrator/pkg/proxy/proxy.go
matthewlouisbrockman: fair

packages/client-proxy/internal/proxy/proxy.go
matthewlouisbrockman: k

packages/client-proxy/internal/proxy/paused_sandbox_resumer_grpc.go
matthewlouisbrockman: k

packages/client-proxy/internal/proxy/paused_sandbox_resumer_grpc.go
matthewlouisbrockman: k

packages/client-proxy/internal/proxy/paused_sandbox_resumer_grpc.go
matthewlouisbrockman: This is used on existing autoresume so not really want to change this right now; maybe a separate todo:/issue for cleanup?

packages/api/internal/handlers/proxy_grpc.go
matthewlouisbrockman: checking oauth first, i think still need the db check for team

packages/api/internal/handlers/client_proxy_oauth.go
matthewlouisbrockman: https://github.com/e2b-dev/infra/pull/2470/commits/b40b17932d270683d7ca73214de7c891d943e0a7 moves the client proxy oauth out tho

packages/api/internal/handlers/client_proxy_oauth.go
matthewlouisbrockman: https://github.com/e2b-dev/infra/pull/2470/commits/b40b17932d270683d7ca73214de7c891d943e0a7

packages/api/internal/handlers/client_proxy_oauth.go
matthewlouisbrockman: https://github.com/e2b-dev/infra/pull/2470/commits/b40b17932d270683d7ca73214de7c891d943e0a7

packages/orchestrator/pkg/proxy/proxy.go
matthewlouisbrockman: https://github.com/e2b-dev/infra/pull/2470/commits/41017d65bdd4671d4cd9db7dc96d801553f0c77c

packages/client-proxy/internal/proxy/proxy.go
matthewlouisbrockman: https://github.com/e2b-dev/infra/pull/2470/commits/da734bd6872ec828e948793d4e96bfbb354f9a5b

packages/client-proxy/internal/proxy/paused_sandbox_resumer_grpc.go
matthewlouisbrockman: https://github.com/e2b-dev/infra/pull/2470/commits/af9208f90c5215caccbeac020b75fb30e4501ebe

packages/client-proxy/internal/proxy/paused_sandbox_resumer_grpc.go
matthewlouisbrockman: https://github.com/e2b-dev/infra/pull/2470/commits/e320d4a6d80cf214fd254a8846488ca614585310

packages/api/internal/handlers/proxy_grpc.go
matthewlouisbrockman: https://github.com/e2b-dev/infra/pull/2470/commits/bb52622cf5f8ee14657549061edf7a367d61dc26 does the oauth check but db i think still needed

packages/api/internal/clusters/discovery/remote.go
matthewlouisbrockman: https://github.com/e2b-dev/infra/pull/2470/commits/1f76ea49d5e7a596433c5135abcd594523f6eb28

iac/modules/job-api/jobs/api.hcl
matthewlouisbrockman: can add a token to the internal i guess but maybe with followup PR because would require separate client proxy deployment?

iac/provider-gcp/nomad-cluster/network/ingress.tf
matthewlouisbrockman: k merged in main https://github.com/e2b-dev/infra/pull/2470/commits/161fb72b38a948a3dc180f81f8eeca2c5e81daf4 there's a h2c backend service now

packages/api/internal/handlers/proxy_grpc.go
matthewlouisbrockman: added a cache for it to the auth store https://github.com/e2b-dev/infra/pull/2470/commits/b0256e9fb3b0bf33d1e5292b1d2fed888aca682b

packages/db/migrations/20260423170000_cluster_auth_org_id.sql
matthewlouisbrockman: done https://github.com/e2b-dev/infra/pull/2470/commits/7f2b4464b2b816cf0ac9df6fead2599c86a1b655

packages/client-proxy/internal/proxy/paused_sandbox_resumer_grpc.go
matthewlouisbrockman: https://linear.app/e2b/issue/ENG-3522/handle-byoc-autoresume for followup on the metadata

packages/client-proxy/internal/proxy/paused_sandbox_resumer_grpc.go
```diff
@@ -19,17 +22,52 @@ import (
...
+func apiGrpcAddressUsesTLS(address string) bool {
+	address = strings.TrimSpace(address)
+
+	host, _, err := net.SplitHostPort(address)
+	if err != nil {
+		host = address
+	}
+	host = strings.Trim(strings.TrimSpace(host), "[]")
+
+	if host == "" || host == "localhost" || strings.HasSuffix(host, ".service.consul") {
```
dobrac: if we're having something consul specific, we should have it interfaced/abstracted so it can be easily replaced

packages/client-proxy/internal/proxy/paused_sandbox_resumer_grpc.go
```diff
@@ -19,17 +22,52 @@ import (
...
+
+	return true
+}
+
+func NewGRPCPausedSandboxResumer(address string, oauthConfig GRPCOAuthConfig) (PausedSandboxResumer, error) {
 	if strings.TrimSpace(address) == "" {
 		return nil, errors.New("api grpc address is required")
 	}
 
+	auth, err := newGrpcResumeAuth(context.Background(), oauthConfig)
```
dobrac: lets pass the context

packages/client-proxy/internal/proxy/paused_sandbox_resumer_grpc.go
```diff
@@ -61,6 +100,12 @@ func (c *grpcPausedSandboxResumer) Resume(ctx context.Context, sandboxId string,
 		ctx = metadata.AppendToOutgoingContext(ctx, proxygrpc.MetadataEnvdAccessToken, envdAccessToken)
 	}
 
+	var err error
+	ctx, err = c.auth.authorize(ctx)
```
dobrac:
```suggestion
	ctx, err := c.auth.authorize(ctx)
```

dobrac (changes requested):
Why using the `http://github.com/coreos/go-oidc` library and not the `https://github.com/zitadel/oidc`?

packages/client-proxy/internal/proxy/paused_sandbox_resumer_grpc.go
matthewlouisbrockman: k, inferring the TLS from the external vs internal grpc variable being present

matthewlouisbrockman: that way, no need to check the actual value https://github.com/e2b-dev/infra/pull/2470/commits/9d720b1bc9b5a77323c9328cb6f0973e8810acd7

packages/client-proxy/internal/proxy/paused_sandbox_resumer_grpc.go
matthewlouisbrockman: k, passing in context https://github.com/e2b-dev/infra/pull/2470/commits/2f99e54c8912f2e9563ad2969be7524bd79b6bc9

packages/client-proxy/internal/proxy/paused_sandbox_resumer_grpc.go
matthewlouisbrockman: fair https://github.com/e2b-dev/infra/pull/2470/commits/1c45a5d791fa9daf4bb0666470bea748138212a3

matthewlouisbrockman:
> Why using the `http://github.com/coreos/go-oidc` library and not the `https://github.com/zitadel/oidc`?

coreos/go-oidc just the minimal library for verifying the token and claims; don't really need zitadel's server

dobrac:
ok, lets keep using the coreos one 👍

sitole (approved):

dobrac (approved):

PR #2481, opened 2026-04-22
fix(compression): race for header finalization

levb:
Summary

  Fixes a Resume-after-Pause deadlock in Checkpoint and layer-pipeline builds (V3 and V4). UFFD/NBD page-fault readers on a just-paused snapshot.

  How

  Header dependency state is now a SetOnce[map] with three entry points (born-final, pending-with-seeded-knownDependencies, self-local via MarkLocallyAvailable). LookupDependency on a pending header answers from
  knownDependencies directly (zero Dep on miss — matches legacy V3 Size()-RPC fallback); no channel wait. V3 Upload finalizes at the top since V3 stores no deps; V4 finalizes per-file after compression, with defer
  CancelOnError propagating upload failures.

  Drops the cross-layer PendingBuildInfo accumulator and the two-phase UploadData + FinalizeHeaders uploader API — each header now owns its own lifecycle.


packages/shared/pkg/storage/header/header.go
```diff
@@ -16,56 +18,56 @@ import (
...
-	t.Builds[buildID] = bd
+	return &Header{
+		Metadata:     &meta,
+		Mapping:      t.Mapping,
+		dependencies: t.dependencies,
+	}
 }
 
+// NewHeader creates a minimal/empty header: no Dependencies map, born final.
+// Use for V3 (uncompressed) paths, skeleton devices, and tests where a
```
dobrac: we should make v4 the default after, not falling back to v3 for tests, etc.

packages/shared/pkg/storage/header/header.go
```diff
@@ -16,56 +18,56 @@ import (
...
-	// RPC and reads uncompressed data when nil.
-	Builds map[uuid.UUID]BuildData
-
-	Mapping []BuildMap
+	Mapping  []BuildMap
+
+	dependencies      map[uuid.UUID]Dependency
+	dependenciesReady chan struct{} // nil for born-final; closed once finalized
+	dependenciesErr   error
+	finalizeOnce      sync.Once
```
dobrac: all of this should be replaceable by SetOnce

packages/shared/pkg/storage/header/serialization_v4.go
```diff
@@ -173,13 +203,11 @@ func deserializeV4(metadata *Metadata, blockData []byte) (*Header, error) {
...
-	h, err := NewHeader(metadata, mappings)
-	if err != nil {
-		return nil, err
+	if pending {
+		return NewHeaderWithPendingDependencies(metadata, mappings, dependencies)
 	}
-	h.Builds = builds
 
-	return h, nil
+	return NewHeaderWithKnownDependencies(metadata, mappings, dependencies)
```
dobrac: why do we need this distinction? Aren't provided dependencies the same as pending that are already resolved?

packages/shared/pkg/storage/header/header.go
levb: Done, but we have 2 maps now (initialDependencies from the constructor) and the dependencies added after uploads (SetOnce).

packages/shared/pkg/storage/header/serialization_v4.go
levb: WithKnownDependencies creates a "final" header; no updates are coming. WithPendingDependencies creates a header with the initial (inherited) map but there are more pending.

In the case of pause - in the case of an overlay device that was dirty - we use Pending since the "self" dependency needs to be set after the upload; but if it's not dirty - we start it off as "already known".

In the P2P case - first we get the header with its inherited dependencies (we need them to resume on the client); later once finalized we add the finalized dependencies. This is why I added the `pending` flag to the serialized header.

Ideally, we should not wait for dependencies here, and defer the dependency resolution all he way until this header is serialized, but I am not quite sure how to code it up yet.

packages/shared/pkg/storage/header/header.go
levb: Once it's stable - it's a superior format for uncompressed, too (avoids Size calls)

levb: https://linear.app/e2b/issue/ENG-3848/switch-to-v4-headers-for-uncompressed-uploads

packages/shared/pkg/storage/header/header.go
```diff
@@ -166,14 +225,21 @@ func (t *Header) GetShiftedMapping(ctx context.Context, offset int64) (BuildMap,
...
-// nil means the build is uncompressed — the caller reads raw bytes instead.
-func (t *Header) GetBuildFrameData(buildID uuid.UUID) *storage.FrameTable {
-	if t.Builds == nil {
-		return nil
+func (t *Header) LookupDependency(ctx context.Context, buildID uuid.UUID) (Dependency, error) {
+	if t.locallyAvailable.Load() && buildID == t.Metadata.BuildId {
+		return Dependency{}, nil
+	}
+
+	if t.IsPending() {
```
levb: TODO: These 3 lines make it all work (some tests deadlock without them), but I confess that I do not fully understand this fallback. Need to get to the bottom of it.

packages/shared/pkg/storage/header/serialization_v4.go
```diff
@@ -1,3 +1,19 @@
+// V4 header binary format:
+//
+//	[ Metadata                      ] // fixed-size, binary.LittleEndian
+//	[ uint32 uncompressedBlockSize  ] // little-endian, size of the inner block
+//	[ LZ4(block)                    ] // block layout below
+//
+// Inner block (LZ4-compressed), all little-endian:
+//
+//	[ byte    pendingFlag           ] // 1 if Dependencies not yet finalized
```
levb: This is to communicate the pending state over the wire in P2P. 0/5 it may be better off as an RPC param and not in the serialized header. Explore.

packages/shared/pkg/storage/header/header.go
levb: Ok, this is because switching to SetOnce resulted in the 2-list approach; this is effectively a 3-way switch on the state. Changed to a switch statement to make it a little more intention revealing.

levb:
Closed in favor of https://github.com/e2b-dev/infra/pull/2532

PR #2482, opened 2026-04-22
refactor(telemetry): add minimal direct OTLP/HTTP export path

ben-fornefeld:
## Summary
- add a minimal direct OTLP/HTTP telemetry export path alongside the existing collector gRPC path
- keep the shared telemetry setup small by relying on standard OTEL exporter environment variables instead of custom config parsing
- preserve existing metric selector helpers and add focused tests for the shared telemetry and tracing middleware behavior

## Motivation
Some services cannot use the internal collector path because of how they are deployed or networked. We still want those services to export telemetry through our shared setup, so this refactor adds a direct path for Grafana's managed OTLP/HTTP endpoint without changing the existing collector-based flow for internal services.


packages/shared/pkg/telemetry/traces.go
```diff
@@ -23,11 +24,15 @@ func (nsb *noopSpanExporter) Shutdown(context.Context) error { return nil }
 func NewSpanExporter(ctx context.Context, extraOption ...otlptracegrpc.Option) (sdktrace.SpanExporter, error) {
 	opts := []otlptracegrpc.Option{
 		otlptracegrpc.WithInsecure(),
-		otlptracegrpc.WithEndpoint(otelCollectorGRPCEndpoint),
+		otlptracegrpc.WithEndpoint(OTELCollectorGRPCEndpoint()),
```
dobrac, line 27: lets pass the endpoint as constructor parameter

packages/shared/pkg/telemetry/main.go
```diff
@@ -68,15 +70,28 @@ func New(ctx context.Context, nodeID, serviceName, serviceCommit, serviceVersion
 	otel.SetMeterProvider(meterProvider)
 
 	// Setup logging
-	logProvider, err := NewLogProvider(ctx, res)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create log provider: %w", err)
-	}
+	var logProvider LogProvider
+	var spanExporter sdktrace.SpanExporter
+	if legacyGRPCEndpoint != "" {
```
dobrac: if this is legacy, what is the migration path? Can we switch to the new path already?

dobrac (changes requested):

packages/shared/pkg/telemetry/main.go
ben-fornefeld: legacy is the wrong symbol naming. i addressed this already before but this is an artifact.

packages/shared/pkg/telemetry/main.go
```diff
@@ -31,16 +30,19 @@ type Client struct {
...
 func New(ctx context.Context, nodeID, serviceName, serviceCommit, serviceVersion, serviceInstanceID string, additional ...attribute.KeyValue) (*Client, error) {
-	if otelCollectorGRPCEndpoint == "" {
+	mode := currentExportMode()
+	if mode == exportModeDisabled {
 		return NewNoopClient(), nil
 	}
 
 	// Setup metrics
-	metricsExporter, err := NewMeterExporter(ctx, otlpmetricgrpc.WithAggregationSelector(func(kind sdkmetric.InstrumentKind) sdkmetric.Aggregation {
+	metricsExporter, err := NewMeterExporter(ctx, WithMeterAggregationSelector(func(kind sdkmetric.InstrumentKind) sdkmetric.Aggregation {
```
dobrac, line 45: the meter exporter has in it's implementation the same http/grpc switch as this New constructor. Can we unite the abstraction so the switch is only one?

packages/shared/pkg/telemetry/logs.go
```diff
@@ -45,3 +50,17 @@ func NewLogProvider(ctx context.Context, res *resource.Resource, extraOpts ...ot
...
+
+func NewHTTPLogProvider(ctx context.Context, res *resource.Resource) (LogProvider, error) {
+	exporter, err := otlploghttp.New(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create logs exporter: %w", err)
+	}
+
+	p := sdklog.NewLoggerProvider(
+		sdklog.WithResource(res),
+		sdklog.WithProcessor(sdklog.NewBatchProcessor(exporter)),
```
dobrac, line 62: these options should be the same as in the grpc one, we should define it in one place to not drift

ben-fornefeld:
closing this - no longer required as we went a different route on the other service which would have needed this

PR #2484, opened 2026-04-23
fix(orch): resolve firecracker version via feature flag in template builder

jakubno:
Resolve the version inside Builder.Build on template manager as the FC binary presents is more tight with template manager version than API version


packages/api/internal/handlers/template_request_build_v3.go
```diff
@@ -115,6 +115,9 @@ func requestTemplateBuild(ctx context.Context, c *gin.Context, a *APIStore, body
 	}
 	span.End()
 
+	// TODO: Stop sending the firecracker version from the API. The orchestrator
```
sitole: @jakubno can you please add issue for cleanup?

packages/orchestrator/pkg/template/build/builder.go
```diff
@@ -123,6 +125,9 @@ func (b *Builder) Build(ctx context.Context, paths storage.Paths, cfg config.Tem
 		featureflags.TeamContext(cfg.TeamID),
 	)
 
+	cfg.KernelVersion = featureflags.DefaultKernelVersion
+	cfg.FirecrackerVersion = b.featureFlags.StringFlag(ctx, featureflags.BuildFirecrackerVersion)
```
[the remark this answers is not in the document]
jakubno: That's the purpose of the PR to move it to template manager

packages/orchestrator/pkg/template/build/builder.go
```diff
@@ -128,6 +129,23 @@ func (b *Builder) Build(ctx context.Context, paths storage.Paths, cfg config.Tem
...
 	cfg.FirecrackerVersion = b.featureFlags.StringFlag(ctx, featureflags.BuildFirecrackerVersion)
 
+	// The API computes HugePages from whatever firecracker version its own
+	// feature-flag lookup returned, under a different LaunchDarkly context
+	// (team/user/tier/cluster vs. builder's template/team). A cross-1.7
+	// divergence between the two resolutions would configure the memfile
+	// page size (2 MiB vs 4 KiB) for a binary we are not about to launch.
+	// Recompute from the firecracker version we actually resolved above so
+	// the page size always matches the binary that runs.
+	if info, err := fcversion.New(cfg.FirecrackerVersion); err == nil {
```
sitole: Let's deprecated gRPC field for huge pages too.

sitole (changes requested):
Is the timeline for removing the firecracker and kernel fields from the API after the next weekly stable release? Afaik, it shouldn't hurt to do it immediately, but maybe someone is using it differently, and we should wait.

jakubno:
> Is the timeline for removing the firecracker and kernel fields from the API after the next weekly stable release? Afaik, it shouldn't hurt to do it immediately, but maybe someone is using it differently, and we should wait.

I would definitely wait 1 release, so we can assyme template managers are rotated and can resolve the version themself

sitole (approved):

PR #2495, opened 2026-04-24
feat(resume-build): add -no-default-route flag to skip in-namespace default route

arkamar:
Allows experimenting with sandbox resume behavior when the guest has no default egress route. Controlled solely via CLI; the new network.Config.SkipDefaultRoute field is not exposed via env to avoid affecting production.


arkamar:
obsoleted by #2497

PR #2496, opened 2026-04-24
Bake clickhouse-client and CNI plugins into cluster disk image

jakubno:
Move two runtime installs out of start-clickhouse.sh into the shared Packer image build:

- clickhouse-client: installed from the official static tarball at a pinned version (default 25.4.5.24, matches clickhouse_version in job-clickhouse). Symlinked as clickhouse-client and guaranteed on PATH via /etc/profile.d/clickhouse.sh.
- CNI plugins: installed to /opt/cni/bin at a pinned version (default v1.6.2). Required by Nomad bridge-mode networking on the ClickHouse nodepool; harmless on other nodepools.

Pre-baking both removes network flakiness at boot and keeps the client version in lockstep with the server.

### DIFF
--- a/iac/nomad-cluster-disk-image/setup/install-clickhouse-client.sh
+++ b/iac/nomad-cluster-disk-image/setup/install-clickhouse-client.sh
@@ -43,18 +43,28 @@ esac
 BASE_URL="https://packages.clickhouse.com/tgz/stable"
 TARBALL="clickhouse-common-static-${VERSION}-${CH_ARCH}.tgz"
 URL="${BASE_URL}/${TARBALL}"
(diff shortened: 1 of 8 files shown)


sitole:
Please replicate same changes to AWS setup

jakubno:
> Please replicate same changes to AWS setup

it's there already, isn't it?

sitole:
> > Please replicate same changes to AWS setup
> 
> it's there already, isn't it?

Yes, you are right. I reviewed mobile and missed that. Sorry for bothering you.

sitole (approved):

PR #2502, opened 2026-04-26
fix: use command substitution for GOMAXPROCS instead of literal string

caiyesd:
`GOMAXPROCS='nproc'` sets the literal string "nproc" rather than the CPU count. Use `$(nproc)` to correctly capture the output of the command.


jakubno:
Hey @caiyesd , could you please sign CLA first? I also think this can be completely removed

ValentaTomas:

PR #2503, opened 2026-04-27
Update edge-api openapi spec

sitole:
Updated edge-api spec with removed deprecated service discovery endpoint.

### DIFF
--- a/spec/openapi-edge.yml
+++ b/spec/openapi-edge.yml
@@ -78,7 +78,6 @@ components:
         - healthy
         - draining
         - unhealthy
-        - standby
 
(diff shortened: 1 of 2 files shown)


spec/openapi-edge.yml
```diff
@@ -78,7 +78,6 @@ components:
         - healthy
         - draining
         - unhealthy
-        - standby
```
sitole, line 81: @jakubno did some work here and this is not correct change. Let's wait on fixing before this pr is merged.

sitole: cc @jakubno

PR #2506, opened 2026-04-27
add H2C ingress backend for BYOC

matthewlouisbrockman:
adds a H2C ingress backend in addition to existing HTTP one to use for future infra for BYOC; it's not used currently but that's fine

### DIFF
--- a/iac/provider-gcp/nomad-cluster/network/ingress.tf
+++ b/iac/provider-gcp/nomad-cluster/network/ingress.tf
@@ -37,7 +37,7 @@ resource "google_compute_health_check" "ingress" {
 resource "google_compute_backend_service" "ingress" {
   name = "${var.prefix}ingress"
 
-  protocol  = "HTTP"
+  protocol  = "H2C"
   port_name = var.ingress_port.name
 
   session_affinity = null
--- a/iac/provider-gcp/nomad-cluster/network/ingress.tf
+++ b/iac/provider-gcp/nomad-cluster/network/ingress.tf
@@ -37,6 +37,27 @@ resource "google_compute_health_check" "ingress" {
 resource "google_compute_backend_service" "ingress" {
   name = "${var.prefix}ingress"
 
+  protocol  = "HTTP"
+  port_name = var.ingress_port.name
+
+  session_affinity = null
+  health_checks    = [google_compute_health_check.ingress.id]
+
+  timeout_sec = var.ingress_timeout_seconds
+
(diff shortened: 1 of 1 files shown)


iac/provider-gcp/nomad-cluster/network/ingress.tf
```diff
@@ -37,7 +37,7 @@ resource "google_compute_health_check" "ingress" {
 resource "google_compute_backend_service" "ingress" {
   name = "${var.prefix}ingress"
 
-  protocol  = "HTTP"
+  protocol  = "H2C"
```
ValentaTomas, line 61: Really not sure if this won't break things.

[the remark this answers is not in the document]
matthewlouisbrockman: switching them to http2_health_check with h2c fails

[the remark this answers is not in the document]
matthewlouisbrockman: alright, this seems to be more of a pain than it's worth, leaving as http/1.1 should be fine

matthewlouisbrockman: adding a second instead with H2C leaving this as HTTP. tested and it did lead to some failed responses with just changing this.

iac/provider-gcp/nomad-cluster/network/ingress.tf
```diff
@@ -55,6 +55,27 @@ resource "google_compute_backend_service" "ingress" {
   }
 }
 
+resource "google_compute_backend_service" "http2_ingress" {
```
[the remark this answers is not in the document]
djeebus: how do we plan on using this?

iac/provider-gcp/nomad-cluster/network/ingress.tf
```diff
@@ -55,6 +55,27 @@ resource "google_compute_backend_service" "ingress" {
...
+  name = "${var.prefix}http2-ingress"
```
dobrac:
```suggestion
  name = "${var.prefix}http2c-ingress"
```

dobrac (approved):

iac/provider-gcp/nomad-cluster/network/ingress.tf
matthewlouisbrockman: kk switched names to h2c

PR #2507, opened 2026-04-28
fix(otel): drop exemplars at collector to fit Mimir items/s budget

ValentaTomas:
## Summary

Foxtrot's Grafana Cloud Mimir tenant is being rate-limited (100k items/s, 300k burst) — samples, metadata, and exemplars are all being dropped, breaking API dashboards. Querying `grafanacloud_instance_*_per_second` directly shows that **exemplars alone account for ~90% of what we ship**:

| Signal | Accepted | Discarded (rate_limited) | Discarded (other) | **Total sent** |
|---|---:|---:|---:|---:|
| samples   |  7,840/s | 16,271/s | 23/s aggregator-too-old | **24,134/s** |
| **exemplars** | 38,651/s | **103,326/s** | **156,341/s** too_old | **298,318/s** |
| metadata  |  2,708/s |  4,357/s | — | **7,065/s** |
| **total** | **49,199/s** | **123,954/s** | **156,364/s** | **🔥 329,517/s** |

24h peak exemplars hit ~420k/s alone. We are not querying exemplars in any dashboard today.

This PR adds a `transform/drop_exemplars` processor that calls `set(exemplars, [])` on every metric data point and wires it into the four Grafana-Cloud-bound metrics pipelines (`metrics`, `metrics/prometheus`, `metrics/rpc_only`, `metrics/host`). The clickhouse-bound `metrics/external` pipeline is intentionally not touched.

### Why exemplars exploded

- OTel Go SDK defaults `OTEL_METRICS_EXEMPLAR_FILTER=trace_based`, so every metric data point inside a sampled span gets a trace-id exemplar attached.
- Tracing is currently unsampled (~42k spans/s peak).
- Histograms `rpc.server.call.duration` (8,820 series) and `rpc.client.call.duration` (3,304 series) each carry ~16 buckets, so each scrape can emit thousands of exemplars.

### Projected impact (after rollout)

```
samples sent     :  24,134/s   (unchanged)
exemplars sent   :       0/s   (stripped)
metadata sent    :   7,065/s   (unchanged)
total            :  31,199/s   ->  31% of the 100k items/s limit
```

Eliminates:
- samples \`rate_limited\` (16k/s -> 0)
- samples \`aggregator-sample-too-old\` on \`orchestrator_sandbox_envd_init_calls_total\` (queue stops backing up)
- exemplars \`rate_limited\` (103k/s -> 0)
- exemplars \`exemplar_too_old\` (156k/s -> 0)
[...]

PR #2508, opened 2026-04-28
fix(otel): disable metric exemplars in Go services

ValentaTomas:
Apply `metric.WithExemplarFilter(exemplar.AlwaysOffFilter)` as the default option in `telemetry.NewMeterProvider`, so every Go service that goes through our shared telemetry helper drops exemplars on histogram/counter measurements. Callers can still override it via `extraOption` if they ever need to.

Exemplars count 1:1 against the Mimir tenant items/s limit, and we don't query them in any dashboard today.

Replaces #2507 (broken — OTTL `set(exemplars, [])` is a silent no-op per upstream maintainers).

### Changes
- `packages/shared/pkg/telemetry/metrics.go` — add `WithExemplarFilter(AlwaysOffFilter)` as a default `NewMeterProvider` option.
- `packages/api/internal/metrics/team.go`, `packages/orchestrator/pkg/metrics/sandboxes.go` — drop the now-redundant explicit filter at the call sites and the `exemplar` import.

### Why SDK option instead of `OTEL_METRICS_EXEMPLAR_FILTER` env var
- Single source of truth — behavior lives next to the rest of the meter setup; no drift between local dev, self-hosted, and prod jobspecs.
- No per-jobspec duplication — no need to add a line to every `*.hcl` / `*.tf` we ever create.
- Self-hosters get it for free without copying an env var into their own deployments.
- Works for `NewAnonymous` / CLI tools (`create-build`, `clean-nfs-cache`, smoke tests) that don't go through Nomad templates.
- Two call sites in this repo were already doing exactly this — they were just doing it ad-hoc per call site rather than at the shared helper.

### DIFF
--- a/iac/modules/job-api/jobs/api.hcl
+++ b/iac/modules/job-api/jobs/api.hcl
@@ -135,7 +135,6 @@ job "api" {
         ANALYTICS_COLLECTOR_API_TOKEN  = "${analytics_collector_api_token}"
         LOGS_COLLECTOR_ADDRESS         = "${logs_collector_address}"
         OTEL_COLLECTOR_GRPC_ENDPOINT   = "${otel_collector_grpc_endpoint}"
-        OTEL_METRICS_EXEMPLAR_FILTER   = "always_off"
 
         REDIS_POOL_SIZE                = "${redis_pool_size}"
         REDIS_CLUSTER_URL              = "${redis_cluster_url}"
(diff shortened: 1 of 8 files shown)


dobrac (approved):

jakubno:
Why we don't do it with
`metric.WithExemplarFilter(exemplar.AlwaysOffFilter)`?

ValentaTomas:
We probably can, but not sure if it matters. Do you prefer the code+rebuild way?

jakubno:
> We probably can, but not sure if it matters. Do you prefer the code+rebuild way?
yes imho it's more reliable, we reuse it in services in other repo as well

jakubno (approved):

PR #2511, opened 2026-04-28
test(uffd): add deterministic race tests for stale-source / madvise-deadlock / faulted-short-circuit

ValentaTomas:
## Summary

Adds a `net/rpc/jsonrpc`-over-unix-socket RPC harness for the userfaultfd cross-process tests, plus three deterministic race tests that demonstrate stale-source / madvise-deadlock / faulted-short-circuit bugs in the REMOVE-event handling path.

**These tests are intentionally failing on this PR — that is the bug demonstration. The production fix lands in stacked PR #2512.**

## Stack / dependency

- Base: `feat/uffd-test-scaffolding` (PR #2475 — generic gated/async cross-process scaffolding).
- This PR depends on PR #2475 merging first. With that, the chain #2475 → this PR → #2512 is reviewable independently of the full free-page-reporting feature PR (#1896).
- Original parent context: PR #1896 (`feat/free-page-reporting`).

## Commits

1. `test(uffd): lift REMOVE-event handling and pageTracker.removed closure from feat/free-page-reporting` — verbatim state lift of `packages/orchestrator/pkg/sandbox/uffd/userfaultfd/` from PR #1896 tip (commit `f31027327`). 8 files, +764/-213. Includes:
   - `pageTracker` `removed` `pageState`
   - REMOVE-event drain in `Serve()` (under `settleRequests.Lock()`)
   - Worker-fault dispatch with switch on tracker state (in its current, **buggy**, parent-loop snapshot form — that is what #2512 fixes)
   - `wakeupPipe` self-pipe + `defaultCopyMode` + `MapMemoryReadyAddr` lifecycle
   - `deferred.go` (deferred pagefault list)
   - `prefault.go` updates for prefetch / WP coordination
   - `remove_test.go` (REMOVE-event integration tests)
   - REMOVE-aware variants of `cross_process_helpers_test.go` / `helpers_test.go` / `fd_helpers_test.go`

2. `test(uffd): add net/rpc/jsonrpc test harness over unix socket` — replaces the parent↔child harness's pile of single-purpose pipes (offsets, ready, gate-cmd, gate-sync) plus `SIGUSR1`/`SIGUSR2` signaling with one bidirectional Unix socket carrying stdlib `net/rpc` + `net/rpc/jsonrpc`. Per-call request IDs are handled by the standard library, so the parent can issue concurrent RPCs while a handler is parked.

   Replaced surface:
   | Old | New |
   |---|---|
   | `SIGUSR1` graceful exit | `Service.Shutdown` |
   | `SIGUSR2` + offsets pipe + `binary.Write` | `Service.PageStates` |
   | ready pipe + ReadAll-on-close handshake | `Service.WaitReady` |
   | gate-cmd / gate-sync byte protocol | `Service.ServePause` / `Service.ServeResume` |

   New on the same channel:
   - `Service.InstallFaultBarrier(addr, point) → token` — arms a deterministic barrier in the child's worker goroutine at one of two test-only hook points (`beforeWorkerRLockHook`, `beforeFaultPageHook`).
   - `Service.WaitFaultHeld(token)` — blocks until the worker reaches the barrier; the RPC reply IS the rendezvous.
   - `Service.ReleaseFault(token)` — lets the parked worker proceed.

   The two hook fields on `Userfaultfd` default to `nil` and are nil-checked on the hot path, so production sees zero behavioral change. They are only assigned in the child's `crossProcessServe` wiring.

   `testConfig` gains a `sourcePatcher` hook so race tests can plant a deterministic sentinel byte into the random source data BEFORE the content file is written, instead of depending on the happenstance value of any randomly-generated byte.

3. `test(uffd): add deterministic stale-source / madvise-deadlock / faulted-short-circuit race tests` — three race tests built on the harness above. None use sleeps, retries, or soak loops; each installs explicit barriers on the child's worker goroutine, drives the racing kernel operation from the parent, and asserts on a concrete post-state.

   - **`TestStaleSourceRaceMissingAndRemove`** — regression for the production fix. Plants sentinel byte `0xc3` into the source page, parks the worker via `barrierBeforeRLock`, fires `MADV_DONTNEED`, waits for the REMOVE batch to commit, releases the worker, then asserts the page is zero-filled. Pre-fix, the worker `UFFDIO_COPY`s the planted sentinel because it captured `source = u.src` in the parent loop before the REMOVE landed; post-fix it re-reads state under RLock and zero-faults. Wallclock < 50ms / variant.

     Pre-fix failure on this branch (`4k` and `hugepage` variants), captured from prior Linux sudo runs:

     ```
     race_test.go:184:
         Error: Not equal:
             expected: 0x0
             actual  : 0xc3
         Test:  TestStaleSourceRaceMissingAndRemove/4k
         Messages: page 1 first byte: want 0 (post-fix zero-fault for `removed` state),
                   got 0xc3 — if this equals the sentinel 0xc3, the worker used a stale
                   `source = u.src` snapshot (regression)
     ```

   - **`TestNoMadviseDeadlockWithInflightCopy`** — liveness regression for the user-visible symptom that originally surfaced the race. Parks the worker via `barrierBeforeFaultPage` (i.e. holding RLock), fires `MADV_DONTNEED`, asserts `madvise` returns within 2s. Catches any future change that accidentally couples `readEvents()` to `settleRequests` as a fast assertion failure rather than a 30m CI timeout. Wallclock < 50ms / variant.

     Assertion shape on regression:

     ```
     DEADLOCK: madvise(MADV_DONTNEED) on page %d did not return within 2s
     while a worker is parked holding settleRequests.RLock — readEvents
     must drain the REMOVE event WITHOUT acquiring settleRequests
     ```

   - **`TestFaultedShortCircuitOrdering`** — smoke test on the REMOVE-then-pagefault batch ordering using the gated harness. Pins the invariant that REMOVE batches drain before pagefault dispatch in a single `Serve` iteration. Wallclock ~120ms / variant.

   The `4k` and `hugepage` variants of `TestStaleSourceRaceMissingAndRemove` are the bug-demonstration failures; the other two currently pass and are forward-looking belt-and-suspenders so any future regression surfaces as an instant assertion rather than a 30-minute CI hang.

## Verification

- This branch is on a Mac workstation; UFFD (`userfaultfd(2)` + `MADV_DONTNEED` semantics) is Linux-only and CGO cross-compile from Darwin is not set up here, so `go build` / `go vet` / sudo test runs were not executed locally on this push. Assertion text above is from prior Linux sudo runs.
- CI on Linux is the source of truth for the pre-fix failure signatures.
- With the fix in #2512 applied on top, all three tests pass `-count=20 -timeout=30s` deterministically.

## Diff size

9 files, ~+1791 / -503 lines vs `feat/uffd-test-scaffolding` (the lift accounts for the bulk; harness rewrite + race tests are the rest).

## Out of scope

No Firecracker version bump, no free-page-reporting feature flag, no template-manager / API plumbing, no proto regen, no CI timeout tweak. Strictly the UFFD test closure.


ValentaTomas:
Reopening to force GitHub Actions to re-run the full test workflow. The previous force-push (rebase onto #2475's branch) did not retrigger the `pull_request` workflow — only the trivial `pull_request_target` reviewer-assigner ran, which is why all checks falsely showed green. See thread for details.

ValentaTomas:
Closing in favor of the cleaner four-PR split:
- #2519 — refactor(uffd-tests): RPC harness over net/rpc/jsonrpc
- #2520 — feat(uffd): UFFD_EVENT_REMOVE + matrix-mode tests
- #2521 — test(uffd): deterministic race tests (intentionally fails — bug demonstration)
- #2512 (rebased) — fix(uffd): re-read state inside worker under RLock (makes #2521 pass)

The lift commit is now decomposed across #2519 (harness) and #2520 (REMOVE handling + matrix). The race tests live in #2521; the fix lives in #2512. Bug demonstration is built into the stack — #2521's CI fails with `page 1 first byte: want 0 ... got 0xc3` and #2512's CI passes with the same tests.

PR #2512, opened 2026-04-28
fix(uffd): read page state inside worker under settleRequests.RLock

ValentaTomas:
## Summary

Move the `pageTracker.get(addr)` switch and the `source = u.src` capture from the parent `Serve()` loop into the worker goroutine, after `settleRequests.RLock()`. Read + act + commit is now atomic with respect to the REMOVE batch.

This PR is deliberately split from #2521 so each PR has a single concern: #2521 lands the deterministic regression tests (the red CI is the bug demo); this PR lands the one-file production fix on top and turns those tests green.

## Changes

- `userfaultfd.go` only (+35 / -27). Conceptually three lines moved and one `continue` turned into `return nil` to fit the goroutine closure.

## Root cause

Pre-fix, a `UFFD_EVENT_REMOVE` arriving between the parent loop reading `pageTracker.get(addr)` / capturing `source = u.src` and the worker getting onto a CPU would commit `removed` under `settleRequests.Lock()` first. The worker would then `UFFDIO_COPY` the **stale** source bytes into the page the kernel had just `MADV_DONTNEED`'d and overwrite `removed` with `faulted`. User-visible symptoms: orchestrator parent `madvise()` blocking forever and pages reappearing with stale src bytes after a remove.

## Stack

- Base: `test/uffd-stale-source-race-tests` (#2521)
- Earlier: #2519 -> #2520 -> #2521 -> **this PR**
- Final target: `main`

## Test plan

- [x] `go build ./...`
- [x] `go vet ./pkg/sandbox/uffd/...`
- [x] `golangci-lint run ./pkg/sandbox/uffd/userfaultfd/... ./pkg/sandbox/uffd/testutils/...`
- [x] `sudo GOMAXPROCS=2 go test -race -timeout 15m -count=1 ./pkg/sandbox/uffd/userfaultfd/...` -- all race tests added by #2521 now pass, including `TestStaleSourceRaceMissingAndRemove/{4k,hugepage}`.
- [x] Soak: `sudo go test -count=20 -timeout=30s -run 'TestStaleSourceRaceMissingAndRemove|TestNoMadviseDeadlockWithInflightCopy|TestFaultedShortCircuitOrdering' ./pkg/sandbox/uffd/userfaultfd/...` -- deterministic pass.

### DIFF
--- a/.github/workflows/pr-tests-arm64.yml
+++ b/.github/workflows/pr-tests-arm64.yml
@@ -1,6 +1,6 @@
 name: ARM64 tests on PRs
 
-on: [workflow_call]
+on: [ workflow_call ]
 
 permissions:
   contents: read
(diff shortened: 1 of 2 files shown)


ValentaTomas:
Folded into #2520 — race-test demonstration and fix now live alongside the REMOVE matrix tests so the stack is just #2519 → #2520.

PR #2513, opened 2026-04-28
feat(uffd): add REMOVE-event handling, pageTracker `removed` state, and deterministic race tests

ValentaTomas:
This PR is the **test side** of a stacked pair targeting `main`:

- **PR A (this one)** — lifts the minimum production-side UFFD subset from
  `feat/free-page-reporting` (#1896) needed to compile and exercise the
  REMOVE/MISSING/COPY interleavings, plus the new `net/rpc/jsonrpc` test
  harness over a unix domain socket and three deterministic race tests.
  The race tests are intentionally configured against the **buggy**
  pre-fix worker dispatch (state read in the parent loop instead of under
  `settleRequests.RLock`), so `TestStaleSourceRaceMissingAndRemove` fails
  on this PR. That failure is the bug demonstration.
- **PR B (stacked)** — `fix/uffd-stale-source-race-on-main` — applies the
  one-file production fix that flips PR A's failing test green.

Why a separate stack instead of just landing in #1896: the user wanted
the stale-source race to be reviewable as a tests-then-fix pair against
`main`, independent of the rest of the free-page-reporting feature work
(Firecracker v1.14, FPR feature flag, template-manager plumbing, proto
regen, etc.). When this stack merges, #1896 will rebase on top and drop
the lifted commits.

---

### 1. Production code lifted from #1896 (commit `c50322b1d`)

Files: `userfaultfd.go`, `deferred.go`, `page_tracker.go`, `prefault.go`,
plus pre-existing scaffolding in `cross_process_helpers_test.go`,
`fd_helpers_test.go`, `helpers_test.go`, and the new `remove_test.go`
(scaffolding form — replaced by the harness in commit 2).

The lift adds:

- `pageTracker.pageState` gains `removed` (distinct from `missing` /
  `faulted`).
- `Serve()` drains the UFFD fd in a per-poll-cycle batch and processes
  `UFFD_EVENT_REMOVE`: REMOVE batches take `settleRequests.Lock()` and
  call `pageTracker.setState(..., removed)` for every page in the
  removed range, **before** any pagefault dispatch in that iteration.
- For each pagefault in the batch, `Serve()` reads `pageTracker` state
  in the **parent loop** (NOT under `settleRequests.RLock`) and captures
  `source = u.src` there before dispatching the worker goroutine.
  **This is the buggy form that PR B fixes.** A `REMOVE` event arriving
  between the parent-loop state read and the worker actually acquiring
  RLock leaves the worker with a stale `source = u.src` snapshot, which
  it then `UFFDIO_COPY`s into a page the kernel just `MADV_DONTNEED`'d.
- `deferred.go` batches up pagefaults that returned EAGAIN/short-circuit
  so the next poll cycle picks them up; a self-pipe `wakeupPipe` wakes
  poll immediately when a deferred fault is enqueued.
- `prefault.go` grows the same batched-event handling.

Out of scope (intentionally NOT lifted from #1896): Firecracker v1.14
bump, free-page-reporting feature flag, template-manager / API plumbing,
proto regen, fcversion sandbox feature, anything outside
`packages/orchestrator/pkg/sandbox/uffd/userfaultfd/`.

Lift size: 8 files, +883 / -193 lines (production + base scaffolding).

### 2. `net/rpc/jsonrpc` test harness over unix socket (commit `fe8d2df75`)

Replaces the previous pile of single-purpose pipes (offsets, ready,
gate-cmd, gate-sync) plus SIGUSR1/SIGUSR2 signals with one bidirectional
unix domain socket carrying stdlib `net/rpc` + `net/rpc/jsonrpc`. Only
the kernel userfaultfd is still handed off out-of-band (via
`ExtraFiles`); source data goes through a temp file.

RPC methods exposed by the child's `Service`:

- `Service.WaitReady` — replaces the ready pipe + ReadAll handshake.
- `Service.PageStates` — replaces SIGUSR2 + offsets pipe + binary.Write protocol; returns `{missing, faulted, removed}` offset slices.
- `Service.ServePause` — replaces gate-cmd byte protocol; pauses the serve loop at the next iteration.
- `Service.ServeResume` — replaces gate-sync byte protocol; resumes a paused serve loop.
- `Service.InstallFaultBarrier` — NEW: arms a deterministic barrier in the child's worker goroutine at one of two hook points (`barrierBeforeRLock` or `barrierBeforeFaultPage`); returns a token.
- `Service.WaitFaultHeld` — NEW: blocks until the worker reaches the installed barrier — the RPC reply IS the rendezvous.
- `Service.ReleaseFault` — NEW: lets the parked worker proceed.
- `Service.Shutdown` — replaces SIGUSR1 graceful exit.

Two new test-only fields land on `Userfaultfd`:
`beforeWorkerRLockHook` and `beforeFaultPageHook`, both
`func(addr uintptr)`. Both default to `nil` and are nil-checked on the
hot path, so production sees zero behavioural change. They are only
assigned in the child's `crossProcessServe` wiring.

`testConfig` gains a `sourcePatcher` hook so race tests can plant a
deterministic sentinel byte into the random source data BEFORE the
content file is written, instead of depending on the happenstance value
of any randomly-generated byte.

The same commit removes `t.Parallel()` from the gated cross-process
tests (`TestRemoveThenWriteGated`, `TestWriteThenRemoveGated`,
`TestFaultedShortCircuitOrdering`). While the handler is in the gated
`paused` state, the user thread that triggered the queued WRITE fault
is suspended in the kernel's pagefault path. From the Go runtime's
perspective that goroutine is "running" (not in syscall) and cannot be
preempted. If a CONCURRENT cross-process test triggers a stop-the-world
GC pause, STW will wait forever for the suspended goroutine to reach a
safe point. Running the gated tests sequentially avoids that
interleaving while leaving every other test on `t.Parallel()`.

### 3. Deterministic race tests (commit `151736b81`)

Three tests, all built on the harness + barrier hooks. None of them use
sleeps, retries, or soak loops; each one installs an explicit barrier,
drives the racing kernel operation from the parent, and asserts on a
concrete post-state.

- **`TestStaleSourceRaceMissingAndRemove`** — regression test for the
  fix landing in PR B. Plants a non-zero sentinel (`0xc3`) into page 1,
  parks the worker via `barrierBeforeRLock`, fires `MADV_DONTNEED` on
  the same page, waits for the REMOVE batch to commit
  (`state == removed`), releases the worker, then asserts the page is
  zero-filled.
  - **Pre-fix (this PR): FAILS** with:
    ```
    page 1 first byte: want 0 (post-fix zero-fault for `removed` state),
    got 0xc3 — if this equals the sentinel 0xc3, the worker used a
    stale `source = u.src` snapshot (regression)
    ```
    `0xc3` is exactly the planted sentinel, proving the worker
    `UFFDIO_COPY`'d stale `u.src` bytes after the page was removed.
  - Post-fix (PR B): passes in ~50 ms / variant.
- **`TestNoMadviseDeadlockWithInflightCopy`** — liveness regression test
  for the user-visible symptom that originally surfaced the bug. Parks
  the worker via `barrierBeforeFaultPage` (i.e. holding RLock,
  mid-handler), fires `MADV_DONTNEED` from the parent, asserts madvise
  returns within 2 s. Pre-fix this currently passes (the lifted
  `readEvents()` does not take `settleRequests`, so madvise is not
  blocked). Future-proofs against any change that accidentally couples
  `readEvents()` to `settleRequests` — would surface as a fast 2 s
  assertion failure rather than a 30 m CI timeout.
- **`TestFaultedShortCircuitOrdering`** — smoke test on the
  REMOVE-then-pagefault batch ordering using the gated harness. Pins
  the invariant that REMOVE batches drain before pagefault dispatch in
  a single Serve iteration. Passes both pre- and post-fix — guards
  against future regressions in batch ordering.

Local sudo run on this branch (Linux, Go 1.25.9):

```
=== RUN   TestStaleSourceRaceMissingAndRemove
--- FAIL: TestStaleSourceRaceMissingAndRemove (0.05s)
    --- FAIL: TestStaleSourceRaceMissingAndRemove/4k       (assertion above, got 0xc3)
    --- FAIL: TestStaleSourceRaceMissingAndRemove/hugepage (assertion above, got 0xc3)
--- PASS: TestNoMadviseDeadlockWithInflightCopy (0.04s)
--- PASS: TestFaultedShortCircuitOrdering       (0.24s)
FAIL
```

> These race tests intentionally fail on this PR. The fix lands in stacked
> PR `fix/uffd-stale-source-race-on-main`. **Merge this stack as a unit**;
> do not merge this PR alone or CI on `main` will go red.

---

## Stacking

- This PR's base: `main`.
- Stacked child PR's base: this branch (`feat/uffd-page-tracker-and-remove-events`).

## Related

- Parent feature work: #1896 (`feat/free-page-reporting`). When this
  stack merges, #1896 will rebase on top and drop the lifted commits.
- The original Cursor Bugbot review on PR #1896 flagged the stale-source
  race against commit `c8b3d63` — that report is what this stack
  addresses.

## Test plan

- [x] `go build ./...` clean against `main` + this PR.
- [x] `go vet ./...` clean.
- [x] `sudo go test -timeout 2m -count=1 -run 'Race|Deadlock|Ordering' ./pkg/sandbox/uffd/userfaultfd/...` — `TestStaleSourceRaceMissingAndRemove` fails with the documented assertion (bug demo); other two pass.
- [x] `sudo go test -timeout 5m -count=1 ./pkg/sandbox/uffd/userfaultfd/...` — all pre-existing tests (`TestRemove*`, `TestRemoveThenFault`, `TestRemoveThenWriteGated`, `TestWriteThenRemoveGated`, `TestMissing*`, `TestMissingWrite*`, `TestAsyncWriteProtection`, `TestSerial*`, `TestParallel*`) pass.


ValentaTomas:
Closing in favor of #2511. We're keeping the Stack A form (#2511 → #2475 → main, #2512 stacked on #2511). #2513/#2514 were a duplicate Option-3 stack opened in parallel; superseded.

PR #2529, opened 2026-04-30
feat(orchestrator): allow overriding with specific ip allow for internal

WWZZ:
Currently DeniedSandboxCIDRs has blocks of denies causing some of our internal endpoints we want to expose to sandbox to be blocked.

I'm adding support here to allow users to configure in .env.* files to allow opening specific cidrs / IPs for internal usage on sandbox if necessary.


dobrac (approved):

dobrac:
Hi @WWZZ, looks great, thank you for the contribution!

PR #2531, opened 2026-04-30
feat(sandbox): Allow support for adding to sandbox hosts entries

WWZZ:
Allow sandboxes the ability to override with specific hosts file entries.

Today sandboxes reaches out to 8.8.8.8, but for our use cases we don't really want to expose those dns out externally everywhere. This way we can add specific dns entries without breaking the current resolver behavior for some of our internal dns needs within sandbox.


WWZZ:
This specific use case does not seem to be necessary for open source

PR #2532, opened 2026-04-30
fix(orch): upload/V4 header race, including P2P

levb:
Fixes a race in the V4 upload + header path, including across orchestrators (P2P).

  - **Per-build upload coordination via `SetOnce` futures.** Replaces `upload_tracker` in `layer_executor` and the ad-hoc plumbing in `build_upload*`. A single `Uploads` registry tracks in-flight uploads per build;
   waiters block on a future that resolves once the upload (and its V4 header) is durable.
  - **Collapsed upload plumbing.** `build_upload.go` / `build_upload_v3.go` / `build_upload_v4.go` are replaced by `upload.go` + `upload_v3.go` + `upload_v4.go` + `uploads.go` under `pkg/sandbox`, with a clearer
  split between the per-version upload step and the registry that coordinates waiters.
  - **Drop V4 header carriage from P2P.** Peers no longer ship header bytes over gRPC; once a peer signals the build is finalized, consumers fetch the canonical header from GCS. Removes a class of staleness bugs
  where the peer-side header could disagree with what later landed in storage.
  - **Redis pub/sub hint for cross-orch upload-finished.** Publishes a small "build finalized" hint so a remote orchestrator waiting on a sibling build wakes immediately instead of polling.
  - Adds a rapid pause/resume integration test that exercises the cross-orch wait path.


packages/shared/pkg/storage/header/metadata.go
```diff
@@ -156,6 +156,11 @@ func (d *DiffMetadata) ToDiffHeader(
 		return nil, fmt.Errorf("failed to create header: %w", err)
 	}
 
+	// Diff headers describe a build whose data is only in the local cache
+	// until the upload pipeline finalizes and SwapHeader publishes the
+	// post-upload header.
+	header.IncompletePendingUpload = true
```
dobrac: it should be hardwired in the constructor

packages/orchestrator/pkg/sandbox/build/header_load.go
```diff
@@ -0,0 +1,67 @@
...
+	hdrPath := storage.Paths{BuildID: buildID.String()}.HeaderFile(string(t))
+	deadline := time.Now().Add(budget)
+
+	backoff := loadV4InitialBackoff
+	for {
+		h, err := header.LoadHeader(ctx, store, hdrPath)
+		if err == nil {
+			return h, nil
+		}
+		if !errors.Is(err, storage.ErrObjectNotExist) {
```
dobrac, line 50: we should accept some amount of failures

packages/orchestrator/pkg/sandbox/uploads.go
```diff
@@ -0,0 +1,250 @@
...
+
+	return !hasSelf
+}
+
+// pollRemoteStorageForHeader polls storage for the post-upload V4 header. Returns nil if
+// storage only has a V3 header (genuine V3 build, no upgrade available);
+// returns the V4 header on success; returns the budget-expired error if the
+// parent is still uploading on a remote orch and never lands.
+func (u *Uploads) pollRemoteStorageForHeader(ctx context.Context, buildID uuid.UUID, t build.DiffType, hint <-chan error) (*header.Header, error) {
+	if u.persistence == nil {
```
dobrac: why do we need this? Maybe we can replace with noop?

packages/orchestrator/pkg/sandbox/uploads.go
```diff
@@ -0,0 +1,250 @@
...
+}
+
+func (u *Uploads) Stop() {
+	u.futures.Stop()
+}
+
+// Start replaces a finished future at the same key; rejects an in-flight one.
+func (u *Uploads) Start(buildID uuid.UUID) (*utils.ErrorOnce, error) {
+	u.startMu.Lock()
+	defer u.startMu.Unlock()
```
dobrac: I think the lock is not necessary. Each build upload should be unique

packages/orchestrator/pkg/sandbox/uploads.go
```diff
@@ -0,0 +1,250 @@
...
+
+	return fut, nil
+}
+
+// Wait returns the parent's post-upload V4 header. Same-orch waits on the local
+// future; cross-orch refreshes from GCS when the locally-cached header is
+// stale, optionally accelerated by a per-call Redis subscription.
+func (u *Uploads) Wait(ctx context.Context, buildID uuid.UUID, t build.DiffType) (*header.Header, error) {
+	ctx, span := tracer.Start(ctx, "wait-for-parent-upload", trace.WithAttributes(
+		attribute.String("build_id", buildID.String()),
```
dobrac: withbuildid

packages/orchestrator/pkg/sandbox/uploads.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+
+	d, err := u.find(ctx, buildID, t)
+	if err != nil {
+		return nil, err
+	}
+
+	h := d.Header()
+	if isStale(h, buildID) {
+		subCtx, cancel := context.WithCancel(ctx)
```
dobrac: lets call subCtx -> ctx

packages/orchestrator/pkg/sandbox/uploads.go
```diff
@@ -0,0 +1,250 @@
...
+		defer cancel()
+
+		fresh, err := u.pollRemoteStorageForHeader(ctx, buildID, t, u.subscribe(subCtx, buildID))
```
dobrac: why fresh is nil if err is nil

packages/orchestrator/pkg/sandbox/uploads.go
```diff
@@ -0,0 +1,250 @@
...
+// future; cross-orch refreshes from GCS when the locally-cached header is
+// stale, optionally accelerated by a per-call Redis subscription.
+func (u *Uploads) Wait(ctx context.Context, buildID uuid.UUID, t build.DiffType) (*header.Header, error) {
+	ctx, span := tracer.Start(ctx, "wait-for-parent-upload", trace.WithAttributes(
+		attribute.String("build_id", buildID.String()),
+		attribute.String("file_type", string(t)),
+	))
+	defer span.End()
+
+	if item := u.futures.Get(buildID); item != nil {
```
dobrac, line 105: can you add comment here that this is for the local path?

packages/orchestrator/pkg/sandbox/uploads.go
```diff
@@ -0,0 +1,250 @@
...
+		return nil, err
+	}
+
+	h := d.Header()
+	if isStale(h, buildID) {
+		subCtx, cancel := context.WithCancel(ctx)
+		defer cancel()
+
+		fresh, err := u.pollRemoteStorageForHeader(ctx, buildID, t, u.subscribe(subCtx, buildID))
+		if err != nil {
```
dobrac: add comment that the polling is only for P2P


We might want to split the two approaches to two methods

packages/orchestrator/pkg/sandbox/uploads.go
```diff
@@ -0,0 +1,250 @@
...
+		}
+	}
+
+	return h, nil
+}
+
+// isStale reports whether h needs to be refreshed from GCS to be usable as a
+// parent in V4 lineage construction. V3 builds (Version<V4) are skipped:
+// they have no Builds map and nothing to upgrade to.
+func isStale(h *header.Header, buildID uuid.UUID) bool {
```
dobrac: please rename to isPending

packages/orchestrator/pkg/sandbox/uploads.go
```diff
@@ -0,0 +1,250 @@
...
+	if h.IncompletePendingUpload {
+		return true
+	}
+	if h.Metadata.Version < header.MetadataVersionV4 {
```
dobrac: this should not be required as V3 should have the IncompletePendingUpload set to false

packages/orchestrator/pkg/sandbox/uploads.go
```diff
@@ -0,0 +1,250 @@
...
+		return false
+	}
+	_, hasSelf := h.Builds[buildID]
```
dobrac: if this is false, the header is fundamentally broken

packages/orchestrator/pkg/sandbox/uploads.go
```diff
@@ -0,0 +1,250 @@
...
+		return nil, nil
+	}
+
+	h, err := build.PollRemoteStorageForHeader(ctx, u.persistence, buildID, t, hint, refreshHeaderBudget)
+	if err != nil {
+		return nil, err
+	}
+	if h.Metadata.Version < header.MetadataVersionV4 {
+		return nil, nil
+	}
```
dobrac: this should not be necessary

packages/orchestrator/pkg/sandbox/uploads.go
```diff
@@ -0,0 +1,250 @@
...
+
+	payload := ""
+	if uploadErr != nil {
+		payload = uploadErr.Error()
+	}
+
+	if err := u.redis.Publish(ctx, uploadDoneChannel(buildID), payload).Err(); err != nil {
+		logger.L().Warn(ctx, "failed to publish upload-done signal",
+			zap.String("build_id", buildID.String()),
```
dobrac: withbuildid

packages/orchestrator/pkg/sandbox/upload_v4.go
```diff
@@ -0,0 +1,109 @@
+package sandbox
```
dobrac, line 1: lets rename back to build_upload

packages/shared/pkg/storage/paths.go
```diff
@@ -71,6 +71,12 @@ func (p Paths) DataFile(name string, ct CompressionType) string {
 	return fmt.Sprintf("%s/%s%s", p.BuildID, name, ct.Suffix())
 }
 
+// HeaderFile returns the storage path for a header sidecar of a data file
+// (e.g. "memfile" → "{buildID}/memfile.header").
+func (p Paths) HeaderFile(name string) string {
+	return fmt.Sprintf("%s/%s%s", p.BuildID, name, HeaderSuffix)
```
dobrac, line 77: lets now reuse this helper for Rootfsheader and MemfileHeader

packages/orchestrator/pkg/sandbox/snapshot.go
```diff
@@ -18,29 +19,13 @@ type Snapshot struct {
 	RootfsDiffHeader  *header.Header
 	Snapfile          template.File
 	Metafile          template.File
+	Paths             storage.Paths
+	BuildID           uuid.UUID
+	ParentBuildID     uuid.UUID
```
dobrac: probably not used anywhere

packages/orchestrator/pkg/sandbox/sandbox.go
```diff
@@ -1117,6 +1117,17 @@ func (s *Sandbox) Pause(
...
+	parentBuildID, err := uuid.Parse(s.Template.Files().BuildID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse parent build id: %w", err)
+	}
+	// Base layers report the source template with the same buildID as self —
+	// flatten that to uuid.Nil so the upload pipeline doesn't self-deadlock
+	// waiting on its own future.
+	if parentBuildID == buildID {
+		parentBuildID = uuid.Nil
+	}
```
dobrac: double check if this can be removed as it might not be needed in Snapshot

packages/orchestrator/pkg/sandbox/upload.go
```diff
@@ -0,0 +1,97 @@
...
+
+	"github.com/e2b-dev/infra/packages/orchestrator/pkg/sandbox/build"
+	"github.com/e2b-dev/infra/packages/shared/pkg/featureflags"
+	"github.com/e2b-dev/infra/packages/shared/pkg/storage"
+	headers "github.com/e2b-dev/infra/packages/shared/pkg/storage/header"
+	"github.com/e2b-dev/infra/packages/shared/pkg/utils"
+)
+
+type Upload struct {
+	id, parent uuid.UUID
```
dobrac: lets rename id to buildID

packages/orchestrator/pkg/sandbox/block/local.go
```diff
@@ -108,7 +112,18 @@ func (d *Local) UpdateHeaderSize() error {
 		return fmt.Errorf("failed to get file info: %w", err)
 	}
 
-	d.header.Metadata.Size = uint64(info.Size())
+	// Clone-then-store so concurrent readers (via Header()) don't observe the
```
dobrac: double check the reason and code how its implemented

dobrac (changes requested):
When we get to fetch in a chunker, we should error out if the header is incomplete - defensive check to catch bugs early

packages/shared/pkg/storage/header/metadata.go
levb: `newDiffHeader` now

packages/orchestrator/pkg/sandbox/build/header_load.go
levb: Accept 3 now

packages/orchestrator/pkg/sandbox/uploads.go
levb: never nil, removed

packages/orchestrator/pkg/sandbox/block/local.go
levb: This is used by baseBuilder, I am not sure if there is concurrency there in accessing the header but this code does not hurt, we should swap-to-a-copy as a general rule.

packages/orchestrator/pkg/sandbox/sandbox.go
levb: This check is a fix for an observed bug during testing, from what I understand in some (base layer?) cases the can be true and was messing things up later.

packages/orchestrator/pkg/sandbox/snapshot.go
levb: still used basically as a param bag from Pause return to AddSandbox; OOS for this PR?

packages/orchestrator/pkg/sandbox/uploads.go
levb: 1 method makes sense in the caller context - just wait on the final header. I added the comment and simplified.

dobrac (approved):

PR #2537, opened 2026-05-01
feat(iac) Support for optional granular gsa / subnets / network tags

WWZZ:
For our security requirements, we need separate subnets / network tags / GSA for each of our resources. This PR supports that use case for GCP


WWZZ:
removing gsa portion for now, keeping it to just subnets

PR #2538, opened 2026-05-01
support for granular subnet and tags

WWZZ:
We need to support granular subnets and tags, this pr allows those be configurable per mig so that each of them can be overridden


ValentaTomas:

PR #2542, opened 2026-05-03
refactor(uffd-tests): pre-fault uffd-registered pages via madvise to eliminate STW-deadlock class

ValentaTomas:
Stacked on #2520. Eliminates the Go-runtime UFFD STW-deadlock class without splitting the harness further or removing parallelism.

## Problem

Test goroutines in the parent read UFFD-registered memory via plain Go (`bytes.Equal(memoryArea[...], ...)`). When the page is missing, the kernel parks the OS thread in TASK_KILLABLE and Linux refuses to deliver SIGURG. Go's STW (e.g. triggered by a concurrent goroutine allocating during a barrier hold) waits for that thread to reach a safe point forever.

## Fix

Drive the actual fault through a syscall — `unix.Madvise(page, MADV_POPULATE_READ/WRITE)` — so the goroutine is in `_Gsyscall` while the kernel is parked. The Go runtime's `entersyscall` transitions the goroutine to `_Gsyscall` before the syscall, detaching it from a P; STW does not wait on `_Gsyscall` goroutines, so GC completes immediately even if the kernel parks the thread inside the pagefault for arbitrarily long. Once `madvise` returns, the page is resident and the subsequent plain Go `bytes.Equal` / `copy` does not fault.

### Why MADV_POPULATE_* and not pread/pwrite on /proc/self/mem

`/proc/self/mem` goes through `access_remote_vm` -> `get_user_pages_remote` with `FOLL_FORCE | FOLL_REMOTE`, which intentionally bypasses UFFD (so a debugger cannot deadlock on a UFFD-controlled page). Empirically, `pread` on a UFFD-MISSING page returns EIO. `MADV_POPULATE_READ`/`MADV_POPULATE_WRITE` (Linux 5.14+) use `get_user_pages` (no `FOLL_REMOTE`), so UFFD fires normally.

## Knock-on cleanups

- Restore `t.Parallel()` on `TestStaleSourceRaceMissingAndRemove` and `TestNoMadviseDeadlockWithInflightCopy` (reverts the workaround in #2520's `caf0377dc`).
- Drop the `//nolint:paralleltest,tparallel` rationale on `TestFaultedShortCircuitOrdering`, `TestRemoveThenWriteGated`, and `TestWriteThenRemoveGated`. Pure cleanup — the hazard is gone.

## Notes

- Production code untouched.
- Cross-process harness unchanged: parent still owns the mmap and UFFD registration, child still serves events.
- `unix.Madvise(MADV_DONTNEED)` already runs in syscall mode, so REMOVE-trigger paths needed no changes.
- The race-test direct page-content assertion (`page[0]` after a barrier-released read) also pre-faults via `MADV_POPULATE_READ` so an ad-hoc load on a `removed` page cannot re-introduce the deadlock.

### DIFF
--- a/packages/orchestrator/pkg/sandbox/uffd/userfaultfd/helpers_test.go
+++ b/packages/orchestrator/pkg/sandbox/uffd/userfaultfd/helpers_test.go
@@ -311,7 +311,12 @@ func (h *testHandler) executeRead(ctx context.Context, op operation) error {
 
 	readBytes := (*h.memoryArea)[op.offset : op.offset+int64(h.pagesize)]
 
-	// bytes.Equal is the first access to uffd-managed memory, triggering the pagefault.
+	// MADV_POPULATE_READ faults the page in via syscall, so the goroutine sits
+	// in _Gsyscall while the kernel waits for UFFDIO_COPY — STW can preempt.
+	if err := unix.Madvise(readBytes, unix.MADV_POPULATE_READ); err != nil {
+		return fmt.Errorf("madvise POPULATE_READ at offset %d: %w", op.offset, err)
+	}
+
 	if !bytes.Equal(readBytes, expectedBytes) {
 		idx, want, got := testutils.FirstDifferentByte(readBytes, expectedBytes)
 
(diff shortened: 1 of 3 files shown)


ValentaTomas:
Folded into #2520. The two commits (madvise pre-fault + restore t.Parallel) now live there.

PR #2543, opened 2026-05-03
fix(uffd): mark received uffd fd FD_CLOEXEC

ValentaTomas:
\`recvmsg\` for the uffd fd doesn't pass \`MSG_CMSG_CLOEXEC\`, and \`ParseUnixRights\`
doesn't set the flag itself. Any later fork/exec in the orchestrator would
inherit the fd. Mark it CLOEXEC right after we parse it. Defensive — no
observed leak today, but cheap insurance.


ValentaTomas:
Will reopen later.

PR #2544, opened 2026-05-03
test(uffd): rapid state-machine + chaos-source coverage

ValentaTomas:
Stacks on `feat/uffd-remove-events-matrix`. Adds two seedable test suites that meaningfully expand coverage. Will rebase onto `refactor/uffd-test-child-owned-memory` once it lands (that branch is the proper base because both tests use `t.Parallel()` pressure, which is only safe with the syscall-based STW fix it carries).

## Architecture note

Both new tests use a **direct parent-side handler** (`directHandler`) rather than the cross-process RPC harness. This means:
- No child process spawned — the mmap, UFFD fd, and `Serve` goroutine all live in the test process.
- Chaos injection works naturally (the chaos source is in the same process as `Serve`).
- `pageStateEntries()` is called directly on `*Userfaultfd` without RPC round-trips.
- t.Cleanup handles teardown; the close-based `serveDone chan struct{}` lets both the test body and t.Cleanup wait concurrently without consuming the channel.

## TestRapidStateMachine

Property-based state-machine fuzzer using `pgregory.net/rapid v1.3.0`. Drives random sequences of `read / write / madvise(MADV_DONTNEED)` actions against a live handler; the model tracks per-page state (`missing` / `faulted` / `zero-faulted` / `removed`). The 4-state model correctly handles remove+re-fault pages that are zero-filled (not from source). Rapid shrinks any failing sequence to a minimal counterexample.

Both pagesize arms (4 KiB, 2 MiB) × both removeEnabled modes run as parallel subtests via `runMatrix`.

Reproduce a failure:
```
RAPID_SEED=<seed-from-fail-output> go test -run='TestRapidStateMachine' ./pkg/sandbox/uffd/userfaultfd/...
```

## TestChaosCloseTerminatesUnderLatency

Wraps `block.Slicer` with a `chaosSource` that injects uniform random `[0, 50ms]` latency per `Slice` call. Fires 64 concurrent `MADV_POPULATE_READ` goroutines, then triggers `Shutdown` and asserts teardown completes within 5s. Catches `Close↔Serve` drain ordering regressions where a slow worker would otherwise wedge `wg.Wait()`.

After `Serve` returns, the UFFD fd is closed so any goroutines still blocked on unresolved faults receive EFAULT and unblock gracefully.

Both pagesize arms × both removeEnabled modes run as parallel subtests via `runMatrix`.

Reproduce a failure:
```
UFFD_CHAOS_SEED=<seed-from-fail-output> go test -run='TestChaosCloseTerminatesUnderLatency' ./pkg/sandbox/uffd/userfaultfd/...
```

## Notes

- `pgregory.net/rapid` is added as a direct test-only dep at v1.3.0; verified it does not appear in non-test builds (`go list -deps ./... | grep pgregory` returns nothing).
- Both tests call `t.Parallel()` on all subtests. The direct in-process handler avoids the Go-runtime↔UFFD STW deadlock that plagued the cross-process tests (all page faults now happen via `MADV_POPULATE_READ/WRITE` in `_Gsyscall`, which is always at a GC safe point).
- Both tests require root (`UFFD_API` syscall). Without root, they skip gracefully.
- Local test run: `sudo go test -race -count=3 -timeout=180s -run='TestRapidStateMachine|TestChaosCloseTerminatesUnderLatency' ./pkg/sandbox/uffd/userfaultfd/...` — PASS (3.9s).


ValentaTomas:
Folded into #2520. Rapid + chaos tests now use the existing cross-process testharness instead of an in-process handler.

PR #2547, opened 2026-05-03
refactor(header): replace IsEmptyBlock with fast IsZero short-circuit

ValentaTomas:
`IsEmptyBlock` was the per-block hot path of `DiffMetadataBuilder.Process` — one call per block during the rootfs scan in `DirectProvider.exportToDiff`. It paid a full SIMD `memcmp` (4 KiB or, on hugepages, 2 MiB) before it could short-circuit, and required two static zero buffers (`EmptyBlock`, `EmptyHugePage`) to compare against.

Replace it with a small `IsZero` helper that uses QEMU's `buffer_is_zero` trick: sample three bytes (first, last, middle) so most non-zero blocks reject from a single cache line. The fallback uses `bytes.Equal` on `b` shifted by one (`b[1:] == b[:n-1]`), which dispatches to the runtime's SIMD `memequal` on amd64/arm64 — same speed as the old comparison against a static buffer when the short-circuit doesn't fire.

`DiffMetadataBuilder.Process` calls `IsZero` directly (the size validation `IsEmptyBlock` did was unreachable in practice). `EmptyBlock` is removed; `EmptyHugePage` stays — `build.go` uses it as the literal zero buffer for `uuid.Nil` reads in `Build.ReadAt`.

Pre-factored from #2546 so it can land on its own.

### DIFF
--- a/packages/shared/pkg/storage/header/diff.go
+++ b/packages/shared/pkg/storage/header/diff.go
@@ -18,8 +18,9 @@ var EmptyHugePage = make([]byte, HugepageSize)
 
 // IsZero reports whether b is all-zero. Samples first/middle/last byte to
 // reject most non-zero buffers from one cache line, then falls back to
-// bytes.Equal(b[:n-1], b[1:]) — true iff every adjacent pair is equal, i.e.
-// all bytes equal b[0] (which the sample already proved is zero).
+// bytes.Equal(b[:n-1], b[1:]): true exactly when every adjacent pair of
+// bytes is equal, i.e. all bytes equal b[0] (which the sample already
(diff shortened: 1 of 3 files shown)


ValentaTomas:
Trimmed `IsZero` doc to one paragraph and dropped the trivial test cases — kept just the cases that exercise the 3-byte sample, the SIMD fallback path, and the byte-just-before-the-last-sample edge case. Also fixed the `EmptyHugePage` callsite reference to `(*build.File).Slice` (was `Build.ReadAt` per Claude review).

bchalios (approved):

PR #2549, opened 2026-05-03
experiment(ext4): noatime+lazytime mounts and lazy mkfs to shrink diffs

ValentaTomas:
Three small ext4 knobs aimed at the same goal as #2546 — keeping snapshot diff size small — but separated here so we can decide each on its own. All "expected positive value, no real downside" in the [research summary](#); none are slam-dunks until measured on a real workload.

## Changes

**1. Guest rootfs mount — `rootflags = discard,noatime,lazytime`**
ext4's `relatime` default still touches the inode-table block roughly once per day per accessed inode. `noatime` removes that. `lazytime` keeps mtime/ctime updates in memory and only persists them at sync, which the snapshot path already triggers — so repeated touches of the same inode collapse into one inode-table write at snapshot time instead of N.

**2. Build-time host loop mount — `loop,discard,noatime,lazytime`**
Same reasoning as (1), applied during template construction so directory walks in the build phase don't dirty inode-table blocks in the template.

**3. `mkfs.ext4 -E lazy_itable_init=1,lazy_journal_init=1,discard`**
The two `lazy_*` options skip eager zero-fill of the inode table and journal at format time, so the freshly-created `rootfs.ext4` file stays maximally sparse on the host. The kernel's `ext4lazyinit` thread fills them on first mount — and with the cache changes from #2546, those zero writes now route through `fallocate(PUNCH_HOLE)` instead of growing the diff. `discard` is included because mke2fs handles it gracefully on regular files (no-op or `PUNCH_HOLE` fallback).

## What's not in here

`inline_data`, `data=writeback`, `bigalloc`, `commit=N`, `journal_async_commit` — all "measure-first" or risky for ephemeral sandboxes. Happy to add separately if anything looks promising in benchmarks.

Stacked on #2546 because the changes share the `rootflags` line and the `mkfs` lazy-init story only pays off once the cache layer punches zero writes.


ValentaTomas:

PR #2552, opened 2026-05-04
feat(fc): drain virtio-balloon free-page-hinting before pause

ValentaTomas:
Drains virtio-balloon **free-page-hinting** before pause so snapshots don't capture pages the guest already considers free.

Balloon install is gated by `free-page-hinting-install` (bool LD flag); kernel-side eligibility is targeted via the LD context (kernel/FC version). On pause we call `start_balloon_hinting(acknowledge_on_stop=true)` and poll `describe_balloon_hinting` until `host_cmd == DONE`, gated by `free-page-hinting-timeout-ms` (int LD flag, ms; 0 = disabled). Reclaimed pages emit `UFFD_EVENT_REMOVE`, already tracked by the parent FPR work.

Hot path is kept minimal: post-drain and post-pause we trigger an FC metrics flush but don't wait for the reader, trading per-pause counter precision for pause latency. System-level FPH activity is observable via the periodic 5 s metrics flush.

Includes `cmd/resume-build -fph-bench` and `scripts/bench-fph.sh` for offline FPR vs FPR+FPH comparison.

Operator must wait for the kernel FPH race fix to roll out before enabling `free-page-hinting-timeout-ms` in prod.


packages/orchestrator/pkg/sandbox/fc/process.go
```diff
@@ -710,6 +712,44 @@ func (p *Process) Pause(ctx context.Context) error {
...
+		}
+		// host_cmd is monotonic and we just called start, so host > 0
+		// after FC accepts it. Require it to guard against transient
+		// nil/zero responses returning a false-positive completion.
+		if host > 0 && guest >= host {
+			return nil
+		}
+		if backoff < 50*time.Millisecond {
+			backoff *= 2
+		}
```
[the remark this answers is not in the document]
ValentaTomas: Fixed in 8b35ec6 — backoff is now `backoff = min(backoff*2, 50*time.Millisecond)`.

packages/orchestrator/pkg/sandbox/fc/process.go
```diff
@@ -710,6 +712,42 @@ func (p *Process) Pause(ctx context.Context) error {
 	return p.client.pauseVM(ctx)
 }
 
+// DrainBalloon triggers a free-page-hinting run and blocks until the guest
+// acknowledges or ctx fires. No-op when the balloon wasn't installed.
+func (p *Process) DrainBalloon(ctx context.Context) error {
+	if !p.balloonInstalled {
+		return nil
```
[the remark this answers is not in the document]
ValentaTomas: Fixed in 7619cc9 — dropped the `balloonInstalled` field; `DrainBalloon` now relies on Firecracker (HTTP 400 `StartBalloonHintingBadRequest` => `not-configured`) and is correct across both create and resume paths.

ValentaTomas:
Waiting for the merge of https://github.com/e2b-dev/infra/pull/2541, but otherwise should be ready.

ValentaTomas:
Before enabling in prod we need to deploy the kernel fix though.

packages/orchestrator/pkg/sandbox/fc/process.go
```diff
@@ -713,6 +714,61 @@ func (p *Process) Pause(ctx context.Context) error {
...
+			return fmt.Errorf("balloon hinting status: %w", err)
+		}
+		// host_cmd is monotonic; require host > 0 to avoid a false-positive
+		// completion before FC has accepted the start.
+		if host > 0 && guest >= host {
+			return nil
+		}
+		backoff = min(backoff*2, 50*time.Millisecond)
+	}
+}
```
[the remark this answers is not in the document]
ValentaTomas: Good catch. Fixed in f0486c18 — the drain now snapshots host_cmd before calling startBalloonHinting and requires a strict bump (host > hostBefore && guest >= host). Strict no-op when counters start at 0, prevents the resume-from-snapshot false-positive.

packages/orchestrator/pkg/sandbox/fc/process.go
```diff
@@ -763,9 +780,11 @@ func (p *Process) DrainBalloon(ctx context.Context) error {
...
 			return fmt.Errorf("balloon hinting status: %w", err)
 		}
-		// host_cmd is monotonic; require host > 0 to avoid a false-positive
-		// completion before FC has accepted the start.
-		if host > 0 && guest >= host {
+		// Require a strict bump over the pre-start baseline to avoid
+		// a false-positive completion before FC's VMM thread has accepted
+		// the start (the API thread ack/VMM bump race), including the
+		// resume-from-snapshot case where counters restore non-zero.
+		if host > hostBefore && guest >= host {
```
kalyazin: Were you able to test that all pages are freed by this point?

It looks like `guest >= host` may be a premature condition because the guest sets the cmd to match the `host_cmd` when it starts reporting free pages, not when it finishes the cycle.

I think the right condition is below:
```diff
-if host > hostBefore && guest >= host {
+if sawBump && host == 1 { // freePageHintDone
```
where `sawBump` is set whenever we see `host > hostBefore`.

See a Firecracker test that confirms it: https://github.com/firecracker-microvm/firecracker/commit/42c70dd75ccca37044c63b0195631acd922e3111 .

kalyazin (dismissed):

packages/orchestrator/pkg/sandbox/fc/process.go
ValentaTomas: I think you are right, fixing it in the follow-up commits.

(I have not tested all pages being freed)

packages/orchestrator/pkg/sandbox/fc/process.go
```diff
@@ -713,6 +717,91 @@ func (p *Process) Pause(ctx context.Context) error {
...
+		// guest_cmd flips to host_cmd on the driver's initial cmd_id_start
+		// descriptor, before any page-blocks are processed, so it cannot
+		// signal completion. Wait for host_cmd to return to freePageHintDone
+		// after a confirmed bump — FC sets that only post-guest-STOP.
+		if host > hostBefore {
+			sawBump = true
+		}
+		if sawBump && host == freePageHintDone {
+			return nil
+		}
```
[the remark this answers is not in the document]
ValentaTomas: Valid bug — fixed in 44585a99. The `sawBump` defense was unnecessary: FC processes `StartFreePageHinting` synchronously under the VMM lock (`rpc_interface.rs::StartFreePageHinting`) and only returns after `start_hinting → update_free_page_hint_cmd(cmd_id)` has set `host_cmd >= 2`. So any post-start observation of `host_cmd == FREE_PAGE_HINT_DONE (1)` is unambiguously end-of-cycle — there's no API-ack/VMM-bump race to defend against. Dropped `sawBump` and the now-redundant pre-start describe; extracted the loop as `pollFphDone` with regression coverage in `drain_balloon_test.go`.

packages/orchestrator/pkg/sandbox/fc/process.go
```diff
@@ -713,6 +717,91 @@ func (p *Process) Pause(ctx context.Context) error {
...
+		// guest_cmd flips to host_cmd on the driver's initial cmd_id_start
+		// descriptor, before any page-blocks are processed, so it cannot
+		// signal completion. Wait for host_cmd to return to freePageHintDone
+		// after a confirmed bump — FC sets that only post-guest-STOP.
+		if host > hostBefore {
+			sawBump = true
+		}
+		if sawBump && host == freePageHintDone {
+			return nil
```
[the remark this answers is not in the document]
ValentaTomas: Valid — same root cause as the Bugbot finding. Fixed in 44585a99 by dropping the `sawBump` requirement (FC's API thread serializes the start under the VMM lock so no bump can be missed defensively). Unit test for the fast-cycle-after-prior-done case in `drain_balloon_test.go`.

packages/orchestrator/pkg/sandbox/fc/process.go
ValentaTomas: **(still reviewing this change)**

Followed up in 44585a99 — went further than your patch and dropped `sawBump` entirely. Reasoning: `rpc_interface.rs::StartFreePageHinting` is processed synchronously under the VMM lock, so by the time the start API returns, `host_cmd` is already bumped to a fresh `cmd_id >= 2`. Any subsequent `host_cmd == FREE_PAGE_HINT_DONE` is then unambiguously end-of-cycle, and a fast cycle that completes between polls (~2.4 ms hugepages cycle vs. our 5 ms initial poll) no longer hangs the loop. Confirm if the synchronous-bump assumption is correct.

For verifying all pages freed: not measured on our side yet. Plan is a Go integration test patterned after `test_orchestrator_drain_condition_is_premature` — allocate+free in guest, call `DrainBalloon`, assert `balloon.free_page_hint_count` and `free_page_hint_freed` deltas from the FC metrics FIFO. Needs a metrics-snapshot helper on `Process`, so deferring to a follow-up. Open to a simpler signal if you have one (e.g. memfile size delta with drain on/off on the same workload).

ValentaTomas:
@cla-bot check

packages/orchestrator/pkg/sandbox/fc/process.go
```diff
@@ -713,6 +717,86 @@ func (p *Process) Pause(ctx context.Context) error {
...
+
+		return nil
+	}
+
+	// acknowledge_on_stop=true so FC writes host_cmd back to freePageHintDone
+	// on the guest's STOP — that transition is what we wait on below.
+	if err := p.client.startBalloonHinting(ctx, true); err != nil {
+		var notConfigured *operations.StartBalloonHintingBadRequest
+		if errors.As(err, &notConfigured) {
+			outcome = "not-configured"
```
kalyazin, line 749: for my education, why do we emit a metric here instead of (or along with) printing a log warning?

packages/orchestrator/pkg/sandbox/fc/process.go
kalyazin: The updated flow looks correct to me. As for the verification, it looks sensible. Maybe we need both: 1) make sure the orch drains properly and also 2) that it actually leads to the memfile reduction.

packages/orchestrator/pkg/sandbox/fc/process.go
ValentaTomas: > `span.SetAttributes(attribute.String("drain-balloon.outcome", outcome))`

This is an attribute on a trace, not a metric here.
I mainly wanted to have this on a trace to check if things are enabled on specific runs.
I think for the metrics (we can have another PR) we can follow with several histograms to collect the overall data generation more granularly (we now mostly check the node egress + storage ingress). We have this info on traces, but that was for individual run analysis.

packages/orchestrator/pkg/sandbox/fc/process.go
ValentaTomas: Both signals wired in aa9bcbd8.

(1) DrainBalloon now emits `drain-balloon.{hinted_count,hinted_bytes}` as span attributes by snapshotting cumulative `balloon.free_page_hint_count` / `free_page_hint_freed` from the FC metrics FIFO before and after the drain loop. `outcome=ok` with `hinted_count=0` is observable as the no-op-drain case.

(2) `cmd/resume-build -fph-bench` reads the same counters after each pause, prints them per iteration and per arm, and warns if the FPR+FPH arm reports `hint_count=0` across all iterations — that's the regression catch for any future "drain returns without doing work" bug independent of the memfile-size signal.

Mechanics: `firecrackerMetrics` extended to parse balloon (FC's SharedIncMetric resets on serialize, so the reader accumulates per-line deltas), `Process.BalloonMetrics` and `Process.FlushAndReadBalloonMetrics` expose the cumulative snapshot. Unit tests cover the accumulator and the JSON-tag mapping.

kalyazin: Thanks, having such a test is very nice.

It took me a while to wire the feature flags to run it locally, but it probably just me. Besides that, I noticed a couple of potentially material issues. Specifically, due to imprecise OpenAPI spec in Firecracker, we misinterpret  a successful `PATCH /balloon/hinting/start` call as a failure. I want to submit a fix upstream to FC, but meanwhile we have to work it around. Also, we can improve handling of balloon stats when the FC process terminates.

I pushed all those changes (3 fixes + 1 feature flag wiring) into a branch based on this one: https://github.com/kalyazin/infra/tree/feat/sandbox-pause-fph-fixes . Please have a look and feel free to cherry-pick those you find valid.

kalyazin: Upstream PR for return code mismatch: https://github.com/firecracker-microvm/firecracker/pull/5884 .

ValentaTomas: Cherry-picked all four fixes onto the branch. Thanks!

iac/modules/job-otel-collector/configs/otel-collector.yaml
```diff
@@ -524,6 +530,16 @@ service:
...
         - clickhouse
+%{ if enable_otel_router_metrics }
+    metrics/external/otel_router:
+      receivers:
+        - otlp
+      processors:
+        - filter/external_metrics
+        - batch
+      exporters:
+        - otlp/otel_router
```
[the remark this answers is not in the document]
ValentaTomas: Unrelated to this PR — comment is on `iac/modules/job-otel-collector/configs/otel-collector.yaml` which came in via a merge with main. Bugbot Autofix already self-flagged this as a false positive.

packages/orchestrator/pkg/sandbox/fc/drain_balloon_test.go
```diff
@@ -0,0 +1,31 @@
+package fc
```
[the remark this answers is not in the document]
ValentaTomas: Fixed in 2eedd3f. Both test files now carry `//go:build linux`.

kalyazin (approved):

kalyazin:
Just wanted to note that with unconditional enabling of FPR/FPH in https://github.com/e2b-dev/infra/pull/2552/changes#diff-264fea9254b55eb530decfded9ded8f53bc7623f8c207656650746796b64f178 , the tests would be failing for FC < 1.14. We may accept it if we believe we won't need to stick with 1.12 in the future.

Or we could harden it a bit:
```diff
diff --git a/packages/orchestrator/cmd/create-build/main.go b/packages/orchestrator/cmd/create-build/main.go
index 8fe6a832b..fa600eba9 100644
--- a/packages/orchestrator/cmd/create-build/main.go
+++ b/packages/orchestrator/cmd/create-build/main.go
@@ -39,6 +39,7 @@ import (
 	"github.com/e2b-dev/infra/packages/orchestrator/pkg/template/build/metrics"
 	artifactsregistry "github.com/e2b-dev/infra/packages/shared/pkg/artifacts-registry"
 	"github.com/e2b-dev/infra/packages/shared/pkg/dockerhub"
+	"github.com/e2b-dev/infra/packages/shared/pkg/fcversion"
 	"github.com/e2b-dev/infra/packages/shared/pkg/featureflags"
 	templatemanager "github.com/e2b-dev/infra/packages/shared/pkg/grpc/template-manager"
 	"github.com/e2b-dev/infra/packages/shared/pkg/logger"
@@ -354,6 +355,13 @@ func doBuild(
 		})
 	}
 
+	// Mirror prod's gating (pkg/template/server/create_template.go:70): only
+	// enable FPR if the chosen Firecracker version actually supports it.
+	// Hardcoding true breaks builds on FC <1.14 with a balloon 400.
+	fcInfo, err := fcversion.New(fc)
+	if err != nil {
+		return fmt.Errorf("invalid firecracker version %q: %w", fc, err)
+	}
 	tmpl := config.TemplateConfig{
 		Version:            templates.TemplateV2LatestVersion,
 		TemplateID:         templateID,
@@ -366,7 +374,7 @@ func doBuild(
 		ReadyCmd:           readyCmd,
 		KernelVersion:      kernel,
 		FirecrackerVersion: fc,
-		FreePageReporting:  true,
+		FreePageReporting:  fcInfo.HasFreePageReporting(),
 		TeamID:             "local",
 		Steps:              steps,
 	}
```

ValentaTomas:
> Just wanted to note that with unconditional enabling of FPR/FPH in [#2552 (changes)](https://github.com/e2b-dev/infra/pull/2552/changes#diff-264fea9254b55eb530decfded9ded8f53bc7623f8c207656650746796b64f178) , the tests would be failing for FC < 1.14. We may accept it if we believe we won't need to stick with 1.12 in the future.
> 
> Or we could harden it a bit:
> 
> ```diff
> diff --git a/packages/orchestrator/cmd/create-build/main.go b/packages/orchestrator/cmd/create-build/main.go
> index 8fe6a832b..fa600eba9 100644
> --- a/packages/orchestrator/cmd/create-build/main.go
> +++ b/packages/orchestrator/cmd/create-build/main.go
> @@ -39,6 +39,7 @@ import (
>  	"github.com/e2b-dev/infra/packages/orchestrator/pkg/template/build/metrics"
>  	artifactsregistry "github.com/e2b-dev/infra/packages/shared/pkg/artifacts-registry"
>  	"github.com/e2b-dev/infra/packages/shared/pkg/dockerhub"
> +	"github.com/e2b-dev/infra/packages/shared/pkg/fcversion"
>  	"github.com/e2b-dev/infra/packages/shared/pkg/featureflags"
>  	templatemanager "github.com/e2b-dev/infra/packages/shared/pkg/grpc/template-manager"
>  	"github.com/e2b-dev/infra/packages/shared/pkg/logger"
> @@ -354,6 +355,13 @@ func doBuild(
>  		})
>  	}
>  
> +	// Mirror prod's gating (pkg/template/server/create_template.go:70): only
> +	// enable FPR if the chosen Firecracker version actually supports it.
> +	// Hardcoding true breaks builds on FC <1.14 with a balloon 400.
> +	fcInfo, err := fcversion.New(fc)
> +	if err != nil {
> +		return fmt.Errorf("invalid firecracker version %q: %w", fc, err)
> +	}
>  	tmpl := config.TemplateConfig{
>  		Version:            templates.TemplateV2LatestVersion,
>  		TemplateID:         templateID,
> @@ -366,7 +374,7 @@ func doBuild(
>  		ReadyCmd:           readyCmd,
>  		KernelVersion:      kernel,
>  		FirecrackerVersion: fc,
> -		FreePageReporting:  true,
> +		FreePageReporting:  fcInfo.HasFreePageReporting(),
>  		TeamID:             "local",
>  		Steps:              steps,
>  	}
> ```

We are gating the enabling here https://github.com/e2b-dev/infra/blob/feat/sandbox-pause-fph/packages/orchestrator/pkg/template/server/create_template.go#L70 and for now we will be using the feature flags context (which should carry the kernel and fc version) to check. At worst I think this should just fail and continue as the baloon device is not there.

packages/orchestrator/pkg/sandbox/fc/process.go
kalyazin: https://github.com/firecracker-microvm/firecracker/pull/5884 is merged. Going to be included in FC 1.16.

PR #2553, opened 2026-05-04
feat(base): disable proactive memory compaction in guest

ValentaTomas:
Adds `vm.compaction_proactiveness=0` to the base template's `/etc/sysctl.conf` so kcompactd no longer runs background page migrations in the guest.

With 2 MiB host-side hugepage backing of guest RAM, every migration dirties a destination hugepage from the host UFFD's perspective and lands in the next memfile diff — with no snapshot-aligned benefit. The pre-pause `compact_memory` write (#2551) does the work deterministically right before we capture state.

Existing templates inherit the change on rebuild.


matthewlouisbrockman (approved):
lgtm, we need to make sure 2551 goes out at same time/first?

PR #2555, opened 2026-05-04
ci: enable -race + add actionlint and govulncheck

ValentaTomas:
## What

Three small, high-signal CI additions:

- **`-race` in unit tests.** `CLAUDE.md` documents that tests should run with `-race`, but the workflow command didn't pass the flag. Enabled now; per-test timeout bumped 20m → 30m to absorb the slowdown. x86 only — race detection is architecture-independent so ARM64 doesn't gain extra coverage.
- **`actionlint`.** Lints workflow YAML; catches expression typos, deprecated runners, and most importantly script-injection patterns. Already paid for itself in this PR (see below). Shellcheck integration disabled on first introduction (~28 mostly style-level findings in pre-existing scripts; will clean up separately).
- **`govulncheck`.** Go team's official vulnerability scanner, matrixed per `go.work` module. Reachability-aware so it only flags CVEs in code paths actually called.

## Side effect: a real injection finding fixed

`actionlint` flagged `pr-no-generated-changes.yml:99`:

```yaml
git push origin HEAD:${{ github.head_ref }}
```

`github.head_ref` is the PR branch name, attacker-controlled. Routed through an env var so it can't be interpreted as shell metacharacters. Same treatment for `inputs.commit`.

## Why these specifically

Of the broader CI-quality candidates I considered, only these made the cut for "real signal, low ongoing maintenance, no platform lock-in". Skipped (each deserves its own PR/owner):

- **CodeQL** — overlaps with GitHub's UI-based "default setup". Verify in repo Security settings before adding a workflow.
- **tflint / tfsec** — large initial findings backlog; better with platform owner.
- **typos / dependabot.yml / container scan** — lower signal or different scope.

## Risk

The `-race` change is the only behavioral one; if pre-existing races surface and block CI, drop it from `pr-tests.yml` and land the rest. Both new workflows are pure additions.

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -1,0 +1,5 @@
+self-hosted-runner:
+  labels:
+    - ci-builder
+    - infra-tests
+    - infra-runner-arm
--- a/.github/workflows/govulncheck.yml
+++ b/.github/workflows/govulncheck.yml
@@ -1,0 +1,30 @@
+name: govulncheck
+
+on: [workflow_call]
+
+permissions:
+  contents: read
+
+jobs:
+  detect-modules:
+    runs-on: ubuntu-24.04
+    outputs:
+      modules: ${{ steps.set-modules.outputs.modules }}
+    steps:
+      - uses: actions/checkout@v5
+      - id: set-modules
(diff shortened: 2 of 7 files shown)


ValentaTomas:
Reverting scope; replacing with a smaller, targeted PR for the actual asks.

PR #2564, opened 2026-05-05
feat(provision): mask periodic rootfs-dirtying timers

ValentaTomas:
Mask periodic timers/daemons that scribble small files into the rootfs (audit, apt-daily, motd-news, man-db, e2scrub, fstrim, logrotate, unattended-upgrades, accounts-daemon, udisks2). Each wake-up dirties at least one 4 KiB block per run with no value for ephemeral sandboxes.

`systemctl mask` is a no-op on units not present, so this is safe across all customer base images.


ValentaTomas:

PR #2565, opened 2026-05-05
feat(ext4): disable journal on rootfs mkfs

ValentaTomas:
Drop `has_journal` from the rootfs mkfs feature list. The jbd2 commit timer dirties journal blocks every 5 s regardless of guest activity, inflating every snapshot diff. Also reclaims ~32 MiB at the front of the rootfs.

Snapshots come from a paused guest (no in-flight writes), so the journal's crash-recovery role across pause/resume is unused.

Risk: a guest kernel panic between snapshots can leave the rootfs inconsistent — resume always loads the previous good snapshot, so blast radius is bounded to one in-flight customer interaction. Want to discuss before flipping universally — feature flag is an easy follow-up.


ValentaTomas:

PR #2566, opened 2026-05-05
feat(ext4): pack metadata + larger flex group on rootfs mkfs

ValentaTomas:
Pass `-G 64` and `-E packed_meta_blocks=1` to `mkfs.ext4` on rootfs creation. Both options cluster ext4 metadata more tightly on disk so small-file metadata dirties land in fewer 4 MiB GCS chunks per snapshot.

mkfs-time only, no semantic change. Pairs with #2563 (inline_data).


ValentaTomas:

ValentaTomas:

ValentaTomas:

PR #2573, opened 2026-05-05
Create IAP for nomad

djeebus:


djeebus:
Turns out it's tedious to get nomad's terraform provider to communicate through IAP, and *really* tedious to convince terraform that IAP is optional, but not necessary. Going to punt on this.

PR #2578, opened 2026-05-06
fix(orch): GCS header request storm in Upload.Wait

levb:
Two fixes that drop orchestrator GCS header traffic on staging from
  ~2700 reads per pause/resume + layered-build trio down to 2.

  ### 1. `fix(build): one-shot LoadHeader on peer transition`

  `File.retryOnTransition` polled `PollRemoteStorageForHeader` with a
  30s budget. The transition is signaled only after the source upload
  finalized, so the header is already in storage — one `LoadHeader`
  suffices. Polling under high peer-transition rates (staging traffic
  sim) inflated GCS reads via backoff retries.

  ### 2. `fix(orch): stop O(ancestors×uploads) GCS header storm in Wait`

  Every pause / template build called `collectAncestorBuilds` → `Wait`
  per ancestor. `Wait` fell through to `LoadHeader` whenever the
  ancestor wasn't in the local templateCache (typical for
  grand-grandparents reached only via mappings). V3 discarded the
  result entirely (sync gate); V4 kept only `BuildData` it already had
  on `srcHeader.Builds`.

  `Wait` now returns immediately when the device's
  `IncompletePendingUpload` bit is clear, waits on the local future
  when one exists, and only polls remote storage when the new
  `peerclient.UploadChecker` confirms a peer is mid-upload.

### DIFF
--- a/packages/orchestrator/pkg/sandbox/build_upload_v4.go
+++ b/packages/orchestrator/pkg/sandbox/build_upload_v4.go
@@ -3,7 +3,6 @@ package sandbox
 import (
 	"context"
 	"fmt"
-	"maps"
 
 	"github.com/google/uuid"
 	"golang.org/x/sync/errgroup"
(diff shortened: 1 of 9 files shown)


packages/orchestrator/pkg/sandbox/build_upload_v4.go
```diff
@@ -98,45 +91,52 @@ func (u *Upload) uploadFramed(
...
 		if err != nil {
-			return nil, fmt.Errorf("wait for ancestor %s/%s: %w", m.BuildId, fileType, err)
+			return fmt.Errorf("wait for ancestor %s/%s: %w", m.BuildId, fileType, err)
 		}
-		// V3 ancestors have Builds=nil (FrameTable is V4-only); their data is
-		// raw bytes and the read path doesn't consult Builds for them. Skip
-		// silently so V4 descendants of V3 ancestors still upload.
-		bd, ok := h.Builds[m.BuildId]
-		if !ok {
+		if h == nil || dst == nil {
```
dobrac, line 130: how/why can the h be nil when err is not nil?

packages/orchestrator/pkg/sandbox/uploads.go
```diff
@@ -93,47 +95,49 @@ func (u *Uploads) Start(buildID uuid.UUID) (*utils.ErrorOnce, error) {
...
+// already carries its BuildData through srcHeader.Builds.
 func (u *Uploads) Wait(ctx context.Context, buildID uuid.UUID, t build.DiffType) (*header.Header, error) {
 	ctx, span := tracer.Start(ctx, "wait-for-parent-upload", trace.WithAttributes(
 		telemetry.WithBuildID(buildID.String()),
 		attribute.String("file_type", string(t)),
 	))
 	defer span.End()
 
+	d, err := u.find(ctx, buildID, t)
+	if err != nil && !errors.Is(err, ErrBuildNotInCache) {
```
dobrac, line 109: can we use the `ErrBuildNotInCache` instead of assuming `d ==nil <=> ErrBuildNotInCache`?

dobrac (changes requested):

dobrac (approved):

packages/orchestrator/pkg/sandbox/uploads.go
dobrac: lets address them in a follow up to not block this PR

packages/orchestrator/pkg/sandbox/build_upload_v4.go
dobrac: lets address them in a follow up to not block this PR

PR #2581, opened 2026-05-06
fix: allow header-based routing when Host is an IP address

tomassrnka:
## Summary

Fixes a regression from #2540 that broke SDK connectivity for any IP-based access to the client-proxy (bridge networks, LAN, on-prem deployments, e2b-local).

- `isLocalRequestHost()` only accepted `localhost` and loopback IPs (`127.0.0.1`, `::1`)
- When the SDK connects via an IP like `192.168.100.66:3002`, routing headers (`E2b-Sandbox-Id`, `E2b-Sandbox-Port`) were silently ignored
- `parseHost()` then tried to extract a sandbox ID from the IP address and failed with `ErrInvalidHost` → HTTP 400

**The fix:** widen `isLocalRequestHost()` from `ip.IsLoopback()` to `ip != nil`.

This is safe because the `shouldParseHeaders` guard exists to prevent header spoofing on per-sandbox hostnames like `49983-sbx.e2b.app`, where the hostname is authoritative. An IP address can never contain sandbox routing info — `parseHost()` will always fail on it — so there is nothing for headers to conflict with. Headers are the only routing mechanism that can work for IP hosts.

The per-sandbox hostname guard is unaffected: headers are still ignored when `r.Host` is a domain like `49983-sbx.e2b.app`.

## How it was found

e2b-local nightly build on `2026-05-05` failed the sandbox smoke test. The build itself succeeded, but `sandbox.commands.run()` returned `HTTP 400` because the client-proxy rejected the bridge network host `192.168.100.66:3002`:

```
WARN  invalid host  {"host": "192.168.100.66:3002"}  handler.go:30
```

The previous night's build (using commit `92ea305dd`, before #2540) passed. Today's build (using `ae68b805a`, after #2540) failed.

## Test plan

- [x] All existing `TestGetTargetFromRequest` tests pass
- [x] Added test cases for private IPv4, 10.x network, and public IPv4 hosts
- [ ] Verify e2b-local nightly build passes with this fix

### DIFF
--- a/packages/shared/pkg/proxy/host.go
+++ b/packages/shared/pkg/proxy/host.go
@@ -56,7 +56,9 @@ func isLocalRequestHost(host string) bool {
 	host = requestHostname(host)
 	ip := net.ParseIP(host)
 
-	return host == "localhost" || (ip != nil && ip.IsLoopback())
+	// An IP address cannot encode sandbox routing info (like {port}-{sandboxId}.{domain}),
+	// so header-based routing is the only mechanism that can work for IP hosts.
+	return host == "localhost" || ip != nil
 }
 
 func SandboxSharedHostDomain(host string) (string, bool) {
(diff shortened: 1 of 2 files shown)


matthewlouisbrockman (approved):
oh.

PR #2585, opened 2026-05-07
fix(peerclient): resolve base path per-call from live FrameTable

levb:
After a peer→storage transition, peerSeekable kept opening base against the original (uncompressed) path captured at construction. With the recent compression work, post-transition reads now target a compressed object, so the cached base resolved to a non-existent path.

P2P routing produced the stale binding (it captured a path while the build was still uncompressed); it also contains the fix. peerSeekable now holds (buildID, basic name, base provider, objType) and composes the actual storage path at base-open time using the CompressionType from the live FrameTable. The base seekable is reopened on ct change. No changes outside peerclient.


packages/orchestrator/pkg/sandbox/template/peerclient/storage.go
```diff
@@ -130,29 +129,41 @@ func newPeerStorageProvider(
...
-		uploaded: p.uploaded,
-		openFn: func(ctx context.Context) (storage.Seekable, error) {
-			return p.base.OpenSeekable(ctx, path, objType)
+	fileName = storage.StripCompression(fileName)
+
+	return &peerSeekable{
+		peerHandle: peerHandle{
+			client:   p.peerClient,
+			buildID:  buildID,
+			fileName: fileName,
```
dobrac: lets rename the variable so we know its without the suffix, actually, shouldn't this be just the type instead of file name?

dobrac (changes requested):

dobrac (approved):

PR #2590, opened 2026-05-07
Dirty pages deduplication between diff and base layer for memory

bchalios:
Adds 4 KiB memfile diff dedup behind `memfile-diff-dedup` to avoid storing pages unchanged from the base when Firecracker reports dirty memory at 2 MiB granularity.

Also fixes related page-granular restore correctness for empty mappings and chunk reads, and runs integration pause/resume coverage across base, best-effort, and direct I/O dedup modes.

Stacked on #2522.


packages/orchestrator/pkg/sandbox/build/cache.go
```diff
@@ -156,6 +156,18 @@ func (s *DiffStore) Add(d Diff) {
...
+// Get'ed yet. Unlike Get, never calls diff.Init — safe for best-effort code
+// paths that must avoid triggering remote I/O.
+func (s *DiffStore) Peek(key DiffStoreKey) Diff {
+	item := s.cache.Get(key)
+	if item == nil {
+		return nil
+	}
+
+	return item.Value()
+}
```
arkamar: We already have a method which does the same thing, Let's throw this one in favor of `Lookup()` instead. There is just one use of `Peek` anyway.

packages/orchestrator/pkg/sandbox/build/build.go
```diff
@@ -120,41 +119,68 @@ func (b *File) ReadAt(ctx context.Context, p []byte, off int64) (n int, err erro
...
 		}
 
-		result, err := diff.Slice(ctx, int64(mappedBuild.Offset), int64(h.Metadata.BlockSize), ft)
-		if err != nil {
-			if retry, swapErr := b.retryOnTransition(ctx, err); retry {
-				continue
-			} else if swapErr != nil {
-				return nil, swapErr
+		if m.BuildId != uuid.Nil {
+			diff := b.store.Peek(GetDiffStoreKey(m.BuildId.String(), b.fileType))
```
arkamar:
```suggestion
			diff, _ := b.store.Lookup(GetDiffStoreKey(m.BuildId.String(), b.fileType))
```
No need to introduce method with the same functionality.

packages/orchestrator/pkg/sandbox/build/cache.go
```diff
@@ -156,6 +156,18 @@ func (s *DiffStore) Add(d Diff) {
...
+
```
arkamar:
```suggestion
```
Now unused with respect to previous comment.

packages/orchestrator/pkg/sandbox/block/memfd.go
```diff
@@ -233,3 +237,59 @@ func (m *MemfdCache) FileSize(ctx context.Context) (int64, error) {
...
+	for len(iovs) > 0 {
+		n, err := unix.Pwritev(fd, iovs, off)
+		if errors.Is(err, unix.EINTR) {
+			continue
+		}
+		if err != nil {
+			return err
+		}
+		if n == 0 {
+			return fmt.Errorf("pwritev: EOF with %d iovec(s) remaining", len(iovs))
```
arkamar, line 286: Just for the awareness, `iovs := [][]byte{{}}` would emit this error as well, but the write would be technically legit. This error is a bit weird because of that.

arkamar (commented):
Noticed few nit things, but wasn't able to finish it completely.

packages/orchestrator/pkg/sandbox/build/build.go
```diff
@@ -120,41 +119,68 @@ func (b *File) ReadAt(ctx context.Context, p []byte, off int64) (n int, err erro
...
+			if m.BuildId != uuid.Nil {
+				size := b.buildFileSize(h, m.BuildId)
+				ft := h.GetBuildFrameData(m.BuildId)
+				diff, derr := b.getBuild(ctx, m.BuildId, size, ft.CompressionType())
+				if derr == nil {
+					if slice, sErr := diff.Slice(ctx, int64(m.Offset), length, ft); sErr == nil {
+						return slice, nil
+					}
+				}
+				// Errors fall through to ReadAt's retry-on-transition path.
```
jakubno, line 145: shouldn't we at least log `derr` and `sErr`?

packages/orchestrator/pkg/sandbox/block/cache.go
```diff
@@ -203,6 +206,220 @@ func (c *Cache) ExportToDiff(ctx context.Context, out *os.File) (*header.DiffMet
...
+	packed := make(map[int64]int64, dirty.GetCardinality())
+	var cum int64
+	for r := range BitsetRanges(dirty, blockSize) {
+		for chunkOff := int64(0); chunkOff < r.Size; chunkOff += blockSize {
+			packed[r.Start+chunkOff] = cum
+			cum += blockSize
+		}
+	}
+	src := func(absOff int64) ([]byte, error) {
+		return c.Slice(packed[absOff], blockSize)
```
jakubno: do we want to be defensive?
```suggestion
		idx, ok := packed[absOff]
		if !ok { 
			return nil, fmt.Errorf("dedup src: %d not packed", absOff)
		}

		return c.Slice(idx, blockSize)
```

packages/orchestrator/pkg/sandbox/uffd/userfaultfd/userfaultfd.go
```diff
@@ -49,10 +49,25 @@ func hasEvent(revents, event int16) bool {
...
+// PageReader is the data source UFFD pulls page contents from on a fault.
+type PageReader interface {
+	ReadAt(ctx context.Context, p []byte, off int64) (int, error)
+}
+
+// pagePool returns HugepageSize-sized scratch buffers shared across all
+// UFFD instances; 4 KiB-page UFFDs reslice the buffer down to pageSize.
+var pagePool = sync.Pool{
+	New: func() any {
+		buf := make([]byte, header.HugepageSize)
```
jakubno, line 61: is the [comment](https://github.com/e2b-dev/infra/pull/2590/changes#r3271368151) legit? Should we add some comment to change it after migrating?

packages/orchestrator/pkg/sandbox/uffd/userfaultfd/userfaultfd.go
```diff
@@ -108,13 +123,16 @@ const (
...
-	for _, region := range m.Regions {
-		if region.PageSize != uintptr(blockSize) {
-			return nil, fmt.Errorf("block size mismatch: %d != %d for region %d", region.PageSize, blockSize, region.BaseHostVirtAddr)
+// NewUserfaultfdFromFd creates a new userfaultfd instance. Page size is
+// taken from the FC-registered regions; all regions must agree.
+func NewUserfaultfdFromFd(fd uintptr, src PageReader, m *memory.Mapping, logger logger.Logger) (*Userfaultfd, error) {
+	if len(m.Regions) == 0 {
+		return nil, errors.New("memory mapping has no regions")
+	}
+	pageSize := m.Regions[0].PageSize
```
jakubno, line 132: maybe worth adding a check that it's either pageSize or hugepageSize

packages/orchestrator/pkg/sandbox/uffd/userfaultfd/userfaultfd.go
jakubno, line 502: if this fail, wake won't run, isn't that a problem?

jakubno (commented):
shouldn't we add `IsCached` to `localDiff` returning `true`?

packages/orchestrator/pkg/sandbox/build/cache.go
ValentaTomas: Done, removed `Peek` and switched the call site to the existing `Lookup`.

packages/orchestrator/pkg/sandbox/build/build.go
ValentaTomas: Done, using `Lookup` here now.

packages/orchestrator/pkg/sandbox/build/cache.go
ValentaTomas: Done, `Peek` is removed.

packages/orchestrator/pkg/sandbox/block/memfd.go
ValentaTomas: Handled by skipping empty iovecs before calling pwritev.

packages/orchestrator/pkg/sandbox/build/build.go
ValentaTomas: Done, both fast-path errors are logged before falling back to ReadAt.

packages/orchestrator/pkg/sandbox/uffd/userfaultfd/userfaultfd.go
ValentaTomas: Done, the pool is now used only for hugepage faults; 4 KiB faults allocate 4 KiB buffers.

packages/orchestrator/pkg/sandbox/uffd/userfaultfd/userfaultfd.go
ValentaTomas: Done, NewUserfaultfdFromFd now rejects page sizes other than 4 KiB or 2 MiB.

packages/orchestrator/pkg/sandbox/uffd/userfaultfd/userfaultfd.go
```diff
@@ -49,10 +49,25 @@ func hasEvent(revents, event int16) bool {
...
+	ReadAt(ctx context.Context, p []byte, off int64) (int, error)
+}
+
+// pagePool returns HugepageSize-sized scratch buffers shared across all
+// UFFD instances; 4 KiB-page UFFDs reslice the buffer down to pageSize.
+var pagePool = sync.Pool{
+	New: func() any {
+		buf := make([]byte, header.HugepageSize)
+
+		return &buf
```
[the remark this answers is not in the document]
ValentaTomas: Done, the hugepage pool is only used for 2 MiB faults now.

packages/orchestrator/pkg/sandbox/block/cache.go
```diff
@@ -203,6 +206,211 @@ func (c *Cache) ExportToDiff(ctx context.Context, out *os.File) (*header.DiffMet
...
+	return cache, &header.DiffMetadata{
+		Dirty:     pageDirty,
+		Empty:     pageEmpty,
+		BlockSize: header.PageSize,
+	}, nil
+}
+
+// drainDirtyPages reconstructs iovs from the bitmap and drains them.
+func drainDirtyPages(ctx context.Context, fd int, src func(absOff int64) ([]byte, error), pageDirty *roaring.Bitmap, blockSize int64) (int64, error) {
+	iovs := make([][]byte, 0, pageDirty.GetCardinality())
```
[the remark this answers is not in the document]
ValentaTomas: Done, dirty pages are now streamed into bounded pwritev batches instead of materializing one iovec per dirty page.

packages/orchestrator/pkg/sandbox/uffd/userfaultfd/userfaultfd.go
ValentaTomas: Fixed by waking after zero-fill if write-protect fails, before returning the error.

ValentaTomas: Also this is only for 4KiB pages.

ValentaTomas (approved):

PR #2595, opened 2026-05-08
fix(volumes): broadcast volume directory creation to all cluster nodes

AdaAibaby:
## Problem

When a volume is created via POST /volumes, the directory is only
created on a single randomly-selected orchestrator node
(`executeOnOrchestratorByClusterID` uses rand.Shuffle and returns on
the first success). Sandbox placement (PlaceSandbox) is independent
and selects nodes based on CPU/RAM availability, with no awareness of
which node holds the volume directory.

This causes an intermittent failure: sandboxes that land on the same
node as the volume get a working NFS mount; sandboxes that land on any
other node fail to mount because the NFS proxy calls
syscall.Mount(path, path, MS_BIND) on a path that does not exist on
that node.

A second, related bug exists in `envd/internal/api/init.go`: when two
concurrent `/init` requests race on `isMountingNFS`, the second request
silently returns `nil`, causing the sandbox to start successfully with
no NFS mount.

## Changes

**volume_util.go**
- Add `executeOnAllClusterNodes`: fans out a gRPC call concurrently to
  every ready node in the cluster using errgroup, returning a combined
  error if any node fails.

**volume_create.go**
- `createVolume` now calls `executeOnAllClusterNodes` instead of
  `executeOnOrchestratorByClusterID`, so the volume directory is
  created on every orchestrator node at creation time.

**volume_delete.go**
- `deleteVolume` likewise broadcasts to all nodes to clean up the
  directory everywhere it was created.

**init.go**
- `setupNFS`: when `isMountingNFS` CAS fails (concurrent request),
  return an explicit error instead of `nil` so the caller receives a
  `400` and can retry, rather than silently succeeding with no mount.

## Testing

- `go build ./packages/api/... ./packages/envd/... packages.` ✅
- `go vet ./packages/api/internal/handlers/... internal.` ✅
- `go test internal.` ✅
- Manual: on a staging cluster with ≥2 orchestrator nodes, creating a
  volume and then creating 10 sandboxes with that volume mount all
  succeed with a valid NFS mount (mount | grep nfs non-empty in every
  sandbox).

### DIFF
--- a/packages/api/internal/handlers/volume_create.go
+++ b/packages/api/internal/handlers/volume_create.go
@@ -181,7 +181,10 @@ func isValidVolumeName(name string) bool {
 }
 
 func (a *APIStore) createVolume(ctx context.Context, clusterID uuid.UUID, volume queries.Volume) error {
-	return a.executeOnOrchestratorByClusterID(ctx, clusterID, func(ctx context.Context, client *clusters.GRPCClient) error {
+	// Broadcast to all nodes so the volume directory exists on every orchestrator
+	// node. Sandbox placement is independent of volume creation and may schedule
+	// a sandbox on any node in the cluster.
+	return a.executeOnAllClusterNodes(ctx, clusterID, func(ctx context.Context, client *clusters.GRPCClient) error {
 		_, err := client.Volumes.CreateVolume(ctx, &orchestrator.CreateVolumeRequest{
 			Volume: toVolumeKey(volume),
 		})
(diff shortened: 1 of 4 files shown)


AdaAibaby:
> @cla-bot check

@cla-bot check

adababys:
> We require contributors to sign our Contributor License Agreement, and we don't have @adababys on file. You can sign our CLA at https://e2b.dev/docs/cla . Once you've signed, post a comment here that says '@cla-bot check'

@cla-bot check

ValentaTomas:

PR #2601, opened 2026-05-09
refactor(api): introduce pluggable service-discovery interface

tvi:
The api binary previously took a *nomadapi.Client directly into the Orchestrator and clusters.Pool to enumerate orchestrator (Firecracker host) and template-manager (template-build host) instances.

This change introduces a Discovery interface so the Nomad-specific listing logic can be swapped out without touching the rest of the orchestrator/clusters code paths:

  - new orchestrator/discovery package with a Discovery interface and a NewNomad implementation that wraps the existing /v1/nodes call
  - Orchestrator.New now takes a discovery.Discovery instead of a *nomadapi.Client; listNomadNodes dispatches through it
  - clusters.NewPool / newLocalCluster take a clustersdiscovery.Discovery instead of a *nomadapi.Client; the existing LocalServiceDiscovery is the only implementation
  - handlers/store.go constructs the Nomad-backed Discovery once at startup and passes it to both consumers
  - client_test.go updated to construct an Orchestrator with the new nodeDiscovery field

Behavior is unchanged: the api still talks to the local Nomad agent exactly as before.

### DIFF
--- a/packages/api/internal/orchestrator/discovery/nomad.go
+++ b/packages/api/internal/orchestrator/discovery/nomad.go
@@ -1,0 +1,56 @@
+package discovery
+
+import (
+	"context"
+	"fmt"
+
+	nomadapi "github.com/hashicorp/nomad/api"
+
+	"github.com/e2b-dev/infra/packages/shared/pkg/consts"
+)
+
+// nomadDiscovery implements Discovery against a local Nomad HTTP agent.
+//
+// In the Nomad deploy each client node has the orchestrator binary running
+// directly (raw_exec) and listening on consts.OrchestratorAPIPort, so every
+// "ready" Nomad node in the configured pool is an orchestrator.
+type nomadDiscovery struct {
+	client   *nomadapi.Client
+	nodePool string
+}
+
+// NewNomad creates a Nomad-backed Discovery.
+func NewNomad(client *nomadapi.Client, nodePool string) Discovery {
+	return &nomadDiscovery{
+		client:   client,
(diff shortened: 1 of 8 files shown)


packages/api/internal/orchestrator/discovery/nomad.go
```diff
@@ -0,0 +1,56 @@
...
+		nodePool = "default"
+	}
+	return &nomadDiscovery{
+		client:   client,
+		nodePool: nodePool,
+	}
+}
+
+func (d *nomadDiscovery) ListNodes(ctx context.Context) ([]Node, error) {
+	_, span := tracer.Start(ctx, "list-nomad-nodes")
```
dobrac: missing use of ctx

packages/api/internal/orchestrator/discovery/nomad.go
```diff
@@ -0,0 +1,56 @@
...
+// "ready" Nomad node in the configured pool is an orchestrator.
+type nomadDiscovery struct {
+	client   *nomadapi.Client
+	nodePool string
+}
+
+// NewNomad creates a Nomad-backed Discovery. nodePool defaults to "default" when
+// empty (matches the historical filter used by listNomadNodes).
+func NewNomad(client *nomadapi.Client, nodePool string) Discovery {
+	if nodePool == "" {
```
dobrac: do we need the fallback?

dobrac (approved):
lgtm! Two nits

dobrac (approved):

PR #2614, opened 2026-05-10
refactor(orchestrator/buildlogger): implement zapcore.Core directly

tvi:
LogEntryLogger previously implemented io.Writer and was plugged into a zap pipeline via zapcore.NewCore(jsonEncoder, logEntryLogger, DebugLevel). Each entry was serialized to JSON by the encoder and then parsed back out by Write to recover the message, level, timestamp, and fields. Two of those parse paths could fail and called logger.L().Error(context.TODO(), ...) because Write has no context.

Implement zapcore.Core on LogEntryLogger instead: Check, Write(Entry, []Field), Sync, With, plus an embedded zapcore.LevelEnabler. The Entry struct already carries level, time, and message, and []Field is encoded straight into a map[string]string via zapcore.MapObjectEncoder, so the JSON encode/parse round-trip and its failure modes disappear, and with them the two context.TODO() call sites.

With returns a childCore that shares the parent's capture slice and mu and only carries the accumulated fields, so logger.With(...) chains do not fork the capture buffer and Lines() still observes every entry.

Wire-up in TemplateCreate is simplified: drop the JSON encoder and the zapcore.NewCore wrapper and tee LogEntryLogger directly with the build logger core. Test helper writeTestBuildLogs constructs zapcore.Entry values instead of marshalling synthetic JSON and calling Write([]byte).

### DIFF
--- a/packages/orchestrator/pkg/template/build/buildlogger/log_entry_logger.go
+++ b/packages/orchestrator/pkg/template/build/buildlogger/log_entry_logger.go
@@ -1,98 +1,107 @@
 package buildlogger
 
 import (
-	"bytes"
-	"context"
-	"encoding/json"
+	"fmt"
 	"sync"
-	"time"
 
-	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
 	"google.golang.org/protobuf/types/known/timestamppb"
 
 	template_manager "github.com/e2b-dev/infra/packages/shared/pkg/grpc/template-manager"
-	"github.com/e2b-dev/infra/packages/shared/pkg/logger"
-	"github.com/e2b-dev/infra/packages/shared/pkg/logs"
 )
 
-type ZapEntry struct {
-	Ts    float64 `json:"ts"`
-	Msg   string  `json:"msg"`
-	Level string  `json:"level"`
-}
-
+// LogEntryLogger is a zapcore.Core that captures every log entry into an
+// in-memory slice of template_manager.TemplateBuildLogEntry. It replaces the
+// older io.Writer-based implementation that round-tripped log lines through
+// JSON; consuming entries directly avoids that and removes the need for any
+// fallback error-logging context.
 type LogEntryLogger struct {
+	zapcore.LevelEnabler
+
 	mu    sync.Mutex
 	lines []*template_manager.TemplateBuildLogEntry
 }
 
+// Compile-time assertion: LogEntryLogger implements zapcore.Core.
+var _ zapcore.Core = (*LogEntryLogger)(nil)
+
 func NewLogEntryLogger() *LogEntryLogger {
 	return &LogEntryLogger{
-		lines: make([]*template_manager.TemplateBuildLogEntry, 0),
+		LevelEnabler: zapcore.DebugLevel,
+		lines:        make([]*template_manager.TemplateBuildLogEntry, 0),
 	}
 }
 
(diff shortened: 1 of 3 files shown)


packages/orchestrator/pkg/template/build/buildlogger/log_entry_logger.go
```diff
@@ -1,100 +1,99 @@
...
 
-				continue
-			}
+func (b *LogEntryLogger) Write(ent zapcore.Entry, fields []zapcore.Field) error {
+	return b.write(ent, nil, fields)
+}
 
-			timestamp := epochToTime(entry.Ts)
+func (b *LogEntryLogger) Sync() error {
+	// In-memory capture; nothing to flush.
```
dobrac: if a write to lines is happening and Sync is used as a barrier, you might access lines while the slice change is happening - causing data race

dobrac (changes requested):

packages/orchestrator/pkg/template/build/buildlogger/log_entry_logger.go
tvi: Looking at the code: `Sync()` doesn't actually access `lines` at all — it just returns nil. So there's no data race in `Sync()` itself. The `Write` and `Lines` methods both properly hold `b.mu` when accessing `b.lines`.

The reviewer's claim is that "if Sync is used as a barrier, you might access lines while the slice change is happening." But `Sync()` here is a no-op that returns immediately — it doesn't read `lines`, so it can't race on them. The actual barrier semantics are provided by the mutex in `Write` and `Lines`.

The reviewer's concern doesn't apply to this code as written. `Sync()` performs no access to `lines`, so it cannot be involved in a data race on that slice. Concurrency safety for `lines` is correctly enforced by `b.mu` in `Write()` and `Lines()`. If the intent is for `Sync()` to act as a barrier, that's a separate design question, but the current implementation does not introduce a race because it touches no shared state.

dobrac: Can you please check if the git history of adding the lock there was a fix or an initial implementation? The synchronization might be here because of guaranteeing that if the template status is done/success, no new lines are pending addition to the lines slice.

tvi: Found it. In commit 9b0cbd50f, the previous no-op `Sync` was changed to acquire the mutex with the comment:

`// No-op for SafeBuffer, as it doesn't have an underlying file to sync`
`// But wait for the mutex to ensure no writes are happening`

The reviewer's intuition is correct — the lock in `Sync` was added intentionally as a barrier, not as a fix for accessing `lines`. It was meant to ensure that when `Sync()` returns, no `Write` is in flight, so subsequent code (e.g., reading the final state when a build transitions to done/success) sees a quiesced logger.

Reverting `Sync()` to a pure no-op (current `7ea52e77a` refactor) drops that barrier semantics. The downstream concern is not a data race on `lines` itself (Write/Lines are still mutex-protected), but rather that callers that previously relied on `Sync()` returning only after all in-flight `Write` calls completed may now observe partial state.

The fix is to restore the `b.mu.Lock(); b.mu.Unlock()` in `Sync()` (with the original comment) so it continues to act as a quiescence barrier. The reviewer was right — please put the lock back.

dobrac (approved):

PR #2620, opened 2026-05-11
fix(envd): stop memory exhaustion when client disconnects from streaming process

arkamar:
When a Start/Connect client disconnects while a process is producing output, the fan-out loop keeps draining the Source channel with no subscribers — discarding every value but keeping the reader goroutines hot. Each reader allocates 32 KiB per read cycle, and with a fast producer envd RSS grows to hundreds of MiB in seconds, OOM-killing other processes in the sandbox.

The fix adds back-pressure: when no active subscribers remain, the fan-out stops consuming from Source, the channel fills, the reader blocks, and the OS pipe back-pressures the child. The child resumes instantly when a new client calls Fork. This is safe because output with no subscribers is already lost — the SDK has no replay mechanism.

Stdout/stderr pipes are replaced with manual `os.Pipe()` so `cmd.Wait()` doesn't close the read-ends prematurely, fixing output truncation on fast commands. After the child is reaped, `SetReadDeadline` on the read-ends lets readers drain buffered data then exit cleanly instead of blocking forever when an orphan grandchild holds the write-end open. Reader goroutines close their own read-ends via defer.

EndEvent buffer is changed from 0 to 1 so the send in `Wait()` succeeds when the fan-out is parked. The `outCtx`/`outCancel` mechanism is removed entirely — a dedicated `readersDone` channel tracks when readers actually exit, which `SendSignal(SIGKILL)` can no longer bypass. `close(m.done)` is moved under the lock so `Fork()` racing with shutdown cannot orphan a subscriber.

### DIFF
--- a/packages/envd/internal/services/process/handler/handler.go
+++ b/packages/envd/internal/services/process/handler/handler.go
@@ -206,12 +206,18 @@ func New(
 				n, readErr := tty.Read(buf)
 
 				if n > 0 {
-					outMultiplex.Source <- rpc.ProcessEvent_Data{
+					event := rpc.ProcessEvent_Data{
 						Data: &rpc.ProcessEvent_DataEvent{
 							Output: &rpc.ProcessEvent_DataEvent_Pty{
 								Pty: buf[:n],
 							},
 						},
+					}
+
+					select {
+					case outMultiplex.Source <- event:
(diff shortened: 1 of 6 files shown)


arkamar:
closing in favor of #2639

PR #2633, opened 2026-05-12
Make possible to configure own vpc and subnet for cluster image build

sitole:


iac/provider-gcp/nomad-cluster-disk-image/main.tf
```diff
@@ -1,8 +1,10 @@
 terraform {
-  required_version = ">= 1.5.0, < 1.6.0"
+  required_version = ">= 1.7.0, < 1.8.0"
```
jakubno: please revert this as we use `  required_version = ">= 1.5.0, < 1.6.0"` in all other places in this repository

jakubno (changes requested):
one nit

iac/provider-gcp/nomad-cluster-disk-image/main.tf
sitole: @djeebus is there a block to update the infra TF? Is it planned on your end, or should I do it?

sitole: @jakubno i reverted required version back to `>= 1.5.0, < 1.6.0` so we can merge

jakubno (approved):

PR #2634, opened 2026-05-12
chore(orch): remove unnecessary `ALLOW_SANDBOX_INTERNET`

jakubno:
API is setting the network config based on that parameter already, so this isn't needed anymore


arkamar (approved):
lgtm

Issue #2637, opened 2026-05-13
[Bug]: Missing teamID filtering in snapshot query, permission security risk, need to fix ownership check logic


AdaAibaby:
### Environment

e2b == main
e2b sdk == 2.20.0

### Timestamp of the issue

2026-05-11 15:00 UTC

### Frequency

Happens every time

### Expected behavior

1. SnapshotCache.Get supports teamID parameter
2. Database query GetLastSnapshot supports filtering by teamID, only returns snapshots accessible to the current team
3. No additional post-hoc ownership check in 4 handlers, data permission is secured at the database layer
4. Strict team-level isolation for snapshot data without unauthorized access

### Actual behavior

1. Existing snapshot query does not support teamID filtering, only queryable by snapshot ID
2. No team permission filtering at the database layer, risk of unauthorized snapshot access
3. TODO comments and post-hoc ownership checks remain in 4 handlers, redundant and insecure
4. No team-level data isolation in cache interface and database queries

### Issue reproduction

1. Call SnapshotCache.Get to retrieve snapshot without teamID permission filtering
2. Check database query logic: GetLastSnapshot does not support teamID parameter
3. Check 4 handler files with TODO: ownership check comments
4. No team isolation at data level, leading to security risks

### Additional context

Need to add new SQL query GetLastSnapshotByTeam with team_id filtering;Regenerate sqlc Go code;Upgrade SnapshotCache.Get to support teamID;Remove redundant post-hoc ownership checks and TODO comments in handlers.

jakubno:
As mentioned in the PR, there isn't really an issue and maintaining 2 methods doing basically the same, where one also do auth introduces more security risk. Feel free to reopen with alternative solution , closing for now.

PR #2638, opened 2026-05-13
fix(snapshot): scope GetLastSnapshot query by teamID to prevent unauthorized cross-team access

AdaAibaby:
Fixes #[2637](https://github.com/e2b-dev/infra/issues/2637)

## Problem

Four handlers (`sandbox_connect`, `sandbox_get`, `sandbox_pause`, `sandbox_resume`) called
`SnapshotCache.Get(ctx, sandboxID)` which fetches a snapshot by sandbox ID only, then
performed a **post-fetch** ownership check:

```go
if lastSnapshot.Snapshot.TeamID != teamID {
    return api.NewError(http.StatusNotFound, "snapshot not found")
}


jakubno:
This isn't really possible as we need to access the snapshot in some places without knowing team ID apriori and we don't want to maintain 2 methods doing the same. Please propose the solution in the issue before starting the implementation next time 🙏🏻

AdaAibaby:
> This isn't really possible as we need to access the snapshot in some places without knowing team ID apriori and we don't want to maintain 2 methods doing the same. Please propose the solution in the issue before starting the implementation next time 🙏🏻

Thanks a lot for the explanation, Jakub — that makes perfect sense.
I didn’t realize some internal paths access snapshots without having the teamID available upfront.
I’ll move the discussion back to #2637​ and propose a solution that works within the current architecture (without introducing a second method).
Appreciate you taking the time to review this 🙏

PR #2642, opened 2026-05-13
feat(api): SOCKS5 egress proxy on sandbox network config (BYOP)

arkamar:
Sandboxes have two related egress problems we cannot solve from our
side: operators want a predictable egress identity (their public IP,
not ours), and they want sandboxes to reach hosts that are reachable
from their own proxy server — corporate networks, VPN endpoints,
internal LAN. This PR adds Bring-Your-Own-Proxy: an optional
`egressProxy: {address, username, password}` on the sandbox network
config tunnels all sandbox TCP egress through a user-supplied SOCKS5
server, with our allow/deny rules still applied before the tunnel.

Scope here is the API + orchestrator plumbing: OpenAPI, DB types,
proto, the API create/get handlers (password omitted from GET), the
`BYOPProxyEnabledFlag` LD gate, and a validator that rejects a SOCKS5
endpoint pointing at our own infra at create time and re-checks on
every dial (DNS-rebind guard). For BYOP-enabled sandboxes only, the
sandbox-netns kernel firewall is narrowed on demand so TCP destined
for the user's internal networks reaches the userspace egress proxy
instead of being dropped — without this the user's LAN destinations
would never make it past the kernel. Non-BYOP sandboxes keep the
existing hard kernel drop.


packages/orchestrator/pkg/server/sandboxes.go
```diff
@@ -287,6 +305,14 @@ func (s *Server) Update(ctx context.Context, req *orchestrator.SandboxUpdateRequ
 		return nil, status.Error(codes.NotFound, "sandbox not found")
 	}
 
+	// Mirror the Create-side BYOP gate; defense-in-depth for direct gRPC callers.
+	if req.GetEgress().GetEgressProxyAddress() != "" && !s.sandboxFactory.EgressProxy().SupportsBYOP() {
+		telemetry.ReportEvent(ctx, "egressProxy update rejected: orchestrator build has no BYOP dialer")
+
+		return nil, status.Error(codes.Unimplemented,
+			"egress proxy is not supported by this orchestrator build")
+	}
```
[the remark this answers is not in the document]
arkamar: should be addressed in 24b9426de30f ("feat(orch/server): gate BYOP egress proxy updates behind LaunchDarkly flag")

packages/orchestrator/pkg/sandbox/network/firewall.go
```diff
@@ -147,7 +152,8 @@ func (fw *Firewall) addSetFilterRule(ipSet *nftables.Set, drop bool) {
 		Chain: fw.filterChain,
 		Exprs: append(append(fw.tapIfaceMatch(),
 			expressions.IPv4DestinationAddress(1),
-			expressions.IPSetLookUp(ipSet, 1)),
+			expressions.IPSetLookUp(ipSet, 1),
+			&expr.Counter{}),
```
jakubno: are we using the counters somewhere?

arkamar: nope, it was for debugging, I dropped the commit.

arkamar:
rebased because of merge conflicts.

arkamar:
FP rebased branch due to merge conflicts.

packages/orchestrator/pkg/server/sandboxes.go
```diff
@@ -105,6 +105,23 @@ func (s *Server) Create(ctx context.Context, req *orchestrator.SandboxCreateRequ
...
+			return nil, status.Error(codes.PermissionDenied,
+				"egress proxy is not enabled for this team")
+		}
+		if !s.sandboxFactory.EgressProxy().SupportsBYOP() {
+			telemetry.ReportEvent(ctx, "egressProxy rejected: orchestrator build has no BYOP dialer")
+
+			return nil, status.Error(codes.Unimplemented,
+				"egress proxy is not supported by this orchestrator build")
+		}
+	}
```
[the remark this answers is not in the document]
arkamar: Orchestrator gRPC should be reachable solely by the API, which already runs ValidateEgressProxy.

packages/api/internal/handlers/sandbox_network_update.go
```diff
@@ -52,6 +55,31 @@ func (a *APIStore) PutSandboxesSandboxIDNetwork(c *gin.Context, sandboxID string
...
+			Password: sharedUtils.DerefOrDefault(ep.Password, ""),
+		}, nil)
+		if err != nil {
+			telemetry.ReportError(ctx, "invalid egress proxy config", err)
+			a.sendAPIStoreError(c, http.StatusBadRequest, fmt.Sprintf("Invalid egress proxy config: %s", err))
+
+			return
+		}
+
+		egressProxy = canonical
```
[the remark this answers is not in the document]
arkamar: The PUT behavior is consistent with other fields in the entry point. The PATCH should be introduced for reconfiguration. GET not sending the password feels natural to me.

packages/shared/pkg/sandbox-network/byop.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+	parts := strings.Split(raw, ",")
+	out := make([]string, 0, len(parts))
+	for _, p := range parts {
+		p = strings.TrimSpace(p)
+		if p == "" {
+			continue
+		}
+		if _, _, err := net.ParseCIDR(p); err != nil {
+			continue
```
jakubno: maybe worth a log?

packages/shared/pkg/sandbox-network/byop.go
```diff
@@ -0,0 +1,194 @@
...
+// a canonical copy (host lower-cased, whitespace trimmed); the input is not
+// mutated. cfg == nil returns (nil, nil). resolve defaults to DefaultHostResolver.
+//
+// Rules:
+//   - Address must parse as "host:port" with a non-zero port.
+//   - Host must be an IP literal or resolve via the provided resolver.
+//   - Every resolved A/AAAA record must NOT be in DeniedSandboxCIDRs.
+//   - If Username == "" then Password must also be "" (no orphan password).
+//   - Username and Password may contain {{sandboxID}} literally.
+func ValidateEgressProxy(ctx context.Context, cfg *EgressProxyConfig, resolve HostResolver) (*EgressProxyConfig, error) {
```
jakubno, line 128: Can we add length validation. Username/password are capped at 255 bytes by RFC 1929

packages/orchestrator/pkg/sandbox/network/firewall.go
```diff
@@ -246,6 +254,38 @@ func (fw *Firewall) installRules() error {
...
+func (fw *Firewall) EnableBYOPProxy() error {
+	if !fw.byopEnabled.CompareAndSwap(false, true) {
+		return nil
+	}
+	fw.conn.FlushChain(fw.filterChain)
+	if err := fw.installRules(true); err != nil {
+		fw.byopEnabled.Store(false)
+
+		return err
+	}
```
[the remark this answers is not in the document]
arkamar: `FlushChain` only buffers a DELRULE; `installRules` commits it together with the new rules in one atomic nftables batch, so a failed flush rolls the whole thing back and the default ruleset stays in place. Verified on a real kernel — the chain is never left empty.

packages/shared/pkg/sandbox-network/byop.go
arkamar: done

packages/shared/pkg/sandbox-network/byop.go
arkamar: yes

packages/shared/pkg/sandbox-network/byop_test.go
```diff
@@ -0,0 +1,250 @@
...
+		{"172.16.0.1", true},
+		{"192.168.5.5", true},
+		{"127.0.0.1", true},
+		{"169.254.169.254", true},
+		{"::1", true},
+		{"fe80::1", true},
+		{"fc00::1", true},
+		// Unspecified addresses are denied as BYOP endpoints even though they
+		// are not part of the kernel DeniedSandboxCIDRs nftables set.
+		{"0.0.0.0", true},
```
jakubno:
```suggestion
		{"0.0.0.0", true},
		{"0.1.2.3", true},
```

packages/orchestrator/pkg/sandbox/network/slot.go
```diff
@@ -287,22 +301,34 @@ func (s *Slot) UpdateInternet(ctx context.Context, egress *orchestrator.SandboxN
...
 	if err != nil {
 		return fmt.Errorf("failed to get slot network namespace '%s': %w", s.NamespaceID(), err)
 	}
 	defer n.Close()
 
+	// Set before mutating: a partial failure must still trigger cleanup.
+	s.firewallCustomRules.Store(true)
+
 	err = n.Do(func(_ ns.NetNS) error {
+		if hasBYOP {
```
jakubno: we're now doing two separate kernel commits per update:

1. `EnableBYOPProxy()` / `DisableBYOPProxy()` — flushes the filter chain and reinstalls it, committing via the `conn.Flush()` inside `installRules`
2. `ReplaceUserRules()` — replaces the user allow/deny sets and commits with its own `conn.Flush()`

Each batch is atomic on its own, but between them the kernel holds a hybrid state: the **new** rule-3 mode with the **previous** user sets. The enable direction is the risky one — rule 3 has already been narrowed to non-TCP (so TCP toward `predefinedDenySet` ranges is no longer dropped in-kernel) while the user sets still reflect the old config. 

This is especially problematic for `Update`, it should land as one transition. 

Suggestion: make `installRules` buffer-only (drop its internal `Flush`) and let callers commit, so `UpdateInternet` can stage mode switch + set replacement and flush once.

packages/orchestrator/pkg/sandbox/network/firewall.go
```diff
@@ -246,6 +254,38 @@ func (fw *Firewall) installRules() error {
 	return nil
 }
 
+// EnableBYOPProxy flushes the filter chain and reinstalls it with Rule 3
+// narrowed to non-TCP. Idempotent; rolls back the flag on failure.
+func (fw *Firewall) EnableBYOPProxy() error {
```
jakubno: similar problem here (it shouldn't happen as we hold lock for the sandbox in API, but imho this would make it safer) 

`fw.conn` is shared across goroutines with no lock

`google/nftables.Conn` buffers operations until `Flush()`, and nothing serializes access to it. Two concurrent `Update` RPCs for the same sandbox interleave their buffered ops on the same connection — e.g. A's `FlushChain` + BYOP rules and B's `FlushChain` + default rules can mix into each other's batches, leaving the chain with a rule set that corresponds to neither request, and `byopEnabled` out of sync with actual kernel state. The `CompareAndSwap` on `byopEnabled` gives idempotency, not mutual exclusion — it doesn't protect the netlink buffer.

This race pre-exists for `ReplaceUserRules`, but this PR widens what runs over the shared conn (chain flush + full reinstall, in two directions), and the consequences of a corrupted transition are worse now that one of the modes weakens the kernel drop.

jakubno (commented):
.

packages/orchestrator/pkg/sandbox/network/slot.go
arkamar: addressed

packages/orchestrator/pkg/sandbox/network/firewall.go
arkamar: addressed

packages/orchestrator/pkg/sandbox/network/firewall.go
```diff
@@ -273,9 +272,26 @@ func (fw *Firewall) ReplaceUserRules(allowedCIDRs, deniedCIDRs []string) error {
...
+// from scratch; on flush failure the kernel keeps the previous ruleset and no
+// in-memory state can desync from it.
+func (fw *Firewall) ApplyRules(byop bool, allowedCIDRs, deniedCIDRs []string) error {
+	fw.mu.Lock()
+	defer fw.mu.Unlock()
+
+	fw.conn.FlushChain(fw.filterChain)
+	fw.installRules(byop)
+	if err := fw.bufferUserRules(allowedCIDRs, deniedCIDRs); err != nil {
+		return err
```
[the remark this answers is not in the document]
arkamar: this is preexisting issue also present in constructor. I would propose to fixing it in a separate PR once this one lands.

jakubno (approved):

packages/api/internal/orchestrator/update_network.go
```diff
@@ -28,14 +29,20 @@ func (o *Orchestrator) UpdateSandboxNetworkConfig(
...
+	// PUT is full-replace: omitting egressProxy clears BYOP.
+	egressConfig := &types.SandboxNetworkEgressConfig{
+		AllowedAddresses: allowedEntries,
+		DeniedAddresses:  deniedEntries,
+		Rules:            rules,
 	}
+	if egressProxy != nil {
+		egressConfig.EgressProxyAddress = egressProxy.Address
+		egressConfig.EgressProxyUsername = egressProxy.Username
+		egressConfig.EgressProxyPassword = egressProxy.Password
```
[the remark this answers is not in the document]
jakubno: Pre-existing problem

PR #2655, opened 2026-05-14
fix(api): use uncancellable context for sandbox removal finish

jakubno:
Ensure finish is always executed

### DIFF
--- a/packages/api/internal/orchestrator/delete_instance.go
+++ b/packages/api/internal/orchestrator/delete_instance.go
@@ -77,7 +77,7 @@ func (o *Orchestrator) RemoveSandbox(ctx context.Context, teamID uuid.UUID, sand
 		}
 	}
 	defer func() {
-		finish(ctx, err)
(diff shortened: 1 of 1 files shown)


arkamar (approved):

packages/api/internal/orchestrator/delete_instance.go
```diff
@@ -77,7 +77,7 @@ func (o *Orchestrator) RemoveSandbox(ctx context.Context, teamID uuid.UUID, sand
 		}
 	}
 	defer func() {
-		finish(ctx, err)
+		finish(context.WithoutCancel(ctx), err)
```
sitole, line 80: Will finish fn implementation, make sure the call cannot hang forever?

Lately, we experienced a situation where the API was unable to delete the sandbox from the orchestrator. In that case, it should try long enough, but eventually still fail instead of overwhelming the API and the underlying backend (Redis/orchestrator gRPC) or similar.

PR #2665, opened 2026-05-15
test(storage): relax retry-delay upper bounds to deflake

ValentaTomas:
`TestRetryableClient_ActualRetryBehavior` asserts retry timing with very tight upper bounds (200ms / 300ms) on jittered 50ms / 100ms base delays. Loaded CI runners occasionally exceed those bounds — the test should assert "retries happened with non-zero spacing", not exact timing.

Widens both upper bounds to 2s. Lower bounds (must be > 0) preserved.


ValentaTomas:
Not sure if this is a proper fix.

PR #2668, opened 2026-05-15
[ENG-4062] perf(api): replace per-Release goroutine with shared publish worker 

jakubno:
This should help with creating unnecessary amount of goroutines + introduce the concept of publisher


packages/api/internal/sandbox/storage/redis/publisher.go
```diff
@@ -0,0 +1,280 @@
...
+	default:
+		p.drop(ctx, routingKey, dropReasonQueueFull, p.metrics.droppedQueueFull)
+	}
+}
+
+func (p *publisher) drop(ctx context.Context, routingKey, reason string, attrs metric.MeasurementOption) {
+	p.metrics.dropped.Add(ctx, 1, attrs)
+
+	n := p.dropped.Add(1)
+	if n%publishDropLogInterval == 1 {
```
arkamar:
```suggestion
	if n%publishDropLogFrequency == 1 {
```
nit: the Interval is a bit misleading here, I am proposing Frequency as it tells the reader that this is a occurrence rate based on a count, not a clock timer.

packages/api/internal/sandbox/storage/redis/publisher.go
```diff
@@ -0,0 +1,280 @@
...
+const (
+	publishQueueDepth = 8192
+
+	// publishTimeout bounds a single PUBLISH round-trip
+	publishTimeout        = 5 * time.Second
+	publishShutdownBudget = 5 * time.Second
+
+	// publishDropLogInterval rate-limits drop warnings: only every Nth
+	// drop produces a log line, with the running total attached.
+	publishDropLogInterval = 64
```
arkamar:
```suggestion
	// publishDropLogFrequency rate-limits drop warnings: only every Nth
	// drop produces a log line, with the running total attached.
	publishDropLogFrequency = 64
```

packages/api/internal/sandbox/storage/redis/publisher.go
```diff
@@ -0,0 +1,280 @@
...
+	"github.com/e2b-dev/infra/packages/shared/pkg/logger"
+	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
+)
+
+const (
+	publishQueueDepth = 8192
+
+	// publishTimeout bounds a single PUBLISH round-trip
+	publishTimeout        = 5 * time.Second
+	publishShutdownBudget = 5 * time.Second
```
arkamar: It looks a bit weird to have those two set to the same value as it implies that the `drainOnShutdown` can process at most one hung Redis call before budget expires. It is worth commenting, at least.

packages/api/internal/sandbox/storage/redis/publisher_test.go
```diff
@@ -0,0 +1,436 @@
...
+		lock, err := storage.locker.Obtain(t.Context(), key, testLockTimeout)
+		require.NoError(t, err)
+		require.NoError(t, lock.Release(context.WithoutCancel(t.Context())))
+	}
+
+	// Allow any in-flight publishes to drain through the worker.
+	require.Eventually(t, func() bool {
+		runtime.GC() //nolint:revive // test-only stability shim
+
+		return runtime.NumGoroutine()-before < 10
```
arkamar: I am bit worried this test will be flaky due to Parallelization.

arkamar (changes requested):
I have only few nits and notes, but otherwise it looks good.

packages/api/internal/sandbox/storage/redis/publisher_test.go
jakubno: Removed the test, it was little bit unnecessary

packages/api/internal/sandbox/storage/redis/publisher.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+
+	select {
+	case p.queue <- routingKey:
+		p.mu.RUnlock()
+	default:
+		p.mu.RUnlock()
+		p.drop(ctx, routingKey, dropReasonQueueFull, p.metrics.droppedQueueFull)
+	}
+}
```
arkamar, line 163:
```suggestion
func (p *publisher) Publish(ctx context.Context, routingKey string) {
	p.mu.RLock()
	defer p.mu.RUnlock()

	if p.closing {
		p.drop(ctx, routingKey, dropReasonClosed, p.metrics.droppedClosed)

		return
	}

	select {
	case p.queue <- routingKey:
	default:
		p.drop(ctx, routingKey, dropReasonQueueFull, p.metrics.droppedQueueFull)
	}
}
```
Is the excluded `p.drop()` optimization meaningful here? `drop()` is just an atomic increment and an occasional logger call. The suggested change is better readable IMO and less error prone with future changes and it practically affects only orchestrator's close path.

arkamar (changes requested):

packages/api/internal/sandbox/storage/redis/publisher.go
jakubno: I though I have changed this, It was probably only in terminal 😅 🙈

arkamar (approved):

PR #2673, opened 2026-05-15
Add OIDC auth provider for API and dashboard API

tvi:
Replace the Supabase-specific HMAC JWT validation in both API and dashboard API with a generic auth-provider verification path that supports OIDC issuers and resolves the internal user UUID via a new `public.user_identities` table keyed on `(oidc_iss, oidc_sub)`. A legacy HMAC strategy is kept as a compatibility shim during the migration.

- New `packages/auth/pkg/auth/oidc` package: per-issuer `Verifier` built from an OIDC discovery document (issuer cross-check, JWKS auto-discovery over HTTPS, refresh via `MicahParks/keyfunc/v3`+`jwkset`). Audience validation supports `MatchAny` (`MatchAll` is rejected by the validator today, kept as a placeholder for future use). `Verify` extracts `iss`/`sub` from the validated claims and resolves the internal UUID via an `IdentityLookup`; the package itself stays db-free.
- New `packages/auth/pkg/auth/legacy` package: HMAC-only `Verifier` for the existing Supabase JWT secret (with the previous 16-byte minimum-length check). Used as one strategy in the aggregating verifier.
- New aggregating `auth.Verifier` (`packages/auth/pkg/auth/verifier.go`) iterates configured strategies and returns the first that succeeds. Old `packages/auth/pkg/auth/jwt.go` (Supabase-specific parser) is deleted.
- `AuthService` no longer holds raw JWT secrets. `ValidateSupabaseToken` is replaced by `ValidateAuthProviderToken` which calls the verifier and rejects tokens that resolve to `uuid.Nil`.
- New migration `20260515120000_create_user_identities_table.sql` introducing `public.user_identities (oidc_iss text, oidc_sub text, user_id uuid)` with composite PK `(oidc_iss, oidc_sub)`, FK to `public.users` with `ON DELETE CASCADE`, and RLS enabled. Two new sqlc queries (`GetUserIdentity`, `UpsertPublicIdentity`).
- `NewAuthIdentityLookup` adapter in the `auth` package wires the read-replica `*authqueries.Queries` into `oidc.IdentityLookup`, mapping `pgx.ErrNoRows` to `oidc.ErrIdentityNotFound`.
- Configuration: API/dashboard API replace `SUPABASE_JWT_SECRETS` with a single `AUTH_PROVIDER_CONFIG` JSON env var (parsed via custom `env` parser). Terraform variables on AWS/GCP Nomad providers updated accordingly, with the existing Supabase secret carried as the default `legacy.hmac.secrets` value.
- OpenAPI: add `AuthProviderTokenAuth` (Bearer) and `AuthProviderTeamAuth` (`X-Team-Id`) security schemes alongside the existing Supabase schemes; regenerate `api.gen.go` in both services. Gin CORS headers and middleware registrations extended to accept the new headers.

Example `AUTH_PROVIDER_CONFIG`:

```json
{
  "jwt": [
    {
      "issuer": {
        "url": "https://issuer.example.com",
        "discoveryURL": "https://issuer.example.com/.well-known/openid-configuration",
        "audiences": ["dashboard-api"],
        "audienceMatchPolicy": "MatchAny"
      },
      "cacheDuration": "5m"
    }
  ],
  "legacy": {
    "hmac": { "secrets": ["legacy-supabase-secret"] }
  }
}
```


spec/openapi-dashboard.yml
```diff
@@ -18,6 +18,14 @@ components:
       type: apiKey
       in: header
       name: X-Supabase-Team
+    AuthProviderTokenAuth:
```
dobrac: The name of Token needs to be alphabetically before team - the parser then persists the order and will auth you before getting the team.

This is the reason why the Supabase1 vs Supabase2 are there with the numbers

spec/openapi.yml
```diff
@@ -27,6 +27,14 @@ components:
       type: apiKey
       in: header
       name: X-Supabase-Team
+    AuthProviderTokenAuth:
```
dobrac: same here with the numbering

packages/auth/pkg/auth/oidc/oidc.go
```diff
@@ -0,0 +1,283 @@
...
+
+	"github.com/MicahParks/jwkset"
+	"github.com/MicahParks/keyfunc/v3"
+	"github.com/golang-jwt/jwt/v5"
+	"github.com/google/uuid"
+)
+
+// oidcHTTPTimout is the timeout used for OIDC discovery and JWKS HTTP
+// requests.
+const oidcHTTPTimout = 10 * time.Second
```
dobrac:
```suggestion
const oidcHTTPTimeout = 10 * time.Second
```

packages/auth/pkg/auth/oidc/oidc.go
```diff
@@ -0,0 +1,283 @@
...
+	if !ok {
+		return uuid.Nil, nil, errors.New("auth provider token is missing iss claim")
+	}
+
+	sub, ok := claimString(claims, "sub")
+	if !ok {
+		return uuid.Nil, nil, errors.New("auth provider token is missing sub claim")
+	}
+
+	userID, err := v.identities.GetUserIdentity(ctx, iss, sub)
```
dobrac, line 143: this will incurr a database request on each Verify call - we should probably cache these locally or in Redis

packages/auth/pkg/auth/verifier.go
```diff
@@ -0,0 +1,138 @@
...
+type Verifier struct {
+	strategies []strategy
+}
+
+func NewVerifier(ctx context.Context, config ProviderConfig, oidcHTTPClient *http.Client, identities oidc.IdentityLookup) (*Verifier, error) {
+	normalized := config.normalize()
+	if err := normalized.validate(); err != nil {
+		return nil, err
+	}
+	if !normalized.Enabled() {
```
dobrac, line 79: should we first check enabled before validation?

dobrac (changes requested):

packages/auth/pkg/auth/verifier.go
tvi: i think its correct this way, first you fill in defaults then you check

packages/auth/pkg/auth/oidc/oidc.go
```diff
@@ -0,0 +1,283 @@
...
+	if err := validateHTTPSURL(doc.JWKSURI, "discovery jwks_uri"); err != nil {
+		return nil, err
+	}
+
+	storage, err := jwkset.NewStorageFromHTTP(doc.JWKSURI, jwkset.HTTPClientStorageOptions{
+		Client:          httpClient,
+		Ctx:             ctx,
+		HTTPTimeout:     oidcHTTPTimout,
+		RefreshInterval: entry.CacheDuration,
+	})
```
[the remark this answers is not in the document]
tvi: good point, but doesnt apply here ctx is long live context

packages/auth/pkg/auth/legacy/legacy.go
```diff
@@ -0,0 +1,108 @@
...
+
+		var userID uuid.UUID
+		if claimValue, ok := claimString(claims, "sub"); ok {
+			if parsed, parseErr := uuid.Parse(claimValue); parseErr == nil {
+				userID = parsed
+			}
+		}
+
+		return userID, claims, nil
+	}
```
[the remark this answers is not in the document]
tvi: fix

packages/api/internal/handlers/store.go
```diff
@@ -200,7 +200,13 @@ func NewAPIStore(ctx context.Context, tel *telemetry.Client, redisClient redis.U
 
 	authCache := sharedauth.NewAuthCache[*types.Team](redisClient)
 	authStore := sharedauth.NewAuthStore(authDB)
-	authService := sharedauth.NewAuthService[*types.Team](authStore, authCache, config.SupabaseJWTSecrets)
+	authHTTPClient := &http.Client{}
```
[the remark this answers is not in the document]
tvi: working as intended

packages/auth/pkg/auth/oidc/oidc.go
```diff
@@ -0,0 +1,283 @@
...
+		if len(typed) == 0 {
+			return "", false
+		}
+		first, ok := typed[0].(string)
+
+		return first, ok && first != ""
+	default:
+		return "", false
+	}
+}
```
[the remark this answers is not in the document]
tvi: deliberate design

packages/auth/pkg/auth/verifier.go
```diff
@@ -0,0 +1,138 @@
...
+		return nil, errors.New("auth provider not enabled")
```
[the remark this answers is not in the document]
tvi: deliberate design Supabase secrets are required at this time

dobrac (approved):

dobrac:
Follow ups:
- The security schema names should be more explicit about the alphabetical order requirement
- There should be a support for using API without AUTH_PROVIDER_CONFIG

PR #2674, opened 2026-05-16
perf(sandbox): reduce memory dirtying from journald and envd logging

ValentaTomas:
Reduces guest memory and rootfs dirtying caused by journald and envd logging during pause-resume cycles.

envd no longer writes to its stdout by default. A new `-verbose` flag opts in for local development; in FC production envd's stdout stays empty so it doesn't dirty journald pages. The HTTP exporter still ships the full debug stream to the orchestrator. Only envd's stderr (Go runtime panics, `log.Fatalf`, bootstrap and MMDS-poll warnings) reaches journald — which is where we actually want envd's lifecycle/fatal output to land.

The MMDS poll (50ms ticker) and the HTTP exporter's per-line "send failed" paths used to write to stderr per iteration / per log line, which 1:1-amplified into journald during outages. Both now log only the first failure of each kind. The exporter additionally caps its in-memory queue at 10k entries with drop-oldest semantics so a wedged collector can't grow it without bound. The remaining stderr writes from those two paths are tagged with the syslog `<4>` (WARNING) prefix so journald classifies them correctly instead of as errors.

The `handler.go` "error reading from pty/stdout/stderr" messages were on stderr but are about envd-handled user processes, not envd internals; they now flow through the zerolog logger.

A journald drop-in caps the remaining (other systemd services) journal at `Storage=persistent` / `SystemMaxUse=8M` / `MaxLevelStore=warning` so the journal can't grow without bound and the writes land on the rootfs (small, ext4-`inline_data`-friendly) rather than dirtying RAM via the default tmpfs path.

### DIFF
--- a/packages/envd/internal/host/mmds.go
+++ b/packages/envd/internal/host/mmds.go
@@ -144,20 +144,21 @@ func PollForMMDSOpts(ctx context.Context, mmdsChan chan<- *MMDSOpts, envVars *ut
 
 	// Polling runs at 50ms — log the first failure of each kind only so a
 	// persistently broken MMDS endpoint doesn't spam journald until ctx
-	// cancellation.
+	// cancellation. Tagged as syslog WARNING (<4>) since MMDS hiccups are
+	// expected during early boot and recoverable.
 	var loggedTokenErr, loggedOptsErr bool
 
 	for {
 		select {
 		case <-ctx.Done():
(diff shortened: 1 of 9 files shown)


ValentaTomas:
Closing in favor of reopening the original #2423 on the same branch.

ValentaTomas:
Closing in favor of the split into #2675 (keep envd logging out of journald) and #2676 (bound and back off the logs exporter and MMDS poll).

PR #2675, opened 2026-05-16
perf(sandbox): keep envd logging out of journald

ValentaTomas:
envd's zerolog stdout writer is now gated behind a new \`-verbose\` flag (default off), so production envd inside FC no longer writes anything to stdout — journald stays clean of per-request debug events. The HTTP exporter still ships full debug to the orchestrator regardless.

A journald drop-in caps the rest of the in-VM journal at \`Storage=persistent\` / \`SystemMaxUse=8M\` / \`MaxLevelStore=warning\` so other systemd services can't grow it without bound either.

\`handler.go\`'s "error reading from pty/stdout/stderr" messages were on raw stderr but are about envd-handled user processes, not envd internals; they now flow through the zerolog logger.

Split out of #2674 (journal-side half).

### DIFF
--- a/packages/envd/Makefile
+++ b/packages/envd/Makefile
@@ -46,7 +46,7 @@ start-docker:
 	-p 8001:8001 \
 	--rm \
 	-i envd-debug \
-	/usr/bin/envd -isnotfc
+	/usr/bin/envd -isnotfc -verbose
 
 build-and-upload:
 	make build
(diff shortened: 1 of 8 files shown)


ValentaTomas:
Fixed in latest push: stdout is now added when `!isFC || verbose`, so `-isnotfc` keeps logging to stdout as before.

dobrac (approved):

PR #2676, opened 2026-05-16
fix(envd): bound the in-memory logs queue

ValentaTomas:
Adds a byte-bounded log queue so the exporter can't grow without bound when the collector is unreachable.

- 8 MiB total queue cap with drop-oldest whole entries (JSON envelopes stay intact); 192 KiB per-line ingestion cap under [Loki's 256 KiB \`max_line_size\`](https://grafana.com/docs/loki/latest/configure/#limits_config) default.
- MMDS opts swapped via \`atomic.Pointer\` instead of pointer + RWMutex.
- HTTP keepalives disabled on the exporter and the MMDS poll client so a pause/resume that changes endpoints doesn't reuse half-dead connections.

ID injection and the send loop are unchanged; wire-compatible with the current orchestrator handler — no deploy-order dependency on #2679.

### DIFF
--- a/packages/envd/internal/logs/exporter/exporter.go
+++ b/packages/envd/internal/logs/exporter/exporter.go
@@ -6,37 +6,39 @@ import (
 	"log"
 	"net/http"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/e2b-dev/infra/packages/envd/internal/host"
 )
 
-const ExporterTimeout = 10 * time.Second
+const (
+	ExporterTimeout = 10 * time.Second
+
+	// Under Loki's 256 KiB max_line_size default.
+	maxLogLineBytes  = 192 << 10
+	maxBufferedBytes = 8 << 20
+)
 
 type HTTPExporter struct {
-	client   http.Client
-	logs     [][]byte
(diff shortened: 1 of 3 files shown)


ValentaTomas:
Addressed in latest push: `LogsCollectorAddress` is now captured under `mmdsLock.RLock()` and used as a local; the error format no longer dereferences `w.mmdsOpts` outside the lock.

packages/envd/internal/logs/exporter/exporter.go
```diff
@@ -87,26 +87,39 @@ func (w *HTTPExporter) listenForMMDSOptsAndStart(ctx context.Context, mmdsChan <
 }
 
 func (w *HTTPExporter) start(ctx context.Context) {
+	// Cap stderr noise: log each error kind at most once per logFloor so a
+	// fast-failing collector (e.g. TCP RST) can't flood journald.
+	const logFloor = time.Minute
+	var lastLoggedJSONErr, lastLoggedSendErr time.Time
```
djeebus, line 93: Might be good to add a "398 error ignored since last log" or something similar

djeebus (approved):

PR #2682, opened 2026-05-17
fix(envd): set OOMScoreAdjust to -999 so kernel can reclaim a leaky envd

ValentaTomas:
\`OOMScoreAdjust=-1000\` makes envd fully immune to the OOM killer. If envd ever leaks past \`memory.max\` for its cgroup (historical snapshots can still carry the pre-0.5.16 fan-out leak), the kernel kills *some other process* in the cgroup and envd keeps growing.

\`-999\` is still extremely low priority but not fully immune. Combined with the existing \`Restart=always\` + \`StartLimitIntervalSec=0\`, a leaky envd self-heals into a fresh instance instead of needing a manual \`pkill envd ; sleep 5\` recipe (Petr Vaněk's Apr-30 finding).

## Test plan
- [ ] On a freshly built sandbox, \`cat /proc/\$(pidof envd)/oom_score_adj\` returns \`-999\`.
- [ ] Inject memory pressure inside the sandbox until the cgroup hits memory.max; verify the kernel eventually kills envd (instead of another process) and systemd restarts it.


ValentaTomas:
Per discussion: skipping — unlikely to help in practice.

PR #2686, opened 2026-05-17
fix(orch): preserve Prefetch in template metadata transition helpers

ValentaTomas:
Preserve `t.Prefetch` in `SameVersionTemplate`/`NewVersionTemplate`/`BasedOn`. Pause was the only caller without a `WithPrefetch` re-attachment, so paused snapshots silently lost the build-time prefetch list.

### DIFF
--- a/packages/orchestrator/pkg/template/metadata/template_metadata.go
+++ b/packages/orchestrator/pkg/template/metadata/template_metadata.go
@@ -129,6 +129,7 @@ func (t Template) BasedOn(
 		Start:        t.Start,
 		FromTemplate: &ft,
 		FromImage:    nil,
+		Prefetch:     t.Prefetch,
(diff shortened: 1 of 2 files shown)


ValentaTomas:
Closing — the Pause-vs-Checkpoint asymmetry is intentional. Prefetch data reflects an observed runtime trajectory of a specific resume, not a property of the template. On Pause the only prefetch we could carry over is the build-time optimize-phase data, which describes a fresh-from-template boot, not the just-paused state. Attaching it would be bandwidth wasted + no coverage for the actually-hot pages of the current state. Checkpoint avoids this by locally resuming and observing; Pause has no symmetric way to do that cheaply. Leaving prefetch absent on Pause snapshots is the honest behavior.

PR #2694, opened 2026-05-17
feat(memfd): async pause/copy (in-flight reads)

ValentaTomas:
Alternative to #2626 with the same async pause but in-flight reads — no
Wait on the read path.

ReadAt/Slice demand-page the requested range from memfd into the cache
file if runCopy hasn't reached it yet, then return the cache mmap slice.
runCopy skips already-cached ranges. The memfd is closed at end of copy;
returned slices stay valid (they point at the cache file mmap, not the
memfd).

Only direct readers of the on-disk file (CachePath, e.g. snapshot
upload) still Wait. Lets a follow-up sandbox resume from a just-paused
snapshot without blocking on the upload.

Gated by MemfdBackgroundCopyFlag; requires UseMemFdFlag (#2522).

### DIFF
--- a/packages/orchestrator/pkg/sandbox/block/device.go
+++ b/packages/orchestrator/pkg/sandbox/block/device.go
@@ -41,6 +41,19 @@ type ReadonlyDevice interface {
 	SwapHeader(h *header.Header)
 }
 
+// DiffSource is the set of methods build.localDiff needs from its backing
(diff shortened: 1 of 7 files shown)


ValentaTomas:
This solution would be ideal for the smallest latency of resumes, but we will be first going with the wait-on-read https://github.com/e2b-dev/infra/pull/2626.

We might revisit in the future.

PR #2698, opened 2026-05-17
feat(rootfs): blanket IPv4 PREROUTING DNAT to 127.0.0.1

ValentaTomas:
Replaces the per-port socat path for IPv4 listeners with one provision-time iptables rule + sysctl:

\`\`\`
net.ipv4.conf.all.route_localnet=1
iptables -t nat -A PREROUTING -i eth0 -p tcp -j DNAT --to-destination 127.0.0.1
\`\`\`

Applied at boot by \`e2b-ipv4-forward.service\`. envd's scanner still tracks IPv6 listeners and keeps socat for those (no \`route_localnet\` for v6). Idempotent: \`iptables -C\` check before \`-A\`.

Supersedes #2697.


ValentaTomas:
Closing this for now as it seems to not be related to the current investigation and this is a bigger rework.

PR #2700, opened 2026-05-17
feat(rootfs): isolate envd in a dedicated network namespace

ValentaTomas:
Moves \`eth0\` + envd into a new \`envd-ns\` so customer iptables changes in the default namespace can no longer break envd's MMDS / \`/init\` path (cf. the recent customer who DNAT'd :80 → :4200 and broke MMDS token fetch).

## Topology

\`\`\`
default ns (customer)            envd-ns
  veth-customer 192.168.250.1     veth-envd  192.168.250.2
                                  eth0       169.254.0.21 (moved in)
\`\`\`

- envd-ns runs as a router: \`POSTROUTING MASQUERADE\` on eth0 for customer outbound; \`PREROUTING DNAT\` to \`192.168.250.1\` for inbound, with a \`RETURN\` exception for port 49983 (envd) so \`/init\` reaches envd directly.
- Customer ns picks up forwarded packets on \`veth-customer\` and the existing #2698 PREROUTING DNAT forwards them to \`127.0.0.1\`.
- envd binds inside envd-ns via systemd \`NetworkNamespacePath=/run/netns/envd-ns\`.

## Notes

- \`e2b-netns.service\` is oneshot, gated \`Before=network-pre.target sysinit.target\`, idempotent.
- Customer's view of the network changes: their \"eth0\" is now \`veth-customer\` (\`192.168.250.1\`) and gateway is \`192.168.250.2\`. Apps binding \`0.0.0.0\` or \`127.0.0.1\` continue to work; apps explicitly binding \`169.254.0.21\` would not (rare).
- Existing per-port socat / port forwarder code is unchanged for this PR; it remains the bridge from \`veth-customer:port\` to \`127.0.0.1:port\` in the customer ns. Combined with #2698 the path works end-to-end.

## Test plan
- [ ] \`ip netns list\` inside the guest shows \`envd-ns\`; \`ip -n envd-ns addr\` shows \`eth0 169.254.0.21\`.
- [ ] Orchestrator can reach \`169.254.0.21:49983\` (envd \`/init\`).
- [ ] Customer process binding \`127.0.0.1:8080\` is reachable from the orchestrator via the standard port-forward path.
- [ ] Customer running \`iptables -t nat -A OUTPUT -d 169.254.169.254 -p tcp --dport 80 -j REDIRECT --to-ports 4200\` no longer affects envd's MMDS fetch.
- [ ] Customer outbound to the public internet via the egress proxy still works.


ValentaTomas:
Refactored per @arkamar's suggestion (91fea54a):
- Veth pair declared in `10-e2b-veth.netdev`.
- Customer-side IP + gateway in `10-e2b-veth-customer.network`.
- `e2b-netns.service` is now a chain of `ExecStart=-` lines for the irreducibly imperative bits (`ip netns add`, `ip link set ... netns`, configuring eth0 + veth-envd inside envd-ns); `-` prefix gives idempotency on re-runs.
- iptables rules still need `-C` check before `-A` to be safely re-runnable across boots, wrapped in `sh -c` (no clean systemd primitive for "insert if missing").

Trade-offs vs. fully-declarative-everywhere:
- `ip netns add` and `ip link set <if> netns <ns>` don't have native systemd directives.
- `systemd-networkd` doesn't run inside `envd-ns`, so the envd-side interface config is in the unit's ExecStart chain.

LMK if you'd like to push further (e.g. install `systemd-networkd` inside envd-ns too) or if this strikes the right balance.

ValentaTomas:
Closing in favor of the dirty self-heal pin (#2701). The dedicated network namespace is still the right structural fix and we'll likely reopen / rebase this once the immediate Bitfrost mitigation lands and we have time to coordinate the systemd-networkd + iptables-in-customer-ns story properly.

PR #2703, opened 2026-05-17
fix(envd): bound /init request body to 1 MiB

ValentaTomas:
Wraps `r.Body` with `http.MaxBytesReader`. Real /init payloads (env vars, CA bundle, mounts, token) are well under 1 MiB; a slow or oversized peer can't hold `initLock` + an unbounded in-memory buffer.


ValentaTomas:
No need to do this, as this is user-passed and therefore mostly desired.

PR #2704, opened 2026-05-17
perf(envd): set ReadHeaderTimeout on HTTP server

ValentaTomas:
Sets `ReadHeaderTimeout: 10s` on the envd http.Server. Read/WriteTimeout intentionally stay 0 (long-lived requests must not be cut), but ReadHeaderTimeout is independent and bounds slow-loris-style attacks against the request header phase only.

### DIFF
--- a/packages/envd/main.go
+++ b/packages/envd/main.go
@@ -197,10 +197,16 @@ func main() {
 			),
 		),
 		Addr: fmt.Sprintf("0.0.0.0:%d", port),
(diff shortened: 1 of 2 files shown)


ValentaTomas:
Closing — ReadHeaderTimeout only bounds the request header phase. Once headers are in, body read and handler execution remain unbounded, so it wouldn't help any of the /init-stuck scenarios.

PR #2705, opened 2026-05-17
fix(envd): bound NFS mount kernel-side retries

ValentaTomas:
Adds `timeo=20,retrans=2` to the NFS mount options so that an unreachable proxy makes the in-kernel mount give up after ~6s instead of sitting in D-state past the userspace 10s timeout.

### DIFF
--- a/packages/envd/internal/api/init.go
+++ b/packages/envd/internal/api/init.go
@@ -243,14 +243,12 @@ var nfsOptions = strings.Join([]string{
 	"nfsvers=3",      // nfs proxy is nfs version 3
 	"noacl",          // no reason for acl in the sandbox
 
-	// Bound the mount against an unreachable proxy to ~10s total:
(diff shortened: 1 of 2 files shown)


ValentaTomas:
Closing — D-state pile-up was overstated. Real impact is only shrinking 75s → 6s in the narrow TCP-connected-but-RPC-hung failure shape; doesn't help the TCP-DROP shape (kernel TCP connect timeout dominates) or the ECONNREFUSED shape (mount fails fast anyway). Not worth the option-tuning maintenance for an unobserved failure mode.

ValentaTomas:
Closing — real-world impact is narrow (only the 'TCP connected but NFS RPC hangs' shape, ~6s instead of ~75s). The orchestrator+initLock+errgroup combo already bounds D-state mount processes per sandbox to (#volumes × 1 in-flight), not unbounded. No reported incident from this shape; revive if we see one.

PR #2706, opened 2026-05-18
fix(envd): replace env vars in /init instead of merging

ValentaTomas:
Replaces the env-var dict on each /init that carries a non-nil `EnvVars` payload, so removing a variable from sandbox config takes effect on the next /init instead of persisting forever.

### DIFF
--- a/packages/envd/internal/utils/envvars.go
+++ b/packages/envd/internal/utils/envvars.go
@@ -1,0 +1,77 @@
+package utils
+
+import "sync"
+
+// EnvVars distinguishes system-set ("internal") entries from user-provided
+// ones; internal entries cannot be touched by the user-facing methods.
+type EnvVars struct {
+	m sync.Map
+}
+
+type envVarEntry struct {
+	value    string
+	internal bool
+}
+
+func NewEnvVars() *EnvVars {
+	return &EnvVars{}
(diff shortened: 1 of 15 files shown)


packages/envd/internal/utils/envvars.go
jakubno, line 1: why don't we just use map + mutex. It would simplify the code and for our use case it'll work perfectly fine

ValentaTomas: We have other env vars like E2B_SANDBOX that we dynamically add and we don't want to be overwritten or removed when we clean the env vars when we pass user env vars on init. I didn't want to do an explicit list of these internal env vars as I'm pretty sure we will forget about adding a new one to another place.

ValentaTomas: The alternative solution is keeping two maps, but at that point it seems easier to just store additional info in one map.

jakubno: I meant map of envVarEntry

ValentaTomas: Switched to a regular map + sync.RWMutex in 7643fa529.

jakubno (approved):

PR #2708, opened 2026-05-18
fix(api): keep API alive until in-flight requests finish

jakubno:
When the API received SIGTERM, in-flight HTTP requests (sandbox pause, snapshot, delete) could be cut off mid-flight, leaving sandboxes in inconsistent state on the orchestrator node.

Root cause: each serve goroutine had `defer cancel()` on the process root context. http.Server.Shutdown stops the listener — which makes ListenAndServe return — and the serve goroutine's deferred cancel fired during the drain. Shutdown's context was a child of the root, so it self-aborted; in-flight r.Context() (BaseContext = root) died with it.

Fixes:
  - Serve goroutines now `defer serveErrCancel()` on a dedicated serveErrCtx (child of ctx) so a serve exit wakes the shutdown watcher without cancelling the root.
  - Shutdown watcher selects on signalCtx OR serveErrCtx and drains HTTP via a bounded child of the still-live root ctx.
  - Bound grpc.Server.GracefulStop via GracefulStopWithTimeout with a Stop() fallback; run both gRPC servers in parallel via sync.WaitGroup.Go.
  - Bump Nomad kill_timeout 30s -> 150s to fit the drain budget (15s LB drain + 75s HTTP + 10s gRPC + cleanup slack).

### DIFF
--- a/iac/modules/job-api/jobs/api.hcl
+++ b/iac/modules/job-api/jobs/api.hcl
@@ -139,9 +139,9 @@ job "api" {
 
     task "start" {
       driver       = "docker"
-      # If we need more than 30s we will need to update the max_kill_timeout in nomad
+      # Budget = shutdownDrainWait (15s) + httpShutdownTimeout (requestTimeout 70s) + grpcShutdownTimeout (10s) + cleanup (30s) + slack.
       # https://developer.hashicorp.com/nomad/docs/configuration/client#max_kill_timeout
-      kill_timeout = "30s"
+      kill_timeout = "150s"
       kill_signal  = "SIGTERM"
 
       resources {
--- a/iac/modules/job-api/jobs/api.hcl
+++ b/iac/modules/job-api/jobs/api.hcl
@@ -139,7 +139,7 @@ job "api" {
 
     task "start" {
       driver       = "docker"
-      # Budget = shutdownDrainWait (15s) + httpShutdownTimeout (requestTimeout 70s + 5s) + grpcShutdownTimeout (10s) + cleanup (30s) + slack.
+      # Budget = shutdownDrainWait (15s) + shutdownTimeout (requestTimeout 70s + 5s) + cleanup (30s) + slack.
       # https://developer.hashicorp.com/nomad/docs/configuration/client#max_kill_timeout
       kill_timeout = "150s"
       kill_signal  = "SIGTERM"
--- a/iac/modules/job-api/jobs/api.hcl
+++ b/iac/modules/job-api/jobs/api.hcl
@@ -139,7 +139,7 @@ job "api" {
 
     task "start" {
       driver       = "docker"
-      # Budget = shutdownDrainWait (15s) + httpShutdownTimeout (requestTimeout 70s) + grpcShutdownTimeout (10s) + cleanup (30s) + slack.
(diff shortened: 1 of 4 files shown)


packages/shared/pkg/grpc/shutdown.go
```diff
@@ -0,0 +1,36 @@
...
+
+	go func() {
+		srv.GracefulStop()
+		close(done)
+	}()
+
+	select {
+	case <-done:
+		return true
+	case <-time.After(d):
```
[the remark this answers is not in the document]
jakubno: Read the docs:
```
// After waits for the duration to elapse and then sends the current time
// on the returned channel.
// It is equivalent to [NewTimer](d).C.
//
// Before Go 1.23, this documentation warned that the underlying
// [Timer] would not be recovered by the garbage collector until the
// timer fired, and that if efficiency was a concern, code should use
// NewTimer instead and call [Timer.Stop] if the timer is no longer needed.
// As of Go 1.23, the garbage collector can recover unreferenced,
// unstopped timers. There is no reason to prefer NewTimer when After will do.
```

iac/modules/job-api/jobs/api.hcl
```diff
@@ -139,9 +139,9 @@ job "api" {
 
     task "start" {
       driver       = "docker"
-      # If we need more than 30s we will need to update the max_kill_timeout in nomad
+      # Budget = shutdownDrainWait (15s) + httpShutdownTimeout (requestTimeout 70s) + grpcShutdownTimeout (10s) + cleanup (30s) + slack.
```
jakubno:
```suggestion
      # Budget = shutdownDrainWait (15s) + httpShutdownTimeout (requestTimeout 70s + 5s) + grpcShutdownTimeout (10s) + cleanup (30s) + slack.
```

packages/api/main.go
```diff
@@ -521,27 +556,39 @@ func run() int {
...
+				l.Warn(ctx, "internal gRPC forced stop after graceful timeout", zap.Duration("budget", grpcShutdownTimeout))
+			}
+		})
+		grpcWG.Go(func() {
+			if !e2bgrpc.GracefulStopWithTimeout(edgeGrpcServer, grpcShutdownTimeout) {
+				l.Warn(ctx, "edge gRPC forced stop after graceful timeout", zap.Duration("budget", grpcShutdownTimeout))
+			}
+		})
+		grpcWG.Wait()
+
```
sitole: Let's drain all 3 servers at same time

sitole (changes requested):
nit

iac/modules/job-api/jobs/api.hcl
```diff
@@ -139,9 +139,9 @@ job "api" {
 
     task "start" {
       driver       = "docker"
-      # If we need more than 30s we will need to update the max_kill_timeout in nomad
+      # Budget = shutdownDrainWait (15s) + httpShutdownTimeout (requestTimeout 70s + 5s) + grpcShutdownTimeout (10s) + cleanup (30s) + slack.
```
jakubno:
```suggestion
      # Budget = shutdownDrainWait (15s) + shutdownTimeout (requestTimeout 70s + 5s) + cleanup (30s) + slack.
```

sitole (approved):

Issue #2710, opened 2026-05-18
failed to reset memory.peak


AdaAibaby:
a lot of <img width="3401" height="1827" alt="Image" src="https://github.com/user-attachments/assets/8d485cd9-a436-4a30-bb13-c5b4f9de5fb2" /> 

failed to reset memory.peak, interval peak semantics degraded  {"service": "orchestrator_template-manager", "internal": true, "pid": 3692383, "trace_id": "f0a5169be0edd839c5775c51193195fb", "span_id": "977ce3dd66bdafb2", "error": "write /sys/fs/cgroup/e2b/sbx-il8ne0g3geuvdtzy7qdsr-8mnmtty6e3guyqvsfjwq/memory.peak: invalid argument"}

arkamar:
This happens with kernels older than 6.12, where the memory.peak reset feature was introduced.

AdaAibaby:
> This happens with kernels older than 6.12, where the memory.peak reset feature was introduced.

You're absolutely correct regarding the kernel version limitation. We have identified two root causes for this invalid argument error:
The file-descriptor-scoped memory.peak reset functionality is fully stabilized and officially available starting from Linux 6.12 (ref: https://kernelnewbies.org/Linux_6.12). On kernels 6.1 ~ 6.11, memory.peak remains read-only, and any write operation will fail with EINVAL.
Our former implementation wrote non-empty content to reset the peak value, which does not conform to kernel specs and would also fail even on compatible kernels.
This PR https://github.com/e2b-dev/infra/pull/2724 resolves both issues thoroughly:
Implements precise kernel version parsing and detection, enabling memory peak reset only on kernels >= 6.12, and silently skipping reset logic on older kernels to eliminate redundant error logs.
Fixes the write logic to use standard zero-length writes that strictly match upstream kernel semantics.
Supplements comprehensive unit tests and integration tests covering various kernel version naming formats and edge cases, guaranteeing consistent behavior across different environments.
Feel free to take a review, all feedback and suggestions are highly welcome!

AdaAibaby:
Upgrade Kernel Version

PR #2724, opened 2026-05-19
fix(cgroup): suppress EINVAL noise on memory.peak reset for kernel < 6.12

AdaAibaby:
Friendly ping for maintainer review. @jakubno
## Problem
https://github.com/e2b-dev/infra/issues/2710
On kernels < 6.12 (e.g. `6.8.0-55-generic`, `6.8.0-110-generic`), the orchestrator emits a WARN log every 5 seconds per running sandbox:

**Root cause — two compounding bugs:**

1. **Wrong write content**: The per-FD `memory.peak` reset requires a
   zero-length write (`write(fd, buf, 0)`). The previous code called
   `WriteString("0")` which writes 1 byte `0x30`. The kernel rejects any
   non-empty write with `EINVAL` on all kernel versions.

2. **Missing kernel version gate**: The per-FD reset mechanism was
   introduced in kernel 6.1 but only became fully stable (writable) in
   kernel 6.12. On kernels 6.1–6.11 the file exists and is readable, but
   any write returns `EINVAL`. The code attempted the write unconditionally,
   flooding logs on every `GetStats` call.

## Solution

- Add `kernelAtLeast(major, minor int) bool` — reads `uname` once at
  startup via `golang.org/x/sys/unix.Uname`, parses the release string
  (handles Ubuntu HWE format `6.8.0-110-generic`, Debian, RHEL suffixes).
- Extract `parseKernelAtLeast(release string, major, minor int) bool` as a
  pure function for unit-testability without syscalls.
- Add `peakResetSupported bool` field to `managerImpl`, set once in
  `NewManager()`.
- In `Create()`: only open `memory.peak` with `O_RDWR` when
  `peakResetSupported == true`; on older kernels `memoryPeakFile` stays
  `nil` — `GetStats` already skips the reset path when the FD is nil.
- Fix the write itself: `memoryPeakFile.Write([]byte{})` (zero-length)
  instead of `WriteString("0")`.

## Behaviour after this fix

| Kernel | `memoryPeakFile` | Reset attempted | `MemoryPeakBytes` |
|--------|-----------------|-----------------|-------------------|
| < 6.12 | `nil` | No | `0` (no tracking) |
| ≥ 6.12 | `O_RDWR` FD | Yes (empty write) | interval peak |

No WARN logs are emitted on kernels < 6.12. The fix is a no-op on
kernels ≥ 6.12 where reset already worked correctly.

## Tests

All tests pass with `go test ./pkg/sandbox/cgroup/...` (non-root) and
`sudo go test ./pkg/sandbox/cgroup/...` (root, real cgroup).

**New non-root unit tests** (no cgroup, no root required):
- `TestKernelAtLeast` — 19 table-driven sub-cases covering exact boundary
  `6.12`, below/above, major crossover, Ubuntu HWE `6.8.0-55-generic` /
  `6.8.0-110-generic` / `6.12.0-110-generic`, Debian, RHEL suffix formats.
- `TestKernelAtLeastMalformed` — 6 malformed inputs; must not panic,
  must return `false`.
- `TestGetStatsWithNilPeakFile` — `getStatsForPath` with `nil` peak FD
  returns valid stats and `MemoryPeakBytes == 0`.

**New root integration tests** (real cgroup):
- `TestNewManagerPeakResetFlag` — `peakResetSupported` matches
  `kernelAtLeast(6, 12)` on the running kernel.
- `TestCreateMemoryPeakFDOnSupportedKernel` — FD is non-nil on ≥ 6.12,
  nil on < 6.12.
- `TestMemoryPeakNoResetOnOldKernel` — simulates old-kernel manager;
  `GetStats` succeeds with no write attempt.
- `TestMemoryPeakResetWritesEmptyBytes` — on ≥ 6.12: empty write
  succeeds; writing `"0"` returns `EINVAL` (documents the original bug).
  Auto-skipped on < 6.12.

## Files changed

- `packages/orchestrator/pkg/sandbox/cgroup/manager.go`
- `packages/orchestrator/pkg/sandbox/cgroup/manager_test.go`


AdaAibaby:
Hi @arkamar,
This change has been verified and works properly on both 6.8.x and 6.12.x. Please kindly review this PR. Feel free to leave any comments. If everything looks good, you may also drop an LGTM. Thanks a lot!

jakubno:
Hey @AdaAibaby, we don't have capacity to support older version of kernel. There are also known kernel bugs, which has impact on `e2b`, so I would suggest to upgrade here

PR #2732, opened 2026-05-19
feat(envd): support user-defined file metadata via xattrs

mishushakov:
## Summary
- Adds optional user-defined file metadata to `POST /files`. Any request header of the form `X-Metadata-<key>: <value>` is persisted as an extended attribute (xattr) on the uploaded file. The `X-Metadata-` prefix is stripped and the remaining header name is lowercased to form the key. Multiple files in one multipart upload receive the same metadata.
- Surfaces that metadata on every `EntryInfo`: the HTTP upload response, and the filesystem gRPC service (`Stat`, `ListDir`, `Move`, `MakeDir`, etc.) via a new `map<string, string> metadata` proto field (field 11). The wiring is centralized in `filesystem.GetEntryInfo`, so all callers pick it up.
- xattrs are stored under the `user.e2b.` namespace — `user.` is required by the Linux VFS for unprivileged xattrs, and `e2b.` namespaces our keys so they don't collide with other tooling writing to `user.*`. Foreign `user.*` xattrs are filtered out and not surfaced.
- Replace-on-upload semantics: each upload rewrites the file's full metadata set. Keys present from a prior upload but absent from the new request are removed, and an upload with no `X-Metadata-*` headers clears all existing metadata (`O_TRUNC` preserves xattrs, so we always rewrite).
- Validation: keys and values must be printable US-ASCII (`0x20`–`0x7E`), else HTTP 400. Keys are capped at 246 bytes (255-byte VFS xattr-name limit minus the `user.e2b.` prefix); values at 1024 bytes.
- Error handling on write: xattr-unsupported filesystems (e.g. `/proc`, `/sys`) are best-effort — the body is persisted and we log a warning rather than fail; `ENOSPC`/`EDQUOT` map to HTTP 507. The response `EntryInfo` reads xattrs back from disk, so it never falsely claims metadata was persisted.

## Implementation
- xattr read/write/validate helpers live in a single cross-platform `packages/shared/pkg/filesystem/xattr.go`, backed by the [`github.com/pkg/xattr`](https://github.com/pkg/xattr) library rather than hand-rolled syscalls. The library handles the size-probe/retry and name parsing that previously lived in our own `xattr_linux.go`.
- `ReadMetadata`/`WriteMetadata` carry the e2b-specific policy on top of the library: `user.e2b.` namespacing, full-set replace semantics, and `ValidateMetadata`. `IsXattrUnsupported` is exported so `upload.go` can treat xattr-less filesystems as best-effort without duplicating the errno check.
- No platform shims needed — `pkg/xattr` is cross-platform, so the metadata tests run on Linux and macOS alike (only `extractStatTimes` keeps a `linux`/`darwin` split, for `Stat_t` field differences).
- Bumps envd version to `0.5.28` and the OpenAPI spec to `0.1.3`.

## Test plan
- [x] `make lint` clean for envd and shared
- [x] `go test ./...` passes for `packages/envd` and `packages/shared/pkg/filesystem` (metadata tests now run on macOS too)
- [ ] Upload a file with `-H 'X-Metadata-author: mish' -H 'X-Metadata-purpose: upload'` and confirm `Stat` returns the same map
- [ ] Confirm xattrs land under `user.e2b.*` via `getfattr -d <path>` inside a sandbox
- [ ] Re-upload the same path with different headers and confirm prior keys are cleared


mishushakov:
@claude review

mishushakov:
@claude review
@codex review

packages/envd/internal/api/upload.go
```diff
@@ -105,6 +132,23 @@ func processFile(r *http.Request, path string, part io.Reader, uid, gid int, log
 		return http.StatusInternalServerError, err
 	}
 
+	if len(metadata) > 0 {
+		if err := filesystem.WriteMetadata(path, metadata); err != nil {
+			switch {
+			case errors.Is(err, syscall.ENOTSUP) || errors.Is(err, syscall.EOPNOTSUPP):
```
jakubno: can you add comment it shouldn't happen on ext4?

Given that maybe returning error would make sense

packages/envd/internal/api/upload.go
```diff
@@ -20,11 +20,38 @@ import (
...
 
-func processFile(r *http.Request, path string, part io.Reader, uid, gid int, logger zerolog.Logger) (int, error) {
+// metadataHeaderPrefix is the request-header prefix used to attach
+// user-defined metadata to file uploads. Each `<metadataHeaderPrefix><key>:
+// <value>` header becomes a `user.<key>` xattr on the uploaded file.
+const metadataHeaderPrefix = "X-Metadata-"
+
+// extractMetadataHeaders returns all `X-Metadata-*` headers from h, with the
+// prefix stripped and keys lowercased. Returns nil if none are present.
+func extractMetadataHeaders(h http.Header) map[string]string {
```
jakubno, line 35: please add validation for name and value sizes

packages/envd/internal/api/upload.go
```diff
@@ -20,11 +20,38 @@ import (
...
+	var metadata map[string]string
+	for name, values := range h {
+		if len(values) == 0 || !strings.HasPrefix(name, metadataHeaderPrefix) {
+			continue
+		}
+		key := strings.ToLower(strings.TrimPrefix(name, metadataHeaderPrefix))
```
jakubno, line 41: I don't see any mention in spec that the key will be converted to lowercase

packages/shared/pkg/filesystem/entry_linux.go
```diff
@@ -13,3 +13,12 @@ func extractStatTimes(base *syscall.Stat_t) statTimes {
...
 }
+
+func readEntryMetadata(path string) map[string]string {
+	metadata, err := ReadMetadata(path)
+	if err != nil {
+		return nil
+	}
+
+	return metadata
+}
```
jakubno: this is useless, it's `ReadMetadata

```suggestion
```

packages/shared/pkg/filesystem/entry.go
```diff
@@ -75,6 +75,8 @@ func GetEntryInfo(path string, fileInfo os.FileInfo) EntryInfo {
 		entry.ModifiedTime = fileInfo.ModTime()
 	}
 
+	entry.Metadata = readEntryMetadata(path)
```
jakubno:
```suggestion
	entry.Metadata = ReadMetadata(path)
```

packages/shared/pkg/filesystem/xattr.go
```diff
@@ -0,0 +1,42 @@
...
+const MetadataXattrPrefix = "user."
+
+// MaxMetadataKeyLen caps the length of a metadata key. The Linux VFS limits
+// the full xattr name (including the `user.` prefix) to 255 bytes.
+const MaxMetadataKeyLen = 255 - len(MetadataXattrPrefix)
+
+// MaxMetadataValueLen caps the length of a metadata value. The kernel allows
+// up to 64KiB per xattr; we cap at 32KiB so the request stays well within
+// per-filesystem limits (ext4/xfs/btrfs all accept this).
+const MaxMetadataValueLen = 32 * 1024
```
jakubno: - ext4: a single xattr value must fit in one filesystem block. On the typical 4 KiB block size, the hard limit is ~4 KiB per value (and all xattrs on an inode must fit in that one block, minus the inline area in the inode). 32 KiB will fail with ENOSPC/E2BIG unless
   the FS was made with a 32 KiB+ block size, which is not standard on x86_64 Linux.

packages/shared/pkg/filesystem/xattr.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+)
+
+// MetadataXattrPrefix is the xattr namespace used to store user-defined
+// metadata that envd surfaces via its file-related APIs.
+const MetadataXattrPrefix = "user."
```
jakubno: shouldn't we make it more unique? something like `user.e2b.`?

packages/envd/internal/services/filesystem/utils.go
```diff
@@ -62,6 +62,7 @@ func entryInfo(path string) (*rpc.EntryInfo, error) {
 		Group:         group,
 		ModifiedTime:  toTimestamp(info.ModifiedTime),
 		SymlinkTarget: info.SymlinkTarget,
+		Metadata:      info.Metadata,
```
jakubno, line 65: should we include it in list as well? could you do some benchmark? It may end up with 2x syscalls

packages/envd/spec/envd.yaml
```diff
@@ -1,7 +1,7 @@
 openapi: 3.0.0
 info:
   title: envd
-  version: 0.1.1
+  version: 0.1.3
```
jakubno, line 4:
```suggestion
  version: 0.1.2
```

packages/envd/spec/envd.yaml
```diff
@@ -116,6 +116,13 @@ paths:
           $ref: "#/components/responses/InternalServerError"
     post:
       summary: Upload a file and ensure the parent directories exist. If the file exists, it will be overwritten.
+      description: |
+        Any request header of the form `X-Metadata-<key>: <value>` is persisted
+        as a user-defined extended attribute on the uploaded file. Keys are
+        lowercased and the `X-Metadata-` prefix is stripped; the resulting
+        metadata is returned on `EntryInfo` lookups (e.g. `Stat`, `ListDir`).
+        Header values must be US-ASCII. Multiple files in a single multipart
```
jakubno: > Header values must be US-ASCII
This isn't enforced

packages/shared/pkg/filesystem/xattr_linux.go
```diff
@@ -0,0 +1,124 @@
...
+		if b == 0 {
+			if i > start {
+				out = append(out, string(buf[start:i]))
+			}
+			start = i + 1
+		}
+	}
+	if start < len(buf) {
+		out = append(out, string(buf[start:]))
+	}
```
jakubno:
```suggestion
	parts := bytes.Split(buf, []byte{0})
	out := make([]string, 0, len(parts))
	for _, p := range parts {
		if len(p) == 0 {
			continue
		}
		out = append(out, string(p))
	}

```

packages/envd/internal/api/upload_metadata_test.go
```diff
@@ -0,0 +1,73 @@
...
+			}
+			for k, v := range tc.want {
+				if got[k] != v {
+					t.Errorf("key %q: got %q, want %q", k, got[k], v)
+				}
+			}
+		})
+	}
+
+	t.Run("multiple values for one header uses first value", func(t *testing.T) {
```
jakubno, line 61: also mention this in spec

jakubno (dismissed):

packages/shared/pkg/filesystem/entry_linux.go
mishushakov: good catch, the function was doing some more things before, it just got trimmed a lot

packages/envd/internal/api/upload.go
```diff
@@ -20,11 +20,38 @@ import (
...
 	"github.com/e2b-dev/infra/packages/envd/internal/utils"
+	"github.com/e2b-dev/infra/packages/shared/pkg/filesystem"
 )
 
 var ErrNoDiskSpace = errors.New("not enough disk space available")
 
-func processFile(r *http.Request, path string, part io.Reader, uid, gid int, logger zerolog.Logger) (int, error) {
+// metadataHeaderPrefix is the request-header prefix used to attach
+// user-defined metadata to file uploads. Each `<metadataHeaderPrefix><key>:
+// <value>` header becomes a `user.<key>` xattr on the uploaded file.
```
mishushakov: user.e2b.key

packages/envd/internal/api/upload.go
```diff
@@ -167,16 +215,21 @@ func (a *API) handlePart(r *http.Request, part *multipart.Part, paths UploadSucc
...
 	}
 
-	return &EntryInfo{
+	entry := &EntryInfo{
 		Path: filePath,
 		Name: filepath.Base(filePath),
 		Type: File,
-	}, http.StatusOK, nil
+	}
+	if persisted, _ := filesystem.ReadMetadata(filePath); len(persisted) > 0 {
```
mishushakov: we should log error at least

packages/envd/internal/api/upload.go
```diff
@@ -338,14 +400,19 @@ func (a *API) handleRawUpload(r *http.Request, u *user.User, uid, gid int, opera
...
 	}
 
-	return UploadSuccess{{
+	entry := EntryInfo{
 		Path: filePath,
 		Name: filepath.Base(filePath),
 		Type: File,
-	}}, http.StatusOK, nil
+	}
+	if persisted, _ := filesystem.ReadMetadata(filePath); len(persisted) > 0 {
```
mishushakov: we should at least log error

packages/envd/spec/envd.yaml
```diff
@@ -144,6 +144,23 @@ paths:
           $ref: "#/components/responses/InternalServerError"
     post:
       summary: Upload a file and ensure the parent directories exist. If the file exists, it will be overwritten.
+      description: |
```
mishushakov, line 147: can we also mention in case of duplicate keys it will always use first like we had in test case?

packages/shared/pkg/filesystem/entry.go
```diff
@@ -75,6 +75,8 @@ func GetEntryInfo(path string, fileInfo os.FileInfo) EntryInfo {
 		entry.ModifiedTime = fileInfo.ModTime()
 	}
 
+	entry.Metadata, _ = ReadMetadata(path)
```
mishushakov: at least log the error

packages/shared/pkg/filesystem/xattr.go
```diff
@@ -0,0 +1,155 @@
...
+// limits. Values larger than this are rejected with HTTP 400.
+const MaxMetadataValueLen = 1024
+
+// ValidateMetadata returns an error if any key/value pair would be rejected
+// by WriteMetadata. Validation is filesystem-agnostic and catches obvious
+// client mistakes before we issue any syscalls.
+//
+// Both keys and values are required to be printable US-ASCII (0x20-0x7E),
+// matching the constraint documented for `X-Metadata-*` request headers.
+func ValidateMetadata(metadata map[string]string) error {
```
mishushakov: we can just limit total size to 2 instead like in S3?

packages/shared/pkg/filesystem/xattr.go
```diff
@@ -0,0 +1,155 @@
...
+		}
+	}
+
+	for k, v := range metadata {
+		if err := xattr.Set(path, MetadataXattrPrefix+k, []byte(v)); err != nil {
+			return err
+		}
+	}
+
+	return nil
```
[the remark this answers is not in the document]
mishushakov: Accepted as a known limitation. There's no atomic multi-xattr syscall, and with the new 4 KiB total-size cap validated up front (42bffc09d), a partial write from ENOSPC is effectively unreachable on ext4 — the whole set is guaranteed to fit before we touch the file. Metadata is best-effort by design.

packages/envd/internal/api/upload.go
```diff
@@ -105,6 +132,27 @@ func processFile(r *http.Request, path string, part io.Reader, uid, gid int, log
...
+			// filesystems such as /sys and /proc that the upload API also
+			// supports. The file body was already persisted, so we log and
+			// continue; the response EntryInfo reads xattrs back from disk
+			// so it won't falsely claim metadata was persisted.
+			logger.Warn().Str("path", path).Err(err).Msg("filesystem does not support xattrs; metadata not persisted")
+		case errors.Is(err, syscall.ENOSPC) || errors.Is(err, syscall.EDQUOT):
+			return http.StatusInsufficientStorage, fmt.Errorf("not enough space for file metadata: %w", err)
+		default:
+			return http.StatusInternalServerError, fmt.Errorf("error writing file metadata: %w", err)
+		}
```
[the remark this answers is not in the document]
mishushakov: Accepted — inherent ordering: xattrs can only be set after the file exists. The common cases are already handled (unsupported FS warns and continues, ENOSPC maps to 507); only unexpected errors return 500 with the body persisted, which matches how the body-write path itself behaves. Not worth re-architecting.

packages/shared/pkg/filesystem/xattr.go
mishushakov: Done in 42bffc09d — dropped the per-value cap and replaced it with a single 4 KiB cap on total metadata size (stored names + values, summed across keys), validated up front so oversized sets get a clean 400 instead of a late ENOSPC. Went with 4 KiB rather than 2 KiB to match ext4's real per-inode xattr-block ceiling. The per-key length cap stays since it's the hard VFS xattr-name limit.

packages/shared/pkg/filesystem/xattr.go
```diff
@@ -0,0 +1,162 @@
...
+		return nil, err
+	}
+
+	var metadata map[string]string
+	for _, name := range names {
+		if !strings.HasPrefix(name, MetadataXattrPrefix) {
+			continue
+		}
+
+		value, err := xattr.Get(path, name)
```
matthewlouisbrockman, line 97: Probably need read-side validation here too. Uploads only write printable ASCII, but users can set user.e2b.* xattrs themselves inside the sandbox. If a value is raw bytes / invalid UTF-8, we put it into map<string,string> and the response can fail to marshal. Quick repro: set user.e2b.bad = []byte{0xff}, call Stat, then proto.Marshal(resp.Msg) fails with "string field contains invalid UTF-8". Probably should skip/sanitize/reject invalid xattrs when reading.

packages/shared/pkg/filesystem/entry.go
```diff
@@ -75,6 +75,12 @@ func GetEntryInfo(path string, fileInfo os.FileInfo) EntryInfo {
 		entry.ModifiedTime = fileInfo.ModTime()
 	}
 
+	// Metadata is best-effort: a read failure shouldn't fail the entry lookup
+	// (Size/Mode/times are still valid), and this helper has no logger to
+	// report it through. Callers that need the error (e.g. the upload handler)
+	// call ReadMetadata directly.
+	entry.Metadata, _ = ReadMetadata(path)
```
matthewlouisbrockman: tested and replicated; we should address

matthewlouisbrockman (changes requested):

packages/shared/pkg/filesystem/entry.go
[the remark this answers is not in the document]
mishushakov: Fixed in 89a17bf98. Added `GetEntryInfoWithoutMetadata` (builds the entry purely from the passed-in `fileInfo`, no xattr syscall) and switched the orchestrator volume service's `toEntry` to it. The volume proto has no metadata field and these paths are the host's view of a chroot-relative path, so the read was both wasted and a footgun. envd's Stat/ListDir still go through `GetEntryInfo` (correct namespace, surfaces metadata).

packages/shared/pkg/filesystem/xattr.go
mishushakov: Fixed in 89a17bf98. ReadMetadata now skips any `user.e2b.*` entry whose key or value isn't valid UTF-8 (`utf8.ValidString`/`utf8.Valid`), so directly-set raw-byte xattrs can't break proto marshaling of EntryInfo.metadata. Our upload path only writes printable ASCII, so well-formed metadata is unaffected. Added TestReadMetadataSkipsInvalidUTF8 with your 0xff repro.

packages/shared/pkg/filesystem/entry.go
mishushakov: Addressed in 89a17bf98 — the non-UTF-8 xattr that broke Stat marshaling is now skipped on read (see the xattr.go thread), and the metadata read is no longer issued from the orchestrator volume path. Could you re-check with your repro when you get a chance?

matthewlouisbrockman (approved):

PR #2735, opened 2026-05-19
Make sandbox traffic routable over ingress

sitole:
- Load balancer setup for sharing sbx connections between client-proxy directly and ingress
- Expose cpu and memory envs for ingress
- Expose env for configuring ingress traffic weight 
- Default weight is zero

`INGRESS_SANDBOX_TRAFFIC_WEIGHT` works are the handbrake when it's set to 0.
Configuration change is not affecting already established connections.

### DIFF
--- a/iac/modules/job-ingress/main.tf
+++ b/iac/modules/job-ingress/main.tf
@@ -12,6 +12,9 @@ locals {
 
     otel_collector_grpc_endpoint = var.otel_collector_grpc_endpoint
   })
+
+  ingress_cpu_count = var.ingress_cpu_count == null ? 1 : var.ingress_cpu_count
+  ingress_memory_mb = var.ingress_memory_mb == null ? 512 : var.ingress_memory_mb
 }
 
 resource "nomad_job" "ingress" {
@@ -19,8 +22,9 @@ resource "nomad_job" "ingress" {
     count         = var.ingress_count
     node_pool     = var.node_pool
     update_stanza = var.update_stanza
-    cpu_count     = var.ingress_cpu_count
-    memory_mb     = var.ingress_memory_mb
+
+    cpu_count = local.ingress_cpu_count
+    memory_mb = local.ingress_memory_mb
 
     control_port          = var.control_port
     ingress_port          = var.ingress_port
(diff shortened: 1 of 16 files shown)


tvi (approved):

iac/provider-gcp/nomad-cluster/network/main.tf
```diff
@@ -303,8 +303,19 @@ resource "google_compute_url_map" "orch_map" {
...
+
+    default_route_action {
+      weighted_backend_services {
+        backend_service = google_compute_backend_service.default["session"].self_link
+        weight          = 100 - var.ingress_sandbox_traffic_weight
+      }
+
+      weighted_backend_services {
+        backend_service = google_compute_backend_service.ingress.self_link
+        weight          = var.ingress_sandbox_traffic_weight
```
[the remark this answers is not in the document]
sitole: this is relevant, blocking merge

iac/modules/job-ingress/main.tf
```diff
@@ -12,15 +12,19 @@ locals {
 
     otel_collector_grpc_endpoint = var.otel_collector_grpc_endpoint
   })
+
+  ingress_cpu_count = var.ingress_cpu_count == null ? 1 : var.ingress_cpu_count
+  ingress_memory_mb = var.ingress_memory_mb == null ? 512 : var.ingress_memory_mb
```
djeebus, line 17: why the extra effort around defaults? passing nulls at the root and then setting defaults in code in the module hides what those defaults are, and complicates the code. if you set the defaults at the root, we may have to do it in two places (aws and gcp) but keeping all the defaults at the root simplifies reading the code.

iac/provider-gcp/nomad-cluster/network/variables.tf
```diff
@@ -109,3 +109,9 @@ variable "additional_api_paths_handled_by_ingress" {
     timeout_sec = optional(number)
   }))
 }
+
+variable "ingress_sandbox_traffic_weight" {
+  type        = number
+  description = "Weight (out of 100) of sandbox traffic routed to the ingress backend. The remainder goes to the session backend."
+  default     = 0
+}
```
djeebus, line 117: If it's out of 100, probably makes more sense to call it a percentage instead of a weight.

iac/modules/job-ingress/main.tf
sitole: good point i will re do it that way

PR #2744, opened 2026-05-20
feat(orchestrator): add dummy orchestrator binary for local API dev

tvi:
The real orchestrator is //go:build linux throughout (firecracker, NBD, NFS, cgroups, network namespaces), so on macOS dev machines there is nothing the api package can talk to. This adds a parallel, platform- agnostic dummy orchestrator binary that implements the orchestrator.proto SandboxService surface against an in-memory map so the api side mostly works end-to-end, plus an empty Chunk/Volume/Info surface and a HTTP /health endpoint that scripts/start-service.sh can probe.

Orchestrator changes:

- cmd/dummy-orchestrator/main.go: new binary entry point, no build tag. Plain grpc.NewServer (no cmux, no telemetry/feature-flags/redis). Wires up SandboxService (the in-memory dummy impl), Chunk/Volume services as Unimplemented embeds so calls return codes.Unimplemented instead of "connection refused", InfoService, and grpc health (status SERVING). Adds a separate HTTP /health endpoint on HEALTH_PORT (default 5018) so the integration-test harness can detect readiness without bringing in cmux. Reads GRPC_PORT (default 5008), HEALTH_PORT, NODE_ID, NODE_LABELS, SERVICE_VERSION; commitSHA is linker-set. Graceful shutdown on SIGINT/SIGTERM.
- pkg/dummyserver/sandbox.go: in-memory SandboxServiceServer. Create stores a clone of the SandboxConfig keyed by sandbox_id and returns a fixed per-process client_id. Update mutates end_time and returns NotFound if the sandbox is missing. List clones the map. Delete and Pause remove the entry (idempotent). Checkpoint returns an empty response. ListCachedBuilds returns an empty list.
- pkg/dummyserver/info.go: static InfoServiceServer. ServiceInfo returns Healthy, role Orchestrator, version/commit/labels from the binary, plus a minimal MachineInfo (cpu_architecture = runtime.GOARCH). ServiceStatusOverride is a no-op.
- Makefile: build-dummy (CGO_ENABLED=0, builds for the host GOOS, -ldflags sets main.commitSHA, outputs bin/dummy-orchestrator) and run-dummy (NODE_ID=\$(HOSTNAME) GRPC_PORT=5008 HEALTH_PORT=5018).

API changes:

- internal/cfg/model.go: new ServiceDiscoveryProviderLocal = "local" constant, accepted by the Parse() validator, plus LOCAL_ORCHESTRATOR_ADDRESS (default 127.0.0.1:5008).
- internal/orchestrator/discovery/local.go: NewLocal(addr) Discovery that always returns one Node at the configured host:port. Falls back to consts.OrchestratorAPIPort when the port is omitted.
- internal/clusters/discovery/static.go: NewStaticDiscovery(nil) so the template-builder discovery branch has an empty-but-non-error implementation in local mode (no template builders on a dev mac).
- internal/handlers/store.go: new case for cfg.ServiceDiscoveryProviderLocal that wires both discoveries without touching Nomad or k8s.

Run flow:

  cd packages/orchestrator && make run-dummy   # grpc :5008, health :5018
  SERVICE_DISCOVERY_PROVIDER=local make -C packages/api run

### DIFF
--- a/packages/api/internal/cfg/model.go
+++ b/packages/api/internal/cfg/model.go
@@ -25,6 +25,10 @@ const (
 	ServiceDiscoveryProviderNomad = "nomad"
 	// ServiceDiscoveryProviderKubernetes queries the in-cluster K8s API (the K8s deploy).
 	ServiceDiscoveryProviderKubernetes = "kubernetes"
+	// ServiceDiscoveryProviderLocal returns a single statically configured
+	// orchestrator address. Used to develop the API against the darwin dummy
+	// orchestrator on macOS, where neither Nomad nor Kubernetes is available.
+	ServiceDiscoveryProviderLocal = "local"
 )
 
 type Config struct {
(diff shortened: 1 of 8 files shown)


matthewlouisbrockman (approved):
just a dummy localhost for testing thing, not deployed, should be fine.

PR #2753, opened 2026-05-20
feat(api): add per-call timeouts for redis backend operations (ENG-4060)

jakubno:
Wrap every redis call in packages/api/internal/sandbox with a context.WithTimeout(ctx, 5s) so a stuck redis round trip cannot block callers (e.g. while holding a distributed lock) indefinitely.

Sites left unchanged: lock.Obtain (already bounded by caller timeout), lock publishReleaseNotification (own 5s timeout), and the long-lived subscription manager pubsub Subscribe.


ValentaTomas:

PR #2757, opened 2026-05-20
refactor: enforce blocked team with allowlist

drankou:
## Summary

Moves blocked-team enforcement from opt-in handler checks into shared middleware for API and dashboard API routes, with explicit allowlists for recovery/read/delete paths.

## Changes

- Added blocked-team middleware for `packages/api` and `packages/dashboard-api`
- Added route allowlists for blocked teams
- Add late team-resolution checks for access-token/user-auth API paths


packages/api/internal/middleware/blocked_team.go
```diff
@@ -0,0 +1,95 @@
...
+		"/templates/:templateID":                        {},
+		"/templates/:templateID/tags":                   {},
+		"/templates/aliases/:alias":                     {},
+		"/templates/:templateID/builds/:buildID/logs":   {},
+		"/templates/:templateID/builds/:buildID/status": {},
+		"/v2/sandboxes":                                 {},
+		"/v2/sandboxes/:sandboxID/logs":                 {},
+		"/volumes":                                      {},
+		"/volumes/:volumeID":                            {},
+	},
```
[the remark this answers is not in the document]
drankou: not relevant

packages/api/internal/handlers/auth.go
```diff
@@ -12,10 +12,28 @@ import (
...
 	"github.com/e2b-dev/infra/packages/auth/pkg/auth"
 	"github.com/e2b-dev/infra/packages/auth/pkg/types"
 )
 
+// applyBlockedTeamCheck mirrors middleware.EnforceBlockedTeam for code
+// paths that resolve the team late (i.e. access-token / user-id auth, where
+// the team is not on the gin context when the middleware runs). Returns a
+// 403 *api.APIError if the team is blocked AND the matched route is not in
+// the blocked-team allowlist.
+func applyBlockedTeamCheck(c *gin.Context, team *types.Team) *api.APIError {
```
drankou: fyi, user-auth routes only resolve the user during auth and may pick the team later in `GetTeam` / `resolveTemplateAndTeam`, so the middleware has no team to check yet. This keeps blocked-team enforcement in place for those late team-resolution paths

packages/api/internal/middleware/blocked_team.go
jakubno, line 1: can we do 1 shared blocked_team middleware which accepts map to filter out the paths 🙏🏻

jakubno (changes requested):

packages/api/internal/handlers/auth.go
```diff
@@ -26,6 +44,10 @@ func (a *APIStore) GetTeam(
 	defer span.End()
 
 	if team, ok := auth.GetTeamInfo(c); ok {
+		if apiErr := applyBlockedTeamCheck(c, team); apiErr != nil {
```
jakubno: this should be already checked if `team, ok := auth.GetTeamInfo(c)` returns `ok`, right? + it will cause e.g. `GetTemplatesAliasesAlias` return error

packages/api/internal/handlers/auth.go
```diff
@@ -52,6 +74,10 @@ func (a *APIStore) GetTeam(
 			}
 		}
 
+		if apiErr := applyBlockedTeamCheck(c, team); apiErr != nil {
```
jakubno: can you add comment that this is rather more restrictive but as this is deprecated, it is okay?

packages/api/internal/handlers/auth.go
```diff
@@ -131,6 +157,10 @@ func (a *APIStore) resolveTemplateAndTeam(
...
```
jakubno: same here

packages/api/internal/handlers/auth.go
```diff
@@ -147,6 +177,10 @@ func (a *APIStore) resolveTemplateAndTeam(
 
 		for _, t := range userTeams {
 			if t.Team.ID == aliasInfo.TeamID {
+				if apiErr := applyBlockedTeamCheck(c, t.Team); apiErr != nil {
```
jakubno: and here

jakubno (dismissed):

packages/api/internal/handlers/auth.go
drankou: yes, dropped

tvi (approved):

PR #2759, opened 2026-05-20
fix(envd): start pty after stream subscription

ValentaTomas:
Starts PTY processes after stream subscribers are registered so early PTY output is not dropped.


ValentaTomas:
This needs more testing, so closing for now.

PR #2760, opened 2026-05-20
feat(metrics): split timer measurements

ValentaTomas:
Emits explicit duration, byte, and count timer instruments while preserving legacy timer measurements.

Verified with `go test ./packages/shared/pkg/telemetry ./packages/shared/pkg/storage ./packages/orchestrator/pkg/sandbox/block/...`.


ValentaTomas:
Closing unrelated.

PR #2765, opened 2026-05-20
ci: retry gotestsum install on transient proxy.golang.org errors

ValentaTomas:
## Summary
- pr-tests and pr-tests-arm64 occasionally cancel a whole run because `go install gotest.tools/gotestsum@v1.13.0` hits an HTTP/2 stream reset against `proxy.golang.org` (`INTERNAL_ERROR; received from peer`).
- Wrap the install in a 3-attempt loop with backoff so a single transient network fault doesn't sink the run. No new dependencies, no behaviour change on the success path.

## Test plan
- Verify both workflows still install gotestsum on first try in green CI.
- Watch for naturally recurring proxy hiccups; retry should silently absorb them.

### DIFF
--- a/.github/workflows/pr-tests-arm64.yml
+++ b/.github/workflows/pr-tests-arm64.yml
@@ -146,7 +146,14 @@ jobs:
         if: matrix.package == 'packages/orchestrator'
 
       - name: Install gotestsum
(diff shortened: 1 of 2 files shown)


ValentaTomas:
Closing this for now, will try to have a more proper solution here.

PR #2766, opened 2026-05-20
perf(clean-nfs-cache): restore dirfd-relative statx

levb:


djeebus (approved):
All the important stuff is covered by tests, and the rest seems reasonable

PR #2768, opened 2026-05-21
fix(orchestrator): preserve checkpoint state

dobrac:
Summary:
- add checkpoint failure details with sandbox runtime state
- mirror RUNNING vs KILLED checkpoint failures in API state
- document checkpoint snapshot behavior as a spec

Tests:
- go test ./internal/orchestrator -run TestCheckpointFailureState


ValentaTomas:

PR #2769, opened 2026-05-21
fix(envd): ignore closed pty read errors

ValentaTomas:
## Summary
Handle PTY master EIO reads as normal teardown in envd, matching the common Linux behavior when the slave side closes.

References:
- https://github.com/creack/pty/issues/184
- https://github.com/owenthereal/upterm/pull/11

## Test plan
`go test ./internal/services/process/...` from `packages/envd`.

### DIFF
--- a/packages/envd/internal/services/process/handler/handler.go
+++ b/packages/envd/internal/services/process/handler/handler.go
@@ -221,7 +221,7 @@ func New(
 					}
 				}
 
-				if errors.Is(readErr, io.EOF) {
+				if errors.Is(readErr, io.EOF) || errors.Is(readErr, syscall.EIO) {
 					break
 				}
 
--- a/packages/envd/pkg/version.go
+++ b/packages/envd/pkg/version.go
@@ -1,3 +1,3 @@
 package pkg
 
-const Version = "0.5.27"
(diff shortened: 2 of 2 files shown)


packages/envd/internal/services/process/handler/handler.go
```diff
@@ -221,7 +221,7 @@ func New(
 					}
 				}
 
-				if errors.Is(readErr, io.EOF) {
+				if errors.Is(readErr, io.EOF) || errors.Is(readErr, syscall.EIO) {
```
wj-e2b, line 224: Shouldn't we break on almost all errors here? There's a bunch more that are possible that are not retryable, like EINVAL

wj-e2b (approved):
Technically fine, but mind as well catch all of them

packages/envd/internal/services/process/handler/handler.go
wj-e2b: EPIPE, etc

ValentaTomas: According to the linked issues, the eio is really expected so we just break gracefully on this one, while logging the more unexpected ones.

PR #2770, opened 2026-05-21
refactor(orchestrator): Fix leaking resources on shutdown and add framework for better lifecycle tracking

wj-e2b:
Pass 1.

Do some mild lifecycle tracking and then fix all the shutdown cleanup bugs.
Verified by shutting down orchestrator and seeing no leaked kernel resources.
Should enable graceful shutdowns.

### DIFF
--- a/packages/orchestrator/pkg/factories/run.go
+++ b/packages/orchestrator/pkg/factories/run.go
@@ -782,6 +782,8 @@ func run(config cfg.Config, opts Options) (success bool) {
 		}
 	}
 
+	orchestratorService.StartDraining(ctx)
+
 	// Wait for services to be drained before closing them
 	if tmpl != nil {
 		err := tmpl.Wait(closeCtx)
@@ -799,27 +801,21 @@ func run(config cfg.Config, opts Options) (success bool) {
 		}
 	}
 
-	closers = append(closers, closer{"sandbox drain", func(context.Context) error {
-		logger.L().Info(ctx, "Starting sandbox drain phase",
-			zap.Bool("forced", config.ForceStop),
-			zap.Int("sandbox_count", sandboxes.Count()),
-		)
+	logger.L().Info(ctx, "Starting sandbox drain phase",
+		zap.Bool("forced", config.ForceStop),
+		zap.Int("sandbox_count", sandboxes.Count()),
+	)
 
-		if config.ForceStop {
-			forceStopSandboxes()
-
-			return nil
-		}
-
+	if config.ForceStop {
+		forceStopSandboxes()
+	} else {
 		err := orchestratorService.DrainSandboxes(closeCtx)
 		if err != nil {
 			logger.L().Warn(ctx, "sandbox drain phase did not complete gracefully; forcing sandbox shutdown", zap.Error(err))
 
 			forceStopSandboxes()
 		}
-
-		return nil
-	}})
+	}
 
 	slices.Reverse(closers)
 	for _, closer := range closers {
--- a/packages/orchestrator/pkg/factories/run.go
+++ b/packages/orchestrator/pkg/factories/run.go
@@ -119,6 +119,16 @@ func (e serviceDoneError) Error() string {
 	return fmt.Sprintf("service %s finished", e.name)
 }
 
+func isServiceDoneError(err error) bool {
+	var sde serviceDoneError
+
+	return errors.As(err, &sde)
+}
+
+func isIgnorableSyncError(err error) bool {
+	return errors.Is(err, syscall.EINVAL)
+}
+
 // Run starts the orchestrator, blocking until shutdown.
 // Returns true on clean shutdown.
 func Run(opts Options) bool {
@@ -234,7 +244,7 @@ func run(config cfg.Config, opts Options) (success bool) {
 	// there's a panic.
 	defer func(g *errgroup.Group) {
 		err := g.Wait()
-		if err != nil {
+		if err != nil && !isServiceDoneError(err) {
 			log.Printf("error while shutting down: %v", err)
 			success = false
 		}
@@ -275,7 +285,7 @@ func run(config cfg.Config, opts Options) (success bool) {
 	}))
 	defer func(l logger.Logger) {
 		err := l.Sync()
-		if err != nil {
+		if err != nil && !isIgnorableSyncError(err) {
 			log.Printf("error while shutting down logger: %v", err)
 			success = false
 		}
@@ -293,7 +303,7 @@ func run(config cfg.Config, opts Options) (success bool) {
 	)
 	defer func(l logger.Logger) {
 		err := l.Sync()
-		if err != nil {
+		if err != nil && !isIgnorableSyncError(err) {
 			log.Printf("error while shutting down sandbox logger: %v", err)
 			success = false
 		}
@@ -311,7 +321,7 @@ func run(config cfg.Config, opts Options) (success bool) {
 	)
 	defer func(l logger.Logger) {
 		err := l.Sync()
-		if err != nil {
+		if err != nil && !isIgnorableSyncError(err) {
 			log.Printf("error while shutting down sandbox logger: %v", err)
 			success = false
 		}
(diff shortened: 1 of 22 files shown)


packages/orchestrator/pkg/factories/run.go
```diff
@@ -346,7 +352,30 @@ func run(config cfg.Config, opts Options) (success bool) {
...
+				if err := closeFn(closeCtx); err != nil {
+					clog.Error(ctx, "error during shutdown", zap.Error(err))
+
+					return err
+				}
+
+				return nil
+			},
+		})
+	}
```
djeebus: These feel like they should be part of the `lifecycle.Manager` struct, rather than helper functions here, no?

packages/orchestrator/pkg/factories/run.go
```diff
@@ -406,11 +435,11 @@ func run(config cfg.Config, opts Options) (success bool) {
 		logger.L().Fatal(ctx, "failed to create template cache", zap.Error(err))
 	}
 	templateCache.Start(ctx)
-	closers = append(closers, closer{"template cache", func(context.Context) error {
+	registerClose("template cache", []string{"feature flags", "limiter"}, func(context.Context) error {
```
djeebus: The dependencies-as-strings make me a little nervous, in terms of resiliency to future changes. At the least they should be consts, but if you're going to build dependency injection from scratch, may as well use types. Or use fx!

packages/orchestrator/pkg/factories/run.go
```diff
@@ -589,24 +620,29 @@ func run(config cfg.Config, opts Options) (success bool) {
...
 		if err != nil {
 			logger.L().Fatal(ctx, "failed to start nfs proxy", zap.Error(err))
 		}
-		closers = append(closers, nfsClosers...)
+		for _, unit := range nfsUnits {
+			if unit.Stop == nil {
+				registerUnit(unit)
+			} else {
+				registerClose(unit.Name, unit.After, unit.Stop)
+			}
```
djeebus: ???

packages/orchestrator/pkg/lifecycle/manager.go
```diff
@@ -0,0 +1,138 @@
...
+		m.names = make(map[string]struct{})
+	}
+
+	for _, unit := range units {
+		if unit.Name == "" {
+			return errors.New("lifecycle unit name is required")
+		}
+
+		if _, ok := m.names[unit.Name]; ok {
+			return fmt.Errorf("duplicate lifecycle unit %q", unit.Name)
```
djeebus: If we allow these units to be replaced, it gives us an extension point to swap out implementations before executing in the future.

packages/orchestrator/pkg/lifecycle/manager.go
```diff
@@ -0,0 +1,138 @@
...
+		}
+
+		m.names[unit.Name] = struct{}{}
+		unit.After = append([]string(nil), unit.After...)
```
djeebus: wtf??

packages/orchestrator/pkg/lifecycle/manager.go
```diff
@@ -0,0 +1,138 @@
...
+
+		if err := unit.Stop(ctx); err != nil {
+			errs = append(errs, fmt.Errorf("stop %q: %w", unit.Name, err))
+		}
+	}
+
+	return errors.Join(errs...)
+}
+
+func (m *Manager) startOrder() ([]Unit, error) {
```
djeebus: This function in particular makes me want to use a library sooner rather than later.

packages/orchestrator/pkg/sandbox/map.go
```diff
@@ -25,26 +27,44 @@ type MapSubscriber interface {
...
+// Map holds sandboxes that are live (running), known active lifecycles,
+// together with a IP-to-sandbox index. The indexes are managed independently.
 //
 // AssignNetwork/NetworkReleased manage the IP map,
 // MarkRunning/MarkStopping manage the live set.
 type Map struct {
-	live    *smap.Map[*Sandbox]
-	network *smap.Map[*Sandbox]
+	live       *smap.Map[*Sandbox]
+	lifecycles *smap.Map[lifecycleEntry]
```
djeebus: Is this a superset of the `live` and `network` maps? Can we remove either of those and exclusively use this new map?

packages/orchestrator/pkg/server/main.go
```diff
@@ -38,6 +39,8 @@ const uploadedBuildsTTL = 1 * time.Hour
 // MaxStartingInstancesPerNode feature flag and resize the semaphore.
 const startingSandboxesLimitRefreshInterval = 30 * time.Second
 
+const sandboxDrainLogInterval = 5 * time.Second
```
djeebus: For sandboxes that may remain up for 48 hours or more, logging every 5 seconds "still not done" is a lot of logs. maybe once a minute is better? or maybe a sliding scale:
- every 5 seconds for a minute
- every minute for an hour
- every 15 minutes for the next 48 hours

packages/orchestrator/pkg/server/main.go
```diff
@@ -149,16 +155,152 @@ func New(ctx context.Context, cfg ServiceConfig) (*Server, error) {
...
+			if err := sbx.Close(ctx); err != nil {
+				errCh <- fmt.Errorf("cleanup sandbox %s/%s: %w", sbx.Runtime.SandboxID, sbx.LifecycleID, err)
+				sbxLog.Error(ctx, "failed to cleanup sandbox after force stop", zap.Error(err))
+
+				return
+			}
+
+			if err := s.proxy.RemoveFromPool(sbx.LifecycleID); err != nil {
+				sbxLog.Warn(ctx, "failed to remove sandbox from proxy pool after force stop", zap.Error(err))
+			}
```
djeebus: Why is this necessary on force stop but not drain?

packages/orchestrator/pkg/factories/run.go
```diff
@@ -782,19 +818,44 @@ func run(config cfg.Config, opts Options) (success bool) {
...
 	}
 
-	slices.Reverse(closers)
-	for _, closer := range closers {
-		clog := globalLogger.With(zap.String("service", closer.name), zap.Bool("forced", config.ForceStop))
-		clog.Info(ctx, "closing")
-		if err := closer.close(closeCtx); err != nil {
-			clog.Error(ctx, "error during shutdown", zap.Error(err))
-			success = false
+	registerClose("sandbox drain", []string{"orchestrator server", "network pool", "device pool"}, func(context.Context) error {
```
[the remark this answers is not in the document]
djeebus: To double down on codex's comment, this close needs to happen as literally the first thing to happen on any signals, before any other closing can happen, and it needs to happen on its own (no concurrent closing can happen at the same time). it's not clear that's happening here.

djeebus:
Before merging, we should do the following:

- update [traffic simulator](https://github.com/e2b-dev/traffic-simulation/blob/main/tests/src/activities.ts) to add some slow sandboxes (even just `sleep $(( (RANDOM % 1800) + 1 ))` would probably work for this purpose)
- deploy this branch to somewhere that can be hit by traffic simulator (not sure how to deploy to a dev cluster, but that would work)
- randomly restart one orchestrator nomad allocation every ~15 minutes
- wait 24 hours
- ensure that every sandbox ran correctly

That should give us enough confidence to say "this is fixed"

packages/orchestrator/pkg/sandbox/map.go
wj-e2b: They're all used for different purposes that are disjoint. We can maybe merge the live one at some point, but this requires a bunch of code changes

packages/orchestrator/pkg/server/main.go
wj-e2b: Fixed

wj-e2b:
@claude review

packages/orchestrator/pkg/sandbox/network/firewall.go
```diff
@@ -109,7 +110,16 @@ func NewFirewall(tapIf string, orchestratorInternalIP string, extraAllowedCIDRs
 }
 
 func (fw *Firewall) Close() error {
-	return fw.conn.CloseLasting()
+	fw.conn.DelTable(&nftables.Table{
+		Name:   tableName,
+		Family: nftables.TableFamilyINet,
+	})
```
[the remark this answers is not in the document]
wj-e2b: These are namespaced

packages/orchestrator/pkg/factories/run.go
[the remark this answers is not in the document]
wj-e2b: Ingress can't be closed because we want to still be able to stop sandboxes from the api.

PR #2771, opened 2026-05-21
fix(orchestrator): report real V4 header compression ratio

ValentaTomas:
V4 header serialization LZ4-compresses its inner block, but the upload metric reported equal uncompressed and stored sizes so the ratio was always 1:1. `StoreHeader` now returns the pre-LZ4 byte count alongside the stored size and the metrics caller uses it. V3 headers stay at 1:1.

### DIFF
--- a/packages/orchestrator/pkg/sandbox/upload_metrics.go
+++ b/packages/orchestrator/pkg/sandbox/upload_metrics.go
@@ -47,13 +47,16 @@ func uploadRatioBp(compressed, uncompressed int64) int64 {
 }
 
 func storeHeaderWithMetrics(ctx context.Context, store storage.StorageProvider, path, fileType string, h *headers.Header) error {
-	stats, err := headers.StoreHeader(ctx, store, path, h)
+	stored, uncompressed, err := headers.StoreHeader(ctx, store, path, h)
 	if err != nil {
 		return err
 	}
 
-	cfg := storage.CompressConfig{Type: stats.Compression.String()}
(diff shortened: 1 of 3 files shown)


packages/orchestrator/pkg/sandbox/upload_metrics.go
```diff
@@ -47,12 +47,16 @@ func uploadRatioBp(compressed, uncompressed int64) int64 {
...
 
 func storeHeaderWithMetrics(ctx context.Context, store storage.StorageProvider, path, fileType string, h *headers.Header) error {
-	size, err := headers.StoreHeader(ctx, store, path, h)
+	stored, uncompressed, err := headers.StoreHeader(ctx, store, path, h)
 	if err != nil {
 		return err
 	}
 
-	recordUploadCompression(ctx, uploadArtifactHeader, fileType, storage.CompressConfig{}, size, size)
+	cfg := storage.CompressConfig{}
```
levb: 0/5 total style nit - probably a little better to compose the config in serialize since that's where we hardcode it, and return the entire config up. But that's another return value.

levb (approved):
LGTM

packages/orchestrator/pkg/sandbox/upload_metrics.go
ValentaTomas: Done — composed in StoreHeader and returned up; caller no longer duplicates the V4→LZ4 dispatch.

PR #2779, opened 2026-05-21
chore: add internal as non default entrypoint

jakubno:


iac/modules/job-ingress/jobs/traefik.toml
```diff
@@ -10,6 +10,7 @@
 
   [entryPoints.web]
     address = ":${ingress_port}"
+    asDefault = true
```
sitole, line 13: Isn't the concert opposite? When exposing a service, I want to force it to have an entrypoint defined; otherwise, I can leak internal service when missing the entrypoint setting.

sitole: from Traefik docs: `If there is no entryPoint with the asDefault option set to true, then the list of default entryPoints includes all HTTP/TCP entryPoints.`

So its safer to have at least one

sitole (approved):

PR #2786, opened 2026-05-22
feat(orch): static NUD_PERMANENT neighbours + ARP/IPv6 off on per-sandbox veth

ValentaTomas:
In `Slot.CreateNetwork`, disable ARP and pin the veth ↔ vpeer L2 next hop as `NUD_PERMANENT` on both sides. The IPv4 neigh table is then completely off the per-sandbox hot path (no GC, no ARP, no `gc_thresh*` impact). IPv6 is disabled host-wide by #2785, so no per-link IPv6 handling needed.

MASQUERADE egress and host-service DNATs are unaffected (pure L3 / DNAT paths).

### DIFF
--- a/packages/db/pkg/testutils/queries/models.go
+++ b/packages/db/pkg/testutils/queries/models.go
@@ -220,7 +220,6 @@ type User struct {
 	CreatedAt time.Time
 	UpdatedAt time.Time
 	ID        uuid.UUID
-	Email     string
 }
 
 type UserIdentity struct {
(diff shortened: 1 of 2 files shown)


ValentaTomas:
Closing — TestSmokeAllFCVersions consistently fails on this branch (sandbox provisioning times out during apt-get inside the FC VM) and isolating which kernel/netns interaction makes the static NUD_PERMANENT entries break in-VM egress would push this beyond the small-PR budget. The IPv4 neighbour overflow at default gc_thresh3=1024 is already addressed by the gc_thresh* bump in #2785; the structural fix (no-GC permanent entries) can be re-attempted as a follow-up.

PR #2787, opened 2026-05-22
perf(orchestrator): pace block.Cache dirty pages via sync_file_range

ValentaTomas:
## Summary

Fresh-template fills write several GB into a per-resume `block.Cache` mmap on md0 within seconds. On affected hosts that exceeds the per-bdi share of `vm.dirty_ratio` and parks orchestrator threads in `balance_dirty_pages` for tens of seconds — observed: a 32 GB-memfile template resuming in ~55 s instead of a few seconds.

- `block.Cache`: hold the cache fd open and call `sync_file_range(WRITE)` after each `WriteAtWithoutLock` so the kernel drains writeback continuously instead of letting dirty pages pile up to the threshold.
- GCP client bootstrap: `vm.dirty_ratio` 20 → 40 and `vm.dirty_background_ratio` 10 → 5 — adds headroom and starts background writeback earlier.

`sync_file_range(SYNC_FILE_RANGE_WRITE)` is non-blocking and runs in kernel flusher threads; mmap reads (UFFD `Slice` → `UFFDIO_COPY`) don't conflict with `PG_writeback`, so this doesn't slow fetch or serving.

## Test plan

- [ ] `go vet`/`go build` for orchestrator block package on linux
- [ ] Time a fresh-template resume on a memory-heavy template before/after on staging
- [ ] Watch `/proc/meminfo` Dirty/Writeback during the fill and gRPC latency dashboard for regression

### DIFF
--- a/iac/provider-gcp/nomad-cluster/scripts/start-client.sh
+++ b/iac/provider-gcp/nomad-cluster/scripts/start-client.sh
@@ -127,6 +127,12 @@ net.ipv4.tcp_max_syn_backlog = 65535
 # Increase the maximum number of memory map areas
 vm.max_map_count=1048576
 
+# Allow more dirty pages and start writeback earlier so a per-resume
+# block.Cache fill on md0 does not pile up enough dirty memory to trigger
+# balance_dirty_pages throttling on these hosts.
+vm.dirty_ratio = 40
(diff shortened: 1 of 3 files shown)


ValentaTomas:

PR #2788, opened 2026-05-22
perf(orch): batch per-sandbox iptables + install host-service redirects globally

ValentaTomas:
## What

Two independent changes against the per-sandbox iptables cost:

**1. Batch the per-sandbox host iptables work.** Introduce a `RuleSet` builder in `pkg/sandbox/network` that accumulates appends/deletes and flushes them via a single `iptables-restore --noflush` per `CreateNetwork` / `RemoveNetwork`. The egress proxy contributes its rules into the same `RuleSet`. The in-namespace SNAT/DNAT rules keep using `*iptables.IPTables` (they live in per-sandbox netns and are not on the global host tables). `RemoveNetwork` uses `ApplyBestEffort` so a partial earlier `CreateNetwork` doesn't abort cleanup.

**2. Install host-service REDIRECT rules globally at boot.** The hyperloop / portmapper / NFS-proxy DNAT redirects all match `-d OrchestratorInSandboxIPAddress` (an unrouted host-side address only reachable from inside a sandbox netns), so the `-i veth-N` constraint is unnecessary. `installGlobalSandboxRules` runs once from `NewStorageLocal` using `AppendUnique`, and the three appends/deletes drop from `CreateNetwork`/`RemoveNetwork`.

## Why

On Ubuntu 22+ `iptables` is `iptables-nft`. Each `iptables -A` reads, mutates, and rewrites the whole table under the xtables lock — total host provisioning cost is **O(N²)** in the table size. At ~150 sandboxes that's already ~1300 nat PREROUTING entries; the 500 ms `Acquire` timeout starts firing well before 1000 sandboxes/node. See `design/2026-04-28-100x-scaling/00-orchestrator-scaling-bottlenecks.md` §1.1.

Batching collapses the ~6–9 shellouts per CreateNetwork into one fork + one rewrite per touched table. Moving the three global redirects out of per-sandbox path eliminates 3 of the 9 rules outright; the remaining per-sandbox rules are still batched. F28 (shrinking `runtime.LockOSThread` window) and the full nftables verdict-map rewrite (target) are out of scope and tracked separately.

## Risk

- `iptables-restore --noflush` is atomic per table — if any line fails to parse, none of that table's lines apply. For adds this is desirable. For deletes we use `ApplyBestEffort`, which falls back to per-rule `iptables.Delete()` on failure so a missing rule does not abort teardown.
- Global redirect install is idempotent (`AppendUnique`); restarts mid-fleet are safe. During the transition, old per-sandbox rules (still present from sandboxes created on the previous binary) coexist with the new global rule — both match the same packets and behavior is unchanged.

## Out of scope (deferred for follow-ups)

- nftables verdict-map keyed by `iifname` — the "target" rewrite. Would collapse per-sandbox `-A` into an atomic O(1) set-element insert.
- `runtime.LockOSThread` window reduction in `CreateNetwork`. Becomes more meaningful after iptables work shrinks further.
- TCP firewall: 3 ports (HTTP/TLS/other) per sandbox still go through batched iptables; collapsing them into one verdict-map needs nftables.

### DIFF
--- a/packages/db/pkg/testutils/queries/models.go
+++ b/packages/db/pkg/testutils/queries/models.go
@@ -220,7 +220,6 @@ type User struct {
 	CreatedAt time.Time
 	UpdatedAt time.Time
 	ID        uuid.UUID
-	Email     string
 }
 
 type UserIdentity struct {
--- a/packages/orchestrator/cmd/resume-build/main.go
+++ b/packages/orchestrator/cmd/resume-build/main.go
@@ -18,7 +18,6 @@ import (
 	"time"
 
 	"github.com/containernetworking/plugins/pkg/ns"
-	"github.com/coreos/go-iptables/iptables"
 	"github.com/google/uuid"
 	"github.com/launchdarkly/go-sdk-common/v3/ldlog"
 	"github.com/launchdarkly/go-sdk-common/v3/ldvalue"
@@ -1525,7 +1524,7 @@ type noEgressProxy struct {
 	network.NoopEgressProxy
 }
 
-func (noEgressProxy) OnSlotCreate(s *network.Slot, _ *iptables.IPTables) error {
+func (noEgressProxy) OnSlotCreate(s *network.Slot, _ *network.RuleSet) error {
 	nsPath := filepath.Join("/var/run/netns", s.NamespaceID())
 
 	handle, err := ns.GetNS(nsPath)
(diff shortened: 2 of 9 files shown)


ValentaTomas:
Superseded by the proper implementation in #

ValentaTomas:
Superseded by #2789 — proper implementation: nftables verdict-map keyed by veth ifname (O(1) per sandbox) instead of batched iptables-restore (O(N) per sandbox).

PR #2789, opened 2026-05-22
perf(orch): replace per-sandbox host iptables with nftables verdict map

ValentaTomas:
Replace the per-sandbox host iptables work with a shared nftables ruleset (`inet e2b_host`) created once at orchestrator boot. All sandbox-keyed behaviour (forward accept, source MASQUERADE, TCP-firewall redirects) hangs off a `sandbox_veths` ifname set; `CreateNetwork` / `RemoveNetwork` mutate the set via one netlink op each — O(1) regardless of fleet size.

Replaces ~6 per-sandbox `iptables` shellouts that ran O(N²) under `iptables-nft` (every `-A` rewrote the whole table under the xtables lock).

Hook priorities (-110 / -90 / -10) run before the legacy iptables-nft compat chains, so leftover per-sandbox iptables rules from older binaries become harmless no-ops during a rolling deploy. The table is dropped and re-added on each orchestrator start.

The `EgressProxy` interface drops its `*iptables.IPTables` parameter — `tcpfirewall.Proxy` no longer needs per-sandbox iptables work (set membership covers it).

### DIFF
--- a/packages/db/pkg/testutils/queries/models.go
+++ b/packages/db/pkg/testutils/queries/models.go
@@ -220,7 +220,6 @@ type User struct {
 	CreatedAt time.Time
 	UpdatedAt time.Time
 	ID        uuid.UUID
-	Email     string
 }
 
 type UserIdentity struct {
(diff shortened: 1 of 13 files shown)


ValentaTomas:

PR #2570, opened 2026-05-05
refactor(storage): consolidate IO reader wrappers

levb:
Unify the scattered io.ReadCloser wrapper types into a single
io_wrappers.go file with interface assertions as a TOC:
offsetReader, instrumentedReader, cancelReader, sectionReader.

Pre-cursor to https://github.com/e2b-dev/infra/pull/2431, taking all non-functional refactoring into a separate PR

PR #2790, opened 2026-05-22
perf(storage): zero-copy GCS multipart upload from disk

ValentaTomas:
Stream each multipart part body straight from the open file via a fresh `io.SectionReader` per attempt instead of allocating a 50 MB chunk per part. Drop per-part `Content-MD5` (final-object integrity is verified by GCS; TLS protects in-transit). Drop the sibling-goroutine SHA-256 in `UploadFileInParallel` — the compressed upload path already hashes the source while reading it, so the uncompressed multipart path no longer needs its own.

Inspired by #2570 (sectionReader pattern) and #2393 (drop per-part MD5).


ValentaTomas:

PR #2803, opened 2026-05-22
fix(compression): correctness findings from compression audit

ValentaTomas:
Compression-audit correctness fixes:

- reject misaligned writes in `Cache.WriteAtWithoutLock`
- cap V4 header LZ4 decompression at the size prefix
- drop corrupt compressed cache entries on size mismatch
- retry GCS 404 within a post-transition window with backoff
- per-Read idle timeout on GCS range reads

### DIFF
--- a/packages/orchestrator/pkg/sandbox/block/cache.go
+++ b/packages/orchestrator/pkg/sandbox/block/cache.go
@@ -342,6 +342,10 @@ func (c *Cache) WriteAtWithoutLock(b []byte, off int64) (int, error) {
 		return 0, nil
 	}
 
+	if int64(len(b))%c.blockSize != 0 || off%c.blockSize != 0 {
+		return 0, fmt.Errorf("misaligned write: len=%d off=%d block=%d", len(b), off, c.blockSize)
+	}
+
 	end := min(off+int64(len(b)), c.size)
 	if end <= off {
 		return 0, nil
--- a/packages/orchestrator/pkg/sandbox/template/peerclient/seekable.go
+++ b/packages/orchestrator/pkg/sandbox/template/peerclient/seekable.go
@@ -7,12 +7,18 @@ import (
 	"io"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"go.uber.org/zap"
 
 	"github.com/e2b-dev/infra/packages/shared/pkg/grpc/orchestrator"
 	"github.com/e2b-dev/infra/packages/shared/pkg/logger"
 	"github.com/e2b-dev/infra/packages/shared/pkg/storage"
+)
+
+const (
+	postTransitionRetryWindow = 30 * time.Second
+	postTransitionRetryDelay  = 250 * time.Millisecond
 )
 
 var _ storage.Seekable = (*peerSeekable)(nil)
@@ -33,13 +39,11 @@ type peerSeekable struct {
 	baseCT storage.CompressionType
 	loaded bool
 
-	// transitionEmitted ensures we signal PeerTransitionedError at most once
-	// after the peer flips uploaded=true. The caller (build.File) reacts by
-	// loading the post-upload header from storage; whether that ends up V4
-	// (compressed) or V3 (no upgrade) determines how subsequent reads route.
-	// Either way, after the first emission we fall through to base so V3
-	// builds don't loop forever against PeerTransitionedError.
-	transitionEmitted atomic.Bool
+	// transitionAt is set on first PeerTransitionedError emission after
+	// uploaded flips to true; nil otherwise. Subsequent base 404s within
+	// postTransitionRetryWindow re-emit PeerTransitionedError so concurrent
+	// readers retry against the post-upload header.
+	transitionAt atomic.Pointer[time.Time]
 }
 
 // getBase returns a base Seekable opened against the storage path composed
@@ -126,8 +130,11 @@ func (s *peerSeekable) OpenRangeReader(ctx context.Context, off int64, length in
 		return res.value, err
 	}
 
-	if s.uploaded != nil && s.uploaded.Load() && s.transitionEmitted.CompareAndSwap(false, true) {
-		return nil, &storage.PeerTransitionedError{}
+	if s.uploaded != nil && s.uploaded.Load() {
+		now := time.Now()
+		if s.transitionAt.CompareAndSwap(nil, &now) {
+			return nil, &storage.PeerTransitionedError{}
+		}
 	}
 
 	base, err := s.getBase(ctx, frameTable.CompressionType())
@@ -135,7 +142,16 @@ func (s *peerSeekable) OpenRangeReader(ctx context.Context, off int64, length in
 		return nil, err
 	}
 
-	return base.OpenRangeReader(ctx, off, length, frameTable)
+	rc, err := base.OpenRangeReader(ctx, off, length, frameTable)
+	// GCS can briefly 404 a just-finalized object; within the retry window
+	// re-emit so build.File reloads the header and retries with backoff.
+	if errors.Is(err, storage.ErrObjectNotExist) {
+		if at := s.transitionAt.Load(); at != nil && time.Since(*at) < postTransitionRetryWindow {
+			return nil, &storage.PeerTransitionedError{RetryAfter: postTransitionRetryDelay}
+		}
+	}
+
+	return rc, err
 }
 
 func (s *peerSeekable) StoreFile(context.Context, string, ...storage.PutOption) (*storage.FrameTable, [32]byte, error) {
--- a/packages/shared/pkg/storage/header/serialization_v4.go
+++ b/packages/shared/pkg/storage/header/serialization_v4.go
@@ -3,6 +3,7 @@ package header
 import (
 	"bytes"
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"io"
 	"slices"
@@ -19,6 +20,8 @@ const v4SizePrefixLen = 4
 
 // v4FlagsLen is the length of the V4 flags byte. Bit 0 = IncompletePendingUpload.
 const v4FlagsLen = 1
+
+const v4MaxUncompressedHeaderSize = 64 << 20
 
 // v4FlagIncomplete is bit 0 of the V4 flags byte: when set, the header
 // describes a build whose upload has not yet finalized (an in-flight diff).
@@ -130,10 +133,17 @@ func deserializeV4(metadata *Metadata, blockData []byte) (*Header, error) {
 	}
 
 	flags := blockData[0]
+	size := binary.LittleEndian.Uint32(blockData[v4FlagsLen:])
+	if size > v4MaxUncompressedHeaderSize {
+		return nil, fmt.Errorf("v4 header uncompressed size %d exceeds cap %d", size, v4MaxUncompressedHeaderSize)
+	}
 
-	decompressed, err := decompressLZ4(blockData[v4FlagsLen+v4SizePrefixLen:])
+	decompressed, err := decompressLZ4(blockData[v4FlagsLen+v4SizePrefixLen:], int(size))
 	if err != nil {
 		return nil, fmt.Errorf("failed to LZ4-decompress v4 header block: %w", err)
+	}
+	if len(decompressed) != int(size) {
+		return nil, fmt.Errorf("v4 header decompressed size %d != prefix %d", len(decompressed), size)
 	}
 
 	reader := bytes.NewReader(decompressed)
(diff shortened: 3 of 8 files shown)


packages/orchestrator/pkg/sandbox/block/cache.go
```diff
@@ -342,6 +342,10 @@ func (c *Cache) WriteAtWithoutLock(b []byte, off int64) (int, error) {
 		return 0, nil
 	}
 
+	if int64(len(b)) < c.blockSize || off%c.blockSize != 0 {
+		return 0, fmt.Errorf("WriteAtWithoutLock requires block-aligned write: len=%d off=%d blockSize=%d", len(b), off, c.blockSize)
+	}
```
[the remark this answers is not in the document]
ValentaTomas: Good catch — fixed in d26c7de. Now requires `len(b) % blockSize == 0` AND aligned offset.

packages/orchestrator/pkg/sandbox/template/peerclient/seekable.go
```diff
@@ -135,7 +142,21 @@ func (s *peerSeekable) OpenRangeReader(ctx context.Context, off int64, length in
 		return nil, err
 	}
 
-	return base.OpenRangeReader(ctx, off, length, frameTable)
+	rc, err := base.OpenRangeReader(ctx, off, length, frameTable)
+	if errors.Is(err, storage.ErrObjectNotExist) && s.withinTransitionWindow() {
+		return nil, &storage.PeerTransitionedError{}
+	}
```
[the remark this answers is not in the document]
ValentaTomas: Valid — fixed by adding the same translation `Size` and `WriteTo` already do: `gcpObject.openRangeReader` now maps `cloud.google.com/go/storage.ErrObjectNotExist` to the local `storage.ErrObjectNotExist` so the B5 transition-window check actually matches on GCS.

packages/orchestrator/pkg/sandbox/template/peerclient/seekable.go
```diff
@@ -135,7 +141,15 @@ func (s *peerSeekable) OpenRangeReader(ctx context.Context, off int64, length in
 		return nil, err
 	}
 
-	return base.OpenRangeReader(ctx, off, length, frameTable)
+	rc, err := base.OpenRangeReader(ctx, off, length, frameTable)
+	if errors.Is(err, storage.ErrObjectNotExist) {
+		at := s.transitionAt.Load()
+		if at != 0 && time.Since(time.Unix(0, at)) < postTransitionRetryWindow {
+			return nil, &storage.PeerTransitionedError{}
```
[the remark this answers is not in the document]
ValentaTomas: Fixed in 60ed9c2bf: the base 404 transition path now re-emits only once by clearing the transition timestamp with CAS.

packages/orchestrator/pkg/sandbox/template/peerclient/seekable.go
```diff
@@ -135,7 +141,17 @@ func (s *peerSeekable) OpenRangeReader(ctx context.Context, off int64, length in
...
 	}
 
-	return base.OpenRangeReader(ctx, off, length, frameTable)
+	rc, err := base.OpenRangeReader(ctx, off, length, frameTable)
+	if errors.Is(err, storage.ErrObjectNotExist) {
+		at := s.transitionAt.Load()
+		if at != 0 && time.Since(time.Unix(0, at)) < postTransitionRetryWindow {
+			if s.transitionAt.CompareAndSwap(at, 0) {
+				return nil, &storage.PeerTransitionedError{}
+			}
```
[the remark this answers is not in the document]
ValentaTomas: Fixed in 8755e377d: concurrent callers all remain retriable, and post-transition base 404 retries now carry a small RetryAfter delay to avoid spinning.

packages/shared/pkg/storage/header/serialization_v4.go
```diff
@@ -242,14 +252,13 @@ func extractRelevantRanges(mappings []BuildMap) map[uuid.UUID][]storage.Range {
...
 
-// decompressLZ4 decompresses an LZ4 frame from V4 header data.
-func decompressLZ4(src []byte) ([]byte, error) {
+func decompressLZ4(src []byte, expected int) ([]byte, error) {
 	r := lz4.NewReader(bytes.NewReader(src))
 
-	data, err := io.ReadAll(r)
-	if err != nil {
+	buf := bytes.NewBuffer(make([]byte, 0, expected))
+	if _, err := io.CopyN(buf, r, int64(expected)+1); err != nil && !errors.Is(err, io.EOF) {
```
jakubno, line 261: can you add comment explaining what we want to achieve here? 🙏🏻 are we expecting unbounded streams?

packages/orchestrator/pkg/sandbox/build_upload_v4.go
```diff
@@ -94,7 +94,9 @@ func (u *Upload) uploadFramed(
 	if err := u.appendAncestorBuilds(ctx, h.Builds, srcHeader.Mapping, fileType); err != nil {
 		return err
 	}
-	h.Builds[u.buildID] = selfBuild
+	if srcPath != "" {
```
jakubno: could you explain this? 🙏🏻

packages/orchestrator/pkg/sandbox/template/peerclient/seekable.go
```diff
@@ -135,7 +144,15 @@ func (s *peerSeekable) OpenRangeReader(ctx context.Context, off int64, length in
 		return nil, err
 	}
 
-	return base.OpenRangeReader(ctx, off, length, frameTable)
+	rc, err := base.OpenRangeReader(ctx, off, length, frameTable)
+	if errors.Is(err, storage.ErrObjectNotExist) {
+		at := s.transitionAt.Load()
```
jakubno: we could use atomic.Pointer[time.Time] so we don't have to do the conversions

packages/orchestrator/pkg/sandbox/template/peerclient/seekable.go
```diff
@@ -135,7 +144,15 @@ func (s *peerSeekable) OpenRangeReader(ctx context.Context, off int64, length in
 		return nil, err
 	}
 
-	return base.OpenRangeReader(ctx, off, length, frameTable)
+	rc, err := base.OpenRangeReader(ctx, off, length, frameTable)
+	if errors.Is(err, storage.ErrObjectNotExist) {
```
jakubno, line 148: this changes the behavior slightly, could you add comment with explanation

packages/shared/pkg/storage/header/serialization_v4.go
ValentaTomas: Bound on the legitimate stream — LZ4 can expand input ~255×, so without `expected+1` a malformed/oversized header would let `io.ReadAll` allocate gigabytes. The size prefix is already capped at `v4MaxUncompressedHeaderSize`, so we know the legitimate length and reject anything longer (caller in `deserializeV4` checks `len(decompressed) != int(size)`). Added a comment in 27eeb26.

packages/orchestrator/pkg/sandbox/build_upload_v4.go
ValentaTomas: When `srcPath == ""` there is no diff to upload, so `selfBuild` stays the zero `BuildData`. Writing it would produce an entry on disk indistinguishable from a real empty-size build; readers should fall through the mapping path instead of trusting a synthetic self entry. Added a comment in 27eeb26.

packages/orchestrator/pkg/sandbox/template/peerclient/seekable.go
ValentaTomas: Switched to `atomic.Pointer[time.Time]` in 27eeb26.

packages/orchestrator/pkg/sandbox/template/peerclient/seekable.go
ValentaTomas: Added a comment in 27eeb26: within `postTransitionRetryWindow` we re-emit `PeerTransitionedError` on every base 404 (not just the first), so concurrent readers all retry; `RetryAfter` keeps `build.File` from spinning during GCS visibility lag.

packages/shared/pkg/storage/header/serialization_v4.go
ValentaTomas: Comment added in 393ddbe — bounded read defends against LZ4 ~255× expansion on malformed input; `deserializeV4` rejects any overrun.

packages/orchestrator/pkg/sandbox/build_upload_v4.go
ValentaTomas: Reverted in 393ddbe — `buildFileSize` already returns 0 for missing entries, so writing/skipping the zero `BuildData` is observationally identical. Dropping it minimizes the diff.

packages/orchestrator/pkg/sandbox/template/peerclient/seekable.go
ValentaTomas: Done in 393ddbe — also collapsed `transitionEmitted` into a single `atomic.Pointer[time.Time]` via CAS, so emission gating + timestamp share one field.

packages/orchestrator/pkg/sandbox/template/peerclient/seekable.go
ValentaTomas: Comment added in 393ddbe.

packages/shared/pkg/storage/storage_google.go
```diff
@@ -265,27 +265,50 @@ func (o *gcpObject) Size(ctx context.Context) (int64, error) {
 }
 
 func (o *gcpObject) openRangeReader(ctx context.Context, off, length int64) (io.ReadCloser, error) {
-	ctx, cancel := context.WithTimeout(ctx, googleReadTimeout)
+	readCtx, cancel := context.WithCancel(ctx)
 
-	reader, err := o.handle.NewRangeReader(ctx, off, length)
+	openTimer := time.AfterFunc(googleReadTimeout, cancel)
+	reader, err := o.handle.NewRangeReader(readCtx, off, length)
+	openTimer.Stop()
```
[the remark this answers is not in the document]
ValentaTomas: Valid race. Fixed in 393ddbe — when `openTimer.Stop()` returns false on a successful open, we close the reader and return an open-deadline error so the caller retries.

packages/shared/pkg/storage/storage_cache_seekable_compressed.go
```diff
@@ -29,23 +29,27 @@ func (c *cachedSeekable) openReaderCompressed(ctx context.Context, offsetU int64
...
+			}
 
-			return nil, fmt.Errorf("decompress cached frame: %w", err)
+			return withNFSGauge(ctx, decompressed), nil
 		}
-
-		return withNFSGauge(ctx, decompressed), nil
-	case !os.IsNotExist(err):
+		f.Close()
+		_ = os.Remove(path)
```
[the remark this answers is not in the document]
ValentaTomas: Valid. Fixed in latest force-push — only `os.Remove` on confirmed size mismatch; transient `Stat` errors leave the file in place and fall through to the miss path.

packages/shared/pkg/storage/storage_google.go
```diff
@@ -265,27 +265,55 @@ func (o *gcpObject) Size(ctx context.Context) (int64, error) {
...
+	readCtx, cancel := context.WithCancel(ctx)
 
-	reader, err := o.handle.NewRangeReader(ctx, off, length)
+	openTimer := time.AfterFunc(googleReadTimeout, cancel)
+	reader, err := o.handle.NewRangeReader(readCtx, off, length)
+	// Stop returning false means the timer fired and cancel was/is being called,
+	// so readCtx is cancelled and reader (if any) is unusable.
+	if !openTimer.Stop() && err == nil {
+		reader.Close()
+		err = errors.New("open deadline exceeded")
```
jakubno: it is on purpose we don't use context.DeadlineExceeded (wrapped)?

packages/shared/pkg/storage/storage_google.go
```diff
@@ -265,27 +265,55 @@ func (o *gcpObject) Size(ctx context.Context) (int64, error) {
...
+		ReadCloser: reader,
+		cancel:     cancel,
+		timer:      time.AfterFunc(googleReadTimeout, cancel),
+	}, nil
 }
 
-// cancelOnCloseReader wraps a ReadCloser and calls a CancelFunc on Close,
-// ensuring the context used to create the reader is cleaned up.
-type cancelOnCloseReader struct {
+type idleTimeoutReader struct {
```
jakubno, line 297:
```suggestion
// idleTimeoutReader fires cancel() after googleReadTimeout with no Read activity (in-flight Read with no progress, or no Read called)
type idleTimeoutReader struct {
```

jakubno (approved):

packages/shared/pkg/storage/storage_google.go
ValentaTomas: No, oversight — switched to `context.DeadlineExceeded` in 819c7a3.

packages/shared/pkg/storage/storage_google.go
ValentaTomas: Applied in 819c7a3.

PR #2808, opened 2026-05-23
refactor(sandbox): extract domain types into sandboxtypes leaf package

tvi:
Extract Sandbox, State, RemoveOpts, errors, Storage and
ReservationStorage interfaces, and related constants into a new
`sandboxtypes` leaf package. This breaks an import cycle that prevented
`store_test.go` from living in package `sandbox` alongside the code it
exercises.

Cycle being broken:

```text
sandbox (incl. _test.go files)
  -> sandbox/storage/memory       (used by tests as backing storage)
    -> sandbox                    (memory imports Sandbox, Storage, errors)
      ... CYCLE
````

Unlike the analogous fix for `nbd/testutils`, the back-edge here is not a
couple of helper functions but the entire memory (and redis) storage
implementations, which legitimately need the `Sandbox` struct, `State`
enum, error sentinels, and the `Storage` interface itself. Moving those
types into a leaf package is the cleanest break:

```text
sandboxtypes              (leaf: Sandbox, State, errors, Storage interface)
  ^         ^
  |         |
sandbox   sandbox/storage/{memory,redis,populate_redis}
sandbox   sandbox/reservations[/redis]
```

All arrows point at `sandboxtypes`; no cycle. Package `sandbox` keeps the
`Store` orchestrator, `Callbacks`, `CreationMetadata`, callback function
types, and now also re-exports every moved symbol via `aliases.go` so
external callers (handlers, orchestrator, utils, metrics, etc.)
continue to reference `sandbox.Sandbox`, `sandbox.ErrNotFound`,
`sandbox.StateRunning`, and so on without import churn.

### Files moved (renames)

```text
sandbox/sandbox.go  -> sandbox/sandboxtypes/sandbox.go
sandbox/states.go   -> sandbox/sandboxtypes/states.go
sandbox/errors.go   -> sandbox/sandboxtypes/errors.go
```

### New files

```text
sandbox/sandboxtypes/storage.go
  - Storage + ReservationStorage interfaces
  - StorageName* constants

sandbox/aliases.go
  - re-exports for backward compatibility
```

Storage backends and reservation packages now import `sandboxtypes`
directly. `store_test.go` is converted from `package sandbox_test` to
`package sandbox` now that the cycle is gone.

No behavior change. All sandbox tests pass; full `api` package builds
and vets cleanly.

### DIFF
--- a/packages/api/internal/sandbox/aliases.go
+++ b/packages/api/internal/sandbox/aliases.go
@@ -1,0 +1,56 @@
+package sandbox
+
+import "github.com/e2b-dev/infra/packages/api/internal/sandbox/sandboxtypes"
+
+// This file re-exports the sandbox domain types defined in the sandboxtypes
+// leaf package. The types live in sandboxtypes (rather than directly in
+// package sandbox) so storage backends like sandbox/storage/memory and
+// sandbox/storage/redis can implement the Storage interface without creating
+// an import cycle back into package sandbox. External callers can continue to
+// reference these symbols through sandbox.Sandbox, sandbox.State, etc.
+
+// Domain types.
+type (
+	Sandbox     = sandboxtypes.Sandbox
+	State       = sandboxtypes.State
+	StateAction = sandboxtypes.StateAction
+	RemoveOpts  = sandboxtypes.RemoveOpts
+
+	TransitionEffect = sandboxtypes.TransitionEffect
+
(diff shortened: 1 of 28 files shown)


ValentaTomas (approved):

ValentaTomas:
Just fix the lint.

PR #2812, opened 2026-05-25
feat(auth): allow http OIDC issuer on loopback hosts

tvi:
The OIDC verifier in packages/auth/pkg/auth/oidc previously enforced an https-only scheme on both the configured issuer URL and the discovery/JWKS URLs returned by it. This made it impossible to use a self-hosted Hydra running on http://localhost:4444/ as an OIDC issuer without standing up a TLS terminator in front of it.

Relax validateURL to accept http when the URL host is a loopback address: the literal name 'localhost', any IPv4 in 127.0.0.0/8, or the IPv6 loopback ::1. The check matches strings + parsed literal IPs only; we deliberately do not resolve DNS so the result stays pure and TOCTOU-free. Mirrors the same carve-out Kubernetes' apiserver applies to --oidc-issuer-url for in-cluster test deployments.

Any non-loopback host still requires https, preserving the production posture. Adds positive tests covering localhost, 127.0.0.1, and the IPv6 loopback; the existing non-https/non-loopback rejection test is unchanged.

### DIFF
--- a/packages/auth/pkg/auth/oidc/config_test.go
+++ b/packages/auth/pkg/auth/oidc/config_test.go
@@ -63,6 +63,36 @@ func TestEntry_Validate(t *testing.T) {
 			wantErr: "must be https",
 		},
 		{
+			// Loopback carve-out: http is allowed when the host is
+			// localhost. Used by the local devenv stack to point at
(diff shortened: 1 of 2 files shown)


ben-fornefeld (approved):
lgtm

PR #2814, opened 2026-05-25
Run dedup page write phase asynchronously

ValentaTomas:
Memfd dedup ran the page write phase synchronously on the Pause RPC, ignoring the bgCopy detachment. Split `dedupPages` into a sync compare and an async drain wrapped in `DedupedMemfdCache` (mirrors `MemfdCache`); reads block via `SetOnce[*Cache].WaitWithContext`, the goroutine owns and closes the memfd. The diff metadata/header still comes back synchronously so `AddSnapshot`/`NewUpload` keep working unchanged. Also coalesces contiguous dirty pages in `drainDirtyPages` (mirrors `Cache.copyProcessMemory`) and moves the best-effort `IsCached`/`uuid.Nil` check to per-page granularity so a single uncached neighbour no longer poisons cached pages of the same block. Integration matrix wires `TESTS_USE_MEMFD` so two of the three variants exercise the new async path end-to-end.

### DIFF
--- a/packages/orchestrator/pkg/sandbox/block/memfd.go
+++ b/packages/orchestrator/pkg/sandbox/block/memfd.go
@@ -6,6 +6,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
+	"time"
 
 	"github.com/RoaringBitmap/roaring/v2"
 	"go.opentelemetry.io/otel/attribute"
@@ -15,6 +17,7 @@ import (
 
 	"github.com/e2b-dev/infra/packages/shared/pkg/logger"
 	"github.com/e2b-dev/infra/packages/shared/pkg/storage/header"
+	"github.com/e2b-dev/infra/packages/shared/pkg/utils"
 )
 
 // Memfd wraps a memfd received from Firecracker. NewFromFd takes ownership of
@@ -121,11 +124,9 @@ func copyFromMemfd(ctx context.Context, cache *Cache, memfd *Memfd, dirty *roari
 // closes the memfd (releasing the hugetlb pages — potentially tens of GB)
 // and reads delegate to the embedded Cache.
 type MemfdCache struct {
-	cache *Cache
-
+	cache  *Cache
 	cancel context.CancelFunc
-	done   chan struct{} // closed by runCopy; happens-before for err
-	err    error
+	done   *utils.SetOnce[struct{}]
 }
 
 // NewCacheFromMemfdAsync starts the memfd→cache copy on a goroutine so
@@ -150,21 +151,19 @@ func NewCacheFromMemfdAsync(
 	if err != nil {
 		return nil, errors.Join(err, memfd.Close())
 	}
+	done := utils.NewSetOnce[struct{}]()
 	if dirty.IsEmpty() {
 		if closeErr := memfd.Close(); closeErr != nil {
 			return nil, errors.Join(fmt.Errorf("close memfd: %w", closeErr), cache.Close())
 		}
+		_ = done.SetValue(struct{}{})
 
-		return &MemfdCache{cache: cache}, nil
+		return &MemfdCache{cache: cache, done: done}, nil
 	}
 
 	// Detach from the request context so the copy outlives Pause.
 	copyCtx, cancel := context.WithCancel(context.WithoutCancel(ctx))
-	m := &MemfdCache{
-		cache:  cache,
-		cancel: cancel,
-		done:   make(chan struct{}),
-	}
+	m := &MemfdCache{cache: cache, cancel: cancel, done: done}
 
 	go m.runCopy(copyCtx, memfd, dirty, blockSize)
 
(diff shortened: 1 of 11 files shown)


packages/orchestrator/pkg/sandbox/block/memfd.go
```diff
@@ -249,18 +257,132 @@ func NewCacheFromMemfdDeduped(
...
+) {
+	ctx, span := tracer.Start(ctx, "dedup-drain")
+	defer span.End()
+
+	writeStart := time.Now()
+	cache, err := dedupDrain(ctx, src, pageDirty, blockSize, d.outPath, directIO)
+	writeDur := time.Since(writeStart)
+
+	if closeErr := memfd.Close(); closeErr != nil {
+		err = errors.Join(err, fmt.Errorf("close memfd: %w", closeErr))
```
jakubno: won't this cause all the in-flight operations calling Wait to fail? Can't it be problematic with e.g. upload snapshot? Before we just logged this

ValentaTomas: Good catch — reverted to a warn log, matching the previous sync behaviour. Drain errors still propagate through Wait but a memfd close failure no longer poisons in-flight readers.

packages/orchestrator/pkg/sandbox/block/memfd.go
```diff
@@ -254,18 +259,134 @@ func NewCacheFromMemfdDeduped(
...
-		logger.L().Warn(ctx, "close memfd after dedup", zap.Error(err))
+	compareDur := time.Since(compareStart)
+
+	// Snapshot counts before exposing the bitmaps via meta; pauseProcessMemory
+	// mutates Empty after we return, which would race with recordDedupAttrs.
+	totalPages := plan.exportedSize / header.PageSize
+	uniquePages := int64(plan.pageDirty.GetCardinality())
+	emptyPages := int64(plan.pageEmpty.GetCardinality())
+
+	meta := &header.DiffMetadata{
```
jakubno, line 268:
```suggestion
	// ownership of pageEmpty after this point transfers to the caller
	// runDrain / dedupDrain must only touch plan.pageDirty from here on.
	meta := &header.DiffMetadata{
```

packages/orchestrator/pkg/sandbox/block/memfd.go
```diff
@@ -254,18 +259,134 @@ func NewCacheFromMemfdDeduped(
 	dirty *roaring.Bitmap,
 	bestEffort bool,
 	directIO bool,
-) (*Cache, *header.DiffMetadata, error) {
+) (*DedupedMemfdCache, *header.DiffMetadata, error) {
+	ctx, span := tracer.Start(ctx, "dedup-pages",
```
jakubno: is this correct name for the span? btw there's already another one with the same name

packages/orchestrator/pkg/sandbox/block/memfd.go
```diff
@@ -254,18 +259,134 @@ func NewCacheFromMemfdDeduped(
...
+	blockSize int64,
+	directIO bool,
+	compareDur time.Duration,
+	totalPages, uniquePages, emptyPages int64,
+) {
+	ctx, span := tracer.Start(ctx, "dedup-drain")
+	defer span.End()
+
+	writeStart := time.Now()
+	cache, err := dedupDrain(ctx, src, pageDirty, blockSize, d.outPath, directIO)
```
jakubno, line 299:
```suggestion
	d.cache, d.err := dedupDrain(ctx, src, pageDirty, blockSize, d.outPath, directIO)
```

packages/orchestrator/pkg/sandbox/block/memfd.go
```diff
@@ -254,18 +259,134 @@ func NewCacheFromMemfdDeduped(
...
+	// Log only: a memfd close failure shouldn't fail in-flight Wait callers
+	// (upload, resume) — the data on disk is already complete.
+	if closeErr := memfd.Close(); closeErr != nil {
+		logger.L().Warn(ctx, "close memfd after dedup drain", zap.Error(closeErr))
+	}
+
+	recordDedupAttrs(ctx, totalPages, uniquePages, emptyPages, compareDur, writeDur)
+
+	d.cache = cache
+	d.err = err
```
jakubno:
```suggestion
```

jakubno (approved):

packages/orchestrator/pkg/sandbox/block/memfd.go
ValentaTomas: Adopted — folded the ownership note into the existing snapshot comment.

packages/orchestrator/pkg/sandbox/block/memfd.go
ValentaTomas: Renamed to `dedup-compare` to disambiguate from the sync `Cache.Dedup` span.

packages/orchestrator/pkg/sandbox/block/memfd.go
ValentaTomas: Done — assigning into `d.cache`/`d.err` directly removes the temporary vars.

packages/orchestrator/pkg/sandbox/block/memfd.go
ValentaTomas: Removed (folded into the direct assignment above).

PR #2823, opened 2026-05-26
perf(header): dictionary-encoded SoA Mappings to shrink cached headers

ValentaTomas:
## Summary

`Header.Mapping` was a flat `[]BuildMap` (40 bytes/entry — `uuid.UUID` is 16 bytes inline). Each `Header` lives in the orchestrator template cache for up to 25h, so on snapshot-heavy nodes the per-entry overhead dominates heap (~60% of orchestrator inuse on a recent staging pprof).

Replace it with a struct-of-arrays `Mappings` type that interns each distinct `BuildId` once in a small dictionary and references it by `uint16` index per entry. Per-entry footprint drops from 40 B to **26 B (~35%)**; the dictionary is a handful of UUIDs.

**Internal representation only.** Wire format (V3/V4), validation, merge/normalize semantics, and the `BuildMap` value type are all unchanged. Construction goes through `MappingsFromSlice`; the hot read path uses `At` / `OffsetAt` / `All` accessors. Serializers and CLI tools that needed the legacy `[]BuildMap` shape call `Mapping.Slice()` at the boundary.

Stacks cleanly on top of (or independent of) #2822 — that PR clipped `NormalizeMappings`' slack capacity; this one shrinks every entry that survives the clip.

## Bench

Per-entry retain on a 10k-mapping header (worst case for the old layout):

| layout | per-entry | 10k-entry header |
|---|---|---|
| before (`[]BuildMap`) | 40 B | ~400 KB |
| after (`Mappings`)    | 26 B | ~260 KB |

On a saturated cache (~45k cached headers × ~1k entries avg) this maps to roughly **~600 MB freed** on top of the clip PR.

## Bonus

`appendAncestorBuilds` previously scanned every mapping entry, de-duplicating BuildIds via a map. Now it walks `Mapping.Builds()` directly — already deduped, no allocation, regardless of mapping count.

## Test plan

- [x] `go test ./packages/shared/pkg/storage/header/...` (round-trip + existing serialization, merge, normalize tests pass)
- [x] `go test ./packages/orchestrator/pkg/sandbox/block/... ./packages/sandbox/build/...`
- [x] `go build ./...` for every module in go.work
- [x] new `Mappings` unit tests + construction benchmark

### DIFF
--- a/packages/orchestrator/cmd/copy-build/main.go
+++ b/packages/orchestrator/cmd/copy-build/main.go
@@ -131,8 +131,8 @@ func NewHeaderFromPath(ctx context.Context, from, headerPath string) (*header.He
 func getReferencedData(h *header.Header, objectType storage.ObjectType) []string {
 	builds := make(map[string]struct{})
 
-	for _, mapping := range h.Mapping {
-		builds[mapping.BuildId.String()] = struct{}{}
+	for _, id := range h.Mapping.Builds() {
+		builds[id.String()] = struct{}{}
 	}
 
 	delete(builds, uuid.Nil.String())
--- a/packages/orchestrator/cmd/inspect-build/main.go
+++ b/packages/orchestrator/cmd/inspect-build/main.go
@@ -105,7 +105,7 @@ func printUsage() {
 
 func printHeader(h *header.Header, source string) {
 	// Validate mappings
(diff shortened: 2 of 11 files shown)


ValentaTomas:
Closing for now, might revisit if we need to further improve the memory because of the mappings.

PR #2827, opened 2026-05-26
feat(storage): write-through compressed templates to NFS on upload

ValentaTomas:
Compressed template uploads skipped the NFS write-through path, so the first read on every node refetched from GCS. Tee each compressed frame to a `.frm` file at its C-space offset during upload, matching the layout `openReaderCompressed` expects.

Gated by the existing `write-to-cache-on-writes` flag.

### DIFF
--- a/packages/shared/pkg/storage/compress_upload.go
+++ b/packages/shared/pkg/storage/compress_upload.go
@@ -147,7 +147,7 @@ func compressStream(ctx context.Context, in io.Reader, cfg CompressConfig, uploa
 			frameSizes = append(frameSizes, FrameSize{U: int32(f.uncompressedSize), C: int32(len(f.compressed))})
 			compressed = append(compressed, f.compressed)
 			if sink != nil {
-				sink(cOffset, f.compressed)
+				sink(ctx, cOffset, f.compressed)
 			}
 			cOffset += int64(len(f.compressed))
 		}
--- a/packages/shared/pkg/storage/compress_upload.go
+++ b/packages/shared/pkg/storage/compress_upload.go
@@ -104,7 +104,7 @@ func (p *part) addFrame(ctx context.Context, uncompressedData []byte, pool *sync
 	})
 }
 
-func compressStream(ctx context.Context, in io.Reader, cfg CompressConfig, uploader partUploader, maxUploadConcurrency int) (*FrameTable, [32]byte, error) {
+func compressStream(ctx context.Context, in io.Reader, cfg CompressConfig, uploader partUploader, maxUploadConcurrency int, sink FrameSink) (*FrameTable, [32]byte, error) {
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
@@ -132,6 +132,7 @@ func compressStream(ctx context.Context, in io.Reader, cfg CompressConfig, uploa
 
 	// Upload loop.
 	var frameSizes []FrameSize
+	var cOffset int64
 	var loopErr error
 	for p := range q {
 		if err := p.compress.Wait(); err != nil {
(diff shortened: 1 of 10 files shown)


packages/shared/pkg/storage/compress_upload.go
```diff
@@ -145,6 +146,10 @@ func compressStream(ctx context.Context, in io.Reader, cfg CompressConfig, uploa
 		for _, f := range p.frames {
 			frameSizes = append(frameSizes, FrameSize{U: int32(f.uncompressedSize), C: int32(len(f.compressed))})
 			compressed = append(compressed, f.compressed)
+			if sink != nil {
+				sink(cOffset, f.compressed)
+			}
```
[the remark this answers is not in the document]
levb: This would be fine as long as the FRM files themselves are complete; if unused - they'll be garbage collected.

levb (approved):
LGTM

packages/shared/pkg/storage/storage_cache_seekable.go
```diff
@@ -336,6 +338,45 @@ func (c *cachedSeekable) StoreFile(ctx context.Context, path string, opts ...Put
...
+		// large uploads with a low concurrency cap pile up thousands of live
+		// goroutines in wg. Pushes backpressure into the sink caller (the
+		// upload loop in compressStream), which is the right place to slow.
+		select {
+		case sem <- struct{}{}:
+		case <-ctx.Done():
+			return
+		}
+
+		c.goCtx(ctx, func(ctx context.Context) {
```
dobrac, line 362: why does this goroutine use the WithoutCancel context?

packages/shared/pkg/storage/storageopts/storageopts.go
```diff
@@ -2,21 +2,33 @@
...
 
 type ObjectMetadata map[string]string
 
 const ObjectMetadataTeamID = "team_id"
 
+// FrameSink fires once per compressed frame produced by a compressed
+// StoreFile, with the frame's absolute C-space offset and bytes. Best-effort
+// and expected to return quickly — implementations should schedule any I/O
+// asynchronously and bound their own concurrency.
+type FrameSink func(ctx context.Context, cOffset int64, compressed []byte)
```
dobrac, line 16: can we make this interface instead?

dobrac (dismissed):

packages/shared/pkg/storage/storage_cache_seekable.go
levb: It outlives the upload cancel, but there is nothing explicit that kills it. It's been like this since before compression, see https://github.com/e2b-dev/infra/blob/758e726c7f2883434882affbafd42dcc201bc734/packages/shared/pkg/storage/storage_cache_seekable.go#L307

packages/shared/pkg/storage/storageopts/storageopts.go
levb: 2/5 it should stay a function ptr - simple contract, easy to use, no other methods expected. Also, if we ever needed to add a method, there is only 1 code path that uses this, would be easy to change then?

packages/shared/pkg/storage/storage_cache_seekable.go
ValentaTomas: Intentional, and pre-existing for the uncompressed write-through path: the cache write is detached from the upload ctx so it isn't aborted mid-write when the upload is cancelled (it still inherits values for tracing/flags). It's tracked by c.wg so it can be awaited. Added a comment on goCtx documenting this.

packages/shared/pkg/storage/storageopts/storageopts.go
ValentaTomas: Keeping it a function type per Lev's reasoning: it's a single-method, single-call-site contract, and a func ptr is the simplest thing that works. Easy to promote to an interface later if a second method/impl shows up.

levb (approved):

PR #2831, opened 2026-05-27
feat(storage): rationalize read path OTEL

levb:
Adds to https://github.com/e2b-dev/infra/pull/2570 which should be merged first.

Adds an orchestrator.read.* metric family with consistent attributes
(file_type/source/codec/outcome) covering each stage of a read, plus
per-layer chunker and build-file timers, so dashboards can attribute
latency end-to-end from sandbox-visible read to backend fetch.

Does not remove any of the prior metrics, this will be done separately
after the dashboards are updated.

Metrics:
  - `orchestrator.file.read_at` build.File.ReadAt - per-fault unit, aggregates all underlying mappings into one record
  - `orchestrator.chunk.slice` Chunker.Slice - per per-mapping unit, source=mmap on cache hit else the backend that served
  - `orchestrator.read.open` - OpenRangeReader (open / TTFB)
  - `orchestrator.read.read` - source-read wall time, raw bytes
  - `orchestrator.read.decompress` - decompress wall time + uncompressed bytes
  - `orchestrator.read.fetch` - total fetch wall time + uncompressed bytes delivered
  - `orchestrator.read.writeback` - NFS cache writeback wall + bytes
  - `orchestrator.read.pipeline.efficiency` - fetch / (open+read+decompress)
  - `orchestrator.read.cache` - NFS hit/miss/writeback events
  - `orchestrator.read.inflight` - concurrent fetches gauge

Spans:
  - `chunk.fetch` - runFetch goroutine span


packages/shared/pkg/storage/compress_decode.go
```diff
@@ -86,17 +105,44 @@ func NewDecompressingReader(inner RangeReader, ct CompressionType) (RangeReader,
...
+	return n, err
 }
 
-func (r *decompressReader) Close(ctx context.Context) error {
+func (r *decompressReader) Close(ctx context.Context) (*ReadStats, error) {
+	// Drain to EOF so any wrapper below observes the full encoded stream. With
+	// LZ4 BlockChecksum=true / Checksum=false the 4-byte EndMark is otherwise
+	// left unread.
+	if r.readErr == nil {
+		_, _ = io.Copy(io.Discard, r.meteredOut)
```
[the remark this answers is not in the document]
levb: killed draining instead, we don't really care - it only may come into play for the last 4 bytes of LZ4 stream that we don't care about anyway.

packages/shared/pkg/telemetry/meters.go
```diff
@@ -548,6 +548,48 @@ func NewTimerFactory(
...
+// ambiguous. The histogram already emits a _count series, so no separate
+// event-count instrument is needed.
+type FloatTimerFactory struct {
+	duration metric.Float64Histogram
+	bytes    metric.Int64Counter
+}
+
+func NewFloatTimerFactory(
+	meter metric.Meter,
+	metricName, durationDescription, bytesDescription, _ string,
```
djeebus: can we remove the last unused argument? do we need to maintain it for some interface?

packages/shared/pkg/storage/read_metrics.go
```diff
@@ -0,0 +1,165 @@
...
+	readPipelineEfficiency.Record(ctx, ratio, attrs)
+}
+
+// StartInflight increments the read.inflight gauge and returns a func that
+// decrements it; defer the returned func so the +1/-1 can't drift apart.
+func StartInflight(ctx context.Context, attrs metric.MeasurementOption) func() {
+	readInflight.Add(ctx, 1, attrs)
+
+	return func() { readInflight.Add(ctx, -1, attrs) }
+}
```
djeebus, line 97: I'll just add the traditional warning here: gauges are rarely the right answer, and often misleading. It might be safer to record "reads begun" and "reads finished" as counters. Alternately, recording both would be reasonable.

TL;DR: gauges only show you a point-in-time every n seconds (30 or 60 by default, depending on system), no way to judge what happened in between.

As an example of what could be hidden by a gauge:
- a process increments and decrements that gauge 1000 times in 5 seconds, in between two readings. The gauge reports 0 at each moment in time, no indication of the flurry of activity.
- 50 reads are started/stopped every second. The gauge shows "50 concurrent reads" but what really happened was that there were 50 reads * 30 seconds = 1500 reads total; no way to differentiate that from "50 sustained reads over 30 seconds"

You can *nearly* always the gauge data from a counter; you can't get the counter data from the gauge. If point-in-time concurrency is valuable, let's keep it; otherwise `increase(total_read_count[$__interval])` tells us "how many reads were started over this period, averaged per second"; not quite the same thing, but similar in terms of diagnostic value.

PR #2832, opened 2026-05-27
Add ClickHouse metric log toggle

mihir-ephemeral:
## What changed

Adds an `enable_metric_logs` input to `iac/modules/job-clickhouse`, defaulting to `true` so existing consumers keep the current behavior.

When set to `false`, the generated ClickHouse config disables:

- `system.metric_log`
- `system.asynchronous_metric_log`

## Why

Some deployments already export ClickHouse metrics through Prometheus/OTel and do not need ClickHouse to persist duplicate internal metric history tables. Disabling these logs avoids background MergeTree merges for those internal tables on smaller/self-hosted deployments.

## Validation

- `git diff --check`
- Did not run Terraform.

### DIFF
--- a/iac/modules/job-clickhouse/configs/config.xml
+++ b/iac/modules/job-clickhouse/configs/config.xml
@@ -47,9 +47,13 @@
 
     <listen_host>0.0.0.0</listen_host>
 
+    %{ if enable_metric_logs }
     <asynchronous_metric_log>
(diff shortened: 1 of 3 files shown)


mihir-ephemeral:
Closing this because we cannot use the upstream contribution path here. We'll keep the ClickHouse module toggle as an internal EphemeralTechnologies fork patch instead.

PR #2834, opened 2026-05-28
feat(orchestrator): add ublk block device provider as alternative to NBD

harmlessii:
## Summary

Add ublk-based block device provider as an alternative to NBD for sandbox rootfs.

## Motivation
- On high-core-count machines (hundreds of cores), NBD device connection takes hundreds of milliseconds. In contrast, ublk device creation takes only ~5ms on the same machine.
- ublk eliminates the user↔kernel context switches that NBD requires per IO — NBD routes every IO through a userspace daemon via sockets (2 context switches + 2 copies per IO), while ublk uses zero-copy io_uring queues.
- No need to pre-allocate a static device pool (kernel assigns device IDs dynamically).

## Changes

- `rootfs/ublk.go` — UblkProvider implementing the Provider interface
- `rootfs/ublk_backend.go` — Minimal adapter binding context to ublk-go callbacks
- `rootfs/ublk_test.go` — Integration tests matching NBD's path_direct_test.go coverage
- `ublk/pool.go` — Device pool with OTel metrics and structured logging

## Prerequisites

- Linux kernel 6.0+ with `ublk_drv` module loaded (`modprobe ublk_drv`)
- Dependency: github.com/e2b-dev/ublk-go

## Testing

All 6 integration tests pass (requires root + ublk_drv):
- Write, WriteAtOffset, Direct4MB, Direct32MB, LargeWrite (1GB dd), LargeRead (1GB dd)

### DIFF
--- a/packages/orchestrator/go.mod
+++ b/packages/orchestrator/go.mod
@@ -30,6 +30,7 @@ require (
 	github.com/dustin/go-humanize v1.0.1
 	github.com/e2b-dev/infra/packages/clickhouse v0.0.0
 	github.com/e2b-dev/infra/packages/shared v0.0.0
+	github.com/e2b-dev/ublk-go v0.1.3
 	github.com/edsrzf/mmap-go v1.2.1-0.20241212181136-fad1cd13edbd
 	github.com/firecracker-microvm/firecracker-go-sdk v1.0.0
 	github.com/getkin/kin-openapi v0.137.0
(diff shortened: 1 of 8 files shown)


harmlessii:
@cla-bot check

jakubno:
Hey @harmlessii , thanks for your PR, we're inclined to add `ublk` support. But we've decided it'll be done by one of our core maintainers

PR #2842, opened 2026-05-28
fix(otel): Use GCP native prometheus exporter so metrics don't get subset under a subkey when queried using prometheus API in Grafana

wj-e2b:


iac/modules/job-otel-collector-nomad-server/configs/otel-collector-nomad-server.yaml
```diff
@@ -145,6 +145,31 @@ processors:
           enabled: true
 %{ endif }
 
+%{ if provider_name == "gcp" && enable_gcp_telemetry_metrics }
```
wj-e2b, line 148: These are reserved names in GCP

dobrac (approved):

PR #2843, opened 2026-05-28
perf(header): drop oversized temp set in newDiffHeader

ValentaTomas:
On snapshot-heavy nodes the temp set in `newDiffHeader` (sized to `len(h.Mapping)` to dedupe build IDs before populating `h.Builds`) dominates live heap: each in-flight `ToDiffHeader` keeps one alive for the duration of the call.

Heap profile from a node right before OOM (126 GB total): `newDiffHeader` alone holds ~20 GB (~16%) — over-sized buckets that GC can't reclaim while many concurrent Pauses are running.

Walk mappings once and dedupe via a set sized to the (small) `sourceBuilds` map instead. Final `h.Builds` is identical.

### DIFF
--- a/packages/shared/pkg/storage/header/header.go
+++ b/packages/shared/pkg/storage/header/header.go
@@ -102,15 +102,20 @@ func newDiffHeader(metadata *Metadata, mapping []BuildMap, sourceBuilds map[uuid
 	}
 
 	if sourceBuilds != nil {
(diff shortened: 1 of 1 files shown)


ValentaTomas:
Superseded by #2844, which contains an equivalent newDiffHeader fix (iterating the deduplicated Mapping.Builds()) as part of the compact-mapping change.

PR #2844, opened 2026-05-28
perf(header): pack cached Header.Mapping into a compact form

ValentaTomas:
Replace the cached `Header.Mapping` (`[]BuildMap`, 40 B/entry) with a struct-of-arrays compact form (~14 B/entry): offset/length/storage as uint32 page-block indices plus a deduplicated per-header build-ID table. On snapshot-heavy nodes the retained mapping slices dominate host RAM (heap profiles show ~80% inuse in `NormalizeMappings`). Mappings are encoded in PageSize units since a 2 MiB-hugepage memory file can carry page-granular diff mappings. In-memory only; the V3/V4 wire format is unchanged.

### DIFF
--- a/packages/shared/pkg/storage/header/compact.go
+++ b/packages/shared/pkg/storage/header/compact.go
@@ -1,0 +1,192 @@
+package header
+
+import (
+	"errors"
+	"fmt"
+	"iter"
+
+	"github.com/google/uuid"
+)
+
+// Mapping is the compact in-memory representation of a Header's mapping list.
+// A merged Header is cached for up to 25h, so on snapshot-heavy nodes these
+// slices dominate host RAM. It shrinks each entry from 40 bytes (a BuildMap) to
+// 14 by encoding offset/length/storage as uint32 block indices, deduplicating
+// BuildId into a per-header table addressed by uint16, and storing the columns
+// as parallel slices. Immutable; read via At / All / Slice.
+type Mapping struct {
+	blockSize uint64
+	builds    []uuid.UUID
+	offsets   []uint32
+	lengths   []uint32
+	storage   []uint32
+	buildIdx  []uint16
+}
+
+// maxBuildsPerHeader caps the per-header build-ID table at the uint16 limit.
+const maxBuildsPerHeader = 1 << 16
+
+// NewMapping packs src into the compact representation. blockSize is the unit
+// for the block indices and must divide every Offset, Length, and
+// BuildStorageOffset in src (callers pass PageSize, the universal granularity).
+func NewMapping(blockSize uint64, src []BuildMap) (Mapping, error) {
+	if blockSize == 0 {
+		return Mapping{}, errors.New("compact mapping: block size cannot be zero")
+	}
+	if len(src) == 0 {
+		return Mapping{blockSize: blockSize}, nil
+	}
+
+	// uint32 page-block index caps a single file at ~16 TiB (PageSize units).
+	const maxBlockIdx = (1 << 32) - 1
+
+	idxByBuild := make(map[uuid.UUID]uint16, 8)
+	builds := make([]uuid.UUID, 0, 8)
+	offsets := make([]uint32, len(src))
+	lengths := make([]uint32, len(src))
+	storage := make([]uint32, len(src))
+	buildIdx := make([]uint16, len(src))
+
+	for i, m := range src {
+		if m.Offset%blockSize != 0 {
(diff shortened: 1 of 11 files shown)


packages/shared/pkg/storage/header/compact.go
```diff
@@ -0,0 +1,200 @@
...
+
+// SearchOffset returns the first index i whose byte offset is strictly greater
+// than off, matching sort.Search semantics on Offset. It compares in block
+// units, so entries are never materialized: entry.OffsetBlocks*blockSize > off
+// iff entry.OffsetBlocks > off/blockSize (integer division).
+func (m Mapping) SearchOffset(off int64) int {
+	if off < 0 {
+		return 0
+	}
+	target := uint64(off) / m.blockSize
```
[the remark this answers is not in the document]
ValentaTomas: Fixed: SearchOffset now returns 0 for an empty mapping before the division, so a zero-value Mapping cannot panic.

packages/shared/pkg/storage/header/header.go
```diff
@@ -81,17 +83,33 @@ func NewHeader(metadata *Metadata, mapping []BuildMap) (*Header, error) {
...
+		// page-aligned invariant the compact form depends on. In practice
+		// production paths always supply mapping explicitly.
+		length := metadata.Size
+		if rem := length % PageSize; rem != 0 {
+			length += PageSize - rem
+		}
 		mapping = []BuildMap{{
 			Offset:             0,
-			Length:             metadata.Size,
+			Length:             length,
```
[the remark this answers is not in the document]
ValentaTomas: Intended: the synthesized fallback (only when no mapping is supplied) rounds Length up to PageSize so it satisfies the compact form alignment; ValidateHeader tolerates the sub-block overhang.

packages/shared/pkg/storage/header/header.go
```diff
@@ -81,17 +83,33 @@ func NewHeader(metadata *Metadata, mapping []BuildMap) (*Header, error) {
...
 			BuildId:            metadata.BuildId,
 			BuildStorageOffset: 0,
 		}}
 	}
 
+	// Encode in PageSize units, not metadata.BlockSize: a file's logical block
+	// size can be larger than its mapping granularity (e.g. a 2 MiB-hugepage
+	// memory file whose diff mappings are page-granular after dedup). PageSize
+	// is the universal floor — every mapping is page-aligned.
+	compact, err := NewMapping(PageSize, mapping)
```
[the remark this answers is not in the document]
ValentaTomas: Safe by invariant: Offset, Length, and BuildStorageOffset are always block-aligned in real headers (V3 included). CreateMapping derives BuildStorageOffset as a cumulative sum of block-aligned Lengths, and the diff block size is always >= PageSize, so every stored offset is a PageSize multiple. The old 123 fixture was a synthetic value that cannot occur in production. The compact form fundamentally requires page-alignment, which the data always satisfies, so no existing snapshot regresses.

packages/shared/pkg/storage/header/header.go
```diff
@@ -81,17 +83,33 @@ func NewHeader(metadata *Metadata, mapping []BuildMap) (*Header, error) {
 	}
 
 	if len(mapping) == 0 {
+		// Round up to PageSize so the synthesized fallback satisfies the
```
levb: The comment is a little complex to parse for me, the code would be immediately clear if used the standard "round up" expression `length := (metadata.Size + PageSize - 1) / PageSize * PageSize`

packages/shared/pkg/storage/header/header.go
```diff
@@ -81,17 +83,33 @@ func NewHeader(metadata *Metadata, mapping []BuildMap) (*Header, error) {
...
 		mapping = []BuildMap{{
 			Offset:             0,
-			Length:             metadata.Size,
+			Length:             length,
 			BuildId:            metadata.BuildId,
 			BuildStorageOffset: 0,
 		}}
 	}
 
+	// Encode in PageSize units, not metadata.BlockSize: a file's logical block
```
levb: i found this comment confusing. Maybe trim?

levb (approved):
Comments/nits only, LGTM otherwise

packages/shared/pkg/storage/header/header.go
ValentaTomas: Fixed in ccf634a71.

packages/shared/pkg/storage/header/header.go
ValentaTomas: Fixed in ccf634a71.

PR #2845, opened 2026-05-28
perf(infra): slightly reduce client base hugepage reservation

ValentaTomas:
Lower the client base hugepage default from 80% to 78% so less host RAM is permanently locked as base hugepages and a bit more remains available for orchestrator heap headroom. This only changes the default; clusters can still override `hugepages_percentage`.


ValentaTomas:
We will modify via env vars.

PR #2846, opened 2026-05-28
feat(observability): include kill_reason in kill-path structured logs

rguliyev:
`kill_reason` already rides on the webhook event, the ClickHouse sandbox_events row, and the orchestrator.sandbox.killed counter. Adds it to the kill-path structured logs too, so log-only investigation in Loki has the same attribution.

### DIFF
--- a/packages/api/internal/orchestrator/delete_instance.go
+++ b/packages/api/internal/orchestrator/delete_instance.go
@@ -34,18 +34,28 @@ func (o *Orchestrator) RemoveSandbox(ctx context.Context, teamID uuid.UUID, sand
 		switch opts.Action {
 		case sandbox.StateActionKill:
 			if errors.Is(err, sandbox.ErrNotFound) {
-				logger.L().Info(ctx, "Sandbox not found, already removed", logger.WithSandboxID(sandboxID))
+				logger.L().Info(ctx, "Sandbox not found, already removed",
+					logger.WithSandboxID(sandboxID),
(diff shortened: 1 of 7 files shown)


packages/api/internal/orchestrator/delete_instance.go
```diff
@@ -96,11 +106,16 @@ func (o *Orchestrator) RemoveSandbox(ctx context.Context, teamID uuid.UUID, sand
...
 	err = o.removeSandboxFromNode(ctx, sbx, opts.Action, opts.Reason)
 	if err != nil {
-		logger.L().Error(ctx, "Error removing sandbox",
+		fields := []zap.Field{
 			zap.String("state_action", opts.Action.Name),
 			zap.Error(err),
 			logger.WithSandboxID(sbx.SandboxID),
-		)
+		}
+		if opts.Action == sandbox.StateActionKill {
```
dobrac, line 114: can we add it always? (same in other places)

dobrac (approved):

packages/api/internal/orchestrator/delete_instance.go
rguliyev: Pause callers (sandbox_pause, evictor's auto-pause branch) leave Reason
empty, and KillReason.String() normalizes empty to "unknown". Without the
gate, every pause failure logs `state_action=pause kill_reason=unknown`,
which reads as an unattributed kill in Loki when it's actually a failed
pause. The gate keeps kill_reason scoped to kill outcomes.

PR #2847, opened 2026-05-28
perf(header): add V5 columnar varint header format

ValentaTomas:
Stacked on #2844. Adds a V5 header format: the mapping section is columnar and varint-coded instead of V4's fixed 40-byte records. Page-granular memfile dedup produces million-entry headers that overflow the cap and compress poorly; V5 keeps the same semantics while making those headers much smaller. Readers understand V5 unconditionally; writes are gated behind `HeaderV5WriteFlag` (default off) for a read-before-write rollout.

### DIFF
--- a/packages/orchestrator/pkg/sandbox/build_upload.go
+++ b/packages/orchestrator/pkg/sandbox/build_upload.go
@@ -52,6 +52,9 @@ func NewUpload(
 	}
 
 	headerVersion := uint64(headers.MetadataVersionV4)
+	if ff != nil {
+		ctx = featureflags.AddToContext(ctx, featureflags.CompressUseCaseContext(useCase))
+	}
 	if ff != nil && ff.BoolFlag(ctx, featureflags.HeaderV5WriteFlag) {
 		headerVersion = headers.MetadataVersionV5
 	}
--- a/packages/orchestrator/pkg/sandbox/build_upload.go
+++ b/packages/orchestrator/pkg/sandbox/build_upload.go
@@ -51,6 +51,9 @@ func NewUpload(
 		return nil, fmt.Errorf("resolve rootfs compress config: %w", err)
 	}
 
+	if useCase != "" {
+		ctx = featureflags.AddToContext(ctx, featureflags.CompressUseCaseContext(useCase))
+	}
 	headerVersion := uint64(headers.MetadataVersionV4)
 	if ff != nil {
 		ctx = featureflags.AddToContext(ctx, featureflags.CompressUseCaseContext(useCase))
--- a/packages/orchestrator/pkg/sandbox/build_upload.go
+++ b/packages/orchestrator/pkg/sandbox/build_upload.go
@@ -55,9 +55,6 @@ func NewUpload(
 		ctx = featureflags.AddToContext(ctx, featureflags.CompressUseCaseContext(useCase))
 	}
 	headerVersion := uint64(headers.MetadataVersionV4)
-	if ff != nil {
-		ctx = featureflags.AddToContext(ctx, featureflags.CompressUseCaseContext(useCase))
-	}
 	if ff != nil && ff.BoolFlag(ctx, featureflags.HeaderV5WriteFlag) {
 		headerVersion = headers.MetadataVersionV5
 	}
--- a/packages/orchestrator/pkg/sandbox/build_upload.go
+++ b/packages/orchestrator/pkg/sandbox/build_upload.go
@@ -66,7 +66,7 @@ func NewUpload(
 		root:           root,
 		useCase:        useCase,
 		objectMetadata: objectMetadata,
-		useV4:          memV4 || rootV4,
+		useV4:          memV4 || rootV4 || headerVersion == headers.MetadataVersionV5,
 		headerVersion:  headerVersion,
 	}
 
--- a/packages/orchestrator/pkg/sandbox/build_upload.go
+++ b/packages/orchestrator/pkg/sandbox/build_upload.go
@@ -29,6 +29,7 @@ type Upload struct {
 	objectMetadata storage.ObjectMetadata
 	future         *utils.ErrorOnce
 	useV4          bool
+	headerVersion  uint64
 }
 
 func NewUpload(
@@ -50,6 +51,11 @@ func NewUpload(
 		return nil, fmt.Errorf("resolve rootfs compress config: %w", err)
 	}
 
+	headerVersion := uint64(headers.MetadataVersionV4)
+	if ff != nil && ff.BoolFlag(ctx, featureflags.HeaderV5WriteFlag) {
+		headerVersion = headers.MetadataVersionV5
+	}
+
 	u := &Upload{
 		buildID:        snap.BuildID,
 		snap:           snap,
@@ -61,6 +67,7 @@ func NewUpload(
 		useCase:        useCase,
 		objectMetadata: objectMetadata,
 		useV4:          memV4 || rootV4,
+		headerVersion:  headerVersion,
 	}
 
 	if uploads != nil {
--- a/packages/shared/pkg/featureflags/flags.go
+++ b/packages/shared/pkg/featureflags/flags.go
@@ -167,6 +167,13 @@ var (
 	// uploads. Independent of compress-config: it changes the header format,
 	// not whether data is compressed.
 	V4HeaderForUncompressedFlag = NewBoolFlag("v4-header-for-uncompressed", false)
+
+	// HeaderV5WriteFlag makes Pause emit the V5 (columnar, varint) header
+	// layout instead of V4. Readers understand V5 regardless of this flag;
+	// keep it off until every orchestrator that may read these headers (incl.
+	// P2P peers) ships V5 read support, then enable it to shrink the large
+	// fragmented headers produced by page-granular memfile dedup.
+	HeaderV5WriteFlag = NewBoolFlag("header-v5-write", false)
 )
 
 type IntFlag struct {
--- a/packages/shared/pkg/featureflags/flags.go
+++ b/packages/shared/pkg/featureflags/flags.go
@@ -168,11 +168,8 @@ var (
 	// not whether data is compressed.
 	V4HeaderForUncompressedFlag = NewBoolFlag("v4-header-for-uncompressed", false)
 
-	// HeaderV5WriteFlag makes Pause emit the V5 (columnar, varint) header
-	// layout instead of V4. Readers understand V5 regardless of this flag;
-	// keep it off until every orchestrator that may read these headers (incl.
-	// P2P peers) ships V5 read support, then enable it to shrink the large
-	// fragmented headers produced by page-granular memfile dedup.
+	// HeaderV5WriteFlag makes Pause emit V5 headers. When enabled it also
+	// supersedes V4HeaderForUncompressedFlag for uncompressed uploads.
 	HeaderV5WriteFlag = NewBoolFlag("header-v5-write", false)
 )
 
--- a/packages/shared/pkg/storage/header/compact.go
+++ b/packages/shared/pkg/storage/header/compact.go
@@ -24,8 +24,12 @@ type Mapping struct {
 	buildIdx  []uint16
 }
 
-// maxBuildsPerHeader caps the per-header build-ID table at the uint16 limit.
-const maxBuildsPerHeader = 1 << 16
+const (
+	nilBuildIdx = math.MaxUint16
+
+	// maxBuildsPerHeader leaves nilBuildIdx reserved for empty regions.
+	maxBuildsPerHeader = nilBuildIdx
+)
 
 // maxBlockIdx is the largest block index representable by the uint32 columns.
 // At PageSize granularity this caps a single file (memfile/rootfs) at ~16 TiB,
@@ -68,14 +72,18 @@ func NewMapping(blockSize uint64, src []BuildMap) (Mapping, error) {
 			return Mapping{}, fmt.Errorf("compact mapping: block index out of uint32 range at entry %d", i)
 		}
 
-		idx, ok := idxByBuild[m.BuildId]
-		if !ok {
-			if len(builds) >= maxBuildsPerHeader {
-				return Mapping{}, fmt.Errorf("compact mapping: more than %d unique build IDs", maxBuildsPerHeader)
+		idx := uint16(nilBuildIdx)
+		if m.BuildId != uuid.Nil {
+			var ok bool
+			idx, ok = idxByBuild[m.BuildId]
+			if !ok {
+				if len(builds) >= maxBuildsPerHeader {
+					return Mapping{}, fmt.Errorf("compact mapping: more than %d unique build IDs", maxBuildsPerHeader)
+				}
+				idx = uint16(len(builds))
+				idxByBuild[m.BuildId] = idx
+				builds = append(builds, m.BuildId)
 			}
-			idx = uint16(len(builds))
-			idxByBuild[m.BuildId] = idx
-			builds = append(builds, m.BuildId)
 		}
 
 		offsets[i] = uint32(offBlocks)
@@ -103,6 +111,9 @@ func newMappingFromColumns(blockSize uint64, builds []uuid.UUID, offsets, length
 		return Mapping{}, fmt.Errorf("compact mapping: column length mismatch (offsets=%d lengths=%d storage=%d buildIdx=%d)", n, len(lengths), len(storage), len(buildIdx))
 	}
 	for i, bi := range buildIdx {
+		if bi == nilBuildIdx {
+			continue
+		}
 		if int(bi) >= len(builds) {
 			return Mapping{}, fmt.Errorf("compact mapping: buildIdx %d at entry %d out of range (%d builds)", bi, i, len(builds))
 		}
@@ -131,10 +142,15 @@ func (m Mapping) Builds() []uuid.UUID { return m.builds }
 // At materializes the i-th entry as a BuildMap. Panics if i is out of range,
 // matching `mapping[i]` semantics.
 func (m Mapping) At(i int) BuildMap {
+	buildID := uuid.Nil
+	if m.buildIdx[i] != nilBuildIdx {
+		buildID = m.builds[m.buildIdx[i]]
+	}
+
 	return BuildMap{
 		Offset:             uint64(m.offsets[i]) * m.blockSize,
 		Length:             uint64(m.lengths[i]) * m.blockSize,
-		BuildId:            m.builds[m.buildIdx[i]],
+		BuildId:            buildID,
 		BuildStorageOffset: uint64(m.storage[i]) * m.blockSize,
 	}
 }
--- a/packages/shared/pkg/storage/header/compact.go
+++ b/packages/shared/pkg/storage/header/compact.go
@@ -27,6 +27,11 @@ type Mapping struct {
 // maxBuildsPerHeader caps the per-header build-ID table at the uint16 limit.
 const maxBuildsPerHeader = 1 << 16
 
+// maxBlockIdx is the largest block index representable by the uint32 columns.
+// At PageSize granularity this caps a single file (memfile/rootfs) at ~16 TiB,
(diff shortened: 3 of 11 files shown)


packages/shared/pkg/storage/header/serialization_v5.go
```diff
@@ -0,0 +1,235 @@
...
+	var n, m uint32
+	if err := binary.Read(reader, binary.LittleEndian, &n); err != nil {
+		return Mapping{}, fmt.Errorf("failed to read mapping count: %w", err)
+	}
+	if err := binary.Read(reader, binary.LittleEndian, &m); err != nil {
+		return Mapping{}, fmt.Errorf("failed to read mapping build count: %w", err)
+	}
+	if m > maxBuildsPerHeader {
+		return Mapping{}, fmt.Errorf("mapping build count %d exceeds maximum %d", m, maxBuildsPerHeader)
+	}
```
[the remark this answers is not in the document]
ValentaTomas: Addressed in the latest push.

packages/shared/pkg/storage/header/serialization_v5.go
```diff
@@ -0,0 +1,235 @@
...
+	for i := range offsets {
+		d, err := binary.ReadUvarint(reader)
+		if err != nil {
+			return Mapping{}, fmt.Errorf("failed to read offset delta %d: %w", i, err)
+		}
+		prevOffset += d
+		if prevOffset > maxBlockIdx {
+			return Mapping{}, fmt.Errorf("offset block %d at entry %d exceeds uint32", prevOffset, i)
+		}
+		offsets[i] = uint32(prevOffset)
```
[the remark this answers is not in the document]
ValentaTomas: Addressed in the latest push.

packages/orchestrator/pkg/sandbox/build_upload.go
```diff
@@ -50,6 +51,11 @@ func NewUpload(
 		return nil, fmt.Errorf("resolve rootfs compress config: %w", err)
 	}
 
+	headerVersion := uint64(headers.MetadataVersionV4)
+	if ff.BoolFlag(ctx, featureflags.HeaderV5WriteFlag) {
+		headerVersion = headers.MetadataVersionV5
+	}
```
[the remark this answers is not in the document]
ValentaTomas: Addressed in the latest push.

packages/shared/pkg/storage/header/serialization_v5.go
```diff
@@ -0,0 +1,256 @@
...
+			return Mapping{}, fmt.Errorf("failed to read mapping build id %d: %w", i, err)
+		}
+	}
+
+	if n == 0 {
+		return newMappingFromColumns(blockSize, builds, nil, nil, nil, nil)
+	}
+	// Each entry needs at least one varint byte per column, so n cannot exceed
+	// the remaining bytes. Bound the allocation against a crafted count.
+	if int(n) > reader.Len() {
```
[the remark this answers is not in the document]
ValentaTomas: Fixed in 7fd2226c7. The V5 decoder now requires the remaining block to contain at least one byte per entry in each of the four columns before allocating column slices, so a crafted N cannot drive multi-column over-allocation. Verified with go test and golangci-lint for pkg/storage/header.

packages/shared/pkg/storage/header/serialization_v5.go
```diff
@@ -0,0 +1,256 @@
...
+		return Mapping{}, fmt.Errorf("mapping count %d exceeds remaining %d bytes", n, reader.Len())
```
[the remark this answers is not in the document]
ValentaTomas: Fixed in 7fd2226c7. The V5 decoder now requires the remaining block to contain at least one byte per entry in each of the four columns before allocating column slices, so a crafted N cannot drive multi-column over-allocation. Verified with go test and golangci-lint for pkg/storage/header.

packages/shared/pkg/storage/header/serialization_v5.go
```diff
@@ -0,0 +1,256 @@
...
+	if metadata.BlockSize == 0 {
+		return nil, errors.New("v5 header has zero block size")
+	}
+	if len(blockData) < v4FlagsLen+v4SizePrefixLen {
+		return nil, fmt.Errorf("v5 header block too short for flags + size prefix: %d bytes", len(blockData))
+	}
+
+	flags := blockData[0]
+	size := binary.LittleEndian.Uint32(blockData[v4FlagsLen:])
+	if int(size) > v4MaxUncompressedHeaderSize {
```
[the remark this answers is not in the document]
ValentaTomas: Fixed in 7fd2226c7. deserializeV5 now matches V4 and compares size as uint64 against the cap, avoiding int overflow on 32-bit platforms before decompression allocation. Verified with go test and golangci-lint for pkg/storage/header.

packages/shared/pkg/featureflags/flags.go
```diff
@@ -167,6 +167,13 @@ var (
 	// uploads. Independent of compress-config: it changes the header format,
 	// not whether data is compressed.
 	V4HeaderForUncompressedFlag = NewBoolFlag("v4-header-for-uncompressed", false)
+
+	// HeaderV5WriteFlag makes Pause emit the V5 (columnar, varint) header
+	// layout instead of V4. Readers understand V5 regardless of this flag;
+	// keep it off until every orchestrator that may read these headers (incl.
+	// P2P peers) ships V5 read support, then enable it to shrink the large
+	// fragmented headers produced by page-granular memfile dedup.
+	HeaderV5WriteFlag = NewBoolFlag("header-v5-write", false)
```
levb, line 173: Nit: Setting this flag also changes the semantics of V4HeaderForUncompressedFlag to use V5, so just comment.

packages/orchestrator/pkg/sandbox/build_upload.go
```diff
@@ -29,6 +29,7 @@ type Upload struct {
 	objectMetadata storage.ObjectMetadata
 	future         *utils.ErrorOnce
 	useV4          bool
+	headerVersion  uint64
```
levb, line 32: 2/5 uint64 is rich for the header version;

1/5 headerVersion can be used/overloaded to represent useV4 aka "use the latest header format for legacy uncompressed uploads", just set `headerVersion=5` when the flag is ON; Either way, the semantics of the flag are to use V5

packages/shared/pkg/storage/header/serialization.go
```diff
@@ -36,7 +42,11 @@ func DeserializeBytes(data []byte) (*Header, error) {
 
 	blockData := data[metadataSize:]
 
-	if metadata.Version >= 4 {
+	if metadata.Version >= MetadataVersionV5 {
```
levb: I'd really like to add CompressionType (of the header) to the Metadata so we can safely try different compression algorithms. However, adding a field to Metadata is quite an ordeal. Can we pack compression type into the high byte of `Version`, 0 being LZ4 since it is already used?

All I would ask for in this PR is to & 0xFFFF for the version number check, so we can change other bits at will?

levb: Also, 1/5 we should tighten the checks to known versions only (`==`) and error if `version` is not parseable.

packages/shared/pkg/storage/header/serialization_v5.go
```diff
@@ -0,0 +1,313 @@
...
+	result[metadataSize] = flags
+	binary.LittleEndian.PutUint32(result[metadataSize+v4FlagsLen:], uint32(len(blockBytes)))
+	copy(result[metadataSize+v4FlagsLen+v4SizePrefixLen:], compressed)
+
+	return result, int64(len(blockBytes)), nil
+}
+
+// writeV5MappingSection writes the columnar, varint-coded mapping:
+//
+//	uint32 N                       non-empty entry count
```
levb: 0/5 nit: maybe Mbuild + builds first, then Nentries followed by the 4 columns?

packages/shared/pkg/storage/header/serialization_v5.go
```diff
@@ -0,0 +1,313 @@
...
+//	uint32 M                       distinct non-empty build-id count
+//	M × [16]byte                   build-id table
+//	N × uvarint                    offset-block deltas (offset[i]-offset[i-1])
+//	N × uvarint                    length in blocks
+//	N × varint  (zig-zag)          storage-block deltas (storage[i]-storage[i-1])
```
levb: 1/5 zig-zag, meaning it is not monotonic??? An simple example here would be NICE.

packages/shared/pkg/storage/header/serialization_v5.go
```diff
@@ -0,0 +1,313 @@
...
+	}
+	if err := binary.Read(reader, binary.LittleEndian, &m); err != nil {
+		return Mapping{}, fmt.Errorf("failed to read mapping build count: %w", err)
+	}
+	if m > maxBuildsPerHeader {
+		return Mapping{}, fmt.Errorf("mapping build count %d exceeds maximum %d", m, maxBuildsPerHeader)
+	}
+
+	builds := make([]uuid.UUID, m)
+	for i := range builds {
```
levb, line 205: nit: for i := range m

packages/shared/pkg/storage/header/serialization_v5.go
```diff
@@ -0,0 +1,313 @@
...
+		return Mapping{}, fmt.Errorf("mapping count %d exceeds remaining %d bytes", n, reader.Len())
+	}
+
+	sparseOffsets := make([]uint32, n)
+	sparseLengths := make([]uint32, n)
+	sparseStorage := make([]uint32, n)
+	sparseBuildIdx := make([]uint16, n)
+
+	var prevOffset uint64
+	for i := range sparseOffsets {
```
levb: range n

packages/shared/pkg/storage/header/serialization_v5.go
```diff
@@ -0,0 +1,313 @@
...
+		return newMappingFromColumns(blockSize, builds,
+			[]uint32{0}, []uint32{uint32(sizeBlocks)}, []uint32{0}, []uint16{nilBuildIdx})
+	}
+	// Each entry needs at least one byte in each of the four columns. Bound the
+	// allocation against a crafted count before allocating the column slices.
+	if uint64(n) > uint64(reader.Len())/4 {
+		return Mapping{}, fmt.Errorf("mapping count %d exceeds remaining %d bytes", n, reader.Len())
+	}
+
+	sparseOffsets := make([]uint32, n)
```
levb: 0/5 "sparse" is a bit misleading here; I do not have a better suggestion though.

levb (dismissed):
Small nits, one large Q - do we want to use V5 to support different compression on the header, or stick with LZ4?

packages/shared/pkg/featureflags/flags.go
ValentaTomas: Fixed in 02cdcd095: the HeaderV5WriteFlag comment now calls out that it supersedes V4HeaderForUncompressedFlag for uncompressed uploads.

packages/orchestrator/pkg/sandbox/build_upload.go
ValentaTomas: Kept headerVersion as uint64 because Metadata.Version and CloneForUpload use uint64, but the effective write semantics are now clearer: HeaderV5WriteFlag selects V5 and supersedes V4HeaderForUncompressedFlag.

packages/shared/pkg/storage/header/serialization.go
ValentaTomas: Fixed in 02cdcd095: serialization/deserialization now dispatch on the low 16 bits and reject unsupported format versions instead of treating every future version as V5/V4.

packages/shared/pkg/storage/header/serialization_v5.go
ValentaTomas: Fixed in 02cdcd095: V5 now writes the build table count/table before the non-empty entry count and columns.

packages/shared/pkg/storage/header/serialization_v5.go
ValentaTomas: Fixed in 02cdcd095: the storage-delta comment now explains why negative deltas can happen when switching builds.

packages/shared/pkg/storage/header/serialization_v5.go
ValentaTomas: Fixed in 02cdcd095.

packages/shared/pkg/storage/header/serialization_v5.go
ValentaTomas: Fixed in 02cdcd095.

packages/shared/pkg/storage/header/serialization_v5.go
ValentaTomas: Fixed in 02cdcd095: renamed the decoded V5 arrays from sparse* to encoded*.

packages/orchestrator/pkg/sandbox/build_upload.go
```diff
@@ -61,6 +67,7 @@ func NewUpload(
 		useCase:        useCase,
 		objectMetadata: objectMetadata,
 		useV4:          memV4 || rootV4,
```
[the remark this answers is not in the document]
ValentaTomas: Fixed in a9e36bc8d. HeaderV5WriteFlag now routes even uncompressed uploads through the V4/V5 upload path so CloneForUpload can emit V5.

packages/shared/pkg/storage/header/serialization_v5.go
```diff
@@ -0,0 +1,313 @@
...
+		return newMappingFromColumns(blockSize, builds,
+			[]uint32{0}, []uint32{uint32(sizeBlocks)}, []uint32{0}, []uint16{nilBuildIdx})
+	}
+	// Each entry needs at least one byte in each of the four columns. Bound the
+	// allocation against a crafted count before allocating the column slices.
+	if uint64(n) > uint64(reader.Len())/4 {
+		return Mapping{}, fmt.Errorf("mapping count %d exceeds remaining %d bytes", n, reader.Len())
+	}
+
+	encodedOffsets := make([]uint32, n)
```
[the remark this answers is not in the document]
ValentaTomas: Fixed in a9e36bc8d. The V5 decoder now rejects mapping counts above a hard cap before allocating the encoded/reconstructed columns, in addition to the remaining-byte guard.

ValentaTomas:
Addressed the remaining allocation guard feedback in 5f1b5ee by bounding V4 build and mapping counts against remaining bytes before allocating. Also added focused regression tests. Test: `go test ./packages/shared/pkg/storage/header`.

packages/orchestrator/pkg/sandbox/build_upload.go
```diff
@@ -50,6 +51,11 @@ func NewUpload(
 		return nil, fmt.Errorf("resolve rootfs compress config: %w", err)
 	}
 
+	headerVersion := uint64(headers.MetadataVersionV4)
+	if ff != nil && ff.BoolFlag(ctx, featureflags.HeaderV5WriteFlag) {
+		headerVersion = headers.MetadataVersionV5
```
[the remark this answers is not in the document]
ValentaTomas: Fixed in f969efd95. NewUpload now adds the upload use-case context before evaluating HeaderV5WriteFlag, matching the compress-config flag reads. Local compile/lint for pkg/sandbox pass.

packages/shared/pkg/storage/header/serialization_v4.go
```diff
@@ -244,6 +189,77 @@ func compressLZ4(data []byte) ([]byte, error) {
...
+func readV4BuildsSection(reader *bytes.Reader) (map[uuid.UUID]BuildData, error) {
+	var numBuilds uint32
+	if err := binary.Read(reader, binary.LittleEndian, &numBuilds); err != nil {
+		return nil, fmt.Errorf("failed to read build count: %w", err)
+	}
+	if numBuilds == 0 {
+		return nil, nil
+	}
+
+	builds := make(map[uuid.UUID]BuildData, numBuilds)
```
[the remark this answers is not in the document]
ValentaTomas: Already fixed on the current branch. readV4BuildsSection bounds numBuilds against the remaining reader length before allocating the map (minimum BuildInfo + frame-table header bytes per build).

ValentaTomas:
Addressed the latest review findings in 3b99cbe: HeaderV5WriteFlag now evaluates with the upload use-case context, V5 accepts rounded final mappings for non-page-aligned sizes, and the V5 mapping-count guard is more conservative before allocating decoded/reconstructed columns. Test: `go test ./packages/shared/pkg/storage/header`.

ValentaTomas:
Fixed the latest CI/review issues in aa3c486: removed the duplicate use-case context addition, tightened the V5 count cap to account for reconstructed gap entries, and fixed the unconvert lint in the non-aligned size test. Test: `go test ./packages/shared/pkg/storage/header`.

packages/shared/pkg/storage/header/serialization_v5.go
```diff
@@ -0,0 +1,318 @@
...
+
+	return h, nil
+}
+
+// readV5MappingSection reads the columnar mapping written by
+// writeV5MappingSection and reconstructs the compact Mapping directly.
+func readV5MappingSection(reader *bytes.Reader, blockSize, size uint64) (Mapping, error) {
+	if size%blockSize != 0 {
+		return Mapping{}, fmt.Errorf("mapping size %d is not block-aligned to %d", size, blockSize)
+	}
```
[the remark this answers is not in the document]
ValentaTomas: Already fixed on the current branch. V5 decode rounds Metadata.Size to page blocks before reconstructing empty gaps, NewMapping allows 65535 non-empty build IDs while reserving 65535 as the nil sentinel, and V5 bounds encoded/reconstructed mapping allocation before allocating. Header tests and lint pass.

packages/shared/pkg/storage/header/compact.go
```diff
@@ -66,14 +72,18 @@ func NewMapping(blockSize uint64, src []BuildMap) (Mapping, error) {
...
-		if !ok {
-			if len(builds) >= maxBuildsPerHeader {
-				return Mapping{}, fmt.Errorf("compact mapping: more than %d unique build IDs", maxBuildsPerHeader)
+		idx := uint16(nilBuildIdx)
+		if m.BuildId != uuid.Nil {
+			var ok bool
+			idx, ok = idxByBuild[m.BuildId]
+			if !ok {
+				if len(builds) >= maxBuildsPerHeader {
+					return Mapping{}, fmt.Errorf("compact mapping: more than %d unique build IDs", maxBuildsPerHeader)
```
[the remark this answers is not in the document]
ValentaTomas: Already fixed on the current branch. V5 decode rounds Metadata.Size to page blocks before reconstructing empty gaps, NewMapping allows 65535 non-empty build IDs while reserving 65535 as the nil sentinel, and V5 bounds encoded/reconstructed mapping allocation before allocating. Header tests and lint pass.

packages/shared/pkg/storage/header/serialization_v5.go
```diff
@@ -0,0 +1,327 @@
...
+		}
+
+		return newMappingFromColumns(blockSize, builds,
+			[]uint32{0}, []uint32{uint32(sizeBlocks)}, []uint32{0}, []uint16{nilBuildIdx})
+	}
+	// Each entry needs at least one byte in each of the four encoded columns,
+	// plus four reconstructed columns below. Bound the crafted count against
+	// the remaining data and a conservative entry cap before allocating them.
+	maxEntriesByBytes := uint64(reader.Len()) / 4
+	maxEntriesByMemory := uint64(v4MaxUncompressedHeaderSize) / 32
```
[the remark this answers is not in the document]
ValentaTomas: Already fixed on the current branch. V5 decode rounds Metadata.Size to page blocks before reconstructing empty gaps, NewMapping allows 65535 non-empty build IDs while reserving 65535 as the nil sentinel, and V5 bounds encoded/reconstructed mapping allocation before allocating. Header tests and lint pass.

levb (approved):
LGTM

PR #2853, opened 2026-05-29
feat: add scheduled cleanup for orphaned Firecracker processes

AdaAibaby:
Add periodic cleanup task to remove orphaned Firecracker processes on orchestrator nodes.

Orphaned FC processes will be automatically cleared regularly to prevent resource leakage and reduce host resource occupation.


jakubno:
I believe this is now addressed by #3086

PR #2858, opened 2026-05-29
perf(orch): metrics for dirty page throttling

kalyazin:
# Summary

## Problem

When resuming a sandbox with a large memory snapshot (e.g. 32 GB), the
orchestrator creates two mmap-backed caches backed by sparse files on the
host's NVMe RAID:

- a read-through cache that materialises memfile chunks fetched from
  remote storage as Firecracker takes page faults;
- a copy-on-write overlay cache that captures guest writes to the rootfs.

Writing tens of gigabytes of dirty pages into block-device-backed files
triggers Linux's per-BDI dirty-page throttle (`balance_dirty_pages`).
The kernel stalls the goroutines doing the writes until the writeback
daemon catches up, blocking the entire resume for ~64 s — roughly 4× longer than the
actual I/O work.

## What this PR adds

**Metrics** (always on, no flag required):

- `orchestrator.blocks.chunks.store` — native histogram of the time to
  fetch each batch of data from remote storage and write it into the
  local mmap cache. During a throttled resume the p99 spikes to
  300–600 ms while the p50 stays low; this ratio is the primary
  diagnostic signal.
- `orchestrator.host.balance_dirty_pages.threads` — counter of kernel
  dirty-page throttle polls observed by the process. A non-zero rate
  confirms the throttle is active; zero rules it out as the cause of
  slow resumes.

### DIFF
--- a/packages/orchestrator/pkg/sandbox/fc/fc_metrics.go
+++ b/packages/orchestrator/pkg/sandbox/fc/fc_metrics.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
 	"go.opentelemetry.io/otel"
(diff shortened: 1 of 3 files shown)


packages/orchestrator/pkg/sandbox/block/cache.go
```diff
@@ -657,9 +762,24 @@ func (c *Cache) WriteZeroesAt(off, length int64) (int, error) {
...
+		}
+		// memfd has no filesystem path; use fstat on the open fd.
+		// stat.Blocks is always in 512-byte units on Linux regardless of backing store.
+		var stat syscall.Stat_t
+		if err := syscall.Fstat(int(c.file.Fd()), &stat); err != nil {
+			return 0, fmt.Errorf("fstat memfd: %w", err)
+		}
+
+		return stat.Blocks * 512, nil
+	}
```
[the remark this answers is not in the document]
kalyazin: addressed

packages/orchestrator/pkg/sandbox/block/cache.go
```diff
@@ -59,8 +59,27 @@ type Cache struct {
...
+	// GC does not close the fd while the mmap is still in use. Nil for path-backed caches.
+	file    *os.File
+	isMemfd bool
 }
 
+// SetWriteTimer wires a TimerFactory into the cache for measuring mmap write duration.
+// Must be called before any writes; not goroutine-safe during setup.
+func (c *Cache) SetWriteTimer(t telemetry.TimerFactory) {
+	c.writeTimer = &t
+}
```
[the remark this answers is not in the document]
kalyazin: fixed

packages/orchestrator/pkg/sandbox/fc/fc_metrics.go
```diff
@@ -66,8 +67,67 @@ var (
...
+// balanceDirtyPagesWchan is the kernel wait-channel symbol written to
+// /proc/self/task/*/wchan when a thread is parked in balance_dirty_pages.
+// Fires for both per-BDI and global dirty-page throttle regardless of the
+// global dirty/MemTotal ratio, making it the only reliable userspace signal
+// for per-BDI throttle on large-RAM nodes.
+const balanceDirtyPagesWchan = "balance_dirty_pages"
+
+// countBalanceDirtyThreads returns the number of OS threads of the current
+// process currently stalled in balance_dirty_pages by reading
+// /proc/self/task/*/wchan. Returns 0 on any read error.
```
bchalios, line 92: Why is 0 here fine? Can't we return an error?

packages/orchestrator/pkg/sandbox/block/cache.go
```diff
@@ -503,8 +580,16 @@ func (c *Cache) Close() (e error) {
 		e = errors.Join(e, fmt.Errorf("error unmapping mmap: %w", err))
 	}
 
-	// TODO: Move to to the scope of the caller
-	e = errors.Join(e, os.RemoveAll(c.filePath))
+	if c.isMemfd {
```
bchalios: These are lots of `if` statements. Maybe it makes sense to extract memfd vs non-memfd functions and call them from here for either specific case

bchalios (commented):
Implementation looks sane. My main concern is that we're now gonna have to keep snapshots in-memory. Maybe a more conservative approach would be, trying to rate limit writing in the disk rather than dropping writes all together.

packages/orchestrator/pkg/sandbox/fc/fc_metrics.go
kalyazin: I'm not sure how we can represent an error in a metric. Do you have an idea?

packages/orchestrator/pkg/sandbox/block/cache.go
kalyazin: dropped the memfd-based fix in favour of https://github.com/e2b-dev/infra/pull/2861 due to concerns about increased resident memory consumption.

bchalios (approved):

PR #2863, opened 2026-05-29
perf(storage): reduce multipart upload copying

ValentaTomas:
## Summary
Remove multipart Content-MD5 hashing and reduce compression/upload allocations by pooling frame input buffers and using a custom multi-slice upload reader.

## Test plan
- `go test ./packages/shared/pkg/storage`


packages/shared/pkg/storage/gcp_multipart.go
```diff
@@ -291,29 +283,48 @@ func (m *MultipartUploader) uploadPart(ctx context.Context, uploadID string, par
...
 }
 
+type multiSliceReader struct {
+	slices [][]byte
+	idx    int
+	off    int
+}
+
+func (r *multiSliceReader) Read(p []byte) (int, error) {
+	var n int
```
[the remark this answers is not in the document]
ValentaTomas: Fixed in ef2d92fa1. multiSliceReader now follows the io.Reader zero-length read contract, returning 0,nil before EOF and 0,io.EOF only once exhausted. Also fixed the shared lint issue. Local storage tests and golangci-lint pass.

levb (dismissed):
See https://github.com/e2b-dev/infra/pull/2869

packages/shared/pkg/storage/compress_upload.go
```diff
@@ -83,11 +102,13 @@ func newPart(index int, parentCtx context.Context, workers int) (*part, context.
...
 
-func (p *part) addFrame(ctx context.Context, uncompressedData []byte, pool *sync.Pool) {
-	frameInPart := &frame{uncompressedSize: len(uncompressedData)}
+func (p *part) addFrame(ctx context.Context, bufPtr *[]byte, n int, pool *sync.Pool) {
+	frameInPart := &frame{uncompressedSize: n}
 	p.frames = append(p.frames, frameInPart)
+	uncompressedData := (*bufPtr)[:n]
 
 	p.compress.Go(func() error {
+		defer inputBufPool.Put(bufPtr)
```
dobrac: lets not access the inputBufPool from here at all - keep the responsibility to the place where gets are happening. Keeping get - put together

If we need to do async callback, lets do that instead, or move the goroutine schedule to the `readLoop` function (if possible)

dobrac (approved):

packages/shared/pkg/storage/compress_upload.go
ValentaTomas: Do you want to move the defer to line 218 here or am I misunderstanding something?

ValentaTomas: I'm not sure if wrapping in callback make sense. This is fundamentally just a `free` for the memory, if we don't call it and because we need it to be async putting the callback there does not feel like it clears up the lifecycle. It is the same method, but not it is also closured.

dobrac: The goal is to keep the Get/Put combination next to each other - in one place. That way it's possible to validate the buffer free when it's stopped being used.<br><br>Now its across two functions which makes it difficult to track

ValentaTomas: We need to free it in the goroutine, so I don't think you can properly validate that here.

ValentaTomas: Wrapped the pooled buffer in an `inputBuf` type that owns the pool. The pool is now private to that type: callers get a buffer via `getInputBuf` and release it via `buf.Free()`, and never touch `inputBufPool` directly. So the get is paired with `Free` as the buffer's own lifecycle methods, which makes the free easy to track even though it has to happen async in the compress goroutine (`defer buf.Free()`).

packages/shared/pkg/storage/compress_upload.go
```diff
@@ -62,6 +62,43 @@ func (m *memPartUploader) Assemble() []byte {
...
+
+func getInputBuf(size int) inputBuf {
+	if v := inputBufPool.Get(); v != nil {
+		bufPtr := v.(*[]byte)
+		if cap(*bufPtr) >= size {
+			*bufPtr = (*bufPtr)[:size]
+
+			return inputBuf{ptr: bufPtr}
+		}
+	}
```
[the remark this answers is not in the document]
ValentaTomas: Fixed in the extracted bufferPool.Get: on the resize path it now grows the pooled buffer in place (reallocating only the backing array through the existing *[]byte) instead of dropping the pooled pointer, so the get/put pairing holds.

PR #2867, opened 2026-05-30
Add cache-aware placement affinity

ValentaTomas:
Add disabled-by-default cache-aware placement affinity using recent successful build placements. Redis state is bounded by TTL/top-N limits and is best-effort.

### DIFF
--- a/packages/api/internal/orchestrator/client.go
+++ b/packages/api/internal/orchestrator/client.go
@@ -226,12 +226,7 @@ func (o *Orchestrator) discoverClusterNode(ctx context.Context, clusterID uuid.U
 	var wg sync.WaitGroup
 	defer wg.Wait()
 
-	instances := cluster.GetOrchestrators()
-	if len(instances) == 0 && cluster.ID == consts.LocalClusterID {
-		instances = cluster.GetTemplateBuilders()
-	}
-
-	for _, instance := range instances {
+	for _, instance := range cluster.GetOrchestrators() {
 		wg.Go(func() {
 			o.connectToClusterNode(ctx, cluster, instance)
 		})
--- a/packages/api/internal/orchestrator/client.go
+++ b/packages/api/internal/orchestrator/client.go
@@ -226,7 +226,12 @@ func (o *Orchestrator) discoverClusterNode(ctx context.Context, clusterID uuid.U
 	var wg sync.WaitGroup
 	defer wg.Wait()
 
-	for _, instance := range cluster.GetOrchestrators() {
+	instances := cluster.GetOrchestrators()
+	if len(instances) == 0 && cluster.ID == consts.LocalClusterID {
+		instances = cluster.GetTemplateBuilders()
+	}
+
+	for _, instance := range instances {
 		wg.Go(func() {
 			o.connectToClusterNode(ctx, cluster, instance)
 		})
--- a/packages/api/internal/orchestrator/create_instance.go
+++ b/packages/api/internal/orchestrator/create_instance.go
@@ -300,10 +300,14 @@ func (o *Orchestrator) CreateSandbox(
 	clusterNodes := o.GetClusterNodes(nodeClusterID)
 
 	labelFilteringEnabled := o.featureFlagsClient.BoolFlag(ctx, featureflags.SandboxLabelBasedSchedulingFlag, featureflags.TeamContext(team.ID.String()), featureflags.SandboxContext(sandboxID))
+	affinityBuildID := sbxData.Build.ID.String()
+	if isResume {
+		affinityBuildID = ""
+	}
 	placementCacheAffinityConfig := placementAffinityConfigFromFlags(ctx, o.featureFlagsClient, featureflags.TeamContext(team.ID.String()), featureflags.TemplateContext(sbxData.TemplateID), featureflags.SandboxContext(sandboxID))
 	var affinityScores map[string]float64
 	if placementCacheAffinityConfig.enabled {
-		affinityScores = o.placementAffinity.scores(ctx, placementCacheAffinityConfig, nodeClusterID, sbxData.Build.ID.String(), sbxData.TemplateID, sbxData.BaseTemplateID)
+		affinityScores = o.placementAffinity.scores(ctx, placementCacheAffinityConfig, nodeClusterID, affinityBuildID, sbxData.TemplateID, sbxData.BaseTemplateID)
 	}
 
 	node, err = placement.PlaceSandbox(ctx, o.placementAlgorithm, clusterNodes, node, sbxRequest, builds.ToMachineInfo(sbxData.Build), labelFilteringEnabled, team.SandboxSchedulingLabels, affinityScores)
@@ -323,7 +327,7 @@ func (o *Orchestrator) CreateSandbox(
 	}
 	o.createdSandboxesCounter.Add(ctx, 1, metric.WithAttributes(attributes...))
 	if placementCacheAffinityConfig.enabled {
-		o.placementAffinity.record(ctx, placementCacheAffinityConfig, nodeClusterID, node.ID, sbxData.Build.ID.String(), sbxData.TemplateID, sbxData.BaseTemplateID)
+		o.placementAffinity.record(ctx, placementCacheAffinityConfig, nodeClusterID, node.ID, affinityBuildID, sbxData.TemplateID, sbxData.BaseTemplateID)
 	}
 
 	telemetry.SetAttributes(ctx, attribute.String("node.id", node.ID))
--- a/packages/api/internal/orchestrator/create_instance.go
+++ b/packages/api/internal/orchestrator/create_instance.go
@@ -346,6 +346,9 @@ func (o *Orchestrator) CreateSandbox(
 		attribute.Bool("is_resume", isResume),
 		attribute.Bool("node_affinity_requested", sbxData.NodeID != nil),
 		attribute.Bool("node_affinity_success", sbxData.NodeID != nil && node.ID == *sbxData.NodeID),
+		// Whether placement landed on a node the cache-affinity layer preferred,
+		// so the feature's hit rate (and any load skew) is observable on rollout.
+		attribute.Bool("cache_affinity_hit", affinityScores[node.ID] > 0),
 	}
 	o.createdSandboxesCounter.Add(ctx, 1, metric.WithAttributes(attributes...))
 	if placementCacheAffinityConfig.enabled {
--- a/packages/api/internal/orchestrator/create_instance.go
+++ b/packages/api/internal/orchestrator/create_instance.go
@@ -300,10 +300,10 @@ func (o *Orchestrator) CreateSandbox(
 	clusterNodes := o.GetClusterNodes(nodeClusterID)
 
 	labelFilteringEnabled := o.featureFlagsClient.BoolFlag(ctx, featureflags.SandboxLabelBasedSchedulingFlag, featureflags.TeamContext(team.ID.String()), featureflags.SandboxContext(sandboxID))
-	placementCacheAffinityEnabled := o.featureFlagsClient.BoolFlag(ctx, featureflags.SandboxPlacementCacheAffinityFlag, featureflags.TeamContext(team.ID.String()), featureflags.TemplateContext(sbxData.TemplateID), featureflags.SandboxContext(sandboxID))
+	placementCacheAffinityConfig := placementAffinityConfigFromFlags(ctx, o.featureFlagsClient, featureflags.TeamContext(team.ID.String()), featureflags.TemplateContext(sbxData.TemplateID), featureflags.SandboxContext(sandboxID))
 	var affinityScores map[string]float64
-	if placementCacheAffinityEnabled {
-		affinityScores = o.placementAffinity.scores(ctx, nodeClusterID, sbxData.Build.ID.String(), sbxData.TemplateID, sbxData.BaseTemplateID)
+	if placementCacheAffinityConfig.enabled {
+		affinityScores = o.placementAffinity.scores(ctx, placementCacheAffinityConfig, nodeClusterID, sbxData.Build.ID.String(), sbxData.TemplateID, sbxData.BaseTemplateID)
 	}
 
 	node, err = placement.PlaceSandbox(ctx, o.placementAlgorithm, clusterNodes, node, sbxRequest, builds.ToMachineInfo(sbxData.Build), labelFilteringEnabled, team.SandboxSchedulingLabels, affinityScores)
@@ -322,8 +322,8 @@ func (o *Orchestrator) CreateSandbox(
 		attribute.Bool("node_affinity_success", sbxData.NodeID != nil && node.ID == *sbxData.NodeID),
 	}
 	o.createdSandboxesCounter.Add(ctx, 1, metric.WithAttributes(attributes...))
-	if placementCacheAffinityEnabled {
-		o.placementAffinity.record(ctx, nodeClusterID, node.ID, sbxData.Build.ID.String(), sbxData.TemplateID, sbxData.BaseTemplateID)
+	if placementCacheAffinityConfig.enabled {
+		o.placementAffinity.record(ctx, placementCacheAffinityConfig, nodeClusterID, node.ID, sbxData.Build.ID.String(), sbxData.TemplateID, sbxData.BaseTemplateID)
 	}
 
 	telemetry.SetAttributes(ctx, attribute.String("node.id", node.ID))
--- a/packages/api/internal/orchestrator/create_instance.go
+++ b/packages/api/internal/orchestrator/create_instance.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net"
 	"net/http"
 	"time"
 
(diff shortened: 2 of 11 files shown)


packages/api/internal/orchestrator/placement/placement.go
```diff
@@ -102,6 +102,9 @@ func PlaceSandbox(ctx context.Context, algorithm Algorithm, clusterNodes []*node
 
 		switch statusCode {
 		case codes.ResourceExhausted:
+			if len(nodesExcluded)+1 < len(clusterNodes) {
+				nodesExcluded[failedNode.ID] = struct{}{}
+			}
```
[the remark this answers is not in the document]
ValentaTomas: Fixed in a46626980. If the last remaining node also returns ResourceExhausted, we now increment the bounded retry counter instead of spinning until context timeout. Local placement tests and lint pass.

packages/api/internal/orchestrator/placement/placement_best_of_K.go
```diff
@@ -163,40 +212,9 @@ func (b *BestOfK) sample(items []*nodemanager.Node, config BestOfKConfig, exclud
...
-		}
-
-		if config.CanFit {
-			if !b.CanFit(n, resources, config) {
-				continue
-			}
-		}
-
-		if config.TooManyStarting {
-			// To prevent overloading the node
```
dobrac, line 193: the comments got lost when moving to isCandidate

packages/api/internal/orchestrator/placement/placement_best_of_K.go
```diff
@@ -111,19 +116,40 @@ func (b *BestOfK) UpdateConfig(config BestOfKConfig) {
...
+			seen[n.ID] = struct{}{}
+		}
+		for _, n := range nodes {
+			if affinity[n.ID] <= 0 {
+				continue
+			}
+			if _, ok := seen[n.ID]; ok {
+				continue
+			}
+			if b.isCandidate(n, config, excludedNodes, resources, buildMachineInfo, filterByLabels, requiredLabels) {
```
dobrac, line 141: doesn't this effectively make the sample obsolete?

packages/api/internal/orchestrator/create_instance.go
```diff
@@ -298,10 +298,23 @@ func (o *Orchestrator) CreateSandbox(
 
 	nodeClusterID := clusters.WithClusterFallback(team.ClusterID)
 	clusterNodes := o.GetClusterNodes(nodeClusterID)
+	if len(clusterNodes) == 0 {
+		o.discoverClusterNode(ctx, nodeClusterID)
+		clusterNodes = o.GetClusterNodes(nodeClusterID)
+	}
```
dobrac, line 320: if this is trying to fix something, it would make sense to move it to the GetClusterNodes I think

packages/api/internal/orchestrator/placement_affinity.go
```diff
@@ -0,0 +1,152 @@
...
+
+func newPlacementAffinity(redisClient redis.UniversalClient) *placementAffinity {
+	if redisClient == nil {
+		return nil
+	}
+
+	return &placementAffinity{redis: redisClient}
+}
+
+func placementAffinityConfigFromFlags(ctx context.Context, ff *featureflags.Client, contexts ...ldcontext.Context) placementAffinityConfig {
```
dobrac: the context should be added to ctx independently, the contexts should not be a parameter

packages/api/internal/orchestrator/create_instance.go
```diff
@@ -298,10 +298,23 @@ func (o *Orchestrator) CreateSandbox(
...
 
 	labelFilteringEnabled := o.featureFlagsClient.BoolFlag(ctx, featureflags.SandboxLabelBasedSchedulingFlag, featureflags.TeamContext(team.ID.String()), featureflags.SandboxContext(sandboxID))
+	affinityBuildID := sbxData.Build.ID.String()
+	if isResume {
+		affinityBuildID = ""
+	}
```
dobrac, line 327: can you add explanation comment for this?

packages/api/internal/orchestrator/placement_affinity.go
```diff
@@ -0,0 +1,152 @@
...
+		topNodes:         int64(jsonInt(v, "topNodes", defaultPlacementAffinityTopNodes, 1, 100)),
+		readTimeout:      time.Duration(jsonInt(v, "readTimeoutMs", defaultPlacementAffinityReadTimeoutMs, placementAffinityMinTimeoutMs, placementAffinityMaxTimeoutMs)) * time.Millisecond,
+		writeTimeout:     time.Duration(jsonInt(v, "writeTimeoutMs", defaultPlacementAffinityWriteTimeoutMs, placementAffinityMinTimeoutMs, placementAffinityMaxTimeoutMs)) * time.Millisecond,
+		maxAffinityScore: float64(jsonInt(v, "maxAffinityScore", defaultPlacementAffinityMaxScore, 1, 1000)),
+		maxScoreBonus:    jsonPPM(v, "maxScoreBonusPpm", defaultPlacementAffinityMaxScoreBonusPpm),
+		buildWeight:      jsonPPM(v, "buildWeightPpm", defaultPlacementAffinityBuildWeightPpm),
+	}
+}
+
+func jsonInt(v ldvalue.Value, key string, fallback, minValue, maxValue int) int {
```
dobrac, line 71: these helpers should either be not required, or the logic should be done in the featureflags package, returning struct only for the config

packages/api/internal/orchestrator/placement_affinity.go
```diff
@@ -0,0 +1,152 @@
...
+func clampInt(v, minValue, maxValue int) int {
+	if v < minValue {
+		return minValue
+	}
+	if v > maxValue {
+		return maxValue
+	}
+
+	return v
+}
```
dobrac, line 86:
```suggestion
func clampInt(v, minValue, maxValue int) int {
	return max(minValue, min(v, maxValue))
}
```

packages/api/internal/orchestrator/placement_affinity.go
```diff
@@ -0,0 +1,152 @@
...
+
+func (a *placementAffinity) record(ctx context.Context, cfg placementAffinityConfig, clusterID uuid.UUID, nodeID, buildID string) {
+	if a == nil || nodeID == "" || buildID == "" || cfg.buildWeight == 0 {
+		return
+	}
+	ctx, cancel := context.WithTimeout(context.WithoutCancel(ctx), cfg.writeTimeout)
+	defer cancel()
+
+	pipe := a.redis.Pipeline()
+	key := placementAffinityKey(clusterID, "build", buildID)
```
dobrac: is it always build? Could be enum instead

packages/api/internal/orchestrator/placement/placement_best_of_K.go
```diff
@@ -61,7 +61,12 @@ func (b *BestOfK) Score(node *nodemanager.Node, resources nodemanager.SandboxRes
 
 	cpuRequested := float64(resources.CPUs)
 
-	return (cpuRequested + float64(reserved) + config.Alpha*usageAvg) / totalCapacity
+	score := (cpuRequested + float64(reserved) + config.Alpha*usageAvg) / totalCapacity
+	if len(affinityScores) > 0 {
+		score -= affinityScores[0][node.ID]
```
dobrac, line 66: is this somehow normalized with the usages?

packages/api/internal/orchestrator/client.go
```diff
@@ -226,7 +226,12 @@ func (o *Orchestrator) discoverClusterNode(ctx context.Context, clusterID uuid.U
 	var wg sync.WaitGroup
 	defer wg.Wait()
 
-	for _, instance := range cluster.GetOrchestrators() {
+	instances := cluster.GetOrchestrators()
+	if len(instances) == 0 && cluster.ID == consts.LocalClusterID {
+		instances = cluster.GetTemplateBuilders()
+	}
```
dobrac: why this?

dobrac (dismissed):

packages/api/internal/orchestrator/create_instance.go
```diff
@@ -298,10 +298,23 @@ func (o *Orchestrator) CreateSandbox(
 
 	nodeClusterID := clusters.WithClusterFallback(team.ClusterID)
 	clusterNodes := o.GetClusterNodes(nodeClusterID)
+	if len(clusterNodes) == 0 {
+		o.discoverClusterNode(ctx, nodeClusterID)
```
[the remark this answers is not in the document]
ValentaTomas: Fixed in a3c5f190e: empty local placement cache refresh now uses Nomad node discovery instead of cluster/template-builder discovery.

packages/api/internal/orchestrator/client.go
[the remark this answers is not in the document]
ValentaTomas: Fixed in a3c5f190e: local empty-pool refresh no longer registers template-builder-only instances as placement nodes.

packages/api/internal/orchestrator/create_instance.go
ValentaTomas: Fixed in a3c5f190e by keeping local refresh on the Nomad discovery path used by placement nodes. GetClusterNodes remains a pure accessor.

packages/api/internal/orchestrator/placement_affinity.go
ValentaTomas: Fixed in a3c5f190e: flag contexts are now added to ctx before calling placementAffinityConfigFromFlags.

packages/api/internal/orchestrator/create_instance.go
ValentaTomas: Fixed in a3c5f190e: added a short comment explaining why resume disables cache affinity.

packages/api/internal/orchestrator/placement_affinity.go
ValentaTomas: Kept the JSON parsing local to this feature for now, but the call site now passes context normally and the parser returns a typed placementAffinityConfig struct.

packages/api/internal/orchestrator/placement_affinity.go
ValentaTomas: Already fixed in the current branch: clampInt now uses max(minValue, min(v, maxValue)).

packages/api/internal/orchestrator/placement_affinity.go
ValentaTomas: Fixed in a3c5f190e: the build affinity kind is now a named constant.

packages/api/internal/orchestrator/placement/placement_best_of_K.go
```diff
@@ -61,7 +61,12 @@ func (b *BestOfK) Score(node *nodemanager.Node, resources nodemanager.SandboxRes
...
+	}
```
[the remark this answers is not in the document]
ValentaTomas: Fixed in a3c5f190e: affinity bonus is now normalized by totalCapacity to match the placement score scale.

packages/api/internal/orchestrator/placement/placement_best_of_K.go
ValentaTomas: Fixed in a3c5f190e: restored the candidate-filter explanation comments on isCandidate.

packages/api/internal/orchestrator/placement/placement_best_of_K.go
ValentaTomas: This intentionally keeps the random sample as the baseline and only appends eligible affinity nodes that were not sampled, so affinity can break cold-cache misses without replacing the sampling behavior.

packages/api/internal/orchestrator/placement/placement_best_of_K.go
ValentaTomas: Fixed in a3c5f190e by normalizing the affinity bonus in Score on the same totalCapacity denominator as the base score.

packages/api/internal/orchestrator/client.go
ValentaTomas: The local fallback was removed from this path in a3c5f190e; local placement refresh now goes through discoverNomadNodes.

packages/api/internal/orchestrator/placement_affinity.go
```diff
@@ -0,0 +1,152 @@
...
+	pipe.ZIncrBy(ctx, key, 1, nodeID)
+	pipe.ZRemRangeByRank(ctx, key, 0, -cfg.topNodes-1)
+	pipe.Expire(ctx, key, cfg.ttl)
+
+	_, err := pipe.Exec(ctx)
+	if err != nil {
+		logger.L().Debug(ctx, "failed to record placement affinity", zap.Error(err))
+	}
+}
```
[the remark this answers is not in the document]
ValentaTomas: Fixed in 0c56ca2e0. Affinity recording now runs asynchronously, so Redis latency no longer blocks the sandbox creation response path. Local orchestrator tests and lint pass.

packages/api/internal/orchestrator/create_instance.go
```diff
@@ -298,10 +300,40 @@ func (o *Orchestrator) CreateSandbox(
...
+					NomadNodeShortID:    "local",
+					OrchestratorAddress: net.JoinHostPort("localhost", env.GetEnv("ORCHESTRATOR_PORT", "5008")),
+					IPAddress:           "localhost",
+				})
+			}
+		} else {
+			o.discoverClusterNode(ctx, nodeClusterID)
+		}
+		clusterNodes = o.GetClusterNodes(nodeClusterID)
+	}
```
dobrac, line 320: what does this fix?

packages/api/internal/orchestrator/create_instance.go
```diff
@@ -298,10 +300,40 @@ func (o *Orchestrator) CreateSandbox(
...
+	var affinityScores map[string]float64
+	if placementCacheAffinityConfig.enabled && node == nil {
+		affinityScores = o.placementAffinity.scores(ctx, placementCacheAffinityConfig, nodeClusterID, affinityBuildID)
+	}
 
-	node, err = placement.PlaceSandbox(ctx, o.placementAlgorithm, clusterNodes, node, sbxRequest, builds.ToMachineInfo(sbxData.Build), labelFilteringEnabled, team.SandboxSchedulingLabels)
+	if affinityScores == nil {
+		node, err = placement.PlaceSandbox(ctx, o.placementAlgorithm, clusterNodes, node, sbxRequest, builds.ToMachineInfo(sbxData.Build), labelFilteringEnabled, team.SandboxSchedulingLabels)
+	} else {
+		node, err = placement.PlaceSandbox(ctx, o.placementAlgorithm, clusterNodes, node, sbxRequest, builds.ToMachineInfo(sbxData.Build), labelFilteringEnabled, team.SandboxSchedulingLabels, affinityScores)
```
dobrac: the nil check here might not be required

packages/api/internal/orchestrator/create_instance.go
ValentaTomas: This handles the case where no orchestrator nodes are connected yet (e.g. right after API startup, before background discovery has populated the cluster). Without it the first CreateSandbox would fail with "no nodes available". For the local cluster it triggers Nomad discovery and, if still empty, connects directly to the local orchestrator (combined/local-dev setup); for remote clusters it syncs the cluster instances on demand. Added an inline comment to clarify.

packages/api/internal/orchestrator/create_instance.go
ValentaTomas: Right, the nil check is unnecessary since PlaceSandbox is variadic and a nil affinity map is handled downstream (lookups return 0). Simplified to always pass affinityScores.

packages/api/internal/orchestrator/placement/placement_best_of_K.go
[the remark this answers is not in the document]
ValentaTomas: Intentional: the affinity bonus is configured directly on the normalized score scale (maxScoreBonus defaults to 0.02, i.e. a fraction-of-capacity nudge), so it is meant to be a consistent absolute preference rather than a per-node-capacity-relative one. Dividing by totalCapacity would make the cache-affinity nudge weaker on larger nodes, which is the opposite of what we want. Keeping it as-is.

packages/api/internal/orchestrator/placement/placement.go
```diff
@@ -53,12 +56,33 @@ func PlaceSandbox(ctx context.Context, algorithm Algorithm, clusterNodes []*node
...
+				// whole exhausted pool since capacity may free up.
+				if len(nodesExhausted) > 0 {
+					exhaustedRetries++
+					if exhaustedRetries >= maxExhaustedRetries {
+						return nil, err
+					}
+					clear(nodesExhausted)
+					time.Sleep(100 * time.Millisecond)
+
+					continue
```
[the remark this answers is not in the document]
ValentaTomas: Fixed: the exhausted-pool retry now waits with a select on ctx.Done() instead of a blocking time.Sleep, so placement returns promptly when the client disconnects rather than looping through the remaining retries.

ValentaTomas:
Temporarily closing.

PR #2873, opened 2026-05-30
Expose internal scheduling metadata

ValentaTomas:
Expose compact scheduling metadata from orchestrator and template-manager without DB or public API changes.


packages/orchestrator/pkg/sandbox/snapshot.go
```diff
@@ -53,3 +56,112 @@ func (s *Snapshot) Close(ctx context.Context) error {
...
+type schedulingBuildContribution struct {
+	buildID      uuid.UUID
+	memfileBytes uint64
+	rootfsBytes  uint64
+}
+
+func NewSnapshotSchedulingMetadata(memfileHeader, rootfsHeader *header.Header, limit int) *orchestrator.SnapshotSchedulingMetadata {
+	if memfileHeader == nil || memfileHeader.Metadata == nil || rootfsHeader == nil || rootfsHeader.Metadata == nil {
+		return nil
+	}
```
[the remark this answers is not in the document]
ValentaTomas: This one is not applicable: header.Mapping is a value type, not a pointer/slice/map, so it cannot be nil. The zero value is safe here: All() iterates over nil internal slices and Len() returns 0.

packages/api/internal/orchestrator/pause_instance.go
```diff
@@ -84,19 +90,35 @@ func snapshotInstance(ctx context.Context, node *nodemanager.Node, sbx sandbox.S
...
+	if err != nil {
+		return fmt.Errorf("marshal snapshot scheduling metadata: %w", err)
 	}
+	if err := o.sqlcDB.UpdateSnapshotSchedulingMetadata(ctx, queries.UpdateSnapshotSchedulingMetadataParams{
+		SandboxID: sandboxID,
+		Metadata:  string(data),
+	}); err != nil {
+		return fmt.Errorf("update snapshot scheduling metadata: %w", err)
+	}
+	o.snapshotCache.Invalidate(context.WithoutCancel(ctx), sandboxID)
```
dobrac: won't this make the snapshotCache completely ineffective? Shouldn't we instead directly update the data?

packages/orchestrator/pkg/sandbox/snapshot.go
```diff
@@ -53,3 +57,112 @@ func (s *Snapshot) Close(ctx context.Context) error {
...
+	if limit < 0 {
+		limit = 0
+	}
+	if limit > 128 {
+		limit = 128
```
dobrac: lets make the 128 const as hard limit

packages/orchestrator/pkg/sandbox/snapshot.go
```diff
@@ -53,3 +57,112 @@ func (s *Snapshot) Close(ctx context.Context) error {
...
 	return nil
 }
+
+type schedulingBuildContribution struct {
+	buildID      uuid.UUID
+	memfileBytes uint64
+	rootfsBytes  uint64
+}
+
+func NewSnapshotSchedulingMetadata(memfileHeader, rootfsHeader *header.Header, limit int) *orchestrator.SnapshotSchedulingMetadata {
```
dobrac: can we add trace here?

dobrac: this should be probably a function as it doesn't have any methods

packages/orchestrator/pkg/sandbox/snapshot.go
```diff
@@ -53,3 +57,112 @@ func (s *Snapshot) Close(ctx context.Context) error {
...
+		add(m.BuildId, m.Length, 0)
+	}
+	for _, m := range rootfsHeader.Mapping.All() {
+		add(m.BuildId, 0, m.Length)
+	}
+
+	baseBuildID := memfileHeader.Metadata.BaseBuildId
+	if baseBuildID == uuid.Nil {
+		baseBuildID = rootfsHeader.Metadata.BaseBuildId
+	}
```
dobrac: could you explain this please? Should we have two different baseBuildIDs?

packages/orchestrator/pkg/sandbox/snapshot.go
```diff
@@ -53,3 +57,112 @@ func (s *Snapshot) Close(ctx context.Context) error {
...
+				} else {
+					chain[len(chain)-1] = schedulingBuildContribution{buildID: baseBuildID}
+				}
+			}
+		}
+	}
+
+	res := &orchestrator.SnapshotSchedulingMetadata{
+		BaseBuildId:         baseBuildID.String(),
+		Generation:          max(memfileHeader.Metadata.Generation, rootfsHeader.Metadata.Generation) + 1,
```
dobrac: why having it united like this?

dobrac (dismissed):

packages/api/internal/orchestrator/pause_instance.go
ValentaTomas: This path is only called after the DB row is updated and the cached snapshot may already contain stale scheduling metadata. Direct cache update would be nicer, but SnapshotCache only exposes Get/Invalidate today; keeping invalidation here is the minimal safe path and preserves cache correctness.

packages/orchestrator/pkg/sandbox/snapshot.go
ValentaTomas: Agreed on keeping this as a hard cap; the current code clamps the configured limit to 128. Left it as the hard maximum in this PR.

packages/orchestrator/pkg/sandbox/snapshot.go
ValentaTomas: This computation is fully local over already-loaded headers and runs inside the pause/checkpoint flow; I did not add a trace span to keep the PR smaller. The important persisted fields are visible on the resulting snapshot metadata.

ValentaTomas: Kept it as a function (`NewSnapshotSchedulingMetadata`) rather than adding methods; the small private contribution struct is only an accumulator for sorting/aggregation.

packages/orchestrator/pkg/sandbox/snapshot.go
ValentaTomas: There is one scheduling base in the persisted metadata. We prefer the memfile base and fall back to rootfs only if the memfile base is nil; this avoids adding two base IDs to the scheduling contract in this PR.

packages/orchestrator/pkg/sandbox/snapshot.go
ValentaTomas: The chain is unified intentionally so placement sees one compact build lineage ordered by total contribution, with separate memfile/rootfs byte columns for weighting/debugging. Splitting into two chains would make the API/placement contract larger without clear benefit here.

packages/orchestrator/pkg/sandbox/snapshot.go
dobrac: why don't we have two baseBuildIDs if effectively there are two?

packages/orchestrator/pkg/sandbox/build/cache.go
```diff
@@ -171,6 +180,51 @@ func (s *DiffStore) Lookup(key DiffStoreKey) (Diff, bool) {
...
+		if stats == nil {
+			stats = &CachedBuildStats{BuildID: buildID}
+			byBuildID[buildID] = stats
+		}
+
+		// FileSize is the on-disk (cached) size; for a partially-fetched diff
+		// loaded from storage it is smaller than the full logical size.
+		cachedSize, err := item.Value().FileSize(ctx)
+		if err != nil || cachedSize < 0 {
+			continue
```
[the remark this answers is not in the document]
ValentaTomas: Valid. StorageDiff.FileSize was the only chunker accessor missing the nil guard that Size/IsCached/Close already have, so CachedBuildStats could panic on a diff cached before Init. Added the guard (returns 0 cached bytes pre-Init).

packages/orchestrator/pkg/sandbox/snapshot.go
```diff
@@ -53,3 +56,15 @@ func (s *Snapshot) Close(ctx context.Context) error {
...
 	return nil
 }
+
+// NewSnapshotSchedulingMetadata derives the metadata from the new snapshot's
+// diff headers so the freshly created build and its contributions are included.
+func NewSnapshotSchedulingMetadata(ctx context.Context, memfileDiffHeader, rootfsDiffHeader *DiffHeader) *orchestrator.SchedulingMetadata {
+	memfileHeader, memfileErr := memfileDiffHeader.WaitWithContext(ctx)
+	rootfsHeader, rootfsErr := rootfsDiffHeader.WaitWithContext(ctx)
+	if memfileErr != nil || rootfsErr != nil {
+		return nil
```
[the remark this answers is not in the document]
ValentaTomas: Scheduling metadata is best-effort enrichment and the header errors already surface in the upload path, so failing pause here would be a regression. Added a warning log so a dropped lineage is observable instead of silent.

packages/orchestrator/pkg/sandbox/snapshot.go
```diff
@@ -53,3 +59,18 @@ func (s *Snapshot) Close(ctx context.Context) error {
...
+	memfileHeader, memfileErr := memfileDiffHeader.WaitWithContext(ctx)
+	rootfsHeader, rootfsErr := rootfsDiffHeader.WaitWithContext(ctx)
+	if memfileErr != nil || rootfsErr != nil {
+		logger.L().Warn(ctx, "omitting scheduling metadata: diff header did not resolve",
+			zap.Error(errors.Join(memfileErr, rootfsErr)))
+
+		return nil
+	}
+
+	return scheduling.FromHeaders(memfileHeader, rootfsHeader)
```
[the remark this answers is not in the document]
ValentaTomas: Fixed: pause no longer waits on the async memfile-dedup header. Scheduling metadata is now derived synchronously from the resolved parent memfile header plus the new rootfs header (both available without dedup), so the async path is preserved. Also scoped the metadata down to ancestry-only (base, generation, union of referenced build IDs) and dropped the per-build byte arrays that were the only reason the resolved mapping was needed.

packages/orchestrator/pkg/scheduling/metadata.go
```diff
@@ -0,0 +1,96 @@
...
+		bBytes := b.memfileBytes + b.rootfsBytes
+		if aBytes == bBytes {
+			return cmp.Compare(a.buildID.String(), b.buildID.String())
+		}
+
+		return cmp.Compare(bBytes, aBytes)
+	})
+	if len(chain) > chainLimit {
+		chain = chain[:chainLimit]
+	}
```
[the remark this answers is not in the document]
ValentaTomas: Raised chainLimit to 64.

jakubno (approved):

PR #2874, opened 2026-05-30
Add rootfs reboot fallback for snapshots

ValentaTomas:
Add a rootfs reboot fallback so selected snapshot flows can skip memory restore while preserving normal resume behavior by default.


ValentaTomas:
Superseded by the smaller clean-branch PR #2875.

PR #2875, opened 2026-05-30
Add rootfs reboot fallback for snapshots

ValentaTomas:
## Summary
- Add a rootfs reboot fallback for paused snapshots when memory artifacts are skipped or unavailable.
- Allow snapshot creation to request disk-only persistence while keeping normal memory snapshots as the default.

## Tests
- go test ./internal/handlers ./internal/orchestrator ./internal/sandbox/...
- go test ./pkg/server ./pkg/sandbox ./pkg/template/metadata -run '^$'
- go test ./pkg/types
- go test ./internal/api


ValentaTomas:
We will start with a minimal https://github.com/e2b-dev/infra/pull/2877 addition.

PR #2878, opened 2026-05-31
fix(api): bound ResourceExhausted retries in sandbox placement

ValentaTomas:
## What

When all nodes report `ResourceExhausted`, keep retrying placement until the request deadline (as before), but wait a jittered, exponentially-growing backoff (capped) after each full pass over the nodes instead of spinning in a tight loop. Exhausted nodes are skipped within a pass so other nodes are tried first. Also log the give-up once (on deadline) instead of per attempt, and add `template_id`/`build_id` to placement failure logs.

## Why

When every node was at its starting-instance limit, the loop retried with no backoff, spinning until the request deadline. A few stuck creates amplified into thousands of log lines/sec and a ~390x allocation-rate spike on the API. Capacity exhaustion is usually transient, so this keeps the original "retry until deadline" behavior (avoiding spurious placement failures) while the backoff caps the work/log volume per request. The build/template id makes the offending build identifiable from logs.

### DIFF
--- a/packages/api/internal/orchestrator/placement/config.go
+++ b/packages/api/internal/orchestrator/placement/config.go
@@ -1,6 +1,16 @@
 package placement
 
+import "time"
+
 const (
 	maxRetries                  = 3
 	maxStartingInstancesPerNode = 3
+
+	// Bound placement retries when nodes report ResourceExhausted so a saturated
+	// fleet can't make PlaceSandbox spin until the request deadline. The backoff
+	// grows exponentially from base to max (with jitter) to avoid hammering the
+	// fleet under sustained capacity pressure.
+	maxResourceExhaustedRetries     = 8
+	resourceExhaustedBackoffBase    = 250 * time.Millisecond
+	resourceExhaustedBackoffMaxWait = 5 * time.Second
 )
--- a/packages/api/internal/orchestrator/placement/config.go
+++ b/packages/api/internal/orchestrator/placement/config.go
@@ -6,11 +6,11 @@ const (
 	maxRetries                  = 3
 	maxStartingInstancesPerNode = 3
 
-	// Bound placement retries when nodes report ResourceExhausted so a saturated
-	// fleet can't make PlaceSandbox spin until the request deadline. The backoff
-	// grows exponentially from base to max (with jitter) to avoid hammering the
-	// fleet under sustained capacity pressure.
-	maxResourceExhaustedRetries     = 8
+	// When every node reports ResourceExhausted, PlaceSandbox keeps retrying
+	// until the request deadline (capacity is usually transient), but waits a
+	// jittered backoff after each full pass over the nodes so a saturated fleet
+	// can't make it spin in a tight loop. The backoff grows exponentially from
+	// base to max.
 	resourceExhaustedBackoffBase    = 250 * time.Millisecond
 	resourceExhaustedBackoffMaxWait = 5 * time.Second
 )
--- a/packages/api/internal/orchestrator/placement/placement.go
+++ b/packages/api/internal/orchestrator/placement/placement.go
@@ -75,20 +75,18 @@ func PlaceSandbox(ctx context.Context, algorithm Algorithm, clusterNodes []*node
 			node, err = algorithm.chooseNode(ctx, clusterNodes, skip, nodemanager.SandboxResources{CPUs: sbxRequest.GetSandbox().GetVcpu(), MiBMemory: sbxRequest.GetSandbox().GetRamMb()}, buildMachineInfo, labelFilteringEnabled, requiredLabels)
 			if err != nil {
 				// No selectable node. If exhausted nodes are the only blocker,
-				// back off (bounded) and retry them; otherwise give up.
+				// back off and retry them until the request deadline (capacity is
+				// usually transient); otherwise give up.
 				if len(nodesExhausted) == 0 {
 					return nil, err
 				}
 
 				exhaustedRetries++
-				if exhaustedRetries >= maxResourceExhaustedRetries {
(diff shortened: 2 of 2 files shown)


packages/api/internal/orchestrator/placement/placement.go
```diff
@@ -42,6 +43,7 @@ func PlaceSandbox(ctx context.Context, algorithm Algorithm, clusterNodes []*node
 	}
 
 	attempt := 0
+	exhaustedRetries := 0
 	for attempt < maxRetries {
```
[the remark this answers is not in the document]
ValentaTomas: Done. Exhausted nodes are now tracked separately (`nodesExhausted`) and excluded from `chooseNode`, so other nodes are tried first; they are only retried after a bounded backoff once every candidate is exhausted.

packages/api/internal/orchestrator/placement/placement.go
```diff
@@ -103,7 +105,21 @@ func PlaceSandbox(ctx context.Context, algorithm Algorithm, clusterNodes []*node
...
+
 			logger.L().Warn(ctx, "Node exhausted, trying another node", logger.WithSandboxID(sbxRequest.GetSandbox().GetSandboxId()), logger.WithNodeID(failedNode.ID))
+
+			// Back off before retrying so we don't hot-spin (and flood logs)
+			// while the whole fleet is at capacity.
+			select {
+			case <-ctx.Done():
+				return nil, fmt.Errorf("request timed out while waiting for available capacity after %d exhausted retries", exhaustedRetries)
+			case <-time.After(resourceExhaustedBackoff):
+			}
```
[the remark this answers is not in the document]
ValentaTomas: Done, same fix as above: `nodesExhausted` are excluded from selection and only cleared/retried after a backoff when all candidates are exhausted.

packages/api/internal/orchestrator/placement/placement.go
```diff
@@ -53,11 +60,35 @@ func PlaceSandbox(ctx context.Context, algorithm Algorithm, clusterNodes []*node
...
+				for id := range nodesExhausted {
+					delete(skip, id)
+				}
+				nodesExhausted = make(map[string]struct{})
+
+				continue
 			}
 
-			node, err = algorithm.chooseNode(ctx, clusterNodes, nodesExcluded, nodemanager.SandboxResources{CPUs: sbxRequest.GetSandbox().GetVcpu(), MiBMemory: sbxRequest.GetSandbox().GetRamMb()}, buildMachineInfo, labelFilteringEnabled, requiredLabels)
+			node, err = algorithm.chooseNode(ctx, clusterNodes, skip, nodemanager.SandboxResources{CPUs: sbxRequest.GetSandbox().GetVcpu(), MiBMemory: sbxRequest.GetSandbox().GetRamMb()}, buildMachineInfo, labelFilteringEnabled, requiredLabels)
```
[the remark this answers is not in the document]
ValentaTomas: Fixed. The bounded backoff now triggers on the `chooseNode` no-node error path when exhausted nodes are the blocker (not a `len(skip) >= len(clusterNodes)` check), so a partial/label- or CPU-filtered pool still backs off and retries instead of failing immediately.

packages/api/internal/orchestrator/placement/placement.go
```diff
@@ -103,14 +139,25 @@ func PlaceSandbox(ctx context.Context, algorithm Algorithm, clusterNodes []*node
 		switch statusCode {
 		case codes.ResourceExhausted:
 			failedNode.PlacementMetrics.Skip(sbxRequest.GetSandbox().GetSandboxId())
-			logger.L().Warn(ctx, "Node exhausted, trying another node", logger.WithSandboxID(sbxRequest.GetSandbox().GetSandboxId()), logger.WithNodeID(failedNode.ID))
+			nodesExhausted[failedNode.ID] = struct{}{}
 		default:
 			nodesExcluded[failedNode.ID] = struct{}{}
 			failedNode.PlacementMetrics.Fail(sbxRequest.GetSandbox().GetSandboxId())
-			logger.L().Error(ctx, "Failed to create sandbox", logger.WithSandboxID(sbxRequest.GetSandbox().GetSandboxId()), logger.WithNodeID(failedNode.ID), zap.Int("attempt", attempt+1), zap.Error(utils.UnwrapGRPCError(err)))
+			logger.L().Error(ctx, "Failed to create sandbox", logger.WithSandboxID(sbxRequest.GetSandbox().GetSandboxId()), logger.WithNodeID(failedNode.ID), logger.WithTemplateID(sbxRequest.GetSandbox().GetTemplateId()), logger.WithBuildID(sbxRequest.GetSandbox().GetBuildId()), zap.Int("attempt", attempt+1), zap.Error(utils.UnwrapGRPCError(err)))
```
djeebus: Would be good to add some newlines here

djeebus (approved):

packages/api/internal/orchestrator/placement/placement.go
ValentaTomas: Done, wrapped the log fields onto separate lines.

packages/api/internal/orchestrator/placement/config.go
```diff
@@ -1,6 +1,16 @@
...
 const (
 	maxRetries                  = 3
 	maxStartingInstancesPerNode = 3
+
+	// When every node reports ResourceExhausted, PlaceSandbox keeps retrying
+	// until the request deadline (capacity is usually transient), but waits a
+	// jittered backoff after each full pass over the nodes so a saturated fleet
+	// can't make it spin in a tight loop. The backoff grows exponentially from
+	// base to max.
+	resourceExhaustedBackoffBase    = 250 * time.Millisecond
```
jakubno, line 14: this is super high, if sandbox hits this, the probability isn't high, but it can happen, start time will be affected by extra 250ms, we should definitely lower it

jakubno (dismissed):

packages/api/internal/orchestrator/placement/config.go
ValentaTomas: This only happens if you went through all the nodes in the cluster—at that point I think there should be relatively high backoff, because originally this really spammed the orchestrators and logs and entered in kinda ddos mode.

ValentaTomas: cc @jakubno when you can

ValentaTomas:
Folded into #2966 (single accumulating branch for the scheduling work).

PR #2882, opened 2026-06-01
fix(nfsproxy): only increment unmounts counter on successful chroot close

AdaAibaby:
OnNetworkRelease incremented [nfs.chroot.unmounts] unconditionally, even when [chroot.Close()] returned an error. This meant the difference between [nfs.chroot.mounts] and [nfs.chroot.unmounts] — which is supposed to reflect the number of active mount namespaces — was silently understated whenever a close failed, masking mount namespace leaks in metrics and alerting.

Fix by skipping the counter increment via continue when Close() returns an error. The warning log and map cleanup are unaffected.

A unit test is added to nfs_test.go that injects one successful and one failing Close() call and asserts the counter is incremented exactly once.

### DIFF
--- a/packages/orchestrator/pkg/nfsproxy/chroot/nfs.go
+++ b/packages/orchestrator/pkg/nfsproxy/chroot/nfs.go
@@ -109,7 +109,10 @@ func (h *NFSHandler) OnNetworkRelease(ctx context.Context, sbx *sandbox.Sandbox)
 				zap.String("path", chroot.Root()),
 				zap.Error(err),
 			)
+
+			continue
 		}
+
 		h.chrootUnmountsCounter.Add(ctx, 1)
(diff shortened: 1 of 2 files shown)


ValentaTomas:

AdaAibaby:
> Closing as stale for now—let's reopen if it becomes relevant again.

Hi @ValentaTomas, thanks for the note. Just to clarify: this PR fixes a metric accounting bug in packages/orchestrator/pkg/nfsproxy/chroot/nfs.go where nfs.chroot.unmounts was incremented even when chroot.Close() failed, masking potential mount namespace leaks. Since the underlying logic in nfs.go hasn't changed and the bug is still present, could we please reopen this to get the fix in? Happy to rebase if needed.

PR #2885, opened 2026-06-01
feat(dashboard-api): add template tags handlers

drankou:
Adds the dashboard-api read endpoints powering the per-template detail view: single-template read (with latest ready build + versioning/storage fields) and tag-grouped assignment history with keyset pagination, search, and sort.

## Endpoints

| Method | Path | Purpose |
|---|---|---|
| GET | `/templates/{templateID}` | Single-template detail (build, vCPU/RAM/disk, envd version, aliases) |
| GET | `/templates/{templateID}/tags/groups` | Ready tag groups with bounded per-tag history |
| GET | `/templates/{templateID}/tags/count` | Distinct ready tag count |
| GET | `/templates/{templateID}/tags/exists` | Whether a tag has any ready assignment |
| GET | `/templates/{templateID}/tags/{tag}/assignments` | Paginated ready assignments for one tag |

## Pagination & sort

- **Tag groups:** `sort ∈ {latest_desc (default), latest_asc, name_asc, name_desc}`, cursor `{sort}|{ts}|{tag}` (rejected on sort mismatch). `tagsLimit ≤ 100`, per-tag `assignmentLimit ≤ 25`.
- **Assignments:** cursor `assigned_at|assignment_id`, `limit ≤ 100`.
- **Search:** `^[a-z0-9._-]{0,64}$`, case-insensitive substring on tag.

## Performance at prod scale

All queries are scoped to a single template, but every assignment row is joined to `env_builds` by PK to evaluate `status_group = 'ready'` and read build timestamps — one random heap fetch per assignment. 

To address this:

**Covering index** (`20260612140000`):

```sql
CREATE INDEX CONCURRENTLY idx_env_builds_id_covering
    ON env_builds (id) INCLUDE (status_group, created_at, finished_at);
```

It carries exactly the three columns the tag queries need from `env_builds`, turning every per-assignment lookup into an **index-only scan**. Validated with `hypopg` on the replica: the planner adopts it in all query variants, with an identical plan shape, so the plan is parameter-value-independent and safe under prepared-statement plan caching. 

**Estimated size ~24GB** — The index build is non-blocking but will take tens of minutes during migration. If tag listing later moves to a projection table (one row per tag, maintained on write), this index becomes droppable.


packages/dashboard-api/internal/handlers/template_tag_assignments.go
```diff
@@ -0,0 +1,154 @@
...
+
+	return *limit
+}
+
+func parseTagAssignmentsCursor(cursor *api.TagAssignmentsCursor) (time.Time, uuid.UUID, error) {
+	defaultID := uuid.MustParse(maxCursorID)
+	if cursor == nil || *cursor == "" {
+		// Sentinel: future timestamp + max UUID returns the newest rows on the first page.
+		return time.Now().UTC().Add(time.Hour), defaultID, nil
+	}
```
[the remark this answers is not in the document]
drankou: removed, its unnecessary

packages/dashboard-api/internal/handlers/template_get.go
```diff
@@ -0,0 +1,74 @@
...
+		Public:        row.Env.Public,
+		Aliases:       row.Aliases,
+		Names:         row.Names,
+		CreatedAt:     row.Env.CreatedAt,
+		UpdatedAt:     row.Env.UpdatedAt,
+		LastSpawnedAt: row.Env.LastSpawnedAt,
+		SpawnCount:    row.Env.SpawnCount,
+		BuildCount:    row.Env.BuildCount,
+		EnvdVersion:   envdVersion,
+	})
```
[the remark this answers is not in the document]
drankou: addressed

packages/db/pkg/dashboard/sql_queries/templates/list_template_tag_groups_by_name_desc.sql
```diff
@@ -0,0 +1,50 @@
...
+    WHERE eba.env_id = sqlc.arg(template_id)::text
+      AND eb.status_group = 'ready'
+      AND (
+          sqlc.arg(search)::text = ''
+          OR strpos(eba.tag, sqlc.arg(search)::text) > 0
+      )
+    GROUP BY eba.tag
+    HAVING
+        sqlc.narg(cursor_tag)::text IS NULL
+        OR eba.tag < sqlc.narg(cursor_tag)::text
```
jakubno: why is this in HAVING?

packages/db/pkg/dashboard/sql_queries/templates/list_template_tag_groups_by_name_desc.sql
```diff
@@ -0,0 +1,50 @@
...
+    SELECT
+        eba.tag,
+        MAX(eba.created_at)::timestamptz AS latest_assigned_at
+    FROM public.env_build_assignments eba
+    JOIN public.env_builds eb ON eb.id = eba.build_id
+    WHERE eba.env_id = sqlc.arg(template_id)::text
+      AND eb.status_group = 'ready'
+      AND (
+          sqlc.arg(search)::text = ''
+          OR strpos(eba.tag, sqlc.arg(search)::text) > 0
```
jakubno, line 13: should tags be case sensitive?

drankou: effectively case-insensitive end-to-end; we lowercase at write time as well as lowercasing the search input before it reaches query

packages/db/pkg/dashboard/sql_queries/templates/list_template_tag_groups_by_name_desc.sql
drankou: no reason here, addressed

ben-fornefeld (approved):
lgtm, please do a final check in which places we can use shared utils like `utils.DerefOrDefault` as a final cleanup pass

Issue #2893, opened 2026-06-02
Cluster total capacity 9*500=4500 but reject create after only ~1200 sandboxes with "too many sandboxes starting on this node"


AdaAibaby:
**Problem Summary**
We have a cluster with 9 nodes, each configured max sandbox limit = 500, total cluster theoretical capacity: 9 × 500 = 4500 sandboxes.
But sandbox creation fails after only ~1200 running sandboxes, gRPC returns ResourceExhausted: too many sandboxes starting on this node, please retry.
No single node reaches its individual 500 quota, orchestrator incorrectly throttles global allocation early.
**Error Log**
SandboxService/Create/unary [ResourceExhausted]: finished call  {"service": "orchestrator_template-manager", "internal": true, "pid": 936536, "protocol": "grpc", "grpc.component": "server", "grpc.service": "SandboxService", "grpc.method": "Create", "grpc.method_type": "unary", "peer.address": "10.254.72.8:50292", "grpc.start_time": "2026-06-02T10:59:52+08:00", "grpc.request.deadline": "2026-06-02T11:01:02+08:00", "grpc.code": "ResourceExhausted", "grpc.error": "rpc error: code = ResourceExhausted desc = too many sandboxes starting on this node, please retry", "grpc.time_ms": "0.536", "trace_id": "c08e21fe0cc8487da4b3f6421aca931f", "span_id": "e50dfaf753372846"}
 
**Expected behavior**
Each node enforces limit = 500 only for active+starting sandboxes on itself;
Allow cluster to scale up to near 4500 total sandboxes before throttling;
Auto cleanup leaked orphan FC processes to release occupied quota.
**Reproduce steps**
Deploy 9 nodes, per-node max=500
Gradually create sandboxes until total ~1200
All new Create requests hit ResourceExhausted error immediately
/cc @jakubno

jakubno:
Hi @AdaAibaby , those are 2 different limits. Concurrently starting sandboxes on the node (`MaxStartingInstancesPerNode`) and max number of running sandboxes on the node (`MaxSandboxesPerNode`)

AdaAibaby:
8 nodes
per node 128c /1.5T
per sbx 4c32g
sbx number  ~1200

PR #2894, opened 2026-06-02
Release semaphore after sandbox stops, not when Create() returns.Fixes ResourceExhausted error when starting many sandboxes concurrently

AdaAibaby:
fix issue https://github.com/e2b-dev/infra/issues/2893

The `startingSandboxes` semaphore was acquired in `Create()` and released via `defer` when `Create()` returned. However, the sandbox continues running in a background goroutine managed by `setupSandboxLifecycle()`. This caused the semaphore slot to be freed before the sandbox actually stopped.

**Impact**: With a limit of 10, after 10 fast-starting sandboxes all released their semaphore slots, the 11th `Create()` call would fail with `ResourceExhausted` error, even though all 10 sandboxes were still running.

## Root Cause

The semaphore was designed to limit **starting** sandboxes, but was being released too early. It should limit **concurrently running** sandboxes instead.

## Solution

1. **Move semaphore release into `setupSandboxLifecycle()` goroutine**: Release happens after `sbx.Close()` completes, ensuring the slot is held for the entire sandbox lifetime.

2. **Add `semaphoreAcquired` flag**: Tracks whether the semaphore was actually acquired, preventing spurious releases in error paths.

3. **Introduce `proxyPool` interface**: Allows tests to inject a no-op proxy implementation without real networking.

4. **Add comprehensive unit tests**: Verify semaphore is held while sandbox runs, released after stop, and handles error cases correctly.

### `packages/orchestrator/pkg/server/main.go`
- Added `proxyPool` interface (contains only `RemoveFromPool` method)
- Changed `Server.proxy` field from `*proxy.SandboxProxy` to `proxyPool` interface

### `packages/orchestrator/pkg/server/sandboxes.go`
- Modified `Create()`: Removed `defer` release, added `semaphoreAcquired` flag
- Modified `setupSandboxLifecycle()`: New signature with `semaphoreAcquired bool` parameter, release moved to end of goroutine
- Modified `Checkpoint()`: Pass `true` for `semaphoreAcquired`

### `packages/orchestrator/pkg/sandbox/testutil.go` (NEW)
- Added `TestSandboxHandle` for controlling sandbox lifecycle in tests
- Added `NewTestSandbox()` helper function

### `packages/orchestrator/pkg/server/semaphore_lifecycle_test.go` (NEW)
- 4 unit tests covering:
  - Semaphore held while running, released after stop
  - No spurious release when flag=false
  - Limit enforcement across concurrent sandboxes
  - Release on error exit

## Testing

- ✅ All 4 new tests pass
- ✅ All existing server tests pass (no regression)
- ✅ Full build passes

## Verification

The fix ensures:
- Semaphore truly limits concurrently **running** sandboxes
- Error paths handled correctly
- Testable without real proxy/networking
- Backward compatible
/cc @jakubno

### DIFF
--- a/packages/orchestrator/pkg/sandbox/testutil.go
+++ b/packages/orchestrator/pkg/sandbox/testutil.go
@@ -1,0 +1,48 @@
+//go:build linux
+
+package sandbox
+
+import (
+	"github.com/e2b-dev/infra/packages/shared/pkg/utils"
+)
+
+// TestSandboxHandle is a handle returned by NewTestSandbox that lets tests
+// control the sandbox lifecycle (signal exit, trigger cleanup).
+// Only intended for use in tests.
+type TestSandboxHandle struct {
+	Sbx  *Sandbox
+	exit *utils.ErrorOnce
+}
+
+// SignalExit unblocks Sandbox.Wait() as if the sandbox process exited cleanly.
+func (h *TestSandboxHandle) SignalExit() {
+	_ = h.exit.SetSuccess()
+}
+
+// SignalExitWithError unblocks Sandbox.Wait() with an error.
+func (h *TestSandboxHandle) SignalExitWithError(err error) {
+	_ = h.exit.SetError(err)
+}
+
+// NewTestSandbox creates a minimal Sandbox suitable for unit tests.
+// The returned handle lets callers control when the sandbox "exits".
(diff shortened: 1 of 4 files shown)


jakubno:
As explained in the issue, it works as intended

PR #2902, opened 2026-06-02
perf(orch): free guest memory incrementally during snapshot copy

bchalios:
Summary

On pause we copy guest memory from the memfd into the snapshot cache while the memfd is still fully resident, holding ~2× guest memory for the duration. This adds the option to punch each range out of the memfd (MADV_REMOVE) right after it's copied, keeping peak at ~1×. Gated behind a new flag, off by default.

Commits

1. Memfd.punchHole — MADV_REMOVE over a huge-page-aligned range (misalignment is a caller bug → error; no clear() fallback since the mapping is PROT_READ).
2. memfd-punch-on-snapshot flag — BoolFlag, default false.
3. Wiring — thread punchSource through copyFromMemfd → NewCacheFromMemfd/Async → ExportMemory → pauseProcessMemory; Pause reads the flag.

Scope

- Memfd path only (the process_vm_readv path has nothing to punch).
- Non-dedup path only (dedup is page-granular, can't punch at huge-page granularity).

⚠️  Safety

The punch is destructive — it zeroes the pages for every mapping of the fd, including Firecracker's. A failed copy leaves both cache and memfd unusable, so it's only correct on the pause-and-discard path. Before enabling, confirm a failed pause discards the sandbox rather than resuming it. Rollback safety is enforced by convention, not code.

### DIFF
--- a/packages/orchestrator/pkg/sandbox/block/memfd.go
+++ b/packages/orchestrator/pkg/sandbox/block/memfd.go
@@ -99,6 +99,7 @@ func NewCacheFromMemfd(
 	filePath string,
 	memfd *Memfd,
 	dirty *roaring.Bitmap,
+	punchSource bool,
 ) (*Cache, error) {
 	ctx, span := tracer.Start(ctx, "export-memory-from-memfd",
 		trace.WithAttributes(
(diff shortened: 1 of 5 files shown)


ValentaTomas:

PR #2904, opened 2026-06-02
feat(dashboard-api): templates list pagination

drankou:
## Summary

Adds new dashboard-api endpoint backed by split per-sort keyset queries:

- `GET /templates` — paginated team templates list with filtering, search, and column sorting.

## Implementation notes

- **Split per-(column, direction) keyset queries** under `packages/db/pkg/dashboard/sql_queries/templates/list_team_templates.sql` — one query each for `name`, `cpu_count`, `memory_mb`, `created_at`, `updated_at` × asc/desc. Shared server-side filtering and `name`/`id` search are inlined into every variant.
- **Default templates** are appended via `env_defaults` inside the query when the team has no `cluster_id`, so a single round-trip returns both team and default rows.
- **Cursor format** `{sort}|{value}|{templateID}` is validated against the request's `sort` to prevent cross-sort cursor reuse.
- **Handler** (`templates_list.go`, `templates_cursor.go`) switches on the parsed `templatesSort` enum to pick the matching `ListTeamTemplatesBy*` query — no dynamic SQL.
- **`GetTeamTemplate`** single-row lookup added alongside, with the same default-template inlining.
- OpenAPI spec (`spec/openapi-dashboard.yml`) and regenerated `api.gen.go` included.

**Known follow-ups**:

- `createdBy.email` is intentionally omitted from the response until the Supabase user migration lands.


packages/db/pkg/dashboard/sql_queries/templates/list_team_templates.sql
ben-fornefeld, line 1: can you double check if these queries use indices correctly / if we should expect performance issues for teams with a lot of templates?

packages/dashboard-api/internal/handlers/utils_templates_list.go
ben-fornefeld, line 1: since this file seems like utilities only, please rename to the convention: `utils_templates_list.go`

packages/dashboard-api/internal/handlers/templates_list.go
ben-fornefeld, line 1: we are doing a lot of nil checks and have a lot of small helper functions here. two things:
- check out the shared utils package we have available inside this repo. we have utilities like `utils.DerefOrDefault` which you should consider.
- if you need to keep some utility functions, consider moving them to `utils_templates_list.go` as well.

packages/db/pkg/dashboard/sql_queries/templates/list_team_templates.sql
drankou: Added new composite index for `envs` : `(team_id, source, created_at DESC, id)` to improve performance on default sorting. + restructured query for efficiency

ben-fornefeld (approved):

PR #2907, opened 2026-06-03
fix(api): honor scheduling labels on resume placement

ValentaTomas:
## Summary
- Make resume placement skip an origin node that no longer matches current scheduling labels.
- Let labeled resume placement target the requested node pool even when the paused snapshot came from a different CPU model.

## Test plan
- mise exec -- go test ./packages/api/internal/orchestrator ./packages/api/internal/orchestrator/placement


ValentaTomas:
Closing for now as not blocking.

PR #2919, opened 2026-06-03
fix(p2p/compression) - upload header synchronization

levb:
### Problem

A descendant B's header doesn't reflect ancestor A's data until B finishes uploading. Resume B early — peer-served, mid-build — and reads of A through B's header can't pick the right path (uncompressed vs .zstd) or hit a peer that just transitioned. Result: silent wrong-bytes or ErrObjectNotExist against the wrong CT.

### Mechanism

Refresh loads A's own {A}/{filetype}.header sidecar, extracts A's full-file self-FT, reopens upstream at the resolved CT, and atomically stores {upstream, fullDiffFrameTable} into StorageDiff.source as one unit. Idempotent latch: concurrent calls short-circuit.

  Diff highlights

  - StorageDiff: source atomic pointer; lock-free resolve (4 branches); RefreshSource → reloadSource(cause) → reloadSourceLocked when need to refresh
  - File.createDiff: owns the flow for creating the diff including obtaining an authoritative Size.
  - peerSeekable.Size: returns PeerTransitionedError on any miss (no caller-FT means no safe CT guess).
  - PeerRouted marker for type-assert discrimination at the build/peerclient boundary.
  - header.LoadHeader → (*Header, int, error), storage.UncompressedFrameTable sentinel, RangeOpener interface with per-call FT.
  - Telemetry: orchestrator.storage.diff.frame_table_refresh (duration + bytes + count) attributed by cause × file_type.


packages/orchestrator/pkg/sandbox/build/cache.go
```diff
@@ -118,8 +121,24 @@ func (s *DiffStore) Close() {
...
+// pending eviction. Returns (nil, false) if the key isn't present. Use
+// Lookup for a side-effect-free probe (e.g. from IsCached).
+func (s *DiffStore) Get(key DiffStoreKey) (Diff, bool) {
+	s.resetDelete(key)
+	item := s.cache.Get(key)
+	if item == nil {
+		return nil, false
+	}
+
+	return item.Value(), true
```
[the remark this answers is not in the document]
levb: the comment was wrong

packages/orchestrator/pkg/sandbox/build/build.go
```diff
@@ -278,73 +314,190 @@ func (b *File) IsCached(ctx context.Context, off, length int64) bool {
...
 	}
 
-	source, err := b.store.Get(ctx, storageDiff)
+	upstream, err := b.openUpstream(ctx, buildID, objType, ft.CompressionType())
 	if err != nil {
-		return nil, fmt.Errorf("failed to get build from store: %w", err)
+		return nil, 0, nil, err
+	}
+
+	return upstream, bd.Size, ft, nil
```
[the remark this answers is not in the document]
levb: a freshly loaded (v4+) header must not miss the self entry

packages/orchestrator/pkg/sandbox/build/storage_diff.go
```diff
@@ -153,3 +179,138 @@ func (b *StorageDiff) IsCached(ctx context.Context, off, length int64) bool {
...
+// (PeerTransitionedError) and resolve (read-time peer-left fallback) funnel
+// through it; the cause attribute distinguishes them in telemetry. A concurrent
+// caller that wins the mutex short-circuits when the latch is already
+// populated, so parallel segment reads on a fresh StorageDiff pay only one
+// header fetch.
+func (b *StorageDiff) reloadSource(ctx context.Context, cause string) error {
+	b.refreshMu.Lock()
+	defer b.refreshMu.Unlock()
+	if b.source.Load().fullDiffFrameTable != nil {
+		return nil
```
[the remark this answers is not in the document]
levb: As I understand it, the scenario described by the bot would require that we switch from non-peer to peer for a layer; we DON'T.

packages/orchestrator/pkg/sandbox/build/build.go
```diff
@@ -278,73 +315,194 @@ func (b *File) IsCached(ctx context.Context, off, length int64) bool {
...
+//   - peer-routed wrapper, PeerTransitionedError → (0, false, nil)  caller refreshes
+//   - peer-routed wrapper, peer RPC failure → (0, false, err)
+//   - NOT peer-routed (resolveProvider cleared between IsActive probe and
+//     OpenSeekable) → (0, false, nil)  caller refreshes
+//
+// Symmetric with readSegment's PeerTransitionedError handling on the read path:
+// peer says "go to storage" → refresh authoritative header → continue.
+// No 404-driven recovery.
+func initialSize(ctx context.Context, upstream storage.Seekable) (size int64, ok bool, err error) {
+	if _, peerRouted := upstream.(peerclient.PeerRouted); !peerRouted {
```
dobrac, line 500: can we avoid the cast "upstream.(peerclient.PeerRouted)" or not really?

levb: This **is** the signal - the upstream object was just resolved, and it's either a peerRouted or not... how else would I figure it out? We could add a return value and call a method but it'd be redundant.

packages/orchestrator/pkg/sandbox/build/build.go
```diff
@@ -278,73 +315,194 @@ func (b *File) IsCached(ctx context.Context, off, length int64) bool {
...
 	if transErr.RetryAfter > 0 {
 		timer := time.NewTimer(transErr.RetryAfter)
 		defer timer.Stop()
-
 		select {
 		case <-timer.C:
 		case <-ctx.Done():
-			return false, ctx.Err()
+			return ctx.Err()
 		}
```
djeebus, line 307: tiny syntactic improvement, but you can use `time.After(duration)` instead of creating and managing a timer yourself:

```golang
select {
case <-time.After(transErr.RetryAfter):
case <-ctx.Done():
	return false, ctx.Err()
}
```

packages/orchestrator/pkg/sandbox/build/build.go
```diff
@@ -278,73 +315,194 @@ func (b *File) IsCached(ctx context.Context, off, length int64) bool {
...
+	if upstream == nil {
+		var err error
+		upstream, err = b.openUpstream(ctx, buildID, objType, initialCT)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if size == 0 {
+		// (d) and degenerate (a) where bd.Size was zero. Ask storage directly.
```
djeebus, line 419: `(d)` and `(a)`?

packages/orchestrator/pkg/sandbox/build/storage_diff.go
```diff
@@ -128,28 +146,146 @@ func (b *StorageDiff) CachePath(context.Context) (string, error) {
...
+	timer.Success(ctx, int64(bytesLoaded))
 
-	return b.chunker.IsCached(ctx, off, length)
+	return h, nil
+}
+
+// resolve picks the (upstream, ft) the next read should use, given the
+// caller's per-mapping FT hint. The contract: if there is no authoritative FT
+// latched AND no peer currently serving this build, we MUST refresh before
+// reading. The latched upstream was opened at the bootstrap-guessed CT path
```
djeebus, line 194: "CT"?

packages/orchestrator/pkg/sandbox/build/storage_diff.go
```diff
@@ -128,28 +146,146 @@ func (b *StorageDiff) CachePath(context.Context) (string, error) {
...
+// RefreshSource reloads the build's header, latches the authoritative FT, and
+// reopens upstream at the resulting CT path. Called by readSegment after a
+// PeerTransitionedError. Idempotent: once the source latch is populated, the
+// post-refresh upstream is base-routed (no peer wrapper) and cannot emit
+// further PeerTransitionedErrors, so a second call is a no-op.
+func (b *StorageDiff) RefreshSource(ctx context.Context) error {
+	return b.reloadSource(ctx, refreshCausePeerTransitioned)
+}
+
+// reloadSource is the idempotent ensure-latched entry. Both RefreshSource
```
djeebus, line 228: "ensure-latched"?

packages/orchestrator/pkg/sandbox/build/storage_diff.go
```diff
@@ -128,28 +153,128 @@ func (b *StorageDiff) CachePath(context.Context) (string, error) {
...
+// header fetch.
+func (b *StorageDiff) reloadSource(ctx context.Context, cause string) error {
+	b.refreshMu.Lock()
+	defer b.refreshMu.Unlock()
+	if b.source.Load().fullDiffFrameTable != nil {
+		return nil
+	}
+
+	return b.reloadSourceLocked(ctx, cause)
+}
```
dobrac, line 242: nit: do we need both these functions?

packages/orchestrator/pkg/sandbox/build/storage_diff.go
```diff
@@ -128,28 +153,128 @@ func (b *StorageDiff) CachePath(context.Context) (string, error) {
...
+// returning a still-uploading peer's incomplete header. Treating it as "no
+// FrameData = uncompressed" would silently corrupt reads of a compressed file.
+// Fail loudly; the read path will retry when the peer transitions and the
+// storage-authoritative header is available.
+//
+// V3 never reaches reloadSourceLocked: getBuild's V3 branch latches an
+// authoritative empty &{} FT at construction, so resolve short-circuits and
+// reloadSource is never called; V3 builds aren't peer-routed so
+// PeerTransitionedError never fires against them either.
+func (b *StorageDiff) reloadSourceLocked(ctx context.Context, cause string) error {
```
dobrac, line 258: this looks to be called only in reloadSource, making merge of the 3 functions to 1

dobrac (approved):

dobrac (approved):

dobrac (approved):

packages/orchestrator/pkg/sandbox/build/build.go
```diff
@@ -283,73 +320,194 @@ func (b *File) IsCached(ctx context.Context, off, length int64) bool {
...
+	case hasEntry:
+		// Our header has a Builds entry for this ancestor. Do NOT latch
+		// bd.FrameData as the StorageDiff's authoritative full-file FT — it is
+		// filtered to only the frames our header references, not the ancestor's
+		// full table.
+		size = bd.Size
+		initialCT = bd.FrameData.CompressionType()
+
+	case h.Metadata.Version >= header.MetadataVersionV4:
+		peerActive := b.store.isActivePeer != nil && b.store.isActivePeer(buildID.String())
```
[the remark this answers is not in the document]
levb: This is a rare case, and it self-heals, just fetches the header over P2P an extra time.

packages/orchestrator/pkg/sandbox/build/storage_diff.go
levb: I'll send a separate PR for this; agreed.

PR #2930, opened 2026-06-04
feat(envd): add optional EntryInfo to watch FilesystemEvent

mishushakov:
Adds an `include_entryinfo` flag to `WatchDirRequest` and `CreateWatcherRequest`; when set, each `FilesystemEvent` carries the affected entry's `EntryInfo` (new `optional EntryInfo entry` field). The entry is populated best-effort and left `nil` for remove/rename-away events where the path no longer exists, so the watch stream/watcher is never torn down by a stat of a gone file. Both the streaming `WatchDir` and the polling `CreateWatcher`/`GetWatcherEvents` paths are supported, with regenerated proto code in `packages/envd` and `packages/shared`. Bumps the envd version to `0.6.2` and adds `watch_test.go` covering entry-present, entry-absent, and removed-entry cases.


packages/envd/internal/services/filesystem/utils.go
```diff
@@ -65,6 +67,26 @@ func entryInfo(path string) (*rpc.EntryInfo, error) {
...
+// Entry info is best-effort: a nil entry is returned (and the watch keeps running) when
+// it cannot be retrieved. A missing entry is expected for remove and rename-away events,
+// so it is not logged; any other failure is logged at warn level.
+func eventEntryInfo(logger *zerolog.Logger, path string) *rpc.EntryInfo {
+	entry, err := entryInfo(path)
+	if err != nil {
+		var connectErr *connect.Error
+		if errors.As(err, &connectErr) && connectErr.Code() == connect.CodeNotFound {
+			return nil
+		}
```
mishushakov: does it matter if it's connectErr or not - the outcome is the same

mishushakov: Good catch — `entryInfo` always returns a `connect.Error`, so the `errors.As` was always true and only the code mattered. Simplified to `connect.CodeOf(err) != connect.CodeNotFound` (still suppresses the log for the expected remove/rename-away case) in 9449fdc77.

matthewlouisbrockman (approved):
may have conflict on the envd 6.2 with another pr so whoever loses the race to merge'll need to get fixed

PR #2934, opened 2026-06-05
feat(auth): support admin token team auth

ben-fornefeld:
## Summary
- add AdminTeamAuth over X-Team-ID for admin-token authenticated requests
- populate the existing team context from admin token + team ID
- update generated API code and spec coverage


tvi (approved):
Double check 
AdminTeamAuth is alphabetically before AdminTokenAuth

tvi (approved):

PR #2942, opened 2026-06-06
feat(envd): add WebSocket raw TCP tunnel endpoint

mishushakov:
Adds a `GET /tunnel` WebSocket endpoint to envd that tunnels raw TCP to a target `host:port` inside the sandbox.

## How it works

- The target is passed via the `host`/`port` query params, validated and dialed (10s timeout) **before** the upgrade, so bad params return `400` and connection failures return `502` as ordinary HTTP responses rather than over a half-open WebSocket.
- After the `101` upgrade the connection becomes a verbatim bidirectional byte relay (binary frames, no framing or prefixes) implemented as a pair of `io.Copy` calls over `websocket.NetConn`. When either direction ends, both sides are torn down so the other `io.Copy` unblocks.
- A 30s keepalive ping (with a 5s pong timeout) keeps the tunnel warm through the upstream proxies' idle timeouts. A failed ping closes the connection immediately so the relay tears down rather than lingering until an upstream idle timeout.
- The default host is `localhost`; any host is allowed since the sandbox is the user's own environment.

## Auth & spec

Authentication reuses the existing `X-Access-Token` check (`AccessTokenAuth`), and the endpoint is defined in `spec/envd.yaml` and generated (`api.gen.go`) like every other envd route.

## Other

- Unit tests cover the echo relay, host defaulting, and the auth/400/502 paths.
- Bumps the envd version to `0.6.7`.
- Uses `github.com/coder/websocket` (already vendored elsewhere in the repo at v1.8.13).


ValentaTomas:
@mishushakov Can you check how big the websocket library is, just to be sure?

mishushakov:
@ValentaTomas according to Claude:

The WebSocket library used in envd is **`github.com/coder/websocket v1.8.13`**.

## Size

| Metric | Value |
|---|---|
| On-disk module (incl. tests, docs) | **344 KB** |
| Non-test Go source | **~4,470 lines** across 29 files |
| External dependencies | **None** |

The `vendor/golang.org/x/...` entries in the dependency list are part of Go's own standard library (vendored into `net/http`, `crypto`, etc.), not dependencies of the websocket package itself. `coder/websocket` (formerly `nhooyr.io/websocket`) is notably dependency-free — it's a minimal, idiomatic implementation that builds directly on the stdlib.

So it's a very lightweight addition: a small, self-contained library with no transitive third-party dependencies pulled into the envd build.

mishushakov:
Note for PR #3047

Any hijacked connection inherits *writeTimeoutConn, so (a) every WS frame write gets a 60s idle deadline, and (b) a type-assert to *net.TCPConn in a WS library will now fail (the embed only preserves the net.Conn interface, not the concrete type).

mishushakov:
Putting this back to drafts. Tomas's feedback was that we need to use a protocol like QUIC that we can reconnect ourselves under unreliable network conditions

ValentaTomas:

PR #2949, opened 2026-06-08
fix(cache): use 512-byte units for stat.Blocks in FileSize

levb:
stat.Blocks is reported in fixed 512-byte units per POSIX, not in the filesystem's optimal I/O block size. Multiplying by statfs Bsize (4096 on ext4) inflated the on-disk size by ~8x. Drop the Statfs call and multiply by the constant 512.

Verified locally that stat.Blocks * 512 matches the actual on-disk allocation while stat.Blocks * statfs.Bsize does not on ext4.

### DIFF
--- a/packages/orchestrator/pkg/sandbox/block/cache.go
+++ b/packages/orchestrator/pkg/sandbox/block/cache.go
@@ -661,18 +661,15 @@ func (c *Cache) WriteZeroesAt(off, length int64) (int, error) {
 // The size might differ from the dirty size, as it may not be fully on disk.
 func (c *Cache) FileSize(_ context.Context) (int64, error) {
 	var stat syscall.Stat_t
-	err := syscall.Stat(c.filePath, &stat)
-	if err != nil {
+	if err := syscall.Stat(c.filePath, &stat); err != nil {
 		return 0, fmt.Errorf("failed to get file stats: %w", err)
 	}
 
-	var fsStat syscall.Statfs_t
-	err = syscall.Statfs(c.filePath, &fsStat)
-	if err != nil {
-		return 0, fmt.Errorf("failed to get disk stats for path %s: %w", c.filePath, err)
-	}
+	// Per POSIX (and Linux man 2 stat), stat.Blocks is always reported in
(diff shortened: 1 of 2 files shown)


ValentaTomas (approved):

ValentaTomas:
One note—we use XFS for the actual cache in prod. Is that still the case with it?

levb:
> One note—we use XFS for the actual cache in prod. Is that still the case with it?

The [man page](https://man7.org/linux/man-pages/man3/stat.3type.html), [stack overflow](https://unix.stackexchange.com/questions/669543/on-what-unix-like-system-filesystem-is-the-st-blocks-field-returned-by-stat) the AI - all agree that 512 is a hardcoded, Linux-specific constant, not a property of the file system. 

man page:
```
    .st_blocks
              This field indicates the number of blocks allocated to the
              file, in 512-byte units.  (This may be smaller than
              .st_size/512 when the file has holes.)
```

From my cluster's orchestrator (XFS confirmed)
```
lev@e2b-orch-client-3sz2:~$ sudo dd if=/dev/zero of=/orchestrator/_512test bs=1M count=10 conv=fsync
10+0 records in
10+0 records out
10485760 bytes (10 MB, 10 MiB) copied, 0.01238 s, 847 MB/s
lev@e2b-orch-client-3sz2:~$ stat /orchestrator/_512test
  File: /orchestrator/_512test
  Size: 10485760        Blocks: 20480      IO Block: 4096   regular file
Device: 9,0     Inode: 1157109     Links: 1
Access: (0644/-rw-r--r--)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2026-06-09 16:44:02.001133545 +0000
Modify: 2026-06-09 16:44:02.006133891 +0000
Change: 2026-06-09 16:44:02.006133891 +0000
 Birth: 2026-06-09 16:44:02.001133545 +0000
lev@e2b-orch-client-3sz2:~$ du -B1 /orchestrator/_512test
10485760        /orchestrator/_512test
lev@e2b-orch-client-3sz2:~$ 
```
Notice **Blocks: 20480** for the 10Mb file despite the **IO Block** size

PR #2958, opened 2026-06-09
feat(orchestrator): add capacity limit and memory-pressure eviction to template mmap cache

AdaAibaby:
**Changes Summary**
model.go

1.Add environment variable TEMPLATE_CACHE_MAX_ENTRIES (integer, default 0 = unlimited)
  When a positive value is set, apply ttlcache.WithCapacity.
  Cache evicts entries by LRU policy once the entry limit is reached.
2.Add environment variable TEMPLATE_CACHE_MIN_FREE_MEMORY_MB (int64, default 0 = disabled)
  When enabled, start a background goroutine to poll host memory via unix.Sysinfo every 5 seconds.
  Continuously evict LRU entries one by one until free memory is above the configured threshold.

cache.go
1.In NewCache, conditionally enable ttlcache.WithCapacity according to the new cache config.
2.In Cache.Start, launch the startMemoryPressureEviction goroutine if the memory threshold is non-zero.
  Eviction logic: select entries with the earliest ExpiresAt timestamp (TTL resets on each access).
  Stop eviction when memory pressure is relieved or the cache is empty.

**Backward Compatibility**
Both new environment variables use 0 as default, which fully keeps the original runtime behavior.
Existing cloud deployments require no extra configuration or code changes.


jakubno:
We don't see any value in supporting limiting number of items and MB and percent are basically equivalent, so we won't accept this contribution

PR #2974, opened 2026-06-10
uffd: unregister empty and removed huge pages from UFFD missing pages tracking

bchalios:
## What

Unregister huge pages from UFFD in two cases:
* At sandbox start-up: all pages for all ranges with backing build Id that points to `nil`
* On demand: pages for which we receive a UFFD_EVENT_REMOVE event

Also, add tests that ensure that everything works as expected

## Why

UFFD handling for known zero pages (ranges with nil block ID) and pages that the guest has deallocated essentially boils down to providing a zero page. This is adds overhead in page fault handling (round-trip: `guest -> host kernel ->orchestrator -> host kernel -> guest`). Un-registering those pages from UFFD trims down the round trip to: `guest -> host kernel -> guest`.

This introduces real overhead to page fault handling. Using a simple `dd`-like command that touches 1GiB on my laptop shows a throughput of ~1.3GiB/sec on `main` vs ~2.0GiB/sec with the changes introduced here.

## How

We know the initially empty pages, so we unregister them before starting the sandbox. We also unregister all pages for which we receive a UFFD_EVENT_REMOVE. The catch is we use UFFD asynchronous write protection to track dirty pages. So, after unregistering the pages we re-register them, this time using only `UFFDIO_REGISTER_MODE_WP`. This way we keep accurate dirty-page info without handling zero pages.

**Note**: this doesn't work with anonymous 4K pages. As, [userfaultfd documentation](https://docs.kernel.org/admin-guide/mm/userfaultfd.html#write-protect-notifications) mentions:

> For anonymous memory, ioctl(UFFDIO_WRITEPROTECT) will ignore none ptes (e.g. when pages are missing and not populated). For file-backed memories like shmem and hugetlbfs, none ptes will be write protected just like a present pte. In other words, there will be a userfaultfd write fault message generated when writing to a missing page on file typed memories, as long as the page range was write-protected before. Such a message will not be generated on anonymous memories by default.

So, at the moment, this behaviour is restricted in 4K pages only

### DIFF
--- a/packages/orchestrator/pkg/sandbox/uffd/uffd.go
+++ b/packages/orchestrator/pkg/sandbox/uffd/uffd.go
@@ -204,6 +204,13 @@ func (u *Uffd) handle(ctx context.Context, sandboxId string, fdExit *fdexit.FdEx
 		u.memfd.Store(memfd)
 	}
 
+	// Drop MISSING tracking for the snapshot's empty ranges before the guest can
+	// fault, so the kernel serves those zero hugepages directly while WP-async
+	// still catches writes.
+	if err = uffd.DropMissingForEmptyRanges(ctx, u.memfile.Header()); err != nil {
+		return fmt.Errorf("failed to drop MISSING tracking for empty ranges: %w", err)
+	}
+
 	u.handler.SetValue(uffd)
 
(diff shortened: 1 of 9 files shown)


ValentaTomas:

PR #2977, opened 2026-06-11
perf(orchestrator): coalesce adjacent frame fetches into ranged reads (sketch)

ValentaTomas:
Sketch for discussion: read-side fetch coalescing, orthogonal to the dedup budgets in #2862.

Today every cold frame miss is one remote round trip, and a read spanning several frames fetches them serially. This PR:

- Chunker: with `chunker-coalesce-frames > 1`, adjacent uncached frames merge into one fetch session per run (capped by the flag), and all sessions a read needs are started before waiting, so disjoint runs fetch concurrently. In-flight sessions are joined; cached frames split runs.
- Seekable cache: multi-frame requests are served with one upstream ranged read per compressed-contiguous group of missing frames. Each frame is still decompressed through its own bounded sub-reader and written back to its own `.frm` file - the NFS cache layout, hit path, and write-back are unchanged.

Flag default `1` = byte-for-byte current behavior. Read path only; worst case is a larger ranged GET, never different data.

Known gaps (sketch, merge-blockers tracked in EN-42): **no cap on concurrent run fetches — disjoint uncached runs all start in parallel, so simultaneous cold resumes on one node can fan out to upstream/NFS without limit; must be bounded by a flag-sized semaphore (and consider sharing a node budget with the prefetcher) before this can be enabled anywhere**. Also: the run path skips the per-frame slab timer and NFS gauge wrappers of the single-frame path; no gap tolerance yet (a cached frame always splits a run); no integration bench. Sim in #2862 predicts ~2-3x fewer round trips for fragmented restores.

Tracking: [EN-42](https://linear.app/e2b/issue/EN-42/read-side-fetch-coalescing-for-cold-sandbox-restores)

### DIFF
--- a/packages/orchestrator/pkg/sandbox/block/streaming_chunk.go
+++ b/packages/orchestrator/pkg/sandbox/block/streaming_chunk.go
@@ -95,23 +95,34 @@ func (c *Chunker) Slice(ctx context.Context, off, length int64, ft *storage.Fram
 		return nil, fmt.Errorf("failed read from cache at offset %d: %w", off, err)
 	}
 
-	// Fetch every chunk the range spans (one fetch session per chunk).
-	end := off + length
-	for cur := off; cur < end; {
-		chunkOff, chunkLen, lerr := c.locateChunk(cur, ft)
-		if lerr != nil {
+	// Coalesced path: merge adjacent uncached frames into single fetch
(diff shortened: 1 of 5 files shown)


ValentaTomas:
Temporarily closing.

PR #2984, opened 2026-06-11
feat(orchestrator): honor ALLOW_SANDBOX_INTERNAL_CIDRS in TCP egress firewall

jakubno:
Follow up to #2529

The infra-level internal allowlist (`ALLOW_SANDBOX_INTERNAL_CIDRS`) was only applied to the per-slot nftables predefined allow set, so it covered non-TCP traffic but was ignored by the TCP firewall proxy. TCP connections to whitelisted internal IPs were still blocked by user deny rules (e.g. `denyOut` `0.0.0.0/0`), and domain-resolved internal IPs were always hard-blocked with no override.

- parse the allowlist once at proxy startup and plumb it into the connection handlers
- check it at highest priority in isEgressAllowed, mirroring the nftables rule order (predefined allow before user rules); checking it before domain rules also prevents proxy-bound connections from being dialed directly to the Host/SNI domain, bypassing the internal proxy
- exempt whitelisted IPs from the always-denied internal ranges in proxyWithIPVerification
- preparse DeniedSandboxCIDRs into DeniedSandboxIPNets at init instead of re-parsing per connection (panics on malformed entries instead of silently skipping them)
- add MatchTypeInternal metric label for these decisions

### DIFF
--- a/packages/orchestrator/pkg/tcpfirewall/proxy.go
+++ b/packages/orchestrator/pkg/tcpfirewall/proxy.go
@@ -47,12 +47,10 @@ type Proxy struct {
 	proxy      *tcpproxy.Proxy
 }
 
-func New(logger logger.Logger, networkConfig network.Config, sandboxes *sandbox.Map, meterProvider metric.MeterProvider, featureFlags *featureflags.Client) *Proxy {
+func New(logger logger.Logger, networkConfig network.Config, sandboxes *sandbox.Map, meterProvider metric.MeterProvider, featureFlags *featureflags.Client) (*Proxy, error) {
 	internalAllowedNets, err := sandbox_network.ParseCIDRs(networkConfig.AllowSandboxInternalCIDRs)
 	if err != nil {
-		logger.Error(context.Background(), "failed to parse ALLOW_SANDBOX_INTERNAL_CIDRS for TCP firewall",
-			zap.Strings("cidrs", networkConfig.AllowSandboxInternalCIDRs),
-			zap.Error(err))
+		return nil, fmt.Errorf("error parsing internal allowed CIDRs: %w", err)
 	}
 
 	p := &Proxy{
@@ -75,7 +73,7 @@ func New(logger logger.Logger, networkConfig network.Config, sandboxes *sandbox.
 
 	sandboxes.Subscribe(p)
 
-	return p
+	return p, nil
 }
 
 func (p *Proxy) OnInsert(_ context.Context, _ *sandbox.Sandbox) {}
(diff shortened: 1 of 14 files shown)


packages/shared/pkg/sandbox-network/firewall.go
```diff
@@ -91,6 +95,46 @@ func IsSpecifiedIPOrCIDR(s string) bool {
...
+			}
+
+			nets = append(nets, &net.IPNet{IP: ip, Mask: net.CIDRMask(bits, bits)})
+
+			continue
+		}
+
+		_, ipNet, err := net.ParseCIDR(entry)
+		if err != nil {
+			return nil, fmt.Errorf("invalid CIDR %q: %w", entry, err)
```
wj-e2b, line 121: Any error drops the whole list, this should be continue.

wj-e2b (changes requested):

packages/shared/pkg/sandbox-network/firewall.go
```diff
@@ -91,6 +95,46 @@ func IsSpecifiedIPOrCIDR(s string) bool {
...
+				ip = v4
+				bits = 8 * net.IPv4len
+			}
+
+			nets = append(nets, &net.IPNet{IP: ip, Mask: net.CIDRMask(bits, bits)})
+
+			continue
+		}
+
+		_, ipNet, err := net.ParseCIDR(entry)
```
wj-e2b, line 119: Also should maybe call TrimSpace here to handle extra spaces

packages/shared/pkg/sandbox-network/firewall.go
jakubno: Why? It doesn't say it's best effort.

This is run during init and will panic if you provide non parse-able data. This is on purpose to prevent some kind of unnoticed misconfiguration

wj-e2b: This is a fair callout, but I saw this piece of code where it wasn't fatal so it wasn't clear to me.

https://github.com/e2b-dev/infra/pull/2984/changes#diff-254503bd517b1fa86124028d5c39d5cb307ee34ea38bb14ddb357c82f10a2453R51

packages/orchestrator/pkg/tcpfirewall/proxy.go
```diff
@@ -38,20 +39,32 @@ type Proxy struct {
...
+	// even when they fall into the always-denied private ranges or user deny rules.
+	internalAllowedNets []*net.IPNet
+
 	proxyRules []proxyRule
 	proxy      *tcpproxy.Proxy
 }
 
-func New(logger logger.Logger, networkConfig network.Config, sandboxes *sandbox.Map, meterProvider metric.MeterProvider, featureFlags *featureflags.Client) *Proxy {
+func New(ctx context.Context, logger logger.Logger, networkConfig network.Config, sandboxes *sandbox.Map, meterProvider metric.MeterProvider, featureFlags *featureflags.Client) *Proxy {
+	internalAllowedNets, err := sandbox_network.ParseCIDRs(networkConfig.AllowSandboxInternalCIDRs)
```
jakubno, line 51: Fix this

packages/shared/pkg/sandbox-network/firewall.go
jakubno: Good catch, thanks for noticing! i wanted to use already parsed data there

wj-e2b (approved):
Will approve for now.

PR #2988, opened 2026-06-11
fix(tests): deflake TestSandboxMemoryIntegrity with explicit resume timeout

ValentaTomas:
`TestSandboxMemoryIntegrity/tmpfs_hash` flakes ~52% on main. Service logs from failed runs show it is never an integrity failure (no hash mismatch ever): the resume call passes an empty body, so the sandbox gets the 15s `SandboxTimeoutDefault`. Hashing the ~480MB tmpfs after resume faults the whole file back through UFFD, which takes >15s on the compressed/dedup shards (zstd1 fails 12/15 recent main runs, lz4 4/15, uncompressed 0/15), so the sandbox is evicted with `kill_reason: timeout` mid-check and every retry gets 502 until the 3min `EventuallyWithT` window expires.

Pass an explicit 300s timeout on resume, matching the sandbox creation timeout.

### DIFF
--- a/tests/integration/internal/tests/orchestrator/sandbox_object_not_found_test.go
+++ b/tests/integration/internal/tests/orchestrator/sandbox_object_not_found_test.go
@@ -22,7 +22,10 @@ func TestSandboxObjectNotFound(t *testing.T) {
 
 	client := setup.GetOrchestratorClient(t, ctx)
 
-	for range 10 {
+	// Under full host load (parallel 100-sandbox tests) the orchestrator can
+	// report ResourceExhausted for minutes before template validation is
+	// reached, so keep the retry window generous.
+	for range 36 {
 		_, err := client.Create(ctx, &orchestrator.SandboxCreateRequest{
 			Sandbox: &orchestrator.SandboxConfig{
 				TemplateId:          "nonexistent-template-id",
@@ -67,6 +70,6 @@ func TestSandboxObjectNotFound(t *testing.T) {
 		return
 	}
 
-	t.Log("failed to create sandbox after 10 retries")
+	t.Log("failed to create sandbox after 36 retries")
 	t.FailNow()
 }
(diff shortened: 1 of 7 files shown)


ValentaTomas:
Added two more flakes from the same family (deadlines too tight for CI load): TestSandboxListPaginationRunningLargerLimit creates 12 sandboxes with the default 30s timeout, so they expire before the list assertions finish (failed 5x in last 15 main runs, always "has 0 of 12"); TestSandboxAutoResumeViaProxy and TestEnvdAccessTokenAutoResumeViaProxy trigger a full snapshot resume through a single HTTP request with a 10s client timeout (failed 4x/2x, always context deadline exceeded).

ValentaTomas:
Covered the remaining recent flakes:

- `TestCommandKillNextApp` / `TestMaskRequestHostAPIParameter`: stream EOFs during heavy `npx create-next-app` / `apt-get install`. Guest kernel logs show soft lockups (`watchdog: BUG: soft lockup ... [npm install]` and `[envd]`, stuck in `clear_page_erms`) — page-fault storms through UFFD wedge envd itself under host load. Moved the npm install into a cached template build (same pattern as #2857) and replaced the netcat apt-get install with the preinstalled python.
- `TestSandboxObjectNotFound`: orchestrator returns ResourceExhausted before template validation while the host is full; 10x5s retries were not enough, extended the window.

The envd soft-lockup under UFFD restore load is a real symptom worth tracking separately; these changes only stop the tests from tripping over it.

tests/integration/internal/tests/orchestrator/sandbox_object_not_found_test.go
```diff
@@ -22,7 +22,10 @@ func TestSandboxObjectNotFound(t *testing.T) {
 
 	client := setup.GetOrchestratorClient(t, ctx)
 
-	for range 10 {
+	// Under full host load (parallel 100-sandbox tests) the orchestrator can
+	// report ResourceExhausted for minutes before template validation is
+	// reached, so keep the retry window generous.
+	for range 36 {
```
levb, line 28: 42 is a better number! 36 is not the Answer to the Ultimate Question of Life, the Universe, and Everything

levb (approved):

tests/integration/internal/tests/orchestrator/sandbox_object_not_found_test.go
ValentaTomas: Done — 42 it is.

PR #2990, opened 2026-06-12
refactor(orchestrator): Refactor orchestrator shutdown to not leak resources

wj-e2b:
- Deny new sandbox/template creation on drain
- Fix leaking network resources
- Add force drain


jakubno (changes requested):
@wj-e2b could we separate it to multiple PRs? 🙏🏻 happy to review afterwards

PR #2995, opened 2026-06-12
Filesystem only snapshots

bchalios:
### What
Adds an opt-in snapshot mode that persists **only the filesystem (rootfs)** and skips the guest memory snapshot. Resuming such a snapshot **cold-boots (reboots)** a fresh VM from the rootfs instead of restoring memory.

- `Sandbox.Pause(WithFilesystemSnapshot())` — pauses without a memory snapshot; uploads only the rootfs + metadata (no memfile/snapfile).
- `Factory.RebootSandbox` — the resume primitive that cold-boots from a snapshot's rootfs, marked routable only once the guest is ready.
- Cold boot now writes the envd MMDS access-token hash, so rebooted secure sandboxes authenticate the same way memory-resumed ones do.
- `cmd/resume-build` gains `-fs-only` and `-reboot` to exercise the full cycle locally.

Normal memory snapshots are unchanged — this mode is strictly opt-in.

### Why
Memory snapshots are the expensive part of a pause (large artifacts, slow upload, more storage). Many workloads don't need warm RAM restored on resume — they only need their disk state. A filesystem-only snapshot is cheaper and smaller to take and store, and also serves as an escape hatch when memory artifacts are slow, oversized, or missing. The trade-off is explicit: a reboot loses RAM, running processes, connections, and unsynced writes; only the synced filesystem survives.

### Scope / not in this PR
Orchestrator-only. No public API/proto/SDK wiring yet — the reboot path is reachable via `resume-build` but not yet over gRPC. API-side resume selection and auto-resume/`connect` gating (so a disk-only snapshot is never *silently* rebooted) are follow-ups. Guest cold-boot speedups live on a separate branch.

### Testing
Verified to build/lint clean per commit. End-to-end pause→reboot validation (synced files persist, uptime resets, correct on-storage layout) is done locally via `resume-build`; secure-sandbox envd auth after reboot still needs validation against a deployment.

### DIFF
--- a/packages/orchestrator/pkg/sandbox/reboot.go
+++ b/packages/orchestrator/pkg/sandbox/reboot.go
@@ -1,0 +1,119 @@
+//go:build linux
+
+package sandbox
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/google/uuid"
+
+	"github.com/e2b-dev/infra/packages/orchestrator/pkg/sandbox/block"
+	"github.com/e2b-dev/infra/packages/orchestrator/pkg/sandbox/fc"
+	"github.com/e2b-dev/infra/packages/orchestrator/pkg/sandbox/template"
+	"github.com/e2b-dev/infra/packages/orchestrator/pkg/template/constants"
+	"github.com/e2b-dev/infra/packages/orchestrator/pkg/units"
+	"github.com/e2b-dev/infra/packages/shared/pkg/fc/models"
+	"github.com/e2b-dev/infra/packages/shared/pkg/grpc/orchestrator"
+	"github.com/e2b-dev/infra/packages/shared/pkg/storage/header"
+	"github.com/e2b-dev/infra/packages/shared/pkg/utils"
+)
+
+const (
+	// minEnvdVersionForKVMClock is the minimum envd version that supports kvm-clock.
+	minEnvdVersionForKVMClock = "0.2.11"
+
+	// rebootEnvdTimeout bounds the systemd boot + envd start; a cold boot needs a
+	// longer window than a memory resume (matches the template build's wait).
+	rebootEnvdTimeout = 60 * time.Second
+)
+
+// RebootSandbox cold-boots a fresh Firecracker VM from the template's rootfs,
+// without restoring guest memory. Used to resume filesystem-only snapshots:
+// guest RAM, processes, and sockets are lost; only the filesystem survives.
+// The sandbox is marked running only after envd is ready, matching
+// ResumeSandbox's routing guarantees; endAt is the caller's absolute end time.
+// IMPORTANT: You must Close() the sandbox after you are done with it.
+func (f *Factory) RebootSandbox(
+	ctx context.Context,
+	t template.Template,
+	config *Config,
+	runtime RuntimeMetadata,
+	endAt time.Time,
+	apiConfigToStore *orchestrator.SandboxConfig,
+) (*Sandbox, error) {
+	ctx, span := tracer.Start(ctx, "reboot sandbox")
+	defer span.End()
+
+	buildID, err := uuid.Parse(t.Files().BuildID)
+	if err != nil {
+		return nil, fmt.Errorf("parse build ID: %w", err)
+	}
+
+	// The masked empty memfile is used only for sizing NoopMemory — guest RAM
+	// is FC's own fresh anonymous memory.
+	pageSize := int64(header.PageSize)
+	if config.HugePages {
+		pageSize = int64(header.HugepageSize)
+	}
+	memfile, err := block.NewEmpty(units.MBToBytes(config.RamMB), pageSize, buildID)
+	if err != nil {
+		return nil, fmt.Errorf("create empty memfile: %w", err)
+	}
+
+	maskedTemplate := template.NewMaskTemplate(t, template.WithMemfile(memfile))
+
+	kvmClock, err := utils.IsGTEVersion(config.Envd.Version, minEnvdVersionForKVMClock)
+	if err != nil {
+		return nil, fmt.Errorf("compare envd version: %w", err)
+	}
+
+	// Sync IO engine so no async writes are in flight if the sandbox is paused again.
+	ioEngine := models.DriveIoEngineSync
+
+	timeout := time.Until(endAt)
+	if timeout <= 0 {
+		return nil, fmt.Errorf("sandbox end time %s is in the past", endAt)
+	}
+
+	sbx, err := f.CreateSandbox(
+		ctx,
+		config,
+		runtime,
+		maskedTemplate,
+		timeout,
+		// Empty rootfs cache path selects the NBD provider, same as a memory
+		// resume, so guest TRIM keeps working and a later pause exports the
+		// overlay diff exactly like a normal resume.
+		"",
+		fc.ProcessOptions{
+			InitScriptPath: constants.SystemdInitPath,
+			KvmClock:       kvmClock,
+			IoEngine:       &ioEngine,
+		},
+		apiConfigToStore,
+		nil,
+		WithDeferredMarkRunning(),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("create sandbox from rootfs: %w", err)
+	}
+
+	// CreateSandbox anchors the lifetime to now; honor the caller's absolute end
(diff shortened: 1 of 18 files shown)


packages/orchestrator/pkg/scheduling/metadata.go
```diff
@@ -20,28 +20,35 @@ const chainLimit = 128
...
-		RootfsDroppedBuilds:  uint32(rootDropped),
+	md := &orchestrator.SchedulingMetadata{
+		RootfsBaseBuildId:   rootfsBase.String(),
+		BuildId:             buildID.String(),
+		RootfsBuildIds:      rootIDs,
+		RootfsBuildBytes:    rootBytes,
+		RootfsDroppedBuilds: uint32(rootDropped),
+	}
+
+	if memfileHeader != nil && memfileHeader.Metadata != nil {
```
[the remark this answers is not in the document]
bchalios: Fixed! `storageTemplate.SchedulingMetadata` now reports rootfs-only scheduling metadata for reloaded fs-only templates instead of dropping it.

packages/orchestrator/pkg/sandbox/reclaim.go
```diff
@@ -108,6 +108,43 @@ func (s *Sandbox) bestEffortReclaim(ctx context.Context) {
...
 }
 
+// guestSync runs sync in the guest via envd so ext4 flushes dirty pages to the
+// virtio disk. Mandatory before a filesystem-only pause: without a memory
+// snapshot the guest page cache is lost, so callers must fail the pause on
+// error instead of persisting a rootfs missing acknowledged writes. Unlike
+// bestEffortReclaim's sync step (LD-flag gated, best-effort), this always runs
+// and always reports failure.
+func (s *Sandbox) guestSync(ctx context.Context) error {
+	const syncTimeout = 5 * time.Second
```
wj-e2b: 5 is pretty short if there's a lot of dirty pages and there's a lot of IO on the box. Maybe this is fine in practice, but we should at least math this out for worst case that every page is dirty and is not flushed for a our maximum disk size.

packages/orchestrator/pkg/sandbox/sandbox.go
wj-e2b, line 1188: This is locked behind a feature flag (within the function call), but without it there's a race between when `sync` is called and when the actual snapshot happens.

packages/orchestrator/pkg/sandbox/reboot.go
```diff
@@ -0,0 +1,128 @@
...
+	pageSize := int64(header.PageSize)
+	if config.HugePages {
+		pageSize = int64(header.HugepageSize)
+	}
+	memfile, err := block.NewEmpty(units.MBToBytes(config.RamMB), pageSize, buildID)
+	if err != nil {
+		return nil, fmt.Errorf("create empty memfile: %w", err)
+	}
+
+	maskedTemplate := template.NewMaskTemplate(t, template.WithMemfile(memfile))
```
wj-e2b, line 66: Templates created before this PR will have inconsistent disks. This is fine normally because we load the dirty pages from memory, but this might create unexpected situation for the user where disk does not have the data they expect

packages/orchestrator/pkg/sandbox/sandbox.go
bchalios: There is a race even with this enabled :) The race window is between the moment the sync finishes and the moment we pause vCPUs. I have a follow up PR, which uses fs freeze semantics (FIFREEZE ioctl), which closes the gap. I kept it separate to make reviewing here easier, but I can include it here if you'd prefer.

packages/orchestrator/pkg/sandbox/reclaim.go
bchalios: That's a good catch. Just a note, the size of data we need to write is bound by the maximum RAM, not disk. 

I've tried to model the theoretical maximum value using a (very conservative) disk throughput value and peeking into the guest RAM value. I've also added a LaunchDarkly feature flag override.

packages/orchestrator/pkg/sandbox/reboot.go
bchalios: That is true, but IMO this can be handled at a higher level. There's a follow up PR which wires the gRPC calls. We can make sure at this level, that we only call RebootSandbox for filesystem-only snapshots (memory file is missing)

packages/orchestrator/pkg/sandbox/reboot.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/e2b-dev/infra/packages/orchestrator/pkg/units"
+	"github.com/e2b-dev/infra/packages/shared/pkg/fc/models"
+	"github.com/e2b-dev/infra/packages/shared/pkg/grpc/orchestrator"
+	"github.com/e2b-dev/infra/packages/shared/pkg/storage/header"
+	"github.com/e2b-dev/infra/packages/shared/pkg/utils"
+)
+
+const (
+	// minEnvdVersionForKVMClock is the minimum envd version that supports kvm-clock.
+	minEnvdVersionForKVMClock = "0.2.11"
```
kalyazin, line 26: Nit. It appears that this is duplicated with create_sandbox.go. Not that it's likely to be updated, but could be extracted into a single const.

packages/orchestrator/pkg/sandbox/reclaim.go
```diff
@@ -108,6 +126,76 @@ func (s *Sandbox) bestEffortReclaim(ctx context.Context) {
...
+// guestSync runs sync in the guest via envd so ext4 flushes dirty pages to the
+// virtio disk. Mandatory before a filesystem-only pause: without a memory
+// snapshot the guest page cache is lost, so callers must fail the pause on
+// error instead of persisting a rootfs missing acknowledged writes. Unlike
+// bestEffortReclaim's sync step (LD-flag gated, best-effort), this always runs
+// and always reports failure.
+func (s *Sandbox) guestSync(ctx context.Context) error {
+	syncTimeout := s.guestSyncTimeout(ctx)
+
+	ctx, span := tracer.Start(ctx, "envd-guest-sync")
```
kalyazin, line 183: would we like a metric here as well to tell how long it usually takes?

kalyazin: In general, do we have enough data to separate the regular pause from fs-only in our metrics?

packages/orchestrator/pkg/sandbox/template/storage_template.go
```diff
@@ -276,18 +276,28 @@ func (t *storageTemplate) Files() storage.CachePaths {
 // rather than the header holders, which stay unset for templates loaded from
 // storage (the headers are resolved internally by NewStorage during Fetch).
 func (t *storageTemplate) SchedulingMetadata(ctx context.Context) *orchestrator.SchedulingMetadata {
-	memfile, memfileErr := t.memfile.WaitWithContext(ctx)
+	// The rootfs is always present; its header carries the build ID and is the
```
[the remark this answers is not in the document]
bchalios: Traced this,  it doesn't reproduce. An fs-only build never has a memfile diff. The only memory artifact tagged with its build ID is the masked empty memfile from RebootSandbox, and block.NewEmpty deliberately sets its block sources to uuid.Nil (not the build ID) so it can't be collapsed/referenced as a real ancestor. A subsequent memory pause inherits those uuid.Nil sources, so the fs-only build ID never lands in the child's memfile Mapping.Builds(), and appendAncestorBuilds (which skips uuid.Nil) never calls Uploads.Wait(fsOnlyBuild, Memfile). The fs-only build is waited on for the rootfs, which it has. @ValentaTomas could you verify my logic here please?

packages/orchestrator/pkg/sandbox/reclaim.go
bchalios: Added a metric for guest sync latency and span attribute to separate between full and filesystem-only snapshots.

packages/orchestrator/pkg/sandbox/sandbox.go
wj-e2b: Yeah this is fine then

wj-e2b (approved):

packages/orchestrator/pkg/sandbox/template/storage_template.go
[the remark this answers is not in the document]
ValentaTomas: I think so.

PR #2996, opened 2026-06-12
fix(orchestrator): harden Firecracker process shutdown

wj-e2b:
Use stdlib pidfd-backed signaling (os.Process is pidfd-backed since Go 1.23) so FC processes cannot be signaled after PID reuse, kill the sandbox cgroup on stop and wait for it to empty via poll on cgroup.events instead of polling cgroup.procs, and keep memory shutdown running even when the exit wait context is canceled so UFFD can exit.


dobrac (approved):
LGTM! Lets resolve first the cursor comment please

PR #2997, opened 2026-06-12
feat(orch): collapse envd's heap into 2 MiB hugepages before pause to cut cold-resume faults

kalyazin:
## Why

A cold (GCS-backed) sandbox resume demand-faults guest RAM through userfaultfd at **2 MiB hugepage granularity**, and each distinct 2 MiB frame is a separate serial fetch from remote storage (~tens of ms each). On the resume critical path the user cgroup is still frozen (`freeze-user-cgroup`), so the only pages faulted before envd answers `/init` are **envd's own**. envd is a long-lived Go process whose heap fragments over time across many distinct guest-physical 2 MiB frames at low fill — so restoring envd touches far more frames than its live byte size implies, and that dominates time-to-envd on a cold resume.

`MADV_COLLAPSE` migrates those scattered 4 KiB live pages into contiguous 2 MiB hugepages, collocating them into *fewer* distinct frames — so the post-pause snapshot faults fewer frames on the next resume. This PR adds an envd endpoint to do that and wires the orchestrator to call it just before pause, behind a flag.

## What

**envd: `POST /collapse`** (`packages/envd/`, version bumped to `0.6.5`)
- Walks `/proc/self/maps`, and for each anonymous read-write region marks it `MADV_HUGEPAGE` then issues `MADV_COLLAPSE` **per 2 MiB chunk** (skipping empty chunks).
- Returns `200` + JSON `CollapseResult{regions,chunks,collapsed,skipped,elapsedMs}` and logs the same stats in-guest.
- Honors the request context: on the orchestrator's timeout or a client disconnect it stops issuing further `MADV_COLLAPSE` calls and returns the progress made so far, mirroring how `/freeze` honors its request context.
- Notes:
  - **Chunk-wise, not whole-VMA:** a single `MADV_COLLAPSE` over a whole region `EINVAL`s on the first empty chunk; iterating 2 MiB windows and skipping empties is what makes it work on a sparse heap.
  - **`MADV_HUGEPAGE` first** sets `VM_HUGEPAGE` so the collapse is eligible regardless of the system THP mode (no global `transparent_hugepage=always` needed).
  - Best-effort: per-chunk failures are counted, never fatal.

**orchestrator: flag-gated pre-pause collapse** (`packages/orchestrator/pkg/sandbox/`)
- New LD flag `collapse-envd-heap` (**default off**). When set, `bestEffortReclaim` calls envd's `/collapse` just before pause, alongside the existing freeze step.
- New LD flag `collapse-envd-heap-timeout-ms` (**default 10000**) bounds the `/collapse` call, so the budget is tunable per rollout without a redeploy; `≤ 0` falls back to the 10 s code default. Collapse is best-effort, so a cut-short run still helps.
- Gated on envd version **≥ 0.6.5** (`MinEnvdVersionForHeapCollapse`): `0.6.5` is the version that introduces `/collapse`, and `0.6.4` already exists in the fleet *without* it, so gating at `0.6.5` prevents POSTing `/collapse` at a 0.6.4 envd that would 404. Best-effort so a failure never blocks pause.
- `callEnvdCollapse` decodes the `CollapseResult` (envd's client stub regenerated for the new 200+body shape).

**Observability** (`packages/shared/pkg/telemetry`, orchestrator)
- Histogram `orchestrator.sandbox.envd.collapse.duration` (round-trip ms, `success` attribute) — the pause-path cost.
- Counter `orchestrator.sandbox.envd.collapse.chunks` split by `result=collapsed|skipped` — chunk attempts (total) vs successes (collapsed).
- Dedicated `envd-collapse` span carrying `collapse.{regions,chunks,collapsed,skipped,duration_ms,success}`, plus a success log line (`"envd heap collapsed"`) so efficacy is visible in Loki even when per-process metric counters reset on an orchestrator restart.

## Validation

- **Unit tests** (`internal/services/memory`): `/proc/self/maps` anon-rw parser, `collapseRange` against a sparsely-populated mmap'd region (one touched page per 2 MiB window), and a `CollapseSelf` smoke test. `go build`/`vet` clean; `golangci-lint` 0 issues across the changed packages.
- **resume-build A/B** (GCS-cold, envd heap physically scattered): cold-resume working set **CTRL 1586 → COLL 465 distinct 2 MiB frames (−71 %)**, latency **32.2 s → 8.2 s (−74 %)**; collapse consolidated ~250 of ~290 chunks.
- **Full managed-orchestrator A/B** (e2b API → orchestrator → Firecracker, `use-nfs-for-snapshots=false` to force GCS-cold): **CTRL 1005 frames / 2010 MB / 22.6 s → COLL 392 frames / 784 MB / 9.5 s** — **−61 % faults, −61 % bytes, −58 % latency**, with **385 chunks consolidated in 428 ms** at pause. Metrics + span confirmed in Mimir/Tempo end-to-end (`service_version 0.2.0-…`).


packages/orchestrator/pkg/sandbox/reclaim.go
```diff
@@ -122,6 +139,73 @@ func (s *Sandbox) envdSupportsCgroupFreeze(ctx context.Context) bool {
...
+	)
+
+	logger.L().Info(ctx, "envd heap collapsed",
+		logger.WithSandboxID(s.Runtime.SandboxID),
+		zap.Int("regions", stats.Regions),
+		zap.Int("chunks", stats.Chunks),
+		zap.Int("collapsed", stats.Collapsed),
+		zap.Int("skipped", stats.Skipped),
+		zap.Int64("duration_ms", elapsedMs),
+	)
```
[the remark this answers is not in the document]
kalyazin: This is a false positive here. `reclaim.go` consumes the **orchestrator's** generated client (`packages/orchestrator/pkg/sandbox/envd/envd.gen.go`), whose `cfg.yaml` sets `prefer-skip-optional-pointer: true` — so `CollapseResult` fields are plain values (`Chunks int`, `Collapsed int`, `ElapsedMs int64`, …), not pointers. `int64(stats.Collapsed)` compiles, and `utils.DerefOrDefault(stats.Collapsed, 0)` would actually fail to compile (can't deref a non-pointer). Confirmed by the passing orchestrator build/lint/ARM64 checks. (The pointer-typed `CollapseResult` is the *envd-side* generated type, which is not used on this path.)

packages/envd/internal/api/collapse.go
```diff
@@ -0,0 +1,57 @@
...
+
+	start := time.Now()
+	stats, err := memory.CollapseSelf()
+	elapsedMs := time.Since(start).Milliseconds()
+	if err != nil {
+		logger.Error().Err(err).Int64("elapsed_ms", elapsedMs).Msg("collapse envd heap")
+		jsonError(w, http.StatusInternalServerError, err)
+
+		return
+	}
```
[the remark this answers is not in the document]
kalyazin: Good catch — fixed. `PostCollapse` now passes `r.Context()` into `memory.CollapseSelf(ctx)`, which checks `ctx.Err()` before each region and before each 2 MiB chunk in `collapseRange`. On the orchestrator’s 10s timeout or a client disconnect it stops issuing further `MADV_COLLAPSE` calls and returns the progress made so far, rather than running on past pause — mirroring how `/freeze` honors the request context.

packages/envd/internal/services/memory/collapse_linux.go
```diff
@@ -0,0 +1,145 @@
...
+			return s, err
+		}
+		s.Regions++
+		rs := collapseRange(ctx, r.start, r.end)
+		s.Chunks += rs.Chunks
+		s.Collapsed += rs.Collapsed
+		s.Skipped += rs.Skipped
+	}
+
+	return s, nil
```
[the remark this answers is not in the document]
kalyazin: Fixed — made cancellation handling uniform. `CollapseSelf` now `break`s out of the region loop on `ctx.Err()` (instead of returning the error), so a timeout/disconnect on any region yields the same outcome: the progress made so far, returned with a nil error and HTTP 200. Collapse is best-effort, so a cut-short run is a partial success rather than an error on some regions and success on others. `collapseRange` already returns its partial stats with no error on cancel, so the whole path is now consistent.

packages/envd/internal/services/memory/collapse_linux.go
```diff
@@ -0,0 +1,129 @@
...
+		start, err := strconv.ParseUint(fields[0][:dash], 16, 64)
+		if err != nil {
+			continue
+		}
+		end, err := strconv.ParseUint(fields[0][dash+1:], 16, 64)
+		if err != nil || end <= start {
+			continue
+		}
+
+		regions = append(regions, region{start: uintptr(start), end: uintptr(end)})
```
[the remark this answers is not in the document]
kalyazin: Addressed: the parsed addresses are now bounded by the platform pointer width before the `uint64`->`uintptr` conversion — `maxUintptr := uint64(math.MaxUint64)`, set to `math.MaxUint32` when `strconv.IntSize == 32`, and the region is skipped if `start`/`end` exceed it. Correct on both widths (envd only runs on 64-bit, where it never trips) and uses the platform-width-keyed bound CodeQL recognizes for `uintptr` narrowing.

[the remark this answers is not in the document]
kalyazin: Addressed: the parsed addresses are now bounded by the platform pointer width before the `uint64`->`uintptr` conversion — `maxUintptr := uint64(math.MaxUint64)`, set to `math.MaxUint32` when `strconv.IntSize == 32`, and the region is skipped if `start`/`end` exceed it. Correct on both widths (envd only runs on 64-bit, where it never trips) and uses the platform-width-keyed bound CodeQL recognizes for `uintptr` narrowing.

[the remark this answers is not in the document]
kalyazin: Resolved by removing the flagged conversion entirely: the maps parsing now uses `github.com/prometheus/procfs` (`procfs.Self().ProcMaps()`), which returns addresses already typed as `uintptr` (`ProcMap.StartAddr`/`EndAddr`). Our code no longer does `strconv.ParseUint(...)` -> `uintptr`, so there is nothing to bound-check — the prior `^uintptr(0)` and `strconv.IntSize` guards did not satisfy the query (CodeQL evaluates the 64-bit branch as a dead compare), and the only bound it would accept (MaxUint32) would wrongly drop real 64-bit heap addresses. Using procfs also drops the hand-rolled `/proc` parser. The library-internal parse is not reported as a PR-introduced alert.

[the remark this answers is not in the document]
kalyazin: Resolved by removing the flagged conversion entirely: the maps parsing now uses `github.com/prometheus/procfs` (`procfs.Self().ProcMaps()`), which returns addresses already typed as `uintptr` (`ProcMap.StartAddr`/`EndAddr`). Our code no longer does `strconv.ParseUint(...)` -> `uintptr`, so there is nothing to bound-check — the prior `^uintptr(0)` and `strconv.IntSize` guards did not satisfy the query (CodeQL evaluates the 64-bit branch as a dead compare), and the only bound it would accept (MaxUint32) would wrongly drop real 64-bit heap addresses. Using procfs also drops the hand-rolled `/proc` parser. The library-internal parse is not reported as a PR-introduced alert.

packages/envd/internal/services/memory/collapse_linux.go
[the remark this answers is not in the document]
kalyazin: Update on the final shape (it changed since my earlier reply): `CollapseSelf` now ends with `return s, ctx.Err()` — `nil` on a complete run, and the cancellation error uniformly when the caller gave up, regardless of which region was being processed (the loop `break`s on `ctx.Err()` and the single tail return reports it). So the inconsistency is gone: a timeout no longer looks like success on the last region and failure on earlier ones. (Returning `ctx.Err()` rather than a partial-success `nil` also keeps the `nilerr` linter happy, since the function does surface the non-nil error it detected.)

packages/envd/internal/services/memory/collapse_linux.go
```diff
@@ -0,0 +1,117 @@
...
+		if m.Perms == nil || !m.Perms.Read || !m.Perms.Write {
+			continue
+		}
+		if m.Pathname != "" {
+			continue
+		}
+		if m.EndAddr <= m.StartAddr {
+			continue
+		}
+		regions = append(regions, region{start: m.StartAddr, end: m.EndAddr})
```
bchalios, line 75: Is that possible somehow? :O

packages/envd/internal/services/memory/collapse_linux.go
```diff
@@ -0,0 +1,117 @@
...
+// /[vdso]/…), neither of which is an anon heap arena.
+func anonRWRegions(maps []*procfs.ProcMap) []region {
+	var regions []region
+	for _, m := range maps {
+		if m.Perms == nil || !m.Perms.Read || !m.Perms.Write {
+			continue
+		}
+		if m.Pathname != "" {
+			continue
+		}
```
bchalios, line 74: would it make sense to look only for `[heap]` here, to (maybe) speed up the process?

packages/envd/internal/services/memory/collapse_linux.go
```diff
@@ -0,0 +1,117 @@
...
+	// Best-effort eligibility hint; ignore errors.
+	_ = madvise(start, end-start, unix.MADV_HUGEPAGE)
+
+	first := (start + hugePageSize - 1) &^ uintptr(hugePageSize-1) // round up to 2 MiB
+	for c := first; c+hugePageSize <= end; c += hugePageSize {
+		if ctx.Err() != nil {
+			return s
+		}
+		s.Chunks++
+		if err := madvise(c, hugePageSize, unix.MADV_COLLAPSE); err == nil {
```
bchalios, line 98: empirically, if this fails once is there a chance that a subsequent call will succeed?

packages/orchestrator/cmd/resume-build/main.go
```diff
@@ -75,6 +75,7 @@ func main() {
 
 	fphTimeoutMs := flag.Int("fph-timeout-ms", 0, "override free-page-hinting-config pause timeout LD flag (0 = use LD default)")
 	reclaim := flag.Bool("reclaim", false, "enable pre-pause reclaim chain (fstrim 500ms, sync 500ms, drop_caches 200ms, compact 1s)")
+	collapse := flag.Bool("collapse", false, "collapse envd's heap before pause (overrides the collapse-envd-heap flag)")
```
bchalios: nit: "collapse", maybe is a bit unintuitive

packages/orchestrator/pkg/sandbox/reclaim.go
```diff
@@ -23,6 +26,12 @@ const reclaimOuterSlack = 500 * time.Millisecond
 // must stay independent of the reclaim shell deadline.
 const freezeTimeout = 2 * time.Second
 
+// collapseTimeout bounds the native POST /collapse call. Collapsing migrates
+// envd's scattered heap pages into hugepages, which is heavier than the freeze
+// sysfs write, so it gets a larger, independent budget. Best-effort: if it is
+// cut short, the partially-collapsed heap is still an improvement.
+const collapseTimeout = 10 * time.Second
```
bchalios: would it make sense to make this configurable (via LD)?

packages/orchestrator/pkg/sandbox/envd.go
```diff
@@ -108,28 +108,72 @@ func (s *Sandbox) callEnvdUnfreeze(ctx context.Context, timeout time.Duration) e
...
+	if err := json.NewDecoder(resp.Body).Decode(&result); err != nil {
+		return envd.CollapseResult{}, fmt.Errorf("decode collapse result: %w", err)
+	}
+
+	return result, nil
+}
+
+// postEnvd issues an authenticated POST to envd's /<path> endpoint with a tight,
+// dedicated deadline and expects 204 No Content.
+func (s *Sandbox) postEnvd(ctx context.Context, timeout time.Duration, path string) error {
```
bchalios, line 145: Did you mean to use this within `callEnvdCollapse`? Otherwise, I don't get why did you re-write it(?)

bchalios (commented):
A few comments/questions. Also, I think you need to bump here envd version to 0.6.5.

packages/orchestrator/cmd/resume-build/main.go
kalyazin: not sure I get it, what are you suggesting?

packages/envd/internal/services/memory/collapse_linux.go
kalyazin: apparently , go doesn't use brk for heap so it doesn't mark those VMAs as `[heap]`.

packages/orchestrator/cmd/resume-build/main.go
kalyazin: ok, I got what you mean, I'll go with `-collapse-envd-heap`

packages/envd/internal/services/memory/collapse_linux.go
kalyazin: I didn't try retrying on EAGAIN. It looks like there are multiple reasons why it can return it and they are not equally likely to succeed on retry. What is certain is it's going to add more latency on the pause path. I'm thinking of going with the simple implementation and see whether a retry is worth it once we get some data. WDYT?

packages/envd/internal/services/memory/collapse_linux.go
kalyazin: probably not, removed :)

packages/orchestrator/pkg/sandbox/reclaim.go
kalyazin: added, thanks

packages/orchestrator/pkg/sandbox/envd.go
kalyazin: updated

kalyazin:
> A few comments/questions. Also, I think you need to bump here envd version to 0.6.5.

Thanks, bumped it.

packages/envd/internal/services/memory/collapse_linux.go
bchalios: I didn't express what I wanted clearly. What I meant was, if `madvise()` fails for the first 2MiB of a region, what's the possibility that the next 2MiB will succeed? Are these independent variables?

packages/orchestrator/pkg/sandbox/reclaim.go
bchalios: do we still need the local `collapseTimeout`? IIUC, if LD is not available the system will be using the locally defined default from here:

```Go
	// CollapseEnvdHeapTimeoutMsFlag bounds the pre-pause POST /collapse call, in
	// milliseconds. Collapse is best-effort: a cut-short run still helps, so this
	// can be tuned per rollout without redeploying. Zero/negative falls back to
	// the code default (collapseTimeout).
	CollapseEnvdHeapTimeoutMsFlag = NewIntFlag("collapse-envd-heap-timeout-ms", 10000)
```

packages/envd/internal/services/memory/collapse_linux.go
kalyazin: I see what you mean. I don't think we have sufficient data to tell. Even if in our testing we see some pattern, it may not be ubiquitous. For example, I wouldn't like to give up the entire region if its first block couldn't collapse for some reason. Another point is, as I understand (although not 100% sure), the main time is spent in actual collapsing (HP allocation and data move) vs the search.

packages/orchestrator/pkg/sandbox/reclaim.go
kalyazin: make sense, updated, thanks!

bchalios (approved):

packages/envd/internal/api/collapse.go
```diff
@@ -0,0 +1,57 @@
...
+	logger := a.logger.With().Str(string(logs.OperationIDKey), logs.AssignOperationID()).Logger()
+
+	start := time.Now()
+	stats, err := memory.CollapseSelf(r.Context())
+	elapsedMs := time.Since(start).Milliseconds()
+	if err != nil {
+		logger.Error().Err(err).Int64("elapsed_ms", elapsedMs).Msg("collapse envd heap")
+		jsonError(w, http.StatusInternalServerError, err)
+
+		return
```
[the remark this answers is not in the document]
kalyazin: Decided to leave this as-is — on reflection it's working-as-intended-enough, and "fixing" it would only buy a bit of log-noise reduction on a path that's essentially never hit.

The only caller is the orchestrator, which bounds `/collapse` with its own `context.WithTimeout`. When that deadline fires, the client aborts and `Do()` returns *before* reading the response — and that same abort (connection close) is what cancels envd's `r.Context()`. So envd only enters the cancellation branch *after* the orchestrator has already stopped listening; it can never read a `200`+partial-stats response on timeout. Returning partial stats instead of `500` is therefore invisible to the orchestrator.

The genuinely useful behavior — stopping `MADV_COLLAPSE` once the caller gives up — already exists via the `ctx` check. All that's left is whether a cancelled best-effort run logs as a `500`/error vs info, and since collapse completes in ~hundreds of ms against a ≥10s budget, that branch is virtually never exercised. Not worth the change (plus an envd version bump) for that.

The related, more impactful issue from this review — the fixed 10s `http.Client.Timeout` silently capping the configurable `collapse-envd-heap-timeout-ms` — is being fixed in #3016.

packages/orchestrator/pkg/sandbox/reclaim.go
```diff
@@ -122,6 +133,82 @@ func (s *Sandbox) envdSupportsCgroupFreeze(ctx context.Context) bool {
...
+	// whenever LD is unavailable or the flag is unset — so there is no separate
+	// local default to keep in sync.
+	timeout := time.Duration(s.featureFlags.IntFlag(ctx, featureflags.CollapseEnvdHeapTimeoutMsFlag,
+		featureflags.SandboxContext(s.Runtime.SandboxID),
+		featureflags.TeamContext(s.Runtime.TeamID),
+		featureflags.TemplateContext(s.Runtime.TemplateID),
+	)) * time.Millisecond
+
+	start := time.Now()
+	stats, err := s.callEnvdCollapse(ctx, timeout)
```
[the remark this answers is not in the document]
kalyazin: Good catch — confirmed. `sandboxHttpClient` has a fixed `Timeout: 10 * time.Second`, so the configurable `collapse-envd-heap-timeout-ms` could only ever *lower* the effective deadline, never raise it past 10s; above that the client aborts first and the collapse is reported failed.

Fixed in #3016: `/collapse` now goes through a dedicated client with no client-level timeout, bounded only by the per-call request context, so the flag budget is honored in full. Scoped to `/collapse` only — `/freeze` and `/unfreeze` keep the shared 10s-capped client, since their fixed 2s deadline is always well under it and the cap is a useful backstop there.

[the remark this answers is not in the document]
kalyazin: Thinking more about it, it isn't a problem. We don't want collapse to take longer than 10 seconds anyway.

PR #3006, opened 2026-06-13
perf(orchestrator): reduce mount namespace copy overhead during sandbox startup

emailcannotbeblank:
## Summary

This PR significantly changes and optimizes the sandbox creation path, greatly improving sandbox startup speed under high concurrency.

## Problem

When creating sandboxes with 100 concurrent requests, sandbox startup is slow. A key step suffers from lock contention, causing the per-sandbox cost to exceed 400ms.

More specifically, `NewProcess` in `packages/orchestrator/pkg/sandbox/fc/process.go` is a critical function in the sandbox creation path.

Previously, this function prepared a bash-based startup flow that included commands similar to:

```bash
unshare -m ...
ip netns exec ns-xxx firecracker ...
```

Both commands copy the mount tree and eventually reach `copy_mnt_ns` in the Linux kernel file `fs/namespace.c`. When each thread copies the mount tree, it calls `namespace_lock` and tries to acquire a global lock. The kernel code is:

```c
static inline void namespace_lock(void)
{
    down_write(&namespace_sem);
}
```

With 100 concurrent threads contending for the same global lock, sandbox startup latency becomes very high.

In addition:

1. Copying a mount tree requires traversing the mount tree, which is expensive. This can be measured with:

   ```bash
   time unshare -m nproc
   ```

   This command can be used to test how long it takes to copy the mount tree.

2. The previous flow performed two mount tree copies (`unshare -m` and `ip netns exec`) and one mount namespace release (the mount namespace created by `unshare -m` being released). All of these operations acquire the global namespace lock.

For these reasons, creating the Firecracker process has very large latency under high concurrency.

## Approach

1. Create a minimal mount tree template.

   Before creating sandboxes, the orchestrator prepares a minimal mount tree and unmounts unnecessary nodes. Later, when creating each sandbox, the mount tree is copied from this minimal template instead of being copied from the full host mount tree. This greatly reduces the time spent copying the mount tree.

2. Optimize the Firecracker process startup flow.

   The previous `unshare -m` and `ip netns exec ns-xxx` flow is replaced with explicit `setns` system calls. This reduces the startup path from two mount tree copies and one mount namespace release to only one mount tree copy.

## Result

This greatly improves the time required to create the Firecracker process.

In `packages/orchestrator/pkg/sandbox/fc/process.go`, the `configure` function is more than 10x faster in our tests, decreasing from over 400ms to under 40ms.

## Discussion

1. This version not only significantly improves startup speed, but also improves sandbox security.

   Previously, the host-side Firecracker process inherited a copy of the host mount tree. If an attacker compromised the Firecracker process, they could observe more of the host filesystem structure. This optimization prunes the mount tree used by the Firecracker process and reduces the attack surface.

2. This optimization reduces the mount tree size and reduces the number of global lock acquisitions.

   Another possible optimization would be to reuse mount namespaces together with reusable network namespaces. Since different sandboxes already reuse network namespace slots, we could prepare a paired mount namespace for each network namespace. When creating a sandbox, after acquiring a network namespace slot, the sandbox could directly use the corresponding mount namespace.

   However, that approach would reduce isolation to some extent. If an attacker compromised the Firecracker process, they might be able to attack other users that later reuse the same network namespace and mount namespace pair.

This PR chooses a more conservative approach: each sandbox still gets its own mount namespace, but that namespace is copied from a minimal template and created with fewer global-lock-heavy operations.


ValentaTomas:
Good idea, but closing as stale for now—let's reopen if it becomes relevant again.

PR #3007, opened 2026-06-15
Fix NBD release retry loop ignoring context cancel

AdaAibaby:
Replace blocking time.Sleep with select + time.After in infinite retry loop of ReleaseDevice.
Now the loop correctly respects context cancellation, preventing process hang during graceful shutdown and timeout.


jakubno:
Worst case it will wait `500 * time.Millisecond`

PR #3011, opened 2026-06-15
Use by-default memfd and background copy

bchalios:
Make memfd for guest memory and background copies during pause/snapshot the default. The behaviour is still controlled by LaunchDarkly (when present) but the default now is to use these two features.

Also, flip -use-memfd to -disable-memfd flag in create-build and resume-build, so that local testing uses memfd by default and apply the same flip in CI testing environments.


jakubno (approved):

dobrac (approved):
make sure to properly specify the FC version dependency

bchalios:
I've also added a commit to change the default Firecracker version to v1.14 which supports using memfd.

jakubno (approved):

.github/actions/build-sandbox-template/action.yml
```diff
@@ -26,7 +26,7 @@ runs:
       env:
         TEMPLATE_ID: "2j6ly824owf4awgai1xo"
         KERNEL_VERSION: "vmlinux-6.1.158-c1a568c"
-        FIRECRACKER_VERSION: "v1.14.1_bd85e43"
+        FIRECRACKER_VERSION: "v1.14.1_431f1fc"
```
jakubno, line 29: Probably for different PR, but we could remove this from build template command and allow overriding with env variable (now make file requires this,  because it passes it as parameter)

PR #3019, opened 2026-06-16
Use FIFREEZE before getting filesystem snapshots

bchalios:
### What

Adds envd `POST /fsfreeze` and `POST /fsthaw` endpoints (FIFREEZE/FITHAW ioctls on the guest rootfs) and wires them into the filesystem-only pause path: when the sandbox's envd supports it, the rootfs is **frozen** before the snapshot instead of merely synced. Older envds keep the existing guest-`sync` behavior. Bumps envd to `0.6.6`.

### Why
A filesystem-only pause has no memory snapshot, so the guest page cache is lost on resume — the rootfs must be quiesced to a consistent on-disk state before the VM is paused. A guest `sync` is **point-in-time**: a write acknowledged after `sync` returns but before `process.Pause` lives only in the page cache and is lost on the reboot resume. `FIFREEZE` both
flushes dirty data *and* blocks further writes until thaw, closing that sync→pause race and guaranteeing the persisted rootfs is consistent.

### How
- **envd:** new `/fsfreeze` and `/fsthaw` handlers freeze/thaw the rootfs (`/`) via the FIFREEZE/FITHAW ioctls. Both are idempotent (already-frozen / not-frozen are no-ops) and serialized by a dedicated lock; a Linux ioctl impl with a non-Linux stub.
- **orchestrator:** a version gate (`MinEnvdVersionForFsFreeze = 0.6.6`) selects fsfreeze when supported. Since `FIFREEZE` already syncs as part of freezing, the redundant guest `sync` is skipped in that case. The thaw is registered on the pause's cleanup, so it fires only on the abort path (a failed pause can't leave the live VM frozen); on success the freeze
is held through `process.Pause` and discarded when the VM is stopped during rootfs export. A duration histogram (tagged `fsfreeze` vs sync, plus `success`/`timeout_ms`) measures both paths.

### DIFF
--- a/packages/orchestrator/pkg/sandbox/reclaim.go
+++ b/packages/orchestrator/pkg/sandbox/reclaim.go
@@ -232,6 +232,20 @@ func (s *Sandbox) envdSupportsCgroupFreeze(ctx context.Context) bool {
 	return ok
 }
 
+// envdSupportsFsFreeze reports whether the sandbox's envd exposes the native
+// /fsfreeze and /fsthaw endpoints. Bad version strings log and return false so
+// the filesystem-only pause falls back to a plain guest sync.
+func (s *Sandbox) envdSupportsFsFreeze(ctx context.Context) bool {
+	ok, err := utils.IsGTEVersion(s.Config.Envd.Version, utils.MinEnvdVersionForFsFreeze)
+	if err != nil {
+		logger.L().Warn(ctx, "fsfreeze version gate: bad envd version", logger.WithSandboxID(s.Runtime.SandboxID), zap.String("envd_version", s.Config.Envd.Version), zap.Error(err))
+
+		return false
+	}
+
+	return ok
+}
+
 // envdSupportsHeapCollapse reports whether the sandbox's envd exposes the native
 // /collapse endpoint. Bad version strings log and return false so we never call
 // an unsupported endpoint.
--- a/packages/orchestrator/pkg/sandbox/reclaim.go
+++ b/packages/orchestrator/pkg/sandbox/reclaim.go
@@ -11,6 +11,7 @@ import (
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
 	"go.opentelemetry.io/otel/metric"
+	"go.opentelemetry.io/otel/trace"
 	"go.uber.org/zap"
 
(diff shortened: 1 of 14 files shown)


packages/orchestrator/pkg/sandbox/reclaim.go
```diff
@@ -170,29 +171,62 @@ func (s *Sandbox) guestSyncTimeout(ctx context.Context) time.Duration {
...
+	if supportsFsFreeze {
+		// fsfreeze flushes the rootfs AND blocks further writes until thaw,
+		// closing the sync->pause race. FIFREEZE already syncs as part of
+		// freezing, so a separate guest sync would be redundant.
+		// If freezing aborted, thaw so we don't leave the live VM's
+		// filesystem frozen; on success the VM is stopped during rootfs
+		// export, so the frozen state is discarded with it and the thaw is a
+		// harmless no-op.
+		cleanup.Add(ctx, func(ctx context.Context) error {
+			s.bestEffortFsthaw(ctx)
```
kalyazin, line 207: were you able to see if this works/helps in practice?

bchalios: not really. I'd need to somehow force `/fsfreeze` to fail. Any ideas are welcome

kalyazin: It looks like the two common options here are seccomp or ptrace. I was thinking of an ad-hoc one-time test that would give us some confidence that this code path is functional without too much of a burden of setting it up.

bchalios: Hmmm, actually I'm wrong. The interesting case is not `/fsfreeze` failing. It's `/fsfreeze` succeeding and a subsequent step of `Pause` fails. Not sure we can seccomp/ptrace this :/ Maybe forcing some timer failing?

kalyazin: hm, it looks like it may only be triggered if FC refuses to pause. Don't know how likely it is.

kalyazin (approved):

packages/orchestrator/pkg/sandbox/reclaim.go
kalyazin: discussed offline, we can handle it in a follow-up.

PR #3021, opened 2026-06-16
feat(envd): split collapse stats into real migrations vs already-huge

kalyazin:
## Why

The pre-pause envd heap collapse records `MADV_COLLAPSE` successes as `collapsed`. But `MADV_COLLAPSE` returns success **both** when it actually migrates scattered base pages into a new hugepage **and** when the window was already a hugepage (a no-op). So the raw `collapsed` count overstates real work, and the collapse-duration/efficacy metrics are diluted by no-op re-collapses (e.g. re-pausing an already-collapsed, resumed sandbox).

## What

- Split the count using the process's **`AnonHugePages` delta** (`/proc/self/smaps_rollup`) across the collapse:
  - `collapsed` now counts only windows **actually migrated** this run.
  - new `already_huge` counts windows that were already huge.
  - invariant: `chunks = collapsed + already_huge + skipped`.
- Fall back to the pre-split meaning (all successes → `collapsed`) when `smaps_rollup` is unreadable, so a missing read never misreports real work as no-ops.
- Plumb `alreadyHuge` through the `/collapse` response (spec + regenerated envd server and orchestrator client) and emit it from the orchestrator as a third `result` bucket (`collapsed | already_huge | skipped`).
- Bump envd `0.6.5 → 0.6.6` (additive response field; version gate stays at 0.6.5).

## Validation

Verified on the dev cluster through the real API/orchestrator. First pause of a freshly-scattered envd heap: `collapsed=379, already_huge=2` (434 ms). Repeat pauses of the resumed (already-collapsed) sandbox: `collapsed≈20, already_huge≈360` (~30 ms). The split correctly attributes real work, and `collapsed` now correlates with collapse duration where the raw success count did not. Lint + unit tests pass.

### DIFF
--- a/packages/envd/internal/services/memory/collapse_linux.go
+++ b/packages/envd/internal/services/memory/collapse_linux.go
@@ -65,23 +65,33 @@ func CollapseSelf(ctx context.Context) (Stats, error) {
 	}
 	after, okAfter := anonHugePagesBytes()
 
-	// Attribute madvise successes to real migrations (AnonHugePages grew) vs
-	// already-huge no-ops. MADV_COLLAPSE is synchronous, so the delta reflects
-	// exactly the hugepages this run created. If smaps_rollup is unreadable, fall
-	// back to the pre-split meaning (all successes counted as Collapsed) rather
-	// than misreporting real work as no-ops.
-	if okBefore && okAfter && after >= before {
-		// Clamp: background THP activity could inflate the delta past what our
-		// madvise calls actually achieved.
-		migrated := min(int((after-before)/hugePageSize), successes)
-		s.Collapsed = migrated
-		s.AlreadyHuge = successes - migrated
-	} else {
-		s.Collapsed = successes
-	}
+	// Attribute madvise successes to real migrations vs already-huge no-ops
+	// using the AnonHugePages delta. MADV_COLLAPSE is synchronous, so the delta
+	// reflects exactly the hugepages this run created.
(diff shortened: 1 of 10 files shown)


packages/envd/internal/services/memory/collapse_linux.go
```diff
@@ -51,14 +60,61 @@ func CollapseSelf(ctx context.Context) (Stats, error) {
...
+	// than misreporting real work as no-ops.
+	if okBefore && okAfter && after >= before {
+		// Clamp: background THP activity could inflate the delta past what our
+		// madvise calls actually achieved.
+		migrated := min(int((after-before)/hugePageSize), successes)
+		s.Collapsed = migrated
+		s.AlreadyHuge = successes - migrated
+	} else {
+		s.Collapsed = successes
+	}
```
bchalios: This seems very racy. It's definitely an approximation of what's changed. Unless I'm missing something, other things might create huge-pages while we're doing this, so do you think it's really useful as a metric distinction?

kalyazin: Yes, that's true. I've been thinking about it this way:
- we currently set THP as madvise globally so envd should be the only user, especially at pause, and only when we explicitly request them
- even if it's diluted by other users, the effect should be small and having this metric at all outweighs its potential imprecision.

bchalios (approved):

PR #3032, opened 2026-06-17
feat(orchestrator): upload metadata.json last as completion marker

ValentaTomas:
## What
Upload `metadata.json` as the final object in the build upload (v3 and v4 paths), after all data files, headers, and the snapfile have been written.

## Why
Its presence then proves the build is fully uploaded. A consumer (e.g. the storage index) can treat a build missing `metadata.json` as incomplete — such as a crashed mid-upload — instead of risking action on a partial object set. GCS object writes are individually atomic, so "all others succeeded, then metadata.json" is a reliable commit marker.


ValentaTomas:
Not needed: per-artifact completeness is already derivable from the headers (present + IncompletePendingUpload=false, written after the data in V4/V5), and provenance comes from custom object metadata (#3033). The storage index can watch the headers instead, so forcing metadata.json to upload last adds no value (and this change had a ctx-cancellation bug in runV4). Closing.

PR #3033, opened 2026-06-17
feat(storage): stamp provenance custom metadata on uploaded objects (incl. headers)

ValentaTomas:
Attach immutable `team_id`, `template_id`, and `origin` (`pause`/`template_build`/`snapshot_template`) as custom object metadata on every uploaded object, including header objects (which previously carried no custom metadata at all). This lets the storage index attribute ownership/usage and classify root kinds from the object it already downloads, without the infra DB. Lineage stays in the header (`base_build_id` + reference edges); mutable, DB-owned state is deliberately excluded.


packages/orchestrator/pkg/server/sandboxes.go
```diff
@@ -890,7 +890,9 @@ func (s *Server) snapshotAndCacheSandbox(
 	}
 
 	objectMetadata := storage.ObjectMetadata{
-		storage.ObjectMetadataTeamID: sbx.Runtime.TeamID,
+		storage.ObjectMetadataTeamID:     sbx.Runtime.TeamID,
+		storage.ObjectMetadataTemplateID: sbx.Runtime.TemplateID,
+		storage.ObjectMetadataOrigin:     storage.ObjectOriginPause,
```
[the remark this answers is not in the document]
ValentaTomas: Fixed for the pause path: it now stamps in.GetTemplateId() (the request's destination template) and origin=pause; checkpoint now stamps origin=snapshot_template. The checkpoint destination (new snapshot-template env) id isn't carried on SandboxCheckpointRequest, so stamping it there needs a proto field — deferred as a follow-up.

packages/orchestrator/pkg/server/sandboxes.go
```diff
@@ -714,7 +714,7 @@ func (s *Server) Checkpoint(ctx context.Context, in *orchestrator.SandboxCheckpo
 
 	sbxlogger.E(sbx).Info(ctx, "Checkpointing sandbox")
 
-	res, err := s.snapshotAndCacheSandbox(ctx, sbx, in.GetBuildId())
+	res, err := s.snapshotAndCacheSandbox(ctx, sbx, in.GetBuildId(), sbx.Runtime.TemplateID, storage.ObjectOriginSnapshotTemplate)
```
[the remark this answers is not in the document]
ValentaTomas: Fixed: SandboxCheckpointRequest now carries template_id; the API passes the destination snapshotTemplateEnvID and the orchestrator stamps in.GetTemplateId() (instead of the source sbx.Runtime.TemplateID), so snapshot-template uploads are indexed under the destination template.

tvi (approved):

packages/orchestrator/orchestrator.proto
```diff
@@ -159,6 +159,7 @@ message SandboxPauseResponse {
 
 message SandboxCheckpointRequest {
   string sandbox_id = 1;
+  string template_id = 2;
```
dobrac: can we find out if anything ever had the id 2? If yes, we need to use 4+

dobrac: this is only metadata and not required for any actual logic, correct? If thats the case, we should pass it as obscure "Record<string, string> type"

ValentaTomas: Checked the history: field 2 was never used on SandboxCheckpointRequest — the earliest definition already had only sandbox_id=1 and build_id=3, and the gap just mirrored SandboxPauseRequest (where 2=template_id). To avoid any ambiguity I moved the new field to 4 anyway.

ValentaTomas: Agreed — done. Replaced the typed template_id with map<string,string> metadata (field 4). The orchestrator treats it as opaque and forwards it verbatim to the object metadata, so no field-level knowledge is needed; the API passes {template_id: <env>} and future provenance keys need no proto change.

dobrac (approved):

Issue #3012, opened 2026-06-15
Optimize Firecracker sandbox startup by reducing mount namespace copies in the orchestrator


emailcannotbeblank:
## Background

We are planning to use E2B infra for large-scale sandbox workloads, including post-training reinforcement learning for large language models and commercial AI agent workloads. We expect to run a large number of sandboxes on high-core-count machines.

However, we found that under large-scale and high-concurrency workloads, sandbox startup latency becomes very high. It is far beyond Firecracker's expected sub-100 ms startup time. With 100 concurrent sandbox creation requests, creating a single sandbox can take more than 1 second in the orchestrator Firecracker startup path.

We investigated this issue and found that the high latency is mainly caused by lock contention during mount namespace operations.

Specifically, in `packages/orchestrator/pkg/sandbox/fc/process.go`, a large amount of time is spent in the Firecracker startup/configure path, especially around:

```go
err = socket.Wait(startCtx, p.firecrackerSocketPath)
```

This is related to how the Firecracker environment is prepared before the Firecracker process becomes ready.

The current flow contains mount namespace related operations such as:

```bash
unshare -m
ip netns exec ns-xxx firecracker ...
```

Both commands can copy the mount tree and eventually reach `copy_mnt_ns` in the Linux kernel file `fs/namespace.c`. When the mount tree is copied, the kernel takes the namespace lock:

```c
static inline void namespace_lock(void)
{
    down_write(&namespace_sem);
}
```

With many concurrent sandbox creation requests, multiple threads contend for this global lock, which makes sandbox startup latency increase sharply.

In addition:

1. Copying a mount tree requires traversing the mount tree, which is expensive when the host mount tree is large.
2. The previous flow performs more than one mount namespace related operation in the Firecracker startup path, including mount tree copies and mount namespace release. These operations can all touch the same namespace lock path.

We also ran a benchmark after creating 1000 sandboxes and then destroying them.

The environment is:

```text
Kernel: 6.6.0-135.0.0.113.oe2403sp3.x86_64
CPU: AMD EPYC 9755
Cores: 512
Preparation: create 1000 sandboxes, then destroy them
```

In the table below, `socket.Wait` is included in the total orchestrator startup/configure time. It is the key step inside the measured orchestrator path, not a separate end-to-end measurement.

| Concurrency | Before total | Before `socket.Wait` | After total | After `socket.Wait` |
|---:|---:|---:|---:|---:|
[...]

PR #3037, opened 2026-06-18
perf(orchestrator): replace `ip netns exec` with `nsenter` to halve mount namespace copies on sandbox startup

AdaAibaby:
https://github.com/e2b-dev/infra/issues/3012
`p.configure()` latency under 100+ concurrent sandboxes:
image
 
image

/cc @jakubno  @dobrac  @ValentaTomas  Would appreciate your review on this change, thanks!


dobrac (changes requested):
Thank you for the contribution!

Can you please verify the mount propagation is still set to private as with `unshare -m`? I believe the `syscall.CLONE_NEWNS` does not carry the same isolation.

AdaAibaby:
@dobrac   — verified on the orchestrator host (kernel 6.8.0-55-generic, util-linux 2.39.3)
  using bpftrace to trace the `mount` tracepoint at the kernel syscall boundary:

  ```bash
  sudo bpftrace -e '
  tracepoint:syscalls:sys_enter_unshare /args->unshare_flags & 131072/ {
      printf("[%s pid=%d] unshare(CLONE_NEWNS=0x%x)\n", comm, pid, args->unshare_flags);
  }
  tracepoint:syscalls:sys_enter_mount /(args->flags & 278528) == 278528/ {
      printf("[%s pid=%d] mount(target=%s, MS_REC|MS_PRIVATE=0x%x)\n",
             comm, pid, str(args->dir_name), args->flags);
  }
  '

  Running both paths concurrently:

  === path A: old — unshare -m ===
  [unshare pid=3291031] unshare(CLONE_NEWNS=0x20000)
  [unshare pid=3291031] mount(target=/, MS_REC|MS_PRIVATE=0x44000)

  === path B: new — CLONE_NEWNS via SysProcAttr + script-first mount --make-rprivate / ===
  [unshare pid=3291033] unshare(CLONE_NEWNS=0x20000)
  [mount   pid=3291033] mount(target=/, MS_REC|MS_PRIVATE=0x44000)

  Both paths produce the identical kernel syscall sequence — unshare(CLONE_NEWNS=0x20000)
  followed immediately by mount(MS_REC|MS_PRIVATE=0x44000, target=/). The only difference
  is which binary issues the second call: in the old path the unshare binary does it
  implicitly before exec; in the new path the mount --make-rprivate / at the top of the
  start script does it explicitly. The resulting propagation state is identical.

  Also updated 5b55eed to correct the Unshareflags comment (CLONE_NEWNS alone does not
  carry MS_PRIVATE; private isolation depends on the explicit mount --make-rprivate / as
  the invariant first command of both script templates).
  ```

PR #3039, opened 2026-06-18
feat(orch): opt-in DSCP marker for sandbox egress (SANDBOX_EGRESS_DSCP)

tomassrnka:
Opt-in DSCP marker on sandbox egress. Off by default.

## Config

`SANDBOX_EGRESS_DSCP` (uint8, default `0`, range `0..63`). `0` disables; `8` (CS1) is the recommended value. Range-validated in `ParseConfig`.

When disabled: no iptables rule, no setsockopt, no `xt_DSCP` kernel-module dependency.

## What it touches

Two paths, both needed for the marker to actually land on the wire:

1. **Direct egress** (UDP/ICMP/raw IP/non-proxied TCP) — iptables `mangle POSTROUTING -o vpeer -j DSCP --set-dscp <n>` inside the sandbox netns.
2. **tcpfirewall TCP proxy hop** — `IP_TOS` + `IPV6_TCLASS` set on the upstream socket via `Dialer.Control` / `ControlContext`. Without this, every TCP connection from a sandbox loses the mark at the local proxy because the upstream socket is host-originated.

IPv6 supported (sets both family options, tolerates `ENOPROTOOPT` from the non-matching one).

## Test plan

- [x] Default-off: deploy without the env var → no mangle rule, orchestrator boots without `xt_DSCP`.
- [x] Enabled: `SANDBOX_EGRESS_DSCP=8` → mangle rule present; `tcpdump` on the host gateway shows `tos 0x20` for TCP/UDP/ICMP from the sandbox.
- [ ] IPv6: TCP to an IPv6-only host carries Traffic Class `0x20`.
- [x] Out of range: `SANDBOX_EGRESS_DSCP=99` → orchestrator refuses to start.


packages/orchestrator/pkg/sandbox/network/network.go
```diff
@@ -250,6 +250,15 @@ func (s *Slot) CreateNetwork(ctx context.Context) (retErr error) {
 		return fmt.Errorf("error creating postrouting rule from vpeer: %w", err)
 	}
 
+	// Tag every packet leaving the sandbox netns with a DSCP marker so downstream
+	// firewalls can distinguish sandbox traffic at L3 regardless of protocol.
+	// Rule lives inside the sandbox netns so it auto-disappears with the namespace
+	// and cannot be flushed from inside the guest (the FC VM only sees virtio-net).
```
arkamar:
```suggestion
```
I would keep only two lines with motivation, those two suggested for removal should be obvious.

arkamar (commented):
The change makes sense to me, I am just wondering if we should make this optional/configurable only for those who need it.

tomassrnka:
> The change makes sense to me, I am just wondering if we should make this optional/configurable only for those who need it.

You can always just set it to 0 by default and that disables this. Maybe env variable would be good next steps though.

packages/orchestrator/pkg/tcpfirewall/handlers.go
```diff
@@ -24,8 +25,39 @@ const (
...
 
 	noHostnameValue = ""
+
+	// sandboxEgressTOS is the IP TOS-byte form of the DSCP CS1 marker applied
+	// to sandbox netns egress by pkg/sandbox/network (--set-dscp 8). DSCP
+	// occupies the top 6 bits of the TOS byte, so 8 << 2 = 32. We set it on
+	// the proxy's upstream sockets so host-originated TCP carries the same
+	// marker as direct (non-proxied) sandbox egress — without this, the
+	// REDIRECT-to-local-proxy hop strips the L3 marker for all TCP traffic.
+	sandboxEgressTOS = 32
```
arkamar: The biggest problem I have with this PR right now is that we have two numbers, DSCP (8) and TOS (32 = 8 << 2), representing the same thing in two different places. It's a ticking bomb, we will have a problem if anyone will change one of them and will forgot about the second one.

I think we should have some sort of shared truth.

packages/orchestrator/pkg/sandbox/network/dscp_egress_test.go
```diff
@@ -0,0 +1,81 @@
...
+	config, err := ParseConfig()
+	require.NoError(t, err)
+
+	// High, fixed index: Populate fills the pool from low indices, so this is
+	// unlikely to collide with a running orchestrator or other tests.
+	const idx = 30000
+	slot, err := NewSlot("dscp-egress-test", idx, config, NewNoopEgressProxy())
+	require.NoError(t, err)
+
+	require.NoError(t, slot.CreateNetwork(context.Background()))
```
arkamar:
```suggestion
	require.NoError(t, slot.CreateNetwork(t.Context()))
```

packages/orchestrator/pkg/sandbox/network/network.go
```diff
@@ -250,6 +250,13 @@ func (s *Slot) CreateNetwork(ctx context.Context) (retErr error) {
 		return fmt.Errorf("error creating postrouting rule from vpeer: %w", err)
 	}
 
+	// Tag every packet leaving the sandbox netns with a DSCP marker so downstream
+	// firewalls can distinguish sandbox traffic at L3 regardless of protocol.
+	err = tables.Append("mangle", "POSTROUTING", "-o", s.VpeerName(), "-j", "DSCP", "--set-dscp", sandboxEgressDSCP)
```
arkamar: I am still a bit worried that adding this unconditionally could potentially break other deployments. I believe we should introduce the guard in this PR.

arkamar (changes requested):

packages/orchestrator/pkg/sandbox/network/network.go
tomassrnka: yes, let's do it properly via env variable

tomassrnka:
@arkamar ready for re-review — both your CHANGES_REQUESTED points addressed in [`193b274ed`](https://github.com/e2b-dev/infra/pull/3039/commits/193b274ed):

**1. Single source of truth.** The two constants (`sandboxEgressDSCP="8"` for iptables, `sandboxEgressTOS=32` for setsockopt) are gone. One canonical `network.Config.SandboxEgressDSCP` field; the TOS byte is computed once at `tcpfirewall.New()` (`networkConfig.SandboxEgressDSCP << 2`). Both consumers read from the same source — no drift possible.

**2. Default-disabled guard.** New env var `SANDBOX_EGRESS_DSCP` (uint8, `envDefault:"0"`, range-validated 0..63 in `ParseConfig`). When `0`:
- No iptables `mangle POSTROUTING` rule is added in `CreateNetwork`.
- `markDSCP` returns `nil` immediately — zero setsockopt calls.
- No `xt_DSCP` kernel-module dependency.

The existing test sets `config.SandboxEgressDSCP = 8` explicitly so it still exercises the enabled path end-to-end.

The PR title + body have been rewritten to reflect the opt-in shape. Follow-up commit [`19d2672a0`](https://github.com/e2b-dev/infra/pull/3039/commits/19d2672a0) removes an unused `context` import (lint catch after the `t.Context()` swap).

packages/orchestrator/pkg/tcpfirewall/handlers.go
```diff
@@ -26,6 +27,40 @@ const (
 	noHostnameValue = ""
 )
 
+// sandboxEgressTOS is the IP TOS-byte form of network.Config.SandboxEgressDSCP
+// (top 6 bits of the TOS byte, i.e. DSCP << 2). Set once by Proxy.New() so the
+// per-connection markDSCP path doesn't need to thread the value through every
+// handler signature. 0 = disabled (no setsockopt), matching the env default.
+var sandboxEgressTOS uint8
```
arkamar: Not an issue right now, but `-race` parameter will make the tests flaky if there will be more tests calling `tcpfirewll.New()` in the future.

arkamar (approved):
I would suggest to trim some comments a little bit, they are quite long and they make the code harder to consume, imho. Otherwise, LGTM.

packages/orchestrator/pkg/tcpfirewall/handlers.go
tomassrnka: fixed & fixed

PR #3041, opened 2026-06-18
Trace template build timeout sources

dobrac:
## Summary
- record context-deadline details when template builds are collapsed to `build timed out`
- add a dedicated span around async template layer upload waits
- annotate snapshot upload spans with template/build/team/layer identifiers and record upload failures

## Verification
- `gofmt`
- `go test ./pkg/template/build/metrics`

Linux orchestrator package compile was not runnable locally on macOS because the userfaultfd/cgo path requires a Linux C toolchain.


rguliyev (approved):
lg

PR #3047, opened 2026-06-19
feat(envd): bound stalled transfers with idle write/read timeouts

mishushakov:
envd serves long-lived streams over HTTP/1.1, so a peer that stalls mid-transfer can pin a handler — and the process/file behind it — indefinitely. This adds a `netlimit` package with a per-write idle timeout (applied globally via a listener wrapper, since a write deadline only arms on genuine TCP backpressure and is safe everywhere) and a per-read idle timeout scoped to file upload (`PostFiles`), where "no bytes for N seconds" is pathological — unlike `StreamInput`/process streams where a quiet gap is normal. It also sets `ReadHeaderTimeout` for slowloris protection (header read only, never bodies or streams) and bumps the envd version to 0.6.7. Both timeouts reset per chunk, so slow-but-progressing transfers survive and only true stalls are reclaimed.


mishushakov:
@codex review
@claude review

mishushakov:
Thanks @gemini-code-assist — I looked into this and it doesn't apply on our Go version (1.26). TCP keep-alive isn't affected by wrapping the accepted conn:

- `http.Server` never sets `SO_KEEPALIVE` itself. Its only `*net.TCPConn` assertions are for `sendfile`/`ReadFrom` and `Hijacker` — not keep-alive. `Server.doKeepAlives()` is *HTTP*-level connection reuse, unrelated to TCP keep-alive and unaffected by the wrapper.
- TCP keep-alive is enabled by the **net package at accept time**, inside the underlying `*net.TCPListener.Accept()`, before `writeTimeoutListener` ever wraps the conn (`net/tcpsock_posix.go` → `newTCPConn`, which enables keep-alive when `ListenConfig.KeepAlive == 0`, i.e. the zero value). It's a socket option on the fd, so wrapping the Go object can't disable it.
- `http.Server.ListenAndServe` itself just calls `net.Listen("tcp", addr)` with the zero-value `ListenConfig`. Our `(&net.ListenConfig{}).Listen(...)` is exactly equivalent, so keep-alive behavior is identical to before this change.

The `tcpKeepAliveListener` that `ListenAndServe` used to install was removed once `ListenConfig` started enabling keep-alives, so there's nothing for the wrapper to bypass here.

(One real effect of wrapping — losing the `sendfile` fast path because we deliberately don't implement `io.ReaderFrom` — is intentional: a single sendfile deadline would span an entire download and kill a legitimate large-but-progressing transfer. See the comment in `netlimit/writedeadline.go`.)

mishushakov:
Thanks @codex — both findings were real and are fixed in 689620e5c.

**1. Body reuse / smuggling (upload.go):** confirmed. `shouldReuseConnection()` → `closedRequestBodyEarly()` type-asserts `w.req.Body.(*body)`; replacing `r.Body` with the `IdleReadCloser` wrapper made that assertion fail, so the server couldn't detect an aborted large upload's undrained bytes and would reuse the connection. Fixed by restoring `r.Body` to the original `*body` in a defer before `PostFiles` returns (and clearing the read deadline). The wrapper is now only in place while the handler reads; net/http does its own early-close/reuse accounting on the native type, exactly as before.

**2. CloseWrite (writedeadline.go):** confirmed. `closeWriteAndWait()` checks `c.rwc.(closeWriter)`, and embedding only `net.Conn` hid `*net.TCPConn.CloseWrite`. Added a `CloseWrite` method that forwards to the underlying conn (returns `errors.ErrUnsupported` otherwise), so the server can half-close before a full close on mid-upload error replies. (`ReaderFrom`/sendfile stays intentionally unforwarded — a single sendfile deadline would span a whole download.)

A full regression test needs a live keep-alive server since the bug only manifests on connection reuse with a real `*body`; the wrapper-level behavior (CloseWrite forwarding) has unit tests.

ValentaTomas:

PR #3053, opened 2026-06-22
feat(storage): make S3 multipart part size and concurrency configurable

sitole:
## What

S3 `StoreFile` multipart uploader used hardcoded `PartSize=10MB`, `Concurrency=8` (80 MB in flight). Now configurable via env:

| Env var | Default | Effect |
|---|---|---|
| `AWS_UPLOAD_PART_SIZE_MB` | `10` | multipart part size (MB) |
| `AWS_UPLOAD_CONCURRENCY` | `8` | parts in flight |

Defaults match prior hardcoded values, so behavior is unchanged unless set.

## Why

On AWS-backed orchestrators, snapshot upload throughput is capped by the uploader once the source disk stops being the bottleneck (e.g. after provisioning gp3 throughput or moving to local NVMe). Raising part size + concurrency lets the S3 PUT use the available disk/network headroom — e.g. `32` MB × `16` = 512 MB in flight vs the current 80 MB.

## Safety

Out-of-range env values degrade gracefully via the AWS SDK: `PartSize` floored at `MinUploadPartSize` (5 MB), `Concurrency<=0` falls back to `DefaultUploadConcurrency` (5), total parts capped at `MaxUploadParts` (10000).

## Test

`go build ./pkg/storage/` passes. No behavior change at defaults.

### DIFF
--- a/packages/shared/pkg/storage/storage_aws.go
+++ b/packages/shared/pkg/storage/storage_aws.go
@@ -17,6 +17,7 @@ import (
 	"github.com/aws/aws-sdk-go-v2/service/s3/types"
 	"go.uber.org/zap"
 
+	"github.com/e2b-dev/infra/packages/shared/pkg/env"
 	"github.com/e2b-dev/infra/packages/shared/pkg/logger"
 )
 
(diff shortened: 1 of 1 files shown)


sitole:
todo: add feature flag support

packages/shared/pkg/storage/storage_aws.go
```diff
@@ -51,13 +63,25 @@ func newAWSStorage(ctx context.Context, bucketName string) (*awsStorage, error)
...
+	partSizeMB, err := env.GetEnvAsInt("AWS_UPLOAD_PART_SIZE_MB", awsDefaultUploadPartSizeMB)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse AWS_UPLOAD_PART_SIZE_MB: %w", err)
+	}
+
+	concurrency, err := env.GetEnvAsInt("AWS_UPLOAD_CONCURRENCY", awsDefaultUploadConcurrency)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse AWS_UPLOAD_CONCURRENCY: %w", err)
+	}
+
```
sitole, line 75: Feature flag support

PR #3073, opened 2026-06-24
chore(client-proxy): mark enable autoresume to true by default

jakubno:
Turn the flag to true by default - it's the default we use in our deployments


arkamar (approved):
Could you please mention in the merge commit message that this is the default we use in our deployments, which should be reasonable motivation for those who will dig for this in git history.

PR #3079, opened 2026-06-24
feat(orchestrator): graceful template-build drain on shutdown

wj-e2b:
Actually honor config.ShutdownDrainTimeout. Before, wg.Wait() and time.Sleep(15s) ignored closeCtx entirely — the select only checked the context once at entry. So a configured shutdown timeout did nothing during the template drain: the process would block past the deadline until builds finished on their own (or indefinitely on a stuck build), then always burn the full 15s.

Also makes Wait return a meaningful error at the deadline. That return value is unused, but it's the signal the later commits wires into forceStopTemplateBuilds().

### DIFF
--- a/packages/orchestrator/pkg/template/server/main.go
+++ b/packages/orchestrator/pkg/template/server/main.go
@@ -27,11 +27,17 @@ import (
 	templatemanager "github.com/e2b-dev/infra/packages/shared/pkg/grpc/template-manager"
 	"github.com/e2b-dev/infra/packages/shared/pkg/logger"
 	"github.com/e2b-dev/infra/packages/shared/pkg/storage"
+	"github.com/e2b-dev/infra/packages/shared/pkg/utils"
 )
 
 type closeable interface {
 	Close() error
 }
+
+// consumerStatusCheckGracePeriod is how long the graceful drain waits, after
+// in-flight builds finish, for consumers to read the final build status before
+// the server shuts down.
+const consumerStatusCheckGracePeriod = 15 * time.Second
 
 type ServerStore struct {
 	templatemanager.UnimplementedTemplateServiceServer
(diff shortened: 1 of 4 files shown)


packages/shared/pkg/utils/apply_all_or_none.go
```diff
@@ -3,23 +3,26 @@ package utils
...
 
 	for _, update := range updates {
 		rollback, err := update(ctx)
 		if err != nil {
-			rollbackCtx := context.WithoutCancel(ctx)
+			rollbackCtx, rollbackCancel := WithoutCancelPreservingDeadline(ctx)
 			for i := len(rollbacks) - 1; i >= 0; i-- {
 				rollbacks[i](rollbackCtx)
 			}
+			rollbackCancel()
```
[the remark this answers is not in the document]
wj-e2b: Intended behavior

dobrac (approved):

packages/orchestrator/pkg/template/server/main.go
```diff
@@ -27,12 +27,18 @@ import (
...
 )
 
 type closeable interface {
 	Close() error
 }
 
+// consumerStatusCheckGracePeriod is how long the graceful drain waits, after
+// in-flight builds finish, for consumers to read the final build status before
+// the server shuts down.
+const consumerStatusCheckGracePeriod = 15 * time.Second
```
rguliyev, line 40: can we pass it via args or envs?

rguliyev (approved):

packages/orchestrator/pkg/template/server/main.go
wj-e2b: None of the existing code hoists this to CLI env vars today. We can change it if it ever needs to be tweaked.

PR #3082, opened 2026-06-25
fix: support password in REDIS_URL via redis:// URL scheme

### LINKED ISSUE #3036: The REDIS_URL environment variable does not support passwords.
Background: For security reasons, I intended to use a password-protected Redis, but it seems that REDIS_URL in the api.hcl environment does not support it and reports an error.

Error log: redis://xxxx:xxxx@redis-xxxxxx.redis.xxxx.com:6379: too many colons in address

Root cause: The value of REDIS_URL includes the redis:// scheme prefix, but the code parses the entire string as a host:port format, leading to the “too many colons” error.

Issue confirmation: The code does not support the redis:// URI format. redis.Options.Addr only accepts host:port, and the password needs to be provided separately via the Password field; however, the current RedisConfig structure does not have a Password field.

Proposed solution: Modify the code to support parsing the URI format. In redis.go, within the branch case config.RedisURL != "", use redis.ParseURL() to parse the full URI.

Test results and conclusion: It successfully parses the password and allows Redis to function normally, while remaining compatible with the previous format. I have modified and tested the code. May I contribute it to the community?

orion-gmx:
Fixes #3036

### DIFF
--- a/packages/shared/pkg/factories/redis.go
+++ b/packages/shared/pkg/factories/redis.go
@@ -113,14 +113,18 @@ func NewRedisClient(ctx context.Context, config RedisConfig) (redis.UniversalCli
 
 		redisClient = redis.NewClusterClient(clusterOpts)
 	case config.RedisURL != "":
-		opts := &redis.Options{
-			Addr:                  config.RedisURL,
-			PoolSize:              poolSize,
-			MinIdleConns:          minIdleConns,
-			ConnMaxIdleTime:       -1,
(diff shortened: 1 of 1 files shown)


orion-gmx:
@cla-bot check

tvi:
Hi @orion-gmx , i rebased your change and incorporated it here https://github.com/e2b-dev/runtime/commit/29462afe56fd06b10b77e5ae2cf0b07a1c313f7d . Thank you for your contribution.

PR #3090, opened 2026-06-25
feat(orchestrator): startup resource reclaim package

wj-e2b:
Add the startupreclaim package, which best-effort reclaims host resources
left behind by sandboxes that did not shut down cleanly: orphaned firecracker
processes (matched to ns-<idx> via /proc netns inode, with cmdline-ancestry
fallback), connected NBD devices, slot network namespaces, sandbox cgroups,
and stale sockets/fifos/overlay files. Failures are logged and counted via
OTel metrics, never fatal.

### DIFF
--- a/packages/orchestrator/pkg/startupreclaim/firecracker.go
+++ b/packages/orchestrator/pkg/startupreclaim/firecracker.go
@@ -1,0 +1,115 @@
+//go:build linux
+
+package startupreclaim
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strconv"
+	"syscall"
+
+	"go.uber.org/zap"
+
+	"github.com/e2b-dev/infra/packages/shared/pkg/logger"
+)
+
+// reclaimFirecrackers kills orphaned firecracker process groups by scanning the
+// host process table. The network slots they used are reclaimed separately by
+// the network reclaim.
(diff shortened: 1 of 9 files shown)


packages/orchestrator/pkg/sandbox/network/storage_local.go
dobrac, line 31: why does this need to be public?

dobrac, line 191: if this is used only in tests, not sure if we need the method at all

rguliyev:
@wj-e2b could you please add monitoring for reclaims labeled by instance, reclaim type, status (success, fail), (more?)

wj-e2b:
> @wj-e2b could you please add monitoring for reclaims labeled by instance, reclaim type, status (success, fail), (more?)

This already exists

packages/orchestrator/pkg/startupreclaim/firecracker.go
```diff
@@ -0,0 +1,115 @@
...
+	parts := bytes.Split(data, []byte{0})
+	cmdline := make([]string, 0, len(parts))
+	for _, part := range parts {
+		cmdline = append(cmdline, string(part))
+	}
+
+	return cmdline, nil
+}
+
+// firecrackerBinaryName mirrors fc.FirecrackerBinaryName; it is duplicated here
```
dobrac, line 103: can we maybe refactor the const so it can be reused?

dobrac (approved):

PR #3091, opened 2026-06-25
orchestrator: index builds into storage-api (ingest client)

ValentaTomas:
## Summary
- Best-effort, feature-flagged (`storage-api-ingest-build`) client that ingests builds into storage-api after their headers are uploaded; a no-op when `STORAGE_API_HOST` is unset, so the orchestrator never hard-depends on the service.
- Minimal ingest-only gRPC proto + generated client (other RPCs added when needed).

## Notes
- Soft error by default (log + continue); the flag can promote it to a hard error.
- Follow-up (tracked): send `lineage_id`/`generation` so storage-api can supersede across a sandbox's pause chain.


ValentaTomas:
Will reopen, for now leaving as reference.

PR #3093, opened 2026-06-26
add multi-distro support for Alpine, CentOS/RHEL besides Debian/Ubuntu

AdaAibaby:
#### Description
Previously custom template builds only supported Debian/Ubuntu images relying on apt package manager, and failed on Alpine, CentOS/RHEL due to distro-specific provisioning scripts.
This change extends template provisioning logic to support mainstream Linux distributions:
Debian / Ubuntu (apt)
Alpine (apk)
CentOS / RHEL (dnf/yum)
The provisioning layer auto-detects target OS distribution, switches to corresponding native package manager, and adapts distro-specific package names & system file paths automatically during template build and provisioning phase.
#### Changes
Add OS distro detection logic to identify base image system
Implement separate provisioning branches for apt / apk / dnf/yum package managers
Normalize package names and file paths compatible with Alpine, CentOS/RHEL
Remove limitation that restricts templates to Debian-family images

/cc @jakubno  @dobrac @ValentaTomas Would appreciate your review on this change, thanks!

### DIFF
--- a/packages/envd/internal/services/process/handler/handler.go
+++ b/packages/envd/internal/services/process/handler/handler.go
@@ -89,6 +89,17 @@ func currentNice() int {
 	return 20 - prio
 }
 
+// findBinary returns the first existing path from candidates, falling back to
+// the bare name (relying on PATH lookup at exec time) if none exist on disk.
+func findBinary(name string, candidates ...string) string {
+	for _, p := range candidates {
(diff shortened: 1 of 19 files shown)


AdaAibaby:
image
image
image
image
image

jakubno:
Should be solved in #3411

PR #3119, opened 2026-06-28
feat: persist synchronously-available layer sizes to env_builds via gRPC

ValentaTomas:
Persists per-artifact layer sizes on the `env_builds` row, returned over gRPC from the build/pause/snapshot flows (no new GCS object metadata).

Only sizes available **without** blocking on the async memfile dedup header are included here (that header can take 30s+ under background-copy):
- `rootfs_mapped_size_bytes`, `rootfs_diff_size_bytes` (rootfs header is built synchronously)
- `memfile_logical_size_bytes` (from the base memfile header)

rootfs logical size is already stored as `total_disk_size_mb`. The truly-async memfile mapped/diff sizes are handled in a follow-up PR (written to the memfile data object metadata during upload).

Wiring:
- proto: new `LayerSizes` message on `SandboxPauseResponse`, `SandboxCheckpointResponse`, and `TemplateBuildMetadata`
- orchestrator computes them from the resolved headers (`Snapshot.SyncLayerSizes`, `getLayerSizes` for template builds)
- API persists via new `SetEnvBuildLayerSizes` query (pause/snapshot are best-effort, non-fatal)

Most of the diff is generated (proto `.pb.go`, sqlc `.sql.go`).


ValentaTomas:
Superseded by #3122: per the latest direction, all layer sizes (logical/mapped/diff, both artifacts) go via GCS object metadata and the gRPC/env_builds path is dropped for now.

PR #3120, opened 2026-06-28
feat(orchestrator): surface soft-delete reason group on tombstone check

ValentaTomas:
The storage-index soft-delete tombstone value is `<reason>:<action_id>`, but the orchestrator only logged the raw marker and the check metric had no reason dimension.

This parses the marker and:
- adds a bounded `reason` group dimension to `orchestrator.build.soft_delete_check` (unexpected/high-cardinality values collapse to `other`)
- logs `reason` and `action_id` as separate fields on the soft-deleted log line

No behavior change to enforcement.


ValentaTomas:
Used for demonstration. Referenced (related) from Linear EN-1184.

PR #3122, opened 2026-06-28
feat(orchestrator): write layer sizes (logical/mapped/diff) to object metadata

ValentaTomas:
Writes each layer's `logical-size`, `mapped-size`, and `diff-size` (all uncompressed, decimal bytes) as custom metadata on the memfile and rootfs data objects, derived from the resolved diff header during the background upload.

Sizes live on the data object rather than the build row because the memfile values depend on the async dedup header, which resolves after the build row is finalized. Deriving them from the header adds no latency to the pause/checkpoint response.

Supersedes #3119 (the gRPC-to-env_builds approach), which is closed.


packages/orchestrator/pkg/sandbox/build_upload.go
```diff
@@ -92,6 +94,28 @@ func NewUpload(
 	return u, nil
 }
 
+// layerSizeMetadata adds the layer's logical, mapped, and diff sizes (all
+// uncompressed, from the diff header) to the base object metadata. They live on
+// the data object because the memfile values depend on the async dedup header.
+func (u *Upload) layerSizeMetadata(h *headers.Header) storage.ObjectMetadata {
+	md := make(storage.ObjectMetadata, len(u.objectMetadata)+3)
```
dobrac: is this always +3?

dobrac (approved):

packages/orchestrator/pkg/sandbox/build_upload.go
ValentaTomas: Only when the diff header is present — reworked so the +3 hint applies only on the path that actually writes the three size keys; the nil-header path now allocates just the base metadata.

PR #3134, opened 2026-06-29
fix(docker-reverse-proxy): fix response body handling in proxy and token acquisition

AdaAibaby:
/cc @jakubno @dobrac  @ValentaTomas Would appreciate your review on this change, thanks!

Fix two response body handling issues in the docker-reverse-proxy package.

### 1. `ModifyResponse` consumes response body without restoring it (`store.go`)

The `ModifyResponse` callback reads the entire response body via `io.ReadAll` on
401 Unauthorized responses for logging purposes, but never restores it. Since
`httputil.ReverseProxy` forwards `resp.Body` to the client *after*
`ModifyResponse` returns, the client receives an **empty 401 response body**
with no error details.

**Fix:** Close the consumed body, restore it with `io.NopCloser(bytes.NewReader(...))`,
and update `ContentLength` so the proxy writes the correct `Content-Length` header.

### 2. Potential panic + duplicate defer in `getToken` (`token.go`)

```go
body := make([]byte, resp.ContentLength)  // panics if ContentLength == -1

### DIFF
--- a/packages/docker-reverse-proxy/internal/handlers/store.go
+++ b/packages/docker-reverse-proxy/internal/handlers/store.go
@@ -49,12 +49,16 @@ func NewStore(ctx context.Context) *APIStore {
 	// Custom ModifyResponse function
 	proxy.ModifyResponse = func(resp *http.Response) error {
 		if resp.StatusCode == http.StatusUnauthorized {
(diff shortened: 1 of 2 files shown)


jakubno:

PR #3135, opened 2026-06-29
fix(api): use correct error variable in processCustomErrors

AdaAibaby:
When multiple security schemes are configured and the first error is ErrNoAuthHeader (skipped by continue), the TeamForbiddenError or TeamBlockedError branch was using the wrong variable `err` (which still pointed to unwrapped[0]) instead of the `errors.As`-extracted typed variable.

This caused the error message to contain "authorization header is missing" instead of the actual forbidden/blocked reason.

Fix: use `teamForbidden.Error()` / `teamBlocked.Error()` (the concrete type extracted by `errors.As`) instead of `err.Error()`. This is more precise than using `errW.Error()`, because `errW` may contain wrapper text (e.g. "failed getting team: ...") while the typed variable returns only the clean user-facing message.

Add comprehensive tests covering:
- TeamForbidden/TeamBlocked after ErrNoAuthHeader (the bug scenario)
- TeamForbidden/TeamBlocked as the only error
- Wrapped TeamForbidden/TeamBlocked errors
- Generic error after ErrNoAuthHeader
- All errors being ErrNoAuthHeader


dobrac (approved):

AdaAibaby:
CI workflows are awaiting maintainer approval to run. Could you approve the workflow? @dobrac Thanks!

PR #3140, opened 2026-06-29
fix(envd): fix process lookup by tag and nil panic in Wait()

AdaAibaby:
Two bugs in the process service:

1. getProcess Range callback return values were inverted:
   - Returned true (continue) on match — should return false (stop)
   - Returned false (stop) on non-match — should return true (continue) This caused tag-based process lookup to fail randomly: it would stop at the first non-matching tagged process instead of continuing to search, and would keep iterating after finding a match.

2. Handler.Wait() unconditionally called p.tty.Close(), but p.tty is nil for non-PTY processes (those started via cmd.Start() at line 429). Every non-PTY process exit would panic with a nil pointer dereference.

### DIFF
--- a/packages/envd/internal/services/process/handler/handler.go
+++ b/packages/envd/internal/services/process/handler/handler.go
@@ -450,7 +450,9 @@ func (p *Handler) Wait() {
 
 	err := p.cmd.Wait()
 
-	p.tty.Close()
+	if p.tty != nil {
+		p.tty.Close()
+	}
 
 	var errMsg *string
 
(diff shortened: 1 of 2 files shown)


leonmeijer:
This is the behavioral superset of #3099 because it also fixes the non-PTY Wait nil dereference, but it omits #3099’s focused lookup regression cases and the required envd behavioral version bump. Please consolidate those pieces into one rebased PR and close the duplicate rather than merge both.

leonmeijer (commented):
The process lookup and non-PTY wait behavior changes leave packages/envd/pkg/version.go at 0.6.7. The repository requires every envd behavior change to bump this value; without it, X-Envd-Version and rollout compatibility logic cannot distinguish binaries with and without these fixes. Please include the version bump.

PR #3145, opened 2026-06-30
fix(envd): use constant-time comparison for signature validation

AdaAibaby:
Replace string inequality operator (!=) with crypto/subtle.ConstantTimeCompare for signature validation in auth.go. The previous implementation was vulnerable to timing attacks, where an attacker could potentially determine the correct signature byte-by-byte by measuring response times.

This is a standard security best practice for comparing cryptographic values such as HMAC signatures, tokens, and hashes.

### DIFF
--- a/packages/envd/internal/api/auth.go
+++ b/packages/envd/internal/api/auth.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"crypto/subtle"
 	"errors"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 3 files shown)


jakubno (dismissed):
@AdaAibaby , could you please fix the lint issue?

AdaAibaby:
> @AdaAibaby , could you please fix the lint issue?

Done. Does this look OK?

arkamar (changes requested):
please, also do the envd version bump.

AdaAibaby:
> please, also do the envd version bump.

Done. Does this look OK?

arkamar (approved):

PR #3138, opened 2026-06-29
fix: handle `InvalidObjectState` error for archived S3 objects

orion-gmx:
## Summary

When objects are transitioned to archived storage classes (e.g. `GLACIER`, `ARCHIVE_FR`, `COLD_ARCHIVE`) via lifecycle rules, S3-compatible backends return an `InvalidObjectState` error on `GetObject`/`HeadObject` instead of the usual `NoSuchKey`. Previously this was unhandled, causing sandbox creation to fail with a generic "Internal" gRPC error.

### `packages/shared/pkg/storage/storage.go`
- Added `ErrObjectArchived` sentinel error for archived/inaccessible objects

### `packages/shared/pkg/storage/storage_aws.go`
- `WriteTo()`: handle `types.InvalidObjectState` → `ErrObjectArchived`
- `OpenRangeReader()`: handle `types.InvalidObjectState` → `ErrObjectArchived`
- `Size()`: handle `types.InvalidObjectState` → `ErrObjectArchived`

### `packages/shared/pkg/storage/metrics.go`
- Added `OutcomeArchived` constant for metrics tracking
- `Outcome()` function now maps `ErrObjectArchived` to the `"archived"` outcome

### `packages/orchestrator/pkg/server/sandboxes.go`
- `Create()`: check for `ErrObjectArchived` and return `FailedPrecondition` with a clear message telling users to rebuild the template

## Testing

- Verified compilation of both `shared/pkg/storage` and `orchestrator/pkg/server` packages
- The `InvalidObjectState` error type is part of the AWS SDK v2 (`github.com/aws/aws-sdk-go-v2/service/s3/types`) and is returned when attempting to read objects in `GLACIER`, `ARCHIVE`, or similar storage classes

Closes #3137

PR #3147, opened 2026-06-30
Improve/archived object map aws error

AdaAibaby:
refactor: extract mapAWSError helper to reduce duplication

Consolidate the duplicated AWS error → sentinel mapping across WriteTo, OpenRangeReader, and Size into a single [mapAWSError()](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) helper function.

This also fixes an inconsistency where Size() checked types.NotFound but the other two methods did not.

Changes: 1 file, +23 -34 (net reduction of 11 lines)


leonmeijer:
This contains #3138 verbatim plus the centralized AWS error mapper, so it should be the single consolidation target. Preserve the existing operation-specific context for unmapped `OpenRangeReader` errors while rebasing, then close #3138 instead of merging both.

PR #3152, opened 2026-07-01
feat(orchestrator): LD-gated ClickHouse write fan-out feature flag

rguliyev:
Add ClickhouseWriteFanoutFlag (defaults false); 
flipping it true in LD enables writes to additional ClickHouse endpoints without a redeploy. 
Default DSN writes are never gated.

### DIFF
--- a/packages/clickhouse/pkg/hoststats/delivery.go
+++ b/packages/clickhouse/pkg/hoststats/delivery.go
@@ -45,6 +45,12 @@ type ClickhouseDelivery struct {
 	conn    driver.Conn
 }
 
+type GatedClickhouseDelivery struct {
+	*ClickhouseDelivery
+
+	ff *featureflags.Client
+}
+
 var tracer = otel.Tracer("github.com/e2b-dev/infra/packages/clickhouse/pkg/hoststats")
 
 const DefaultBatcherName = "sandbox-host-stats"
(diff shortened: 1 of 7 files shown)


packages/clickhouse/pkg/hoststats/delivery.go
```diff
@@ -96,9 +106,15 @@ func (c *ClickhouseDelivery) Push(stat SandboxHostStat) error {
 	return c.batcher.Push(stat)
 }
 
-// Close waits for queued items to flush. ctx is ignored: honoring it would
-// leak the flush goroutine onto a connection the caller is about to tear
-// down. Real deadlines require plumbing ctx through batcher.Stop.
+func (c *GatedClickhouseDelivery) Push(stat SandboxHostStat) error {
+	if c.ff != nil && c.ff.BoolFlag(context.Background(), featureflags.ClickhouseWriteFanoutFlag) {
```
dobrac, line 110: would it be possible to pass proper context here to have the ability to do e.g. gradual rollout or some targeting?

dobrac (approved):

packages/clickhouse/pkg/hoststats/delivery.go
rguliyev: Would need to change `hoststats.Delivery.Push` to take a ctx, plus every caller, doable but much wider than this PR.

That said, for this flag:

- LD already attaches the service/environment context automatically (see #3100), so percentage rollouts across orchestrator instances work today with no per-request ctx.
- Per-tenant / per-sandbox targeting on a write kill switch is unusual — you'd end up with half a tenant's stats written and half dropped across the fleet.
- The events side already takes ctx (Publish has one), so if per-request targeting ever becomes needed there, it's available without any interface change.

PR #3155, opened 2026-07-01
fix: WrapContextAsUserError should not misclassify internal timeouts as user cancellations

### LINKED ISSUE #3154: Bug: WrapContextAsUserError misclassifies internal timeouts as user-initiated cancellations
## Bug Description

`WrapContextAsUserError` in `packages/orchestrator/pkg/template/build/builderrors/errors.go` treats **all** errors containing `context.Canceled` as user-initiated cancellations, reporting `"build was cancelled"`. However, `context.Canceled` can also originate from **internal timeouts** (e.g., envd init timeout in `WaitForEnvd`), which are not user cancellations. This causes misleading error messages that make debugging extremely difficult.

## Root Cause

The error propagation chain:

1. **`WaitForEnvd`** (`sandbox.go:~1790`) creates a child context via `context.WithCancelCause(ctx)`. When the 60s timeout fires, it calls `cancel(errors.New("syncing took too long"))`.

2. **`doRequestWithInfiniteRetries`** (`envd.go:93`) detects `ctx.Done()` and returns:
   ```go
   return nil, requestCount, fmt.Errorf("%w with cause: %w", ctx.Err(), context.Cause(ctx))
   // ctx.Err() == context.Canceled  ← wrapped into the error chain
   ```

3. The error bubbles up through `initEnvd` → `WaitForEnvd` → `CreateSandbox.Sandbox()` → `BuildLayer` → `Build`.

4. **`Build`** (`builder.go:~193`) has a deferred call:
   ```go
   e = builderrors.WrapContextAsUserError(e)
   ```
   Which checks:
   ```go
   if errors.Is(err, context.Canceled) {
       return phases.NewPhaseBuildError(phases.PhaseMeta{}, ErrCanceled)
   }
   ```
   Since the error chain contains `context.Canceled` (from step 2), it matches — and the **real cause** ("syncing took too long" / envd failed to start) is **discarded and replaced** with `"build was cancelled"`.

## Impact

- Users see `"build was cancelled"` when their build actually failed due to envd startup timeout, a crashed Firecracker process, or other internal issues.
- The real failure reason is completely lost, making it nearly impossible for users to diagnose and fix their builds.
- Metrics that distinguish user cancellations from internal failures are polluted.

## Reproduction

Build a template with a base image where envd takes longer than 60 seconds to initialize (e.g., a heavy image on a slow network where provisioning installs many packages). The build will fail with `"build was cancelled"` instead of a message indicating envd startup timeout.

## Suggested Fix

`WrapContextAsUserError` should only classify `context.Canceled` as a user cancellation when the **Build-level context** itself was canceled (i.e., via `TemplateBuildDelete` or the cancellation watcher in `create_template.go`). Internal child-context cancellations should not be treated as user cancellations.

One approach: check `ctx.Err()` of the Build context directly in the defer, rather than scanning the error chain with `errors.Is`:

```go
defer func() {
    if ctx.Err() != nil {
        // ctx is the build-level context, so this IS a user/external cancellation
        e = errors.Join(e, ctx.Err())
        e = builderrors.WrapContextAsUserError(e)
    }
    // Otherwise, do not wrap — let the real error propagate
}()
```

Alternatively, `WrapContextAsUserError` could accept the build context and only match when `ctx.Err() == context.Canceled`, rather than using `errors.Is` which traverses the entire wrapped error chain.

## Affected Files

- `packages/orchestrator/pkg/template/build/builderrors/errors.go` — `WrapContextAsUserError`
- `packages/orchestrator/pkg/template/build/builder.go:~185-194` — deferred error wrapping
- `packages/orchestrator/pkg/sandbox/envd.go:93` — wraps `ctx.Err()` (`context.Canceled`) into the error chain

AdaAibaby:
## Summary

Fixes https://github.com/e2b-dev/infra/issues/3154

`WrapContextAsUserError` previously used `errors.Is(err, context.Canceled)` which traverses the entire error chain. This caused internal child-context cancellations (e.g., `WaitForEnvd` 60s timeout calling `cancel(errors.New("syncing took too long"))`) to be misclassified as user-initiated cancellations, replacing the real error with a misleading `"build was cancelled"` message.

### `builderrors/errors.go`
- `WrapContextAsUserError` now accepts the **build-level context** as a parameter
- Instead of `errors.Is(err, context.Canceled)`, it checks `buildCtx.Err()` directly
- Only classifies as user cancellation/timeout when the build context itself is done
- Internal child-context cancellations now propagate their real error messages

### `builder.go`
- Updated the call site to pass `ctx` (the build-level context) to `WrapContextAsUserError`

### `builderrors/errors_test.go` (new)
- Added 6 test cases covering:
  - Nil error passthrough
  - Existing user error preserved
  - Internal timeout NOT misclassified (the key regression test)
  - User cancellation correctly classified
  - Build deadline exceeded correctly classified
  - Full WaitForEnvd error chain simulation

## Before vs After

| Scenario | Before | After |
|----------|--------|-------|
| envd fails to start within 60s | `"build was cancelled"` ❌ | Real error: `"failed to init new envd: context canceled with cause: syncing took too long"` ✅ |
| User cancels build | `"build was cancelled"` ✅ | `"build was cancelled"` ✅ |
| Build deadline exceeded | `"build timed out"` ✅ | `"build timed out"` ✅ |

### DIFF
--- a/packages/orchestrator/pkg/template/build/builder.go
+++ b/packages/orchestrator/pkg/template/build/builder.go
@@ -190,7 +190,7 @@ func (b *Builder) Build(ctx context.Context, paths storage.Paths, cfg config.Tem
 			e = errors.Join(e, ctx.Err())
 		}
 
-		e = builderrors.WrapContextAsUserError(e)
+		e = builderrors.WrapContextAsUserError(ctx, e)
 		if e != nil {
 			childSpan.RecordError(e, trace.WithAttributes(
 				telemetry.WithTemplateID(cfg.TemplateID),
--- a/packages/orchestrator/pkg/template/build/builderrors/errors.go
+++ b/packages/orchestrator/pkg/template/build/builderrors/errors.go
@@ -26,9 +26,10 @@ func IsUserError(err error) bool {
 	return phases.UnwrapPhaseBuildError(err) != nil
 }
 
-// WrapContextAsUserError wraps context.Canceled as a user error if no user error already exists.
-// This ensures that user-initiated cancellations are tracked as user errors in metrics.
-func WrapContextAsUserError(err error) error {
+// WrapContextAsUserError wraps context errors as user errors only when the build-level
+// context itself was canceled or timed out. This prevents internal child-context
+// cancellations (e.g., envd init timeout) from being misclassified as user cancellations.
+func WrapContextAsUserError(buildCtx context.Context, err error) error {
 	if err == nil {
 		return nil
 	}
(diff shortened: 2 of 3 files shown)


dobrac (approved):

AdaAibaby:
CI workflows are awaiting maintainer approval to run. Could you approve the workflow? @dobrac Thanks!

PR #3164, opened 2026-07-01
fix(orchestrator): include phase/step context and original error chain in build cancellation messages

### LINKED ISSUE #3160: fix(orchestrator): generic 'build was cancelled' error hides actual failure cause
## Problem

When a template build is cancelled or times out, the orchestrator returns only a generic `build was cancelled` message. This hides the actual failure cause (e.g., envd crash, provisioning failure) and makes debugging very difficult for users.

### Root Cause

In `builderrors.WrapContextAsUserError()`, any `context.Canceled` or `context.DeadlineExceeded` error is replaced with a bare `ErrCanceled` / `ErrTimedOut`, discarding the entire error chain that contains diagnostic information.

### Example

When envd crashes during a build, the actual error chain is:
```
wait for envd: failed to init new envd: failed to init envd: context canceled
```
But the user only sees:
```
build was cancelled
```

## Proposed Fix

Three targeted changes (+53/-9 lines):

1. **`PhaseBuildError.Error()`** — include phase/step context in the error message
2. **`phases.Run()`** — wrap context cancellation errors with the active phase metadata
3. **`WrapContextAsUserError()`** — preserve the original error chain instead of replacing it

### Before / After

| Scenario | Before | After |
|----------|--------|-------|
| Cancel during phases.Run | `build was cancelled` | `context canceled (phase: provision, step: 1)` |
| envd crash | `build was cancelled` | `build was cancelled: wait for envd: failed to init new envd: context canceled` |
| User cancel | `build was cancelled` | `build was cancelled: context canceled` |
| Build timeout | `build timed out` | `build timed out: context deadline exceeded` |

Fully backward compatible — `errors.Is(err, ErrCanceled)` still works.

orion-gmx:
Fixes #3160

## Problem

When a build is cancelled or times out, `WrapContextAsUserError()` replaces the entire error chain with a bare `ErrCanceled` / `ErrTimedOut`, so the user only sees `build was cancelled` with no diagnostic context about which phase failed or what the underlying error was.

### `packages/orchestrator/pkg/template/build/phases/errors.go`
- `PhaseBuildError.Error()` now includes phase and step in the formatted message
- Added nil guard for `Err` field
- Skips appending phase/step suffix when the inner error chain already carries identical metadata (avoids duplication)

### `packages/orchestrator/pkg/template/build/phases/phase.go`
- Added `wrapContextErr` closure in `Run()` that annotates `context.Canceled` / `DeadlineExceeded` errors with the active phase metadata
- Applied to all error return points in the builder loop

### `packages/orchestrator/pkg/template/build/builderrors/errors.go`
- `WrapContextAsUserError()` now extracts phase/step from any existing `PhaseBuildError` and re-wraps with the standardized `ErrCanceled`/`ErrTimeout` sentinel
- Introduced `sanitizedError` wrapper type that cleans up newlines in `Error()` while preserving the original error chain via `Unwrap()`
- When `err` is wrapped by an outer error (e.g. `errors.Join`), the full context is preserved instead of unwrapping to just `pbe.Err`

## Before / After

| Scenario | Before | After |
|---|---|---|
| Cancel during phases.Run | `build was cancelled` | `build was cancelled: context canceled (phase: provision, step: 1)` |
| envd crash | `build was cancelled` | `build was cancelled: wait for envd: failed to init new envd: context canceled` |
| User cancel | `build was cancelled` | `build was cancelled: context canceled` |
| Build timeout | `build timed out` | `build timed out: context deadline exceeded` |

## Backward Compatibility

Fully compatible — `errors.Is(err, ErrCanceled)` and `errors.Is(err, ErrTimedOut)` still work because the sentinels are wrapped with `%w`.

### DIFF
--- a/packages/orchestrator/pkg/template/build/builderrors/errors.go
+++ b/packages/orchestrator/pkg/template/build/builderrors/errors.go
@@ -5,6 +5,8 @@ package builderrors
 import (
 	"context"
 	"errors"
+	"fmt"
+	"strings"
 
 	"github.com/e2b-dev/infra/packages/orchestrator/pkg/template/build/phases"
 	template_manager "github.com/e2b-dev/infra/packages/shared/pkg/grpc/template-manager"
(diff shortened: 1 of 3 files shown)


orion-gmx:
@chatgpt-codex-connector review 83a910fb2

PR #3168, opened 2026-07-01
ci: build infra images for tango

rguliyev:
Adds Tango to the Infra image build/upload matrix.\n\nThis follows the existing workflow pattern for staging, foxtrot, and juliett. The job uses the GitHub environment named `tango` and reads `vars.INFISICAL_MACHINE_IDENTITY_ID` from that environment.\n\nPrerequisite before the workflow can run successfully:\n- the Tango Infra Infisical identity from terraform/infisical has been applied and configured as the `INFISICAL_MACHINE_IDENTITY_ID` GitHub environment variable for `tango`.

### DIFF
--- a/.github/workflows/build-and-upload-images.yml
+++ b/.github/workflows/build-and-upload-images.yml
@@ -22,6 +22,7 @@ jobs:
           - staging
           - foxtrot
           - juliett
+          - tango
     environment: ${{ matrix.environment }}
     concurrency:
       group: build-images-${{ matrix.environment }}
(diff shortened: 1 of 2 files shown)


dobrac (approved):

djeebus (approved):

rguliyev:
Addressed the envd promotion gap called out in the Codex review. The PR now adds Tango to `.github/workflows/promote-envd.yml` as a dispatch target and includes it in the promotion matrix, so the versioned `envd.<sha>` artifact produced by the build workflow can be promoted to the default `envd` object for Tango as well.

PR #3176, opened 2026-07-02
feat(api): discover orchestrators via nomad service

wj-e2b:
Switch discovery to use nomad's service discovery instead of blindly checking every node in the default nodepool. Lets us use multiple nodepools for cutting over orchestrator from system jobs to service jobs.

Existing orchestrator-ee system job already registers to the service.


djeebus (approved):

dobrac (approved):

packages/api/internal/orchestrator/discovery/nomad.go
```diff
@@ -13,45 +13,78 @@ import (
...
 
-	out := make([]Node, 0, len(nodes))
-	for _, n := range nodes {
-		shortID := n.ID
-		if len(shortID) > consts.NodeIDLength {
-			shortID = shortID[:consts.NodeIDLength]
+			out = append(out, Node{
+				ShortID:             shortID,
+				IPAddress:           reg.Address,
+				OrchestratorAddress: net.JoinHostPort(reg.Address, strconv.Itoa(reg.Port)),
```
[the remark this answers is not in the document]
wj-e2b: This is not worse than it is right now, where any node in the node pool has this issue.

packages/api/internal/cfg/model.go
```diff
@@ -48,6 +48,24 @@ type Config struct {
...
+	// NomadOrchestratorLegacyDiscoveryEnabled enables a node-pool-based
+	// discovery FALLBACK unioned with the service-based discovery above: all
+	// ready Nomad nodes in the "default" pool are assumed to run an
+	// orchestrator on the well-known port. It covers orchestrator jobs
+	// deployed from jobspecs that register their service with an empty
+	// Address (pre-port-label-fix), which service discovery skips as
+	// unroutable, and removes any rollout ordering constraint between the
+	// API and orchestrator releases. Set to false once no legacy orchestrator
+	// jobs remain. Used when ServiceDiscoveryProvider=nomad.
+	NomadOrchestratorLegacyDiscoveryEnabled bool `env:"NOMAD_ORCHESTRATOR_LEGACY_DISCOVERY_ENABLED" envDefault:"true"`
```
dobrac, line 67: can we do this using feature flags to allow fast rollbacks?

dobrac (approved):

packages/api/internal/cfg/model.go
wj-e2b: Yes, but it's pretty minor in the scheme of things. You would almost never flip this unless you were completely done migrating. This is for disabling the old default nodepool discovery. This is more for my testing.

dobrac: sounds good, thx

PR #3179, opened 2026-07-02
[EN-1410] feat(clickhouse): per-team TTL on sandbox_events

jakubno:
Add events_ttl_days column stamped on each event so retention can vary. Table TTL switches from fixed 7 days to per-row toIntervalDay(events_ttl_days); 

ttl_only_drop_parts disabled for this table since parts now mix expiry times. Missing/zero TTL falls back to the historical 7-day default.

### DIFF
--- a/packages/clickhouse/migrations/20260702120000_add_sandbox_events_ttl_days.sql
+++ b/packages/clickhouse/migrations/20260702120000_add_sandbox_events_ttl_days.sql
@@ -1,0 +1,34 @@
+-- +goose Up
+-- Retention of the event in days
+ALTER TABLE sandbox_events_local
+    ADD COLUMN IF NOT EXISTS events_ttl_days Int64 DEFAULT 7 CODEC (ZSTD(1));
+
+ALTER TABLE sandbox_events
+    ADD COLUMN IF NOT EXISTS events_ttl_days Int64 DEFAULT 7;
+
+-- Switch the fixed 7-day table TTL to a per-row TTL driven by the team's
+-- retention. Existing rows use the column DEFAULT (7), so nothing expires
+-- earlier than before.
+ALTER TABLE sandbox_events_local
+    MODIFY TTL toDateTime(timestamp) + toIntervalDay(events_ttl_days);
+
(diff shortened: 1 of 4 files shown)


packages/clickhouse/migrations/20260702120000_add_sandbox_events_ttl_days.sql
```diff
@@ -0,0 +1,32 @@
...
+-- Retention of the event in days
+ALTER TABLE sandbox_events_local
+    ADD COLUMN IF NOT EXISTS events_ttl_days Int64 DEFAULT 7 CODEC (ZSTD(1));
+
+ALTER TABLE sandbox_events
+    ADD COLUMN IF NOT EXISTS events_ttl_days Int64 DEFAULT 7;
+
+-- Switch the fixed 7-day table TTL to a per-row TTL driven by the team's retention
+ALTER TABLE sandbox_events_local
+    MODIFY TTL toDateTime(timestamp) + toIntervalDay(events_ttl_days);
```
jakubno, line 11:
```suggestion
    MODIFY TTL toDateTime(timestamp) + toIntervalDay(if(events_ttl_days <= 0, 7, if(events_ttl_days > 3650, 3650, events_ttl_days)));
```

packages/clickhouse/pkg/events/delivery.go
```diff
@@ -109,6 +111,11 @@ func (c *ClickhouseDelivery) Publish(_ context.Context, _ string, event events.S
 
 	eventData := string(eventDataJson)
 
+	ttlDays := event.EventsTTLDays
+	if ttlDays <= 0 {
+		ttlDays = events.DefaultEventsTTLDays
+	}
+
```
arkamar, line 121:
```suggestion
	ttlDays := event.EventsTTLDays
	if ttlDays <= 0 {
		ttlDays = events.DefaultEventsTTLDays
	}
	if ttlDays > 365 {
		ttlDays = 365
	}

```
With respect to gemini review. I don't think we need to do the SQL-side guard, but maybe we should cap it here?

arkamar (approved):

PR #3184, opened 2026-07-02
feat(api): filter snapshots by name

mishushakov:
Adds an optional `name` query parameter to `GET /snapshots`, mirroring the metadata filter on sandbox listing (`{ query: { name: 'some name' } }`). The supplied name is parsed with the shared `id` helpers (namespace validated against the requesting team) and resolved to a template ID via `templateCache.ResolveAlias` — the same resolution path used by sandbox create and snapshot create — then applied as a plain `env_id` filter in the list query, scoped to the team. A caller may pass the bare alias (`my-snapshot`), the namespaced form (`my-team/my-snapshot`), or the raw snapshot ID returned by the API; a `:tag` suffix filters to snapshots that have a successful build with that tag. Regenerated the OpenAPI (`api.gen.go`) and sqlc (`list_team_snapshot_templates.sql.go`) code accordingly.


packages/db/queries/snapshots/list_team_snapshot_templates.sql
ben-fornefeld, line 1: have you checked the query plans on this?

mishushakov: let Fable make a benchmark

Screenshot 2026-07-02 at 18 57 40

matthewlouisbrockman (approved):
looks good, u tested the query time.

djeebus (approved):

PR #3194, opened 2026-07-03
fix(orchestrator): enforce sandbox TTL on the node via WaitForExit

### LINKED ISSUE #3193: orchestrator: node has no independent sandbox timeout enforcement — VMs leak indefinitely if API evictor stops
## Summary

The orchestrator node has no independent mechanism to kill sandboxes when their TTL expires. A `WaitForExit()` function exists that would implement this, but **it is never called anywhere in the codebase**. As a result, the API-layer evictor is the *sole* mechanism that enforces sandbox timeouts — making it a single point of failure.

### 1. `WaitForExit` is dead code

`packages/orchestrator/pkg/sandbox/sandbox.go:1719` defines:

```go
func (s *Sandbox) WaitForExit(ctx context.Context) error {
    timeout := time.Until(s.GetEndAt())
    select {
    case <-time.After(timeout):
        return errors.New("waiting for exit took too long")
    case <-ctx.Done():
        return nil
    case <-s.exit.Done():
        ...
    }
}
```

A grep over the entire repository confirms **zero call sites** — this function is never invoked.

### 2. The lifecycle goroutine only waits for Firecracker to exit naturally

`setupSandboxLifecycle` (`packages/orchestrator/pkg/server/sandboxes.go:1039`) runs:

```go
go func() {
    ctx, _ = tracer.Start(context.WithoutCancel(ctx), ...)

    waitErr := sbx.Wait(ctx)  // blocks on exit.WaitWithContext — only unblocks when FC exits
    ...
    sbx.Close(ctx)
}()
```

`sbx.Wait()` blocks until the Firecracker process exits or `Stop()` is called from outside. There is no timeout race. The node's in-memory `endAt` field is updated via gRPC `Update`, but no goroutine on the node ever reads it to trigger a kill.

### 3. The only timeout-kill path is the API evictor

`packages/api/internal/orchestrator/evictor/evict.go` polls Redis every **50 ms**:

```go
ticker := time.NewTicker(50 * time.Millisecond)
// ...
sbxs, _ := e.store.ExpiredItems(ctx)  // ZRANGEBYSCORE -inf now
for _, item := range sbxs {
    go e.evictSandbox(ctx, item)  // gRPC Remove → node.Stop()
}
```

This is the *only* code path that kills an expired sandbox. The node itself does nothing.

## Failure Scenario

1. Sandbox created with `endAt = now + 5m`
2. API service pod restarts (rolling deploy, OOM, crash)
3. Evictor goroutine stops — expired sandboxes are no longer discovered
4. `sbx.Wait()` in the lifecycle goroutine blocks forever (no timeout)
5. Firecracker VM continues running on the node indefinitely
6. Every node accumulates zombie VMs; CPU/memory leak unbounded until node reboot

This also means **any transient API outage (even seconds) can cause VMs to outlive their TTL** — the evictor catches up only after the API recovers and the next 50 ms tick fires.

## Evidence

```
$ grep -rn "WaitForExit" packages/ --include="*.go"
packages/orchestrator/pkg/sandbox/sandbox.go:1719:func (s *Sandbox) WaitForExit(ctx context.Context) error {
```

One result — the definition. No callers.

## Proposed Fix

Wire `WaitForExit` into `setupSandboxLifecycle` so the node enforces timeouts independently:

```go
func (s *Server) setupSandboxLifecycle(ctx context.Context, sbx *sandbox.Sandbox) {
    go func() {
        ctx, childSpan := tracer.Start(context.WithoutCancel(ctx), "stop sandbox-lifecycle", trace.WithNewRoot())
        defer childSpan.End()

        waitErr := sbx.WaitForExit(ctx)
        if waitErr != nil {
            sbxlogger.I(sbx).Error(ctx, "sandbox lifecycle ended", zap.Error(waitErr))
            // If we timed out, the sandbox is still running — stop it.
            if stopErr := sbx.Stop(ctx); stopErr != nil {
                sbxlogger.I(sbx).Error(ctx, "failed to stop sandbox after timeout", zap.Error(stopErr))
            }
        }

        if cleanupErr := sbx.Close(ctx); cleanupErr != nil {
            sbxlogger.I(sbx).Error(ctx, "failed to cleanup sandbox", zap.Error(cleanupErr))
        }

        if closeErr := s.proxy.RemoveFromPool(sbx.LifecycleID); closeErr != nil {
            sbxlogger.I(sbx).Warn(ctx, "errors when manually closing connections to sandbox", zap.Error(closeErr))
        }

        sbxlogger.E(sbx).Info(ctx, "Sandbox stopped")
    }()
}
```

Note: `WaitForExit` computes the timeout once (`time.Until(GetEndAt())`). A follow-up improvement would be to make this timer resettable so that KeepAlive extensions (via gRPC `Update` → `SetEndAt`) are respected without waiting for the next API-driven eviction. For now, the node-side kill serves as a safety net against indefinite leaks.

## Impact

- **Severity**: High — resource leak, potential node exhaustion
- **Trigger**: Any API service restart or sustained outage
- **Components**: `packages/orchestrator` (node), `packages/api` (evictor)

AdaAibaby:
## Problem

`WaitForExit` was defined in `sandbox.go` but never called anywhere in the codebase. The API-layer evictor (polling Redis every 50 ms) was the sole mechanism to kill expired sandboxes. When the API service restarts, the evictor stops and Firecracker VMs accumulate on nodes indefinitely.

Closes #3193

### `packages/orchestrator/pkg/server/sandboxes.go`

Replace `sbx.Wait` with `sbx.WaitForExit` in `setupSandboxLifecycle`. The lifecycle goroutine now races against the sandbox TTL. On timeout, `Stop()` is called explicitly before the existing `Close`/cleanup path.

### `packages/orchestrator/pkg/sandbox/sandbox.go`

Replace the one-shot `time.After` in `WaitForExit` with a `time.NewTimer` loop that re-checks `endAt` after each fire. A KeepAlive that calls `SetEndAt` mid-flight now correctly resets the node-side deadline.

### `packages/orchestrator/pkg/sandbox/wait_for_exit_test.go` (new)

5 unit tests covering all branches of `WaitForExit`:
- `TestWaitForExit_AlreadyExpired` — past endAt returns error immediately
- `TestWaitForExit_ExitBeforeTTL` — clean exit before TTL returns nil
- `TestWaitForExit_ExitWithError` — FC error is wrapped and returned
- `TestWaitForExit_KeepAliveExtendsTTL` — `SetEndAt` extension resets the timer; sandbox is not killed at original deadline
- `TestWaitForExit_ContextCancelled` — ctx cancel returns nil

All pass with `-race`.

## Behaviour after this fix

| Scenario | Before | After |
|---|---|---|
| API service restarts | VMs run until node reboot | Node kills VM at `endAt` independently |
| KeepAlive extends TTL | Node ignores extension (timer not reset) | Node re-arms timer on each fire |
| Normal sandbox exit | FC exits → cleanup runs | Unchanged |
| API evictor fires before `endAt` | Kills normally | Unchanged (Stop is idempotent) |

## Testing

```
cd packages/orchestrator && go test -run TestWaitForExit -v -race -count=1 ./pkg/sandbox/
```

### DIFF
--- a/packages/orchestrator/pkg/sandbox/sandbox.go
+++ b/packages/orchestrator/pkg/sandbox/sandbox.go
@@ -1720,20 +1720,33 @@ func (s *Sandbox) WaitForExit(ctx context.Context) error {
 	ctx, span := tracer.Start(ctx, "sandbox-wait-for-exit")
 	defer span.End()
 
-	timeout := time.Until(s.GetEndAt())
-
-	select {
-	case <-time.After(timeout):
-		return errors.New("waiting for exit took too long")
-	case <-ctx.Done():
-		return nil
-	case <-s.exit.Done():
-		err := s.exit.Error()
-		if err == nil {
-			return nil
+	for {
+		endAt := s.GetEndAt()
+		timeout := time.Until(endAt)
+		if timeout <= 0 {
+			return errors.New("sandbox TTL expired")
 		}
 
-		return fmt.Errorf("fc process exited prematurely: %w", err)
+		timer := time.NewTimer(timeout)
+		select {
+		case <-timer.C:
+			// Re-check in case SetEndAt extended the deadline (e.g. via KeepAlive).
+			if s.GetEndAt().After(endAt) {
+				continue
+			}
+			return errors.New("waiting for exit took too long")
+		case <-ctx.Done():
+			timer.Stop()
+			return nil
+		case <-s.exit.Done():
+			timer.Stop()
+			err := s.exit.Error()
+			if err == nil {
+				return nil
+			}
+
+			return fmt.Errorf("fc process exited prematurely: %w", err)
+		}
 	}
 }
 
--- a/packages/orchestrator/pkg/sandbox/wait_for_exit_test.go
+++ b/packages/orchestrator/pkg/sandbox/wait_for_exit_test.go
@@ -1,0 +1,137 @@
+//go:build linux
+
+package sandbox
+
+import (
+	"context"
+	"errors"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/e2b-dev/infra/packages/shared/pkg/utils"
+)
+
+func newTestSandboxForExit(endAt time.Time) *Sandbox {
+	sbx := &Sandbox{
+		Metadata: &Metadata{},
+		exit:     utils.NewErrorOnce(),
+	}
+	sbx.SetEndAt(endAt)
+
+	return sbx
+}
+
+// TestWaitForExit_AlreadyExpired verifies that WaitForExit returns an error
+// immediately when endAt is already in the past.
+func TestWaitForExit_AlreadyExpired(t *testing.T) {
+	t.Parallel()
+
+	sbx := newTestSandboxForExit(time.Now().Add(-1 * time.Second))
+
+	err := sbx.WaitForExit(context.Background())
+
+	require.Error(t, err)
+}
+
+// TestWaitForExit_ExitBeforeTTL verifies that WaitForExit returns nil when
+// the sandbox exits cleanly before its TTL expires.
+func TestWaitForExit_ExitBeforeTTL(t *testing.T) {
+	t.Parallel()
+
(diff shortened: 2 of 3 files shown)


jakubno:
We want to keep the sandbox TTL handling on API level only

PR #3199, opened 2026-07-03
[EN-1048] fix(api): heal sandbox missing from eviction index

jakubno:
- Adds executionID to global eviction index to prevent TOCTOU
- Adds heal routine, which adds sandbox to global eviction index, if out of sync


packages/api/internal/sandbox/storage/redis/utils.go
```diff
@@ -25,14 +28,50 @@ var (
...
+
+	if i := strings.LastIndexByte(rest, ':'); i > 0 {
+		if _, err := uuid.Parse(rest[i+1:]); err == nil {
+			return teamID, rest[:i], rest[i+1:], true
+		}
+	}
 
-	return
+	return teamID, rest, "", true
 }
```
arkamar, line 92: This reads very weird. Function comment claims that the `sandboxID` never contains `:`, but it can return one with `:` if the UUID check fails.

Why not to write it with `strings.Split()`, which is more simple to read:
```suggestion
// parseExpirationMember parses both member formats:
// "teamID:sandboxID:executionID" (current) and "teamID:sandboxID" (legacy).
// executionID == "" marks a legacy member. Sandbox IDs never contain ':';
// execution IDs are always UUIDs, so the tail is an execution tag iff it
// parses as one.
func parseExpirationMember(member string) (teamID, sandboxID, executionID string, ok bool) {
	parts := strings.Split(member, ":")
	switch len(parts) {
	case 2: // legacy
		return parts[0], parts[1], "", true
	case 3:
		if _, err := uuid.Parse(parts[2]); err != nil {
			return "", "", "", false
		}
		return parts[0], parts[1], parts[2], true
	default:
		return "", "", "", false
	}
}
```

packages/api/internal/sandbox/storage/redis/heal.go
```diff
@@ -0,0 +1,162 @@
...
+	if err != nil {
+		return 0, fmt.Errorf("failed to read team index: %w", err)
+	}
+	if len(sandboxIDs) == 0 {
+		return 0, nil
+	}
+
+	// Per-team MGET: all keys share the team hash tag (cluster slot safe).
+	keys := utils.Map(sandboxIDs, func(id string) string { return getSandboxKey(teamID, id) })
+	vals, err := s.redisClient.MGet(ctx, keys...).Result()
```
arkamar, line 137: `SMEMBERS` and `MGET` might return large replies for teams with many sandboxes. I am worried if the unbounded calls could cause troubles in such a cases.

packages/api/internal/sandbox/storage/redis/items.go
```diff
@@ -105,10 +119,39 @@ func (s *Storage) ExpiredItems(ctx context.Context) ([]sandboxtypes.Sandbox, err
...
 				logger.L().Debug(ctx, "ExpiredItems: Sandbox marked as expried in index, but state say otherwise", logger.WithSandboxID(sbx.SandboxID), logger.Time("end_time", sbx.EndTime))
 
+				// Re-score the drifted member to the stored EndTime so it
+				// stops occupying the expired scan window on every tick.
+				// XX: never resurrect a member a concurrent Remove deleted.
+				if ref.executionID != "" {
+					rescores = append(rescores, redis.Z{
+						Score:  float64(sbx.EndTime.UnixMilli()),
+						Member: ref.member,
+					})
```
arkamar, line 152: It seems there's a small race between the `MGET` and `ZADD XX`, a concurrent Update that shortens the EndTime in that window gets clobbered back to the stale later score, delaying eviction until it passes. It seems to be rare and bounded, so probably fine to accept.

arkamar (changes requested):
Not necessarily request for changes, it's rather a request for discussion.

packages/api/internal/sandbox/storage/redis/items.go
jakubno: Yes, the race is there, but to trigger it we would need to have a drifted score (itself an error path) and a concurrent Update inside a single ExpiredItems pass. Double-rare. 

I think it's okay to accept it here as working around it would make this more complex

Also the sandbox is never removed early as there an extra check to prevent that

packages/api/internal/sandbox/storage/redis/heal.go
jakubno: Good point, thanks for noticing this!

packages/api/internal/sandbox/storage/redis/main.go
```diff
@@ -51,23 +95,32 @@ func NewStorage(
...
+		metrics:     metrics,
 	}, nil
 }
 
 // Start subscribes to the global PubSub channel and launches the publish
-// worker. Blocks until the context is cancelled or Close is called.
+// worker and the expiration index healer. Blocks until the context is
+// cancelled or Close is called.
 func (s *Storage) Start(ctx context.Context) {
 	pubDone := make(chan struct{})
```
arkamar, line 123: I think the condex is technically right and the solution is simple:
```suggestion
	ctx, cancel := context.WithCancel(ctx)
	defer cancel()

	pubDone := make(chan struct{})
```

arkamar (approved):
I think the `cancel()` for `startHealer()` makes sense. Otherwise LGTM.

PR #3209, opened 2026-07-06
feat: make GCS DirectPath opt-in via GCS_ENABLE_DIRECT_PATH

ValentaTomas:
#2149 hardcoded `EnableDirectPath(false)` when it cleaned up the gRPC env vars, so today the flag can't be enabled at all. This restores the env knob from #2133 but with **default off** — no behavior change for the orchestrator (which is where DirectPath performed badly).

The storage compactor's Cloud Batch workers are pure bulk readers where Direct Connectivity pays off (LB bypass, higher per-connection throughput at PB-scale reads); they'll set `GCS_ENABLE_DIRECT_PATH=true` per submit. The client library already pairs its DirectPath default with `AllowNonDefaultServiceAccount(true)`, which our override doesn't touch, so the flag works for non-default SAs.

### DIFF
--- a/packages/shared/pkg/storage/storage_google.go
+++ b/packages/shared/pkg/storage/storage_google.go
@@ -42,7 +42,7 @@ const (
 	googleBackoffMultiplier        = 2
 	googleMaxAttempts              = 10
 	defaultGRPCConnectionPoolSize  = 8
-	defaultGCSEnableDirectPath     = false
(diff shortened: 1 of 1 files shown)


ValentaTomas:
Closing for now — deliberately keeping DirectPath hardcoded off until campaign numbers show the transport is actually a bottleneck.

PR #3214, opened 2026-07-07
Paginate S3 prefix deletes

tomassrnka:
DeleteObjectsWithPrefix only deleted the first ListObjectsV2 page (1000 keys max), so prefixes with more than 1000 objects left the rest behind. Now it pages through all of them. Also refuses an empty prefix so a bad call can't wipe the whole bucket.


tomassrnka:
bugbot run

tvi (approved):

PR #3216, opened 2026-07-07
[EN-1603] fix(api): prevent uint64 underflow in node allocated metrics

jakubno:
`OptimisticRemove` subtracted from unsigned counters without a guard. When the periodic `ServiceInfo` poll overwrote the local counter after a sandbox already left the orchestrator's map, the subsequent remove underflowed

### DIFF
--- a/packages/api/internal/orchestrator/nodemanager/node.go
+++ b/packages/api/internal/orchestrator/nodemanager/node.go
@@ -215,7 +215,6 @@ func (n *Node) OptimisticRemove(ctx context.Context, res SandboxResources) {
 	n.metricsMu.Lock()
 	defer n.metricsMu.Unlock()
 
-	// Directly subtract from the current metrics view.
 	cpu := uint32(res.CPUs)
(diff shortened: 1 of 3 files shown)


packages/api/internal/orchestrator/nodemanager/node.go
```diff
@@ -215,7 +215,17 @@ func (n *Node) OptimisticRemove(ctx context.Context, res SandboxResources) {
 	n.metricsMu.Lock()
 	defer n.metricsMu.Unlock()
 
-	// Directly subtract from the current metrics view
-	n.metrics.CpuAllocated -= uint32(res.CPUs)
-	n.metrics.MemoryAllocatedBytes -= uint64(res.MiBMemory) * 1024 * 1024
+	// Directly subtract from the current metrics view.
```
arkamar: nit: this comment no longer fits here, it should be moved to the subtraction place or removed.
```suggestion
```

arkamar (approved):

PR #3225, opened 2026-07-08
chore(api): remove legacy expiration member migration code [EN-1602]

jakubno:
Drops the legacy two-part member format: dual ZREM in Remove, migrate-on-write in Update, lazy upgrade in ExpiredItems, legacy half of healer ZMSCORE check, and the legacy_upgraded sweep metric.

### DIFF
--- a/packages/api/internal/sandbox/storage/redis/heal.go
+++ b/packages/api/internal/sandbox/storage/redis/heal.go
@@ -55,7 +55,6 @@ func (s *Storage) startHealer(ctx context.Context) {
 	}
 }
 
-// TODO: remove after migration
 // healerEnabled reports whether the heal pass should run. A nil feature flag
 // client (tests) falls back to the flag's default.
 func (s *Storage) healerEnabled(ctx context.Context) bool {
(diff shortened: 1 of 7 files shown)


packages/api/internal/sandbox/storage/redis/heal.go
```diff
@@ -55,7 +55,6 @@ func (s *Storage) startHealer(ctx context.Context) {
 	}
 }
 
-// TODO: remove after migration
```
arkamar, line 58: Are we going to keep the `expiration-index-healer` flag for unseeable future?

arkamar (approved):

packages/api/internal/sandbox/storage/redis/heal.go
jakubno: Few days, then I would remove it, but it isn't related to migrationn

jakubno: [EN-1653: Remove `expiration-index-healer` flag](https://linear.app/e2b/issue/EN-1653/remove-expiration-index-healer-flag)

PR #3235, opened 2026-07-08
ci: speed up integration tests

dobrac:
## Why

PR wall time for integration tests was ~21.5 min, gated by the slowest job of a 3-config compression matrix where every job rebuilds everything and runs the full suite. Target: **under 10 minutes**.

Baseline (`28977209000`, zstd1 = long pole): Build Packages 2m26s, host init + template + services ~4m20s, tests 14m07s, total **21m24s**.

Parallelism experiments on this PR showed each host saturates server-side (Firecracker boots + zstd compression) right at `go test -parallel=4`: at 8, per-test times inflated 70–140% for ~1 min net gain and 10 template builds flaked; at 12, envd inits starved (`syncing took too long`) and 87 tests failed. Client-side concurrency can't cut wall time — more hosts can.

## What

**Sharding for wall time (the main change)**
- PRs run the production-like zstd1 config split into **four shards on parallel runners**, sized by measured test time (`TEST_SHARD` in `tests/integration/Makefile`):
  - `templates-builds-1` / `templates-builds-2`: the `^TestTemplateBuild*` tests (real FC builds, ~1400s of summed test time that barely parallelizes on one host), split into balanced halves by name prefix — new tests fall into shard 2 automatically.
  - `sandboxes`: `api/sandboxes` + the light packages (proxies, metrics, volumes).
  - `rest`: everything else via `go list` (auto-picks-up new packages), including the non-build template tests.
- Push to main keeps the full 3-config matrix running the whole suite per config (`full-matrix` input).

**Setup cut from ~7 min to ~4.5 min per job**
- Package builds run concurrently instead of serially, with a dedicated salted Go cache (setup-go keys purely on dep files, so it kept restoring stale `-N -l` artifacts and recompiling everything; after merge, push-main runs will keep the salted cache warm for PRs).
- ClickHouse migrations run goose directly on the host (`migrate-host`, pinned via go.mod tool directive) — building the migrator image cost ~2 min/job even with gha layer caching, for a container whose migrations are volume-mounted anyway.
- postgres/clickhouse/redis start in the background right after checkout (new `start-databases` action), overlapping the Go build; `start-services` waits with bounded health loops (previously infinite) on pinned images.
- `-gcflags=all=\"-N -l\"` is now opt-out (`DEBUG_GCFLAGS`, empty in CI): race detection stays on, optimizations stay on.

**Less test work**
- New `RequestTemplateWithoutBuild` helper: 7 tests that only exercise template metadata/authz no longer build full FC templates (~60–160s each). Verified none of these API paths filter on build status.
- envd process tests: fixed 10s/5s sleeps replaced with `EventuallyWithT` polling.

**Learned the hard way**
- The template build cannot overlap a running orchestrator: `create-build` manages its own sandbox network state and the two conflict, wedging the provisioning VM (reverted after one red run).

## Measured result (run `29041851323`)

| Shard | Tests step | Total job |
|---|---|---|
| zstd1-sandboxes | 2m37s | 5.7 min |
| zstd1-templates-1 | 3m20s | 6.4 min |
| zstd1-templates-2 | 4m12s | 6.8 min |
| zstd1-rest | 4m01s | **7.2 min** |

**PR wall time: 21.4 min → 7.2 min (−66%)**, runner usage ~50 → ~26 min. Verified across the shard junits: all 219 top-level tests run exactly once (none missing, none duplicated), 0 failures.

## Not changed (checked and rejected)
- Raising `-parallel` (measurements above; knob kept for local/future tuning).
- `MAX_PARALLEL_MEMFILE_SNAPSHOTTING`: dead config, nothing reads it.
- host-init gsutil caching: the whole step is ~20s.
- Tag-assignment tests keep real builds: `POST /templates/tags` requires a ready build.

## Follow-up worth investigating
The compression path is the dominant remaining cost: the same suite runs 3x faster uncompressed than under zstd1 with `compress_workers=8`, and the host saturates at 4 concurrent FC operations. Likely production-relevant contention, worth profiling independently of CI.

### DIFF
--- a/.github/actions/build-packages/action.yml
+++ b/.github/actions/build-packages/action.yml
@@ -4,15 +4,52 @@ description: "Compiles and builds Go services."
 runs:
   using: "composite"
   steps:
-    - name: Setup Go
-      uses: ./.github/actions/go-setup-cache
+    - name: Set up Go
+      uses: actions/setup-go@v6
+      with:
+        go-version-file: go.work
+        # Caching is handled explicitly below. setup-go keys its cache purely
+        # on the dependency files, so it could never store a build cache for
+        # the optimized (DEBUG_GCFLAGS="") binaries this job builds - every
+        # run would restore stale artifacts and recompile everything.
+        cache: false
+
+    - name: Restore Go caches
+      uses: actions/cache@v5
+      with:
+        path: |
+          ~/go/pkg/mod
+          ~/.cache/go-build
+        # The "opt" salt separates this cache from setup-go caches populated
+        # by other jobs with debugger-friendly build flags.
+        key: integration-go-opt-v1-${{ runner.os }}-${{ hashFiles('go.work', 'packages/*/go.mod', 'packages/*/go.sum', 'tests/integration/go.mod', 'tests/integration/go.sum') }}
+        restore-keys: |
+          integration-go-opt-v1-${{ runner.os }}-
 
     - name: Build Packages
       run: |
-        make -C tests/integration build-debug
-        make -C packages/db build-debug
-        make -C packages/orchestrator build-debug
-        make -C packages/api build-debug
-        make -C packages/envd build-debug
-        make -C packages/client-proxy build-debug
+        # The builds are independent, so run them concurrently: the Go build
+        # cache and module cache are safe for concurrent use and shared
+        # dependencies compile only once.
+        pkgs="tests/integration packages/db packages/orchestrator packages/api packages/envd packages/client-proxy"
+
+        pids=""
+        for pkg in $pkgs; do
+          log="/tmp/build-$(basename "$pkg").log"
+          make -C "$pkg" build-debug > "$log" 2>&1 &
+          pids="$pids $!:$pkg"
+        done
+
+        status=0
+        for entry in $pids; do
+          pid="${entry%%:*}"
+          pkg="${entry#*:}"
+          if ! wait "$pid"; then
+            echo "::error::build failed for $pkg"
+            status=1
+          fi
+          echo "===== $pkg ====="
+          cat "/tmp/build-$(basename "$pkg").log"
+        done
+        exit $status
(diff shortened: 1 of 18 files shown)


dobrac:
Parallelism tuning results from this PR's CI runs (zstd1 shard):

| TEST_PARALLELISM | Result | Test step | Notes |
|---|---|---|---|
| 4 (baseline, main) | pass | 14m07s | |
| 8 | pass | 13m11s | per-test times +70-140% (summed test time 4166s -> 7595s), 10 template-build flakes rescued by `--rerun-fails=1` |
| 12 | **87 failures** | 9m17s | envd init starvation: `failed to init envd: syncing took too long` x44 in orchestrator log |

The host saturates server-side (FC VM boots + zstd compression with 8 workers per operation), so raising client-side `-parallel` past 4 only adds contention. Kept CI at 4; the `TEST_PARALLELISM` knob remains for local runs and future tuning if the compression path gets cheaper.

This strengthens the follow-up note: the compression path is the dominant cost (uncompressed shard runs the same suite in 4m51s vs 14m07s for zstd1).

dobrac:
Final result after the sharding + setup work: **PR wall time 21.4 min → 7.2 min**, under the 10-minute budget with ~3 min of headroom.

| Shard | Tests step | Total job |
|---|---|---|
| zstd1-sandboxes | 2m37s | 5.7 min |
| zstd1-templates-1 | 3m20s | 6.4 min |
| zstd1-templates-2 | 4m12s | 6.8 min |
| zstd1-rest | 4m01s | 7.2 min |

Coverage verified from the shard junits: all 219 top-level tests run exactly once across the four shards (none missing, none duplicated), 0 failures, no rerun-rescues. Steady-state should improve slightly further once push-main runs keep the salted Go cache warm for PR branches.

tomassrnka:
Adopted in #3479

PR #3245, opened 2026-07-09
fix: remove release from release please

charlie-e2b:


rguliyev (commented):
## DevOps review

The design direction is good — tag-triggered publishing is a much better shape than the deleted per-package `release_created` jobs (re-publish becomes "re-run the tag's workflow", RCs work by hand-pushing a tag, and the DRY single publish job replaces 3 near-identical ones). I verified build parity against the deleted jobs and the Dockerfiles on the branch (docker-reverse-proxy `COMMIT_SHA`+`VERSION`, client-proxy `COMMIT_SHA` only, clickhouse-migrator none with `packages/clickhouse` context — all match), the `clickhouse-migrator ↔ packages/clickhouse` mapping in both the tag step and publish.yml, the step-skip logic on unknown tags, and the commit-title parsing against real squash-merge subjects in this repo (`chore(main): release clickhouse-migrator 1.0.1 (#3227)` parses correctly). One blocker though, confirmed in release-please's source:

### 🔴 Blocker — `skip-github-release: true` never swaps `autorelease: pending` → `autorelease: tagged`, so release-please stops opening release PRs after the first merge

The label swap lives inside `createReleasesForPullRequest` ([manifest.ts#L1307-L1320](https://github.com/googleapis/release-please/blob/main/src/manifest.ts)), which only runs for PRs that produced ≥1 candidate release. With config-level `skip-github-release: true`, `strategy.buildRelease()` returns early ("Release skipped from strategy config", `strategies/base.ts`), so there are **zero** candidates and the merged release PR keeps its `autorelease: pending` label forever. On the next push, `createPullRequests()` → `findMergedReleasePullRequests()` finds it and aborts: *"There are untagged, merged release PRs outstanding - aborting"* (`manifest.ts:937`) — **no new release PR is ever created again**.

The new "Tag release" step creates the git tag but never touches the PR labels, so the first release after this merges will work and the second will silently never get a release PR.

**Fix**: after minting the tag, swap the labels on the release PR (its number is right there in the `(#N)` suffix of the merge title):
```bash
gh pr edit "${pr_number}" --remove-label "autorelease: pending" --add-label "autorelease: tagged"
```
Do it in the already-exists path too, so a re-run converges. (Alternative worth a thought: keep GitHub Releases and let release-please do tags+labels itself — since the action already runs with the App token, its tags *would* trigger `publish.yml`. If the goal is only "no Releases page", the label fix keeps your approach intact.)

### 🟡 Major — only the push head is inspected; coalesced release merges are silently dropped

With `separate-pull-requests: true`, two release PRs merged in quick succession (or any batch push landing >1 merge commit) means only the head commit's component gets tagged/published — the comment documents manual-tag recovery, but this is a foreseeable multi-package scenario with no detection. Cheap hardening: scan the push's commits (`git log --format=%s ${{ github.event.before }}..${GITHUB_SHA}` or `github.event.commits`) instead of only `-1`, or at least emit a `::warning::` when a non-head commit in the push matches the release pattern.

### 🟢 Minor / nits

- `workflow_dispatch` with a nonexistent tag fails at the checkout step with a confusing git error; a cheap `gh api .../git/ref/tags/${TAG}` existence check in the resolve step would fail it with a clear message.
- The squash-title dependence is called out in comments (👍), but it also silently relies on the repo's squash-title setting staying "PR title"-shaped; for a single-commit release PR GitHub's default squash subject is the *commit's* message. It happens to match today because release-please's branch commit message equals the PR title — worth one line in the comment so nobody "fixes" the repo merge settings out from under this.
- The recovery/RC procedures live only in YAML comments; a short note in the contributor docs would help whoever hits this at 2am.
- PR body is empty and `updated to fix possible publish on merge main` isn't a great commit message for archaeology on a change this behavioral.

### 👍 Verified good

- Commit-title parsing is injection-safe: component is validated against a hardcoded `case` allowlist, version comes from the manifest via `jq` (not the attacker-influenceable title, which is only used in an equality check), and `gh api -f` passes literals — worst case is a clean `exit 1`.
- The manifest-vs-title version cross-check is a nice guard against stale release PRs merged out of order.
- App-token gating is right: `GITHUB_TOKEN`-pushed tags wouldn't trigger `on: push`, and the step hard-fails with an actionable message instead of minting a dead tag.
- Changelog continuity with lightweight tags checks out: release-please's release lookup falls back to listing repo tags, and `manifest.ts` even special-cases `skipGithubRelease` there — no GitHub Release object needed (conditional on the label blocker being fixed).
- Never re-pushing tags + no `:latest` is exactly the right shape for the `immutableTags=true` e2b-artifacts repos. One consequence to be aware of: if a version's image is ever *bad* (published but broken), the immutable tag means recovery is a new patch release, not a re-publish — which is arguably a feature.

djeebus (approved):

PR #3247, opened 2026-07-10
fix(dashboard-api): return user_id from /admin/users/bootstrap and formalize response schema

AdaAibaby:
## Problem

The `/admin/users/bootstrap` endpoint omits `user_id` from its declared OpenAPI response (`TeamResolveResponse` only has `id` and `slug`). The Next.js dashboard reads this field to populate the session cookie; when it is absent the session is created without a `userId`, `getAuthContext()` returns null on every request, and the user is redirected back to `/sign-in` in a loop — visible as `ERR_TOO_MANY_REDIRECTS` in incognito/fresh sessions.

### `spec/openapi-dashboard.yml`
- Add `AdminUserBootstrapResponse` schema with required fields `id`, `slug`, and `user_id`
- Point the `/admin/users/bootstrap` 200 response at the new schema instead of `TeamResolveResponse`
  (`TeamResolveResponse` itself is unchanged to avoid affecting `/admin/teams/bootstrap` and `/teams/resolve`)

### `packages/dashboard-api/internal/api/api.gen.go`
- Regenerated via `go generate ./internal/api/` to include the new `AdminUserBootstrapResponse` struct

### `packages/dashboard-api/internal/handlers/admin_users_bootstrap.go`
- Use the typed `api.AdminUserBootstrapResponse{Id, Slug, UserId}` instead of an untyped `gin.H` map

### `packages/db/migrations/20260708120000_add_teams_email_lower_index.sql`
- Add `-- +goose NO TRANSACTION` directive so `CREATE INDEX CONCURRENTLY` does not fail when the migration test harness wraps migrations in a transaction block

## Test plan

- [ ] `TestPostAdminUsersBootstrap_ResponseIncludesUserID` passes
- [ ] `go build ./...` in `packages/dashboard-api` succeeds
- [ ] Incognito sign-in flow completes without redirect loop on dev


AdaAibaby:
Fixed in e289eb47. Three issues addressed:

**Gemini high —  compile error**: Not a bug.  was added in Go 1.25; this repo runs Go 1.26.3.

**Gemini/Codex P2 —  LIMIT 1 ambiguity**: Replaced with  (returns all matches, no LIMIT). Callers now require exactly one result:
-  DB fallback: 0 → 404, 1 → proceed, 2+ → 409 Conflict  
-  email-match migration: skips if 0 or 2+ matches (treats as no pre-Ory account found)

**Codex P1 — SSO team bypass in DB fallback**: Fixed.  DB fallback now fails closed when  — returns 503 instead of bypassing the org membership check.

PR #3249, opened 2026-07-10
[EN-1673] feat(orchestrator): ensure requested free disk before finalize

gm-e2b:
[EN-1673] feat(orchestrator): ensure requested free disk before finalize
    
Template build steps can consume the free space allocated by the base
builder, leaving newly built templates below the DiskSizeMB target.
Resizing after finalize would make the rootfs inconsistent with its
Firecracker memory snapshot, while layered rootfs artifacts cannot use
the existing local-file resize path directly.

Add a feature-gated ensure-free-disk phase after user steps and before
finalize. Hash the layer from the parent hash, the ensure-free-disk key,
and the requested disk size. Reuse cached ensure artifacts for ordinary
cached builds, while fresh and forced builds bypass lookup, execute the
phase, and refresh the cache mapping through the normal upload pipeline.

Measure ext4 through a detached host-side NBD overlay. When the target is
already satisfied, emit a header-and-metadata-only filesystem layer that
preserves the source mappings without uploading a rootfs data body. When
growth is required, create a larger sparse overlay, zero-fill its tail,
run e2fsck, resize2fs, and e2fsck again, then export only the changed
extents. Compose and validate resized headers without mutating the source
header.

Keep the resize NBD device scoped to a single lifecycle: open it for the
filesystem operations, flush it, and close it exactly once before
exporting the backing cache. Move block-device flushing onto
DirectPathMount so template growth and sandbox rootfs cleanup share the
same BLKFLSBUF implementation.

Publish layer metadata only after all snapshot artifacts are uploaded,
ensuring cache lookups cannot observe incomplete artifacts. Finalize then
cold-boots from the filesystem-only ensure layer, keeping the disk and
memory snapshot coherent.
    
DiskSizeMB remains a best-effort free-space target because ext4 metadata
and finalize writes may consume part of the resulting capacity. Also
preserve correct e2fsck bitmask handling and cancellation-safe NBD
partial-open cleanup.
    
Part of: EN-1673
Related: EN-896

### DIFF
--- a/packages/orchestrator/pkg/sandbox/nbd/path_direct.go
+++ b/packages/orchestrator/pkg/sandbox/nbd/path_direct.go
@@ -18,6 +18,7 @@ import (
 	"github.com/Merovius/nbd/nbdnl"
 	"go.opentelemetry.io/otel"
 	"go.uber.org/zap"
+	"golang.org/x/sys/unix"
 
 	"github.com/e2b-dev/infra/packages/orchestrator/pkg/sandbox/block"
 	"github.com/e2b-dev/infra/packages/shared/pkg/featureflags"
@@ -238,6 +239,33 @@ func (d *DirectPathMount) Open(ctx context.Context) (retDeviceIndex uint32, err 
 	telemetry.ReportEvent(ctx, "connected to NBD")
 
 	return deviceIndex, nil
+}
+
+// Flush writes all pending data through the NBD connection and then clears the
+// kernel's block-device buffers. Call this before reading or exporting the
+// backend directly so it cannot observe writes that are still cached by Linux.
+func (d *DirectPathMount) Flush(ctx context.Context) error {
+	ctx, span := tracer.Start(ctx, "direct-path-mount-flush")
+	defer span.End()
+
+	path := GetDevicePath(d.deviceIndex)
+	file, err := os.Open(path)
+	if err != nil {
+		return fmt.Errorf("open NBD device for flush: %w", err)
+	}
+	defer func() {
+		if err := file.Close(); err != nil {
+			logger.L().Warn(ctx, "failed to close NBD device after flush", zap.Error(err), zap.String("path", path))
+		}
+	}()
+
+	// BLKFLSBUF completes all dirty pages as NBD writes to the in-process
+	// backend and invalidates the cache; fsync would only add the unsupported NBD_CMD_FLUSH.
+	if err := unix.IoctlSetInt(int(file.Fd()), unix.BLKFLSBUF, 0); err != nil {
+		return fmt.Errorf("flush NBD device buffers: %w", err)
+	}
+
+	return nil
 }
 
 func (d *DirectPathMount) Close(ctx context.Context) error {
--- a/packages/orchestrator/pkg/template/build/layer/layer_executor.go
+++ b/packages/orchestrator/pkg/template/build/layer/layer_executor.go
@@ -261,8 +261,6 @@ func (lb *LayerExecutor) PauseAndUpload(
 	ctx, childSpan := tracer.Start(ctx, "pause-and-upload")
 	defer childSpan.End()
 
-	userLogger.Debug(ctx, fmt.Sprintf("Processing layer: %s", meta.Template.BuildID))
-
 	// snapshot is automatically cleared by the templateCache eviction
 	snapshot, err := sbx.Pause(
 		ctx,
@@ -273,8 +271,22 @@ func (lb *LayerExecutor) PauseAndUpload(
 		return fmt.Errorf("error processing vm: %w", err)
 	}
 
+	return lb.UploadSnapshot(ctx, userLogger, snapshot, hash, meta, buildOrigin)
+}
+
+// UploadSnapshot caches and uploads a snapshot produced without a running VM.
+func (lb *LayerExecutor) UploadSnapshot(
+	ctx context.Context,
+	userLogger logger.Logger,
+	snapshot *sandbox.Snapshot,
+	hash string,
+	meta metadata.Template,
+	buildOrigin storage.ObjectOrigin,
+) error {
+	userLogger.Debug(ctx, fmt.Sprintf("Adding layer to cache: %s", meta.Template.BuildID))
+
 	// Add snapshot to template cache so it can be used immediately
-	err = lb.templateCache.AddSnapshot(
+	err := lb.templateCache.AddSnapshot(
 		context.WithoutCancel(ctx),
 		meta.Template.BuildID,
 		snapshot.MemorySnapshot.DiffHeader,
@@ -291,7 +303,7 @@ func (lb *LayerExecutor) PauseAndUpload(
 	}
 
 	// Upload snapshot async, it's added to the template cache immediately
-	userLogger.Debug(ctx, fmt.Sprintf("Saving: %s", meta.Template.BuildID))
+	userLogger.Debug(ctx, fmt.Sprintf("Uploading layer: %s", meta.Template.BuildID))
 
 	objectMetadata := lb.BuildContext.Config.ObjectMetadata(buildOrigin)
 
@@ -333,6 +345,9 @@ func (lb *LayerExecutor) PauseAndUpload(
 			return fmt.Errorf("error uploading snapshot: %w", err)
 		}
 
+		// Publish the recipe hash only after every artifact has been uploaded. A
+		// cache lookup can then resolve the hash to this build ID without observing
+		// a snapshot whose header, metadata, or data is still missing.
 		if err := lb.index.SaveLayerMeta(ctx, hash, cache.LayerMetadata{
 			Template: cache.Template{
 				BuildID: meta.Template.BuildID,
--- a/packages/orchestrator/pkg/template/build/metrics/metrics.go
+++ b/packages/orchestrator/pkg/template/build/metrics/metrics.go
@@ -15,10 +15,11 @@ import (
 type Phase string
 
 const (
-	PhaseBase     Phase = "base"
-	PhaseSteps    Phase = "steps"
-	PhaseFinalize Phase = "finalize"
-	PhaseOptimize Phase = "optimize"
+	PhaseBase       Phase = "base"
+	PhaseSteps      Phase = "steps"
+	PhaseResizeDisk Phase = "resize-disk"
+	PhaseFinalize   Phase = "finalize"
+	PhaseOptimize   Phase = "optimize"
 )
 
 // BuildResultType represents the type of build result
--- a/packages/orchestrator/pkg/template/build/phases/ensurefreedisk/builder.go
+++ b/packages/orchestrator/pkg/template/build/phases/ensurefreedisk/builder.go
@@ -1,0 +1,253 @@
+//go:build linux
+
+// Package ensurefreedisk grows a quiescent rootfs after user steps and before
+// finalize cold-boots it.
+package ensurefreedisk
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+	"time"
+
+	"github.com/google/uuid"
+	"go.opentelemetry.io/otel/attribute"
+	"go.uber.org/zap"
+
+	"github.com/e2b-dev/infra/packages/orchestrator/pkg/sandbox"
+	sbxtemplate "github.com/e2b-dev/infra/packages/orchestrator/pkg/sandbox/template"
+	"github.com/e2b-dev/infra/packages/orchestrator/pkg/template/build/buildcontext"
+	"github.com/e2b-dev/infra/packages/orchestrator/pkg/template/build/layer"
+	"github.com/e2b-dev/infra/packages/orchestrator/pkg/template/build/metrics"
+	"github.com/e2b-dev/infra/packages/orchestrator/pkg/template/build/phases"
+	"github.com/e2b-dev/infra/packages/orchestrator/pkg/template/build/storage/cache"
+	"github.com/e2b-dev/infra/packages/orchestrator/pkg/template/metadata"
+	"github.com/e2b-dev/infra/packages/shared/pkg/logger"
+	"github.com/e2b-dev/infra/packages/shared/pkg/storage"
+	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
+	"github.com/e2b-dev/infra/packages/shared/pkg/units"
+)
+
+const (
+	resizeDiskName    = "resize-disk"
+	resizeDiskTimeout = time.Hour
+)
+
+type EnsureFreeDiskBuilder struct {
+	buildcontext.BuildContext
+
+	sandboxFactory *sandbox.Factory
+	layerExecutor  *layer.LayerExecutor
+	templateCache  *sbxtemplate.Cache
+	index          cache.Index
+}
+
+func New(
+	buildContext buildcontext.BuildContext,
+	sandboxFactory *sandbox.Factory,
+	layerExecutor *layer.LayerExecutor,
+	templateCache *sbxtemplate.Cache,
+	index cache.Index,
+) *EnsureFreeDiskBuilder {
+	return &EnsureFreeDiskBuilder{
+		BuildContext:   buildContext,
+		sandboxFactory: sandboxFactory,
+		layerExecutor:  layerExecutor,
+		templateCache:  templateCache,
+		index:          index,
+	}
+}
+
+func (b *EnsureFreeDiskBuilder) Prefix() string {
+	return resizeDiskName
+}
+
+func (b *EnsureFreeDiskBuilder) String(context.Context) (string, error) {
+	return "Resizing disk", nil
+}
+
+func (b *EnsureFreeDiskBuilder) Metadata() phases.PhaseMeta {
+	return phases.PhaseMeta{
+		Phase:    metrics.PhaseResizeDisk,
+		StepType: resizeDiskName,
+	}
+}
+
+func (b *EnsureFreeDiskBuilder) Hash(_ context.Context, sourceLayer phases.LayerResult) (string, error) {
+	return cache.HashKeys(
+		sourceLayer.Hash,
+		resizeDiskName,
+		strconv.FormatInt(b.Config.DiskSizeMB, 10),
+	), nil
+}
+
+func (b *EnsureFreeDiskBuilder) Layer(
+	ctx context.Context,
+	sourceLayer phases.LayerResult,
+	hash string,
+) (phases.LayerResult, error) {
+	finalMetadata := sourceLayer.Metadata
+	finalMetadata.Template = metadata.TemplateMetadata{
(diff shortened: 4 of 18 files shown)


packages/orchestrator/pkg/template/build/phases/ensurefreedisk/sizing.go
```diff
@@ -0,0 +1,65 @@
...
+		return 0, fmt.Errorf("disk target overflows: %d MiB", diskMB)
+	}
+
+	return units.MBToBytes(diskMB), nil
+}
+
+func computeGrownSize(currentSize, targetFree, freeBefore, blockSize int64) (int64, error) {
+	if currentSize <= 0 || targetFree < 0 || freeBefore >= targetFree || blockSize != int64(header.RootfsBlockSize) {
+		return 0, fmt.Errorf("invalid grow geometry: size=%d target=%d free=%d block=%d", currentSize, targetFree, freeBefore, blockSize)
+	}
```
jakubno: can we do it as separate checks so we can return nicer error?

packages/orchestrator/pkg/template/build/phases/ensurefreedisk/sizing.go
```diff
@@ -0,0 +1,65 @@
...
+	if diskMB < 0 || diskMB > math.MaxInt64>>units.MBShift {
+		return 0, fmt.Errorf("disk target overflows: %d MiB", diskMB)
+	}
+
+	return units.MBToBytes(diskMB), nil
+}
+
+func computeGrownSize(currentSize, targetFree, freeBefore, blockSize int64) (int64, error) {
+	if currentSize <= 0 || targetFree < 0 || freeBefore >= targetFree || blockSize != int64(header.RootfsBlockSize) {
+		return 0, fmt.Errorf("invalid grow geometry: size=%d target=%d free=%d block=%d", currentSize, targetFree, freeBefore, blockSize)
```
jakubno: why is `blockSize` here, when it isn't actually used?

packages/orchestrator/pkg/template/build/phases/ensurefreedisk/sizing.go
```diff
@@ -0,0 +1,65 @@
...
+		size += add
+	}
+
+	return size, nil
+}
+
+func validateSourceGeometry(sourceSize, blockSize int64, sourceHeader *header.Header) error {
+	if sourceHeader == nil || sourceHeader.Metadata == nil {
+		return errors.New("source rootfs header metadata is missing")
+	}
```
jakubno: little bit unrelated to this PR, but 
```
	blockSize := sourceRootfs.BlockSize()
```
run before this check. `BlockSize()` always access `header` and `Metadata`. Maybe we should validate header already in `NewStorage`?

packages/orchestrator/pkg/template/build/layer/layer_executor.go
```diff
@@ -333,6 +346,9 @@ func (lb *LayerExecutor) PauseAndUpload(
 			return fmt.Errorf("error uploading snapshot: %w", err)
 		}
 
+		if !publishLayerMeta {
```
jakubno: Can we add comment here 🙏🏻

packages/orchestrator/pkg/template/build/layer/layer_executor.go
```diff
@@ -273,8 +271,23 @@ func (lb *LayerExecutor) PauseAndUpload(
...
+func (lb *LayerExecutor) UploadSnapshot(
+	ctx context.Context,
+	userLogger logger.Logger,
+	snapshot *sandbox.Snapshot,
+	hash string,
+	meta metadata.Template,
+	buildOrigin storage.ObjectOrigin,
+	publishLayerMeta bool,
+) error {
+	userLogger.Debug(ctx, fmt.Sprintf("Processing layer: %s", meta.Template.BuildID))
```
jakubno: is this log correct? it's upload, why does it say processing?

packages/orchestrator/pkg/template/build/phases/ensurefreedisk/sizing.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"errors"
+	"fmt"
+	"math"
+
+	"github.com/e2b-dev/infra/packages/shared/pkg/storage/header"
+	"github.com/e2b-dev/infra/packages/shared/pkg/units"
+)
+
+func calculateTargetFree(diskMB int64) (int64, error) {
```
jakubno: i would either rename it to something like `targetFreeBytes(targetFreeMB int64)` or just inline it

packages/orchestrator/pkg/template/build/phases/ensurefreedisk/grow.go
```diff
@@ -0,0 +1,263 @@
...
+		return nil, fmt.Errorf("open NBD device: %w", err)
+	}
+
+	return &offlineDevice{mnt: mnt, path: nbd.GetDevicePath(idx)}, nil
+}
+
+func (d *offlineDevice) close(ctx context.Context) error {
+	if d == nil || d.mnt == nil {
+		return nil
+	}
```
jakubno:
```suggestion
	if d.mnt == nil {
		return nil
	}
```

packages/orchestrator/pkg/template/build/phases/ensurefreedisk/grow.go
```diff
@@ -0,0 +1,263 @@
...
+	cleanupCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), offlineCleanupTimeout)
+	defer cancel()
+
+	err := d.mnt.Close(cleanupCtx)
+	if err == nil {
+		d.mnt = nil
```
jakubno: is there a reason why we want to allow calling close mutliple times? I think this is incorrect and dangerous as we don't really control if the device was released or not. If it was we could potentially disconnect device for another build

jakubno: we actually call this function 2 times

packages/orchestrator/pkg/template/build/phases/ensurefreedisk/grow.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+	removeErr := os.Remove(f.Name())
+	if errors.Is(removeErr, os.ErrNotExist) {
+		removeErr = nil
+	}
+
+	return errors.Join(closeErr, removeErr)
+}
+
+func flushDevice(path string) error {
```
jakubno: Similar code already exists in pkg/sandbox/rootfs/nbd.go:162, combining BLKFLSBUF with rootfs.flush. Best reuse: move logic onto DirectPathMount.Flush() and call it from both paths. This also removes inconsistent ordering and duplicate fsyncs.

packages/orchestrator/pkg/template/build/phases/ensurefreedisk/grow.go
```diff
@@ -0,0 +1,263 @@
...
+		Metadata: &meta,
+		Mapping:  sourceHeader.Mapping,
+		Builds:   sourceHeader.Builds,
+	}, buildID)
+	if err != nil {
+		return nil, nil, 0, fmt.Errorf("build grown rootfs header: %w", err)
+	}
+	if err := grownHeader.Mapping.Validate(grownHeader.Metadata.Size, header.PageSize); err != nil {
+		return nil, nil, 0, fmt.Errorf("validate grown rootfs mapping: %w", err)
+	}
```
jakubno: could we move this to `ToResizedDiffHeader` function living on `DiffMetadata`? and extend `ToDiffHeader` a little. It should prevent some duplication and simplify the code

packages/orchestrator/pkg/template/build/phases/ensurefreedisk/builder.go
```diff
@@ -0,0 +1,208 @@
...
+
+	grownBuildID, err := uuid.Parse(currentLayer.Metadata.Template.BuildID)
+	if err != nil {
+		return phases.LayerResult{}, fmt.Errorf("parse grown build id: %w", err)
+	}
+
+	// Measure the source and produce a grown diff only when the target is not met.
+	rootfsDiff, grownHeader, res, err := b.growAndExport(ctx, sourceRootfs, grownBuildID, b.Config.DiskSizeMB)
+	if err != nil {
+		return phases.LayerResult{}, phases.NewPhaseBuildError(b.Metadata(), fmt.Errorf("ensure free disk space: %w", err))
```
jakubno: the error will be returned to the error, it shouldn't contain any internal error, check other places using `NewPhaseBuildError`

jakubno (changes requested):
why do we always skip cache, wouldn't it be enough for the hash to depend on free size + previous hash

(it's fine to create a separate layer for this step)

packages/orchestrator/pkg/template/build/phases/ensurefreedisk/sizing.go
gm-e2b: Done.

packages/orchestrator/pkg/template/build/phases/ensurefreedisk/sizing.go
gm-e2b: We can remove, it's duplicated from `validateSourceGeometry`. But we still to check if the the block size of NBD is was we expect to avoid miscalculating free memory from number of free block: 
https://github.com/e2b-dev/infra/blob/make-disk-free-promise-real-en-1673/packages/orchestrator/pkg/template/build/core/filesystem/ext4.go#L220

packages/orchestrator/pkg/template/build/phases/ensurefreedisk/sizing.go
gm-e2b: Done.

packages/orchestrator/pkg/template/build/layer/layer_executor.go
gm-e2b: No longer needed, as discussed offline, we decided to make `ensure-free-disk` phase always emitting an header artifact, even when no resizing happened. This simplify the caching of this layer in both cases where we have a normal build and a forced rebuild (`force = true`).

It was needed by the previous design, where ensure artifacts were uploaded for finalize but not published to the layer cache. Ensure now always produces a reusable artifact, including the no-resize case, so every caller publishes layer metadata.

packages/orchestrator/pkg/template/build/phases/ensurefreedisk/builder.go
gm-e2b: Removed `NewPhaseBuildError` and made this internal error.

packages/orchestrator/pkg/template/build/phases/ensurefreedisk/grow.go
gm-e2b: Good catch, done.

packages/orchestrator/pkg/template/build/layer/layer_executor.go
gm-e2b: Clarified this into two log records: one for cache update and another for upload to object storage.

packages/orchestrator/pkg/template/build/phases/ensurefreedisk/grow.go
gm-e2b: Done. I moved NBD flushing onto `DirectPathMount.Flush()` and use it from both the regular rootfs provider and `ensure-free-disk`. The shared method issues `BLKFLSBUF` once; Linux performs the block-device sync before invalidating its cache, so the previous duplicate `fsyncs` and inconsistent ordering are gone.

packages/orchestrator/pkg/template/build/phases/ensurefreedisk/sizing.go
gm-e2b: Refactored code to check header pointers before the calls to `Size()` and `BlockSize`.

packages/orchestrator/pkg/template/build/phases/ensurefreedisk/grow.go
gm-e2b: Yes, this is a bug, I refactored the code to ensure the `.close()` method is called only once.

packages/orchestrator/pkg/template/build/phases/ensurefreedisk/grow.go
```diff
@@ -0,0 +1,274 @@
...
+) (freeAfter int64, e error) {
+	// Expose the detached overlay as a normal host block device for e2fsprogs.
+	device, err := b.openOfflineDevice(ctx, backend)
+	if err != nil {
+		return 0, err
+	}
+	defer func() { e = errors.Join(e, device.close(ctx)) }()
+
+	// Repair before resizing, then verify the enlarged filesystem.
+	if _, err := filesystem.CheckIntegrity(ctx, device.path, true); err != nil {
```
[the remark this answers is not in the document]
gm-e2b: Known issue, will open a ticket.

packages/orchestrator/pkg/sandbox/nbd/path_direct.go
```diff
@@ -240,6 +241,33 @@ func (d *DirectPathMount) Open(ctx context.Context) (retDeviceIndex uint32, err
...
 
+// Flush writes all pending data through the NBD connection and then clears the
+// kernel's block-device buffers. Call this before reading or exporting the
+// backend directly so it cannot observe writes that are still cached by Linux.
+func (d *DirectPathMount) Flush(ctx context.Context) error {
+	ctx, span := tracer.Start(ctx, "direct-path-mount-flush")
+	defer span.End()
+
+	path := GetDevicePath(d.deviceIndex)
+	file, err := os.OpenFile(path, os.O_RDWR, 0)
```
jakubno: we used `os.Open` before, which uses `O_RDONLY`. Do we need `O_RDWR` or how did it work before?

packages/orchestrator/pkg/template/build/metrics/metrics.go
```diff
@@ -17,6 +17,7 @@ type Phase string
 const (
 	PhaseBase     Phase = "base"
 	PhaseSteps    Phase = "steps"
+	PhaseResize   Phase = "ensure-free-disk"
```
jakubno: unite the naming 🙏🏻

packages/orchestrator/pkg/template/build/phases/ensurefreedisk/builder.go
```diff
@@ -0,0 +1,252 @@
...
+}
+
+func (b *EnsureFreeDiskBuilder) Build(
+	ctx context.Context,
+	userLogger logger.Logger,
+	_ string,
+	sourceLayer phases.LayerResult,
+	currentLayer phases.LayerResult,
+) (phases.LayerResult, error) {
+	ctx, cancel := context.WithTimeout(ctx, time.Hour)
```
jakubno: can we move `time.Hour` to constant 🙏🏻

packages/orchestrator/pkg/sandbox/nbd/path_direct.go
```diff
@@ -240,6 +241,33 @@ func (d *DirectPathMount) Open(ctx context.Context) (retDeviceIndex uint32, err
...
+		return fmt.Errorf("open NBD device for flush: %w", err)
+	}
+	defer func() {
+		if err := file.Close(); err != nil {
+			logger.L().Warn(ctx, "failed to close NBD device after flush", zap.Error(err), zap.String("path", path))
+		}
+	}()
+
+	// BLKFLSBUF syncs the block device before invalidating its buffer cache, so
+	// an additional file.Sync call would only repeat the same flush.
```
jakubno:
```suggestion
	// BLKFLSBUF completes all dirty pages as NBD writes to the in-process
	// backend and invalidates the cache; fsync would only add the unsupported NBD_CMD_FLUSH.
```

packages/orchestrator/pkg/template/build/phases/ensurefreedisk/builder.go
```diff
@@ -0,0 +1,252 @@
...
+
+	layerBuildID, err := uuid.Parse(currentLayer.Metadata.Template.BuildID)
+	if err != nil {
+		return phases.LayerResult{}, fmt.Errorf("parse ensure layer build id: %w", err)
+	}
+
+	// Measure the source and produce either an empty or grown rootfs layer.
+	// The underlying error may contain internal details (device paths, syscall
+	// failures, ...), so it must not be surfaced to the user as-is via
+	// phases.NewPhaseBuildError; keep it as a plain internal error instead.
```
jakubno:
```suggestion
```

packages/orchestrator/pkg/template/build/phases/ensurefreedisk/builder.go
```diff
@@ -0,0 +1,252 @@
...
+				zap.Int64("free_before_bytes", res.freeBefore),
+				zap.Int64("free_after_bytes", res.freeAfter),
+			)
+		}
+
+		userLogger.Info(ctx, fmt.Sprintf(
+			"ensure-free-disk: free %d -> %d MiB (target %d MiB)",
+			units.BytesToMB(res.freeBefore), units.BytesToMB(res.freeAfter),
+			units.BytesToMB(res.target),
+		))
```
jakubno, line 201: as it's kinda expected to undershot little, i wouldn't prompt users to ask
```suggestion
		userLogger.Info(ctx, fmt.Sprintf(
			"ensure-free-disk: free %d -> %d MiB",
			units.BytesToMB(res.freeBefore), units.BytesToMB(res.freeAfter),
		))
```

jakubno (changes requested):
few NITs, otherwise LGTM! 🎉  
- ARCHITECTURE.md:302 should be updated

packages/orchestrator/pkg/sandbox/nbd/path_direct.go
gm-e2b: I got confused, apparently `BLKFLSBUF` also works with fd in `O_RDONLY`.

packages/orchestrator/pkg/template/build/metrics/metrics.go
gm-e2b: Done.

packages/orchestrator/pkg/template/build/phases/ensurefreedisk/builder.go
gm-e2b: Done.

packages/orchestrator/pkg/sandbox/nbd/path_direct.go
gm-e2b: Done.

packages/orchestrator/pkg/template/build/phases/ensurefreedisk/builder.go
gm-e2b: Done.

jakubno (approved):

PR #3286, opened 2026-07-14
[EN-1038] feat(db): add template disk entitlement schema

gm-e2b:
Separate the default template free-space target from the maximum total root filesystem size that a team may build.

Add nullable default_free_disk_size_mb and max_disk_size_mb columns to tiers, plus extra_max_disk_size_mb to add-ons. Keep extra_disk_mb as the add-on contribution to default free space and add constraints preventing invalid negative or inverted tier values.

Keep the existing team_limits view unchanged for current consumers and introduce team_limits_v2 with the new effective default and maximum fields. Preserve compatibility with legacy add-on writers by treating a null extra maximum as equal to extra_disk_mb.

Leave data population and runtime adoption to later rollout steps so this commit remains an expand-only schema change.

Regenerate the database models and add focused migration coverage for the new columns, the unchanged V1 view, the V2 contract, security invoker behavior, and effective tier-plus-add-on calculations.

Part of: EN-1038

### DIFF
--- a/packages/db/migrations/20260714091414_add_template_disk_entitlements.sql
+++ b/packages/db/migrations/20260714091414_add_template_disk_entitlements.sql
@@ -1,0 +1,56 @@
+-- +goose Up
+-- +goose StatementBegin
+
+ALTER TABLE "public"."tiers"
+    -- Free-rootfs target in the existing MiB convention.
+    ADD COLUMN IF NOT EXISTS "default_free_disk_size_mb" bigint,
+    -- Total logical-rootfs ceiling before active add-ons.
+    ADD COLUMN IF NOT EXISTS "max_disk_size_mb" bigint;
+
+ALTER TABLE "public"."addons"
+    -- Total-ceiling increment in the existing MiB convention.
+    ADD COLUMN IF NOT EXISTS "extra_max_disk_size_mb" bigint;
+
+-- Populate existing tiers in every environment before the effective limits are exposed.
+UPDATE "public"."tiers"
+SET "default_free_disk_size_mb" = "disk_mb",
+    "max_disk_size_mb" = "disk_mb" + 25000;
+
+-- Populate rows that predate the schema expansion in every environment. Keep the column nullable
+-- because legacy writers may still omit it during the rollout.
+UPDATE "public"."addons"
+SET "extra_max_disk_size_mb" = "extra_disk_mb"
+WHERE "extra_max_disk_size_mb" IS NULL;
+
+-- Every existing row was populated above, and every future row must supply both values.
+ALTER TABLE "public"."tiers"
+    ALTER COLUMN "default_free_disk_size_mb" SET NOT NULL,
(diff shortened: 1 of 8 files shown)


packages/db/migrations/20260714091414_add_template_disk_entitlements.sql
jakubno, line 1: can we add if not exists to the statements🙏🏻

packages/db/migrations/20260714091415_create_team_limits_v2.sql
```diff
@@ -0,0 +1,40 @@
+-- +goose Up
+-- +goose StatementBegin
+
+CREATE VIEW "public"."team_limits_v2"
```
jakubno: Do we really need limits v2? They're just new columns, so we could just replace the old view.

jakubno (changes requested):

packages/db/migrations/20260714091414_add_template_disk_entitlements.sql
gm-e2b: Added for new columns. Adding constraint checks does not support `IF NOT EXISTS`.

packages/db/migrations/20260714091415_create_team_limits_v2.sql
gm-e2b: Main reason was to ensure a safe goose migration, so concurrent reads to the `team_limits` view are not blocked by the migration's transaction. But since this is only a schema migration, I expect the transaction to be fast. So I replaced `team_limits` view, which also simplifies following code (no need for code migration to use V2 view).

packages/db/migrations/20260714091414_add_template_disk_entitlements.sql
```diff
@@ -0,0 +1,53 @@
...
+        CHECK (default_free_disk_size_mb >= 0),
+    ADD CONSTRAINT "tiers_max_disk_size_mb_check"
+        CHECK (max_disk_size_mb > 0),
+    ADD CONSTRAINT "tiers_default_free_disk_size_lte_max_check"
+        CHECK (default_free_disk_size_mb <= max_disk_size_mb);
+
+ALTER TABLE "public"."addons"
+    ADD CONSTRAINT "addons_extra_max_disk_size_mb_check"
+        CHECK (extra_max_disk_size_mb >= 0);
+
```
jakubno, line 39: `default_free_disk_size_mb` and `max_disk_size_mb` should be non null here

packages/db/sqlc.yaml
```diff
@@ -21,6 +21,15 @@ overrides:
           import: "time"
           type: "Time"
           pointer: true
+      # sqlc does not infer nullability for these view expressions.
```
jakubno: this won't be needed afterwards

jakubno (changes requested):

jakubno (approved):

PR #3290, opened 2026-07-15
feat(orchestrator): add NetworkAssignHook for sandbox lifecycle extensions

rguliyev:
Sandbox editions sometimes need bounded, local prep after a sandbox's identity/network slot are assigned but before its guest runs - a gap between the two existing extension points: EgressProxy fires before identity exists, MapSubscriber's OnInsert fires only after the guest is already running.

Add NetworkAssignHook (OnNetworkAssign), called once per start attempt right after AssignNetwork in CreateSandbox/ResumeSandbox, tagged with a NetworkAssignReason since these carry different guest-state guarantees. Called synchronously with the caller's own ctx and no derived timeout: the implementation is responsible for bounding its own duration. A panic is recovered, since there is no other recovery layer between this call and the process's main goroutine. Can never fail a sandbox start; a returned error is logged and swallowed.

No-op everywhere by default; zero behavior change until an edition wires one in.

Benchmarked on the no-op path: ~10ns/op, 0 B/op, 0 allocs/op - noise next to a sandbox create/resume (ms-s).


packages/orchestrator/pkg/sandbox/sandbox.go
```diff
@@ -373,6 +379,53 @@ func NewFactory(
...
+					logger.WithLifecycleID(sbx.LifecycleID),
+					zap.Any("panic", r))
+			}
+		}()
+		if err := f.startHook.BeforeStart(hookCtx, sbx, reason); err != nil {
+			logger.L().Warn(ctx, "sandbox start hook failed, continuing",
+				logger.WithSandboxID(sbx.Runtime.SandboxID),
+				logger.WithLifecycleID(sbx.LifecycleID),
+				zap.Error(err))
+		}
```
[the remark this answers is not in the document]
rguliyev: Fixed in 822bd935f - suppress the redundant log when BeforeStart returns context.Canceled/DeadlineExceeded, since that's already logged by the timeout/cancellation branch.

packages/orchestrator/pkg/sandbox/sandbox.go
```diff
@@ -373,6 +379,53 @@ func NewFactory(
...
+	}()
+
+	select {
+	case <-done:
+	case <-hookCtx.Done():
+		logger.L().Warn(ctx, "sandbox start hook did not complete before timeout, proceeding without waiting further",
+			logger.WithSandboxID(sbx.Runtime.SandboxID),
+			logger.WithLifecycleID(sbx.LifecycleID))
 	}
```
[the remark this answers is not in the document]
rguliyev: Fixed in 822bd935f - now checks ctx.Err() to distinguish parent-context cancellation from startHookTimeout actually elapsing, and logs accordingly.

packages/orchestrator/pkg/sandbox/start_hook.go
```diff
@@ -0,0 +1,71 @@
...
+	StartReasonResume
+	// StartReasonReboot: a cold boot from a filesystem-only snapshot; guest
+	// memory, and therefore any prior network connection state, is not
+	// preserved.
+	StartReasonReboot
+	// StartReasonThrowawayResume: a resume that will not be promoted to a
+	// live, routable sandbox (see WithoutLiveRegistration) and whose network
+	// egress is denied for its entire lifetime (see WithDenyEgress).
+	StartReasonThrowawayResume
+)
```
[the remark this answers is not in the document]
rguliyev: Fixed in 822bd935f - added StartReason.String() (create/resume/reboot/throwaway_resume/unknown), with test coverage.

packages/orchestrator/pkg/sandbox/start_hook.go
```diff
@@ -0,0 +1,88 @@
...
+	// established network connection) may be preserved across this call.
+	StartReasonResume
+	// StartReasonReboot: a cold boot from a filesystem-only snapshot; guest
+	// memory, and therefore any prior network connection state, is not
+	// preserved.
+	StartReasonReboot
+	// StartReasonThrowawayResume: a resume that will not be promoted to a
+	// live, routable sandbox (see WithoutLiveRegistration) and whose network
+	// egress is denied for its entire lifetime (see WithDenyEgress).
+	StartReasonThrowawayResume
```
wj-e2b: What does this mean?

packages/orchestrator/pkg/sandbox/start_hook.go
```diff
@@ -0,0 +1,88 @@
...
+	StartReasonUnknown StartReason = iota
+	// StartReasonCreate: a brand-new sandbox, cold-booting for the first time.
+	StartReasonCreate
+	// StartReasonResume: a memory-snapshot resume; guest state (including any
+	// established network connection) may be preserved across this call.
+	StartReasonResume
+	// StartReasonReboot: a cold boot from a filesystem-only snapshot; guest
+	// memory, and therefore any prior network connection state, is not
+	// preserved.
+	StartReasonReboot
```
wj-e2b: Cold boot would be more clear

packages/orchestrator/pkg/sandbox/sandbox.go
```diff
@@ -1014,6 +1088,12 @@ func (f *Factory) ResumeSandbox(
 		return nil
 	})
 
+	reason := StartReasonResume
+	if ropts.skipLiveRegistration {
+		reason = StartReasonThrowawayResume
+	}
+	f.runStartHook(ctx, sbx, reason)
```
wj-e2b: This is called start hook, but the usage from belt requires this to be post network setup. The name isn't clear where the hook is placed and there's no comments. A person without context reading this could assume that this hook gets fired pre-sandbox create or post-sandbox create and it's ambigious. The actual implementation is that it's post-network.

packages/orchestrator/pkg/sandbox/start_hook.go
rguliyev: Fixed in d29bd9177 - this refers to the pause-resume prefetch harvest, an internal warm-pool mechanism that briefly resumes a sandbox (network-isolated, unregistered) purely to record its resume working set into the page cache ahead of time, then discards it. Added that explanation directly to the doc comment (see ThrowawayResumeOptions) so it's not opaque anymore.

packages/orchestrator/pkg/sandbox/start_hook.go
rguliyev: I'd push back on renaming Reboot itself: this package's existing StartType enum already documents both StartTypeCreate ("cold boot (template build)") and StartTypeReboot ("cold boot from a snapshot rootfs") as varieties of cold boot, so ColdBoot wouldn't actually disambiguate this from Create, it'd collide with vocabulary already used for the other case. Reboot also matches the existing RebootSandbox factory method 1:1. I did fix the substance of the comment though (d29bd9177): it now says "a filesystem-only snapshot boot" instead of leading with "cold boot," which is the part that was actually unclear.

packages/orchestrator/pkg/sandbox/sandbox.go
rguliyev: Fixed in d29bd9177 - renamed StartHook/BeforeStart to NetworkAssignHook/OnNetworkAssign across both repos, matching the existing MapSubscriber convention (OnInsert/OnNetworkRelease are named after the exact event they fire on, not a generic verb). OnNetworkAssign is now explicitly the setup-side counterpart to OnNetworkRelease, and the name itself says where it fires (right after AssignNetwork) instead of requiring a reader to go find the doc comment. Also added explicit "do not move this call" comments at both call sites (CreateSandbox/ResumeSandbox) stating the ordering invariant.

tvi (approved):

wj-e2b (approved):

PR #3304, opened 2026-07-17
fix(clickhouse): avoid server-tz drift in sandbox/team metrics filters

Piwriw:
## Summary

- Replace `DateTime64` query parameters with `Int64` Unix-nanosecond parameters converted through `fromUnixTimestamp64Nano`.
- Apply the fix to:
  - `sandbox_metrics_gauge` queries in `packages/clickhouse/pkg/sandbox.go`
  - `team_metrics_sum` and `team_metrics_gauge` queries in `packages/clickhouse/pkg/team.go`
  - Existing sandbox and build log queries
- Centralize ClickHouse timestamp conversion in `packages/clickhouse/pkg/timestamp`, including safe clamping for dates outside the `int64` nanosecond range.

## Why

`clickhouse.DateNamed(..., clickhouse.Seconds)` serializes a UTC `time.Time` as a wall-clock string without an offset, such as `'2025-02-17 00:00:00'`.

When a `DateTime64(9)` column has no explicit timezone, ClickHouse interprets that value using the server timezone. On a non-UTC server such as `Asia/Shanghai`, the query window shifts by the timezone offset. This can produce empty results for short ranges and stale or incorrect results for longer ranges.

The issue has existed since the original metrics migrations:

- `20250717135224_sandbox_metrics.sql`
- `20250801113224_team_metrics.sql`

The sandbox logs reader introduced in #3236 already used Unix nanoseconds. This PR applies the same timezone-independent approach to the older metrics readers and moves the shared conversion into a dedicated package.

## Affected Endpoints

- `GET /sandboxes/{sandboxID}/metrics`
- `GET /teams/{teamID}/metrics`
- `GET /teams/{teamID}/metrics/max`

## Implementation

Unix nanoseconds represent an absolute instant. Passing them as `Int64` values and converting them with `fromUnixTimestamp64Nano` avoids timezone-dependent string parsing:

```sql
timestamp >= fromUnixTimestamp64Nano({start_time:Int64})
AND timestamp <= fromUnixTimestamp64Nano({end_time:Int64})
```

The shared `timestamp.UnixNano` helper also clamps far-future and far-past values before conversion, preventing `time.Time.UnixNano` overflow.

Returned timestamps and their JSON serialization are unchanged.

## Testing

- Added focused tests for timestamp conversion, timezone normalization, and `int64` range clamping.
- Verified the ClickHouse timestamp, sandbox logs, and root client packages.

No schema changes or migrations are required.

### DIFF
--- a/packages/clickhouse/pkg/sandbox.go
+++ b/packages/clickhouse/pkg/sandbox.go
@@ -24,6 +24,28 @@ type Metrics struct {
 	DiskUsed       float64   `ch:"disk_used"`
 }
 
+// maxUnixSecondsForCH is the largest Unix-second value whose nanosecond
+// representation fits in a signed int64. time.Time.UnixNano() is undefined
+// beyond ~2262-04-11; API handlers accept Unix-second query params without
+// an upper bound, so far-future sentinels (e.g. 9999999999 = year 2286)
+// must be clamped before conversion to avoid wrapping to negative int64
+// and silently breaking fromUnixTimestamp64Nano filter windows.
+const maxUnixSecondsForCH = (1<<63 - 1) / int64(time.Second)
+
+// unixNanoForCH returns t as nanoseconds since the Unix epoch, clamped to
+// the int64-representable range. Always use this instead of t.UTC().UnixNano()
+// when feeding values to ClickHouse's fromUnixTimestamp64Nano.
+func unixNanoForCH(t time.Time) int64 {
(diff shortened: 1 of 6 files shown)


Piwriw:
Could someone familiar with the ClickHouse readers take a review pass? Happy to address any feedback.

Piwriw:
@codex review

Piwriw:
@codex review

PR #3317, opened 2026-07-21
fix(docker-reverse-proxy): add HTTP client timeouts and fix unsafe body read

dongjiang1989:
## Problem

The `docker-reverse-proxy` service has several HTTP client/server safety issues:

1. **HTTP Server has no timeouts** — the `http.Server` is created without `ReadHeaderTimeout` or `IdleTimeout`, leaving it vulnerable to [Slowloris attacks](https://en.wikipedia.org/wiki/Slowloris_(computer_security)) and idle connection leaks.

2. **Token fetch uses `http.DefaultClient` with no timeout** — `getToken()` in `token.go` calls `http.DefaultClient.Do(r)` to reach GCP Artifact Registry. If the registry is unreachable, the goroutine blocks indefinitely, leading to goroutine leaks under load.

3. **Unsafe body read on error path** — `make([]byte, resp.ContentLength)` panics when `ContentLength == -1` (chunked transfer encoding), and `resp.Body.Read()` does not guarantee reading the full response. Additionally there is a redundant `defer resp.Body.Close()`.

4. **Reverse proxy uses default `http.Transport`** — `httputil.NewSingleHostReverseProxy` is created without a custom transport, meaning there are no dial, TLS, or response-header timeouts for proxied requests to GCP.

### `main.go`
- Add `ReadHeaderTimeout: 10s` to the HTTP server to mitigate Slowloris.
- Add `IdleTimeout: 120s` to reclaim idle keep-alive connections.
- Note: `ReadTimeout`/`WriteTimeout` are intentionally **not** set because the proxy streams large Docker blob payloads.

### `token.go`
- Introduce a package-level `tokenClient` (`&http.Client{Timeout: 10s}`) and replace `http.DefaultClient`. 10s is generous for a small JSON token call within GCP's internal network.
- Fix the double `defer resp.Body.Close()` — keep only the outer one.
- Replace `make([]byte, resp.ContentLength)` + `resp.Body.Read(body)` with `io.ReadAll(resp.Body)`, which correctly handles chunked encoding and guarantees a full read.

### `store.go`
- Attach a custom `http.Transport` to the reverse proxy with:
  - `DialContext` timeout: 10s (TCP connection establishment)
  - `KeepAlive`: 30s
  - `TLSHandshakeTimeout`: 10s
  - `ResponseHeaderTimeout`: 30s (only waits for headers; body transfer is uncapped)
  - `IdleConnTimeout`: 90s (Go default)
  - `MaxIdleConns`: 100, `MaxIdleConnsPerHost`: 10 (all traffic goes to a single host `{region}-docker.pkg.dev`, so the per-host limit is the effective cap)

## Verification

```bash
cd packages/docker-reverse-proxy
go build ./...   # ✅
go vet ./...     # ✅
golangci-lint run ./...  # ✅ 0 issues
```

## Notes

- This is a bug fix — no new behavior, only safety guardrails.
- No issue was opened beforehand per [CONTRIBUTING.md](CONTRIBUTING.md) ("Bug fixes are always welcome and can skip this step if the problem is clear").
- Timeout values were chosen conservatively for the GCP-internal deployment topology and can be tuned via env vars in a follow-up if needed.

### DIFF
--- a/packages/docker-reverse-proxy/internal/handlers/store.go
+++ b/packages/docker-reverse-proxy/internal/handlers/store.go
@@ -5,9 +5,11 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"net"
 	"net/http"
 	"net/http/httputil"
 	"net/url"
+	"time"
 
 	"github.com/e2b-dev/infra/packages/db/client"
 	authdb "github.com/e2b-dev/infra/packages/db/pkg/auth"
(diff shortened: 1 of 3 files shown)


AdaAibaby:

PR #3319, opened 2026-07-21
feat(api): add sandbox workload identity permission

gm-e2b:
feat(api): add sandbox IAM workload token configuration                                                                                               
                                                                                                                                                         
Add an optional iam.tokens map to sandbox create, keyed by a                                                                                          
caller-chosen token name, where each definition carries an exact                                                                                      
audience and tokenType. A non-empty, validated map enables workload                                                                                   
identity for the sandbox; there is no separate enabled flag.                                                                                          
                                                                                                                                                     
Admission requires a non-empty audience and tokenType "JWT-SVID",                                                                                     
preserves the audience exactly as received, and rejects an empty token                                                                                
name or any non-null filePath (file-based delivery is unsupported) with                                                                               
a 400 naming the offending field. The definitions are propagated to the                                                                               
orchestrator in SandboxConfig.iam and persisted in the running-sandbox                                                                                
(Redis) and paused-snapshot (Postgres) state, so they survive                                                                                         
pause/resume and orchestrator re-sync. A resume or fork restores the                                                                                  
configuration from the snapshot, so a fork inherits the parent                                                                                        
sandbox's IAM configuration; the resumed/forked execution still obtains                                                                               
a freshly generated execution ID.                                                                                                                     
                                                                                                                                                     
The API mints, signs, injects, or delivers no credential.                                                                                             
                                                                                                                                                     
Refs: EN-1860


spec/openapi.yml
```diff
@@ -752,11 +752,48 @@ components:
...
+          description: Audience of the workload token, stored exactly as provided.
+        tokenType:
+          type: string
+          description: Workload token type.
+        filePath:
+          type: string
+          nullable: true
+          description: >-
+            Reserved for future file-based delivery. File delivery is not
+            supported: a non-null value is rejected. Leave absent or null.
```
sitole: Reserving `filePath` now seems premature. Adding an optional field to the schema later is fully backward-compatible, so this buys nothing today — meanwhile every generated SDK exposes a field whose only behavior is a guaranteed 400. Simpler to drop it from the spec entirely and introduce it when file-based delivery actually ships. If the intent is to lock the name into SDKs ahead of time, that reasoning belongs in the PR description.

packages/api/internal/handlers/sandbox_create.go
```diff
@@ -197,6 +197,14 @@ func (a *APIStore) PostSandboxes(c *gin.Context) {
 		envdAccessToken = &accessToken
 	}
 
+	iamCfg, iamErr := buildSandboxIam(body.Iam)
```
sitole, line 202: Lets add feature flag returning this is unsupported feature so we can gate keep and allow for each team during beta testing.

packages/api/internal/handlers/sandbox_create.go
```diff
@@ -336,6 +345,53 @@ func (a *APIStore) PostSandboxes(c *gin.Context) {
...
+
+	reject := func(field, msg string) *api.APIError {
+		return &api.APIError{
+			Code:      http.StatusBadRequest,
+			Err:       fmt.Errorf("%s: %s", field, msg),
+			ClientMsg: fmt.Sprintf("%s: %s", field, msg),
+		}
+	}
+
+	tokens := make(map[string]types.SandboxIamToken, len(*iam.Tokens))
```
sitole, line 380: Lets add reasonlable limit on number of tokens.

sitole (changes requested):

packages/api/internal/handlers/sandbox_create.go
gm-e2b: Addressed, I added max 5 tokens for now.

packages/api/internal/handlers/sandbox_create.go
gm-e2b: Done.

spec/openapi.yml
gm-e2b: Makes sense. Dropped.

sitole (approved):

PR #3320, opened 2026-07-21
Export rootfs in the background during pause()/snapshot()

bchalios:
## What

When we are snapshotting a sandbox, during pause()/snapshot() we create a snapshot of both memory and rootfs.
Exporting the rootfs boils down to reflinking the overlay to the local cache path and then (asynchronously) uploading the cache to GCS.

This exporting process happens synchronously. We have seen that, _typically_ reflinking is very fast as it deals with metadata (rather than moving data around).

This PR makes the export asynchronous (optional, guarded by a feature flag). In the case of asynchronous export of the rootfs, we don't observe the reflink latency in the pause, but we _might_ observe it during resume time **if** the resume happens while the asynchronous seal operation is still ongoing.

## Why

We have seen that, whereas the reflink operation is typically fast, it can experience long tails when there's a lot of dirty data in the page cache that needs to be written to the backing storage.

For example, when we are taking a snapshot after writing 2GiB of data in the disk immediately after the write finished (E1) and 30 seconds after the write had completed and:

* E1 — pause immediately after write: the ~2 GB NVMe flush burst happens inside the pause window (host Dirty ~2 GB → 0 during the pause). → pause = 6.8 s
* E2 — 30 s settle before pause: the flush burst happens during the settle; Dirty is already 0 when the pause fires. → pause = 0.9 s
* Same write, same node — only difference is whether the flush overlaps the pause.

This nicely scales with amount of time we wait after the write:
* settle  0 s  →  ~4.2 s   (6.8 s tail)
* settle  5 s  →   3.78 s
* settle 10 s  →   3.71 s
* settle 20 s  →   1.53 s
* settle 30 s  →  ~0.7–0.9 s

### DIFF
--- a/packages/orchestrator/pkg/sandbox/block/cache.go
+++ b/packages/orchestrator/pkg/sandbox/block/cache.go
@@ -106,18 +106,62 @@ func (c *Cache) isClosed() bool {
 	return c.closed.Load()
 }
 
-func (c *Cache) ExportToDiff(ctx context.Context, out *os.File) (*header.DiffMetadata, error) {
-	ctx, childSpan := tracer.Start(ctx, "export-to-diff")
-	defer childSpan.End()
-
-	c.mu.Lock()
-	defer c.mu.Unlock()
+// DiffMetadata returns the dirty/empty diff metadata from the tracker without
+// copying any block data. It lets a deferred/background seal build the diff
+// header (and scheduling metadata) synchronously while the actual reflink copy
+// (ExportToDiff) runs off the critical path. The result matches what
+// ExportToDiff computes internally, provided the cache is frozen (no writes) in
+// between — which is guaranteed once the sandbox has been stopped and the cache
+// ejected.
+func (c *Cache) DiffMetadata() (*header.DiffMetadata, error) {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
 
 	if c.isClosed() {
 		return nil, NewErrCacheClosed(c.filePath)
 	}
 
 	if c.mmap == nil {
+		return header.NewDiffMetadata(c.blockSize, nil, nil), nil
+	}
+
+	dirty, empty := c.tracker.Export()
+
+	return header.NewDiffMetadata(c.blockSize, dirty, empty), nil
+}
+
+func (c *Cache) ExportToDiff(ctx context.Context, out *os.File) (*header.DiffMetadata, error) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	return c.exportToDiffLocked(ctx, out, nil)
+}
+
+// ExportToDiffWithMetadata copies the dirty ranges described by meta to out,
+// using meta's bitmap instead of re-reading the tracker. Callers that captured a
+// DiffMetadata earlier (e.g. the deferred rootfs seal, which reads it at setup
+// and exports later in the background) use this so the copied ranges are
+// guaranteed to match a header built from the same bitmap read.
+func (c *Cache) ExportToDiffWithMetadata(ctx context.Context, out *os.File, meta *header.DiffMetadata) (*header.DiffMetadata, error) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	return c.exportToDiffLocked(ctx, out, meta)
+}
+
+func (c *Cache) exportToDiffLocked(ctx context.Context, out *os.File, meta *header.DiffMetadata) (*header.DiffMetadata, error) {
+	ctx, childSpan := tracer.Start(ctx, "export-to-diff")
+	defer childSpan.End()
+
+	if c.isClosed() {
+		return nil, NewErrCacheClosed(c.filePath)
+	}
+
+	if c.mmap == nil {
+		if meta != nil {
+			return meta, nil
+		}
+
 		return header.NewDiffMetadata(c.blockSize, nil, nil), nil
 	}
 
@@ -137,8 +181,11 @@ func (c *Cache) ExportToDiff(ctx context.Context, out *os.File) (*header.DiffMet
 		logger.L().Warn(ctx, "error syncing file", zap.Error(err))
 	}
 
-	dirty, empty := c.tracker.Export()
-	diffMetadata := header.NewDiffMetadata(c.blockSize, dirty, empty)
+	diffMetadata := meta
+	if diffMetadata == nil {
+		dirty, empty := c.tracker.Export()
+		diffMetadata = header.NewDiffMetadata(c.blockSize, dirty, empty)
+	}
 
 	dst := int(out.Fd())
 	var writeOffset int64
--- a/packages/orchestrator/pkg/sandbox/build/deferred_diff.go
+++ b/packages/orchestrator/pkg/sandbox/build/deferred_diff.go
@@ -1,0 +1,102 @@
+//go:build linux
+
+package build
+
+import (
+	"context"
+
+	"github.com/e2b-dev/infra/packages/shared/pkg/storage"
+	"github.com/e2b-dev/infra/packages/shared/pkg/utils"
+)
+
+// deferredDiff is a Diff whose backing data is produced asynchronously. It is
+// returned synchronously from a pause that seals the rootfs in the background:
+// the cache key and block size are known up front (so DiffStore.Add and the
+// upload's compress-config validation work immediately), while every data-
+// bearing method blocks on the inner promise until the background seal resolves
+// the real Diff.
+//
+// The producer MUST always resolve the promise — with the sealed Diff on success
+// or an error on failure — otherwise the data methods (and Close) block forever.
+type deferredDiff struct {
+	cacheKey  DiffStoreKey
+	blockSize int64
+	inner     *utils.SetOnce[Diff]
+}
+
+var _ Diff = (*deferredDiff)(nil)
+
+// NewDeferredDiff wraps a promise of a Diff. cacheKey and blockSize are the
+// synchronously-known identity of the diff; inner is resolved by the background
+// sealer with the materialized Diff (or an error).
+func NewDeferredDiff(cacheKey DiffStoreKey, blockSize int64, inner *utils.SetOnce[Diff]) Diff {
+	return &deferredDiff{
+		cacheKey:  cacheKey,
+		blockSize: blockSize,
+		inner:     inner,
+	}
+}
+
+func (d *deferredDiff) CacheKey() DiffStoreKey {
+	return d.cacheKey
+}
+
+func (d *deferredDiff) BlockSize() int64 {
+	return d.blockSize
+}
+
+func (d *deferredDiff) CachePath(ctx context.Context) (string, error) {
+	inner, err := d.inner.WaitWithContext(ctx)
+	if err != nil {
+		return "", err
+	}
+
+	return inner.CachePath(ctx)
+}
+
+func (d *deferredDiff) ReadAt(ctx context.Context, p []byte, off int64, ft *storage.FrameTable) (int, error) {
+	inner, err := d.inner.WaitWithContext(ctx)
+	if err != nil {
+		return 0, err
+	}
+
+	return inner.ReadAt(ctx, p, off, ft)
+}
+
+func (d *deferredDiff) Slice(ctx context.Context, off, length int64, ft *storage.FrameTable) ([]byte, error) {
+	inner, err := d.inner.WaitWithContext(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	return inner.Slice(ctx, off, length, ft)
+}
+
+func (d *deferredDiff) Size(ctx context.Context) (int64, error) {
+	inner, err := d.inner.WaitWithContext(ctx)
+	if err != nil {
+		return 0, err
+	}
+
+	return inner.Size(ctx)
+}
+
+func (d *deferredDiff) FileSize(ctx context.Context) (int64, error) {
+	inner, err := d.inner.WaitWithContext(ctx)
+	if err != nil {
+		return 0, err
+	}
+
+	return inner.FileSize(ctx)
+}
+
+// Close waits for the seal to resolve and closes the materialized diff. If the
+// seal failed there is nothing to close (the producer cleans up the partial file
+// on error), so only close when the diff actually materialized.
+func (d *deferredDiff) Close() error {
+	if inner, err := d.inner.Wait(); err == nil {
+		return inner.Close()
+	}
+
+	return nil
+}
--- a/packages/orchestrator/pkg/sandbox/build/deferred_diff.go
+++ b/packages/orchestrator/pkg/sandbox/build/deferred_diff.go
@@ -4,10 +4,18 @@ package build
 
 import (
 	"context"
+	"errors"
 
 	"github.com/e2b-dev/infra/packages/shared/pkg/storage"
 	"github.com/e2b-dev/infra/packages/shared/pkg/utils"
 )
+
+// ErrDeferredSealFailed marks a deferred rootfs seal that failed. The seal runs
+// exactly once, so the failure is permanent: the promise stays settled with this
+// error and every subsequent data-bearing method on the diff returns it. Callers
+// such as the pause-upload retry loop match on it to stop retrying a diff that
+// can never materialize (rather than burning the whole retry budget on it).
+var ErrDeferredSealFailed = errors.New("deferred rootfs seal failed")
 
 // deferredDiff is a Diff whose backing data is produced asynchronously. It is
 // returned synchronously from a pause that seals the rootfs in the background:
--- a/packages/orchestrator/pkg/sandbox/build/deferred_diff.go
+++ b/packages/orchestrator/pkg/sandbox/build/deferred_diff.go
@@ -53,6 +53,19 @@ func (d *deferredDiff) BlockSize() int64 {
 	return d.blockSize
 }
 
+// sealed reports, without blocking, whether the background seal has resolved the
+// promise (with the materialized diff or an error). Cache eviction uses it to
+// skip a not-yet-sealed diff rather than block on the data methods (which wait
+// on the seal) or evict a fresh, still-in-flight snapshot.
+func (d *deferredDiff) sealed() bool {
+	select {
+	case <-d.inner.Done:
+		return true
+	default:
+		return false
+	}
+}
+
 func (d *deferredDiff) CachePath(ctx context.Context) (string, error) {
 	inner, err := d.inner.WaitWithContext(ctx)
 	if err != nil {
--- a/packages/orchestrator/pkg/sandbox/build/deferred_diff_test.go
+++ b/packages/orchestrator/pkg/sandbox/build/deferred_diff_test.go
@@ -1,0 +1,92 @@
+//go:build linux
+
+package build
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/e2b-dev/infra/packages/shared/pkg/utils"
+)
+
+// makeLocalDiff writes `data` to a fresh local diff file and materializes it.
+func makeLocalDiff(t *testing.T, blockSize int64, data []byte) Diff {
+	t.Helper()
+
+	f, err := NewLocalDiffFile(t.TempDir(), "build-id", Rootfs)
+	require.NoError(t, err)
+	_, err = f.WriteAt(data, 0)
+	require.NoError(t, err)
+
+	d, err := f.CloseToDiff(blockSize)
+	require.NoError(t, err)
+
+	return d
+}
+
+// TestDeferredDiff_IdentityIsSynchronous verifies CacheKey and BlockSize resolve
+// without blocking on the (still-unresolved) inner promise, so DiffStore.Add and
+// the upload's compress-config validation work the moment the diff is created.
+func TestDeferredDiff_IdentityIsSynchronous(t *testing.T) {
+	t.Parallel()
+
+	inner := utils.NewSetOnce[Diff]()
+	key := GetDiffStoreKey("build-id", Rootfs)
+	d := NewDeferredDiff(key, 4096, inner)
+
+	require.Equal(t, key, d.CacheKey())
+	require.EqualValues(t, 4096, d.BlockSize())
+}
+
+// TestDeferredDiff_DelegatesAfterResolve verifies the data methods block until
+// the promise resolves, then delegate to the materialized diff.
+func TestDeferredDiff_DelegatesAfterResolve(t *testing.T) {
+	t.Parallel()
+
+	blockSize := int64(4096)
+	data := make([]byte, blockSize)
+	for i := range data {
+		data[i] = 0x5A
+	}
+
+	materialized := makeLocalDiff(t, blockSize, data)
+	t.Cleanup(func() { _ = materialized.Close() })
+
+	inner := utils.NewSetOnce[Diff]()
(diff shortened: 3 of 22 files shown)


bchalios:
Thanks for the reviews. Status on each finding:

### Addressed in the current revision
- **Sync fallback for non-NBD providers** (gemini high / Cursor "Missing sync fallback for DirectProvider"): `DirectProvider.PrepareExportDiff` now returns the new sentinel `rootfs.ErrDeferredExportNotSupported`, and `Pause` detects it via `errors.Is` and falls back to the synchronous `pauseProcessRootfs` path (clearing `deferRootfsExport`). Enabling the flag on a Direct-provider sandbox degrades to synchronous export instead of failing the pause.
- **Sentinel error** (gemini medium, `nbd.go` / `direct.go`): added `ErrDeferredExportNotSupported` in the `rootfs` package.
- **Ejected `sealCache` leak on pause error** (gemini high / Cursor "Ejected cache leaked on pause error"): the ejected cache is closed by a `cleanup` guarded on `started`, so if Pause aborts before `startSeal` runs the cache is closed there; once the seal owns it, the goroutine closes it.

### Fixed in the incoming revision
- **Abort cleanup races with seal** (Cursor medium): good catch. The `sealCache` close was gated on `started`, but the `diffPromise.SetError` abort-resolver was not — so on the success path it still ran on snapshot release and could poison the promise, racing the seal's `SetValue` and dropping the sealed diff. The abort-resolver is now gated on `started` too, and `started` is an `atomic.Bool` (written on the pause goroutine, read on the cleanup goroutine). After the seal starts the resolver is a no-op; it only poisons the promise on the genuine abort path (before `startSeal`), where it must, so the deferred diff's `Close` never blocks on a seal that will never resolve.

### Intended tradeoff — proposing a follow-up rather than blocking here
- **Failed seal breaks cached snapshot** (Cursor high): this is the deliberate async-durability tradeoff of deferring the export, and it's feature-flag-gated and pause-only. Bounding the risk:
  - The seal is a local `copy_file_range` reflink; it only fails on genuine disk/IO errors (ENOSPC, EIO), which would fail the synchronous path too — just at a different time.
  - Pause already returns before the snapshot upload completes (the upload is fire-and-forget on `uploadsWG`), so "pause RPC succeeds, durability resolves shortly after" is the existing model; a failed seal extends that window to the local diff.

  The genuine difference from `main` is that a failed **seal** leaves the *local* cache entry unusable (its promise errors), whereas a failed **upload** on `main` still leaves a usable local diff for same-node resume. If we want to close that gap, I'd do it as a follow-up: on seal failure, evict the build's local template-cache entry so a same-node resume falls through cleanly instead of hitting a poisoned diff. That eviction is cross-layer (the entry is added in `server.snapshotAndCacheSandbox`; the failure is detected in `sandbox.runDeferredRootfsExport`), so I'd prefer to keep it out of this PR. Happy to file it.

bchalios:
Two more from the latest Cursor pass — both valid, both fixed in the incoming revision:

- **Empty deferred rootfs breaks peer lookup** (Cursor medium): correct — `AddSnapshot` only skips a concrete `*NoDiff`, so a `deferredDiff` that later resolves to `NoDiff` (a pause with no dirty filesystem blocks) would sit in the `DiffStore` and make peer `LookupDiff` succeed, then return `NoDiffError` from `Slice`/`Size` instead of `ErrNotAvailable`. Fixed at the source: `setupDeferredRootfsExport` already has the dirty bitmap synchronously (it builds the header before the seal), so when `diffMetadata.Dirty.IsEmpty()` it now returns a real `*NoDiff` directly — closing the ejected cache and skipping the background seal entirely (nothing to reflink). That matches the synchronous path exactly, so `AddSnapshot` omits it from the store and peer lookup keeps returning `ErrNotAvailable`.

- **Upload waits on rootfs before others** (Cursor medium): correct — `runV3`/`runV4` resolved `RootfsDiff.CachePath` (which blocks on the seal) *before* building the errgroup, serializing the memfile/snapfile/metadata uploads behind the reflink. Moved the rootfs `CachePath` resolution *into* its own errgroup goroutine in both paths, so only the rootfs upload waits on the seal and the other objects upload concurrently with it. (The rootfs *header* is built synchronously in the deferred path, so only the data path was gated.)

Build + vet + race tests green.

bchalios:
Two more from the re-review of the pushed fixes — both addressed in the incoming revision:

- **Harvest resumes before rootfs seal** (Cursor medium): right — the prefetch harvest's throwaway warm resume reads the rootfs, which with deferred export is a promise-backed diff that only resolves once the background seal finishes, so the resume would block on (and spend its budget against) the reflink. Fixed by making the harvest wait for the seal *before* its timed resume: `snapshotResult` now carries the rootfs diff, and `harvestResumePrefetchAsync` calls `rootfsDiff.CachePath(hCtx)` first — returns immediately on the synchronous/`NoDiff` paths, blocks until the seal on the deferred path, and if the seal fails or the harvest deadline fires first it records a `skipped` outcome and never touches a half-sealed snapshot (same skip semantics as a missed start slot). So the two flags now cooperate instead of the harvest reliably failing.

- **Deferred path skips rootfs metrics** (Cursor low): correct — `setupDeferredRootfsExport` built `diffMetadata` but never called `recordSnapshotDiff`, so the rootfs size/ratio metrics went dark with the flag on. Added the same `recordSnapshotDiff(ctx, "rootfs", diffMetadata, originalHeader)` call the synchronous `pauseProcessRootfs` path makes.

Build + vet + race tests (server + sandbox + build) all green.

bchalios:
Good catch — and this one was introduced by the V3 half of my upload reorder, so it's on me.

**V3 uploads rootfs header before seal** (Cursor high): correct. In `runV3` the rootfs header and body are separate errgroup goroutines, and the header resolves synchronously at pause time — so moving `CachePath` into only the body goroutine let the header goroutine finalize and publish the rootfs header to storage while the background seal was still running. A subsequent seal failure (which doesn't retry) would then leave a completed header with no body. `runV4` doesn't have this because its `uploadFramed` does body-then-header in one goroutine behind the same `CachePath` gate.

Fixed by gating the V3 rootfs **header** goroutine on the seal as well (`RootfsDiff.CachePath(egCtx)` before `storeHeaderWithMetrics`). Now both rootfs goroutines wait on the seal — so if it fails, neither the header nor the body is published — while memfile/snapfile/metadata still overlap the reflink. No-op on the synchronous path. Build + vet + race tests green.

bchalios:
**Peers see unsealed rootfs as ready** (Cursor medium): correct that, with the P2P-chunk-transfer flag and deferred export both on, `peerRegistry.Register` advertises the build before the seal finishes, so a peer's `Size`/`Slice` blocks on our reflink (or gets `Internal` if the seal fails) rather than `ErrNotAvailable`. Leaving it as-is intentionally for this PR — it's the same async-durability tradeoff as the deferred seal itself, and the impact is bounded and non-corrupting:

- A peer blocks at most for the seal duration, then its chunk client times out and falls back to fetching from GCS — a latency hit, not a failure.
- If the seal *fails*, the build's own upload fails too, so the build is unavailable everywhere regardless; the peer error is just one symptom of that already-documented failure mode.
- It only bites when both feature flags are enabled together, which isn't the initial rollout.

The clean fix is to advertise to peers only once the rootfs is sealed. I'm deferring it to a focused follow-up rather than bolting it on here, because doing it correctly has to preserve the `Register`→`Unregister` happens-before ordering: a naive background `Register`-after-seal races the upload's `Unregister` (the upload only finishes after the seal) and can leak a peer registration until its TTL. The right shape is to register inside the upload flow right after `CachePath` resolves and before the upload, which is a small refactor of the upload path — better as its own change. I'll file it.

packages/orchestrator/pkg/sandbox/sandbox.go
```diff
@@ -1848,6 +1884,159 @@ func pauseProcessRootfs(
...
+	if err != nil {
+		return nil, nil, nil, errors.Join(fmt.Errorf("building rootfs diff header: %w", err), sealCache.Close())
+	}
+
+	// No dirty filesystem blocks: the seal would produce an empty diff, i.e. the
+	// same *NoDiff the synchronous path returns. Return NoDiff directly (and skip
+	// the background seal — nothing to reflink) so AddSnapshot omits it from the
+	// DiffStore and peer LookupDiff keeps returning ErrNotAvailable rather than an
+	// entry whose Slice/Size yield NoDiffError. Close the ejected cache now since
+	// no seal will own it.
```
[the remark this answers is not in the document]
bchalios: This is valid. The instrumentation was added (and merged in main) after this PR was first created and was missing after I rebased. I have re-worked the code flow to measure the rootfs export latency in both cases. In the case of deferred export, we measure the critical path, i.e. setting up the export, not the export itself.

This is in purpose, so that we can measure accurately the effect of deferring the export in the pause/snapshot latency.

packages/orchestrator/pkg/sandbox/build/deferred_diff.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+func (d *deferredDiff) FileSize(ctx context.Context) (int64, error) {
+	inner, err := d.inner.WaitWithContext(ctx)
+	if err != nil {
+		return 0, err
+	}
+
+	return inner.FileSize(ctx)
+}
```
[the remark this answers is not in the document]
bchalios: Typically true, even if 1. not very likely to happen: the layer being sealed is the last one added, so very unlikely that the eviction routine will pick it up, 2. even if it does, the seal will eventually finish and 3. empirically the sealing finishes with a p90 < 1sec, so any slow down should be within this range.

In any case, we added logic to skip considering layers that have not been sealed yet.

packages/orchestrator/pkg/sandbox/sandbox.go
```diff
@@ -1848,6 +1908,159 @@ func pauseProcessRootfs(
...
+	buildID uuid.UUID,
+	blockSize int64,
+	diffPromise *utils.SetOnce[build.Diff],
+) error {
+	diffFile, err := build.NewLocalDiffFile(s.config.DefaultCacheDir, buildID.String(), build.Rootfs)
+	if err != nil {
+		return s.failRootfsSeal(diffPromise, fmt.Errorf("create rootfs diff file: %w", err))
+	}
+
+	if _, err := sealCache.ExportToDiff(ctx, diffFile.File); err != nil {
```
kalyazin: claude is telling me that `DiffMetadata` and `ExportToDiff` both read the tracker to get the bitmap, but at different times, and there is no structural guarantee that they are exactly the same. Can we pass the metadata in here?
```go
func (c *Cache) ExportToDiffWithMetadata(ctx context.Context, out *os.File, meta *header.DiffMetadata) error
```

packages/orchestrator/pkg/sandbox/sandbox.go
```diff
@@ -1796,29 +1815,70 @@ func buildProvisionalMemfile(
...
+	cleanup *Cleanup,
+) (d build.Diff, h *header.Header, startSeal func(context.Context), e error) {
 	ctx, span := tracer.Start(ctx, "process-rootfs")
 	defer span.End()
-
 	// Duration of the rootfs export+diff, split by fs_only (runs for both pause
 	// kinds) so the fs-only pause latency can be decomposed into quiesce + rootfs.
+	// This is the pause CRITICAL-PATH rootfs cost: the full export for the
+	// synchronous path, but only the eject/setup for the deferred path — the
+	// background reflink seal (runDeferredRootfsExport) is intentionally excluded.
```
kalyazin, line 1906: it looks like with excluding the reflink seal, we lose visibility on the reflink latency. Shall we add a new metric for that?

kalyazin: also that way we should be able to detect errors on the async path, otherwise they're silent?

packages/orchestrator/pkg/sandbox/block/cache.go
```diff
@@ -106,6 +106,30 @@ func (c *Cache) isClosed() bool {
 	return c.closed.Load()
 }
 
+// DiffMetadata returns the dirty/empty diff metadata from the tracker without
```
kalyazin, line 109: PR summary: we should make it clearer that the latency isn't eliminated in the case of immediate resume - it migrates to the pause.

packages/orchestrator/pkg/sandbox/build/deferred_diff.go
```diff
@@ -0,0 +1,115 @@
...
+
+// Close waits for the seal to resolve and closes the materialized diff. If the
+// seal failed there is nothing to close (the producer cleans up the partial file
+// on error), so only close when the diff actually materialized.
+func (d *deferredDiff) Close() error {
```
kalyazin, line 117: this seems to be only called on disk pressure-caused chunk eviction, but isn't on the TTL-caused eviction. Doesn't look material with 25h TTL, but becomes a problem if we choose to shorten TTL for some reason.

packages/orchestrator/pkg/sandbox/sandbox.go
kalyazin: `rootfs copy is synchronous today` is no longer true.

packages/orchestrator/pkg/server/prefetch_harvest.go
```diff
@@ -190,7 +190,25 @@ func (s *Server) harvestResumePrefetchAsync(
 		)
 
 		start := time.Now()
-		pages, outcome, err := harvester.run(hCtx, sbx, res.meta, res.upload, buildID, objectMetadata, consume)
+
+		// With deferred rootfs export the just-paused snapshot's rootfs diff is
+		// sealed (reflinked) in the background, and the throwaway warm resume below
+		// reads the rootfs. Wait for the seal to finish here instead of letting the
```
kalyazin, line 195: can we not count it into the harvestDurationHistogram and add record this wait as a separate metric?

packages/orchestrator/pkg/sandbox/sandbox.go
```diff
@@ -1848,6 +1908,159 @@ func pauseProcessRootfs(
 	return rootfsDiff, rootfsHeader, nil
 }
 
+// setupDeferredRootfsExport ejects the writable cache and stops the sandbox
+// (destroy path), then prepares the deferred rootfs diff + header from the frozen
+// ejected cache. It returns a startSeal closure that reflinks the cache into the
+// diff in the background, so the pause returns without paying the reflink stall.
+// Only safe on the suspend path, where nothing reads the diff before the seal
+// completes.
+func (s *Sandbox) setupDeferredRootfsExport(
```
kalyazin, line 2001: would it make sense to add tests to cover failure branches here?

packages/orchestrator/pkg/sandbox/build/deferred_diff_test.go
```diff
@@ -11,6 +11,21 @@ import (
 	"github.com/e2b-dev/infra/packages/shared/pkg/utils"
 )
 
+// TestDeferredDiff_SealedIsNonBlocking verifies sealed() reports readiness
+// without blocking on the promise, so cache eviction can skip an unsealed diff
+// instead of stalling on it.
+func TestDeferredDiff_SealedIsNonBlocking(t *testing.T) {
```
kalyazin, line 17: can we have a test for the fix in the commit, ie the diff is skipped when unsealed?

packages/orchestrator/pkg/sandbox/sandbox.go
```diff
@@ -1848,6 +1908,159 @@ func pauseProcessRootfs(
...
+	blockSize int64,
+	diffPromise *utils.SetOnce[build.Diff],
+) {
+	ctx, span := tracer.Start(ctx, "deferred-rootfs-export")
+	defer span.End()
+
+	if err := s.sealCacheToDiff(ctx, sealCache, buildID, blockSize, diffPromise); err != nil {
+		logger.L().Error(ctx, "deferred rootfs export failed", zap.Error(err))
+	} else {
+		telemetry.ReportEvent(ctx, "rootfs diff sealed (deferred)")
```
kalyazin, line 2107: claude says we don't report 'exported rootfs' event on the deferred path.

packages/orchestrator/pkg/sandbox/sandbox.go
bchalios: done!

packages/orchestrator/pkg/sandbox/sandbox.go
bchalios: Added a metric specifically for the sealing part.

packages/orchestrator/pkg/sandbox/block/cache.go
bchalios: updated the PR summary.

packages/orchestrator/pkg/sandbox/build/deferred_diff.go
bchalios: added logic to close an unsealed `deferredDiff` in a go routine (without cancel) in case of eviction.

packages/orchestrator/pkg/sandbox/sandbox.go
bchalios: fixed

packages/orchestrator/pkg/server/prefetch_harvest.go
bchalios: done!

packages/orchestrator/pkg/sandbox/sandbox.go
bchalios: Added failure branch tests covering the eject-failure early return.

packages/orchestrator/pkg/sandbox/build/deferred_diff_test.go
bchalios: added a test that ensures that the diff is skipped when unsealed and then later evicted after it was sealed.

packages/orchestrator/pkg/sandbox/sandbox.go
bchalios: adding missing telemetry events

packages/orchestrator/pkg/sandbox/sandbox.go
```diff
@@ -1796,29 +1817,70 @@ func buildProvisionalMemfile(
...
 
-	rootfsDiffFile, err := build.NewLocalDiffFile(cacheDir, buildId.String(), build.Rootfs)
+	if pauseOpts.deferRootfsExport {
+		rootfsDiff, rootfsHeader, startSeal, err := s.setupDeferredRootfsExport(ctx, buildID, originalHeader, cleanup)
+		switch {
+		case errors.Is(err, rootfs.ErrDeferredExportNotSupported):
+			// The provider (e.g. DirectProvider) can't defer; fall through to the
+			// synchronous export below. Safe because PrepareExportDiff returns this
+			// sentinel before ejecting/stopping anything.
+			pauseOpts.deferRootfsExport = false
```
kalyazin, line 1932: claude is saying we can never read this false, including the histogram function above (processRootfsDurationHistogram) where it could be helpful.

packages/orchestrator/pkg/sandbox/deferred_export_test.go
```diff
@@ -0,0 +1,101 @@
...
+	require.NoError(t, err)
+	require.Equal(t, blockData, got, "deferred diff must contain the dirtied block")
+	require.NoError(t, diff.Close())
+}
+
+// TestSetupDeferredRootfsExport_PrepareError verifies the earliest failure branch:
+// when ejecting the writable cache fails, setup propagates the error and returns
+// no diff/header/seal — so Pause falls back to the synchronous export rather than
+// registering a deferred diff whose seal would never run.
+func TestSetupDeferredRootfsExport_PrepareError(t *testing.T) {
```
kalyazin, line 99: claude thinks it would also be nice to test the call order of the cleanups in setupDeferredRootfsExport in the case of a later failure to avoid lockups.

packages/orchestrator/pkg/sandbox/sandbox.go
kalyazin: like deferRootfsExport could be an attribute to that metric

packages/orchestrator/pkg/sandbox/deferred_export_test.go
bchalios: added

packages/orchestrator/pkg/sandbox/sandbox.go
bchalios: done

packages/orchestrator/pkg/sandbox/sandbox.go
```diff
@@ -1921,6 +1992,175 @@ func pauseProcessRootfs(
...
+	// Export using the metadata captured at setup so the sealed data matches the
+	// header built from the same bitmap read (rather than re-reading the tracker).
+	if _, err := sealCache.ExportToDiffWithMetadata(ctx, diffFile.File, meta); err != nil {
+		return s.failRootfsSeal(diffPromise, errors.Join(fmt.Errorf("export rootfs diff: %w", err), diffFile.Close()))
+	}
+	telemetry.ReportEvent(ctx, "exported rootfs")
+
+	diff, err := diffFile.CloseToDiff(blockSize)
+	if err != nil {
+		return s.failRootfsSeal(diffPromise, fmt.Errorf("materialize rootfs diff: %w", err))
```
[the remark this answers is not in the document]
bchalios: Good catch — fixed in d545ded0d. The leak was in `CloseToDiff`: on a Sync/Stat/materialization failure it only closed the fd (`defer f.File.Close()`) without `os.Remove`, so the partial file orphaned in `DefaultCacheDir` with nothing registering it in the DiffStore (eviction can't reclaim a file it doesn't know about). Fixed at the `CloseToDiff` layer so every caller — the deferred seal, the synchronous pause path, and `ensurefreedisk/grow` — removes its partial file on failure. Added `TestLocalDiffFileCloseToDiffRemovesPartialOnError`.

packages/shared/pkg/featureflags/flags.go
```diff
@@ -186,6 +186,15 @@ var (
...
 	PeerToPeerAsyncCheckpointFlag = NewBoolFlag("peer-to-peer-async-checkpoint", false)
 
+	// DeferRootfsExportFlag moves the rootfs diff seal (the reflink, which forces a
+	// synchronous host->NVMe writeback) off the pause critical path: pause() ejects
+	// the cache and stops the sandbox, then reflinks the diff in the background so
+	// the call returns without the writeback stall. Applied only to the suspend
+	// (pause) path, where nothing reads the diff until a later resume. Off by
+	// default; falls back to the synchronous export when off or on a non-NBD
+	// provider.
+	DeferRootfsExportFlag = NewBoolFlag("defer-rootfs-export", false)
```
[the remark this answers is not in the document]
bchalios: Done in d545ded0d — added a **Deferred rootfs export** sub-bullet to the Pause section of `docs/ARCHITECTURE.md`. It documents the flag-gated behavior: the orchestrator ejects the writable COW cache during pause and returns, then seals it into the rootfs diff (reflink) in the background — so the async upload and any origin-node resume/prefetch that reads the rootfs diff wait on the seal, and a seal failure is permanent (the upload fails fast rather than retrying).

kalyazin (approved):
LGTM

The only thing is claude flags the following as a minor finding, and I can't see how realistic it is. Up to you if you find it worth fixing.
>  Seal goroutine isn't on a shutdown WaitGroup — after drainUploads times out the process can still exit mid-copy_file_range. Now bounded in blast radius by
  the CloseToDiff cleanup for in-process failures, but an exit-time partial still orphans.

PR #3325, opened 2026-07-21
fix(orchestrator): reject standby while draining

jakubno:
Prevent changing `draining` status to `standby`

---
<a href="https://app.blacksmith.sh/e2b-dev/codesmith/infra/pr/3325"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787230464&installation_model_id=14389&pr_number=3325&repository=e2b-dev%2Finfra&return_to=https%3A%2F%2Fgithub.com%2Fe2b-dev%2Finfra%2Fpull%2F3325&signature=5bd8332b7b87b6e3a7c34aa1116334c44d8840a8679c67483cc43896e713ba2b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>/codesmith</code> with what you need. Autofix is disabled.</sup>


spec/openapi.yml
```diff
@@ -3516,6 +3516,8 @@ paths:
       responses:
         "204":
           description: The node status was changed successfully
+        "400":
+          $ref: "#/components/responses/400"
```
arkamar:
```suggestion
        "409":
          $ref: "#/components/responses/409"
```
Not a blocker, but wouldn't `409 Conflict` fit here better for this. I am maybe a bit lost on http statuses and grpc codes. `code.FailedPrecondition` also sounds more aligned to the issue than `InvalidArgument`.

packages/api/internal/handlers/admin.go
```diff
@@ -68,6 +70,12 @@ func (a *APIStore) PostNodesNodeID(c *gin.Context, nodeId api.NodeID) {
 
 	err = node.SendStatusChange(ctx, body.Status)
 	if err != nil {
+		if status.Code(err) == codes.InvalidArgument {
```
arkamar:
```suggestion
		if status.Code(err) == codes.FailedPrecondition {
```
Ok, we should also change this now we changed the http status. The description for Invalid resource is 
```
	// InvalidArgument indicates client specified an invalid argument.
	// Note that this differs from FailedPrecondition. It indicates arguments
	// that are problematic regardless of the state of the system
	// (e.g., a malformed file name).
```
while for FailedPrecodition:
```
	// FailedPrecondition indicates operation was rejected because the
	// system is not in a state required for the operation's execution.
	// For example, directory to be deleted may be non-empty, an rmdir
	// operation is applied to a non-directory, etc.
	//
	// A litmus test that may help a service implementor in deciding
	// between FailedPrecondition, Aborted, and Unavailable:
	//  (a) Use Unavailable if the client can retry just the failing call.
	//  (b) Use Aborted if the client should retry at a higher-level
	//      (e.g., restarting a read-modify-write sequence).
	//  (c) Use FailedPrecondition if the client should not retry until
	//      the system state has been explicitly fixed. E.g., if an "rmdir"
	//      fails because the directory is non-empty, FailedPrecondition
	//      should be returned since the client should not retry unless
	//      they have first fixed up the directory by deleting files from it.
	//  (d) Use FailedPrecondition if the client performs conditional
	//      REST Get/Update/Delete on a resource and the resource on the
	//      server does not match the condition. E.g., conflicting
	//      read-modify-write on the same resource.
```
which fits better here.

packages/orchestrator/pkg/service/service_info.go
```diff
@@ -140,7 +142,9 @@ func convertMachineInfo(machineInfo machineinfo.MachineInfo) *orchestratorinfo.M
 
 func (s *Server) ServiceStatusOverride(ctx context.Context, req *orchestratorinfo.ServiceStatusChangeRequest) (*emptypb.Empty, error) {
 	logger.L().Info(ctx, "service status override request received", zap.String("status", req.GetServiceStatus().String()))
-	s.info.SetStatus(ctx, req.GetServiceStatus())
+	if !s.info.OverrideStatus(ctx, req.GetServiceStatus()) {
+		return nil, status.Error(codes.InvalidArgument, "cannot change node status from draining to standby")
```
arkamar:
```suggestion
		return nil, status.Error(codes.FailedPrecondition, "cannot change node status from draining to standby")
```

packages/orchestrator/pkg/service/service_info_test.go
```diff
@@ -0,0 +1,28 @@
...
+
+	info := &ServiceInfo{}
+	info.SetStatus(t.Context(), orchestratorinfo.ServiceInfoStatus_Draining)
+	server := &Server{info: info}
+
+	_, err := server.ServiceStatusOverride(t.Context(), &orchestratorinfo.ServiceStatusChangeRequest{
+		ServiceStatus: orchestratorinfo.ServiceInfoStatus_Standby,
+	})
+
+	require.Equal(t, codes.InvalidArgument, status.Code(err))
```
arkamar:
```suggestion
	require.Equal(t, codes.FailedPrecondition, status.Code(err))
```

arkamar (changes requested):
One more, sorry I wasn't explicit in previous comment. Anyway, I hope I caught all places as I did that in webui.

arkamar (approved):

PR #3342, opened 2026-07-23
fix(orchestrator): restore //go:build linux tags on template metadata package

### LINKED ISSUE #3316: orchestrator/pkg/template/metadata fails to compile on non-Linux (Darwin) after build tags removed
### What's wrong

PR #3299 removed the `//go:build linux` constraints from `template_metadata.go` and `prefetch.go` in `packages/orchestrator/pkg/template/metadata`, making those files eligible for non-Linux builds. But both still import `github.com/e2b-dev/infra/packages/orchestrator/pkg/sandbox/block`, and the `block` package is almost entirely Linux-only (`range.go`, `cache.go`, `memfd.go`, `overlay.go`, `device.go`, `tracker.go`, etc. all carry `//go:build linux`). Only `prefetch_tracker.go` supplies the shared data types.

As a result, `GOOS=darwin` builds of the metadata/prefetch packages fail: they import a package with no buildable Go files on Darwin. This blocks local development and tooling on macOS.

### Where

- `packages/orchestrator/pkg/template/metadata/template_metadata.go` (imports `sandbox/block`, no build tag): https://github.com/e2b-dev/infra/blob/main/packages/orchestrator/pkg/template/metadata/template_metadata.go#L1-L15
- `packages/orchestrator/pkg/template/metadata/prefetch.go` (same import, no build tag): https://github.com/e2b-dev/infra/blob/main/packages/orchestrator/pkg/template/metadata/prefetch.go#L1-L11
- `packages/orchestrator/pkg/sandbox/block/range.go` still `//go:build linux` (representative of the package): https://github.com/e2b-dev/infra/blob/main/packages/orchestrator/pkg/sandbox/block/range.go#L1

### How to reproduce

1. `GOOS=darwin GOARCH=amd64 go build ./packages/orchestrator/pkg/template/metadata/...`
2. Build fails because the imported `sandbox/block` package has no Go files buildable for Darwin.

### Suggested fix

Separate the portable metadata/prefetch data types (e.g. `PrefetchBlockEntry`) from the Linux-only `block` implementation so the metadata package can compile cross-platform, or add a minimal non-Linux stub for the `block` package. If Darwin builds of this package aren't intended to be supported, restoring a build constraint on the metadata files would also resolve the mismatch.

Found while running [Ito](https://ito.ai) (AI code review, free for open source) against recently merged PRs — full analysis: https://app.ito.ai/share/dacb5bb0-ce18-415f-aa26-0c706e869d42.

AdaAibaby:
## Summary

- `packages/orchestrator/pkg/template/metadata/template_metadata.go` and `prefetch.go` both import `packages/orchestrator/pkg/sandbox/block`, which is a Linux-only package (all its files carry `//go:build linux`).
- The build constraints on these two files were dropped in a prior refactor, causing compilation failures on Darwin and other non-Linux platforms.
- Restores `//go:build linux` at the top of both files.

Fixes #3316

## Test plan

- [ ] `go build ./packages/orchestrator/...` passes on Linux (verified on dev)
- [ ] `GOOS=darwin go build ./packages/orchestrator/...` no longer fails on the metadata package
/cc @jakubno @dobrac @ValentaTomas @arkamar @tvi Looking forward to your code review.


tvi (approved):

arkamar:
I am bit surprised with this as I think I tested it with `GOOS=darwin go build ./packages/orchestrator/pkg/template/metadata`. Anyway, this will break the `copy-build` command. Not sure if any darwin user uses it.

arkamar:
Anyway, #3299 should have add only those parts from block and metadata packages that should be OS agnostic.

arkamar:
I don't think this is fixing the real issue, see my comments in related issue #3316, `metadata` package builds just fine with `GOOS=darwin`.

PR #3367, opened 2026-07-23
trigger migrations issues on merge queue

djeebus:
this should prevent dupe migration versions from being merged


djeebus:
okay, good point. I'll merge this with #3366 , since it'll fix some things here too

PR #3368, opened 2026-07-23
fix(db): earlier autovacuum on snapshots

ValentaTomas:
Applies the same per-table autovacuum override `env_builds` already received (20260723030000) to `public.snapshots`.

**Why this is correct:** the change is a storage-parameter update only — it alters when autovacuum schedules a pass, never what any query reads or writes. It takes a brief ShareUpdateExclusive lock (no rewrite, no downtime) and is trivially revertible with `RESET`.

**What it improves:** the default trigger fires only after dead tuples reach a fixed fraction of the table, and on a large table that means cleanup starts far too late — meanwhile every read of this table pays visibility checks against the accumulated dead row versions, so scan and lookup costs creep up between passes. This table sits on hot read paths (snapshot lookups during sandbox lifecycle operations), so keeping its dead-tuple population small keeps those paths at their baseline cost. Triggering earlier also makes each pass smaller and cheaper — frequent small vacuums instead of rare huge ones.


ValentaTomas:
CI note: the ARM64 shard failures here are inherited from the goose version collision on main (two migrations sharing `20260723120000`) — every DB-touching shard on every open PR fails the same way. #3369 renumbers the later one; once it lands I'll rerun the checks here, which should then be green. The migration in this PR is unrelated to the failure.

djeebus (approved):

PR #3372, opened 2026-07-24
fix(db): earlier autovacuum on envs

ValentaTomas:
Applies the per-table autovacuum override `env_builds` received (20260723030000) to `public.envs`. Companion to #3368 (snapshots); completes the family for the hot write-path tables.

**Why this is correct:** scheduling-only storage-parameter change — no query semantics change, brief ShareUpdateExclusive lock, no rewrite, revertible via `RESET`.

**What it improves:** this table backs template upserts and is probed by several hot lookup paths, so it accumulates dead row versions continuously under normal operation. With the default fractional trigger, a large table can go a very long time before cleanup starts, and until then every upsert and probe wades through dead versions — the per-statement cost of the busiest write path degrades slowly and invisibly. An earlier trigger keeps the dead-version population bounded, which keeps upsert latency flat and also keeps planner statistics fresh (the analyze factor), preventing plan-quality drift on these paths.


ValentaTomas:
CI note: any red DB-touching shards here are inherited from the goose version collision on main — #3369 fixes it; checks rerun once that lands. This migration is unrelated to the failure.

djeebus (approved):

PR #3374, opened 2026-07-24
fix(envd): replace time.Sleep with ticker in ScanAndBroadcast for prompt shutdown

### LINKED ISSUE #3356: fix(envd): port scanner ScanAndBroadcast uses time.Sleep instead of select+ticker, delaying shutdown
## Summary

`ScanAndBroadcast()` in the envd port scanner uses a bare `time.Sleep` to pace scans. Because `Destroy()` closes the `scanExit` channel but the goroutine checks it **before** sleeping (not after), a `Destroy()` call that arrives during the sleep is not noticed until the sleep completes.

**Location:** `packages/envd/internal/port/scan.go:44`

```go
func (s *Scanner) ScanAndBroadcast() {
    for {
        processes, _ := net.Connections("tcp")
        for _, sub := range s.subs.Items() {
            sub.Signal(processes)
        }
        select {
        case <-s.scanExit:
            return
        default:
            time.Sleep(s.period)  // ← shutdown waits up to `period` here
        }
    }
}
```

## Root cause

The `select` only tests `scanExit` at the **top** of each iteration, before the sleep. Once `time.Sleep` starts, a concurrent `Destroy()` closes `scanExit` but the goroutine doesn't wake up — it sleeps the full `period` before looping back and seeing the closed channel.

If `period` is long (the default is not specified here, but envd calls `NewScanner(portScannerInterval)` — if that's on the order of seconds) shutdown is delayed unnecessarily.

## Proposed fix

Replace the `time.Sleep` with a `time.After` inside the select so `Destroy()` can interrupt the sleep immediately:

```go
func (s *Scanner) ScanAndBroadcast() {
    for {
        processes, _ := net.Connections("tcp")
        for _, sub := range s.subs.Items() {
            sub.Signal(processes)
        }
        select {
        case <-s.scanExit:
            return
        case <-time.After(s.period):
        }
    }
}
```

Or use a `time.Ticker` (more efficient — avoids a new timer allocation per iteration):

```go
func (s *Scanner) ScanAndBroadcast() {
    ticker := time.NewTicker(s.period)
    defer ticker.Stop()
    for {
        processes, _ := net.Connections("tcp")
        for _, sub := range s.subs.Items() {
            sub.Signal(processes)
        }
        select {
        case <-s.scanExit:
            return
        case <-ticker.C:
        }
    }
}
```

## Impact

With the current code, every envd shutdown is delayed by up to `portScannerInterval` waiting for the port scanner goroutine to wake from its sleep and notice the exit signal. The fix makes shutdown instantaneous for this goroutine.

AdaAibaby:
fixes https://github.com/e2b-dev/infra/issues/3356

## Problem

`ScanAndBroadcast()` used `time.Sleep` inside a `default` select case:

```go
select {
case <-s.scanExit:
    return
default:
    time.Sleep(s.period)  // ← Destroy() is invisible here
}
```

`Destroy()` closes `scanExit`, but the goroutine only checks `scanExit` at the **top** of each iteration — before the sleep starts. Once `time.Sleep` is entered, a concurrent `Destroy()` call is not noticed until the sleep completes. With `portScannerInterval = 1 s`, every envd shutdown was delayed by up to 1 second waiting for the port scanner goroutine.

## Fix

Replace `time.Sleep` with `time.NewTicker` so `scanExit` and the tick interval are waited on simultaneously in a single `select`:

```go
func (s *Scanner) ScanAndBroadcast() {
    ticker := time.NewTicker(s.period)
    defer ticker.Stop()

    for {
        processes, _ := net.Connections("tcp")
        for _, sub := range s.subs.Items() {
            sub.Signal(processes)
        }
        select {
        case <-s.scanExit:
            return          // exits immediately on Destroy()
        case <-ticker.C:
        }
    }
}
```

`Destroy()` now unblocks the goroutine instantly, regardless of how far into the current interval it is.

Using a `Ticker` (instead of `time.After`) also avoids allocating a new timer object on every loop iteration.

## Tests

Two new tests in `scan_test.go`:

- **`TestScanAndBroadcastDestroyExitsPromptly`** — creates a scanner with a 5 s period, lets it enter the select block, calls `Destroy()`, and asserts the goroutine exits within 200 ms (not after 5 s).
- **`TestScanAndBroadcastDestroyBeforeSleep`** — calls `Destroy()` before `ScanAndBroadcast` starts, verifying the goroutine still exits promptly.

Both pass in < 100 ms total.

/cc @jakubno @dobrac @ValentaTomas


arkamar (approved):
Thanks for the change, LGTM. I think the tests are not needed for this, as they could easily become flaky, at least as they were written. The timing assertions could false-fail on a loaded runner. The fix itself is a straightforward select on scanExit, verifiable by inspection.

I removed the additional comment from `ScanAndBroadcast()` as it described the implementation detail rather than the contract, and the original one-liner still covers what the function does.

I pushed these changes directly to your branch, along with the envd version bump to 0.6.11. The remaining diff is just the sleep→ticker change, which is exactly what we want here.

PR #3377, opened 2026-07-24
chore(db): drop the duplicate team-rename migration copy

ValentaTomas:
The 20260723120000 collision got fixed twice in parallel — #3366 renamed the team-rename migration to 20260723120001 and #3369 to 20260723172400 — and both renames merged, so main now carries the same migration content under two versions and goose applies it twice. Double-application is verified harmless (the backup table is IF NOT EXISTS and the updates match zero rows on a second pass), so this is hygiene, not a hotfix: drop the later copy, keep 20260723120001 as canonical. Environments that already recorded 20260723172400 as applied simply retain a stale version row, which goose tolerates.


ValentaTomas:
Superseded by #3378 (revert of #3369 — same tree change, cleaner provenance).

PR #3385, opened 2026-07-24
fix(orchestrator): survive SIGBUS from failing disks under mmap'd caches

arkamar:
A single bad 4K sector on a node's local NVMe crashed the whole
orchestrator and killed every sandbox on the node: the kernel cannot
return EIO for a memory access, so a failed page-in from an mmap'd cache
file delivers SIGBUS, which Go turns into an unrecoverable fatal error
inside the Cache.ReadAt memmove.

RunFaultSafe converts such faults into a regular error carrying the
fault address (debug.SetPanicOnFault + recover); other panics still
crash. It guards the in-process mmap copies: build segment reads, NBD
dispatch read/write/write-zeroes and the dedup compare. A bad block now
fails a single read or NBD request instead of the whole orchestrator,
and the affected file is just a local cache of remote storage anyway.
The guard is cheap enough to sit on the hot read path.

Recovered faults are logged and counted in
orchestrator.block.memory_fault, which should stay at zero on a healthy
node.


ValentaTomas:
Good find, though, cannot the same thing happen for the Cache struct that also uses mmap?

packages/orchestrator/pkg/sandbox/build/build.go
```diff
@@ -237,21 +237,40 @@ func (b *File) readSegments(ctx context.Context, p []byte, segments []readSegmen
...
+// readSegmentFaultSafe converts memory faults from the diff's mmap'd cache
+// into errors, so a bad disk block fails one read, not the process.
+func (b *File) readSegmentFaultSafe(ctx context.Context, p []byte, s readSegment) error {
+	err := block.RunFaultSafe(ctx, func() error { return b.readSegment(ctx, p, s) })
+	var faultErr *block.MemoryFaultError
+	if errors.As(err, &faultErr) {
+		cachePath, _ := s.diff.CachePath(ctx)
+		logger.L().Error(ctx, "memory fault reading build segment; local disk under the cache is likely failing",
+			zap.Error(err),
+			zap.String("cache_path", cachePath),
```
jakubno: `s.diff.CachePath(ctx)` on the fault path resolves to `DedupedMemfdCache.Path` for a memfd-backed diff, which calls `d.Wait(ctx)` and blocks until the dedup drain completes. In the sequential branch of readSegments the ctx here is the caller's request context rather than the errgroup's already-cancelled gctx, so a fault taken during the inflight-serve window

```suggestion
		cacheKey := string(s.diff.CacheKey())
		logger.L().Error(ctx, "memory fault reading build segment; local disk under the cache is likely failing",
			zap.Error(err),
			zap.String("diff_cache_key", cacheKey)
```

jakubno (approved):
lgtm! 

I added few related follow up tickets [EN-2000](https://linear.app/e2b/issue/EN-2000) and [EN-2003](https://linear.app/e2b/issue/EN-2003)

packages/orchestrator/pkg/sandbox/build/build.go
```diff
@@ -237,21 +237,40 @@ func (b *File) readSegments(ctx context.Context, p []byte, segments []readSegmen
...
+// readSegmentFaultSafe converts memory faults from the diff's mmap'd cache
+// into errors, so a bad disk block fails one read, not the process.
+func (b *File) readSegmentFaultSafe(ctx context.Context, p []byte, s readSegment) error {
+	err := block.RunFaultSafe(ctx, func() error { return b.readSegment(ctx, p, s) })
+	var faultErr *block.MemoryFaultError
+	if errors.As(err, &faultErr) {
+		cacheKey := string(s.diff.CacheKey())
+		logger.L().Error(ctx, "memory fault reading build segment; local disk under the cache is likely failing",
+			zap.Error(err),
+			zap.String("diff_cache_key", cacheKey)
```
arkamar:
```suggestion
			zap.String("diff_cache_key", cacheKey),
```

packages/orchestrator/pkg/sandbox/block/cache.go
```diff
@@ -325,8 +325,22 @@ func (c *Cache) Dedup(
...
 	compareStart := time.Now()
-	plan, err := dedupCompare(ctx, src, base, dirty, blockSize, bestEffort, budget)
-	if err != nil {
+	var plan *dedupPlan
+	if err := RunFaultSafe(ctx, func() error {
+		var compareErr error
+		plan, compareErr = dedupCompare(ctx, src, base, dirty, blockSize, bestEffort, budget)
+
+		return compareErr
+	}); err != nil {
```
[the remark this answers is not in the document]
arkamar: This one seemed to fit to this PR as well, therefore I added the guard in commit 79be12fb7.

jakubno (approved):

PR #3394, opened 2026-07-25
fix(api): let the analytics collector address carry a port

tvi:
## Problem

Every sandbox start/stop against a collector configured as `host:port` logs:

```
ERROR error sending Analytics event {"error": "rpc error: code = Unavailable desc = name resolver error: produced zero addresses"}
```

`NewAnalytics` did `net.JoinHostPort(host, "443")` unconditionally, so `ANALYTICS_COLLECTOR_HOST=localhost:5051` became the target `[localhost:5051]:443`, which cannot resolve. Analytics events are dropped; only the log line surfaces it.

## Changes

- Default the port to 443 only when the host has none, so the managed deployment's bare hostname behaves exactly as before while an explicit `host:port` is used as-is.
- Add `ANALYTICS_COLLECTOR_TLS` (`envDefault:"true"`). The transport was hardcoded to TLS 1.3 with the system root pool, which no plaintext collector can serve; setting it to `false` selects insecure credentials. Deployed configs are unaffected by the default.
- `gRPCApiKey.RequireTransportSecurity()` now mirrors that flag rather than `!env.IsLocal()`. The old check only recognized `ENVIRONMENT=local`, so any other non-prod environment name would have made gRPC refuse to attach the API key to a plaintext connection, failing at client creation.
- Tests for target resolution and the no-op client returned for an empty host.

## Testing

`go test ./internal/analytics_collector/... ./internal/cfg/...`

Manually verified against a plaintext collector on `localhost:5051`: the connection reaches `READY` at target `localhost:5051`, and a deliberately wrong API key comes back as `Unauthenticated: invalid API key`, confirming the per-RPC credentials are attached over the insecure transport.

### DIFF
--- a/packages/api/internal/analytics_collector/auth.go
+++ b/packages/api/internal/analytics_collector/auth.go
@@ -2,16 +2,19 @@ package analyticscollector
 
 import (
 	"context"
-
-	"github.com/e2b-dev/infra/packages/shared/pkg/env"
 )
 
 type gRPCApiKey struct {
 	apiKey string
+	// requireTransportSecurity mirrors the transport credentials the
+	// connection was created with. gRPC refuses to attach per-RPC credentials
+	// that require transport security to a plaintext connection, so a
+	// plaintext collector (local development) must relax this.
+	requireTransportSecurity bool
 }
 
-func newGRPCAPIKey(apiKey string) *gRPCApiKey {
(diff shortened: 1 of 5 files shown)


matthewlouisbrockman (approved):
as long as stays localhost only fine, otherwise plaintext gonna need to worry but should be good

PR #3395, opened 2026-07-25
chore(db): reword stats-migration comment

ValentaTomas:
Comment-only rewording of the 20260725100500 migration; no SQL statements change. Safe for already-migrated databases: goose tracks applied versions by number and does not checksum file content, so this neither re-runs nor invalidates the migration anywhere.


ValentaTomas:
Superseded: the comment rewording was folded into the tip commit directly (history rewritten in place).

PR #3396, opened 2026-07-25
fix(db): drop duplicate snapshots sandbox_id index

ValentaTomas:
## Summary

`idx_snapshots_sandbox_id` is an exact duplicate of `snapshots_sandbox_id_unique` (same single column, same order). The unique index serves all lookups; the copy has **0 lifetime scans** in `pg_stat_user_indexes`, while every snapshot write maintains it for nothing.

`DROP INDEX CONCURRENTLY` with the same timeout/INVALID-leftover handling as the recent drop-index precedent. Down rebuilds it concurrently (unbounded timeout, manual operation by design).


ValentaTomas:
closing for now, will revisit

PR #3397, opened 2026-07-25
fix(db): drop the unused plain env_builds status index

ValentaTomas:
## Summary

`idx_env_builds_status` (plain btree on raw `status`) has **a negligible number of lifetime scans** in `pg_stat_user_indexes`. Status-filtered access runs on the purpose-built indexes instead (`idx_env_builds_env_status_created`, `idx_env_builds_team_status_group`, `idx_env_builds_team_status_pagination`, and the covering index including `status_group`).

env_builds is the most write-amplified table in the registry (0% HOT updates, 8 indexes), so every dropped index removes real WAL and dead-tuple volume from each build insert/supersede.

Draft on purpose: requesting an owner's confirmation that nothing planned depends on a full-table raw-status scan path before this lands.

### DIFF
--- a/packages/db/migrations/20260725040800_drop_env_builds_status_index.sql
+++ b/packages/db/migrations/20260725040800_drop_env_builds_status_index.sql
@@ -5,8 +5,8 @@
 -- hangs unbounded nor dies to a short session default.
 SET statement_timeout = '1h';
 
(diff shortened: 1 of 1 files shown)


ValentaTomas:
closing for now, will revisit

PR #3398, opened 2026-07-25
fix(auth): debounce team_api_keys.last_used to one write per key per minute

ValentaTomas:
## Summary

`UpdateLastTimeUsed` fires on every authenticated request and is one of the largest sources of dead-tuple churn on `team_api_keys`, for a column that is minute-grade observability metadata. An in-process per-key debounce (60s window) keeps `last_used` fresh to the minute while removing almost all of that write load; with multiple auth replicas the worst case is one write per key per replica per minute.


ValentaTomas:
closing for now, will revisit

PR #3411, opened 2026-07-27
feat(orch): distro-aware template base-image provisioning

tomassrnka:
Resolve the base image's distro from its `/etc/os-release` ID and drive provisioning from a declared per-family profile — Debian/Ubuntu (apt), the RHEL family (dnf/microdnf/yum), Arch (pacman), and Alpine on OpenRC (apk) — instead of probing for a package manager. Unsupported or identity-less images are rejected with a clear build-log error.

Once merged, the biggest change is adduser -> useradd and introduction of /usr/local/bin/e2b-seed-certs instead of long one-liner for envd. 

Supersedes #3381 (reopened from a clean branch, no bot-thread churn). NixOS support follows in a stacked PR #3412.


packages/orchestrator/pkg/template/build/phases/base/distro/distro.go
```diff
@@ -0,0 +1,161 @@
...
+			"nftables", "iputils", "jq", "bash",
+		},
+		PkgQueryBody: `pacman -Q "$1" >/dev/null 2>&1`,
+		// -Syu (not -Sy then -S): Arch documents partial upgrades as unsupported.
+		PkgInstall:   `pacman -Syu --noconfirm --needed "$@"`,
+		InitBinary:   "/usr/lib/systemd/systemd",
+		TimeSyncUnit: "chronyd",
+		AdminGroup:   "wheel",
+		CABundle:     "/etc/ssl/certs/ca-certificates.crt",
+		CARefresh:    "update-ca-trust extract",
```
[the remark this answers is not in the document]
tomassrnka: False positive. On Arch the `ca-certificates` package ships `/etc/ssl/certs/ca-certificates.crt` as a symlink to the p11-kit bundle (`/etc/ca-certificates/extracted/tls-ca-bundle.pem`) that `update-ca-trust extract` regenerates, so the bundle already resolves at `$E2B_CA_BUNDLE` with no manual `ln -sf` — unlike RHEL, where update-ca-trust writes only under /etc/pki. Confirmed against the Arch `update-ca-trust(8)` man page and ca-certificates-utils, and it already boots on real KVM (envd started). Restored the one-line rationale in 7f299d2f4a.

packages/orchestrator/pkg/template/build/phases/base/distro/distro.go
```diff
@@ -0,0 +1,163 @@
...
+	},
+	{
+		Key:  "alpine",
+		Init: InitOpenRC,
+		IDs:  []string{"alpine"},
+		// shadow provides useradd/usermod (busybox's adduser takes different flags).
+		Packages: []string{
+			"openrc", "shadow", "openssh", "sudo", "chrony", "socat", "curl",
+			"ca-certificates", "fuse3", "iptables", "git", "nfs-utils", "less",
+			"nftables", "iputils", "jq", "bash",
```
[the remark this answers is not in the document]
tomassrnka: Confirmed and fixed in 8138f67a01. Alpine gets `ionice` from `util-linux-misc` (busybox does not ship the applet — verified against the Alpine package contents index), so `wrapperPrefix` was silently skipping the reset and user processes inherited envd realtime IO from `supervise-daemon --ionice 1:4`. Added `util-linux-misc` to the Alpine package list.

packages/orchestrator/pkg/template/build/core/rootfs/files/seed-certs.sh.tpl
```diff
@@ -0,0 +1,45 @@
...
+fi
+
+if [ ! -s /etc/ssl/certs/ca-certificates.crt ]; then
+    if command -v update-ca-certificates >/dev/null 2>&1; then
+        update-ca-certificates
+    else
+        # Provisioning guarantees the bundle on every supported family;
+        # reaching this means the image diverged after the build.
+        echo "e2b-seed-certs: CA bundle missing and no update-ca-certificates on this image; TLS trust will be degraded" >&2
+    fi
```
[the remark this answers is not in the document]
tomassrnka: Real — fixed in d700a178e0. The `command -v update-ca-certificates` guard silently no-opped on the RHEL family and Arch (they ship `update-ca-trust`), so the finalize pack never merged CAs added in later build layers and the cold-start fallback could not regenerate. Both `packCertBundleCmd` and `e2b-seed-certs` now probe `update-ca-trust extract` as well; on RHEL the seed path also copies the bundle from `/etc/pki` to the Debian-named path envd appends to (Arch's extract emits it directly). NixOS has neither tool and correctly stays a no-op — its bundle comes from the image configuration.

tomassrnka:
bugbot run

packages/orchestrator/pkg/template/build/phases/finalize/configure.sh
```diff
@@ -10,23 +10,51 @@ TEMPLATE_ID={{ .TemplateID }}
...
+    exit 1
+fi
 passwd -d user
-echo "user ALL=(ALL:ALL) NOPASSWD: ALL" >>/etc/sudoers
+# Skip the append when the entry is already present.
+if grep -q '^user ALL=(ALL:ALL) NOPASSWD: ALL' /etc/sudoers; then
+    echo "sudoers entry already present"
+else
+    echo "user ALL=(ALL:ALL) NOPASSWD: ALL" >>/etc/sudoers
+fi
```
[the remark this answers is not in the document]
tomassrnka: False positive — busybox `chmod` accepts the option after the mode. Tested in a live Alpine sandbox built from this branch:

```
which_chmod=/bin/chmod
/bin/chmod -> /bin/busybox
chmod 777 -R /tmp/cm  ->  rc=0
recursed_mode=777          # -R was honoured, not treated as a filename
perms_usrlocal=777         # the real configure.sh line already ran at build time
```

`/usr/local` being 777 in the shipped image is the direct evidence that this exact line succeeded during finalize. Alpine template builds also completed green three times at tip on real KVM (14s / 13s cold), which they could not have done if `set -e` had aborted the script here.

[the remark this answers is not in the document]
tomassrnka: Following up on my own reply: the stated failure is still a false positive — busybox on Alpine 3.24 accepts `chmod 777 -R` and recursed correctly in a live sandbox, and `/usr/local` ships at 777 — but I was wrong to close this with no change.

`chmod 777 -R` depends on options being accepted after an operand, which POSIX does not require. GNU chmod permits it only while `POSIXLY_CORRECT` is unset, so that spelling is one environment variable away from breaking on Debian/Ubuntu/Fedora/Arch too, not just on a non-GNU chmod. The correct spelling costs nothing.

Swapped both occurrences to `chmod -R 777` in 36ad3fdd12.

packages/orchestrator/pkg/template/build/core/rootfs/rootfs.go
```diff
@@ -233,6 +258,15 @@ func additionalOCILayers(
...
 		storage.GuestEnvdPath: {Bytes: envdFileData, Mode: 0o777},
 
+		// Systemd preset policy for envd. provision.sh removes /etc/machine-id,
+		// so the template's next boot is a systemd FIRST boot — and on first
+		// boot PID1 applies the distro preset policy to all units. On the
+		// RHEL family that policy ends with "disable *" (and the systemd RPM
+		// scriptlet's preset-all does the same during provisioning), which
+		// deletes envd's autostart symlink no matter how it was created.
+		// A 00- preset sorts before every distro policy file and wins.
+		"etc/systemd/system-preset/00-e2b.preset": {Bytes: []byte("enable envd.service\n"), Mode: 0o644},
```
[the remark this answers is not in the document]
tomassrnka: False positive — verified on a live Fedora sandbox built from this branch on real KVM:

```
systemctl is-enabled chronyd   -> enabled
systemctl is-active  chronyd   -> active
/etc/systemd/system/multi-user.target.wants/chronyd.service -> /usr/lib/systemd/system/chronyd.service
machine-id                     -> present (regenerated on first boot)
chronyd processes (/proc scan) -> 1
```

The wants symlink `systemctl enable "$E2B_TIMESYNC_UNIT"` creates survives the first boot; the `disable *` preset policy does not remove units that were explicitly enabled, so chrony autostarts normally on the RHEL family. `Reference ID : 00000000` just means unsynchronised — the dev slot has no `/dev/ptp0` under nested virt and falls back to the NTP pool, which is provision.sh's documented path.

packages/envd/internal/services/process/handler/handler.go
```diff
@@ -89,6 +89,25 @@ func currentNice() int {
...
+	if p, err := lookPath("nice"); err == nil {
+		prefix += fmt.Sprintf("%s -n %d ", p, niceDelta)
+	}
+
+	return prefix
+}
+
 func New(
 	ctx context.Context,
 	user *user.User,
```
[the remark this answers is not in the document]
tomassrnka: Correct, and a real miss — CLAUDE.md line 138 requires `pkg/version.go` to be bumped on every behavioral change, and `wrapperPrefix` resolving ionice/nice via `exec.LookPath` (and omitting either when absent) is exactly that. Bumped 0.6.11 -> 0.6.12 in b39c12643b.

packages/orchestrator/pkg/template/build/phases/base/distro/init.go
```diff
@@ -0,0 +1,132 @@
...
+# chrony still syncs in the background, nothing needs to wait for it.
+# masking a unit that doesn't exist on this distro still succeeds (systemctl
+# mask just writes the /dev/null symlink), so a failure here is real.
+systemctl mask chrony-wait.service
+
+echo "Disable slow boot units not needed in the sandbox"
+# binfmt registrations (foreign-arch exec) take ~1s of CPU early in boot and
+# compete with envd start; e2scrub is for LVM-backed ext4 only.
+systemctl mask systemd-binfmt.service
+systemctl mask e2scrub_reap.service`,
```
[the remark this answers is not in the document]
tomassrnka: Partly right, and the part that is right was a real bug — thanks.

Measured on real KVM before the fix:

| family | `is-enabled` | `is-active` | sshd procs |
|---|---|---|---|
| arch | **disabled** | **inactive** | **0** |
| fedora | enabled | active | 1 |
| ubuntu | (ssh.service in wants) | — | listening on :22 |

So the RHEL half of the claim is wrong — the RPM scriptlet enables `sshd` and the `disable *` preset does not undo it. But **Arch was genuinely shipping with sshd_config written and the daemon never enabled**, exactly as described.

Fixed in 4b29d52061: the SSH unit name is now carried per profile the way `E2B_TIMESYNC_UNIT` already is (`ssh` on Debian, `sshd` on RHEL/Arch/Alpine) and the systemd `e2b_init_setup` enables it; the OpenRC branch already did `rc-update add sshd`. After, on the same slot:

```
arch:   is-enabled=enabled  is-active=active  sshd procs=1
fedora: is-enabled=enabled  is-active=active  sshd procs=1
alpine: sshd procs=1 (OpenRC)
```

All four families rebuilt green with `https=200`.

tomassrnka:
bugbot run

packages/orchestrator/pkg/template/build/phases/base/distro/init.go
```diff
@@ -0,0 +1,139 @@
...
+# provide "net") without touching the kernel-managed eth0.
+printf 'auto lo\niface lo inet loopback\n' > /etc/network/interfaces
+if [ -e /etc/init.d/networking ]; then
+    rc-update add networking boot
+else
+    echo "OpenRC networking service not present on this image; services needing 'net' must not be enabled"
+fi
+
+echo "Enable time synchronization ($E2B_TIMESYNC_UNIT)"
+rc-update add "$E2B_TIMESYNC_UNIT" default
```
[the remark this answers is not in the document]
tomassrnka: The logic is right — the message said services needing `net` must not be enabled and the next line enabled `chronyd`, which does declare `need net`. Self-contradictory, so worth cleaning up.

But the severity is overstated: that `else` is unreachable on any image we support. Checked on a live Alpine sandbox from this branch:

```
networking_script = present
networking_owner  = /etc/init.d/networking is owned by openrc-0.63.2-r0
boot runlevel     = bootmisc hostname localmount networking sysctl
default runlevel  = chronyd envd sshd
chronyd need      = need net
chronyd procs     = 2      (running)
```

`/etc/init.d/networking` ships with `openrc` itself, which is the first entry in the Alpine profile's package list. If `openrc` were somehow absent, `rc-update` would not exist either and the block would have failed ~30 lines earlier, so the branch cannot fire in practice.

Fixed the contradiction in e89a2eb235 rather than the control flow: the warning now states the actual consequence (nothing provides `net`, so time sync will fail to start) and notes why the branch is unreachable. Keeping `chronyd` enabled is deliberate — if it ever did fire, a failure visible in the boot log beats a sandbox silently running without time sync.

tomassrnka:
```
% e2b sandbox create pr3411-rocky

Use the following link to inspect this Sandbox live inside the E2B Dashboard️:
↪ https://e2b.dev/dashboard/inspect/sandbox/idxvoxkc298wzot8wp0gt

Terminal connecting to template pr3411-rocky with sandbox ID idxvoxkc298wzot8wp0gt
~ $ cat /etc/os-release
NAME="Rocky Linux"
VERSION="9.3 (Blue Onyx)"
ID="rocky"
ID_LIKE="rhel centos fedora"
VERSION_ID="9.3"
PLATFORM_ID="platform:el9"
PRETTY_NAME="Rocky Linux 9.3 (Blue Onyx)"
ANSI_COLOR="0;32"
LOGO="fedora-logo-icon"
CPE_NAME="cpe:/o:rocky:rocky:9::baseos"
HOME_URL="https://rockylinux.org/"
BUG_REPORT_URL="https://bugs.rockylinux.org/"
SUPPORT_END="2032-05-31"
ROCKY_SUPPORT_PRODUCT="Rocky-Linux-9"
ROCKY_SUPPORT_PRODUCT_VERSION="9.3"
REDHAT_SUPPORT_PRODUCT="Rocky Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="9.3"
~ $
logout

Closing terminal connection to template pr3411-rocky with sandbox ID idxvoxkc298wzot8wp0gt
-----
% e2b sandbox create pr3411-arch

Use the following link to inspect this Sandbox live inside the E2B Dashboard️:
↪ https://e2b.dev/dashboard/inspect/sandbox/ida3blmjj6qbwgay2yoif

Terminal connecting to template pr3411-arch with sandbox ID ida3blmjj6qbwgay2yoif
[user@e2b ~]$ cat /etc/os-release
NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
VERSION_ID=20260726.0.562117
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://gitlab.archlinux.org/groups/archlinux/-/issues"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo
[user@e2b ~]$
logout
```

tomassrnka:
@cla-bot check

packages/envd/internal/services/process/handler/handler.go
```diff
@@ -89,6 +89,25 @@ func currentNice() int {
...
 }
 
+// wrapperPrefix builds the priority-tool part of the process wrapper from
+// whatever the image actually ships. ionice/nice are util-linux/
+// coreutils conveniences that minimal and busybox-based images (Alpine, UBI)
+// may lack or keep elsewhere than /usr/bin — a missing helper must degrade to
+// running the command without that priority adjustment, never to a failed
+// spawn (exit 127 killed every process on such images). lookPath is injected
+// for testability; production passes exec.LookPath.
+func wrapperPrefix(niceDelta int, lookPath func(string) (string, error)) string {
```
dobrac: The `wrapperPrefix` function is nice tool specific, the function name should reflect that rather than having comment explaining it.

The ionice command also contains specific arguments that are not reflected in the function arguments, making the function narrowly specific without the function name reflecting it.
Either include that in the name, or parametrize the function.

packages/orchestrator/pkg/template/build/core/rootfs/files/provision-runner.sh.tpl
```diff
@@ -0,0 +1,33 @@
...
+$BB sh /usr/local/bin/provision.sh 2>&1 | $BB sed "s/^/{{ .ProvisionLogPrefix }}/"
+
+# Flush filesystem changes to disk before the snapshot.
+$BB sync
+if command -v fsfreeze >/dev/null 2>&1; then
+    fsfreeze --freeze /
+else
+    # No util-linux on this image; the double sync flushes the ext4 journal
+    # and the VM is paused before the snapshot is taken.
+    echo "fsfreeze not available on this image; using sync-only flush"
```
dobrac: It might be valuable to prefix this message with the `ProvisionLogPrefix` to expose the message to the user too for full transparency.

packages/orchestrator/pkg/template/build/core/rootfs/rootfs.go
```diff
@@ -41,16 +42,40 @@ var tracer = otel.Tracer("github.com/e2b-dev/infra/packages/orchestrator/pkg/tem
...
+// it feeds filesHash: a change here must rotate the fallback provision
+// version like any other baked-layer change.
+var enableSymlinks = map[string]string{
+	// Enable envd service autostart. The target MUST be absolute: the link
+	// lives in multi-user.target.wants/, so a relative target would resolve
+	// inside that directory and dangle — and provision.sh's offline
+	// `systemctl enable $E2B_TIMESYNC_UNIT` prunes dangling .wants symlinks,
+	// silently disabling envd on distros where the link dangles.
+	"etc/systemd/system/multi-user.target.wants/envd.service": "/etc/systemd/system/envd.service",
+	// NOTE: chrony autostart is enabled by provision.sh via `systemctl enable
```
dobrac, line 55: Is the chrony removal here fully equivalent function wise in terms of timing too?

packages/orchestrator/pkg/template/build/phases/base/distro/distro.go
```diff
@@ -0,0 +1,179 @@
...
+	return ids
+}
+
+// ShellSelector generates the POSIX-sh block provision.sh sources: it switches
+// on the guest's $E2B_DISTRO_ID and defines the profile's packages, shell
+// functions, init path, time-sync unit, admin group and CA handling. An
+// unrecognized id exits 1 with a customer-visible error.
+func ShellSelector() string {
+	var b strings.Builder
+	b.WriteString(`case "$E2B_DISTRO_ID" in` + "\n")
```
dobrac, line 160: What do you think about using the Go templating instead?

packages/orchestrator/pkg/template/build/phases/base/distro/distro.go
```diff
@@ -0,0 +1,179 @@
...
+		Key:  "debian",
+		Init: InitSystemd,
+		IDs:  []string{"debian", "ubuntu"},
+		Packages: []string{
+			"systemd", "systemd-sysv", "openssh-server", "sudo", "chrony", "socat",
+			"curl", "ca-certificates", "fuse3", "iptables", "git", "nfs-common",
+			"less", "nftables", "iputils-ping", "jq",
+		},
+		PkgQueryBody: `dpkg-query -W -f='${Status}' "$1" 2>/dev/null | grep -q "install ok installed"`,
+		PkgInstall:   "apt-get -q update\n    DEBIAN_FRONTEND=noninteractive DEBCONF_NOWARNINGS=yes apt-get -qq -o=Dpkg::Use-Pty=0 install -y --no-install-recommends \"$@\"",
```
dobrac: nit: new lines here might not be that nice compared to simply doing `;` or `&&`

packages/orchestrator/pkg/template/build/phases/base/distro/distro.go
```diff
@@ -0,0 +1,179 @@
...
+
+var Profiles = []Profile{
+	{
+		Key:  "debian",
+		Init: InitSystemd,
+		IDs:  []string{"debian", "ubuntu"},
+		Packages: []string{
+			"systemd", "systemd-sysv", "openssh-server", "sudo", "chrony", "socat",
+			"curl", "ca-certificates", "fuse3", "iptables", "git", "nfs-common",
+			"less", "nftables", "iputils-ping", "jq",
```
dobrac, line 53: Could we add some guarantees that the installed packages will be the same base set conceptually in all profiles?

packages/orchestrator/pkg/template/build/phases/base/provision.sh
```diff
@@ -13,13 +13,30 @@ echo "Starting provisioning script"
...
+# rejected by name rather than guessed.
+echo "Detecting base image distribution"
+if [ -r /etc/os-release ]; then
+    . /etc/os-release
+    E2B_DISTRO_ID="${ID:-unknown}"
+else
+    E2B_DISTRO_ID="unknown (image has no /etc/os-release)"
+fi
+
+{{ .DistroSelector }}
```
dobrac, line 29: Could we instead of embedding shell script serialized in a string - which set variables that are loosely defined/used in the provision.sh - rather provide the profiles list struct and then have the distro selection script directly in the provision.sh?

packages/orchestrator/pkg/template/build/phases/base/distro/init.go
```diff
@@ -0,0 +1,143 @@
...
+if [ -e /etc/init.d/sshd ]; then
+    rc-update add sshd default
+else
+    echo "sshd service not present on this image; skipping"
+fi`,
+}
+
+// indentBlock indents every non-empty line of a shell block for embedding
+// inside the generated selector function bodies.
+func indentBlock(s, prefix string) string {
```
dobrac, line 134: Is this used anywhere?

packages/orchestrator/pkg/template/build/phases/base/distro/init.go
```diff
@@ -0,0 +1,143 @@
...
+	// InitSystemd — Debian/Ubuntu, RHEL/Fedora, Arch (systemd via envd.service).
+	InitSystemd InitSystem = "systemd"
+	// InitOpenRC — Alpine (busybox init → OpenRC via the baked /etc/init.d/envd).
+	InitOpenRC InitSystem = "openrc"
+)
+
+// initSetup is the provisioning-time shell block per init system. Bodies may
+// reference the selector's profile variables (e.g. $E2B_TIMESYNC_UNIT), defined
+// in the same case arm before the function is called.
+var initSetup = map[InitSystem]string{
```
dobrac, line 21: can we move these parts to the Go template (and separate files) instead?

packages/orchestrator/pkg/template/build/phases/base/provision.sh
```diff
@@ -32,18 +49,29 @@ done
 
 if [ -n "$MISSING" ]; then
     echo "Missing packages detected, installing:$MISSING"
-    apt-get -q update
-    DEBIAN_FRONTEND=noninteractive DEBCONF_NOWARNINGS=yes apt-get -qq -o=Dpkg::Use-Pty=0 install -y --no-install-recommends $MISSING
+    # shellcheck disable=SC2086
```
dobrac, line 61: Why do we need to do this?

packages/orchestrator/pkg/template/build/phases/finalize/configure.sh
```diff
@@ -10,34 +10,62 @@ TEMPLATE_ID={{ .TemplateID }}
...
 
 echo "Add sudo to 'user' with no password"
-usermod -aG sudo user
+# Admin group differs by distro (sudo on Debian/Ubuntu, wheel elsewhere); the
+# NOPASSWD sudoers entry below is what actually grants privileges. Neither
+# group existing is a real error.
+if getent group sudo >/dev/null; then
+    usermod -aG sudo user
+elif getent group wheel >/dev/null; then
+    usermod -aG wheel user
```
dobrac, line 58: This is already defined in the `Profiles` in the `distro.go` file. We should reuse the same definition.

dobrac (approved):
Looks great! Mostly minor nits, structure organization, cleanups.

The test coverage is poor. There should be at least integration tests for the different distributions checking the template build succeeds.

packages/envd/internal/services/process/handler/handler.go
tomassrnka: fair, renamed to more meaningful name `ioniceNicePrefix`

packages/orchestrator/pkg/template/build/core/rootfs/rootfs.go
tomassrnka: Both variants take effect before first boot

packages/orchestrator/pkg/template/build/phases/base/distro/distro.go
tomassrnka: Agreed, this is prob nice thing to do, will do in follow-up PR

packages/orchestrator/pkg/template/build/phases/base/distro/init.go
tomassrnka: Yes — ShellSelector() in distro.go calls it to indent each e2b_init_setup() body.

packages/orchestrator/pkg/template/build/phases/base/provision.sh
tomassrnka: TL;DR - making sure this works on ash correctly;

The # shellcheck disable=SC2086 only silences the warning (SC2086) — the unquoted $MISSING deliberately word-splits so each package lands as its own argument; quoting it would pass them all as a single argument. ShellCheck's suggested fix (an array) isn't available here: provision.sh is POSIX sh and must run under busybox ash.

packages/orchestrator/pkg/template/build/phases/base/distro/init.go
tomassrnka: Ok, will do in follow-up PR

packages/orchestrator/pkg/template/build/phases/base/provision.sh
tomassrnka: Ok, but again as follow-up PR, not right now

packages/orchestrator/pkg/template/build/phases/base/distro/distro.go
tomassrnka: Will fix in #3437

packages/orchestrator/pkg/template/build/phases/finalize/configure.sh
```diff
@@ -10,34 +10,61 @@ TEMPLATE_ID={{ .TemplateID }}
...
+    fi
 fi
 
 echo "Add sudo to 'user' with no password"
-usermod -aG sudo user
+# Admin group differs by distro (sudo on Debian/Ubuntu, wheel elsewhere); the
+# NOPASSWD sudoers entry below is what actually grants privileges. The group
+# comes from the distro profile, resolved and persisted by provisioning —
+# defined once in distro.go, not re-probed here.
+. /usr/local/share/e2b/distro.env
```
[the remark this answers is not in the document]
tomassrnka: This is actually super-valid, will fix.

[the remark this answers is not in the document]
tomassrnka: Fixed in 06f2fa9e96 — finalize and the USER step now source distro.env only when present and fall back to probing sudo/wheel for parents provisioned before the file existed (FROM-template builds reuse the parent rootfs without re-provisioning, so the file can legitimately be absent). Fallback path exercised under set -eu; new-image path unchanged.

packages/orchestrator/pkg/template/build/core/rootfs/files/envd.service.tpl
```diff
@@ -46,7 +46,9 @@ LimitCORE=infinity
 # guaranteed present for the sandbox's routable lifetime. The only gap is guest
 # units that auto-start and egress over TLS before /init; that is accepted
 # (revisit if a template needs boot-time egress).
-ExecStartPre=/bin/sh -c 'mountpoint -q /etc/ssl/certs || { mkdir -p /run/e2b/certs && { tar -C /run/e2b/certs -xf /usr/local/share/e2b/ssl-certs.tar 2>/dev/null || cp -a /etc/ssl/certs/. /run/e2b/certs/ 2>/dev/null; }; mount --bind /run/e2b/certs /etc/ssl/certs; } && ([ -s /etc/ssl/certs/ca-certificates.crt ] || update-ca-certificates)'
+# The seeding logic is shared with the OpenRC service (Alpine) and never fails
+# the unit — a degraded trust store is recoverable, a dead envd is not.
+ExecStartPre=/usr/local/bin/e2b-seed-certs
```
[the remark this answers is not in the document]
tomassrnka: not introduced here, will fix separately

PR #3425, opened 2026-07-28
feat(metrics): label pause/resume telemetry by fs_only

bchalios:
Makes filesystem-only **pause and resume** distinguishable in metrics, so the Filesystem-only Snapshots dashboard can stop relying on "(all pauses)" context panels.

## Change
- **New histogram `orchestrator.sandbox.pause.duration`** — Pause handler, attrs `fs_only` + `success` → fs-only pause call-count / error-rate / e2e latency.
- **`fs_only` on `orchestrator.snapshot.upload.failed`** — threaded via `snapshotResult`.
- **`fs_only` on `orchestrator.sandbox.create.duration`** — set at the reboot/resume fork, so e2e create/resume latency splits **filesystem-only reboot vs memory restore**. Combined with the existing `sandbox.resume` bool: `resume=false`→fresh create; `resume=true,fs_only=false`→memory resume; `resume=true,fs_only=true`→fs-only reboot.

## Why
fs-only-ness was recorded only as a *span* attribute (`fs-only-snapshot`), never a metric label, so pause/resume e2e latency, error rate, and upload failures couldn't be scoped to fs-only. Reboot-vs-resume was already distinguishable on the envd-init/uffd metrics (`start_type`); this adds it to the **e2e** `create.duration` as well.

## Notes
- `Pause` uses named returns so the deferred metric can read the final error.
- Pure additive instrumentation; no pause/resume/snapshot logic changed.
- Follow-up: monitoring-repo PR to switch the dashboard's pause/upload/resume panels to `fs_only="true"` once this deploys.

Verified: `go build`/`vet`/`golangci-lint` clean, `go test ./pkg/server/` passes.

### DIFF
--- a/packages/orchestrator/pkg/server/sandboxes.go
+++ b/packages/orchestrator/pkg/server/sandboxes.go
@@ -82,15 +82,19 @@ func (s *Server) Create(ctx context.Context, req *orchestrator.SandboxCreateRequ
 	defer childSpan.End()
 
 	isResume := req.GetSandbox().GetSnapshot()
+	// fsOnly is set at the resume fork below when this takes the filesystem-only
+	// reboot path (vs a memory restore), so create/resume e2e latency can be
+	// split reboot vs memory — mirroring the fs_only pause label. Combined with
+	// sandbox.resume: resume=false → fresh create; resume=true,fs_only=false →
+	// memory resume; resume=true,fs_only=true → filesystem-only reboot.
+	var fsOnly bool
 	createStart := time.Now()
 	defer func() {
-		if createErr != nil {
-			return
-		}
-
 		s.sandboxCreateDuration.Record(ctx, time.Since(createStart).Milliseconds(),
 			metric.WithAttributes(
 				attribute.Bool("sandbox.resume", isResume),
+				attribute.Bool("fs_only", fsOnly),
+				attribute.Bool("success", createErr == nil),
 			),
 		)
 	}()
@@ -228,6 +232,7 @@ func (s *Server) Create(ctx context.Context, req *orchestrator.SandboxCreateRequ
 
 	var sbx *sandbox.Sandbox
 	if meta.IsFilesystemOnly() {
+		fsOnly = true
 		sbx, err = s.sandboxFactory.RebootSandbox(
 			ctx,
 			template,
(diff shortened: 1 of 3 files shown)


packages/orchestrator/pkg/server/sandboxes.go
```diff
@@ -645,10 +653,23 @@ func recordSandboxKill(ctx context.Context, counter metric.Int64Counter, killRea
...
 
+	// Record pause duration split by fs_only vs memory (the gRPC RPC metric
+	// can't distinguish them) and success, so dashboards can scope pause
+	// call-count / error-rate / latency to filesystem-only pauses.
+	pauseStart := time.Now()
+	defer func() {
+		s.sandboxPauseDuration.Record(ctx, time.Since(pauseStart).Milliseconds(),
+			metric.WithAttributes(
+				attribute.Bool("fs_only", in.GetFilesystemOnly()),
+				attribute.Bool("success", err == nil),
```
kalyazin, line 665: Not strictly a part of this, but Claude is telling me that we don't differentiate success/failure for Create like we do for Pause. Seems like we can make it uniform via:
```go
        defer func() { 
                s.sandboxCreateDuration.Record(ctx, time.Since(createStart).Milliseconds(),
                        metric.WithAttributes(
                                attribute.Bool("sandbox.resume", isResume),
                                attribute.Bool("fs_only", fsOnly),
                                attribute.Bool("success", createErr == nil),
                        ),
                )
        }()
```

bchalios: Good call — done in 35ff85035. `create.duration` now records on every exit with a `success` attribute (matching `pause.duration`), so create/resume error rate is queryable too, and pause/create are uniform.

One behavioral note: it now also records *failed* creates (previously success-only), so any existing create-latency panel that wants the old semantics should add `success="true"`.

bchalios: Applied your snippet exactly in `35ff85035` — `create.duration` now records on every exit with `sandbox.resume` + `fs_only` + `success`, uniform with `pause.duration`, so create/resume error-rate is queryable too.

One behavioral change to flag: it now records **failed** creates as well (previously success-only via an early `return`), so any existing create-latency panel that wants the old semantics should filter `success="true"`.

Reopened the thread so it's yours to close — lmk if the semantics look right (and whether you'd like the `success="true"` filter added to the current create.duration dashboard panels as part of the follow-up).

kalyazin (approved):

PR #3433, opened 2026-07-29
pass placeholder for byoc clusters

djeebus:


packages/api/internal/handlers/volume_create.go
```diff
@@ -174,8 +175,12 @@ func (a *APIStore) PostVolumes(c *gin.Context) {
 	c.JSON(http.StatusCreated, result)
 }
 
-func (a *APIStore) getVolumeType(ctx context.Context) string {
+func (a *APIStore) getVolumeType(ctx context.Context, team *types.Team) string {
 	volumeType := a.featureFlags.StringFlag(ctx, featureflags.DefaultPersistentVolumeType)
+	if volumeType == "" && team != nil && team.ClusterID != nil {
```
dobrac: can the team be ever nil? Also, isn't our default clusterID uuid.Empty? 

Either, don't we have a helper for checking the BYOC cluster?

djeebus: No idea. Not worth the potential panic to avoid the 10 or so characters though.

packages/api/internal/handlers/volume_create.go
```diff
@@ -203,6 +213,10 @@ func (a *APIStore) getVolumeType(ctx context.Context, team *types.Team) string {
 		return volumeType
 	}
 
+	if team != nil && team.ClusterID != nil {
+		return a.config.PlaceholderPersistentVolumeType
+	}
```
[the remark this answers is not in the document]
djeebus: That's acceptable for now. Currently byoc clusters can have exactly zero persistent volumes; this PR gets them exactly one, as long as its available on all nodes.

packages/db/queries/volumes/volumes.sql
```diff
@@ -1,6 +1,6 @@
 -- name: CreateVolume :one
-INSERT INTO volumes (team_id, volume_type, name)
-VALUES (@team_id, @volume_type, @name)
+INSERT INTO volumes (id, team_id, volume_type, name)
+VALUES (@id, @team_id, @volume_type, @name)
```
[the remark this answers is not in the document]
djeebus: There's a trade-off here: if we create the row first, then we have to update it afterwards if/when the volume is created. If we create the volume first, then we need to generate an ID for the creation.

packages/api/internal/cfg/model.go
```diff
@@ -110,7 +110,8 @@ type Config struct {
 
 	AuthProvider sharedauth.ProviderConfig `env:"AUTH_PROVIDER_CONFIG"`
 
-	DefaultPersistentVolumeType string `env:"DEFAULT_PERSISTENT_VOLUME_TYPE"`
+	DefaultPersistentVolumeType     string `env:"DEFAULT_PERSISTENT_VOLUME_TYPE"`
+	PlaceholderPersistentVolumeType string `env:"PLACEHOLDER_PERSISTENT_VOLUME_TYPE" envDefault:"REPLACEME"`
```
dobrac: I would change this to something sane by default

packages/api/internal/handlers/volume_create.go
```diff
@@ -203,6 +205,10 @@ func (a *APIStore) getVolumeType(ctx context.Context, team *types.Team) string {
 		return volumeType
 	}
 
+	if team != nil && team.ClusterID != nil {
```
dobrac, line 208: do we have any helper for this check to make it obvious?

dobrac (approved):

packages/api/internal/handlers/volume_create.go
djeebus: No, we only have one that finds a Cluster ID and resolves a pointer to a zero uuid

packages/api/internal/handlers/volume_create.go
```diff
@@ -89,38 +89,19 @@ func (a *APIStore) PostVolumes(c *gin.Context) {
...
-	case err != nil:
-		a.sendAPIStoreError(c, http.StatusInternalServerError, "Error when creating volume")
-		telemetry.ReportCriticalError(ctx, "error when creating volume", err)
-
-		return
-	default:
 	}
 
-	if err := a.createVolume(ctx, clusterID, volume); err != nil {
+	response, err := a.createVolume(ctx, clusterID, volume)
```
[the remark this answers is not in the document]
djeebus: This is true, but we're accepting this as a limitation for now.

packages/orchestrator/pkg/volumes/volume_create.go
```diff
@@ -33,5 +36,11 @@ func (s *Service) CreateVolume(ctx context.Context, request *orchestrator.Create
...
 	}
 
-	return &orchestrator.CreateVolumeResponse{}, nil
+	response := &orchestrator.CreateVolumeResponse{}
+
+	if vt := request.GetVolume().GetVolumeType(); vt != "" {
+		response.VolumeType = &vt
+	}
+
+	return response, nil
```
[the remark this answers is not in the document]
djeebus: The comment is awkward, it's not orchestrator but middleware between api and orchestrator that fixes it. This is fine.

PR #3435, opened 2026-07-29
feat(api): per-region volume type defaults from node-derived region

rguliyev:
Resolve a new volume's type from `DEFAULT_PERSISTENT_VOLUME_TYPE_BY_REGION`, keyed by the single `region=` node label of the ready nodes the team's scheduling labels select — teams stay untouched; Terraform auto-appends the label to every client cluster. Node labels answer *where* the team runs; the map is the only policy source for *what* a volume there should be, so a region mounting several types (e.g. a new default introduced next to a legacy one) never needs runtime guessing.

Precedence: LaunchDarkly override (team-targetable) → region map → deployment-wide default. Zero or multiple matched regions fail open to the default.

Pairs with e2b-dev/terraform#873, which emits the map by deriving it at plan time: a region's single type resolves automatically, the deployment default wins when mounted in a multi-type region, and the plan fails when a multi-type region has no explicit `default_persistent_volume_type_by_region` entry.

### DIFF
--- a/packages/api/internal/handlers/volume_create.go
+++ b/packages/api/internal/handlers/volume_create.go
@@ -6,12 +6,14 @@ import (
 	"fmt"
 	"net/http"
 	"regexp"
+	"strings"
 
 	"github.com/gin-gonic/gin"
 	"github.com/google/uuid"
 
 	"github.com/e2b-dev/infra/packages/api/internal/api"
 	"github.com/e2b-dev/infra/packages/api/internal/clusters"
+	"github.com/e2b-dev/infra/packages/auth/pkg/types"
 	"github.com/e2b-dev/infra/packages/db/pkg/dberrors"
 	"github.com/e2b-dev/infra/packages/db/queries"
 	clustershared "github.com/e2b-dev/infra/packages/shared/pkg/clusters"
@@ -72,9 +74,12 @@ func (a *APIStore) PostVolumes(c *gin.Context) {
 		return
 	}
 
-	ctx = featureflags.AddToContext(ctx, featureflags.VolumeContext(body.Name))
+	ctx = featureflags.AddToContext(ctx,
+		featureflags.VolumeContext(body.Name),
+		featureflags.TeamContext(team.ID.String()),
+	)
 
-	volumeType := a.getVolumeType(ctx)
+	volumeType := a.getVolumeType(ctx, team)
 	if volumeType == "" {
 		a.sendAPIStoreError(c, http.StatusInternalServerError, "No persistent volume type is configured")
 		telemetry.ReportCriticalError(ctx, "default persistent volume type is not configured", nil)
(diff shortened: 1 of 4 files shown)


ValentaTomas (approved):

packages/api/internal/handlers/volume_create.go
```diff
@@ -174,13 +184,45 @@ func (a *APIStore) PostVolumes(c *gin.Context) {
...
+		if !ok {
+			continue
+		}
+
+		if matched && volumeType != matchedType {
+			return "", fmt.Errorf("%w: label %q maps to %q, label %q maps to %q",
+				errConflictingVolumeTypeSchedulingLabels, matchedLabel, matchedType, label, volumeType)
+		}
+
+		matched, matchedLabel, matchedType = true, label, volumeType
```
djeebus: This feels too complicated, and probably implies that we're doing something that's more complicated than it should be. Maybe we can default by region rather than node label?

packages/api/internal/handlers/volume_create.go
```diff
@@ -174,13 +184,84 @@ func (a *APIStore) PostVolumes(c *gin.Context) {
...
+	// mounted on the team's nodes: it may not be usable there at all.
+	if len(mounted) > 1 && !slices.Contains(mounted, a.config.DefaultPersistentVolumeType) {
+		logger.L().Warn(ctx, "team's nodes mount several volume types, none of them the default one",
+			zap.String("team_id", team.ID.String()),
+			zap.Strings("required_labels", requiredLabels),
+			zap.Strings("volume_types", mounted),
+		)
+	}
+
+	return a.config.DefaultPersistentVolumeType
```
djeebus: I think this only works with a single mounted volume type. as soon as we upgrade to a new one (for example, what we did in foxtrot/us-west1), this returns the default, which is probably wrong.

for example, let's say that in us-west1 we default users to FANCYVOLUME, but we have SLOWVOLUME attached to nodes in us-west3. a month later, we find out we can do better, and create a FASTVOLUME in us-west3. we still need SLOWVOLUME mounted, so now we have two mounts in us-west3, with FASTVOLUME as the new default. mounted will now contain two items, which will log a warning and return FANCYVOLUME, which is definitely wrong

I think we need to track volume types, not by label, but by region, and have defaults per region. that should keep everything sane

rguliyev: Agreed — node labels are topology (what a pool *can* mount), a default is policy (what a new volume *should* be), and they only coincide while a region mounts one type. Reworked to your suggestion in 961e429cc: defaults are region-keyed (`DEFAULT_PERSISTENT_VOLUME_TYPE_BY_REGION`), and the team's region is resolved from the nodes' `region=` labels, so team scheduling labels stay untouched.

Walking your scenario: us-west3 nodes advertise `region=us-west3` no matter how many types they mount; the map says `us-west3 → FASTVOLUME`; SLOWVOLUME stays mounted for existing volumes; FANCYVOLUME can't leak in because the answer comes from the map, never from what happens to be mounted.

The map isn't hand-maintained either — the paired terraform change (e2b-dev/terraform#873, 92ca664) derives it at plan time: a region with a single type resolves automatically, a multi-type region resolves to the deployment default when that's one of its types (today's foxtrot state — stays green), and otherwise the **plan fails** until someone names the new default explicitly. So the FASTVOLUME decision gets made in review at the moment it first exists, not guessed at runtime.

PR #3437, opened 2026-07-29
test(orch): per-distro template-build coverage

tomassrnka:
Adds a unit test that holds every distro profile's package set to a shared capability map, and integration tests that build a template from an Ubuntu, Fedora, Arch and Alpine base image plus one unsupported image that must be rejected with a readable reason. Stacked on #3411 — retarget to main once that merges. Heads-up on cost: the first run pulls and provisions four new base images, so the templates package gets slower until the base-layer cache warms.


tomassrnka:
bugbot run

jakubno (approved):

PR #3459, opened 2026-07-30
fix(orch): fall back to ID_LIKE with a warning instead of rejecting

tomassrnka:
Switching provisioning to the declared os-release id dropped every Debian derivative that used to work when we probed for a package manager — Kali declares `ID=kali ID_LIKE=debian`. Unknown ids now retry each `ID_LIKE` token and warn that the profile is a best-effort guess.

`rhel`/`ol`/`amzn` are guarded **before** the fallback — both Oracle and Amazon Linux declare `ID_LIKE=fedora` and would otherwise be silently re-admitted past the rejection the rhel profile documents.

Behavior after this change:
- Known id → unchanged.
- Unknown id with a matching `ID_LIKE` token → provisions as that family, with a customer-visible `WARNING` that it's best effort and untested.
- `rhel`/`ol`/`amzn` → still rejected fast with the kernel/kABI reason.
- Id matching nothing, or no `/etc/os-release` → still exits 1.

Covered by `TestUnknownIDFallsBackToIDLike` and `TestRejectedIDsAreNotReachableViaIDLike` (guard must precede the fallback).

### DIFF
--- a/packages/orchestrator/pkg/template/build/phases/base/distro/distro.go
+++ b/packages/orchestrator/pkg/template/build/phases/base/distro/distro.go
@@ -201,7 +201,11 @@ func ShellSelector() string {
 	// Selection lives in a function so it can be retried per ID_LIKE token.
 	// Assignments and function definitions inside a POSIX-sh function are
 	// global, so the caller sees the profile the same way it always has.
+	// The match is reported via e2b_profile_matched, not the return status —
+	// a function called as an if-condition runs with errexit suppressed,
+	// which would swallow Bootstrap failures inside a matched arm.
 	b.WriteString("e2b_select_profile() {\n")
+	b.WriteString("  e2b_profile_matched=\n")
 	b.WriteString(`  case "$1" in` + "\n")
 	for _, p := range Profiles {
 		fmt.Fprintf(&b, "  %s)\n", strings.Join(p.IDs, "|"))
@@ -219,13 +223,14 @@ func ShellSelector() string {
 		fmt.Fprintf(&b, "    e2b_ca_refresh() { %s; }\n", p.CARefresh)
 		fmt.Fprintf(&b, "    E2B_INIT_SYSTEM=%q\n", p.Init)
 		fmt.Fprintf(&b, "    e2b_init_setup() {\n%s\n    }\n", indentBlock(initSetup[p.Init], "        "))
-		fmt.Fprintf(&b, "    return 0\n")
+		fmt.Fprintf(&b, "    e2b_profile_matched=1\n")
 		fmt.Fprintf(&b, "    ;;\n")
 	}
-	fmt.Fprintf(&b, "  *)\n    return 1\n    ;;\n")
+	fmt.Fprintf(&b, "  *)\n    ;;\n")
 	b.WriteString("  esac\n}\n\n")
 
-	b.WriteString(`if ! e2b_select_profile "$E2B_DISTRO_ID"; then` + "\n")
+	b.WriteString(`e2b_select_profile "$E2B_DISTRO_ID"` + "\n")
+	b.WriteString(`if [ -z "$e2b_profile_matched" ]; then` + "\n")
 
 	// Deliberate rejections are checked before the fallback, so they keep
 	// failing fast with their own reason instead of being matched by ID_LIKE.
@@ -239,7 +244,8 @@ func ShellSelector() string {
 
 	b.WriteString("  e2b_like_match=\n")
 	b.WriteString(`  for e2b_like in $E2B_ID_LIKE; do` + "\n")
-	b.WriteString(`    if e2b_select_profile "$e2b_like"; then` + "\n")
+	b.WriteString(`    e2b_select_profile "$e2b_like"` + "\n")
+	b.WriteString(`    if [ -n "$e2b_profile_matched" ]; then` + "\n")
 	b.WriteString("      e2b_like_match=$e2b_like\n")
 	b.WriteString("      break\n")
 	b.WriteString("    fi\n")
--- a/packages/orchestrator/pkg/template/build/phases/base/distro/distro.go
+++ b/packages/orchestrator/pkg/template/build/phases/base/distro/distro.go
@@ -1,8 +1,9 @@
 // Package distro makes template-build provisioning distro-aware: it selects a
 // declared per-family Profile by the base image's /etc/os-release ID rather than
 // probing for a package manager. Supported: the systemd family (Debian/Ubuntu,
-// Fedora/RHEL/CentOS/Rocky/Alma, Arch) and Alpine on OpenRC; anything else is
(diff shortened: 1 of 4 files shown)


packages/orchestrator/pkg/template/build/phases/base/distro/distro.go
```diff
@@ -201,14 +219,40 @@ func ShellSelector() string {
...
 		fmt.Fprintf(&b, "    E2B_INIT_SYSTEM=%q\n", p.Init)
 		fmt.Fprintf(&b, "    e2b_init_setup() {\n%s\n    }\n", indentBlock(initSetup[p.Init], "        "))
+		fmt.Fprintf(&b, "    return 0\n")
 		fmt.Fprintf(&b, "    ;;\n")
 	}
-	fmt.Fprintf(&b, "  *)\n")
+	fmt.Fprintf(&b, "  *)\n    return 1\n    ;;\n")
+	b.WriteString("  esac\n}\n\n")
+
+	b.WriteString(`if ! e2b_select_profile "$E2B_DISTRO_ID"; then` + "\n")
```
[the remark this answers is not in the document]
tomassrnka: Fixed in a6ae351d2 — the match is reported via `e2b_profile_matched` and `e2b_select_profile` is invoked as a plain statement, so errexit stays live inside profile arms (a failing Bootstrap now aborts at the failing command). A unit test rejects any `if e2b_select_profile` invocation to keep it that way.

packages/orchestrator/pkg/template/build/phases/base/distro/distro.go
```diff
@@ -201,14 +219,40 @@ func ShellSelector() string {
...
+	b.WriteString("  e2b_like_match=\n")
+	b.WriteString(`  for e2b_like in $E2B_ID_LIKE; do` + "\n")
+	b.WriteString(`    if e2b_select_profile "$e2b_like"; then` + "\n")
+	b.WriteString("      e2b_like_match=$e2b_like\n")
+	b.WriteString("      break\n")
+	b.WriteString("    fi\n")
+	b.WriteString("  done\n")
+
+	b.WriteString(`  if [ -z "$e2b_like_match" ]; then` + "\n")
 	fmt.Fprintf(&b, "    echo \"[provision] ERROR: unsupported base image distribution: ID='${E2B_DISTRO_ID:-unknown}'.\" >&2\n")
```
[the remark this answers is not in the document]
tomassrnka: Fixed in a6ae351d2 — the match is reported via `e2b_profile_matched` and `e2b_select_profile` is invoked as a plain statement, so errexit stays live inside profile arms (a failing Bootstrap now aborts at the failing command). A unit test rejects any `if e2b_select_profile` invocation to keep it that way.

packages/orchestrator/pkg/template/build/phases/base/distro/distro.go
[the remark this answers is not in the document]
tomassrnka: Fixed in a6ae351d2 — the match is reported via `e2b_profile_matched` and `e2b_select_profile` is invoked as a plain statement, so errexit stays live inside profile arms (a failing Bootstrap now aborts at the failing command). A unit test rejects any `if e2b_select_profile` invocation to keep it that way.

packages/orchestrator/pkg/template/build/phases/base/distro/distro.go
arkamar:
```suggestion
// Fedora/RHEL/CentOS/Rocky/Alma, Arch) and Alpine on OpenRC. Derivatives are provisioned
// via ID_LIKE as a best-effort guess with a warning.
```
This PR makes the package description stale. Not sure if the suggestion will work here.

packages/orchestrator/pkg/template/build/phases/base/distro/distro.go
```diff
@@ -178,13 +178,35 @@ func SupportedIDs() []string {
...
-// unrecognized id exits 1 with a customer-visible error.
+// functions, init path, time-sync unit, admin group and CA handling.
+//
+// An id we don't know falls back to $E2B_ID_LIKE, the derivative-to-parent
+// pointer from os-release: Kali declares ID=kali ID_LIKE=debian, and before
+// provisioning switched on the declared id, every such Debian derivative worked
+// by accident because we probed for a package manager instead. That fallback
+// warns rather than fails — the profile is a best-effort guess at that point —
+// but an id matching nothing, and an image with no os-release at all, still
+// exits 1 rather than being provisioned against a guessed family.
```
arkamar: This `ShellSelector` doc comment now reads more like a changelog rather than an API contract.

packages/orchestrator/pkg/template/build/phases/base/distro/distro.go
```diff
@@ -201,14 +223,42 @@ func ShellSelector() string {
...
-	fmt.Fprintf(&b, "  *)\n")
+	fmt.Fprintf(&b, "  *)\n    ;;\n")
+	b.WriteString("  esac\n}\n\n")
+
+	b.WriteString(`e2b_select_profile "$E2B_DISTRO_ID"` + "\n")
+	b.WriteString(`if [ -z "$e2b_profile_matched" ]; then` + "\n")
+
+	// Deliberate rejections are checked before the fallback, so they keep
+	// failing fast with their own reason instead of being matched by ID_LIKE.
+	fmt.Fprintf(&b, "  case \"$E2B_DISTRO_ID\" in\n")
```
arkamar, line 234: The guard matches only the exact $E2B_DISTRO_ID, but the ID_LIKE loop below never checks its tokens against RejectedIDs, so ID=cloudlinux ID_LIKE="rhel fedora" skips past rhel and provisions as the rhel family with just a warning.

arkamar (changes requested):
few nits

packages/orchestrator/pkg/template/build/phases/base/distro/distro.go
tomassrnka: Fixed in 8e74ec33c — took your suggestion, plus a closing clause so the rejection behavior stays documented: kernel-dependent ids (RejectedIDs) and anything unmatched are rejected with a clear error.

packages/orchestrator/pkg/template/build/phases/base/distro/distro.go
tomassrnka: Fair — rewritten in 8e74ec33c to state the contract only (unknown ids retry ID_LIKE tokens in order, best effort with a warning; RejectedIDs / no match / no os-release exit 1). The Kali history lives in the commit message.

packages/orchestrator/pkg/template/build/phases/base/distro/distro.go
tomassrnka: i'm ok with the pass here

arkamar (approved):

PR #3483, opened 2026-07-31
feat(snapshots): in place resume & fs-only snapshots

bchalios:
TBD


bchalios:
Closing as superseded. The feature shipped into `main` via the private source repo's one-way export rather than from this branch: `resolveFilesystemBoot` is in `packages/api/internal/handlers/sandbox_connect.go` and `sandbox_resume.go` with `sandbox_resume_test.go` alongside, and the `memory` field is in `spec/openapi.yml`.

Worth noting for anyone reading the diff stat: the +21k/-2k across 100 files is not this feature. Only about a third of the files touch it — the bulk is `dashboard-api` (26), `db` (25), `auth` (16), clickhouse and a NixOS base-image workflow, which is `main`'s own movement showing through because the branch base is six weeks old. Nothing to salvage there.

Branch left in place.

PR #3493, opened 2026-08-01
fix(orchestrator): treat a failed or signalled e2fsck as a failed integrity check

tomassrnka:
Linear: https://linear.app/e2b/issue/EN-1697

**Broken:** `filesystem.CheckIntegrity` could return nil for a rootfs that was never actually checked — a corrupted filesystem passed the integrity gate used by the OCI, base-provision, rootfs and resize-disk build phases.

**Root cause:** the guard `exitCode > accExitCode` misses `ExitCode() == -1` (e2fsck never started, or killed by a signal — OOM, context cancel), so the error was swallowed. The ordered `>` compare also misread e2fsck's bitmask statuses: status 3 (`1|2`, corrected + reboot) failed even though both bits are individually accepted.

**Fix:** treat -1 (not started / signalled) as a failed check; accept a status only when every set bit is accepted.

**Repro:** new `TestCheckIntegrityExitStatus` (fake `e2fsck` on PATH) — 3 of 9 cases fail on unmodified main (signal → nil error, start failure → nil error, status 3 spuriously rejected); the other 6 pin existing behaviour.

**Verification:** all 9 cases + full package suite pass; `gofmt`/`go vet` clean; `go build ./...` on `packages/orchestrator` OK. Run in a `golang:1.26` linux container (package is `//go:build linux`).

Note: the Linear issue predicts a nil-pointer panic — not reachable (`(*os.ProcessState).ExitCode()` guards a nil receiver, returns -1); silent success was the actual defect, and the nil case is still handled explicitly.

### DIFF
--- a/packages/orchestrator/pkg/template/build/core/filesystem/ext4.go
+++ b/packages/orchestrator/pkg/template/build/core/filesystem/ext4.go
@@ -40,6 +40,15 @@ const (
 	reservedBlocksPercentage = int64(0)
 )
 
+// e2fsck reports its outcome as a bitmask, so several conditions can be
+// combined in a single exit status. See the "EXIT CODE" section of e2fsck(8);
+// the statuses not named here (4 uncorrected errors, 8 operational error,
+// 16 usage error, 32 canceled, 128 shared-library error) are always failures.
+const (
+	e2fsckErrorsCorrected   = 1
+	e2fsckRebootRecommended = 2
+)
+
 func Make(ctx context.Context, rootfsPath string, sizeMb int64, blockSize int64) error {
 	ctx, tuneSpan := tracer.Start(ctx, "make-ext4")
 	defer tuneSpan.End()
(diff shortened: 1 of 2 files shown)


tomassrnka:
Supporting evidence found while self-reviewing: `ReplayJournal`, ~15 lines below `CheckIntegrity` in the same file, already encodes exactly the semantics this PR adds:

```go
if !errors.As(err, &exitErr) || exitErr.ExitCode() < 0 || exitErr.ExitCode()&^(1|2) != 0 {
    return string(output), fmt.Errorf("error replaying ext4 journal: %w\n%s", err, output)
}
```

Same three rules: exec failure is an error, a negative (signalled) status is an error, and the accepted statuses are matched as the `1|2` bitmask rather than by ordering. So `CheckIntegrity` was the odd one out rather than the deliberate exception, and this change brings it in line with the convention already used for the other e2fsck invocation in the package.

That also addresses the "could this newly fail builds where e2fsck is unavailable?" question: `ReplayJournal` already hard-fails in exactly that situation, so a missing or killed `e2fsck` is already treated as fatal on this path today.

I kept the `cmd.ProcessState` form rather than copying the `errors.As(&exitErr)` idiom because the exit status is also needed for the `filesystem.e2fsck.exit_code` span attribute; happy to switch to the sibling's idiom if you prefer them identical.

tomassrnka:
Blast-radius audit, since this changes a gate rather than a log line. All six `CheckIntegrity` call sites already treat a returned error as a build failure, so the newly surfaced failures fail the build instead of silently shipping an unchecked rootfs:

- `core/rootfs/rootfs.go:224` → wraps and returns
- `core/oci/oci.go:217` and `:229` → wrap and return
- `phases/ensurefreedisk/grow.go:241` / `:247` → `pre-resize e2fsck` / `post-resize e2fsck`
- `phases/base/builder.go:240` → wraps and returns
- `phases/base/provision.go:249` → calls with `fix=false`, and on error retries with `fix=true`; that retry's error is checked and returned

The only behaviour change on a *healthy* filesystem is that exit status 3 (`1|2`) is now accepted instead of failing the build, which is strictly fewer false failures. The stricter direction is limited to the two cases where e2fsck produced no verdict at all — never started, or killed by a signal — which previously returned `nil`.

PR #3479, opened 2026-07-31
ci: rebase #3235 onto main + shard the compressed allow-list runs

tomassrnka:
## What

Rebase + extension of #3235 (cc @dobrac), squashed onto current main after #3462 landed. Opened as a **draft for a measuring run** — the shard timings from this PR's CI decide the final shard balance.

Two things changed since #3235 was written, and this branch reconciles both:

- #3462 already landed the coverage philosophy (uncompressed runs the whole suite once; zstd1/lz4 only re-run the allow-list). This rebase **keeps that decision** and applies #3235's sharding on top — PRs now shard the *uncompressed* config, not zstd1.
- The distro-family template tests (#3437/#3444) landed into what is now templates shard 2, so the name-prefix split needs rebalancing from this run's junits.

## PR-path matrix (8 jobs; push to main unchanged: 3 unsharded configs)

| Job | Shard | Runs |
|---|---|---|
| uncompressed-templates-1/2 | `^TestTemplateBuild` split by name prefix | 11 + 9 real-build tests |
| uncompressed-sandboxes | sandboxes + metrics + volumes + proxies | 121 tests |
| uncompressed-rest | everything else, `-skip '^TestTemplateBuild'` | 86 tests |
| zstd1-templates / lz4-templates | templates package ∩ allow-list | 6 snapshot-build entries |
| zstd1-other / lz4-other | package complement ∩ allow-list | 33 entries |

Static partition verified: 227 top-level tests land exactly once across the uncompressed shards; templates/no-templates is an exact package complement. `TESTS_ONLY` composes only with package-pure shards (the name-split shards' own `-run` would collide — guarded, fails loudly). Unknown shard names fail loudly instead of silently running the whole suite.

## Carried from #3235 (unchanged in spirit)

- DB containers start right after checkout and overlap the Go builds (`start-databases`); `start-services` waits with bounded health loops. Docker Hub pre-pull retries (#3410) folded in, now covering the pinned tags + otel.
- ClickHouse migrations via goose on the host (`migrate-host`) — no migrator image build.
- Salted optimized Go cache; `DEBUG_GCFLAGS` opt-out (race stays on, `-N -l` off in CI). envd keeps the musl-static link from #3444, gcflags parametrized.
- 7 metadata/authz template tests use `RequestTemplateWithoutBuild`; envd process tests poll instead of fixed sleeps.

## Expected (to be confirmed by this run)

[...]

PR #3499, opened 2026-08-01
ci: allow-list completeness lint + two-phase service startup

tomassrnka:
Two follow-ups from #3479, one commit each (+ review fixes).

**Allow-list completeness lint.** `select-tests.sh` rejects stale allow-list entries; nothing caught the missing direction — a new snapshot-path test nobody lists silently never runs compressed. `check-allowlist-completeness.sh` (wired into the existing `check-tests-allowlist` CI step) requires every test touching a snapshot-lifecycle symbol to be listed or carry `// compression-tests:excluded <reason>`. Package-local wrappers must join its symbol list, so tests can't hide behind helpers. Running it on main found 6 rotted-out tests (cross-team fork/resume/connect, envd CA-cert fs-only reboot) — added; 15 fixture-only tests got exclusion markers. Grey call for review: the two `TestEgressFirewall*PersistsAfterResume` are marked excluded per the tsv's egress exclusion, but read arguably as snapshot-integrity — flip if you disagree.

**Two-phase service startup.** `start-service.sh` gains `start`/`wait` subcommands (legacy form intact) and prints the log tail on health-wait timeout (was diagnostics-free). Boot order: otel ∥ orchestrator, then API ∥ client-proxy. Measured honestly: **no wall-time win** (~2s; the step is dominated by migrations/seed) — the value is the timeout diagnostics; happy to drop the reordering if you prefer minimal diff.

Verified: partition proof 227 tests exactly once; lint red/green paths CI-tested; marker form survives the repo formatter.

### DIFF
--- a/.github/actions/start-services/action.yml
+++ b/.github/actions/start-services/action.yml
@@ -121,21 +121,15 @@ runs:
         PERSISTENT_VOLUME_MOUNTS="test-volume-type:${test_volume_dir}"
         export PERSISTENT_VOLUME_MOUNTS
 
-        # The backgrounded warm-up pull in start-databases discards its exit
-        # status; this synchronous call is the actual guarantee (near-instant
-        # when the warm-up succeeded) so a rate-limited pull retries instead
-        # of failing the collector's 30s health window. The tag comes from
-        # the Makefile that runs the image.
+        # The warm-up pull in start-databases discards its exit status; this
+        # synchronous call is the guarantee (near-instant when warm). Tag
+        # comes from the Makefile that runs the image, so bumps can't strand it.
         OTEL_IMAGE=$(grep -oE 'otel/opentelemetry-collector-contrib:[0-9.]+' packages/otel-collector/Makefile | head -1)
         ./scripts/pull-retry.sh "$OTEL_IMAGE"
 
-        # Boot in two concurrent phases instead of four serial start+wait
-        # chains. Phase 1: otel-collector and orchestrator boot together -
-        # neither depends on the other (the services' OTLP exporters retry
-        # until the collector is up). Phase 2: API and client-proxy (needed
-        # for auto-resume tests) boot together once the orchestrator is
-        # healthy, preserving the orchestrator-before-API ordering the
-        # serial flow guaranteed.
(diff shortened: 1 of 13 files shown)


tomassrnka:
## Measured results (run [30693622934](https://github.com/e2b-dev/infra/actions/runs/30693622934), all green)

**Commit 1 (allow-list lint): validated.** The lint runs in every integration shard via `check-tests-allowlist`, passes on the fixed tree, and fails correctly when a marker is removed. One interaction found the hard way: the repo formatter rewrites the marker to `// compression-tests:excluded` (hyphen ≠ machine-directive pattern), so the lint accepts the formatted form — that's the third commit.

**Commit 2 (two-phase startup): no measured wall win — being honest about it.** Start Services came in at 40–46s vs 42–48s baseline (~2s, within noise). The step's cost is dominated by DB health-wait + migrations + seed, not the four service boots — prebuilt binaries turn healthy in 1–2 polls, so the serial waits I parallelized were already near-free. What the commit still delivers: the log tail on health-wait timeout (previously a timeout produced zero diagnostics) and the `start`/`wait` split for anyone restructuring later. If you'd rather keep the diff minimal, I can drop the two-phase reordering and keep only the script's subcommands + failure diagnostics — say the word.

**Unrelated to this PR but visible in the run**: shards queued 2.1–6.2 min because three pushes in ~10 min stacked runs against the runner group's 24-slot ceiling (wall 17.3 min despite healthy shard durations). Same story as the #3479 measurements — the max-concurrency raise is worth doing.

tomassrnka:
Both bot findings were real and are fixed in 0036a31f1: receiver-method name mangling + missing function-end reset (claude), and multiline/variable-arg autopause forms (codex). Each was reproduced with a probe file before fixing; the probe now trips all three paths and the tree is green without it.

tomassrnka:
cursor's Low was right — my variable-arg alternative matched the `f` in `false`. Fixed in 97937edae: the scan now strips literal `WithAutoPause(false)`/`WithAutoResume(false)` from a line before matching, so any other argument shape flags and only the explicit opt-out stays quiet. Probe-verified: false-only test quiet, true/variable-arg tests flagged.

PR #3504, opened 2026-08-01
fix(orchestrator): report the current network egress when listing sandboxes

tomassrnka:
### What was broken

Updating a sandbox's network via the API propagates correctly to the orchestrator, but the change was invisible to the orchestrator's gRPC sandbox list. Two consequences, both in the ticket:

- the sandbox detail API served outdated network information
- the API re-syncs its store from this list on redeploy, so an updated sandbox silently reverted to its create-time egress — and the next pause/resume rebuilt it from that stale config, losing the update for real

### Root cause

`List` returned `sbx.APIStoredConfig` verbatim. That field is a create-time snapshot (its own doc comment marks it deprecated, "used to store the config to allow API restarts"); it is assigned exactly twice — create and resume — and never updated.

The update path writes somewhere else entirely: `Update` → `transitionEgress` → `applyNetworkEgress` → `Config.SetNetworkEgress`, which sets `sandbox.Config.Network.Egress`. The two are provably distinct objects, not aliases: create does `network := proto.CloneOf(req.GetSandbox().GetNetwork())` for `sandbox.Config` while storing `req.GetSandbox()` as `APIStoredConfig`.

### Fix

`List` overlays the live egress onto the stored config. The stored config is shared and handed out by reference, so it is **cloned** before the overlay, and only when the two actually diverge — the unchanged case returns the same pointer as before, no allocation.

Fixed in `List` rather than in `applyNetworkEgress` deliberately: that function only receives `*sandbox.Config`, not the `*Sandbox`, and `Config.mu` does not protect `APIStoredConfig`, so a write-through would add an unsynchronised mutation of a pointer `List` hands out.

`allow_internet_access` is corrected in the same clone, narrowly. The API folds `false` into a `0.0.0.0/0` denied CIDR on create *and on every resume*, so a stale `false` would re-apply the deny-all and undo an update that re-enabled the internet — that is the "settings lost after pause/resume" half of the ticket. Only one case is corrected: stored `false` where the live egress no longer denies everything. The converse needs nothing (the deny-all rides in the denied CIDRs themselves), and `nil` stays `nil` because unset means "never explicitly set" in the public API.

`Checkpoint` was considered and deliberately left alone: it passes `sbx.Config` *and* `APIStoredConfig` to `ResumeSandbox`, which stores the same `*sandbox.Config` pointer on the resumed sandbox, so the live egress is already carried and `List` overlays it there too. Applying the helper there is provably a no-op.

### How verified

New `Test_server_List_ReportsCurrentEgress` in `sandboxes_test.go`, driving state directly through `sbx.Config.SetNetworkEgress` (not through `s.Update()` — `transitionEgress` always rolls back without a real netns). Before the fix, 3 of 6 subtests fail and 3 pass:

```
--- FAIL: .../reports_the_updated_egress_without_mutating_the_stored_config
        expected: []string{"9.9.9.9/32"}   actual: []string{"1.0.0.0/8"}
--- FAIL: .../reports_egress_cleared_back_to_nil
        Expected nil, but got: &SandboxNetworkEgressConfig{AllowedCidrs:["1.0.0.0/8"]}
--- FAIL: .../clears_a_stale_allow_internet_access=false_once_the_deny-all_is_gone
--- PASS: .../returns_the_stored_config_untouched_when_nothing_changed
--- PASS: .../keeps_allow_internet_access=false_while_the_deny-all_stands
--- PASS: .../leaves_an_unset_allow_internet_access_unset
```

The three passing ones guard against over-reach; the pre-existing `Test_server_List` passes throughout. All pass after.

Full `./pkg/server/...` suite green under `-race` — 25 tests, 0 skips. `go build`, `go vet`, `gofmt` clean; `golangci-lint` v2.11.4 (repo-pinned, installed in-container) reports 0 issues.

Not verified: run on linux/**arm64** rather than amd64 (local Docker constraint); no arch-specific code is involved. The end-to-end `Update → List` path is not exercised by a unit test for the rollback reason above — the link between them is `sbx.Config`, whose write side `sandboxes_update_test.go` already covers. The API-side consequence is reasoned from `create_instance.go` / `update_network.go`, not covered by a `packages/api` test.

Linear: https://linear.app/e2b/issue/EN-717/fix-sandbox-list-in-orch-grpc-to-return-current-network-setup

### DIFF
--- a/packages/orchestrator/pkg/server/sandboxes.go
+++ b/packages/orchestrator/pkg/server/sandboxes.go
@@ -36,6 +36,7 @@ import (
 	"github.com/e2b-dev/infra/packages/shared/pkg/logger"
 	sbxlogger "github.com/e2b-dev/infra/packages/shared/pkg/logger/sandbox"
 	"github.com/e2b-dev/infra/packages/shared/pkg/retry"
+	sandbox_network "github.com/e2b-dev/infra/packages/shared/pkg/sandbox-network"
 	"github.com/e2b-dev/infra/packages/shared/pkg/storage"
 	"github.com/e2b-dev/infra/packages/shared/pkg/telemetry"
 	"github.com/e2b-dev/infra/packages/shared/pkg/utils"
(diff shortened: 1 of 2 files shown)


jakubno:
I would close this in favor of #3519

PR #3510, opened 2026-08-01
fix(aws): stop s3fs caching missing FC/kernel/envd objects

tomassrnka:
Linear: https://linear.app/e2b/issue/EN-386

**Broken:** on AWS BYOC hosts, an FC version (or kernel/envd build) uploaded to S3 after a host had already tried to read it stayed invisible on that host — sandboxes kept failing; only known workaround was unmount → clean dir → remount.

**Root cause:** all four s3fs mounts in `start-client.sh` pass `-o enable_noobj_cache`, which is a no-op — negative ("does not exist") caching is the s3fs default and the real toggle is `-o disable_noobj_cache`. Every missed lookup is remembered for `stat_cache_expire` (default 900 s), and these buckets are written out-of-band, so a cached miss outlives the upload by up to 15 minutes.

**Fix:** replace (not append — later flag wins) `enable_noobj_cache` with `disable_noobj_cache` on the four mounts.

**Repro (MinIO + s3fs 1.93, the Ubuntu 24.04 base-AMI version, exact production flags):** miss → upload → read: production flags not found for >100 s; flag removed, identical (confirms the no-op); `stat_cache_expire=20` visible after ~40 s (confirms the expiry is the bound); `disable_noobj_cache` visible after ~1 s. With no prior lookup the object is always visible immediately, isolating the negative cache as the cause.

**Verification:** the `disable_noobj_cache` run above is the passing case. `shellcheck -S warning` reports only the pre-existing SC2155 on line 43.

### DIFF
--- a/iac/provider-aws/modules/nodepool-client/scripts/start-client.sh
+++ b/iac/provider-aws/modules/nodepool-client/scripts/start-client.sh
@@ -75,24 +75,32 @@ modprobe nbd nbds_max=4096
 mkdir -p /fc-vm
 
 # Mount envd buckets
+#
+# disable_noobj_cache: s3fs caches "this object does not exist" answers for
+# stat_cache_expire (900s by default). These buckets are written out-of-band -
+# a new Firecracker/kernel/envd version is uploaded after a host has already
+# looked for it - so a cached miss makes the freshly uploaded object invisible
+# on that host until the entry expires or the mount is recreated.
+# Note that "enable_noobj_cache" is the s3fs default, so passing it was a no-op;
+# "disable_noobj_cache" is the option that actually turns the negative cache off.
 envd_dir="/fc-envd"
 mkdir -p $envd_dir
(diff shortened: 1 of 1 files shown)


iac/provider-aws/modules/nodepool-client/scripts/start-client.sh
```diff
@@ -75,24 +75,32 @@ modprobe nbd nbds_max=4096
...
 
 # Mount envd buckets
+#
+# disable_noobj_cache: s3fs caches "this object does not exist" answers for
+# stat_cache_expire (900s by default). These buckets are written out-of-band -
+# a new Firecracker/kernel/envd version is uploaded after a host has already
+# looked for it - so a cached miss makes the freshly uploaded object invisible
+# on that host until the entry expires or the mount is recreated.
+# Note that "enable_noobj_cache" is the s3fs default, so passing it was a no-op;
+# "disable_noobj_cache" is the option that actually turns the negative cache off.
```
sitole: Please remove slop comment

sitole (approved):

PR #3511, opened 2026-08-01
fix(envd): mount sandbox volumes with nolock so file locks don't hang

tomassrnka:
Linear: https://linear.app/e2b/issue/EN-1109/volume-file-locking-hangs-forever-nfsv3-nlm-never-answered

**Broken:** every file lock (`flock`, `fcntl`/`lockf`) on a Volume-backed path hung forever — blocking and non-blocking variants alike. Reads/writes/renames were fine, so `go build`, `cargo`, SQLite and zsh `compinit` silently wedged. Reported by a Volumes private-beta customer.

**Root cause:** volumes come from the orchestrator's userspace NFSv3 proxy, which serves no NLM — the portmapper answers `0` for nlockmgr (100021). The mount string set neither `nolock` nor `local_lock=`, so the kernel default `local_lock=none` forwarded every lock to NLM, and `hard` retried the bind forever.

**Repro:** real Linux NFSv3 client against the real proxy + portmapper (two containers). Before: `flock`/`fcntl`/`lockf`/`sqlite3` all exit 124 (timed out), task in state D at `rpc_wait_bit_killable` via `nfs_flock → nlmclnt_lock → rpc_call_sync`. Same mount + `nolock` (`local_lock=all`): every op exits 0 in ~2 ms, sqlite3 OK.

**Semantics trade-off (please agree explicitly):** `nolock` makes locks process-local to a single sandbox — two sandboxes mounting the same volume lose cross-sandbox `flock`/`fcntl` mutual exclusion. Today that case hangs (loud); after this it silently doesn't coordinate. The ticket accepts this ("acceptable per the existing volumes semantics"), but it is a real behaviour change. Within-sandbox locking still works — verified on the fixed mount: with a lock held, `flock -n` from a second process exits 1 and `flock -x` blocks 2.5 s until release.

**Rollout:** ships with an envd release; no runtime toggle (envd has no config surface), so already-running sandboxes keep the old behaviour until they get the new envd.

**Optional follow-ups (not done here):** make the portmapper return an error for 100021 instead of silently `0` (a recurrence would fail fast, not hang); the go-nfs fork returns `PROC_UNAVAIL` rather than `PROG_UNAVAIL` for unknown programs.

### DIFF
--- a/packages/envd/internal/api/init.go
+++ b/packages/envd/internal/api/init.go
@@ -359,13 +359,8 @@ var nfsOptions = strings.Join([]string{
 	"nfsvers=3",      // nfs proxy is nfs version 3
 	"noacl",          // no reason for acl in the sandbox
 
-	// Handle file locks locally instead of over NLM. The nfs proxy serves no
-	// nlockmgr (program 100021) — the portmapper only advertises nfs (100003)
-	// and mountd (100005), so a GETPORT for nlockmgr answers 0. Without this
-	// the kernel default (local_lock=none) sends every flock/fcntl to NLM,
-	// and "hard" makes it retry that bind forever, so any lock on a volume
-	// wedges in D state. Locks are therefore process-local to one sandbox,
-	// matching volume semantics: there is no cross-sandbox lock coordination.
+	// The nfs proxy serves no nlockmgr, so NLM lock calls wedge forever (D state)
+	// under "hard". Locks are sandbox-local: no cross-sandbox lock coordination.
 	"nolock",
 
 	// disable caching so that pause/resume works correctly
--- a/packages/envd/internal/api/init.go
+++ b/packages/envd/internal/api/init.go
@@ -359,6 +359,15 @@ var nfsOptions = strings.Join([]string{
 	"nfsvers=3",      // nfs proxy is nfs version 3
 	"noacl",          // no reason for acl in the sandbox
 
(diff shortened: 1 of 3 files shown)


tomassrnka:
Validated the Codecov "1 test failed" report (`TestWorkdirPermissionDenied`) — it is **not caused by this change**. Evidence:

1. **The shard is green; this was a first-attempt flake that passed on retry.** `tests/integration/Makefile:127,153` runs `go tool gotestsum --rerun-fails=1`, so the JUnit XML Codecov ingests records both the failed first attempt *and* the passing rerun. Every `integration-tests / run` shard reports `pass`.

2. **The changed code is unreachable in that test.** `nfsOptions` has exactly one consumer — `mountNFS` (`packages/envd/internal/api/init.go:471`) — which is only called from inside `for _, volume := range mounts` in `setupNFS`. `TestWorkdirPermissionDenied` creates a sandbox with no volume mounts and operates on `/tmp/restricted-workdir`, so no NFS mount is ever performed and the option string is never used.

3. **The failure signature is unrelated.** `invalid_argument: protocol error: incomplete envelope: unexpected EOF` is a truncated Connect RPC exec stream — a transport-level hiccup, not a filesystem or mount-option error. A `nolock` regression would surface as a lock hang, not a truncated envelope.

No code change made in response.

PR #3517, opened 2026-08-01
fix(orchestrator): wait for detached snapshot uploads before finishing the drain

tomassrnka:
Linear: https://linear.app/e2b/issue/EN-679

**Broken:** pausing the last sandbox on a draining orchestrator can lose the snapshot — the node declares itself drained and is terminated while the memfile/rootfs are still uploading.

**Root cause, two halves:** (1) the drain check is blind to uploads — `Pause` calls `MarkStopping` at its start and the sandbox lifecycle ends via `Sandbox.Close` during `Pause` itself, while the actual upload is detached into `uploadsWG` (retry budget up to 2 h); `DrainSandboxes` returned once `Sandboxes.Count() == 0` and the lifecycles map emptied — both true while bytes are still in flight — and `uploadsInFlight` was only consulted later in `Server.Close`, after the drain had already reported an empty node. (2) `orchestrator.hcl` set no `kill_timeout`, so Nomad's 5 s default applied — shorter than the drain's own 15 s status-propagation sleep, so the graceful drain added in #3069 never got to run. Every sibling job sets one (client-proxy 24h, ingress 24h, template-manager 70m, api 150s); clients already allow `max_kill_timeout = 24h`.

**Fix:** `DrainSandboxes` keeps polling until live count, lifecycles, and `uploadsInFlight` are all zero (logging the transition once); the orchestrator task gets `kill_timeout = "24h"` / `kill_signal = "SIGTERM"`; `FORCE_STOP=true` remains the immediate-stop escape hatch.

**Verification:** new `TestDrainSandboxesWaitsForInFlightSnapshotUploads` and `TestDrainSandboxesWaitsForUploadStartedByTheLastSandbox` (the latter replays the EN-679 sequence: last sandbox → `MarkStopping` → `MarkStopped` with an upload registered). Both fail with `main.go` reverted ("DrainSandboxes returned while a snapshot upload was in flight"); after, the package passes with `-race`. `gofmt`, `go vet`, `golangci-lint` v2.11.4 (0 issues) clean.

**Reviewer input wanted on the HCL half:** the Go change is unit-verified, but `kill_timeout = "24h"` is a cluster behaviour change I could not exercise — 24h matches client-proxy/ingress, which also wait on user workloads; say the word if you want it bounded nearer the 2 h upload budget.

### DIFF
--- a/iac/modules/job-orchestrator/jobs/orchestrator.hcl
+++ b/iac/modules/job-orchestrator/jobs/orchestrator.hcl
@@ -56,6 +56,18 @@ job "orchestrator-${latest_orchestrator_job_id}" {
     task "start" {
       driver = "raw_exec"
 
+      # The orchestrator drains on SIGTERM: it marks itself draining, waits for
+      # its sandboxes to exit and then for the detached snapshot uploads of the
+      # sandboxes that were paused on the way out. Without an explicit
+      # kill_timeout Nomad SIGKILLs after 5s, which is shorter than the drain's
+      # own 15s status-propagation wait, so the drain never got to run and a
+      # snapshot still uploading was lost. The clients allow up to 24h
+      # (max_kill_timeout in run-nomad.sh); FORCE_STOP=true is the escape hatch
+      # for an immediate stop.
+      # https://developer.hashicorp.com/nomad/docs/configuration/client#max_kill_timeout
+      kill_timeout = "24h"
+      kill_signal  = "SIGTERM"
+
       restart {
         attempts = 0
       }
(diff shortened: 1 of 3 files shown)


tomassrnka:
Checked the cursor High: it does not apply to this job. `orchestrator` is `type = "system"` with no `update` stanza, so no deployment exists and no `progress_deadline` default applies — the kill_timeout-vs-progress_deadline validation only fires for deployment-bearing groups. Verified empirically: rendered this job with the 24h kill_timeout and `nomad job validate` (v2.0.4) passes. The siblings that pair 24h/24h (client-proxy, ingress) are service jobs with update stanzas — different rules. No change made.

PR #3528, opened 2026-08-04
fix(orchestrator): restore dpkg-owned dirs under /etc/ssl/certs before cert bundle

### LINKED ISSUE #3518: Template build: finalize phase silently discards build-layer writes to /etc/ssl/certs
# Template build: finalize phase silently discards build-layer writes to `/etc/ssl/certs`, leaving `ca-certificates-java` permanently half-configured (every `apt-get` in sandboxes exits 100)

## Summary

Any template that installs a JRE (directly, or transitively e.g. via `libreoffice` → `default-jre`) ships a snapshot where:

1. `/etc/ssl/certs/java/` (owned by `ca-certificates-java`, created by dpkg during the build) is **missing** at sandbox runtime, and
2. `ca-certificates-java` is in **half-configured (`iF`)** state in the dpkg database,

even though **every build step succeeded** — inside the build layers, `dpkg -l ca-certificates-java` reports `ii` and the directory exists.

The consequence is that **every subsequent `apt-get install` inside every sandbox created from the template exits with code 100** (dpkg fails while configuring the pending `ca-certificates-java`), even though the requested package actually installs. This is especially painful for agent workloads that install packages on demand and treat non-zero exit codes as failures. Java TLS is also broken (`$JAVA_HOME/lib/security/cacerts` is a dangling symlink to `/etc/ssl/certs/java/cacerts`).

## Root cause

Three mechanisms interact:

**(a) `/etc/ssl/certs` is a tmpfs bind mount inside the guest.**
On every guest boot, `envd.service` runs `ExecStartPre=/usr/local/bin/e2b-seed-certs` (`packages/orchestrator/pkg/template/build/core/rootfs/files/seed-certs.sh.tpl`), which bind-mounts a tmpfs over `/etc/ssl/certs`, seeded from `/usr/local/share/e2b/ssl-certs.tar` (or `cp -aL` from the underlying dir). Therefore **everything dpkg writes under `/etc/ssl/certs` during build steps goes into the tmpfs and never reaches the NBD-backed rootfs**. When `ca-certificates-java` is installed in a build layer, dpkg unpacks `/etc/ssl/certs/java` into the tmpfs; the package configures fine and asserts `ii` — but the directory only exists in that boot's tmpfs.

**(b) The finalize phase boots a fresh VM instead of resuming the last build layer.**
`packages/orchestrator/pkg/template/build/phases/finalize/builder.go` uses `layer.NewCreateSandbox(...)` (line ~193). On this fresh boot, `e2b-seed-certs` re-seeds the tmpfs from the tar / underlying rootfs — neither of which contains `java/` — so the directory silently vanishes between the last build step and finalization.

**(c) `packCertBundle` re-triggers the broken state right before the snapshot.**
The build's last guest step (`packCertBundleCmd` in `packages/orchestrator/pkg/template/build/phases/finalize/configure.go`) runs `update-ca-certificates`. This invokes the `jks-keystore` hook shipped by `ca-certificates-java`, which activates its dpkg trigger and — because it runs outside a maintainer script (`DPKG_MAINTSCRIPT_PACKAGE` unset) — executes `dpkg --triggers-only --pending` itself. The triggered postinst writes to `/etc/ssl/certs/java/cacerts` **without creating the directory** (arguably a Debian bug too, but the directory is package-shipped and always exists on normal systems), fails with `java.io.FileNotFoundException`, and dpkg records `half-configured` in `/var/lib/dpkg/status` on the rootfs. `update-ca-certificates` swallows the hook failure, so **the build reports success** and the broken state is snapshotted.

Since sandbox creation is a pure Firecracker snapshot resume (no systemd boot, no tmpfiles, no re-provisioning), the state can never self-heal at runtime.

### Evidence: guest `dpkg.log` timeline from an affected build

```
16:18:19  trigproc libc-bin ...                          <- last build layer, all OK
16:18:20  trigproc dictionaries-common ...               <- ii asserted here, /etc/ssl/certs/java exists (tmpfs)
          [finalize fresh boot at 16:18:48 — tmpfs re-seeded, java/ gone]
16:18:57  status triggers-pending ca-certificates-java
16:18:57  trigproc ca-certificates-java
16:18:57  status half-configured ca-certificates-java    <- snapshotted
```

Failure inside the triggered postinst:

```
org.debian.security.UnableToSaveKeystoreException: There was a problem saving the new Java keystore.
Caused by: java.io.FileNotFoundException: /etc/ssl/certs/java/cacerts (No such file or directory)
dpkg: error processing package ca-certificates-java (--configure): ... exit status 1
```

And in every sandbox created from the template:

```
$ sudo apt-get install -y sl; echo $?
...
Errors were encountered while processing:
 ca-certificates-java
E: Sub-process /usr/bin/dpkg returned an error code (1)
100
```

## Why this contradicts the documented contract

The comment on `packCertBundleCmd` states the tar must equal the trust store the guest would regenerate at boot, *"including CAs added in user layers or start/ready, registered or not"*, and `tests/integration/internal/tests/envd/ca_cert_build_test.go` explicitly asserts that certs dropped during build steps survive into the baked bundle. The intent that build-layer trust-store changes persist is clear — the `ca-certificates-java` keystore (and the resulting dpkg state) is a case this contract misses.

## Reproduction

```python
template = (
    Template()
    .from_template("code-interpreter-v1")   # any Debian/Ubuntu base
    .apt_install(["default-jre-headless"])  # pulls in ca-certificates-java
    # optional: assert healthy state inside the build — it passes:
    .run_cmd("dpkg -l ca-certificates-java | grep '^ii' && test -d /etc/ssl/certs/java", user="root")
)
```

Then in a sandbox created from the built template:

```bash
dpkg -l ca-certificates-java        # -> iF  (half-configured)
ls /etc/ssl/certs/java              # -> No such file or directory
sudo apt-get install -y sl; echo $? # -> 100
```

## Suggested fix

In `packCertBundleCmd`, before running `update-ca-certificates`:

```sh
mkdir -p /etc/ssl/certs/java   # or: for d in $(dpkg -L ca-certificates-java 2>/dev/null | grep '^/etc/ssl/certs'); do mkdir -p "$d"; done
```

and/or run `dpkg --configure -a || true` afterwards and fail the build (or at least warn) if any package is left in a broken state — today a failed trigger is silently baked into the snapshot.

A more general option: have the finalize boot seed the tmpfs additively (preserve entries present in the underlying dir / previous layer's tmpfs that are missing from the tar) so package-owned directories under `/etc/ssl/certs` survive the fresh boot. The current behavior means *any* build-layer write under `/etc/ssl/certs` other than the bundle itself is silently discarded, which is surprising given the tested contract above.

## Workaround (for other template authors hitting `apt exit 100`)

In a build layer, move the Java keystore off the tmpfs and neutralize the trigger:

```sh
mkdir -p /etc/ssl/certs/java && dpkg --configure -a
update-ca-certificates -f || true
mkdir -p /usr/local/share/java-cacerts
cp -L /etc/ssl/certs/java/cacerts /usr/local/share/java-cacerts/cacerts
JHOME=$(dirname $(dirname $(readlink -f /usr/bin/java)))
ln -sfn /usr/local/share/java-cacerts/cacerts "$JHOME/lib/security/cacerts"
sed -i 's/^cacerts_updates=.*/cacerts_updates=no/' /etc/default/cacerts
```

With `cacerts_updates=no`, the finalize-time trigger becomes a successful no-op, the package stays `ii`, and Java reads its trust store from a rootfs path. Verified working.

## Environment

- E2B cloud, template built via Python SDK (`Template().from_template("code-interpreter-v1")`), Debian 13 guest
- `ca-certificates-java` 20240118, OpenJDK 21
- Observed on builds as of 2026-08-01; root cause traced against current `main` of this repo

AdaAibaby:
Fixes #3518

## What was broken

Any template that installs a JRE (directly or transitively) ships a snapshot where:

-  is missing at sandbox runtime
-  is in **half-configured** () state in the dpkg database

causing every subsequent  inside those sandboxes to exit with code 100.

## Root cause

Three mechanisms interact (full analysis in #3518):

1.  is a tmpfs bind-mount inside the guest — dpkg writes from build layers go there and never reach the NBD-backed rootfs.
2. The finalize phase boots a **fresh VM**;  re-seeds the tmpfs from , which never contained , so the directory silently vanishes.
3.  then runs , which activates 's dpkg trigger. The trigger's  fails with  because  is missing.  **swallows the hook failure** and exits 0 — the build reports success and the  state is baked into the snapshot.

## Fix

** — ** (primary fix):

Before :
1. Walk dpkg's file database and recreate any missing directories under 
2. Run  to resolve pending triggers now that their required paths exist

After :
3. Check for remaining  packages and **fail the build explicitly** — preventing silent broken snapshots going forward

**** (defense in depth):

Seed the tmpfs additively: copy the underlying rootfs directory first, then overlay the tar on top. The tar remains authoritative for cert files; the rootfs copy ensures subdirectories present on the rootfs but absent from an older tar survive the fresh-VM boot.

## Testing

Reproduced the issue (template with ) and confirmed:
- Before:  → ;  → exit 100
- After:  → ;  → exit 0

### DIFF
--- a/packages/orchestrator/pkg/template/build/core/rootfs/files/seed-certs.sh.tpl
+++ b/packages/orchestrator/pkg/template/build/core/rootfs/files/seed-certs.sh.tpl
@@ -18,9 +18,13 @@
 if ! mountpoint -q /etc/ssl/certs; then
     mkdir -p /run/e2b/certs
     if [ -f /usr/local/share/e2b/ssl-certs.tar ]; then
+        # Seed the underlying rootfs directory first so package-owned subdirectories
+        # (e.g. /etc/ssl/certs/java/ from ca-certificates-java) that exist on the
+        # rootfs but are absent from the tar survive the fresh-VM boot. The tar
+        # overlay then takes precedence for any file present in both sources.
+        cp -aL /etc/ssl/certs/. /run/e2b/certs/ 2>/dev/null || true
         if ! tar -C /run/e2b/certs -xf /usr/local/share/e2b/ssl-certs.tar; then
-            echo "e2b-seed-certs: ssl-certs.tar extraction failed; seeding from the live cert dir instead" >&2
-            cp -aL /etc/ssl/certs/. /run/e2b/certs/
+            echo "e2b-seed-certs: ssl-certs.tar extraction failed; running with rootfs-only certs" >&2
         fi
     else
         # Only expected during the base-layer boot, before finalize packs the tar.
--- a/packages/orchestrator/pkg/template/build/phases/finalize/configure.go
+++ b/packages/orchestrator/pkg/template/build/phases/finalize/configure.go
@@ -43,8 +43,42 @@ var ConfigureScriptTemplate = tt.Must(tt.New("provisioning-finish-script").Parse
 // already refreshes the store per family; this step only has to merge CAs that
 // later build layers dropped in, which is a Debian/Alpine convention anyway.
 const packCertBundleCmd = `set -e
+# Restore any package-owned directories under /etc/ssl/certs that exist in
+# dpkg's file database but are absent in this VM's tmpfs. They were originally
+# created by dpkg into a prior build-layer's tmpfs; finalize boots a fresh VM
+# and e2b-seed-certs re-seeds /etc/ssl/certs from ssl-certs.tar, which never
+# contained those subdirectories, so they silently vanish before this step.
+# The canonical case: ca-certificates-java ships /etc/ssl/certs/java/ and its
+# postinst writes cacerts there — if the directory is missing the trigger fails
+# with FileNotFoundException, update-ca-certificates swallows the error, and
+# the package is baked into the snapshot in half-configured (iF) state.
+if command -v dpkg-query >/dev/null 2>&1; then
+	dpkg-query -W -f='${db:Status-Abbrev} ${Package}\n' 2>/dev/null \\
+		| awk '/^.i /{print $2}' \\
+		| while IFS= read -r pkg; do dpkg -L "$pkg" 2>/dev/null; done \\
+		| grep '^/etc/ssl/certs/' \\
+		| sort -u \\
+		| while IFS= read -r path; do
+			[ -e "$path" ] || mkdir -p "$path"
+		  done
+fi
+# Resolve any dpkg triggers still pending now that their required paths exist.
+if command -v dpkg >/dev/null 2>&1; then
+	dpkg --configure -a
+fi
 if command -v update-ca-certificates >/dev/null 2>&1; then
 	update-ca-certificates
+fi
+# Fail the build explicitly if any package is still half-configured (iF).
+# update-ca-certificates swallows hook failures, so without this check a
+# broken dpkg state would be silently baked into the snapshot and every
+# subsequent apt-get inside the sandbox would exit 100.
+if command -v dpkg >/dev/null 2>&1; then
+	broken=$(dpkg -l 2>/dev/null | awk '/^iF /{print $2}' | tr '\n' ' ')
+	if [ -n "$broken" ]; then
+		printf 'e2b cert bundle: broken dpkg packages after configure: %s\n' "$broken" >&2
+		exit 1
+	fi
 fi
 mkdir -p /usr/local/share/e2b
 tar -C /etc/ssl/certs -chf /usr/local/share/e2b/ssl-certs.tar .


0xez:
Thanks for picking this up — the approach matches the root cause analysis in #3518 exactly (dpkg-owned dir restore + `dpkg --configure -a` + additive tmpfs seed + explicit build failure on broken state). One blocking issue and one minor suggestion from the issue author:

## Blocking: `\\` line continuations inside a Go raw string literal

`packCertBundleCmd` is a backtick (raw) string literal, so `\\` is **not** collapsed by the Go compiler — the guest shell receives two literal backslash characters. The shell then interprets `\\<newline>` as an escaped backslash (a stray `\` argument) followed by an *unescaped* newline, so the command terminates and the next line starts with `|`:

```
sh: syntax error near unexpected token `|'
```

Since the script runs under `set -e` (and a syntax error aborts regardless), this would fail the finalize step for **every** Debian/Ubuntu template, not just the affected ones. The continuations need to be single backslashes:

```go
+	dpkg-query -W -f='${db:Status-Abbrev} ${Package}\n' 2>/dev/null \
+		| awk '/^.i /{print $2}' \
```

(The `\n` inside the `-f=` format is fine as-is — dpkg-query itself interprets that escape, so it should stay literal.)

Easy to verify: `go test` won't catch it, but extracting the const and running it through `sh -n` will, and might be worth adding as a unit test alongside the existing template build tests.

## Minor: the `dpkg -L` loop is O(all packages)

Walking every installed package through `dpkg -L` spawns thousands of processes and does scattered reads on the lazily-fetched NBD rootfs — the exact cost profile this code path was designed to avoid (per the comment above `packCertBundleCmd`). A single-pass equivalent:

```sh
grep -hs '^/etc/ssl/certs/' /var/lib/dpkg/info/*.list | sort -u | while IFS= read -r path; do
	[ -e "$path" ] || mkdir -p "$path"
done
```

This reads each `.list` file once and needs no per-package subprocesses. (`-s` suppresses errors on non-dpkg systems, keeping the existing `command -v dpkg-query` guard optional.)

Happy to re-test against our reproduction template (`from_template("code-interpreter-v1")` + `apt_install(["default-jre-headless"])`) once updated — we have the before/after verification pipeline from #3518 ready to run.

leonmeijer:
This is superseded by 6f17229a2a75695de19cbbba4f8f65a7f13ec6c5 on `main`. That implementation restores dpkg-owned certificate paths, validates the resulting tree, and includes focused tests without the raw-string continuation bug previously identified here. This PR can be closed as incorporated.

PR #3531, opened 2026-08-04
fix(azure): add copy-public-builds support for PROVIDER=azure

michalsuba-e2b:
`make copy-public-builds` only handled `PROVIDER=aws`/gcp; `azure` fell through to the gcp else-branch and built a malformed `gs://-fc-kernels/` path from the unset `GCP_BUCKET_PREFIX`.

Adds the azure arm: download from the same public GCS bucket (`gcloud storage cp` — avoids `gsutil`'s Python version gate), upload via `az storage blob upload-batch --auth-mode login` into the `fc-kernels`/`fc-versions`/`fc-busybox` containers addressed by `AZURE_STORAGE_ACCOUNT_NAME`. Containers are unprefixed, unlike AWS/GCP's per-purpose-prefixed buckets — the BYOC Azure storage account already carries the unique prefix in its own name. The `--auth-mode login` data-plane call works because the BYOC terraform grants the apply runner `Storage Blob Data Contributor` on that account for exactly this flow.

Extracted unchanged from the `2026-06-Azure-PoC` branch, where it has run end-to-end against live Azure BYOC environments. Consumed by e2b-dev/byoc via its submodule pin (companion byoc PR bumps it).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -114,6 +114,19 @@ ifeq ($(PROVIDER),aws)
 	rm -rf ./.kernels
 	rm -rf ./.firecrackers
 	rm -rf ./.busybox
+else ifeq ($(PROVIDER),azure)
+	mkdir -p ./.kernels
+	mkdir -p ./.firecrackers
+	mkdir -p ./.busybox
+	gcloud storage cp -r "gs://e2b-prod-public-builds/kernels/*" ./.kernels/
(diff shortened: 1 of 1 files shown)


michalsuba-e2b:
Closing in favor of e2b-dev/belt#1330 — same diff, re-targeted per the belt-first process (development in e2b-dev/belt `oss/infra`, automation syncs to this repo).

PR #3534, opened 2026-08-05
fix(ci): harden composite action inputs in run scripts

SashaMIT:
## Summary
- CI hygiene for `.github/actions/integration-tests` and `unit-tests` composites.
- Bind free-string inputs (`orchestrator-make-path`, `test-path`) under `env:` before `run:` script use.
- `working-directory:` / `with:` / `if:` expressions left as-is (not shell script text).
- Defense-in-depth for Actions composites — not framed as a vulnerability ticket.

## Test plan
- [ ] Workflow YAML review
- [ ] Existing integration/unit workflows still invoke the composites


SashaMIT:
@cla-bot check

PR #3535, opened 2026-08-05
fix(envd): fail closed privileged routes when access token unset

SashaMIT:
## Summary
- Secure-by-default hardening for `envd` auth after `/init`.
- When no access token is configured, reject non-allowlisted privileged routes (including `POST /upgrade`) instead of falling open.
- Keeps existing behavior when a token *is* set, and keeps handover pre-init fail-closed for non-init/health paths.
- Defense-in-depth / secure defaults — not framed as a vulnerability ticket.

## Test plan
- [ ] Code review of `WithAuthorization` branches
- [ ] Existing envd auth tests (if any) still pass
- [ ] Confirm `/upgrade` returns 401 when token unset post-init

### DIFF
--- a/packages/envd/internal/api/auth.go
+++ b/packages/envd/internal/api/auth.go
@@ -79,6 +79,18 @@ func (a *API) WithAuthorization(handler http.Handler) http.Handler {
 			jsonError(w, http.StatusUnauthorized, errors.New("envd not initialized"))
 
 			return
+
+		case a.initialized.Load() && !allowedPath:
+			// After /init, if no access token is configured, fail CLOSED for
+			// privileged routes (including POST /upgrade) instead of allowing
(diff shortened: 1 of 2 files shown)


jakubno:
Closing this, because the new arm can only break, never protect:

- **`secure: true` sandbox** — the token is set, so the existing `a.accessToken.IsSet()` arm already enforces it and the new arm never runs.
- **Non-secure sandbox** — no token, so the new arm 401s everything. But that's open by design; `secure` is the documented opt-in for envd auth.
- **Live-upgraded envd before its post-upgrade `/init`** — already handled by the `a.handover != nil` arm, which deliberately separates "no token was configured" from "the token hasn't been restored yet."

So it can't fire on a sandbox that was meant to be authenticated

PR #3536, opened 2026-08-05
fix(envd): require sha256 for upgrade body

SashaMIT:
## Summary
- Require `X-Envd-Upgrade-Sha256` (lowercase hex SHA-256 of the request body) when `POST /upgrade` carries a non-empty body; reject with 400 if the header is missing, malformed, or mismatches — **before** writing `/usr/bin/envd.next`.
- Empty-body upgrade (restart / re-exec without a new binary) is unchanged.
- Orchestrator `CallEnvdUpgrade` hashes the local envd binary and sets the header when streaming the body.
- Complements auth fail-closed work in #3535 with content integrity (secure defaults / hardening).

## Test plan
- [x] `go test ./internal/services/process/ -run WriteUpgradeBinary` (packages/envd) — missing / malformed / wrong / correct hash paths
- [x] `go build` packages/envd
- [ ] CI: `go test` packages/orchestrator/pkg/sandbox (`TestCallEnvdUpgrade_SetsBodySHA256Header`, linux) — not executable on darwin host here
- [ ] Manual: resume-path live upgrade still delivers via orchestrator with matching digest

### DIFF
--- a/packages/envd/internal/services/process/upgrade_deliver.go
+++ b/packages/envd/internal/services/process/upgrade_deliver.go
@@ -1,0 +1,77 @@
+package process
+
+import (
+	"crypto/sha256"
+	"crypto/subtle"
+	"encoding/hex"
+	"errors"
+	"io"
+	"os"
+)
+
+// UpgradeSha256Header carries the lowercase hex SHA-256 digest of a non-empty
+// POST /upgrade request body. envd refuses to materialize the body on disk
+// unless this header is present, well-formed, and matches the bytes — a
+// content-integrity check complementary to the authenticated trigger and the
+// fixed-path allowlist on DefaultUpgradeBinPath.
+const UpgradeSha256Header = "X-Envd-Upgrade-Sha256"
+
+var (
+	ErrUpgradeSHA256Missing   = errors.New("missing X-Envd-Upgrade-Sha256 header")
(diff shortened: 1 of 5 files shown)


jakubno:
Thanks — this one is carefully built, and the compatibility story holds up (`resolveEnvdUpgradePath` refuses `downgrade`/`same_version` keyed on the live envd version). But I don't think the check buys what it's meant to.

The digest is supplied by the same request as the body, so it proves self-consistency rather than integrity: anyone who can POST a body can compute its SHA-256. Against a caller that can reach `/upgrade` it's a no-op, and transport truncation is already covered — the orchestrator sets `ContentLength` (`packages/orchestrator/pkg/sandbox/envd.go:245-247`), so a short body errors and the handler returns 500 without exec'ing.

`WriteUpgradeBinary` also hashes the in-memory buffer and then writes, never re-reading `destPath` — so it doesn't cover the one case a post-transfer digest is good for, a bad write.

One regression worth flagging regardless: swapping the streaming `io.Copy` for `io.ReadAll` buffers the whole binary in guest memory, inside a memory-constrained VM, while the workload is frozen but still resident.

SashaMIT:
Fair on all counts. The same-request digest only proves self-consistency, and you are right about the io.ReadAll regression: buffering the binary in a memory-constrained guest with the workload resident is a bad trade for a check that does not buy integrity. Thanks for the careful read, both points noted for future submissions.

PR #3541, opened 2026-08-06
fix(docker-reverse-proxy): restore resp.Body after io.ReadAll in ModifyResponse

### LINKED ISSUE #3540: bug(docker-reverse-proxy): TCP connection leak on 401 response — resp.Body not restored after io.ReadAll
## Summary

In `packages/docker-reverse-proxy/internal/handlers/store.go`, the `ModifyResponse` hook reads the response body with `io.ReadAll` on 401 responses but never restores it. `httputil.ReverseProxy` then attempts to stream the (now-empty) body to the downstream client. Because the body was consumed outside the proxy's copy loop, the underlying TCP connection cannot be returned to the keep-alive pool and is silently abandoned.

## Affected code

```go
// store.go:56-64
proxy.ModifyResponse = func(resp *http.Response) error {
    if resp.StatusCode == http.StatusUnauthorized {
        respBody, _ := io.ReadAll(resp.Body)
        log.Printf("Unauthorized request:[%s] %s\n", resp.Request.Method, respBody)
        // resp.Body is NOT restored — body is at EOF, connection cannot be reused
    }
    return nil
}
```

## Why this leaks

`httputil.ReverseProxy` reuses the upstream TCP connection only when it successfully drains `resp.Body` through its own `io.Copy` path. After `io.ReadAll` empties the body out-of-band, the transport marks the connection as non-reusable and closes it.

## Impact

- Every 401 from GCP Artifact Registry leaks one TCP connection permanently.
- Under sustained unauthorized-access or credential rotation, the connection pool is exhausted — subsequent image-pull requests block.
- docker-reverse-proxy is on the critical path for all `e2b.Sandbox` launches with custom templates.

## Fix

```diff
 if resp.StatusCode == http.StatusUnauthorized {
     respBody, _ := io.ReadAll(resp.Body)
+    resp.Body = io.NopCloser(bytes.NewReader(respBody))
     log.Printf("Unauthorized request:[%s] %s\n", resp.Request.Method, respBody)
 }
```

Restoring `resp.Body` is the standard pattern for `ModifyResponse` hooks that inspect the body.

AdaAibaby:
Fixes #3540

## Problem

The `ModifyResponse` hook in `store.go` calls `io.ReadAll(resp.Body)` to log 401 error bodies, but never restores `resp.Body`. After the read, the body is at EOF. When `httputil.ReverseProxy` subsequently tries to copy the body to the downstream client, it reads zero bytes — but the transport's keep-alive reuse logic requires the body to be drained through the proxy's own copy path. Because it wasn't, the upstream TCP connection is marked non-reusable and abandoned.

One TCP connection leaks per 401 response from GCP Artifact Registry.

## Fix

Restore `resp.Body` with `io.NopCloser(bytes.NewReader(respBody))` immediately after `io.ReadAll`, so `ReverseProxy` can stream the body normally and return the connection to the pool.

```diff
 respBody, _ := io.ReadAll(resp.Body)
+resp.Body = io.NopCloser(bytes.NewReader(respBody))
 log.Printf("Unauthorized request:[%s] %s\n", resp.Request.Method, respBody)
```

This is the standard pattern for `ModifyResponse` hooks that need to inspect the response body (documented in the `httputil.ReverseProxy` source).

## Testing

- `go build ./packages/docker-reverse-proxy/...` passes
- `go vet ./packages/docker-reverse-proxy/...` passes

### DIFF
--- a/packages/docker-reverse-proxy/internal/handlers/store.go
+++ b/packages/docker-reverse-proxy/internal/handlers/store.go
@@ -1,6 +1,7 @@
 package handlers
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"io"
@@ -56,10 +57,10 @@ func NewStore(ctx context.Context) *APIStore {
 	proxy.ModifyResponse = func(resp *http.Response) error {
 		if resp.StatusCode == http.StatusUnauthorized {
 			respBody, _ := io.ReadAll(resp.Body)
+			resp.Body = io.NopCloser(bytes.NewReader(respBody))
 			log.Printf("Unauthorized request:[%s] %s\n", resp.Request.Method, respBody)
 		}
 
-		// You can also modify the response here if needed
 		return nil
 	}
 


AdaAibaby:

PR #3543, opened 2026-08-06
fix(envd/port): prevent ScanAndBroadcast from blocking when a subscriber stops receiving

### LINKED ISSUE #3542: bug(envd/port): ScanAndBroadcast goroutine blocks indefinitely when a subscriber stops receiving, rendering Scanner.Destroy() ineffective
## Summary

`ScanAndBroadcast` iterates all subscribers and calls `sub.Signal()` synchronously. `Signal()` sends on an **unbuffered** `Messages` channel with no timeout or select escape. When a subscriber's receiver exits (e.g. `StartForwarding` returns on context cancellation), the send blocks forever. `ScanAndBroadcast` is then stuck inside the subscriber loop and **never reaches the `scanExit` select**, making `Scanner.Destroy()` permanently ineffective. The scanner cannot be stopped until the process exits.

---

### `scan.go:45-60` — broadcast loop with no escape after Signal()
```go
func (s *Scanner) ScanAndBroadcast() {
    ticker := time.NewTicker(s.period)
    defer ticker.Stop()

    for {
        processes, _ := net.Connections(tcp)
        for _, sub := range s.subs.Items() {
            sub.Signal(processes)   // blocks indefinitely if receiver exited
        }
        select {           // ← never reached when Signal() blocks
        case <-s.scanExit:
            return
        case <-ticker.C:
        }
    }
}
```

### `scanSubscriber.go:35-49` — unconditional blocking send on unbuffered channel
```go
func (ss *ScannerSubscriber) Signal(proc []net.ConnectionStat) {
    if ss.filter == nil {
        ss.Messages <- proc     // blocks if nobody is receiving
    } else {
        ...
        ss.Messages <- filtered // blocks if nobody is receiving
    }
}

// NewScannerSubscriber uses an unbuffered channel:
Messages: make(chan []net.ConnectionStat),
```

### `forward.go:101-108` — receiver exits without draining Messages
```go
func (f *Forwarder) StartForwarding(ctx context.Context) {
    for {
        select {
        case <-ctx.Done():
            return   // exits immediately, no drain of f.scannerSubscriber.Messages
        case procs, ok := <-f.scannerSubscriber.Messages:
            ...
        }
    }
}
```

---

## Failure sequence

1. `ctx` passed to `StartForwarding` is cancelled (sandbox shutdown or upgrade failure).
2. `StartForwarding` receives `<-ctx.Done()` and returns — no longer consuming `Messages`.
3. `ScanAndBroadcast` fires on its 1-second ticker, calls `sub.Signal(processes)`.
4. `Signal()` tries to send on the now-unread `Messages` channel — **blocks indefinitely**.
5. `defer portScanner.Destroy()` runs (closes `scanExit`), but `ScanAndBroadcast` is stuck in step 4 and can never observe it.
6. The goroutine leaks for the remainder of the process lifetime.

This is acknowledged in `main.go:439`:
```go
// TODO: shutdown synchronization needs to be revisited.
```

---

## Secondary issue — latent panic via Unsubscribe

`Scanner.Unsubscribe` closes `Messages`:

```go
func (s *Scanner) Unsubscribe(sub *ScannerSubscriber) {
    s.subs.Remove(sub.ID())
    sub.Destroy()   // closes Messages
}
```

If `Unsubscribe` is ever called while `ScanAndBroadcast` is blocked inside `Signal()` (trying to send to the same `Messages` channel), the result is a **send on closed channel → panic**, crashing the entire envd daemon. This is currently avoided only because `Unsubscribe` is not called in the active path — not because the design prevents it.

---

## Impact

- **`Scanner.Destroy()` is silently ineffective** whenever a subscriber has stopped reading. The scanner cannot be cleanly shut down.
- **All subscribers are serialized**: a single slow or stopped receiver stalls the broadcast loop for every other subscriber. If additional subscribers are added in future, one stuck receiver blocks all port-forward updates for the entire sandbox.
- **Latent panic**: any future call to `Unsubscribe` while the scanner is mid-`Signal()` crashes envd.

---

## Fix

**Option A (minimal)**: Use a non-blocking send in `Signal()`:
```go
func (ss *ScannerSubscriber) Signal(proc []net.ConnectionStat) {
    msg := proc
    if ss.filter != nil {
        // build filtered slice...
    }
    select {
    case ss.Messages <- msg:
    default:  // drop if receiver is not ready
    }
}
```

**Option B (preferred)**: Add a `done` channel to `ScannerSubscriber` and select on it inside `Signal()`, so the broadcaster can be unblocked when the subscriber is torn down:
```go
select {
case ss.Messages <- msg:
case <-ss.done:
}
```
This also fixes the `Unsubscribe` panic: closing `done` unblocks any in-flight `Signal()` before `close(Messages)` is called.

AdaAibaby:
Fixes #3542

## Problem

`Signal()` sent on an unbuffered `Messages` channel with no escape path. When a subscriber's receiver exited (e.g. `StartForwarding` returning on `ctx.Done()`), the channel send blocked indefinitely. `ScanAndBroadcast` was then stuck inside the subscriber loop, never reaching the `scanExit` select — making `Scanner.Destroy()` permanently ineffective.

A secondary hazard: `Destroy()` closed `Messages` directly. If `Signal()` was concurrently blocked sending to `Messages` when `Destroy()` ran, the close triggered a `"send on closed channel"` panic (reproducible with the new tests below).

## Fix

Add a `done chan struct{}` to `ScannerSubscriber`.

- `Destroy()` now closes `done` instead of `Messages`.
- `Signal()` selects on both `Messages` and `done`, so it returns immediately when the subscriber is torn down rather than blocking.
- `ScanAndBroadcast` is never stalled by a stopped subscriber and can always observe `scanExit`.

```diff
+	done     chan struct{}

+	done: make(chan struct{}),

 func (ss *ScannerSubscriber) Destroy() {
-	close(ss.Messages)
+	close(ss.done)
 }

 func (ss *ScannerSubscriber) Signal(proc []net.ConnectionStat) {
-    ss.Messages <- proc  // blocked forever if receiver exited
+    select {
+    case ss.Messages <- msg:
+    case <-ss.done:      // returns immediately after Destroy()
+    }
 }
```

## Tests

Three new tests in `scanner_test.go` (all run with `-race`):

| Test | What it verifies |
|------|-----------------|
| `TestScannerSubscriber_Signal_DoesNotBlockAfterDestroy` | `Signal()` returns within 1s after `Destroy()` |
| `TestScanner_Destroy_ExitsAfterSubscriberStopped` | `Scanner.Destroy()` stops `ScanAndBroadcast` even when subscriber is stopped |
| `TestScanner_Destroy_NoPanicOnConcurrentSignal` | Concurrent `Unsubscribe` + `ScanAndBroadcast` does not panic |

**Before fix**: `TestScannerSubscriber_Signal_DoesNotBlockAfterDestroy` panics with `send on closed channel`; `TestScanner_Destroy_ExitsAfterSubscriberStopped` times out.

**After fix**: all 6 tests in the port package pass (`go test -race -count=1 ./packages/envd/internal/port/...`).

### DIFF
--- a/packages/envd/internal/port/scanSubscriber.go
+++ b/packages/envd/internal/port/scanSubscriber.go
@@ -12,6 +12,7 @@ type ScannerSubscriber struct {
 	logger   *zerolog.Logger
 	filter   *ScannerFilter
 	Messages chan ([]net.ConnectionStat)
+	done     chan struct{}
 	id       string
 }
 
@@ -21,6 +22,7 @@ func NewScannerSubscriber(logger *zerolog.Logger, id string, filter *ScannerFilt
 		id:       id,
 		filter:   filter,
 		Messages: make(chan []net.ConnectionStat),
+		done:     make(chan struct{}),
 	}
 }
 
@@ -28,23 +30,32 @@ func (ss *ScannerSubscriber) ID() string {
 	return ss.id
 }
 
+// Destroy signals the subscriber to stop. Any Signal() call blocked waiting
+// for a receiver will return immediately. The Messages channel is left open so
+// in-progress receivers can drain it; callers exit via ctx.Done() instead.
 func (ss *ScannerSubscriber) Destroy() {
-	close(ss.Messages)
+	close(ss.done)
 }
 
+// Signal delivers a port scan result to the subscriber. If the receiver is not
+// ready (e.g. it has stopped consuming after context cancellation), Signal
+// returns immediately rather than blocking, so ScanAndBroadcast is never
+// stalled by a single slow or stopped subscriber.
 func (ss *ScannerSubscriber) Signal(proc []net.ConnectionStat) {
-	// Filter isn't specified. Accept everything.
-	if ss.filter == nil {
-		ss.Messages <- proc
-	} else {
-		filtered := []net.ConnectionStat{}
+	msg := proc
+	if ss.filter != nil {
+		filtered := make([]net.ConnectionStat, 0, len(proc))
 		for i := range proc {
-			// We need to access the list directly otherwise there will be implicit memory aliasing
-			// If the filter matched a process, we will send it to a channel.
+			// Access the slice directly to avoid implicit memory aliasing.
 			if ss.filter.Match(&proc[i]) {
 				filtered = append(filtered, proc[i])
 			}
 		}
-		ss.Messages <- filtered
+		msg = filtered
+	}
+
+	select {
+	case ss.Messages <- msg:
+	case <-ss.done:
 	}
 }
--- a/packages/envd/internal/port/scanner_test.go
+++ b/packages/envd/internal/port/scanner_test.go
@@ -1,0 +1,107 @@
+package port
+
+import (
+	"testing"
+	"time"
+
+	"github.com/rs/zerolog"
+)
+
+// TestScannerSubscriber_Signal_DoesNotBlockAfterDestroy verifies that Signal()
+// returns immediately when the subscriber has been destroyed, rather than
+// blocking forever on the unbuffered Messages channel.
+func TestScannerSubscriber_Signal_DoesNotBlockAfterDestroy(t *testing.T) {
+	t.Parallel()
+
+	l := zerolog.Nop()
+	sub := NewScannerSubscriber(&l, "test", nil)
+	sub.Destroy()
+
+	done := make(chan struct{})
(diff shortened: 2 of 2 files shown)


jakubno:
addressed in 1b65106ac3e4bcf2217f1964f2781a379a8d23b0

PR #3549, opened 2026-08-07
nfsproxy: make NFS file-handle cache limit configurable via NFS_PROXY_CACHE_LIMIT

### LINKED ISSUE #3548: nfsproxy: hardcoded cacheLimit=1024 causes ESTALE errors on large npm/pip installs
## Symptom

After mounting a persistent volume into a sandbox, large package installs (e.g. `npm i -g @openai/codex`) fail with:

```
npm error code UNKNOWN
npm error syscall write
npm error errno -116
npm error UNKNOWN: unknown error, write
```

`errno -116` = `ESTALE` (Stale NFS file handle).

## Root cause

`packages/orchestrator/pkg/nfsproxy/proxy.go:25` hard-codes the NFS file-handle LRU size:

```go
const cacheLimit = 1024
handler = helpers.NewCachingHandler(handler, cacheLimit)
```

`CachingHandler` maps `uuid → (filesystem, path)` via an LRU. When the cache fills, the oldest entry is evicted. If the client then sends an RPC using that evicted handle, `FromHandle` can no longer resolve it and returns `NFSStatusStale` (NFS status 70), which the Linux kernel translates to `ESTALE`:

```go
// helpers/cachinghandler.go:100
return nil, []string{}, &nfs.NFSStatusError{NFSStatus: nfs.NFSStatusStale}
```

**Amplifying factor**: the 1024-slot LRU is **shared across all sandboxes on the node** — not per-sandbox. Under concurrent load, effective slots per sandbox are even fewer.

## Why 1024 is not enough

`@openai/codex` fetches 6 platform-specific optional packages simultaneously (linux-x64/arm64, darwin-x64/arm64, win32-x64/arm64). npm's reify phase issues concurrent `CREATE`/`RENAME`/`WRITE` RPCs for each; every file and directory consumes a handle slot. 1024 is exhausted quickly.

Similar workloads that trigger this:
- Any large `npm i` with per-platform optional packages
- `pip install` of ML frameworks with many transitive dependencies
- Extracting large archives into the volume

## Proposed fix

Replace the hardcoded constant with an env-var-configurable value so operators can tune it **without recompiling the orchestrator**:

- Add `NFS_PROXY_CACHE_LIMIT int` to the orchestrator config (`pkg/cfg/model.go`) with `envDefault:"16384"`
- Raise the default from 1024 → 16384 (each entry ≈150 bytes, 16384 entries ≈ 2.4 MB — negligible)
- Thread the value through `nfsproxy/cfg.Config` → `nfsproxy.NewProxy()`

**Longer-term**: give each sandbox/mount its own `CachingHandler` instance to eliminate cross-sandbox LRU contention entirely.

AdaAibaby:
Fixes #3548.

## Problem

`CachingHandler` in go-nfs maps NFS file handles (UUID) to `(filesystem, path)` via an LRU. The LRU size was hardcoded at `const cacheLimit = 1024`. When a sandbox installs a large package — for example `npm i -g @openai/codex`, which pulls 6 platform-specific optional packages simultaneously — the cache fills and oldest entries are evicted. Any subsequent RPC using an evicted handle returns `NFS3ERR_STALE`, which Linux translates to `ESTALE` (errno 116):

```
npm error code UNKNOWN
npm error syscall write
npm error errno -116
npm error UNKNOWN: unknown error, write
```

The problem is amplified by the fact that the single 1024-slot LRU was shared across **all sandboxes on the node**.

## Fix

Replace the hardcoded constant with a configurable env var so operators can tune the cache **without recompiling the orchestrator**:

| | Before | After |
|---|---|---|
| Cache size | `const cacheLimit = 1024` (hardcoded) | `NFS_PROXY_CACHE_LIMIT` (env var, default `16384`) |
| Memory cost at default | ~150 KB | ~2.4 MB (+2.25 MB per node) |
| Tunable at runtime | No | Yes — change Nomad job env and redeploy |

## Changes

- `pkg/cfg/model.go`: add `NFSProxyCacheLimit int` with `env:"NFS_PROXY_CACHE_LIMIT" envDefault:"16384"`
- `pkg/nfsproxy/cfg/model.go`: add `CacheLimit int` field
- `pkg/nfsproxy/proxy.go`: remove `const cacheLimit = 1024`; use `config.CacheLimit`
- `pkg/factories/run.go`: thread `config.NFSProxyCacheLimit` → `nfscfg.Config.CacheLimit`

## Not in scope

Per-sandbox cache isolation (each mount gets its own `CachingHandler`) would eliminate cross-sandbox LRU contention entirely — tracked in #3548 as a longer-term improvement.

### DIFF
--- a/packages/api/internal/handlers/sandbox_kill.go
+++ b/packages/api/internal/handlers/sandbox_kill.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/gin-gonic/gin"
 	"github.com/google/uuid"
+	"go.uber.org/zap"
 
 	"github.com/e2b-dev/infra/packages/api/internal/db"
 	"github.com/e2b-dev/infra/packages/api/internal/orchestrator"
@@ -69,12 +70,26 @@ func (a *APIStore) DeleteSandboxesSandboxID(
 	case err == nil:
 		killedOrRemoved = true
 	case errors.Is(err, orchestrator.ErrSandboxNotFound):
-		logger.L().Debug(ctx, "Running sandbox not found", logger.WithSandboxID(sandboxID))
+		logger.L().Info(ctx, "Running sandbox not found, checking for snapshot",
+			logger.WithSandboxID(sandboxID),
+			logger.WithTeamID(teamID.String()),
+		)
 	case errors.Is(err, orchestrator.ErrSandboxOperationFailed):
+		logger.L().Error(ctx, "Failed to kill sandbox on orchestrator node",
+			logger.WithSandboxID(sandboxID),
+			logger.WithTeamID(teamID.String()),
+			zap.Error(err),
+		)
+		telemetry.ReportError(ctx, "error killing sandbox on node", err)
 		a.sendAPIStoreError(c, http.StatusInternalServerError, fmt.Sprintf("Error killing sandbox: %s", err))
 
 		return
 	default:
+		logger.L().Error(ctx, "Unexpected error removing sandbox",
+			logger.WithSandboxID(sandboxID),
+			logger.WithTeamID(teamID.String()),
+			zap.Error(err),
+		)
 		telemetry.ReportError(ctx, "error killing sandbox", err)
 		a.sendAPIStoreError(c, http.StatusInternalServerError, fmt.Sprintf("Error killing sandbox: %s", err))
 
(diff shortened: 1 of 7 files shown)


Merlin0220:
We encountered the same issue. As a temporary workaround, we increased the cache size to 16 * 1024. We later tried switching to a per-sandbox cache, but found that it increased sandbox creation time. We have not yet identified the exact cause.

leonmeijer (commented):
NFS_PROXY_CACHE_LIMIT accepts zero or negative values and NewProxy passes them directly to helpers.NewCachingHandler. The pinned go-nfs helper ignores the LRU constructor errors, leaves nil caches, and then panics on the first handle operation; zero-value Config callers have the same failure. Validate a positive limit and give nfsproxy.Config a safe default before constructing the handler.

PR #3551, opened 2026-08-08
nfsproxy: preserve caching handler through middleware chain

### LINKED ISSUE #3552: nfsproxy: middleware chain hides CachingHandler and disables directory verifier caching
## Symptom

The NFS proxy's directory verifier cache is silently bypassed.

Paginated `READDIR` and `READDIRPLUS` requests cannot reuse the directory listing associated with the cookie verifier returned by the first page. Each subsequent request reads and sorts the directory again.

If the directory changes between pages, the newly calculated verifier differs from the verifier supplied by the client, causing go-nfs to return `NFS3ERR_BAD_COOKIE`:

```go
if obj.Cookie > 0 && obj.CookieVerif > 0 && verifier != obj.CookieVerif {
    return &NFSStatusError{NFSStatusBadCookie, nil}
}
```

Even when the directory does not change, the repeated `ReadDir` calls add unnecessary filesystem work for large paginated listings.

## Root cause

The proxy creates a `helpers.CachingHandler` before applying its middleware:

```go
handler = helpers.NewCachingHandler(handler, cacheLimit)

if config.Tracing {
    handler = tracing.WrapWithTracing(handler, config)
}

if config.Metrics {
    handler = metrics.WrapWithMetrics(handler, config)
}

if config.Logging {
    handler = logged.WrapWithLogging(ctx, handler, config)
}

handler = recovery.WrapWithRecovery(ctx, handler)
```

go-nfs checks the outermost handler for the optional `nfs.CachingHandler` interface:

```go
if vh, ok := userHandle.(CachingHandler); verifier != 0 && ok {
    entries := vh.DataForVerifier(path, verifier)
    if entries != nil {
        return entries, verifier, nil
    }
}
```

The tracing, metrics, logging, and recovery handlers only implement `nfs.Handler`. They do not expose or forward:

```go
VerifierFor(path string, contents []fs.FileInfo) uint64
DataForVerifier(path string, verifier uint64) []fs.FileInfo
```

Because recovery middleware is always installed, the outermost handler never satisfies `nfs.CachingHandler`, regardless of whether tracing, metrics, or logging are enabled.

As a result, go-nfs never reaches the verifier cache maintained by `helpers.CachingHandler`:

```go
func (c *CachingHandler) VerifierFor(path string, contents []fs.FileInfo) uint64 {
    id := hashPathAndContents(path, contents)
    c.activeVerifiers.Add(id, verifier{path, contents})
    return id
}

func (c *CachingHandler) DataForVerifier(path string, id uint64) []fs.FileInfo {
    if cache, ok := c.activeVerifiers.Get(id); ok {
        return cache.contents
    }

    return nil
}
```

## Why it matters

NFS directory cookies are indexes into the sorted directory listing:

```go
cookie := uint64(i + 2)
```

For pagination to remain consistent, subsequent requests should use the same listing that produced the original cookie verifier.

Without verifier caching:

- Every page performs another `ReadDir` and sort.
- Insertions or removals can shift cookie indexes between pages.
- Directory changes can cause `NFS3ERR_BAD_COOKIE`.
- Large directory listings perform unnecessary repeated filesystem operations.

## Proposed fix

Preserve `nfs.CachingHandler` through the complete middleware chain:

- Implement `VerifierFor` and `DataForVerifier` on the tracing handler.
- Implement `VerifierFor` and `DataForVerifier` on the metrics handler.
- Implement `VerifierFor` and `DataForVerifier` on the logging handler.
- Implement `VerifierFor` and `DataForVerifier` on the recovery handler.
- Forward both calls to the wrapped handler when it implements `nfs.CachingHandler`.
- Add compile-time assertions that each middleware implements `nfs.CachingHandler`.
- Preserve panic recovery around verifier operations in the recovery handler.
- Add a test covering the complete middleware chain and verifying that arguments and return values reach the underlying caching handler.

Merlin0220:
Fixes https://github.com/e2b-dev/infra/issues/3552.
## Problem

The NFS proxy creates a `helpers.CachingHandler` to cache directory listings by their NFS cookie verifier. However, the caching handler is subsequently wrapped by tracing, metrics, logging, and recovery middleware.

go-nfs detects verifier caching support by asserting the outermost handler as `nfs.CachingHandler`:

```go
handler, ok := userHandle.(nfs.CachingHandler)
```

The middleware handlers only implemented `nfs.Handler`, so this assertion failed before reaching the underlying caching handler. Because the recovery middleware is always installed, verifier caching was bypassed regardless of whether the optional tracing, metrics, or logging middleware was enabled.

As a result, paginated `READDIR` and `READDIRPLUS` requests could not reuse the directory listing associated with their verifier and had to read the directory again.

## Fix

Preserve `nfs.CachingHandler` through the complete middleware chain by forwarding `VerifierFor` and `DataForVerifier` to the wrapped handler.

| Behavior | Before | After |
|---|---|---|
| Outermost handler implements `nfs.CachingHandler` | No | Yes |
| Verifier cache reaches `helpers.CachingHandler` | No | Yes |
| Directory listings can be reused across paginated requests | No | Yes |

## Changes

- Add `nfs.CachingHandler` compile-time interface assertions to the tracing, metrics, logging, and recovery handlers.
- Forward `VerifierFor` and `DataForVerifier` through each middleware layer.
- Preserve panic recovery for verifier cache operations in the recovery handler.
- Add a middleware-chain test that verifies the outer handler satisfies `nfs.CachingHandler` and forwards arguments and return values to the underlying handler.

## Testing

```text
go test ./pkg/nfsproxy/... -count=1
go test -race ./pkg/nfsproxy/... -count=1
```

## Risk: activeVerifiers Cache Key Collision Across Sandboxes

Enabling verifier caching introduces a potential cache-key collision in
`activeVerifiers`. If the mount path and an enumerated file name are identical,
they may produce the same ID. This could cause cache entries to be reused across
different sandboxes, violating cache isolation and potentially leading to
incorrect verifier or file-handle resolution.

If this scenario is possible, the current cache should not be used as-is.
The cache should either be disabled or redesigned with mount-point-scoped, collision-resistant keys.

This risk is related to [#3555](https://github.com/e2b-dev/infra/issues/3555).
The proposed solution is to allocate an independent file-handle cache for each
mount point, using the mount-point identity as the cache
shard key. This prevents cache eviction, lookup, and synchronization in one
sandbox from affecting another.

### DIFF
--- a/packages/orchestrator/pkg/nfsproxy/logged/handler.go
+++ b/packages/orchestrator/pkg/nfsproxy/logged/handler.go
@@ -3,6 +3,7 @@ package logged
 import (
 	"context"
 	"fmt"
+	iofs "io/fs"
 	"net"
 
 	"github.com/go-git/go-billy/v5"
@@ -17,7 +18,10 @@ type loggedHandler struct {
 	config cfg.Config
 }
 
-var _ nfs.Handler = (*loggedHandler)(nil)
+var (
+	_ nfs.Handler        = (*loggedHandler)(nil)
+	_ nfs.CachingHandler = (*loggedHandler)(nil)
+)
 
 func WrapWithLogging(ctx context.Context, handler nfs.Handler, config cfg.Config) nfs.Handler {
 	return loggedHandler{ctx: ctx, inner: handler, config: config}
@@ -95,4 +99,22 @@ func (e loggedHandler) HandleLimit() int {
 	defer finish(e.ctx, nil)
 
 	return e.inner.HandleLimit()
+}
+
+func (e loggedHandler) VerifierFor(path string, contents []iofs.FileInfo) uint64 {
+	handler, ok := e.inner.(nfs.CachingHandler)
+	if !ok {
+		return 0
+	}
+
+	return handler.VerifierFor(path, contents)
+}
+
+func (e loggedHandler) DataForVerifier(path string, verifier uint64) []iofs.FileInfo {
+	handler, ok := e.inner.(nfs.CachingHandler)
+	if !ok {
+		return nil
+	}
+
+	return handler.DataForVerifier(path, verifier)
 }
--- a/packages/orchestrator/pkg/nfsproxy/metrics/handler.go
+++ b/packages/orchestrator/pkg/nfsproxy/metrics/handler.go
@@ -3,6 +3,7 @@ package metrics
 import (
 	"context"
 	"fmt"
+	iofs "io/fs"
 	"net"
 
 	"github.com/go-git/go-billy/v5"
@@ -16,7 +17,10 @@ type metricsHandler struct {
 	config cfg.Config
 }
 
-var _ nfs.Handler = (*metricsHandler)(nil)
+var (
+	_ nfs.Handler        = (*metricsHandler)(nil)
+	_ nfs.CachingHandler = (*metricsHandler)(nil)
+)
 
 func WrapWithMetrics(handler nfs.Handler, config cfg.Config) nfs.Handler {
 	return &metricsHandler{inner: handler, config: config}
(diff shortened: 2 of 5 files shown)


Merlin0220:
> We require contributors to sign our Contributor License Agreement, and we don't have @Merlin0220 on file. You can sign our CLA at https://e2b.dev/docs/cla . Once you've signed, post a comment here that says '@cla-bot check'

@cla-bot check

leonmeijer (commented):
This exposes one process-wide verifier cache across every mounted sandbox, but the pinned go-nfs caching handler derives a verifier only from the namespace-relative path and entry names and DataForVerifier does not validate the path or filesystem. Two sandboxes with the same directory path and names can therefore collide and receive each others cached FileInfo listings on subsequent READDIR pages. Isolate verifier state by mounted filesystem or include and validate mount identity before enabling this passthrough.

PR #3554, opened 2026-08-09
fix(api): expose sandbox placement failure reason in error message

### LINKED ISSUE #3553: api: "Failed to place sandbox" gives no diagnostic - expose why no node accepted the sandbox
## Problem

When sandbox creation fails because no node can host it, the client receives a generic 500 error:

This tells the user nothing about *why* placement failed. The actual cause is buried in server logs as an internal error and is never surfaced. Common root causes include:

- **CPU model mismatch**: the template was built on a newer CPU generation (e.g. Emerald Rapids, model 207) but all cluster nodes run an older generation (e.g. Ice Lake, model 106). Cross-generation placement is asymmetric — an n2 build can resume on an n4 node, but not the reverse.
- **Label mismatch**: the team or template requires scheduling labels (e.g. `gpu`, `fast-disk`) that no available node carries.
- **All nodes not-accepting**: nodes are draining, unhealthy, or saturated.

Without seeing which filter eliminated all nodes it is impossible to self-diagnose or file a useful bug report.

## Root cause

`BestOfK.chooseNode` returns a `FailedToPlaceSandboxError` whose `Error()` string includes `machine=...` and `labels=...` constraints, but `create_instance.go` stores that in the internal `Err` field and returns a hardcoded `ClientMsg: "Failed to place sandbox"` that never reaches the client.

## Proposed fix

1. Track per-filter rejection counts inside `BestOfK.sample()` (not-accepting, cpu-incompatible, label-filtered, excluded).
2. Embed those counts plus the build CPU constraints in `FailedToPlaceSandboxError.Error()`.
3. Propagate `err.Error()` to `ClientMsg` in `create_instance.go`.

**Example new error message:**
```
Failed to place sandbox: no compatible node found (38 nodes checked: 0 not-accepting, 38 cpu-incompatible, 0 label-filtered, 0 excluded); build cpu: arch=x86_64 family=6 model=207
```

This lets the caller immediately identify a CPU-generation mismatch or label gap without digging through server logs.

AdaAibaby:
Fixes #3553

## What

When `BestOfK.chooseNode` finds no eligible node it returns a `FailedToPlaceSandboxError`, but `create_instance.go` only uses the internal `Err` field and returns the opaque `"Failed to place sandbox"` to the client.

This PR makes the error actionable:

1. **`BestOfK.sample()`** now counts how many nodes were rejected at each filter stage: `not-accepting`, `cpu-incompatible`, `label-filtered`, `excluded`.
2. **`FailedToPlaceSandboxError.Error()`** emits a structured message with those counts plus the build CPU constraints and required labels.
3. **`create_instance.go`** propagates `err.Error()` to `ClientMsg` so the caller receives the full diagnosis.

**Before:**
```
500: Failed to place sandbox
```

**After:**
```
500: Failed to place sandbox: no compatible node found (38 nodes checked: 0 not-accepting, 38 cpu-incompatible, 0 label-filtered, 0 excluded); build cpu: arch=x86_64 family=6 model=207
```

## Why this is safe to expose

The message reveals CPU generation and scheduling labels — information already visible through the admin `GET /nodes` endpoint. No secrets or topology details are disclosed.

## Changes

- `packages/api/internal/orchestrator/placement/placement_best_of_K.go` — add `nodeRejectionCounts`, thread through `sample()`, populate `FailedToPlaceSandboxError`, update `Error()`
- `packages/api/internal/orchestrator/create_instance.go` — `ClientMsg` now includes `err.Error()`
- `packages/api/internal/orchestrator/placement/placement_best_of_K_test.go` — update assertions for new error format

### DIFF
--- a/packages/api/internal/orchestrator/create_instance.go
+++ b/packages/api/internal/orchestrator/create_instance.go
@@ -357,6 +357,12 @@ func (o *Orchestrator) CreateSandbox(
 			o.maybeRemapResumeOriginNode(ctx, snapshotSandboxID, team, sbxData.NodeID, placed.WarmedNode)
 		}
 
+		logger.L().Error(ctx, "failed to place sandbox",
+			logger.WithSandboxID(sandboxID),
+			logger.WithTemplateID(sbxData.TemplateID),
+			zap.Error(err),
+		)
+
 		return sandbox.Sandbox{}, &api.APIError{
 			Code:      http.StatusInternalServerError,
 			ClientMsg: "Failed to place sandbox",
--- a/packages/api/internal/orchestrator/create_instance.go
+++ b/packages/api/internal/orchestrator/create_instance.go
@@ -359,7 +359,7 @@ func (o *Orchestrator) CreateSandbox(
 
 		return sandbox.Sandbox{}, &api.APIError{
 			Code:      http.StatusInternalServerError,
-			ClientMsg: "Failed to place sandbox",
+			ClientMsg: fmt.Sprintf("Failed to place sandbox: %s", err),
 			Err:       fmt.Errorf("failed to place sandbox: %w", err),
 		}
 	}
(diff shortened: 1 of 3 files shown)


leonmeijer:
This change is already incorporated on `main` by 063aa20c871116c80915ab1f5c2e81d804e6b2d0, including typed and sanitized placement-failure reasons plus regression tests. Merging this branch would duplicate the current implementation, so this PR can be closed as superseded.

PR #3558, opened 2026-08-13
fix(orchestrator): close TOCTOU race in Cleanup.Add/AddPriority

### LINKED ISSUE #3557: bug(orchestrator): Cleanup.Add has TOCTOU race — cleanup functions silently dropped after Run()
## Summary

`Cleanup.Add` and `Cleanup.AddPriority` have a TOCTOU (time-of-check/time-of-use) race against `Cleanup.run` that silently discards cleanup functions without executing them, causing permanent resource leaks.

## Root Cause

In `packages/orchestrator/pkg/sandbox/cleanup.go`:

**`run()` sets `hasRun` before acquiring the lock (line 80 vs 82):**
```go
func (c *Cleanup) run(ctx context.Context) {
    c.hasRun.Store(true)   // ← written outside the lock
    c.mu.Lock()
    defer c.mu.Unlock()
    // ... drains c.cleanup and c.priorityCleanup
}
```

**`Add()` reads `hasRun` before acquiring the lock (line 40 vs 49):**
```go
func (c *Cleanup) Add(ctx context.Context, f func(ctx context.Context) error) {
    if c.hasRun.Load() == true {   // ← read outside the lock
        // run f immediately
        return
    }
    c.mu.Lock()
    defer c.mu.Unlock()
    c.cleanup = append(c.cleanup, f)   // ← appended after run() may have drained
}
```

## Race Window

```
Goroutine A (Add):    hasRun.Load() == false  →  [not yet locked]
Goroutine B (run):    hasRun.Store(true)  →  Lock()  →  drain all cleanup  →  Unlock()
Goroutine A (Add):    Lock()  →  append(f)   ← f is now in a drained slice, never executed
```

`sync.Once` in `Run()` prevents `run()` from being called a second time, so `f` is permanently lost — no log, no error, no signal.

## Impact

The `Cleanup` type is the central resource-release mechanism for Firecracker sandbox lifecycle. The package has **28 call sites** across sandbox startup (`sandbox.go`, `resume.go`, `reboot.go`, etc.) registering operations such as:

- Closing overlay filesystems
- Releasing network slots
- Removing Firecracker and UFFD socket files
- Stopping the Firecracker process

Any one of these silently dropped during a concurrent error-path teardown leaves a leaked resource on the host until the orchestrator restarts.

## Fix

Move `hasRun.Store(true)` inside the lock in `run()`, and add a double-check in `Add()` / `AddPriority()` after acquiring `mu`:

```go
func (c *Cleanup) Add(ctx context.Context, f func(ctx context.Context) error) {
    // Optimistic fast path.
    if c.hasRun.Load() {
        err := f(context.WithoutCancel(ctx))
        if err != nil {
            logger.L().Error(ctx, "failed to run function after cleanup has run", zap.Error(err))
        }
        return
    }

    c.mu.Lock()
    // Double-check: run() may have completed between the Load above and here.
    if c.hasRun.Load() {
        c.mu.Unlock()
        err := f(context.WithoutCancel(ctx))
        if err != nil {
            logger.L().Error(ctx, "failed to run function after cleanup has run", zap.Error(err))
        }
        return
    }
    c.cleanup = append(c.cleanup, f)
    c.mu.Unlock()
}

func (c *Cleanup) run(ctx context.Context) {
    c.mu.Lock()
    defer c.mu.Unlock()

    c.hasRun.Store(true)  // now set under the lock — closes the race window

    // ... rest unchanged
}
```

The same double-check applies to `AddPriority`.

AdaAibaby:
## Problem

Closes #3557

`Cleanup.Add` and `Cleanup.AddPriority` had a TOCTOU race against `Cleanup.run` that silently discarded cleanup functions, causing permanent resource leaks on the host.

**The race window:**

```
Goroutine A (Add):   hasRun.Load() == false  →  [not yet holding mu]
Goroutine B (run):   hasRun.Store(true)  →  Lock()  →  drain slice  →  Unlock()
Goroutine A (Add):   Lock()  →  append(f)   ← f is in a drained slice, never executed
```

`sync.Once` in `Run()` prevents `run()` from firing a second time, so `f` is permanently lost — no log, no error, no signal.

With **28 call sites** in the sandbox startup path registering operations like closing overlay filesystems, releasing network slots, removing Firecracker/UFFD sockets, and stopping the Firecracker process, any one of these dropped in a concurrent teardown leaves a leaked resource until the orchestrator restarts.

## Fix

Two changes to `cleanup.go`:

**1. Move `hasRun.Store(true)` inside the lock in `run()`**

```go
// before
func (c *Cleanup) run(ctx context.Context) {
    c.hasRun.Store(true)   // ← outside lock
    c.mu.Lock()
    defer c.mu.Unlock()
    ...
}

// after
func (c *Cleanup) run(ctx context.Context) {
    c.mu.Lock()
    defer c.mu.Unlock()
    c.hasRun.Store(true)   // ← inside lock, closes the window
    ...
}
```

**2. Add a double-check of `hasRun` after acquiring `mu` in `Add`/`AddPriority`**

```go
c.mu.Lock()
if c.hasRun.Load() {   // ← double-check: catches goroutines that lost the race
    c.mu.Unlock()
    err := f(context.WithoutCancel(ctx))
    ...
    return
}
defer c.mu.Unlock()
c.cleanup = append(c.cleanup, f)
```

The optimistic fast-path `Load()` before the lock is kept to avoid lock contention in the common post-run case (already-torn-down sandbox), but is no longer the only gate.

## Test plan

- [ ] Existing sandbox lifecycle tests pass
- [ ] Manual: concurrent `Add` and `Run` calls no longer drop functions (race detector: `go test -race ./packages/orchestrator/pkg/sandbox/...`)

### DIFF
--- a/packages/orchestrator/pkg/sandbox/cleanup.go
+++ b/packages/orchestrator/pkg/sandbox/cleanup.go
@@ -37,7 +37,7 @@ func (c *Cleanup) AddNoContext(ctx context.Context, f func() error) {
 }
 
 func (c *Cleanup) Add(ctx context.Context, f func(ctx context.Context) error) {
-	if c.hasRun.Load() == true {
+	if c.hasRun.Load() {
 		err := f(context.WithoutCancel(ctx))
 		if err != nil {
 			logger.L().Error(ctx, "failed to run function after cleanup has run", zap.Error(err))
@@ -47,13 +47,25 @@ func (c *Cleanup) Add(ctx context.Context, f func(ctx context.Context) error) {
 	}
 
 	c.mu.Lock()
+	// Double-check under lock: run() may have completed between the Load above
+	// and acquiring mu, silently dropping f into an already-drained slice.
+	if c.hasRun.Load() {
+		c.mu.Unlock()
+
+		err := f(context.WithoutCancel(ctx))
+		if err != nil {
+			logger.L().Error(ctx, "failed to run function after cleanup has run", zap.Error(err))
+		}
+
+		return
+	}
 	defer c.mu.Unlock()
 
 	c.cleanup = append(c.cleanup, f)
 }
 
 func (c *Cleanup) AddPriority(ctx context.Context, f func(ctx context.Context) error) {
-	if c.hasRun.Load() == true {
+	if c.hasRun.Load() {
 		err := f(context.WithoutCancel(ctx))
 		if err != nil {
 			logger.L().Error(ctx, "failed to run priority function after cleanup has run", zap.Error(err))
@@ -63,6 +75,18 @@ func (c *Cleanup) AddPriority(ctx context.Context, f func(ctx context.Context) e
 	}
 
 	c.mu.Lock()
+	// Double-check under lock: run() may have completed between the Load above
+	// and acquiring mu, silently dropping f into an already-drained slice.
+	if c.hasRun.Load() {
+		c.mu.Unlock()
+
+		err := f(context.WithoutCancel(ctx))
+		if err != nil {
+			logger.L().Error(ctx, "failed to run priority function after cleanup has run", zap.Error(err))
+		}
+
+		return
+	}
 	defer c.mu.Unlock()
 
 	c.priorityCleanup = append(c.priorityCleanup, f)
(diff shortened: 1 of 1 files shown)


leonmeijer:
This duplicates #3136 but is the smaller, current-main, issue-linked patch. I recommend keeping this PR as the single implementation and closing #3136 after carrying over any useful concurrency regression case.

PR #3560, opened 2026-08-13
fix(envd): guard MMDS-poll goroutine with CAS to prevent unbounded accumulation on /init retries

### LINKED ISSUE #3559: bug(envd): PostInit spawns unbounded MMDS-poll goroutines — one per /init retry
## Summary

PostInit unconditionally spawns a new goroutine on every call to poll MMDS. The orchestrator drives /init via an infinite retry loop, so when MMDS is not immediately reachable, goroutines accumulate inside the microVM without bound.

## Root Cause

packages/envd/internal/api/init.go lines 209-213:

go func() {
    ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
    defer cancel()
    host.PollForMMDSOpts(ctx, a.mmdsChan, a.defaults.EnvVars)
}()

No guard prevents concurrent goroutine accumulation. initLock serialises body processing but the goroutine spawns after initLock.Release(1) so serialisation does not protect the spawn.

PollForMMDSOpts creates its own http.Client with DisableKeepAlives and ticks every 50ms issuing 2 HTTP requests per tick. Each goroutine lives up to 60 seconds.

## Impact

At 100ms retry interval, 60 seconds of MMDS unavailability produces 600 concurrent goroutines, each making 40 connection-attempts/sec to 169.254.169.254 — 24000 TCP connections/sec total. This exhausts the VM ephemeral port range and can crash the envd control plane mid-resume.

## Fix

Add an atomic.Bool guard to API struct and use CompareAndSwap in PostInit:

if a.mmdsPollRunning.CompareAndSwap(false, true) {
    go func() {
        defer a.mmdsPollRunning.Store(false)
        ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
        defer cancel()
        host.PollForMMDSOpts(ctx, a.mmdsChan, a.defaults.EnvVars)
    }()
}

Caps concurrent MMDS-poll goroutines at 1 while preserving re-poll across resume cycles.

AdaAibaby:
## Problem

Closes #3559

`PostInit` unconditionally spawns a new `PollForMMDSOpts` goroutine on **every call**:

```go
go func() { //nolint:contextcheck // TODO: fix this later
    ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
    defer cancel()
    host.PollForMMDSOpts(ctx, a.mmdsChan, a.defaults.EnvVars)
}()
```

The orchestrator retries `/init` with infinite backoff until the sandbox responds. During normal Firecracker boot MMDS fills asynchronously, so several retries arrive before MMDS is ready. Each goroutine:

- lives for up to **60 seconds**
- calls MMDS every **50 ms** for both token and opts = up to **1 200 HTTP requests**

With N concurrent retries this produces **O(N) goroutine stacks** and **O(N × 1200) MMDS connections** accumulating silently inside the VM until they time out.

## Fix

Add `mmdsPollRunning atomic.Bool` to `API` (alongside the existing `isMountingNFS` and `initialized` atomics) and gate the goroutine spawn with `CompareAndSwap`:

```go
if a.mmdsPollRunning.CompareAndSwap(false, true) {
    go func() { //nolint:contextcheck
        ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
        defer cancel()
        host.PollForMMDSOpts(ctx, a.mmdsChan, a.defaults.EnvVars)
        a.mmdsPollRunning.Store(false)  // reset so next /init can re-poll if needed
    }()
}
```

**Behaviour preserved:**
- First `/init` — CAS succeeds, goroutine starts, MMDS opts delivered normally.
- Subsequent retries while goroutine is running — CAS fails, no new goroutine; the already-running goroutine delivers the opts.
- If goroutine times out before MMDS is ready — flag resets to `false`, the next `/init` retry starts a fresh poll.

## Changes

| File | Change |
|------|--------|
| `packages/envd/internal/api/store.go` | Add `mmdsPollRunning atomic.Bool` field with explaining comment |
| `packages/envd/internal/api/init.go` | Wrap goroutine spawn in `CompareAndSwap`; reset flag on exit |

## Testing

The fix is a pure atomicity change on an existing `atomic.Bool` pattern already used for `isMountingNFS` in the same struct. No new dependencies.

### DIFF
--- a/packages/envd/internal/api/init.go
+++ b/packages/envd/internal/api/init.go
@@ -206,11 +206,19 @@ func (a *API) PostInit(w http.ResponseWriter, r *http.Request) {
 		a.initialized.Store(true)
 	}
 
-	go func() { //nolint:contextcheck // TODO: fix this later
-		ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
-		defer cancel()
-		host.PollForMMDSOpts(ctx, a.mmdsChan, a.defaults.EnvVars)
-	}()
+	// Only one MMDS-poll goroutine may be in flight at a time. Without this
+	// guard every /init retry (the orchestrator retries with infinite backoff
+	// until the sandbox responds) spawns a fresh goroutine that lives for 60 s
+	// and fires up to 1 200 HTTP requests — O(N) goroutines and O(N*1200)
+	// connections accumulate silently until they time out.
+	if a.mmdsPollRunning.CompareAndSwap(false, true) {
+		go func() { //nolint:contextcheck
+			ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
+			defer cancel()
+			host.PollForMMDSOpts(ctx, a.mmdsChan, a.defaults.EnvVars)
+			a.mmdsPollRunning.Store(false)
+		}()
+	}
 
 	w.Header().Set("Cache-Control", "no-store")
 
--- a/packages/envd/internal/api/store.go
+++ b/packages/envd/internal/api/store.go
@@ -66,6 +66,13 @@ type API struct {
 	// re-adopted (possibly hostile) guest process can neither drive an upgrade
 	// nor be handed a running workload before /init has re-established auth.
 	initialized atomic.Bool
+
+	// mmdsPollRunning is set to true while a PollForMMDSOpts goroutine is
+	// in flight. CompareAndSwap on this flag prevents PostInit from spawning
+	// a new polling goroutine on every /init retry — each goroutine sends up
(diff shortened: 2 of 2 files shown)


leonmeijer (commented):
packages/envd/internal/api/init.go changes retry and MMDS polling behavior but leaves packages/envd/pkg/version.go at 0.6.13. The repository requires every envd behavior change to bump that version; otherwise X-Envd-Version and rollout compatibility logic cannot distinguish binaries with and without this fix. Please bump the envd version in this PR.

PR #3564, opened 2026-08-13
fix(envd): cap PostInit request body at 1 MiB with MaxBytesReader

### LINKED ISSUE #3563: bug(envd): PostInit reads request body with io.ReadAll and no size limit, enabling OOM via oversized payload
## Summary

`PostInit` in `packages/envd/internal/api/init.go` reads the request body with an unbounded `io.ReadAll`:

```go
// packages/envd/internal/api/init.go ~L141
body, err := io.ReadAll(r.Body)  // no MaxBytesReader
defer memguard.WipeBytes(body)
```

Any process inside the VM that can reach the envd HTTP port can send an arbitrarily large body, allocating heap memory until envd is OOM-killed.

## Why it matters

envd is the **sandbox control plane** — it manages file I/O, process execution, cgroup freezing, NFS mounts, and live-upgrade handover. An OOM kill of envd:

- leaves all user processes running but orphaned (no envd to receive commands)
- prevents graceful sandbox teardown (cleanup callbacks, slot release)
- breaks any in-progress pause/resume sequence, potentially corrupting snapshot state

The `/init` endpoint is excluded from `authExcludedPaths` but **does accept unauthenticated bodies** — the auth check happens *after* the body is fully read. A guest process (e.g. user code running inside the sandbox) that discovers the envd port can therefore trigger OOM without any credentials.

### Secondary issue: `memguard.WipeBytes` security value is reduced

`memguard.WipeBytes(body)` is deferred to scrub the access token from heap memory. But if `body` contains the token and is first replicated into a large allocation (e.g. a 500 MiB body), Go's allocator may have already copied the slice header or the GC may have paged parts to disk before the wipe runs. Capping the body to a size where the entire buffer fits comfortably in memory preserves the intent of the wipe.

## Fix

Wrap `r.Body` with `http.MaxBytesReader` before reading. The largest legitimate `/init` payload contains `EnvVars` (many vars) and `CaBundle` (multiple PEM certs); 1 MiB is a generous upper bound that no real orchestrator payload will approach:

```go
if r.Body != nil {
    // Cap body to 1 MiB. /init carries credentials and CA bundles but no
    // bulk data. An unbounded io.ReadAll lets a guest process OOM envd,
    // the sandbox control plane, by sending an oversized body.
    r.Body = http.MaxBytesReader(w, r.Body, 1<<20)
    body, err := io.ReadAll(r.Body)
    defer memguard.WipeBytes(body)
    if err != nil {
        var maxErr *http.MaxBytesError
        if errors.As(err, &maxErr) {
            logger.Error().Msg("request body exceeds 1 MiB limit")
            w.WriteHeader(http.StatusRequestEntityTooLarge)
        } else {
            logger.Error().Msgf("Failed to read request body: %v", err)
            w.WriteHeader(http.StatusBadRequest)
        }
        return
    }
    ...
```

No changes needed to imports (`errors`, `io`, `net/http` are all already imported).

## Severity

Medium. The envd HTTP port is not externally exposed; exploitation requires code execution inside the VM. However sandboxes are explicitly designed to run untrusted LLM-generated code, so "arbitrary code inside the VM" is the normal threat model, not an unusual escalation.

AdaAibaby:
## Problem

Closes #3563

`PostInit` read the full request body with no size cap:

```go
body, err := io.ReadAll(r.Body)  // no MaxBytesReader
defer memguard.WipeBytes(body)
```

Any process inside the VM that can reach the envd HTTP port can send an arbitrarily large body, allocating heap memory until envd is OOM-killed. The auth check happens *after* the full body is read, so no credentials are required.

envd is the **sandbox control plane** — OOM-killing it orphans all user processes, breaks graceful teardown (slot release, cleanup callbacks), and can corrupt pause/resume state.

**Secondary issue:** `memguard.WipeBytes(body)` is intended to scrub the access token from heap memory after use. If `body` is a multi-hundred-MiB allocation, Go's GC may have already paged parts of it to disk or the allocator may have moved the backing array before the deferred wipe runs, reducing the security guarantee.

## Fix

Wrap `r.Body` with `http.MaxBytesReader(w, r.Body, 1<<20)` before `io.ReadAll`. 1 MiB is a generous upper bound — the largest legitimate `/init` payload carries `EnvVars` and `CaBundle` (multiple PEM certs) but no bulk data.

Oversized requests now return **413 Request Entity Too Large** via a dedicated `*http.MaxBytesError` branch; all other read errors keep the existing 400 path.

## Changes

Single file, `packages/envd/internal/api/init.go` (+17/-2 lines):

| Change | Detail |
|--------|--------|
| `r.Body = http.MaxBytesReader(w, r.Body, 1<<20)` | Cap before `io.ReadAll` |
| `errors.As(err, &maxErr)` branch | Return 413 on oversized body |
| Existing 400 path preserved | All other read errors unchanged |

No import changes needed (`errors`, `io`, `net/http` already imported).

## Testing

- Normal orchestrator `/init` payloads are well under 1 MiB — no behaviour change for legitimate callers.
- A request with a body > 1 MiB now gets 413 and envd memory is unaffected.

### DIFF
--- a/packages/envd/internal/api/init.go
+++ b/packages/envd/internal/api/init.go
@@ -137,13 +137,28 @@ func (a *API) PostInit(w http.ResponseWriter, r *http.Request) {
 	logger := a.logger.With().Str(string(logs.OperationIDKey), operationID).Logger()
 
 	if r.Body != nil {
+		// Cap the body to 1 MiB before reading. The largest legitimate /init
+		// payload contains EnvVars and a CaBundle, but no bulk data; 1 MiB is
+		// a generous upper bound that no real orchestrator payload approaches.
+		// Without this limit a guest process can send an unbounded body and
+		// OOM-kill envd, the sandbox control plane, without any credentials.
+		// Capping also preserves the intent of memguard.WipeBytes below: if
+		// the body were huge, partial GC scans could page the secret to disk
+		// before the wipe runs.
+		r.Body = http.MaxBytesReader(w, r.Body, 1<<20)
 		// Read raw body so we can wipe it after parsing
 		body, err := io.ReadAll(r.Body)
 		// Ensure body is wiped after we're done
 		defer memguard.WipeBytes(body)
 		if err != nil {
-			logger.Error().Msgf("Failed to read request body: %v", err)
-			w.WriteHeader(http.StatusBadRequest)
+			var maxErr *http.MaxBytesError
+			if errors.As(err, &maxErr) {
+				logger.Error().Msg("request body exceeds 1 MiB limit")
+				w.WriteHeader(http.StatusRequestEntityTooLarge)
+			} else {
+				logger.Error().Msgf("Failed to read request body: %v", err)
+				w.WriteHeader(http.StatusBadRequest)
+			}
 
 			return
 		}


leonmeijer (commented):
The new 1 MiB limit rejects an envd init payload that is still valid under the public sandbox contract: spec/openapi.yml does not cap envVars, while the orchestrator serializes the complete env map and CA bundle into POST /init and treats a 413 as a fatal create or resume failure. Add an equivalent admission limit at the public API boundary or choose a bound that is already guaranteed there so a successfully admitted sandbox cannot fail only at envd initialization.

This envd behavior change also leaves packages/envd/pkg/version.go at 0.6.13. Without the required version bump, X-Envd-Version and compatibility or rollout logic cannot distinguish nodes that enforce the new body limit.

Issue #3566, opened 2026-08-14
sandbox/storage: stale team-index SET entries accumulate when sandbox keys are lost without going through Remove()


AdaAibaby:
## Summary

The per-team sandbox index (`sandbox:storage:{teamID}:index`, a Redis SET) accumulates stale entries that are never cleaned up when sandbox keys disappear outside the normal `Remove()` path.

## Root cause

`ExpiredItems()` sweeps orphaned ZSET members (entries in `sandbox:storage:global:expiration` whose corresponding sandbox key is gone), but only removes the ZSET member — it does **not** remove the sandboxID from the team index SET:

```go
// items.go — current behaviour
if raw == nil {
    staleMembers = append(staleMembers, ref.member)  // ZREM the ZSET member ✓
    orphanCount++
    continue  // team index entry is NOT cleaned ✗
}
```

No other code path cleans the team index for absent sandbox keys:
- `TeamItems` / `fetchSandboxBatch` silently skip `nil` MGET results with a `continue`
- The healer (`healExpirationIndex`) only fills missing ZSET members, never prunes the team index
- `Remove()` / `removeSandboxScript` cleans both atomically, but only runs on the normal removal path

The team index SET therefore grows without bound whenever sandbox keys disappear externally — most commonly via Redis `maxmemory` key eviction or a Redis restart with incomplete persistence.

## Observed impact

Teams with high sandbox churn (frequent create / pause / resume cycles) combined with Redis key eviction can accumulate thousands of stale SET members, inflating Redis memory and slowing `TeamItems` (SMEMBERS + MGET fan-out is proportional to SET size, including stale entries).

## Trigger conditions

- Redis `maxmemory` policy that can evict individual sandbox keys (`allkeys-lru`, `allkeys-random`, etc.)
- Redis restart with RDB/AOF disabled or incomplete
- Any path that deletes a sandbox key without calling `sandboxStore.Remove()`

Frequent `autoPause` / `autoResume` amplifies the issue by increasing per-team sandbox churn and Redis memory pressure, raising the probability of key eviction.

## Fix

When `ExpiredItems` finds an orphaned ZSET member (MGET returned nil), also `SREM` the sandboxID from the team index. MGET already confirmed the sandbox key is absent, so we cannot unindex a live sandbox. A concurrent `Add` that races the SREM will `SADD` the sandboxID back immediately, so the worst outcome is a brief gap in `TeamItems` results for one eviction cycle.

PR #3567, opened 2026-08-14
fix(api): prune stale team index entries when ZSET orphans are swept

AdaAibaby:
https://github.com/e2b-dev/infra/issues/3566
When ExpiredItems found an orphaned ZSET member (sandbox key already gone), it removed the ZSET member but left the sandboxID sitting in the per-team SET index. These stale entries accumulate indefinitely: TeamItems silently skips nil MGET results, the healer only fills missing ZSET members, and Remove only cleans the team index on the normal removal path.

Fix: when an orphaned ZSET member is confirmed absent (nil MGET result), SREM the sandboxID from the team index in the same sweep. The MGET has already confirmed the key is gone, so we cannot accidentally unindex a live sandbox. A concurrent Add that writes the key after our MGET will SADD the sandboxID back, so the worst outcome is a brief gap in TeamItems results for one eviction cycle.

The memberRef struct gains a teamID field so the orphan handler has the context it needs without restructuring the surrounding loop.

Also replace the addSandboxScript and removeSandboxScript Lua scripts with TxPipelined: they contained no conditional logic that required Lua atomicity, and MULTI/EXEC is sufficient for these cluster-safe key pairs.

### DIFF
--- a/packages/api/internal/sandbox/storage/redis/items.go
+++ b/packages/api/internal/sandbox/storage/redis/items.go
@@ -102,13 +102,11 @@ func (s *Storage) ExpiredItems(ctx context.Context) ([]sandboxtypes.Sandbox, err
 		for i, raw := range batch.cmd.Val() {
 			ref := batch.refs[i]
 
-			// Sandbox key gone but ZSET entry remains — orphaned. Members are
-			// execution-scoped, so this removal can never unindex a fresh
-			// execution concurrently re-added under the same sandbox ID.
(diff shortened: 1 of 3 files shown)


leonmeijer (commented):
ExpiredItems performs the missing-key MGET and the team-index SREM as separate operations. A resume can Add a new execution under the same sandbox ID between them, after which this cleanup removes the live sandbox from its team index permanently. Recheck key absence and remove the index member atomically in one same-slot transaction or script.

PR #3570, opened 2026-08-14
feat(api): fall back to node gRPC when Redis is unavailable for sandbox listing

### LINKED ISSUE #3569: feat(api): fallback to orchestrator-node gRPC when Redis is unavailable for sandbox listing
## Problem

When Redis is unavailable or unresponsive, `GET /v2/sandboxes` returns 500 and running sandboxes are completely invisible to users. There is no degraded-mode fallback.

### Call chain

```
GET /v2/sandboxes
  └─ GetV2Sandboxes handler
       └─ orchestrator.GetSandboxes(ctx, teamID, states)
            └─ sandboxStore.TeamItems(ctx, teamID, states)
                 └─ Redis: SMEMBERS sandbox:storage:{teamID}:index
                           MGET   sandbox:{teamID}:{sandboxID}...
                           ↑ blocks / errors → 500, no data
```

Running sandbox state lives **only** in Redis. If Redis hangs, the entire listing path is dead.

## Ground truth already exists — just not exposed

`keepInSync` already queries every orchestrator node via gRPC every 20 seconds:

```
Node.Sync()
  └─ n.GetSandboxes(ctx)   // gRPC client.Sandbox.List — zero Redis dependency
       └─ store.Reconcile(ctx, activeInstances, n.ID)
```

Each orchestrator node holds the authoritative list of VMs it is running. This path is used only for internal reconciliation today; it is not reachable from the list API.

## Paused sandboxes are unaffected

`getPausedSandboxes` queries PostgreSQL (`GetSnapshotsWithCursor`) and is completely independent of Redis. Only running sandboxes are impacted.

## Proposed fix

In `orchestrator.GetSandboxes`, when the Redis `TeamItems` call returns an error (or exceeds a short deadline), fall back to aggregating results from `o.nodes` via `node.GetSandboxes()`, filtering by `teamID`:

```go
func (o *Orchestrator) GetSandboxes(ctx context.Context, teamID uuid.UUID, states []sandbox.State) ([]sandbox.Sandbox, error) {
    sandboxes, err := o.sandboxStore.TeamItems(ctx, teamID, states)
    if err == nil {
        return sandboxes, nil
    }

    // Redis unavailable — fall back to querying nodes directly
    logger.L().Warn(ctx, "Redis unavailable, falling back to node gRPC for sandbox list", zap.Error(err))
    return o.listSandboxesFromNodes(ctx, teamID, states)
}
```

`listSandboxesFromNodes` fans out `node.GetSandboxes()` across `o.nodes.Items()` in parallel, filters by `teamID`, and deduplicates by `sandboxID`.

## Trade-offs

| | Redis path | Node gRPC fallback |
|---|---|---|
| Latency | ~1 ms | ~node-count × gRPC RTT (parallel) |
| Consistency | Strong (single store) | Best-effort (per-node snapshot) |
| Availability | Fails with Redis | Works as long as ≥1 node answers |

The fallback result is slightly less consistent (a sandbox mid-migration may appear on two nodes briefly) but is far better than a hard 500.

AdaAibaby:
Closes #3569

## Problem

`GET /v2/sandboxes` is entirely Redis-dependent for running sandboxes. When Redis hangs or becomes unresponsive, `TeamItems` blocks until the client timeout and returns an error, causing a hard 500 with no data. Paused sandboxes (served from PostgreSQL) are unaffected.

## Solution

When `TeamItems` returns any error, `GetSandboxes` falls back to fanning out gRPC `Sandbox.List` calls to every connected orchestrator node in parallel. This is the exact same data source `keepInSync` already uses every 20 s for reconciliation — it bypasses Redis entirely and queries the nodes that actually own the VMs.

### Fallback behaviour

| Scenario | Result |
|---|---|
| Redis healthy | Redis result, no change |
| Redis error, ≥1 node responds | Merged, filtered, deduplicated node results |
| Redis error, all nodes fail | Error (same outcome as before, clearer message) |
| Redis error, no nodes connected | Error |

Per-node failures are logged as warnings and skipped so a single bad node cannot suppress results from healthy ones.

### Deduplication

A sandbox mid-migration may transiently appear on two nodes. Results are deduplicated by `sandboxID` before being returned.

## Code changes

**`list_instances.go`**
- `GetSandboxes`: try Redis first; on error log a warning and call `getSandboxesFromNodes`
- `getSandboxesFromNodes`: parallel fan-out over `o.nodes`, filter by teamID + states, deduplicate
- `sandboxNodeLister` interface: minimal interface satisfied by `*nodemanager.Node`, makes the inner function unit-testable without a live gRPC server

**`list_instances_test.go`** (new)
- `TestGetSandboxesFromNodes_NoNodes` — returns error when node list is empty
- `TestGetSandboxesFromNodes_AllNodesFail` — returns error when every node fails
- `TestGetSandboxesFromNodes_FiltersByTeam` — cross-team sandbox is excluded
- `TestGetSandboxesFromNodes_FiltersByState` — wrong-state sandbox is excluded
- `TestGetSandboxesFromNodes_DeduplicatesAcrossNodes` — duplicate sandboxID appears once
- `TestGetSandboxesFromNodes_PartialNodeFailure` — one failing node does not suppress results from the other

## Trade-offs

The fallback result is slightly less consistent than Redis: a sandbox mid-migration can appear on two nodes briefly (handled by deduplication) and very recent state changes written only to Redis may not be reflected. This is acceptable given the alternative is a hard 500.

### DIFF
--- a/packages/api/internal/orchestrator/list_instances.go
+++ b/packages/api/internal/orchestrator/list_instances.go
@@ -56,8 +56,16 @@ func (o *Orchestrator) GetSandboxes(ctx context.Context, teamID uuid.UUID, state
 // in parallel, filters by teamID and states, and deduplicates by sandboxID (a
 // sandbox mid-migration may transiently appear on two nodes).
 //
-// Per-node errors are logged and skipped. If every node fails, an error is
-// returned.
+// The function is fail-closed: if any node returns an error the whole call
+// fails. This is intentional — callers such as template deletion and admin
+// team-kill make safety decisions from the result, and a partial listing could
+// miss sandboxes on the failed node, leading to incorrect actions.
+//
+// Note: node.GetSandboxes reconstructs sandbox records via NewSandbox, which
+// always sets State = StateRunning. Sandboxes that are mid-transition
+// (StatePausing, StateSnapshotting) are returned as StateRunning during the
+// fallback. State filtering therefore works correctly only for callers that
+// include StateRunning in their allowed set, which all current callers do.
 func getSandboxesFromNodes(ctx context.Context, nodes []namedLister, teamID uuid.UUID, states []sandboxtypes.State) ([]sandbox.Sandbox, error) {
 	if len(nodes) == 0 {
 		return nil, fmt.Errorf("no orchestrator nodes connected")
@@ -83,17 +91,15 @@ func getSandboxesFromNodes(ctx context.Context, nodes []namedLister, teamID uuid
 
 	seen := make(map[string]struct{})
 	var out []sandbox.Sandbox
-	var nodeErrors int
 
 	for i, r := range results {
 		if r.err != nil {
-			nodeErrors++
-			logger.L().Warn(ctx, "Node gRPC fallback: failed to list sandboxes from node",
+			logger.L().Error(ctx, "Node gRPC fallback: failed to list sandboxes from node — aborting to avoid partial result",
 				zap.Error(r.err),
 				logger.WithNodeID(nodes[i].id),
 			)
 
-			continue
+			return nil, fmt.Errorf("node %s failed to respond during Redis fallback: %w", nodes[i].id, r.err)
 		}
 
 		for _, sbx := range r.sandboxes {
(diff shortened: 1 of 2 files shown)


jakubno:
Closing as mentioned in the issue

PR #3587, opened 2026-08-18
fix(fc): slot-driven guest IPv6 config; default to IPv4-only to eliminate Happy Eyeballs latency

### LINKED ISSUE #3585: kernel: ipv6.disable=0 with no IPv6 routing causes Happy Eyeballs latency on all outbound sandbox connections
## Symptom

Every outbound connection from a sandbox to an external host that has AAAA records takes 250 ms or more longer than expected. The extra latency appears on the **first TCP connection per destination** — pip install, npm install, HTTP API calls, curl, etc.

```bash
# inside sandbox
time curl -s https://pypi.org -o /dev/null
# real    0m0.582s   (should be ~0.1s from same region)
```

The delay is consistent and reproducible; once the TCP connection is established, transfer speed is normal.

## Root cause

**`packages/orchestrator/pkg/sandbox/fc/kernel_args.go:96-97`** explicitly enables IPv6 inside every Firecracker guest:

```go
"ipv6.disable":  "0",   // IPv6 enabled
"ipv6.autoconf": "1",   // SLAAC enabled
```

However **`packages/orchestrator/pkg/sandbox/fc/process.go:378`** configures the guest network with an IPv4-only `ip=` kernel parameter:

```go
// IPv4 configuration - format: [local_ip]::[gateway_ip]:[netmask]:hostname:iface:dhcp_option:[dns]
ipv4 := fmt.Sprintf("%s::%s:%s:instance:%s:off:%s",
    p.slot.NamespaceIP(), p.slot.TapIPString(), p.slot.TapMaskString(),
    p.slot.VpeerName(), p.slot.TapName())
```

The host side of the tap device only configures IPv4 NAT rules (`network.go:265`); there is no IPv6 router advertisement (radvd) or IPv6 default route on the host tap interface.

Result:
- IPv6 is active in the guest kernel
- SLAAC runs on the tap interface but finds no IPv6 router → only a link-local `fe80::` address is obtained
- No global unicast IPv6 address, no default IPv6 route
- **All IPv6 traffic is unroutable**, but the protocol stack is fully live

When the guest connects to any dual-stack host, the Linux kernel's address selection (RFC 6724) prefers IPv6:

```
connect("api.openai.com:443")
  → DNS: returns both A + AAAA
  → kernel tries IPv6 AAAA address first
  → no route → EHOSTUNREACH / NDP solicitation times out
  → fallback to IPv4 A record  (~250 ms per RFC 8305)
  → connection finally succeeds
  → every first-connection pays 250 ms+
```

The `/etc/sysctl.conf` written by **`packages/orchestrator/pkg/template/build/phases/base/provision.sh`** also has no IPv6 disable entry — only `fs.inotify.max_user_watches` and `vm.compaction_proactiveness` are set — so the guest OS never suppresses IPv6 at the sysctl layer either.

## Why it matters

- **Every sandbox** is affected, regardless of workload
- Any pip, npm, cargo, apt fetch that hits a dual-stack host pays the penalty per connection
- Latency-sensitive API calls (OpenAI, GitHub, GCP, AWS) all have AAAA records
- The penalty compounds: a package install that opens 20 TCP connections silently loses 5+ seconds

## Proposed fix

Change `kernel_args.go:96` to disable IPv6 until the networking layer supports a complete IPv6 stack:

```go
// Before
"ipv6.disable":  "0",
"ipv6.autoconf": "1",

// After
"ipv6.disable": "1",
// ipv6.autoconf is redundant when disable=1, remove it
```

This is a single-line kernel cmdline change with no other code impact. If full IPv6 support is planned in the future, the correct order is: first configure IPv6 routing on the host tap (RA, prefix delegation, ip6tables), then set `ipv6.disable=0`.

AdaAibaby:
Fixes #3585.

## Problem

`kernel_args.go` hardcoded `ipv6.disable=0` and `ipv6.autoconf=1`, enabling IPv6 inside every Firecracker guest. However the host-side network setup only configures IPv4 — no router advertisement is sent on the host tap interface, so SLAAC produces only an unroutable
`fe80::` link-local address.

With IPv6 nominally enabled but unroutable, the Linux kernel's address selection (RFC 6724) prefers AAAA records. On every outbound connection to a dual-stack host the kernel tries IPv6 first, hits EHOSTUNREACH or waits for a silent NDP timeout, then falls back to IPv4
under the Happy Eyeballs ~250 ms delay (RFC 8305).

Every sandbox is affected. pip, npm, cargo, apt, and any HTTP API call to a major service (OpenAI, GitHub, PyPI, GCP, AWS) has AAAA records. A package install that opens 20 TCP connections silently loses 5+ seconds.

## Fix

Rather than hardcoding `ipv6.disable=1`, the guest's IPv6 state is now derived from whether the host slot actually has an IPv6 router configured:

- **`Slot.HasIPv6Router() bool`** — single source of truth. Returns `false` today (no RA on the tap); flip to `true` (or read from real IPv6 slot state) when the host gains a full IPv6 stack.
- **`ProcessOptions.IPv6RouterConfigured bool`** — carries the slot's answer into `buildKernelArgs` without coupling the kernel-args layer to the network package. Set from `Slot.HasIPv6Router()` in `Process.Create`.
- **`buildKernelArgs`** — selects `ipv6.disable=0` or `=1` based on the flag. The zero-value default (`false`) preserves safe behavior: guests without an IPv6 router boot IPv4-only.

This means today's sandboxes get `ipv6.disable=1` (fixing the Happy Eyeballs regression), and future dual-stack support requires only flipping `HasIPv6Router()` — no scattered code changes.

## Changes

| File | What changed |
|---|---|
| `network/slot.go` | `HasIPv6Router() bool` — documented extension point, returns `false` |
| `fc/process.go` | `ProcessOptions.IPv6RouterConfigured`; set from `slot.HasIPv6Router()` before `buildKernelArgs` |
| `fc/kernel_args.go` | `ipv6.disable` derived from `options.IPv6RouterConfigured` instead of hardcoded |
| `fc/kernel_args_test.go` | Updated existing cases (expect `ipv6.disable=1`); added `ipv6 router configured` case (expect `ipv6.disable=0`) |
| `docs/ARCHITECTURE.md` | Documents IPv4-only guest default and the dual-stack upgrade path |


blancadesal:
Tested this on a bare-metal server (Debian 13/trixie, kernel 6.12) while trying to run the `DEV-LOCAL.md` local-dev flow end-to-end.

Reproduced the bug on `main`: `apt-get` inside the guest tried an unroutable IPv6 address first (`Cannot initiate the connection to deb.debian.org:80 (2a04:4e42:...)`) before falling back to IPv4.

Cherry-picked this PR onto the same environment, rebuilt the orchestrator, and confirmed a clean build with no IPv6 attempts – `apt-get` went straight to IPv4. 🎉

PR #1705, opened 2026-01-09
feat: memory prefetching

dobrac:
> [!NOTE]
> Introduces build-time memory access mapping and runtime prefetching to reduce sandbox startup latency.
> 
> - Adds `PrefetchTracker` and extends UFFD to record page-fault order/types and expose `PrefetchData`; adds `Prefault` API to proactively map pages
> - New runtime prefetcher (`uffd/prefetch`) that fetches/copies pages in the background during resume; worker counts driven by feature flags
> - New build phase **optimize** collects memory access data by resuming the template multiple times, intersects results, and writes ordered prefetch mapping to template metadata; invalidates template cache after upload
> - Extends metadata schema with `prefetch.memory { indices, access_types, block_size }`
> - Refactors sandbox resume/init with `utils.Promise`; replaces async channel networking with `getNetworkSlot` promise
> - Improves cleanup behavior (late Add executes immediately), makes UFFD ready signaling robust and `fdexit`/UFFD `Close` idempotent; adds `Cache.Invalidate`
> - Adds tests for promise and prefetch mapping; minor cleanup (remove unused `NewTrackerFromBitset`)
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit be2a21c8fd724f7b1d03612d82fa0721261d89cb. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #3639, opened 2026-09-15
fix(build): gate the optimize phase's metadata re-upload on finalize's uploads

JooKS-me:
## Problem

The optimize phase (`packages/orchestrator/pkg/template/build/phases/optimize`, from #1705)
runs right after finalize, but finalize's `PauseAndUpload` pushes its uploads — snapshot
data and the pre-prefetch metadata — onto the build's `UploadErrGroup` and returns
immediately. The only wait for that group runs at the builder level, after **every**
phase has completed.

So the optimize phase collects its intersection mapping and calls `updateMetadata` with
no ordering against the finalize goroutine: **two writers race on the same metadata
object for the build**. An interleaving where the finalize upload finishes last clobbers
the just-uploaded prefetch mapping — a silent loss that would manifest as "prefetch
mapping sometimes missing" for freshly built templates, with nothing in the logs.

## Solution

`finalizeUploadsSettled()` gates the optimize phase's metadata re-upload on the build's
`UploadErrGroup`:

- **Normal path**: the wait returns after finalize's uploads settle, then
  `updateMetadata` publishes the prefetch-enriched metadata. The mapping can no longer
  interleave with the finalize upload.
- **Failed upload**: publishing is skipped — the remote build is incomplete, and the
  same error resurfaces at the builder-level `UploadErrGroup.Wait()`, so the build
  still fails loudly rather than dangling a mapping over uploads that never landed.

The gate sits after `collectMemoryPrefetchMapping` (the expensive part), so the wait
overlaps the tail of the finalize upload rather than adding to the critical path.


JooKS-me:
@cla-bot check

PR #3644, opened 2026-09-16
fix(envd): skip MMDS polling in containerized deployments

Luckydog691:
## What was broken

`main.go` guards its startup MMDS poll with `if !isNotFC`, and the `-isnotfc` flag is documented as "run outside of Firecracker (skips MMDS poll and HTTP log exporter)". The `/init` handler still started the same 60s polling loop unconditionally: every attempt opens a TCP connection to `169.254.169.254:80`, which with no MMDS endpoint reachable stays half-open in `SYN_SENT` for the whole poll window.

That dangling connection becomes fatal on checkpoint: a gVisor (runsc) checkpoint taken while it exists panics in the TCP endpoint save path (`resetConnectionLocked` dereferences a nil snd/rcv), so checkpointing fails deterministically for the whole poll window.

## Fix

Mirror the guard `main.go` already applies to its startup poll: skip the MMDS polling in `/init` when `isNotFC` is set. Firecracker-based deployments are unchanged.

## Verification

- In a containerized deployment, the `SYN_SENT` connection to `169.254.169.254:80` is observable for ~60s after `/init` before the change; it never appears after it.
- Checkpointing a runsc sandbox inside that window fails before the change; a checkpoint + restore round trip succeeds after it.
- `go build ./...` and `go test ./internal/api/` pass.

Supersedes #3638. Does not address #3559, which concerns `/init` retries in Firecracker deployments.


Luckydog691:
> We require contributors to sign our Contributor License Agreement, and we don't have @Luckydog691 on file. You can sign our CLA at https://e2b.dev/docs/cla . Once you've signed, post a comment here that says '@cla-bot check'

@cla-bot check

Luckydog691:
> We require contributors to sign our Contributor License Agreement, and we don't have @Luckydog691 on file. You can sign our CLA at https://e2b.dev/docs/cla . Once you've signed, post a comment here that says '@cla-bot check'

@cla-bot check

PR #3649, opened 2026-09-18
fix(storage): truncate local files on overwrite

### LINKED ISSUE #3642: Local filesystem writes can retain stale tail data when overwriting shorter objects
## Problem

The Local FS storage backend opens existing objects with `os.O_RDWR|os.O_CREATE` but without `os.O_TRUNC`:

```go
handle, err := os.OpenFile(o.path, os.O_RDWR|os.O_CREATE, 0o644)
```

`fsObject.Put` writes from offset zero, so overwriting an existing object with shorter content leaves the old tail in the file. The same behavior affects the uncompressed `StoreFile` path that uses `getHandle(false)`.

Example:

```text
existing object: abcdefgh
new object:      xy
read result:     xycdefgh
```

This makes the Local FS backend inconsistent with object-storage backends, where a put replaces the complete object. It can corrupt local snapshots, cache files, or metadata when an object is rewritten with a smaller payload.

## Proposed fix

Use truncation for complete-object writes, either by adding `os.O_TRUNC` to the write handle or by explicitly truncating and resetting the offset before copying:

```go
handle, err := os.OpenFile(o.path, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0o644)
if err != nil {
    return nil, err
}
```

The truncating mode should only be used for complete replacement writes. Read, range-read, append, and multipart/random-offset paths should retain their existing semantics.

## Tests

Please add two separate regression tests:

1. `Put` path:
   - Write a longer payload to an object.
   - Overwrite it with a shorter payload using `Put`.
   - Verify `Size` equals the new payload length.
   - Verify reading returns exactly the new payload, with no stale suffix.

2. Direct, uncompressed `StoreFile` path:
   - Store a longer source file into an object.
   - Overwrite it with a shorter source file using uncompressed `StoreFile`.
   - Verify `Size` equals the new source file length.
   - Verify reading returns exactly the new source content, with no stale suffix.

The compressed `StoreFile` path is a separate implementation and should not be conflated with this regression.

Relevant code:

- `packages/shared/pkg/storage/storage_fs.go`
- `fsObject.Put`
- `fsObject.StoreFile`
- `fsObject.getHandle`

tttboy123:
## Summary

- truncate Local FS objects before complete replacement writes
- clear stale `.uncompressed-size` sidecars after successful uncompressed replacements
- cover `Put` and uncompressed `StoreFile` across shorter, longer, equal-length, and empty replacements
- cover compressed-to-uncompressed replacement through both public write paths

## Why

The Local FS backend opened existing files without `O_TRUNC`. Writing a shorter replacement started at offset zero but retained the previous suffix, unlike object-storage put semantics.

A compressed object also leaves an `.uncompressed-size` sidecar that `Size()` prefers. If the same object is later replaced through an uncompressed path, that sidecar must be removed or `Size()` reports the old value even when the replacement bytes are correct.

The truncation flag remains limited to `getHandle(false)`, whose only callers are complete-object replacement paths. Sidecar cleanup happens only after a successful uncompressed copy. Read/range-read and compressed `StoreFile` behavior remain unchanged.

## Validation

- replacement matrix for `Put` and uncompressed `StoreFile`: shorter, longer, equal-length, and empty
- compressed-to-uncompressed sidecar regression for both write paths
- focused Local FS suite
- race-enabled Local FS suite
- all storage tests not requiring Docker/testcontainers
- package compile-only test
- `golangci-lint run --new-from-rev=upstream/main ./...` from `packages/shared` (0 issues)

The complete storage suite was also attempted; its Docker-backed Azure, GCS, and S3 integration tests cannot start locally because rootless Docker is unavailable.

Closes #3642.


tttboy123:
Follow-up scenario verification found one additional Local FS replacement edge case and it is now included in `b30f27fb9`.

A compressed write creates an `.uncompressed-size` sidecar, and `Size()` prefers that value. Before this follow-up, replacing the same object through `Put` or uncompressed `StoreFile` fixed the bytes but left the old sidecar, so a 5-byte replacement could still report size 63. The new regression reproduced that on both public paths before the fix.

The uncompressed paths now remove the stale sidecar only after a successful copy. I also expanded both replacement regressions to cover shorter, longer, equal-length, and empty payloads.

Fresh verification after the update:
- focused Local FS smoke suite: pass;
- focused Local FS race suite: pass;
- all non-Docker storage tests: pass;
- package compile check: pass;
- golangci-lint: 0 issues;
- git diff check: pass.

The Docker-backed integration tests remain unavailable locally because rootless Docker is not installed.

dobrac (approved):

dobrac:
Thanks for the Local FS overwrite fix! The change has been merged here: https://github.com/e2b-dev/runtime/commit/a334b20e1e4dcae2c7d4f26da839acc1dfbefe5d

PR #3654, opened 2026-09-21
feat(featureflags): read int flag fallbacks from the environment

wei-w-wang:
On a deployment without LAUNCH_DARKLY_API_KEY, every flag resolves to its compiled-in fallback. For the per-node limits that is a hard ceiling only a rebuild can change: "max-sandboxes-per-node" falls back to 200, which is sized for cloud node types, not for the host it runs on.

On a self-hosted node with 384 CPUs and 1.5 TB of memory, the orchestrator refuses the 201st sandbox with max number of running sandboxes on node reached (200), even though the host has capacity for several times that.

## Change

Adds 'envIntOr', the int counterpart of the existing 'envBoolOr', and applies it to the two per-node limits:

- max-sandboxes-per-node, set via MAX_SANDBOXES_PER_NODE` (fallback 200)
- max-starting-instances-per-node, set via MAX_STARTING_INSTANCES_PER_NODE (fallback 3)

The override replaces the *fallback*, not the evaluated value:
- Without a LaunchDarkly key: it's the value the offline store serves.
- With a key: it's the default for flags the environment doesn't define. A defined LaunchDarkly flag still wins, so live configuration is not bypassed.

Unset, empty, unparseable, and non-positive values keep the existing fallback. Rejecting non-positive values means a typo can't silently stop a node from accepting sandboxes, and it matches the "Must be > 0" comment on MaxStartingInstancesPerNode.

Only these two flags opt in; no other behavior changes.

## Testing
- "go test ./pkg/featureflags/... -run TestEnvIntOr -v" and "golangci-lint run ./pkg/featureflags/..." (run from packages/shared) Pass
- Verified on a 384-CPU, 1.5 TB node: with MAX_SANDBOXES_PER_NODE=1000, the orchestrator accepts sandboxes past 200 and the refusal no longer appears.

## Notes
Other int flags could opt in the same way if useful; I kept this to the two limits that block self-hosted capacity.


wei-w-wang:
@cla-bot check

PR #3655, opened 2026-09-22
feat(storage): optional fsync durability + stale temp-file sweep for local FS (follow-up to #3649)

AdaAibaby:
## Summary

Follow-up to #3649 / a334b20, which switched the local FS backend to temp+rename atomic replacement. This adds two robustness hardenings on top of that base:

1. **Optional fsync durability.** `replaceFile` now takes a `durable bool`. When set, it fsyncs the temp file **before** the rename and fsyncs the parent directory **after**. Without this, `rename` can be persisted while the new contents are still only in the page cache — a crash then leaves a *renamed-but-empty/truncated* object (worse than the original stale-tail bug, and against the "put replaces the complete object" semantics #3642 aimed for). Off by default (`STORAGE_FS_FSYNC_DURABLE`, package var so config/tests can toggle without threading a param through every caller) to preserve page-cache throughput for high-frequency cache writes; callers storing crash-critical objects (snapshots, headers) can enable it.

2. **Stale temp-file sweep.** `SweepStaleTempFiles(dir, olderThan)` reclaims orphan `.<object>.tmp-*` files that a crash (OOM kill, power loss) leaves behind between `CreateTemp` and `Rename`. Age-gated so a temp file belonging to an in-flight write is never removed. Shares `tmpFileMarker` with `replaceFile` so the create/sweep naming can never drift.

## Why

- The merged temp+rename removed stale-tail corruption and torn concurrent reads, but `rename` without `fsync` is not crash-durable, and a crashed replacement leaks a temp file. On nodes with OOM-kill history and disk-usage-gated cache eviction, leaked temp files silently consume the quota.

## Validation

- `go test ./pkg/storage/` focused FS/sweep/durable suites: pass
- `-race` on the FS/sweep/durable subset: pass
- `go vet ./pkg/storage/`: clean
- `golangci-lint v2 run --new-from-rev` on the changed files: 0 issues
- Docker-backed S3/GCS/Azure/minio testcontainer tests not run locally (rootless Docker unavailable) — unchanged from #3649.

## Not included (proposed, needs maintainer call)

`SweepStaleTempFiles` is a public utility with no internal caller yet. The natural wiring is a one-shot best-effort sweep of the local cache root at orchestrator/template-manager startup (age gate e.g. 15m). Left out of this PR to keep the blast radius on `pkg/storage` only; happy to wire it into startup if you point me at the preferred hook.

Refs #3642, #3649.


AdaAibaby:
@cla-bot check

